PR #289, opened 2015-07-30
Support pyca/Cryptography

vladimir-v-diaz:
Add support for pyca/cryptography.  `pyca_crypto_key.py` contains all of the functionality provided by `pycrypto_keys.py`, with the exception of generating encrypted files with: 3DES with CBC mode, and PBKDF1+MD5 to strengthen passphrase

* [x] Resolve remaining import issues in unit tests

* [x] Support or drop encrypted PEMs with pyca/cryptography

### DIFF
--- a/tests/test_pyca_crypto_keys.py
+++ b/tests/test_pyca_crypto_keys.py
@@ -84,9 +84,10 @@ class TestPyca_crypto_keys(unittest.TestCase):
                       crypto_keys.create_rsa_signature, '', data)
    
     # Check for invalid 'data'.
-    #crypto_keys.create_rsa_signature(private_rsa, '')
+    self.assertRaises(TypeError,
+                      crypto_keys.create_rsa_signature, private_rsa, '')
     
-    self.assertRaises(tuf.CryptoError,
+    self.assertRaises(tuf.FormatError,
                       crypto_keys.create_rsa_signature, private_rsa, 123)
 
     # Check for missing private key.
@@ -120,9 +121,9 @@ class TestPyca_crypto_keys(unittest.TestCase):
                                                       'invalid_method',
                                                       public_rsa, data)
     
-    # Check for invalid signature and data.
-    self.assertRaises(AttributeError, crypto_keys.verify_rsa_signature, signature,
-                                       method, public_rsa, 123)
+    # Check for invalid 'signature' and 'data' arguments.
+    self.assertRaises(tuf.FormatError, crypto_keys.verify_rsa_signature,
+                      signature, method, public_rsa, 123)
     
     self.assertEqual(False, crypto_keys.verify_rsa_signature(signature, method,
                             public_rsa, b'mismatched data'))
--- a/tests/test_pyca_crypto_keys.py
+++ b/tests/test_pyca_crypto_keys.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python
+
+"""
+<Program Name>
+  test_pyca_crypto_keys.py
+
+<Author> 
+  Vladimir Diaz <vladimir.v.diaz@gmail.com>
+
+<Started>
+  June 3, 2015. 
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Test cases for 'pyca_crypto_keys.py'.
+"""
+
+# Help with Python 3 compatibility, where the print statement is a function, an
+# implicit relative import is invalid, and the '/' operator performs true
+# division.  Example:  print 'hello world' raises a 'SyntaxError' exception.
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+import unittest
+import logging
+
+import tuf
+import tuf.log
+import tuf.formats
+import tuf.pyca_crypto_keys as crypto_keys
+
+logger = logging.getLogger('tuf.test_pyca_crypto_keys')
+
+public_rsa, private_rsa = crypto_keys.generate_rsa_public_and_private()
+FORMAT_ERROR_MSG = 'tuf.FormatError raised.  Check object\'s format.'
+
+
+class TestPyca_crypto_keys(unittest.TestCase):
+  def setUp(self):
+    pass
+
+
+  def test_generate_rsa_public_and_private(self):
+    pub, priv = crypto_keys.generate_rsa_public_and_private()
+    
+    # Check format of 'pub' and 'priv'.
+    self.assertEqual(None, tuf.formats.PEMRSA_SCHEMA.check_match(pub),
+                     FORMAT_ERROR_MSG)
+    self.assertEqual(None, tuf.formats.PEMRSA_SCHEMA.check_match(priv),
+                     FORMAT_ERROR_MSG)
+
+    # Check for invalid bits argument.  bit >= 2048 and a multiple of 256.
+    self.assertRaises(tuf.FormatError,
+                      crypto_keys.generate_rsa_public_and_private, 1024)
+   
+    crypto_keys.generate_rsa_public_and_private(2050)
+
+    self.assertRaises(tuf.FormatError,
+                      crypto_keys.generate_rsa_public_and_private, '2048')
+  
+  
+  
+  def test_create_rsa_signature(self):
+    global private_rsa
+    global public_rsa
+    data = 'The quick brown fox jumps over the lazy dog'.encode('utf-8')
+    signature, method = crypto_keys.create_rsa_signature(private_rsa, data)
+
+    # Verify format of returned values.
+    self.assertNotEqual(None, signature)
+    self.assertEqual(None, tuf.formats.NAME_SCHEMA.check_match(method),
+                     FORMAT_ERROR_MSG)
+    self.assertEqual('RSASSA-PSS', method)
+
+    # Check for improperly formatted arguments.
+    self.assertRaises(tuf.FormatError,
+                      crypto_keys.create_rsa_signature, 123, data)
+    
+    self.assertRaises(TypeError,
+                      crypto_keys.create_rsa_signature, '', data)
+   
+    # Check for invalid 'data'.
+    #crypto_keys.create_rsa_signature(private_rsa, '')
+    
+    self.assertRaises(tuf.CryptoError,
+                      crypto_keys.create_rsa_signature, private_rsa, 123)
+
+    # Check for missing private key.
+    self.assertRaises(tuf.CryptoError,
+                      crypto_keys.create_rsa_signature, public_rsa, data)
+
+
+
+  def test_verify_rsa_signature(self):
+    global public_rsa
+    global private_rsa
+    data = 'The quick brown fox jumps over the lazy dog'.encode('utf-8')
+    signature, method = crypto_keys.create_rsa_signature(private_rsa, data)
+
+    valid_signature = crypto_keys.verify_rsa_signature(signature, method, public_rsa,
+                                                data)
+    self.assertEqual(True, valid_signature)
+
+    # Check for improperly formatted arguments.
+    self.assertRaises(tuf.FormatError, crypto_keys.verify_rsa_signature, signature,
+                                       123, public_rsa, data)
+    
+    self.assertRaises(tuf.FormatError, crypto_keys.verify_rsa_signature, signature,
+                                       method, 123, data)
+    
+    self.assertRaises(tuf.FormatError, crypto_keys.verify_rsa_signature, 123, method,
+                                       public_rsa, data)
+    
+    self.assertRaises(tuf.UnknownMethodError, crypto_keys.verify_rsa_signature,
+                                                      signature,
+                                                      'invalid_method',
+                                                      public_rsa, data)
+    
+    # Check for invalid signature and data.
+    self.assertRaises(tuf.CryptoError, crypto_keys.verify_rsa_signature, signature,
+                                       method, public_rsa, 123)
+    
+    self.assertEqual(False, crypto_keys.verify_rsa_signature(signature, method,
+                            public_rsa, b'mismatched data'))
+
+    mismatched_signature, method = crypto_keys.create_rsa_signature(private_rsa,
+                                                             b'mismatched data')
+    
+    self.assertEqual(False, crypto_keys.verify_rsa_signature(mismatched_signature,
+                            method, public_rsa, data))
+
+
+# Run the unit tests.
+if __name__ == '__main__':
+  unittest.main()
--- a/tests/test_pyca_crypto_keys.py
+++ b/tests/test_pyca_crypto_keys.py
@@ -121,7 +121,7 @@ class TestPyca_crypto_keys(unittest.TestCase):
                                                       public_rsa, data)
     
     # Check for invalid signature and data.
-    self.assertRaises(tuf.CryptoError, crypto_keys.verify_rsa_signature, signature,
+    self.assertRaises(AttributeError, crypto_keys.verify_rsa_signature, signature,
                                        method, public_rsa, 123)
     
     self.assertEqual(False, crypto_keys.verify_rsa_signature(signature, method,
--- a/tuf/formats.py
+++ b/tuf/formats.py
@@ -151,6 +151,9 @@ LOGLEVEL_SCHEMA = SCHEMA.Integer(lo=0, hi=50)
 NAME_SCHEMA = SCHEMA.AnyString()
 NAMES_SCHEMA = SCHEMA.ListOf(NAME_SCHEMA)
 
+# A string representing data.
+DATA_SCHEMA = SCHEMA.AnyString()
+
 # Supported hash algorithms.
 HASHALGORITHMS_SCHEMA = SCHEMA.ListOf(SCHEMA.OneOf(
   [SCHEMA.String('md5'), SCHEMA.String('sha1'),
--- a/tuf/keys.py
+++ b/tuf/keys.py
@@ -149,7 +149,7 @@ _DEFAULT_RSA_KEY_BITS = 3072
 
 # The crypto libraries to use in 'keys.py', set by default or by the user.
 # The following cryptography libraries are currently supported:
-# ['pycrypto', 'pynacl', 'ed25519']
+# ['pycrypto', 'pynacl', 'ed25519', 'pyca-cryptography']
 _RSA_CRYPTO_LIBRARY = tuf.conf.RSA_CRYPTO_LIBRARY
 _ED25519_CRYPTO_LIBRARY = tuf.conf.ED25519_CRYPTO_LIBRARY
 _GENERAL_CRYPTO_LIBRARY = tuf.conf.GENERAL_CRYPTO_LIBRARY
@@ -587,7 +587,7 @@ def check_crypto_libraries(required_libraries):
                                      _available_crypto_libraries:
     message = 'The '+repr(_RSA_CRYPTO_LIBRARY)+' crypto library specified'+ \
       ' in "tuf.conf.RSA_CRYPTO_LIBRARY" could not be imported.'
-    raise tuf.UnsupportedLibraryError(message)
+    raise tuf.UnsupportedLibraryError(message+repr(_available_crypto_libraries))
   
   if 'ed25519' in required_libraries and _ED25519_CRYPTO_LIBRARY not in \
                                          _available_crypto_libraries:
@@ -971,10 +971,15 @@ def import_rsakey_from_encrypted_pem(encrypted_pem, password):
     public, private = \
       tuf.pycrypto_keys.create_rsa_public_and_private_from_encrypted_pem(encrypted_pem,
                                                                          password)
+    public = format_rsakey_from_pem(public)
+    private = format_rsakey_from_pem(private, check_private=True)
+
   elif _RSA_CRYPTO_LIBRARY == 'pyca-cryptography':
     public, private = \
       tuf.pyca_crypto_keys.create_rsa_public_and_private_from_encrypted_pem(encrypted_pem,
                                                                          password)
+    public = format_rsakey_from_pem(public)
+    private = format_rsakey_from_pem(private, check_private=True)
   
   else: #pragma: no cover
     message = 'Invalid crypto library: ' + repr(_RSA_CRYPTO_LIBRARY) + '.'
@@ -1001,7 +1006,7 @@ def import_rsakey_from_encrypted_pem(encrypted_pem, password):
 
 
 
-def format_rsakey_from_pem(pem):
+def format_rsakey_from_pem(pem, check_private=False):
   """
   <Purpose> 
     Generate an RSA key object from 'pem'.  In addition, a keyid identifier for
@@ -1026,6 +1031,9 @@ def format_rsakey_from_pem(pem):
     pem:
       A string in PEM format.
 
+    check_private:
+      A boolean indicating whether the check for a private key in PEM format. 
+
   <Exceptions>
     tuf.FormatError, if 'pem' is improperly formatted.
 
@@ -1048,8 +1056,14 @@ def format_rsakey_from_pem(pem):
   # Ensure the PEM string has a valid header and footer.  Although a simple
   # validation of 'pem' is performed here, a fully valid PEM string is
   # needed to later successfully verify signatures.
-  pem_header = '-----BEGIN PUBLIC KEY-----'
-  pem_footer = '-----END PUBLIC KEY-----'
+  if check_private:
+    pem_header = '-----BEGIN RSA PRIVATE KEY-----'
+    pem_footer = '-----END RSA PRIVATE KEY-----'
+  
+  else:
+    pem_header = '-----BEGIN PUBLIC KEY-----'
+    pem_footer = '-----END PUBLIC KEY-----'
+  
   header_start = 0
   footer_start = 0
 
--- a/tuf/keys.py
+++ b/tuf/keys.py
@@ -63,10 +63,11 @@ import binascii
 # http://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings
 import warnings
 
-# 'pycrypto' is the only currently supported library for the creation of RSA
-# keys.
+# 'pycrypto' and 'cryptography' are the only currently supported libraries for
+# the creation of RSA keys.
 # https://github.com/dlitz/pycrypto
-_SUPPORTED_RSA_CRYPTO_LIBRARIES = ['pycrypto']
+# https://github.com/pyca/cryptography
+_SUPPORTED_RSA_CRYPTO_LIBRARIES = ['pycrypto', 'pyca-cryptography']
 
 # The currently supported libraries for the creation of ed25519 keys and
 # signatures.  The 'pynacl' library should be installed and used over the slower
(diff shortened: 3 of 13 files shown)


tuf/formats.py
```diff
@@ -151,6 +151,9 @@
 NAME_SCHEMA = SCHEMA.AnyString()
 NAMES_SCHEMA = SCHEMA.ListOf(NAME_SCHEMA)
 
+# A string representing data.
+DATA_SCHEMA = SCHEMA.AnyString()
```
aaaaalbert: I wonder if that will do what it should, with the Python2/3 `str`/ `bytes` dichotomy.

tuf/keys.py
```diff
@@ -74,11 +75,11 @@
 # if 'pynacl' is unavailable.
 _SUPPORTED_ED25519_CRYPTO_LIBRARIES = ['ed25519', 'pynacl']
 
-# 'pycrypto' is the only currently supported library for general-purpose
-# cryptography, with plans to support pyca/cryptography.
+# 'pycrypto' and 'cryptography' are the only currently supported library for
+# general-purpose cryptography, with plans to support pyca/cryptography.
```
aaaaalbert: `with plans to support pyca/cryptography.`? --- Implemented in this very PR.

tuf/keys.py
```diff
@@ -93,6 +94,15 @@
 except ImportError: # pragma: no cover
   pass
 
+# Import the pyca/Cryptography library so that RSA and generate-purpose
+# cryptography primitives are supported. 
+try:
+  import cryptography
```
aaaaalbert: AFAICT `cryptography` isn't used as a library in this module; same for `Crypto` above. Sub-modules of both are imported in the respective `tuf.*_keys` modules.

tuf/keys.py
```diff
@@ -223,6 +233,9 @@ def generate_rsa_key(bits=_DEFAULT_RSA_KEY_BITS):
   if _RSA_CRYPTO_LIBRARY == 'pycrypto':
     public, private = tuf.pycrypto_keys.generate_rsa_public_and_private(bits)
   
+  elif _RSA_CRYPTO_LIBRARY == 'pyca-cryptography':
```
aaaaalbert: This idiom (testing for things in `_RSA_CRYPTO_LIBRARY` or the ED25519 counterpart) is so common in this module that my suggestion would be to consider to refactor it along the lines of
```
try:
  import one_rsa_lib as rsa_lib
except ImportError:
  pass
try:
  import another_rsa_lib as rsa_lib
....
```

Also saves you tons of identically set up checks later on when you go and check for specific implemtations to be availble (and then call identically-named functions from them).

tuf/keys.py
```diff
@@ -139,7 +149,7 @@
 
 # The crypto libraries to use in 'keys.py', set by default or by the user.
 # The following cryptography libraries are currently supported:
-# ['pycrypto', 'pynacl', 'ed25519']
+# ['pycrypto', 'pynacl', 'ed25519', 'pyca-cryptography']
```
aaaaalbert: Extends https://github.com/theupdateframework/tuf/blob/develop/tuf/conf.py#L86 (so `conf` probably needs an update as well).

tuf/keys.py
```diff
@@ -832,6 +848,17 @@ def verify_signature(key_dict, signature, data):
       else:
         valid_signature = tuf.pycrypto_keys.verify_rsa_signature(sig, method,
                                                                  public, data) 
+    elif _RSA_CRYPTO_LIBRARY == 'pyca-cryptography': 
+      if 'pyca-cryptography' not in _available_crypto_libraries: # pragma: no cover
+        message = 'Metadata downloaded from the remote repository specified' +\
+          ' an RSA signature.  Verifying RSA signatures requires pyca/cryptography.' +\
+          '\n$ pip install cryptography, or pip install tuf[tools].'
+        raise tuf.UnsupportedLibraryError(message)
```
aaaaalbert: I think the error message is wrong. *Your config* (from which `_RSA_CRYPTO_LIBRARY` was set) requires *this particular* lib. Generally speaking, either RSA lib would suffice.

Same goes for https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R846 above.

tuf/keys.py
```diff
@@ -1016,8 +1053,66 @@ def format_rsakey_from_pem(pem):
...
   # validation of 'pem' is performed here, a fully valid PEM string is
   # needed to later successfully verify signatures.
-  pem_header = '-----BEGIN PUBLIC KEY-----'
-  pem_footer = '-----END PUBLIC KEY-----'
+  public_pem = extract_pem(pem) 
+
+  # Begin building the RSA key dictionary. 
+  rsakey_dict = {}
+  keytype = 'rsa'
+  public = public_pem
```
aaaaalbert: To what extent does this local variable help?

tuf/keys.py
```diff
@@ -1016,8 +1053,66 @@ def format_rsakey_from_pem(pem):
...
+  rsakey_dict['keyid'] = keyid
+  rsakey_dict['keyval'] = key_value
+
+  return rsakey_dict
+
+
+
+
+
+def extract_pem(pem, private_pem=False):
```
aaaaalbert: I think you can an erroneous `pem` past the checks contained in here, e.g. by appending multiple copies of the `pem_header` in front.

tuf/keys.py
```diff
@@ -1016,8 +1053,66 @@ def format_rsakey_from_pem(pem):
...
+
+  <Exceptions>
+    tuf.FormatError, if 'pem' is improperly formatted.
+
+  <Side Effects>
+    Only the public and private portion of the PEM is extracted.  Leading or
+    trailing whitespace is not included in the PEM string returned.
+
+  <Returns>
+    A PEM string (excluding leading and trailing newline characters.
```
aaaaalbert: I.e. it returns `pem_header` + key material + `pem_footer`, right?

(Also, this line is missing a closing `)`.)

tests/test_pyca_crypto_keys.py
```diff
@@ -0,0 +1,140 @@
...
+    self.assertEqual(None, tuf.formats.PEMRSA_SCHEMA.check_match(pub),
+                     FORMAT_ERROR_MSG)
+    self.assertEqual(None, tuf.formats.PEMRSA_SCHEMA.check_match(priv),
+                     FORMAT_ERROR_MSG)
+
+    # Check for invalid bits argument.  bit >= 2048 and a multiple of 256.
+    self.assertRaises(tuf.FormatError,
+                      crypto_keys.generate_rsa_public_and_private, 1024)
+   
+    crypto_keys.generate_rsa_public_and_private(2050)
```
aaaaalbert: This should fail, as 2050 mod 256 != 0, but why is there no `assertRaises` around the call?

tests/test_pyca_crypto_keys.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Verify format of returned values.
+    self.assertNotEqual(None, signature)
+    self.assertEqual(None, tuf.formats.NAME_SCHEMA.check_match(method),
+                     FORMAT_ERROR_MSG)
+    self.assertEqual('RSASSA-PSS', method)
+
+    # Check for improperly formatted arguments.
+    self.assertRaises(tuf.FormatError,
+                      crypto_keys.create_rsa_signature, 123, data)
```
aaaaalbert: I believe this will raise a `TypeError` instead, as `int`s have no `len`: https://github.com/vladimir-v-diaz/tuf/blob/ef2ab091db1aa87d3feec3b0732135d65be3bb16/tuf/pyca_crypto_keys.py#L310

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+
+  # Verify the signature, but only if the private key has been set.  The private
+  # key is a NULL string if unset.  Although it may be clearer to explicitly
+  # check that 'private_key' is not '', we can/should check for a value and not
+  # compare identities with the 'is' keyword.  Up to this point 'private_key'
+  # has variable size and can be an empty string.
+  if len(private_key):
+    
+    # pyca/cryptography's expected exceptions when generating RSA key object:
+    # "ValueError/IndexError/TypeError:  When the given key cannot be parsed
```
aaaaalbert: Taking this at face value, doesn't the `try/except` block then let `IndexError`s raise? If so, this should be added to the docstring.

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+  signature = None
+
+  # Verify the signature, but only if the private key has been set.  The private
+  # key is a NULL string if unset.  Although it may be clearer to explicitly
+  # check that 'private_key' is not '', we can/should check for a value and not
+  # compare identities with the 'is' keyword.  Up to this point 'private_key'
+  # has variable size and can be an empty string.
+  if len(private_key):
+    
+    # pyca/cryptography's expected exceptions when generating RSA key object:
```
aaaaalbert: Are we generating a key object at all? (I suppose this should read "..when generating an RSA signature").

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+                      salt_length=hashes.SHA256().digest_size), hashes.SHA256())
+                          
+    except ValueError: #pragma: no cover
+      raise tuf.CryptoError('The private key could not be deserialized.')
+ 
+    except cryptography.exceptions.UnsupportedAlgorithm: #pragma: no cover
+      message = 'The private key is encrypted with an unsupported algorithm.'
+      raise tuf.CryptoError(message)
+   
+    # Generate an RSSA-PSS signature.  Raise 'tuf.CryptoError' for the expected
```
aaaaalbert: It's hard to see where `tuf.CryptoError` could come from, as all the ingredients to `rsa_signer` seem to come from the `cryptography` lib: https://github.com/vladimir-v-diaz/tuf/blob/ef2ab091db1aa87d3feec3b0732135d65be3bb16/tuf/pyca_crypto_keys.py#L70 , https://github.com/vladimir-v-diaz/tuf/blob/ef2ab091db1aa87d3feec3b0732135d65be3bb16/tuf/pyca_crypto_keys.py#L91 , https://github.com/vladimir-v-diaz/tuf/blob/ef2ab091db1aa87d3feec3b0732135d65be3bb16/tuf/pyca_crypto_keys.py#L102

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+    # pyca/cryptography exceptions.
+    rsa_signer.update(data)
+    signature = rsa_signer.finalize()
+  
+  else:
+    raise TypeError('The required private key is unset.')
```
aaaaalbert: I would consider this a `ValueError` instead, see also https://github.com/vladimir-v-diaz/tuf/blob/ef2ab091db1aa87d3feec3b0732135d65be3bb16/tuf/pyca_crypto_keys.py#L970-L971

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+
+      verifier.update(data.encode('utf-8'))
+      
+      # verify() raises 'cryptograpahy.exceptions.InvalidSignature' if the
+      # signature is invalid.
+      try: 
+        verifier.verify()
+        valid_signature = True 
+      
+      except cryptography.exceptions.InvalidSignature as e:
```
aaaaalbert: `e` isn't used.

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+        pass
+ 
+    # Raised by load_pem_public_key(). 
+    except ValueError:
+      raise tuf.CryptoError('The PEM could not be decoded successfully.')
```
aaaaalbert: `tuf.CryptoError` isn't mentioned in the docstring.

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+
+    # Raised by load_pem_public_key().
+    except cryptography.exceptions.UnsupportedAlgorithm:
+      raise tuf.Cryptography('The private key type is not supported.')
```
aaaaalbert: `tuf.Cryptography` doesn't seem to be an existing exception class. Probably meant `tuf.CryptoError`.

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+  
+  else:
```
aaaaalbert: This `else` branch isn't covered by the unit tests.

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+                                  salt_length=hashes.SHA256().digest_size), 
+                                  hashes.SHA256())
+
+      verifier.update(data.encode('utf-8'))
+      
+      # verify() raises 'cryptograpahy.exceptions.InvalidSignature' if the
+      # signature is invalid.
+      try: 
+        verifier.verify()
+        valid_signature = True
```
aaaaalbert: This feels kind of round-about. What's wrong with
```
try:
  verifier.verify()
  return True
except ......:
  return False
```

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+      Data used by tuf.keys.create_signature() to generate
+      'signature'.  'data' (a string) is needed here to verify 'signature'.
+
+  <Exceptions>
+    tuf.UnknownMethodError.  Raised if the signing method used by
+    'signature' is not one supported by tuf.keys.create_signature().
+    
+    tuf.FormatError. Raised if 'signature', 'signature_method', or 'public_key'
+    are improperly formatted.
+
```
aaaaalbert: Not all exceptions are mentioned (see comments below).

Furthermore, not all exceptions are triggered by the unit tests.

tuf/pyca_crypto_keys.py
```diff
@@ -0,0 +1,1017 @@
...
+  tuf.formats.DATA_SCHEMA.check_match(data)
+
+  # Verify whether the private key of 'public_key' produced 'signature'.
+  # Before returning the 'valid_signature' Boolean result, ensure 'RSASSA-PSS'
+  # was used as the signing method.
+  valid_signature = False
+
+  # Verify the signature with pyca/cryptography if the signature method is
+  # valid, otherwise raise 'tuf.UnknownMethodError'.
+  if signature_method == 'RSASSA-PSS':
```
aaaaalbert: Instead of the long `if` clause, why not get unsupported signature methods out of the way upfront, just as is done with the `SCHEMA` checks? Decreases indent, untangles pre-flight checks and actual business logic:
```
if signature_method != "RSASSA-PSS":
  raise tuf.UnknownMethodError(signature_method)
# Implicit `else` follows....
```

vladimir-v-diaz:
I'd like to document (on this pull request) all of the feedback given thus far, so I'm including more of Albert's comments here...

`keys.py`:
* Lots of repeating special-casing for when one crypto lib is available vs the other. They are used exclusively, i.e not multiple of them at the same time, anyway, right? (If you really wanted to check for a lib to be available, see if there's a module with the expected name in the scope.) I see a lot of potential for `try: import a_lib as lib` etc. to fix this in one central place. Then, the various functions call into an umbrella lib name, and "just work".
* Lots of explicit line continuation using backslashes. I find this hard to read, and it's easy to work around this very often. E.g.,
```
message = "multi " + \
    "line " + \
    "error message"
raise MyException(message)
```

versus

```
raise MyException("line "
    "continuation "
    "for free, inside parentheses!")
```

* There were a few rather long comments which explain things that are obvious from the following line of code anyway (or so I found). They weren't added recently, but they made me feel I needed to be extra cautious looking at a few places, when in fact a dictionary got assigned a new item.


Things not related to this PR:
https://github.com/theupdateframework/tuf/blob/develop/tuf/formats.py#L181
Doesn't check that the bit size is a power of two / multiple of 256. See also https://github.com/theupdateframework/tuf/blob/develop/tuf/formats.py#L184 where the same check is missing.


https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R250
Redundant comments IMHO.


https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R38
Why do the links appear here of all places?

https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R43
This *is* keys.py!

https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R316
Where exactly is the *fallback*? Not in https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R322 it seems, where the error is just raised.


https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R264 and https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R158
Smell like refactor. Set the key dict later on, remove local variables.


https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R415
I propose for this to raise an error when private==True but "private" not in key_value.


https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R673
s/conformat/conformant


https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R689
Docstring outdated, this PR adds pyca-cryptography.

https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R676 and https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R812
Long-winding comment with little extra info


https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R906
Why do we care? No "generate" inside of this function! Should this indicate that the function won't read in / use keys of bit sizes less than 2048? I don't see that checked.

https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R1009
Again, doesn't seem to enforce the lower 2048 bits size limit.



https://github.com/theupdateframework/tuf/pull/289/files?diff=split#diff-a3be85d1a9209356a591a86fedcad4a5R1125
This may paste private key material of a slightly-off formatted key into the error log.

tuf/keys.py
awwad: Resolved by [commit](https://github.com/vladimir-v-diaz/tuf/commit/f6d7302293d23ab2094f865464ebbf17b2633168) in [PR #311](https://github.com/theupdateframework/tuf/pull/311)

tuf/keys.py
awwad: Not resolved by PR #311.

tests/test_pyca_crypto_keys.py
awwad: Resolved in PR #311 (Line removed). Presume it was left in by mistake.

tuf/keys.py
awwad: Error message improved [here in PR #311](https://github.com/theupdateframework/tuf/pull/311/files#r52197784). I don't love how it's structured, still, but I haven't yet grokked enough to be confident in that. (What is determining which library to use? User selection in config? Will see.)

TODO.

tuf/keys.py
awwad: Resolved [here in PR #311 with removal of variable](https://github.com/theupdateframework/tuf/pull/311/files#diff-a3be85d1a9209356a591a86fedcad4a5R1070).

tuf/keys.py
awwad: Not resolved in PR #311. (Seems worth doing to my naive eyes. Any issues with it?)

tuf/keys.py
awwad: I don't understand this comment. Slow_Start_Grace_Period is extended..?

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311 by removal of reference to an IndexError. Also, the particulars for Value and Type errors are broken out separately. Index errors are not expected or mentioned by PR #311 code and would be raised normally.

tuf/pyca_crypto_keys.py
awwad: Apparently resolved (agreed to?) by removal of offending comment / reference to tuf.CryptoError in PR #311.

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311, switched to ValueError.

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311 by removal of "as e"

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311.

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311

tuf/keys.py
aaaaalbert: IIRC this comment was meant to point at a similar comment in `conf.py`, https://github.com/theupdateframework/tuf/blob/7f91f56802dd6b851ca6bc5645744b9617011d55/tuf/conf.py#L93-L95 (I probably typoed the line number, and/or they changed due to commits since.)

The point being: Both `keys.py` and `conf.py` should mention the same supported crypto libraries.

tuf/keys.py
awwad: Unclear if resolved:

Vlad, your intention with [the comments added here in PR #311](https://github.com/theupdateframework/tuf/pull/311/files#diff-a3be85d1a9209356a591a86fedcad4a5R1097) is to indicate that (thorough) validation here is not necessary because we can trust that pyca/cryptography and PyCrypto will perform any necessary validation, and any tinkering will result in a pem that is unusable?

tuf/keys.py
awwad: Resolved in PR #311.

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311 by removal of comment, presumably indicating agreement that such an error is not expected.

tuf/pyca_crypto_keys.py
awwad: NOT Resolved, I think.

PR #311:
 - adds mention of fact that tuf.CryptoError can be raised
 - adds mention of fact that tuf.FormatError can be raised by 'data' being improperly formatted

The obvious exceptions are thus handled in PR #311.

PR #311 does not, however, appear to add any tests that would trigger exceptions, and it is implied in Albert's comment that this is needed. (Is that a matter of coding style? I don't see anything like that in [the style guide](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)...?

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311.

tuf/pyca_crypto_keys.py
awwad: Resolved in PR #311. Improved flow.

tuf/pyca_crypto_keys.py
awwad: (Should it be? Will discuss with Vlad.)

tuf/pyca_crypto_keys.py
awwad: Unclear if resolved.

PR #311 changes the comment in the direction of the initial suggestion here, which is then counter to the edited suggestion. If it confused Albert, then it should likely be adjusted again to read something like "Converts a private key from PEM format to a pyca/cryptography object".

awwad:
This comment (Will call it No. 4) isn't addressed by PR #313. (BTW, this 3-PR commenting business - 289 having comments to be addressed in 313 per 311 - is getting a little convoluted. ): )
> https://github.com/theupdateframework/tuf/blob/develop/tuf/formats.py#L181
> Doesn't check that the bit size is a power of two / multiple of 256. See also https://github.com/theupdateframework/tuf/blob/develop/tuf/formats.py#L184 where the same check is missing.

SantiagoTorres:
Hate to say this, but I told you :P

awwad:
Item --9 in the long comment list constitutes, I think, a suggestion that the temps "message" and "keyid" be removed from functions generate_ed25519_key() (L264-341) and generate_rsa_key() (L158-258), as they are not necessary and don't really contribute to readability. PR #313 removes message, but not keyid. I'm not sure if that's intentional. NBD.

awwad:
Item --10 resolved to fill in logic gap, add explicit error message, and clarifying comments by PR #313.

awwad:
Item --11 addressed by typo correction in PR #313.

awwad:
Item --12 resolved in code comments in PR #313.

awwad:
Item --13 is NOT addressed in PR #313 (but I think it's NBD, personally).

awwad:
Item --14: NOT addressed in PR #313. I'm a bit confused about this myself: I don't see why this code comment is here, either. This is an RSA key import, but the comment makes it seem like we're generating a new key. My guess is that you're explaining the doctest lines that follow, but that information doesn't seem relevant to my eyes, anyway. Thoughts?

> Although the PyCrypto crytography library called sets a 1024-bit minimum key size, generate() enforces a minimum key size of 2048 bits.  If 'bits' is unspecified, a 3072-bit RSA key is generated, which is the key size recommended by TUF.

awwad:
Item --14: Addressed by removal of the comment in PR #313.

awwad:
Item --15: Not addressed. See --14.

awwad:
Item --16: Resolved by logic [here](https://github.com/theupdateframework/tuf/pull/313/files#diff-a3be85d1a9209356a591a86fedcad4a5R1150) and [here](https://github.com/theupdateframework/tuf/pull/313/files#diff-a3be85d1a9209356a591a86fedcad4a5R1164) in PR #313, in that if the pem is treated as private, its contents will not be included in the exception.

awwad:
Rounding it out all in one place:
Item --1: NOT addressed. (Broad redundancy possibly replaceable by two try-import-as statements.) See comments [here](https://github.com/theupdateframework/tuf/pull/289#discussion_r48360374).
Item --2: Mostly resolved. Some left over. See comments I made.
Item --3: Verbose comments. Think largely resolved. NBD.
Item --5: NOT addressed in PR #313, but NBD.
Items --6, --7, and --8: Resolved in PR #313.

That rounds it all out, I think.

List of unresolved:
* Noteworthy: --1, --4, --15 
* NBD: --5, --9, --13

tuf/keys.py
awwad: Ah. Resolved by PR #311 clarifying comments.

vladimir-v-diaz:
@awwad says that these pull requests have now been reviewed.

Good work, Sebastien.  I''ll get started on updating the pull rquests soon.

vladimir-v-diaz:
@awwad 
I edited some of your comments to add some of my own.

vladimir-v-diaz:
@awwad 

I made additional commits to this pull request to address some of your comments.  Merging!

PR #292, opened 2015-10-05
Just fixing a link in the readme that pointed to a moved webpage for Virtualenv.

awwad:


vladimir-v-diaz:
Thanks for the pull request.  I reviewed your fixes and the links now point to valid pages.

PR #295, opened 2015-10-28
Specify version numbers, rather than file lengths and hashes, in snapshot.json

vladimir-v-diaz:
List version numbers (rather than file sizes and hashes) of the metadata available on the repository in the top-level `Snapshot` role.

First, this makes updates more efficient by reducing the amount of metadata that needs to be downloaded.

Second, delegated metadata that clients have not lazily evaluated are vulnerable to Rollback attacks if the Timestamp and Snapshot roles are compromised. A Rollback Attack is when "an attacker presents a software update system with older files than those the client has already seen, causing the client to use files older than those the client knows about."

Suppose that there are two delegated roles available on the repository, roles A and B. The client requests a file that is provided by the A role, and version 2 of A is lazily downloaded and cached. That is, B is not downloaded at all. If at this point an attacker compromises Timestamp and snapshot and tries to trick the client into installing version 1 of A, the attack will not succeed because the client will not accept a version of A that is less than the client's currently cached version of 2. The attacker, however, can trick the client into installing an old version of B because the client has not downloaded and cached B yet.

One way to prevent Rollback attacks against any delegated role provided by the repository is to always download them all. Unfortunately, this approach will be inefficient for repositories that provide many delegated roles.

Another approach is to include the current version number of every delegated role available on the repository in the snapshot role. In the example above, the snapshot role would have specified that version 2 of A is currently available, and version 3 of B. With this setup, clients may lazily download A and not have to worry about Rollback attacks against B because it would know from the snapshot role that version 3 is the latest.


dachshund:
Hey @vladimir-v-diaz, thanks for making this happen! :+1: 

I've a quick question. Do we ever delete anything from the snapshot metadata? I ask because suppose the snapshot metadata at version 1 says, "The Django metadata is at version 42". Then, attackers compromise the repository and publish new snapshot metadata at version 2 which says, "There is no Django metadata." Then, they publish new snapshot metadata at version 3 which says, "The Django metadata is at version 41." If a user downloads all three snapsnot metadata and deletes previously-published entries that are later "unpublished", then the user would be susceptible to a rollback attack. We could solve this problem by using something like a "shadow" snapshot metadata to cache everything ever listed in the snapshot metadata, even if they are later deleted.

Sorry I haven't had time to review the code, which is why I'm asking this dumb question. Let me know. Thanks!

vladimir-v-diaz:
This is a good find.

We cache previous versions of metadata (in /metadata/previous/*), but this approach wouldn't work with the attack you've described because the client only caches the last version of metadata.  How about caching all metadata that is downloaded?  It would be nice to retain all this metadata, but it might be take up unnecessary disk space...

When you say a "shadow" snapshot metadata, will this be a separate file or we are slightly tweaking what is included in snapshot.json?  For example, 

```
{
"_type": "Snapshot",
"meta": {"django.json": {versions: 1-5, 8, 10}, ...}
}
```
Where 10 is the latest version number of the Django project.

dachshund:
By "shadow" snapshot metadata, I mean a file in the user's cache that only adds or updates entries, but never removes them.

Yes, if a file only ever grows, then disk space could be a problem. However, if we assume one developer per project on PyPI, then the snapshot metadata is in the order of a few megabytes. Even if the file grows to ten times the size of that, I think most computers (except for things like embedded devices) can spare dozens of megabytes. What do you think?

Another thing the attacker can do now is to "fast-forward" version numbers after a repository compromise. For example, suppose that before the compromise the Django metadata was at version 1. Then, after the compromise, the attacker may say that the Django metadata is at version 3000 when actually it is still at version 1. This attack is risky, because the attack will be caught if the attacker "fast-forwards" metadata that the user actually ends up checking. At worst, this should only be a denial-of-service attack.

To recover from a "fast-forward" attack after the repository is recovered after a compromise, one solution is to "reset" the "shadow" snapshot metadata. I think we should be careful in "resetting" the "shadow" snapshot metadata. Let me think about this a bit more as I write the paper, and get back to you soon.

vladimir-v-diaz:
Is there a reason why we want this "shadow" snapshot metadata to be a separate file?

I was thinking of keeping a history of released versions in `snapshot.json`.  Clients would normally reject any version that is less than what's in the history.  If the repository wants to remove some role, it can mark it as removed in `snapshot.json` with a '-' (or whatever). 

Example:
{
"versions": {1-12, 14, -}
}

Another option for resetting metadata is to allow clients to reset snapshot.json to a safe, older version when the Root file revokes a key.

dachshund:
The "shadow" snapshot metadata is one way to not forget anything, even if the compromised snapshot role "unpublishes"/deletes anything later.

Would you please explain why the snapshot metadata should contain all previous versions? Why not just say "14" in the previous example, and let it stay that way? If that role is now not trusted to sign a package, it won't appear in the current chain of delegations, but we can still preserve its last version number in the snapshot.

vladimir-v-diaz:
I was thinking of preserving all previous versions of metadata in `snapshot.json` to support your goal of not forgetting anything and preventing the rollback attack that you mentioned.  Instead of having a separate "shadow" metadata to store this past information I thought it would simpler to just do it in `snapshot.json`.

I guess we can preserve a role's last version number in the snapshot, even if the repository no longer trusts it.  That is, a role is never deleted from `snapshot.json` which is something we'll likely want.  We'd have to tweak the definition of the snapshot to mention that is lists the current, or last trusted, version numbers of metadata on the repository.

So there are three options we can consider (after our discussion this morning):
1. Mark a role as deleted in snapshot (with a special symbol) and list previously trusted versions.
2. Have a separate "shadow" snapshot metadata to keep a history of versions.
3. preserve the last version in snapshot even if it wont appear in the current chain of delegations.

tuf/client/updater.py
```diff
@@ -1561,85 +1659,73 @@ def _update_metadata_if_changed(self, metadata_role,
...
       None.
 
     <Side Effects>
-      If there is no fileinfo currently loaded for 'metada_filename',
+      If there is no versioninfo currently loaded for 'metada_filename',
       try to load it.
 
     <Returns>
-      Boolean.  True if the fileinfo has changed, false otherwise.
+      Boolean.  True if the versioninfo has changed, false otherwise.
```
aaaaalbert: 1: "[...], *False* otherwise." (note capital F to match Python's `False` constant)
2: `True` is only returned when the `versioninfo` is *strictly numerically greather than* it was before. See [code below](https://github.com/theupdateframework/tuf/pull/295/files#diff-852078e79350a80482b37e47cae64ed3R1712).

tuf/client/updater.py
```diff
@@ -1561,85 +1659,73 @@ def _update_metadata_if_changed(self, metadata_role,
...
 
 
-  def _fileinfo_has_changed(self, metadata_filename, new_fileinfo):
+  def _versioninfo_has_changed(self, metadata_filename, new_versioninfo):
     """
     <Purpose>
-      Non-public method that determines whether the current fileinfo of
-      'metadata_filename' differs from 'new_fileinfo'.  The 'new_fileinfo'
+      Non-public method that determines whether the current versioninfo of
+      'metadata_filename' differs from 'new_versioninfo'.  The 'new_versioninfo'
```
aaaaalbert: See comment below -- the comparison of version numbers is `>`, not `!=`.

tuf/client/updater.py
```diff
@@ -1561,85 +1659,73 @@ def _update_metadata_if_changed(self, metadata_role,
 
 
 
-  def _fileinfo_has_changed(self, metadata_filename, new_fileinfo):
+  def _versioninfo_has_changed(self, metadata_filename, new_versioninfo):
```
aaaaalbert: See comments below. The method isn't really checking for *change*, but *increase* in version number. It should perhaps be called `_file_has_been_updated`.

tuf/client/updater.py
```diff
@@ -1651,32 +1737,67 @@ def _update_fileinfo(self, metadata_filename):
...
-  
- 
+    # Extract the version information from the trusted snapshot role and save
+    # it to the 'self.versioninfo' store.
+    if metadata_filename == 'timestamp.json':
+      trusted_versioninfo = \
+        self.metadata['current']['timestamp']['version']
+
+    # When updating snapshot.json, the client either (1) has a copy of
+    # snapshot.json, or (2) in the process of obtaining it by first downloading
```
aaaaalbert: Minor wording issue:
"    # snapshot.json, or (2) *is* in the process of obtaining it by first downloading"

tuf/client/updater.py
```diff
@@ -2721,20 +2842,15 @@ def download_target(self, target, destination_directory):
...
-      raise tuf.Error(message)
+    try:
+      os.makedirs(target_dirpath)
+    
+    except OSError as e:
+      if e.errno == errno.EEXIST:
+        pass
+      
+      else:
+        raise
```
aaaaalbert: Mention other potential `os` errors in docstring, e.g. (on Mac OS X) `OSError: [Errno 13] Permission denied: 'writeable/dir/with/readonly/subdir/NEW'`

tuf/client/updater.py
```diff
@@ -2721,20 +2842,15 @@ def download_target(self, target, destination_directory):
...
     
-    else:
-      message = repr(target_dirpath) + ' does not exist.'
-      logger.warning(message)
-      raise tuf.Error(message)
+    try:
+      os.makedirs(target_dirpath)
+    
+    except OSError as e:
+      if e.errno == errno.EEXIST:
```
aaaaalbert: Note: `errno.EEXIST` is only raised when the full dir to be `makedirs`ed already exists (e.g. `makedirs("foo/bar/baz")`, where "foo/bar/baz" was there already).

If only parts of the required directory structure exist, this is not an error, and the deeper-nested parts are created within the outer existing structure (e.g. "foo/bar/baz/frobble" inside of "foo/bar").

tuf/formats.py
```diff
@@ -996,6 +1020,42 @@ def make_fileinfo(length, hashes, custom=None):
...
+    version_number:
+      An integer representing the version of a particular metadata role.
+      The dictionary returned by this function is expected to be included
+      in Snapshot metadata.
+
+  <Exceptions>
+    tuf.FormatError, if the 'VERSIONINFO_SCHEMA' to be returned
+    does not have the correct format.
+
+  <Side Effects>
```
aaaaalbert: I'm inclined to list None as the side effects. The return value and possible exception are described elsewhere.

tuf/formats.py
```diff
@@ -996,6 +1020,42 @@ def make_fileinfo(length, hashes, custom=None):
...
+    describes both metadata and target files.
+
+  <Arguments>
+    version_number:
+      An integer representing the version of a particular metadata role.
+      The dictionary returned by this function is expected to be included
+      in Snapshot metadata.
+
+  <Exceptions>
+    tuf.FormatError, if the 'VERSIONINFO_SCHEMA' to be returned
```
aaaaalbert: Proposed edit: " tuf.FormatError, if the *dict* to be returned [...]"

tuf/formats.py
```diff
@@ -996,6 +1020,42 @@ def make_fileinfo(length, hashes, custom=None):
...
+
+  <Returns>
+    A dictionary conformant to 'VERSIONINFO_SCHEMA', containing the version
+    information of a metadata role.
+  """
+
+  versioninfo = {'version' : version_number}
+
+  # Raise 'tuf.FormatError' if 'versioninfo' is improperly formatted.
+  VERSIONINFO_SCHEMA.check_match(versioninfo)
```
aaaaalbert: Might do this to emphasize the exclusive-or of paths out of this function:
```Python
try:
   VERSIONINFO_SCHEMA.check_match(versioninfo) 
except:
  raise
else:
  return versioninfo
```

tuf/repository_lib.py
```diff
@@ -452,71 +464,68 @@ def _delete_obsolete_metadata(metadata_directory, snapshot_metadata,
...
-  return written_metadata_content, written_metadata_digests
+  return written_metadata_content
 
 
 
 
 
-def _strip_consistent_snapshot_digest(metadata_filename, consistent_snapshot):
+def _strip_consistent_snapshot_version_number(metadata_filename,
+                                              consistent_snapshot):
```
aaaaalbert: The docstring is far from complete.

tuf/repository_lib.py
```diff
@@ -452,71 +464,68 @@ def _delete_obsolete_metadata(metadata_directory, snapshot_metadata,
...
 
-  # Strip the digest if 'consistent_snapshot' is True.
-  # Example:  'targets/unclaimed/13df98ab0.django.json'  -->
+  # Strip the version number if 'consistent_snapshot' is True.
+  # Example:  'targets/unclaimed/10.django.json'  -->
   # 'targets/unclaimed/django.json'
   if consistent_snapshot:
     dirname, basename = os.path.split(metadata_filename)
-    embeded_digest = basename[:basename.find('.')]
+    embeded_version_number = basename[:basename.find('.')]
```
aaaaalbert: What about a plain `embedded_version_number, basename = basename.split(".", 1)`?

(Also note spelling of "embedded").

tuf/repository_lib.py
```diff
@@ -452,71 +464,68 @@ def _delete_obsolete_metadata(metadata_directory, snapshot_metadata,
...
     
-    # Ensure the digest, including the period, is stripped.
+    # Ensure the version number, including the period, is stripped.
     basename = basename[basename.find('.') + 1:]
     
     metadata_filename = os.path.join(dirname, basename)
```
aaaaalbert: `metadata_filename` was passed in from the caller originally. I (personally) find overwriting its contents with the "versionless" filename rather confusing, and would suggest using a separate variable for this.

tuf/repository_lib.py
vladimir-v-diaz: I don't always include full docstrings for private methods, unless it is heavily used or complicated.

tuf/client/updater.py
awwad: 1. Resolved in [321](https://github.com/theupdateframework/tuf/pull/321/files#diff-852078e79350a80482b37e47cae64ed3R1699)
2. Remains the case. Docstring and function name updated to be more consistent with this behavior. in PR #321. Will look through to see if there are any implications code calling `_versioninfo_has_been_updated`.

tuf/client/updater.py
awwad: Resolved in PR #321, see [above](https://github.com/theupdateframework/tuf/pull/295#discussion_r54261363).

tuf/client/updater.py
awwad: Resolved in PR #321, see [above](https://github.com/theupdateframework/tuf/pull/295#discussion_r54261363).

tuf/client/updater.py
awwad: Resolved [here in 321](https://github.com/theupdateframework/tuf/pull/321/files#diff-852078e79350a80482b37e47cae64ed3R1769).

tuf/client/updater.py
awwad: Resolved [here in 321](https://github.com/theupdateframework/tuf/pull/321/files#diff-852078e79350a80482b37e47cae64ed3R2817).

tuf/client/updater.py
awwad: Not quite resolved in 321: see [this comment](https://github.com/theupdateframework/tuf/pull/321/files#r54270274)

tuf/formats.py
awwad: Resolved in PR #321.

tuf/formats.py
awwad: Resolved in PR #321

tuf/formats.py
awwad: Resolved in PR #321

tuf/repository_lib.py
awwad: NBD, but NOT resolved in [PR 321 here](https://github.com/theupdateframework/tuf/pull/321/files#diff-9f585d5478a17bd9e0f6cd5f7484ab60R520).

Also, thanks, Albert - I find myself doing the ugly find thing often enough, and yours is clearer and shorter, yes.

(Spelling resolved)

tuf/repository_lib.py
awwad: Resolved in [PR 321 here](https://github.com/theupdateframework/tuf/pull/321/files#diff-9f585d5478a17bd9e0f6cd5f7484ab60R525).

tuf/repository_lib.py
vladimir-v-diaz: Switched to using split() instead.  I don't quite enjoy reading split(".", 1) <-- The second argument might not be obvious to newcomers / inexperienced Python programmers), but accessing list elements in `basename[:basename.find('.')` is harder to read in this case.

tuf/client/updater.py
vladimir-v-diaz: I updated the comment to explain that only OSError.eexist is thrown if the full directory already exists.

tuf/client/updater.py
vladimir-v-diaz: I forgot to edit `test_updater.py` to use the new function name.  The unit tests should all pass now.

tuf/client/updater.py
```diff
@@ -1493,47 +1580,58 @@ def _update_metadata_if_changed(self, metadata_role,
...
+                                        ['meta'] \
+                                        [uncompressed_metadata_filename]
 
     # Check for the availability of compressed versions of 'snapshot.json',
     # 'targets.json', and delegated Targets (that also start with 'targets').
     # For 'targets.json' and delegated metadata, 'referenced_metata'
     # should always be 'snapshot'.  'snapshot.json' specifies all roles
-    # provided by a repository, including their file lengths and hashes.
+    # provided by a repository, including their version numbers.
     if metadata_role == 'snapshot' or metadata_role.startswith('targets'):
```
awwad: (Caveat: this is not related to this change and doesn't need to be fixed for this PR, but I think it should be fixed at some point, if I understand correctly.)

I dislike the whole split-conditional structure here.
(If conditional 1: make temp variable for use if conditionals 1 and 2. If conditional 2: continue logic.)

If I understand correctly, this is better:

```
    if ( 'gzip' in self.metadata['current']['root']['compression_algorithms']
        and (metadata_role == 'snapshot' or metadata_role.startswith('targets'))
        ):
      gzip_metadata_filename = uncompressed_metadata_filename + '.gz'
      compression = 'gzip'
      logger.debug('Compressed version of ' +
                repr(uncompressed_metadata_filename) + ' is available at ' +
                repr(gzip_metadata_filename) + '.')
    else:
        logger.debug('Compressed version of ' +
                     repr(uncompressed_metadata_filename) + ' not available.')
```

* There is no longer a case where the metadata role is snapshot or starts with targets but gzip is not an accepted compresison algorithm in which the temp gzip_metadata_filename is created. (Is this actually needed in some way? If so, I'd say suspect that might be worse.)
* Removed the "\"s in favor of parenthetical-based line continuation.
* Unfortunately had to leave a "):" on its own line because of line lengths.

It's correct syntax in the frame I've stuck it in, but I haven't tried to run tests on it. I'll loop back to try it out another time so I'm sure it doesn't break anything. LMK if I'm wrong at your leisure. (:

vladimir-v-diaz: I've cleaned up these if-else clauses (removed the line continuation markers and relocated the temp variable).  Let me know if readability is improved.

awwad: Better addressed in https://github.com/vladimir-v-diaz/tuf/commit/ddd06f620507ceae6c6ad783f300610674429efb.

Issue #303, opened 2016-01-13
Make log file name configurable


davedoesdev:
At the moment, it's always `tuf.log` in the current working directory.

vladimir-v-diaz:
I like this feature request.  It had occurred to me in the past, but I was fine with it being fixed during development.

Instead of having a fixed path for the log file (See https://github.com/theupdateframework/tuf/blob/develop/tuf/log.py#L77), we can make it configurable by adding an option to [conf.py](https://github.com/theupdateframework/tuf/blob/develop/tuf/conf.py) (which is where we store all configurable settings).

What about another option to disable file logging entirely?

davedoesdev:
Yes, that would be useful too.

vladimir-v-diaz:
Options to disable file logging entirely and configure the logger file were implemented in #304 
Let us know if this solution works for you.

davedoesdev:
That works fine - ta

davedoesdev:
Unfortunately, just importing `tuf.conf` (in order to set the log file) sets it to and creates `tuf.log`:

```shell
david@david-Latitude-E6440:~/dtuf$ rm tuf.log 
david@david-Latitude-E6440:~/dtuf$ python
Python 2.7.10 (default, Oct 14 2015, 16:09:02) 
[GCC 5.2.1 20151010] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import tuf.conf
>>> 
david@david-Latitude-E6440:~/dtuf$ ls
docs  htmlcov   python      requirements.txt  setup.py
dtuf  Makefile  README.md   setenv-hub.sh     test
dxf   pylintrc  README.rst  setenv-local.sh   tuf.log
```

davedoesdev:
https://github.com/theupdateframework/tuf/pull/307

vladimir-v-diaz:
Thanks for providing a fix for this!

vladimir-v-diaz:
This issue was fixed in PR #307 and #308

PR #304, opened 2016-01-14
Allow users to specify (via conf.py) the path to save log files.  Issue #303

vladimir-v-diaz:
Initial implementation for issue #303

### DIFF
--- a/tuf/log.py
+++ b/tuf/log.py
@@ -66,7 +66,6 @@ from __future__ import unicode_literals
 
 import logging
 import time
-import os
 
 import tuf
 import tuf.formats
--- a/tuf/log.py
+++ b/tuf/log.py
@@ -104,6 +104,10 @@ formatter = logging.Formatter(_FORMAT_STRING)
 # set by default.
 console_handler = None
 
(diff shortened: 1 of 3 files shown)


SantiagoTorres:
Nice :+1:

tuf/log.py
```diff
@@ -104,18 +105,25 @@
...
+  file_handler.setLevel(_DEFAULT_FILE_LOG_LEVEL)
+  file_handler.setFormatter(formatter)
+else:
+  pass
+
 
 # Set the logger and its settings.
 logger = logging.getLogger('tuf')
 logger.setLevel(_DEFAULT_LOG_LEVEL)
-logger.addHandler(file_handler)
```
SantiagoTorres: I'm wondering if this, and the previous if can be merged together for simplicity. They check for the same condition.

vladimir-v-diaz: I now set the logger (prior to adding a potential file handler) to avoid these multiple if-else statements.

davedoesdev:
Yes, that's great - thanks

PR #301, opened 2016-01-04
Fix Python 3 errors

davedoesdev:
- Signing data should be byte strings
- Don't log exception outside handler (#300)

PR #305, opened 2016-01-15
Fix Python 3 encoding issues in pcyca_crypto_keys.py

davedoesdev:


vladimir-v-diaz:
The test case failures listed in https://github.com/theupdateframework/tuf/pull/301#issuecomment-171894356 are now passing on my end.

Thanks!  Merging.

PR #307, opened 2016-01-19
Don't import tuf.log in __init__.py so importing tuf.conf doesn't start logging

davedoesdev:


vladimir-v-diaz:
:+1: The log file is no longer created if `tuf.conf` is imported:
```Bash

(tuf2.7)$ ls tuf.log
ls: cannot access tuf.log: No such file or directory
(tuf2.7)$ python
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import tuf.conf
>>> 
(tuf2.7)$ ls tuf.conf
ls: cannot access tuf.conf: No such file or directory
```

I think `test_init.py` will now have to directly import `tuf.log` though, otherwise a warning is raised about a missing handler.

```Bash
(tuf2.7)$ python test_init.py 
No handlers could be found for logger "tuf.test_keys"
..
Ran 2 tests in 0.000s

OK
```

Note that we incorrectly named the logger for this module.  It should be `tuf.test_init` instead of `tuf.test_keys`

davedoesdev:
Shall I do that on this PR or will you do it?

vladimir-v-diaz:
I've created a pull request to fix the issues in my last reply.  I'll go ahead and merge your PR.  Please let me know if #308 is not working as intended.

PR #311, opened 2016-01-25
Address comments to the `pyca/Cryptography` pull request

vladimir-v-diaz:
@awwad 

Can you please review this pull request?  You will only have to verify that I've covered @aaaaalbert 's feedback in PR #289.

Note: Feedback listed in this [comment](https://github.com/theupdateframework/tuf/pull/289#issuecomment-171392644) to be covered in a separate PR.


tuf/keys.py
```diff
@@ -86,7 +86,9 @@
 # default.  https://github.com/pyca/ed25519
 _available_crypto_libraries = ['ed25519']
 
-# Import the PyCrypto library so that RSA keys are supported.
+# Try to import TUF's PyCrypto module (pycrypto_keys.py), and verify and record
+# that the PyCrypto library is available.  pycrypto_keys.py is used here for
+# general-purpose cryptography and RSA.
 try:
   import Crypto
```
awwad: This import is unnecessary and adds nothing. Ditto line 103 below (import cryptography). [Albert's comment](https://github.com/theupdateframework/tuf/pull/289/files#r48359877) on PR #289 likely was trying to make this point.
 
0. This file per se does not use Crypto.
1. Importing tuf.pycrypto_keys should (and does, I think - will test) import whatever it, in turn, needs.
2. If importing tuf.pycrypto_keys fails, then we already know that we cannot use it and thereby will not be supporting Crypto.
3. This also doesn't help the user figure out what's missing because they're not seeing any messages if either import fails (which is fine).

So nothing is added.

tuf/keys.py
```diff
@@ -94,8 +96,9 @@
 except ImportError: # pragma: no cover
   pass
 
-# Import the pyca/Cryptography library so that RSA and generate-purpose
-# cryptography primitives are supported. 
+# Try to import TUF's pyca/Cryptography module (pyca_crypto_keys.py), and
+# verify and record that pyca/Cryptography's library is available.
+# pyca_crypto_keys.py is used for general-purpose cryptography and RSA.
 try:
   import cryptography
```
awwad: This import is unnecessary and adds nothing. Ditto line 93 above (import Crypto). [Albert's comment](https://github.com/theupdateframework/tuf/pull/289/files#r48359877) on PR #289 likely was trying to make this point.
 
0. This file per se does not use 'cryptography'.
1. Importing tuf.pyca_crypto_keys should (and does, I think - will test) import whatever it, in turn, needs.
2. If importing tuf.pyca_crypto_keys fails, then we already know that we cannot use it and thereby will not be supporting 'cryptography'.
3. This also doesn't help the user figure out what's missing because they're not seeing any messages if either import fails (which is fine).

So nothing is added.

tuf/keys.py
```diff
@@ -840,21 +843,23 @@ def verify_signature(key_dict, signature, data):
   if keytype == 'rsa':
     if _RSA_CRYPTO_LIBRARY == 'pycrypto':
```
awwad: Issue, IMO:
 - While better than prev version, instruction to user fails to cover alternative of switching configuration to other library. Suggest something like "You could install PyCrypto (pip install PyCrypto or pip install tuf[tools]) or try switching your configuration (tuf/conf.py) to use pyca-cryptography if that is available instead)." Or, better yet, you could even issue that message specifically if another value *is* in _available_crypto_libraries! Something like (after checking _available_crypto_libaries!) "tuf/conf.py indicates that pycrypto should be used, but it is not available/installed. You could install it (pip install PyCrypto or pip install tuf[tools]), or you could change the value of RSA_CRYPTO_LIBRARY in tuf/conf.py to 'pyca-cryptography', which appears to be available on this system." (I also considered suggesting a setting "either" that would choose whichever is available, in priority order, but that could change if the preferred library were suddenly available due to some other installation. That seems vaguely risky. Might be worth considering anyway, since would be more hassle-free for user?)

Small issue, probably OK:
 - The structure here is a little redundant. Could be checking if _RSA_CRYPTO_LIBRARY is in supported crypto list and in _available_crypto_libraries. Still would need sloppy user instruction case, though, anyway. NBD.

tuf/pyca_crypto_keys.py
```diff
@@ -309,15 +309,11 @@ def create_rsa_signature(private_key, data):
   # has variable size and can be an empty string.
   if len(private_key):
     
-    # pyca/cryptography's expected exceptions when generating RSA key object:
-    # "ValueError/IndexError/TypeError:  When the given key cannot be parsed
-    # (possibly because the passphrase is wrong)."
-    # A passphrase or password is not used to 'private_key', since it should
-    # not be encrypted.
+    # Generate an RSSA-PSS signature.  Raise 'tuf.CryptoError' for any of the
```
awwad: While this addresses [Albert's initial comment on PR #289](https://github.com/theupdateframework/tuf/pull/289/files#r48976488), it looks like this does not provide a clear explanation of what is happening. I suggest something along the lines of Albert's edited comment: "Converts a private key from PEM format to a pyca/cryptography object".

tuf/keys.py
vladimir-v-diaz: Great comment.

I'd like to note here (for documentation purposes) that integrators are not normally expected to directly import modules like `keys.py`, `pycrypto_keys.py`, etc., but rather the repository tools.

`repository_tool.py` provides functionality that doesn't require PyCrypto nor PyCA's Cryptography, so we instead raise exceptions when a caller tries to use a function that does requires either of these libraries (depending on what is specified in 'tuf.conf').

tuf/pyca_crypto_keys.py
vladimir-v-diaz: On lines 315-316 below, I essentially say exactly that.  Should I move those comments to line 312?

vladimir-v-diaz:
@awwad 
Thanks for reviewing this pull request.  I'll try to avoid post-commit reviews from now on.  Merging!

PR #313, opened 2016-01-29
Add changes for second set of review comments

vladimir-v-diaz:
See PR https://github.com/theupdateframework/tuf/pull/311

(This PR is specifically a response to [this single-post collection of review comments on PR #289](https://github.com/theupdateframework/tuf/pull/289#issuecomment-171392644) --Seb.)


SantiagoTorres:
Why not push to the same branch?

vladimir-v-diaz:
From experience, people tend to avoid reviewing a pull request if there are many changes.  I wanted to break it up into more manageable chunks.

This pull request is also for a different set of comments provided by Albert.

SantiagoTorres:
Of course, but if you reviewed *comments*, then the diff shouldn't change too much right(e.g., this diff is 20 lines of net change)?. I'm ot trying to be nosy, whatever works best for everyone :+1:

awwad:
While it doesn't pertain to this PR, as you noted, there's still a big segment in tuf/keys.py with exception messages that should be changed to match this [code style guideline](https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length) on line continuation. [Albert suggested the use of implied continuation inside parentheticals (matches code style guideline)](https://github.com/theupdateframework/tuf/pull/289#issuecomment-171392644), and you made a few changes elsewhere, but this section remains.
tuf/keys.py:572 (for example)

    message = 'The ' + repr(_RSA_CRYPTO_LIBRARY) + ' crypto library specified' +\
      ' in "tuf.conf.RSA_CRYPTO_LIBRARY" is not supported.\n' +\
      'Supported crypto libraries: ' + repr(_SUPPORTED_RSA_CRYPTO_LIBRARIES) + '.'
    raise tuf.UnsupportedLibraryError(message)

should instead be:

    raise tuf.UnsupportedLibraryError('The ' repr(_RSA_CRYPTO_LIBRARY) 
      ' crypto library specified in "tuf.conf.RSA_CRYPTO_LIBRARY" is not'
      ' supported. Supported crypto libraries: '
      repr(_SUPPORTED_RSA_CRYPTO_LIBRARIES) '.')

Much more readable.

tuf/keys.py
```diff
@@ -554,7 +573,7 @@ def check_crypto_libraries(required_libraries):
   tuf.formats.REQUIRED_LIBRARIES_SCHEMA.check_match(required_libraries)
  
   # The checks below all raise 'tuf.UnsupportedLibraryError' if the general,
```
awwad: I would probably move the checks below into their own function. (That said, I'm not sure why this is the place to make sure they're available in the first place, so maybe I'm missing something.)

tuf/keys.py
```diff
@@ -1232,8 +1266,8 @@ def encrypt_key(key_object, password):
  
   # check_crypto_libraries() should have fully verified _GENERAL_CRYPTO_LIBRARY.
   else: # pragma: no cover
-    message = 'Invalid crypto library: ' + repr(_GENERAL_CRYPTO_LIBRARY) + '.'
-    raise tuf.UnsupportedLibraryError(message) 
+    raise tuf.UnsupportedLibraryError('Invalid crypto library:'
+      ' ' + repr(_GENERAL_CRYPTO_LIBRARY) + '.')
```
awwad: Unnecessary explicit concatenation, unnecessary string separation - multiple concatenation formats.

Should be this:
 `raise tuf.UnsupportedLibraryError('Invalid crypto library: ' repr(_GENERAL_CRYPTO_LIBRARY) '.')`

instead of the current:
 `raise tuf.UnsupportedLibraryError('Invalid crypto library:'  ' ' + repr(_GENERAL_CRYPTO_LIBRARY) + '.')`

vladimir-v-diaz: Isn't that a syntax error?

```Python
>>> raise ValueError('Invalid crypto library: ' repr('hello') '.')
  File "<stdin>", line 1
    raise ValueError('Invalid crypto library: ' repr('hello') '.')
                                                   ^
SyntaxError: invalid syntax
```

vladimir-v-diaz: I wanted to address comment [#1](https://github.com/theupdateframework/tuf/pull/289#issuecomment-171392644) in a separate pull request, so intentionally left the large block of exception messages in check_crypto_libraries() untouched (which might change if we modify how we import some of these crypto modules).

I'll update this block in this pull request instead.

awwad: >.< Whoops, sorry. String literals only in implicit concatenation in parentheses. Stupid repr. ):

PR #323, opened 2016-02-26
Fix #322 by detecting expiry of stale files. initial attempt

### LINKED ISSUE #322: Client: Freeze attack issue - expiry of metadata may not be detected

awwad:
Toying with this fix. See Issue #322.


awwad:
Running unit tests.

awwad:
Replacing this PR with a cleaner one: #324.

PR #324, opened 2016-03-10
Fix #322 freeze attack by detecting expiry of stale metadata files

### LINKED ISSUE #322: Client: Freeze attack issue - expiry of metadata may not be detected

awwad:
This fixes Issue #322. Please see that issue!
It replaces PR #323, which got too messy.

**Update:
This PR also:**
- :bug::bug: Fixes [two additional bugs issues identified below](https://github.com/theupdateframework/tuf/pull/324#issuecomment-197543256), both causing unprintable NoWorkingMirrorError exceptions - one in python 3 only and the other in either 2 or 3 with certain mirror name issues.
- adds .DS_Store to .gitignore

This started as a manual squash commit of detect_expiry_322, to avoid merge conflicts.

Comment lines from the individual commits include:
1. Fix #322 by detecting expiry of stale files. initial attempt
2. temp commit of files from Soma
3. removing freeze_attack_stale_expiry and leaving the test added to indefinite freeze attack
4. fixing indefinite freeze attack test: now incorporates old reject-freshly-downloaded-but-expired-timestamp test as well as reject-stale-already-present-but-expired-snapshot test
5. small refinements to indefinite freeze attack test
6. Pulled the recursion out of the except block in refresh() to avoid unprintable nested exceptions.
7. Added comments to the last commit (retry_once)
8. Merge pull request #1 from awwad/detect_expiry_322_temp (removing cruft in another branch)

### DIFF
--- a/tests/test_indefinite_freeze_attack.py
+++ b/tests/test_indefinite_freeze_attack.py
@@ -181,18 +181,18 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
 
   def test_without_tuf(self):
     # Two tests are conducted here.
-    # Test 1: If an expired timestamp is downloaded, is it recognized as such?
-    # Test 2: If we find that the timestamp acquired from a mirror indicates
+    # Test 1: If we find that the timestamp acquired from a mirror indicates
     #         that there is no new snapshot file, and our current snapshot
     #         file is expired, is it recognized as such?
+    # Test 2: If an expired timestamp is downloaded, is it recognized as such?
 
-    # Test 2 Begin:
+    # Test 1 Begin:
     #
     # See description of scenario in Test 2 in the test_with_tuf method.
     # Without TUF, Test 2 is tantamount to Test 1, and so we skip this test.
 
 
-    # Test 1 Begin:
+    # Test 2 Begin:
     #
     # 'timestamp.json' specifies the latest version of the repository files.
     # A client should only accept the same version of this file up to a certain
@@ -241,13 +241,13 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
 
   def test_with_tuf(self):
     # Two tests are conducted here.
-    # Test 1: If an expired timestamp is downloaded, is it recognized as such?
-    # Test 2: If we find that the timestamp acquired from a mirror indicates
+    # Test 1: If we find that the timestamp acquired from a mirror indicates
     #         that there is no new snapshot file, and our current snapshot
     #         file is expired, is it recognized as such?
+    # Test 2: If an expired timestamp is downloaded, is it recognized as such?
 
 
-    # Test 2 Begin:
+    # Test 1 Begin:
     #
     # Addresses this issue: https://github.com/theupdateframework/tuf/issues/322
     #
@@ -337,7 +337,7 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
 
 
 
-    # Test 1 Begin:
+    # Test 2 Begin:
     #
     # 'timestamp.json' specifies the latest version of the repository files.
     # A client should only accept the same version of this file up to a certain
@@ -349,16 +349,13 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
     # TUF client. The TUF client performs a refresh of top-level metadata, which
     # includes 'timestamp.json'.
     
-    timestamp_path = os.path.join(self.repository_directory, 'metadata',
-                                  'timestamp.json')
-    
     # Modify the timestamp file on the remote repository.  'timestamp.json'
     # must be properly updated and signed with 'repository_tool.py', otherwise
     # the client will reject it as invalid metadata.  The resulting
     # 'timestamp.json' should be valid metadata, but expired (as intended).
     repository = repo_tool.load_repository(self.repository_directory)
  
-    key_file = os.path.join(self.keystore_directory, 'timestamp_key') 
+    key_file = os.path.join(self.keystore_directory, 'timestamp_key')
     timestamp_private = repo_tool.import_rsa_privatekey_from_file(key_file,
                                                                   'password')
 
--- a/tests/test_indefinite_freeze_attack.py
+++ b/tests/test_indefinite_freeze_attack.py
@@ -180,13 +180,26 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
 
 
   def test_without_tuf(self):
-    # Test 1:
-    # Scenario:
+    # Two tests are conducted here.
+    # Test 1: If an expired timestamp is downloaded, is it recognized as such?
+    # Test 2: If we find that the timestamp acquired from a mirror indicates
+    #         that there is no new snapshot file, and our current snapshot
+    #         file is expired, is it recognized as such?
+
+    # Test 2 Begin:
+    #
+    # See description of scenario in Test 2 in the test_with_tuf method.
+    # Without TUF, Test 2 is tantamount to Test 1, and so we skip this test.
+
+
+    # Test 1 Begin:
+    #
     # 'timestamp.json' specifies the latest version of the repository files.
     # A client should only accept the same version of this file up to a certain
     # point, or else it cannot detect that new files are available for download.
     # Modify the repository's timestamp.json' so that it expires soon, copy it
     # over the to client, and attempt to re-fetch the same expired version. 
+    #
     # A non-TUF client (without a way to detect when metadata has expired) is
     # expected to download the same version, and thus the same outdated files.
     # Verify that the same file size and hash of 'timestamp.json' is downloaded.
@@ -225,11 +238,6 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
     # Verify 'download_fileinfo' is equal to the current local file.
     self.assertEqual(download_fileinfo, fileinfo)
 
-    # Test 2:
-    # See description of scenario in Test 2 in the test_with_tuf method.
-    # Without TUF, Test 2 is tantamount to Test 1, and so it is not tested
-    # again here.
-
 
   def test_with_tuf(self):
     # Two tests are conducted here.
@@ -237,9 +245,11 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
     # Test 2: If we find that the timestamp acquired from a mirror indicates
     #         that there is no new snapshot file, and our current snapshot
     #         file is expired, is it recognized as such?
+
+
+    # Test 2 Begin:
     #
-    # Test 2 addresses this issue:
-    #   https://github.com/theupdateframework/tuf/issues/322
+    # Addresses this issue: https://github.com/theupdateframework/tuf/issues/322
     #
     # If time has passed and our snapshot (or any targets role) is expired, and
     # the mirror whose timestamp we fetched doesn't indicate the existence of a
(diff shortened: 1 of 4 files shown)


awwad:
Tests all look good to me. Running full tox again on 2.7 and 3.4. LMK if you think this is OK, Vlad?

tests/test_indefinite_freeze_attack.py
```diff
@@ -15,6 +15,14 @@
     than verifying text output), use pre-generated repository files, and
     discontinue use of the old repository tools. -vladimir.v.diaz
 
+  March 9, 2016.
+    Additional test added relating to issue:
+    https://github.com/theupdateframework/tuf/issues/322
+    If a metadata file is not downloaded (no indication of a new version
```
vladimir-v-diaz: I'm not sure if this should instead say, "... a metadata file is not updated ..." or "... the expiration of the locally trusted metadata should be detected."

tests/test_indefinite_freeze_attack.py
```diff
@@ -215,9 +224,108 @@ def test_without_tuf(self):
...
+    # the software update system / application / user. This test creates that
+    # scenario. The correct behavior is to raise an exception.
+    #
+    # Background: Expiration checks were previously conducted
+    # (ensure_not_expired) when the metadata file was downloaded. If no new
+    # metadata file was downloaded, no expiry check would occurs. (Exception:
+    # root was checked for expiration at the beginning of each refresh() cycle,
+    # and timestamp was always checked because it was always fetched.) Snapshot
+    # and targets were never checked if the user does not have evidence that
+    # they have changed.
```
vladimir-v-diaz: I think this comment block would be more helpful to have (or duplicated) in `updater.py`.  The <Purpose> section of refresh() would be an ideal location.

tests/test_indefinite_freeze_attack.py
```diff
@@ -215,9 +224,108 @@ def test_without_tuf(self):
...
+    #   https://github.com/theupdateframework/tuf/issues/322
+    #
+    # If time has passed and our snapshot (or any targets role) is expired, and
+    # the mirror whose timestamp we fetched doesn't indicate the existence of a
+    # new snapshot version, we still need to check that it's expired and notify
+    # the software update system / application / user. This test creates that
+    # scenario. The correct behavior is to raise an exception.
+    #
+    # Background: Expiration checks were previously conducted
+    # (ensure_not_expired) when the metadata file was downloaded. If no new
```
vladimir-v-diaz: To make it easier to locate for readers unfamiliar with the code, consider specifying the module or full function name: `updater._ensure_not_expired()`.

tests/test_indefinite_freeze_attack.py
```diff
@@ -215,9 +224,108 @@ def test_without_tuf(self):
...
+    #I think that I only expect tuf.ExpiredMetadata error here. A
+    # NoWorkingMirrorError indicates something else in this case - unavailable
+    # repo, for example.
+    else:
+      self.fail("TUF failed to detect expired stale snapshot metadata. Freeze.")
+
+
+
+
+    # Part 1:
```
vladimir-v-diaz: These labels are a bit confusing (what is Part 1?  Is this test 1?).  It's not immediately obvious where test 1 ends and test 2 begins, but I think Test 2 is started first (which is unexpected).

tests/test_indefinite_freeze_attack.py
```diff
@@ -215,9 +224,108 @@ def test_without_tuf(self):
...
+
+    except tuf.ExpiredMetadataError as e:
+      logger.info("Test: Refresh #2 - failed as expected. Stale-expired "
+                  "snapshot case generated a tuf.ExpiredMetadataError exception"
+                  ". Test pass.")
+    #I think that I only expect tuf.ExpiredMetadata error here. A
+    # NoWorkingMirrorError indicates something else in this case - unavailable
+    # repo, for example.
+    else:
+      self.fail("TUF failed to detect expired stale snapshot metadata. Freeze.")
```
vladimir-v-diaz: It might be clearer to use the word "locally" here as well. The other case would be when the server serves the client stale metadata.

To be more precise: "Freeze attack was successful" instead of just "Freeze"

tests/test_indefinite_freeze_attack.py
vladimir-v-diaz: Test 1 / Part 1 can be cleaned up a bit (e.g., we don't have to reload the repository, reload the keys for the Timestamp role, etc.)

tuf/client/updater.py
```diff
@@ -643,6 +643,13 @@ def refresh(self, unsafely_update_root_if_necessary=True):
         logger.info('An expired Root metadata was loaded and must be updated.')
         raise
 
+    # If an exception is raised during the metadata update attempts, we will
+    # attempt to update root metadata once by recursing with a special argument
+    # to avoid further recursion. We use this bool and pull the recursion out
+    # of the except block so as to avoid unprintable nested NoWorkingMirrorError
```
vladimir-v-diaz: I'm curious as to why we are getting an unprintable NoWorkingMirrorError exception.  Is this a bug?  I'm not sure if the `retry_once` boolean is a workaround for the unprintable exceptions.

tests/test_indefinite_freeze_attack.py
awwad: Updated that: 
https://github.com/awwad/tuf/commit/8adab689ffb78cdff92d980347a2da93f2250222

I think that should be better. LMK if you think it needs more work.

tests/test_indefinite_freeze_attack.py
```diff
@@ -171,12 +180,26 @@ def tearDown(self):
 
 
   def test_without_tuf(self):
-    # Scenario:
+    # Two tests are conducted here.
```
vladimir-v-diaz: Technically, Test 2 is skipped.  But I see why you'd want to acknowledge the two possible cases for stale metadata.

tuf/client/updater.py
vladimir-v-diaz: Instead of using a second boolean (`unsafely_update_root_if_necessary` being the first one) to take care of retries, would it work to stick the following in the two places that `retry_once` is used?

```Python
self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH)
self.refresh(unsafely_update_root_if_necessary=False)
```

I find it confusing to use two boolean variables for the same thing, unless its an issue with printing nested exceptions.  If it's the latter, I'd like to fix the source of the problem.

awwad: I'll try to reproduce the issue again. Resolved it by moving that out of the except block, but perhaps there's a better way.

Was this:

```
ERROR: test_with_tuf (__main__.TestIndefiniteFreezeAttack)
  File "test_indefinite_freeze_attack.py", line 284, in test_with_tuf
  File "/home/s/w/tuf/tuf/client/updater.py", line 699, in refresh
	self.refresh(unsafely_update_root_if_necessary=False)
  File "/home/s/w/tuf/tuf/client/updater.py", line 651, in refresh
	referenced_metadata='timestamp')
  File "/home/s/w/tuf/tuf/client/updater.py", line 1674, in _update_metadata_if_changed
	expected_versioninfo['version'], compression)
  File "/home/s/w/tuf/tuf/client/updater.py", line 1468, in _update_metadata
	compression_algorithm)
  File "/home/s/w/tuf/tuf/client/updater.py", line 1189, in _get_metadata_file
	raise tuf.NoWorkingMirrorError(file_mirror_errors)
tuf.NoWorkingMirrorError: <unprintable NoWorkingMirrorError object>

Ran 2 tests in 16.959s

FAILED (errors=1)

```

vladimir-v-diaz: Can you please verify that _get_metadata_file() is correctly using logger.exception()?

See: https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py#L1340-L1343

logger.exception() should only be called from an exception handler.  Python's documentation: https://docs.python.org/2/library/logging.html#logging.Logger.exception

awwad: Ah.

It's incorrect here:
https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py#L1152-L1155

awwad:
Sidenote: there is not enough information in the code about why "unsafely update root if necessary" is unsafe.

vladimir-v-diaz:
I tried to explain here: https://github.com/awwad/tuf/blob/detect_expiry_322_clean/tuf/client/updater.py#L620-L624

Previously, "safely" fetching root.json required knowing its hash and file size (as specified in snapshot.json).  Now, safely fetching root.json is knowing its exact version number.  A root.json is unsafely fetched without knowing its version number (i.e.,, snapshot.json might have expired, thus the clients is unsure of the currently trusted version).

awwad:
tuf.client.updater.refresh(): Removed the retry_once stuff, waiting for a fix to the python 3 string issue causing unprintable exceptions.

test_indefinite_freeze_attack: Removed forced failure conditions that I'd left in for debugging the string issue. All should be in order again. As for [this comment on cleaning up the test](https://github.com/theupdateframework/tuf/pull/324#discussion_r56023657), I've tried to cut out pieces, but end up with behavior that surprises me. It may be that I'm failing to understand something.

For example, within test_with_tuf, [for test 2, the signing key for timestamp is loaded again at current line 362, even though it was loaded earlier at current line 278](https://github.com/awwad/tuf/blob/detect_expiry_322_clean/tests/test_indefinite_freeze_attack.py#L362). When I comment out the second loading, I get a failure to write the repository at [line 371](https://github.com/awwad/tuf/blob/detect_expiry_322_clean/tests/test_indefinite_freeze_attack.py#L371) with a complaint that there are not enough signatures to sign timestamp metadata. This leads me to guess that loading the repository again causes us to lose the timestamp keys. (Is that expected?) So I try commenting out [the repository reload at 356](https://github.com/awwad/tuf/blob/detect_expiry_322_clean/tests/test_indefinite_freeze_attack.py#L356), but somehow that leads to a different complaint at [line 371's repository.write() call](https://github.com/awwad/tuf/blob/detect_expiry_322_clean/tests/test_indefinite_freeze_attack.py#L371): that the root keys aren't loaded. The root keys *aren't* loaded, but they weren't necessary anywhere in this test previously (if I reloaded the repository and the timestamp keys). Why are the root keys necessary in this case?

awwad:
I've created [a branch on my repo that reliably produces the unrelated error of the possible bug noted](https://github.com/awwad/tuf/commit/76f5ca5806de30f57424503599831b4cb413d51e) and will in the future address it in a separate issue & PR.

Vlad, please let me know if there's anything left to do with this PR.

vladimir-v-diaz:
I think the only remaining issue is the unprintable exception.
You can override __unicode__() (as we do with __str__()) so that exception messages are correctly printed in Python 3.

See:
https://github.com/awwad/tuf/blob/detect_expiry_322_clean/tuf/__init__.py#L332

awwad:
It wasn't a need for __unicode__(), but we figured it out. There were two bugs in tuf/__init__.py class NoWorkingMirrorError. 
See c4ef697 (use of dictionary.iteritems() not possible in python 3) and 7cd20fe (uninstantiated logging object).

awwad:
Hokay. NOW I think this PR is done. Hehehe.

vladimir-v-diaz:
I just recalled that import of `tuf.log` and `logging` were removed here:
https://github.com/theupdateframework/tuf/commit/1c51b8d732d553e87a3b4555b369294c75f2389e

That would explain the uninstantiated logging object.  I verified that we didn't break the commit above.

vladimir-v-diaz:
One reason why we might have missed the exceptions raised in __init__(): http://stackoverflow.com/questions/25374679/exception-in-python-exception-classs-str-missed

awwad:
Yeah, exceptions in __str__() are sneaky.

awwad:
WRT [logging removal](https://github.com/theupdateframework/tuf/pull/324#issuecomment-197900259):
Right, I remember some traffic about this a while back. Should I remove the logging stuff again?

vladimir-v-diaz:
The issue that Alex fixed only occurs if you import tuf.log in __init__().  We now only import `logging`, so it should be fine.  I tried to import 'tuf.conf', and the 'tuf.log' file is not created (as we wish).

vladimir-v-diaz:
Correction: only occurs if you import `log.py`

awwad:
Messed around with it and it seems OK to me on edge cases, too. Okay, so nothing else I'm aware of. I leave it to you. ^_^

vladimir-v-diaz:
I reviewed the changes and made edits here: https://github.com/theupdateframework/tuf/commit/dd8a7eb0dd155cdb672e27d1d86d4cc07064d77b

Please merge those changes with this pull request (if all my changes are good to go).

PR #325, opened 2016-03-17
Awwad detect expiry 322 clean

### LINKED ISSUE #322: Client: Freeze attack issue - expiry of metadata may not be detected

vladimir-v-diaz:
This fixes Issue #322. Please see that issue!
It replaces PR #323, which got too messy.

Update:
This PR also:

    :bug::bug: Fixes two additional bugs issues identified below, both causing unprintable NoWorkingMirrorError exceptions - one in python 3 only and the other in either 2 or 3 with certain mirror name issues.
    adds .DS_Store to .gitignore

This started as a manual squash commit of detect_expiry_322, to avoid merge conflicts.

Comment lines from the individual commits include:
1. Fix #322 by detecting expiry of stale files. initial attempt
2. temp commit of files from Soma
3. removing freeze_attack_stale_expiry and leaving the test added to indefinite freeze attack
4. fixing indefinite freeze attack test: now incorporates old reject-freshly-downloaded-but-expired-timestamp test as well as reject-stale-already-present-but-expired-snapshot test
5. small refinements to indefinite freeze attack test
6. Pulled the recursion out of the except block in refresh() to avoid unprintable nested exceptions.
7. Added comments to the last commit (retry_once)
8. Merge pull request #1 from awwad/detect_expiry_322_temp (removing cruft in another branch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ build/*
 .coverage
 .tox/*
 tests/htmlcov/*
+.DS_Store
--- a/tests/test_indefinite_freeze_attack.py
+++ b/tests/test_indefinite_freeze_attack.py
@@ -181,18 +181,18 @@ class TestIndefiniteFreezeAttack(unittest_toolbox.Modified_TestCase):
 
   def test_without_tuf(self):
     # Two tests are conducted here.
-    # Test 1: If an expired timestamp is downloaded, is it recognized as such?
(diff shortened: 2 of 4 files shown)


awwad:
For the full comment history on this pull request, see [this PR](https://github.com/theupdateframework/tuf/pull/324).

PR #326, opened 2016-04-06
Remaining TUF 1.0 tasks

vladimir-v-diaz:
Implementation of remaining TUF 1.0 tasks:

* redesign of repository tool
* support role delegations that resemble a graph
* store all delegations in one directory
* update documentation to follow the new behaviour of the repository tool
* Prevent freeze attacks where a client if left with stale, local metadata
* Update repository data (used in tests) following all of the changes
- [ ] WIP (Minor refactor of developer tool)


trishankkarthik:
@JustinCappos I don't mean to make life more difficult, but could we flesh out Section 5.1 (The client application) a bit more?

Right now, there are not enough details for an independent programmer to implement a TUF client using that section. The programmer has to refer to the Python reference implementation.

I think it'd be a good idea to flesh out more details, because the UPTANE RFC will have skip over the TUF bits for brevity.

I'd love to contribute as I find the space and time, but I can't commit to it right now.

vladimir-v-diaz:
I think this is something Sebastien and I can work on, since it overlaps with changes he's making to support the Director role for Uptane (just finished discussing it in person).  We can work on a more detailed client workflow for TUF 1.0, and then Sebastien can expand it for the changes needed for Uptane.

JustinCappos:
Okay, I think this is a good plan.

vladimir-v-diaz:
Note: I created a "master" branch and we'll edit the "develop" branch (containing the latest 1.0 changes) directly.

Issue #340, opened 2016-07-15
Public Log Proposal


ecordell:
This proposes a change to the TUF spec to support a tamper-evident public log of changes to the root authority in a TUF repository and to allow trust-pinning for delegated roles.

There is further discussion of this idea at https://github.com/docker/notary/issues/458 and https://github.com/docker/notary/issues/835
## Motivation

From the TUF spec:

>   To replace a compromised root key or any other top-level role key, the root
>    role signs a new root.json file that lists the updated trusted keys for the
>    role.  When replacing root keys, an application will sign the new root.json
>    file with both the new and old root keys until all clients are known to have
>    obtained the new root.json file (a safe assumption is that this will be a
>    very long time or never).

Currently, rotating the root keys requires signing with old keys indefinitely. Although this does successfully allow rotation, supporting even moderately aggressive rotation schemes (say, having a root threshold of 2/3 available keys and rotating those keys quarterly) results in a large number of old keys that need to be kept around.

Note that this may be less of a concern in TUF as-specified:

> Periodically, the software update system using the framework
> instructs the framework to check each repository for updates.

If clients can be assumed to be constantly polling, there may be an opportunity to remove older keys. But this is not the case for package managers in general - current partial/full implementations in PyPI, RubyGems, and Notary do not have "periodic" polling as one of their properties.
## Proposal

The canonical (server) TUF metadata repository should version the root.json file any time root keys change, and should point to a "parent" root.json.

```
{
    "signed": {
        "_type": "Root",
        // ...
        "parent": "123"  // sha256 of previous root.json
    },
    "parentSignatures" {
        // current metadata signed with parent keys/threshold
    },
    "signatures": [
        // current metadata signed with current keys/threshold
    ]
}
```

When a client gets new metadata with a `parent` and `parentSignatures` block, it does the following:
- requests the parent metadata file from the server by hash
- verifies the hash
- using the public keys and threshold of the parent, calculates the signatures in `parentSignatures` of the new metadata
- repeats the process if the parent also has a `parent` block, until either the hash matches the current local metadata or the pulled metadata doesn't have a parent block.
### Example

Rotating a single root key from `A` to `B`
- switch public key in public portion to point to `B.pub`
- calculate hash of original metadata and put it in the `parent` block of the new metadata
- sign new metadata with `B.key` and put in `signatures` block.
- sign new metadata with `A.key` and put in `parentSignatures` block.
- publish new metadata
- old metadata should still be accessible by hash (not name)
[...]

PR #341, opened 2016-07-21
Add root versioning for root key rotation

ecordell:
This addresses #340 by adding versioning to the root role.

This doesn't currently enforce that new versions of the root be signed by a threshold of previous root keys - should that be optional or enforced?


JustinCappos:
>
> This doesn't currently enforce that new versions of the root be signed by
> a threshold of previous root keys -

should that be optional or enforced?


If the new root file isn't signed by a threshold of the previous root keys,
it won't be trusted by clients.   Perhaps the documentation should explain
this...

Thanks,
Justin

ecordell:
@vladimir-v-diaz  Updated with root versioning enforced on both the repo (unless using `write_partial()` and the client updater. See `test_updater_root_rotation.py` for client behavior.

I think we can open a new issue for trust pinning later. Thanks!

ecordell:
@vladimir-v-diaz @JustinCappos Any thoughts on this? Would it help for me to go through and comment on the diffs to call out particular changes?

vladimir-v-diaz:
Your changes are pretty clear.  No need to comment on the diffs.

I will merge this (and resolve any potential conflicts) after making a few changes of my own to affected modules (issues found while working on remaining code coverage).

Good job on implementing issue #341.

jakedt:
Hey @vladimir-v-diaz, any update on this? Last update was 25 days ago. Evan is currently working on rebasing, but after that is there any hangup to getting this in this week?

vladimir-v-diaz:
Sorry for the delay.  Rebasing any minor conflicts would be appreciated.  We are currently making a major change to how metadata is loaded/written to disk, but I can resolve any issues with this PR after the change and then merge.  This change and Trust Pinning (see below) should be completed this week in time for a demo.

Note: We are also working to finalize the Trust Pinning proposal and implementation.
https://github.com/theupdateframework/taps/blob/master/tap-0001.md
https://github.com/theupdateframework/taps/blob/master/tap-0002.md

Trust Pinning TAP: https://github.com/awwad/taps/tree/tap-0004-trust-pinning

ecordell:
@vladimir-v-diaz I don't see any public discussion to chime in on, but I do have some suggestions w.r.t. the metadata storage. In particular I think there should be an abstraction layer such that the tests can write to the memory instead of disk for the tests.

ecordell:
@vladimir-v-diaz @JustinCappos I've rebased and fixed all of the tests related to doing so

vladimir-v-diaz:
Thanks for keeping this pull request current with the "develop" branch.  I am done implementing our changes (locally) and currently resolving the new conflicts with this pull request.  If something is unclear to me, I'll ask questions here.

tests/test_key_revocation.py
```diff
@@ -385,9 +388,6 @@ def test_root_key_revocation(self):
       for mirror_exception in exception.mirror_errors.values():
         self.assertTrue(isinstance(mirror_exception, tuf.BadSignatureError))
 
-    # Load the previous Root signing key so that the the client can update
-    # successfully.
-    repository.root.load_signing_key(self.role_keys['root']['private'])
```
vladimir-v-diaz: Haven't finished looking at the other changes carefully, but why was this moved to lines 370 - 372?  The purpose of this test condition (starting on line 388) is to verify that the client rejects the root file because it doesn't trust any of its keys (the root key hasn't signed it yet).

tuf/repository_lib.py
```diff
@@ -1980,14 +2019,15 @@ def write_metadata_file(metadata, filename, version_number,
...
     dirname, basename = os.path.split(written_filename)
-    
     basename = basename.split(METADATA_EXTENSION, 1)[0]
     version_and_filename = str(version_number) + '.' + basename + METADATA_EXTENSION 
     written_consistent_filename = os.path.join(dirname, version_and_filename)
 
-    logger.info('Linking ' + repr(written_consistent_filename))
-    os.link(written_filename, written_consistent_filename)
+    #logger.info('Linking ' + repr(written_consistent_filename))
+    #os.link(written_filename, written_consistent_filename)
```
vladimir-v-diaz: These two lines are commented out.  Why is this a copy, rather than a hard link?

tests/test_root_versioning.py
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env python
+
+"""
+<Program Name>
+  test_root_versioning.py
```
vladimir-v-diaz: We currently have tests for specific modules (e.g., test_keydb.py) and integration tests / attacks (test_replay_attack.py, i.e., they end in "attack").  Not sure how we should handle the additional tests of this PR.

tuf/repository_lib.py
ecordell: When making changes, always hard linking old versions to the current will result in all previous versions (1.root, 2.root, etc) pointing to the current (root.json). 

Writing the consistent version and then linking the current to that (e.g. linking root.json -> 3.root.json) should work, but requires more changes because now "write current" doesn't happen in every case. Holding two copies is a small constant increase in storage (#versions -> #versions + 1)

Either solution is fine, I left the commented code to remind myself about the issue and clearly forgot to come back to it!

tuf/repository_lib.py
```diff
@@ -121,11 +121,18 @@ def _generate_and_write_metadata(rolename, metadata_filename, write_partial,
...
   # Generate the appropriate role metadata for 'rolename'. 
-  if rolename == 'root':
+  if rolename == 'root': 
+    tuf.roledb.update_roleinfo(rolename, roleinfo)
+
     metadata = generate_root_metadata(roleinfo['version'],
-                                      roleinfo['expires'], consistent_snapshot,
+                                      roleinfo['expires'], 
+                                      previous_keyids,
+                                      previous_threshold,
```
vladimir-v-diaz: The `previous_keyids` and `previous_threshold` parameters are not actually used in `generate_root_metadata()`

tuf/repository_lib.py
vladimir-v-diaz: We probably want to support the second solution (for clients / devices with limited space).  Creating a true copy or a hard link should also probably be an option (i.e., tuf.conf.consistent_method = copy (or link)).

I'll try to address before merging.

tuf/repository_lib.py
```diff
@@ -121,11 +121,18 @@ def _generate_and_write_metadata(rolename, metadata_filename, write_partial,
...
   # attributes, such as version number, expiration, etc.
   roleinfo = tuf.roledb.get_roleinfo(rolename) 
-
+  previous_keyids = roleinfo.get('previous_keyids', [])
+  previous_threshold = roleinfo.get('previous_threshold', 1)
+  
   # Generate the appropriate role metadata for 'rolename'. 
-  if rolename == 'root':
+  if rolename == 'root': 
+    tuf.roledb.update_roleinfo(rolename, roleinfo)
```
vladimir-v-diaz: Not sure why `update_roleinfo()` is needed here.  `roleinfo` isn't updated in any of previous lines.

```
  roleinfo = tuf.roledb.get_roleinfo(rolename) 
  previous_keyids = roleinfo.get('previous_keyids', [])
  previous_threshold = roleinfo.get('previous_threshold', 1)
  
  # Generate the appropriate role metadata for 'rolename'. 
  if rolename == 'root': 
    tuf.roledb.update_roleinfo(rolename, roleinfo)
```

`update_roleinfo()` isn't called like this any of the if-clauses for the other top-level rolenames.

tuf/repository_lib.py
```diff
@@ -198,13 +206,28 @@ def _generate_and_write_metadata(rolename, metadata_filename, write_partial,
...
+  def should_write():
+    # Always write if partial writing
+    if write_partial:
+      return True
+    
+    # In the normal case, we should write metadata if the threshold is meet
+    write = tuf.sig.verify_threshold(signable, roleinfo['threshold'], roleinfo['signing_keyids'], rolename)
+    
+    # The root must also be signed by previous root keys and threshold
+    if rolename == 'root' and len(previous_keyids) > 0 and previous_threshold is not None:
```
vladimir-v-diaz: I think `previous_threshold` might never be set `None`.

According to the following line (on line 125), it can either be 1 or a valid threshold number enforced by tuf.roledb:
```previous_threshold = roleinfo.get('previous_threshold', 1)```

File: tuf/repository_lib.py [hunk 7]
```diff
@@ -1869,8 +1905,11 @@ def sign_metadata(metadata_object, keyids, filename):
     if key['keytype'] in SUPPORTED_KEY_TYPES:
       if 'private' in key['keyval']:
         signed = signable['signed']
-        signature = tuf.keys.create_signature(key, signed)
-        signable['signatures'].append(signature)
+        try:
+          signature = tuf.keys.create_signature(key, signed)
+          signable['signatures'].append(signature)
+        except:
```
vladimir-v-diaz: We should catch specific exception(s) here, or at least log the caught exception if the bare `except` is used.

tuf/roledb.py
```diff
@@ -124,6 +124,8 @@ def create_roledb_from_root_metadata(root_metadata, repository_name='default'):
     if rolename == 'root':
       roleinfo['version'] = root_metadata['version']
       roleinfo['expires'] = root_metadata['expires']
+      roleinfo['previous_keyids'] = root_metadata.get('signing_keyids', [])
+      roleinfo['previous_threshold'] = root_metadata.get('threshold', 1)
```
vladimir-v-diaz: `root_metadata` doesn't have a 'signing_keyids' or `threshold` entry.  I think `roleinfo.get('keyids')` and `roleinfo.get('threshold')` are needed here instead.

See https://github.com/theupdateframework/tuf/pull/341/files#diff-43de4bd3342179b5413e1971566ed4cfR123
and https://github.com/theupdateframework/tuf/blob/develop/tests/repository_data/repository/metadata/root.json#L58-L82

tests/test_key_revocation.py
ecordell: I'll think about how to improve this test and update. Without changing the location, the updater can't update to the next version at all - signing a version without a previous root key breaks the chain for all subsequent versions.

tests/test_root_versioning.py
ecordell: Happy to move this around, however there are other files like `test_key_revocation` for example.

tuf/repository_lib.py
ecordell: Will fix - I think this is just an artifact of rebasing so much

tests/test_key_revocation.py
vladimir-v-diaz: I am in the process of integrating this branch with some design changes and resolving merge conflicts.  I'll try take care of any remaining issues and let you know of any problems that remain.

tests/test_root_versioning.py
vladimir-v-diaz: I guess we can append `_integration` to these types of tests, including `test_key_revocation.py`

Examples:
test_keydb.py (unit test)
test_replay_attack.py (attack test)
test_root_versioning_integration.py (integration test)
test_key_revocation_integration.py (integration test)

tests/test_key_revocation.py
vladimir-v-diaz: FYI
To prevent breaking the chain of subsequent versions of root, here is the solution that I implemented...

(1) Add a method to allow individual roles to be written and (2) add a flag to this method to control whether version numbers are automatically incremented.  In this way, we can write the temporary root file (with the missing, required root key) for testing purposes and then regenerate the valid root file with the expected version number and signatures.

For example (Note, some steps removed to keep the example small):
```Python
repository.root.remove_verification_key(public_root_key)
repository.writeall()
# The writeall() above automatically increments root to version 2 (supposing it's
# currently at version 1.

# Verify that the client rejects the invalid root file....
try:
  updater.refresh()
except tuf.BadSignatureError:
  ...

# Now regenerate a valid root file, where its version number = 2
repository.root.add_verification_key(public_root_key)
repository.write('root', increment_version_number=False)                    
repository.write('snapshot')                                                
repository.write('timestamp')

# Successfully refresh root...
updater.refresh()
self.assertEqual(updater.metadata['current']['root']['version'], 2)
```

tuf/client/updater.py
```diff
@@ -650,89 +650,73 @@ def refresh(self, unsafely_update_root_if_necessary=True):
...
       # The caller explicitly requested not to unsafely fetch an expired Root.
       else:
         logger.info('An expired Root metadata was loaded and must be updated.')
         raise
 
-    # If an exception is raised during the metadata update attempts, we will
-    # attempt to update root metadata once by recursing with a special argument
-    # (unsafely_update_root_if_necessary) to avoid further recursion.
 
+    self._update_root_metadata(root_metadata)
```
vladimir-v-diaz: This is a major change to the update process.  Ideally, this sort of change should be mentioned in the PR summary and reflected in the proposal and TUF specification.

The docstring should also be updated to state that root.json is now always updated before the other top-level roles.  I'll take care of updating the documentation.  I was debugging a problem and was confused as to why the update process was behaving this way.  Hadn't noticed this one line change to refresh().

tuf/client/updater.py
```diff
@@ -650,89 +650,73 @@ def refresh(self, unsafely_update_root_if_necessary=True):
...
+    new_root_metadata = \
+      tuf.util.load_json_string(new_root_metadata_file.read().decode('utf-8'))
+    new_version = new_root_metadata['signed']['version'] + 1
+    
+    for version in range(next_version, new_version):
+      # Temporarily set consistent snapshot. Will be updated to whatever is set in the latests root.json after 
+      # running through the intermediates with `_update_metadata`
+      self.consistent_snapshot = True
+      self._update_metadata('root', tuf.conf.DEFAULT_ROOT_REQUIRED_LENGTH, version=version, 
+                            compression_algorithm=compression_algorithm)
```
vladimir-v-diaz: According to the new update process, top-level metadata is now downloaded in `root -> timestamp -> snapshot -> targets -> (delegations)` order.  When updating root, the client updates root from the current version up to the latest (including intermediate versions).  Supposing the current version = 1 and the latest version = 5, the client fetches 1.root.json -> 2.root.json -> ... -> 5.root.json.  Currently, root.json is downloaded if any of the top-level metadata cannot be successfully updated, and immediately (if needed) after updating snapshot.json.  The snapshot.json role lists root.json, and timestamp.json provides timeliness for all the other top-level metadata.

As I understand the modified update process, lines 707 - 719 attempt to fetch the latest root to determine the `new_version`.  One problem I see with this approach is that the fetched root file is not verified.  Is this really the latest version?  Since root is not updated after snapshot.json nor its latest version referenced from snapshot.json, I assume the "latest" version of root fetched remotely is treated as up-to-date, until it expires?

tuf/client/updater.py
ecordell: Sorry about that - yeah, this was necessary when implementing but I forgot to update the spec.

tuf/client/updater.py
```diff
@@ -650,89 +650,73 @@ def refresh(self, unsafely_update_root_if_necessary=True):
...
+    
+      <Returns>
+        None.
+    """
+    next_version = current_root_metadata['version'] + 1
+    new_root_metadata_file = self._get_metadata_file('root', 'root.json', tuf.conf.DEFAULT_ROOT_REQUIRED_LENGTH, None, 
+                                                     compression_algorithm=compression_algorithm)
+    new_root_metadata = \
+      tuf.util.load_json_string(new_root_metadata_file.read().decode('utf-8'))
+    new_version = new_root_metadata['signed']['version'] + 1
```
ecordell: Yes, that's why and I agree with your suggestion.

tuf/client/updater.py
ecordell: If the client retrieves what it thinks is the newest root (because of the version listed under `root.json`) but a newer/older version is referenced in the snapshot, the client update process should fail (and will). This means that either the snapshot file hasn't been updated to reference the newest root (which would make it invalid) or there is a newer root available (because of a race condition of someone updating root after you started the update process).

I think it's fine as it stands; but TAP5 changes this behavior, correct?

tuf/client/updater.py
```diff
@@ -1205,6 +1189,19 @@ def _get_metadata_file(self, metadata_role, remote_filename,
       logger.error('Failed to update {0} from all mirrors: {1}'.format(
                        remote_filename, file_mirror_errors))
       raise tuf.NoWorkingMirrorError(file_mirror_errors)
+    
+    
+  def _verify_root_chain(self, role, current, next):
```
ecordell: That's a better name, I agree. Perhaps `_verify_root_chain_link`?

tuf/repository_lib.py
ecordell: Agreed

ecordell:
Thanks for the review @vladimir-v-diaz! I was under the impression you were going to rebase this - would you prefer I do that and make the changes you suggested?

vladimir-v-diaz:
I have made most of the changes to my local branch (includes other design changes).  I will push everything to my remote branch (once I tie up loose ends and have unit tests passing) so you can take a look at what I've changed.

tuf/client/updater.py
vladimir-v-diaz: TAP5 does propose that root.json not be listed in snapshot.json, but it hasn't been finalized yet and is currently undergoing changes.    We'll deal with any of the changes proposed in TAP5 separately.

I'll keep root.json in snapshot.json for this PR.

I am not sure that root.json is properly updated by snapshot.json.  Here is how top-level
metadata is updated in refresh():

```Python
self._update_root_metadata(root_metadata)                                                                               
# Use default but sane information for timestamp metadata, and do not       
self._update_metadata_if_changed('snapshot',
          referenced_metadata='timestamp')           
self._update_metadata_if_changed('targets')
```

I'll try to create a test condition for this to verify.

ecordell: > I am not sure that root.json is properly updated by snapshot.json

It doesn't update the root again if the conditions I listed hold, it simply fails the update process (the hash in snapshot will not be correct).

PR #342, opened 2016-07-28
Add description of fast-forward attack to SECURITY.md

vladimir-v-diaz:
@JustinCappos I'd like to add a description of the fast-forward attack to the security page, and then reference it from the specification.  Have I generalized and given a clear description of the fast-forward attack?


vladimir-v-diaz:
Also updated the specification to include the fast-forward attack, and some that were missing.

vladimir-v-diaz:
@JustinCappos 
I think you might have missed this notification.

PR #345, opened 2016-08-02
Update README.rst

akanbrown:
I did "pip install libsodium" instead of "pip install libNaCl". New users might make the same mistake.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -179,7 +179,7 @@ that need to generate TUF repository files, such as metadata,
 cryptographic keys, and signatures. Whereas the minimal install can only
 verify ed25519 signatures and is intended for sofware updater clients,
 ``tuf[tools]`` provides repository maintainers secure ed25519 key and
-signature signature verification with `PyNaCl <https://pynacl.readthedocs.io/en/latest/>`_,
+signature verification with `PyNaCl <https://pynacl.readthedocs.io/en/latest/>`_,
 a Python binding to the Networking and Cryptography Library.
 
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
We wish for users to execute `pip install tuf[tools]`. This one command will install multiple dependencies (e.g., 'pycrypto>=2.6.1', 'pynacl>=0.2.3') that the user will need for the repository tools.  

Maybe we should instead say, "... signature verification with Python bindings to the Networking and Cryptography Library" and link to their site.  PyNaCl relies on libsodium, a portable C library, but perhaps we don't need to mention that here.  libsodium is installed during PyNaCl's installation.

README.rst
```diff
@@ -180,7 +180,7 @@ cryptographic keys, and signatures. Whereas the minimal install can only
 verify ed25519 signatures and is intended for sofware updater clients,
 ``tuf[tools]`` provides repository maintainers secure ed25519 key and
 signature verification with `PyNaCl <https://pynacl.readthedocs.io/en/latest/>`_,
-a Python binding to the Networking and Cryptography Library.
+a Python binding to the Networking and Cryptography (NaCl) library.
```
vladimir-v-diaz: Should (NaCl) go after `library`?

akanbrown: I think it's okay as is. That's the format they used on their website:

`PyNaCl: Python binding to the Networking and Cryptography (NaCl) library`

What do you think?

vladimir-v-diaz:
I missed that.  This is okay with me also.  Merging!

PR #349, opened 2016-08-04
Fix all_targets in updater

ecordell:
After the most recent merges, the tests for the updater were broken.


vladimir-v-diaz:
Thanks!

PR #352, opened 2016-08-05
Support Unix shell-style wildcards in trusted paths

vladimir-v-diaz:


vladimir-v-diaz:
Note: Need to verify if we want support for only the wildcard operator (*), or/and also the **? (matches any character)**, **[seq] (matches sequence)**, or **[!seq] (does not match sequence)** operators.

packages/foo*.tar.gz

packages/foo?.1.tar.gz

packages/foo[0-4].1.tar.gz

packages/foo[!1-2].1.tar.gz

I thought these would be okay to support, as well

PR #357, opened 2016-08-08
Update specification to list previous_hash field in root.json [WIP]

vladimir-v-diaz:
@JustinCappos What do you think of the `previous_hash` entry?  Does it successfully support hash-chained histories?  Should we also optionally allow all roles to list it in their metadata? (in my edit to tuf-spec.txt, it is only specified for root.json.)


JustinCappos:
At a minimum, it must be in timestamp too, because I think the Docker folks
need this for their use case.

If it's optional, what happens if it is missing sometimes?  Can one just
add / remove it as they like?

PR #363, opened 2016-08-22
Update README.rst

akanbrown:


vladimir-v-diaz:
Thanks for testing and adding instructions for OS X.

PR #365, opened 2016-08-26
Fixes data loss on unplug the power on download files. If we would no…

julianhille:
…t fsync here and update a metadata and then unplug power we lose data.


julianhille:
This does NOT fix the issue if you unplug the power cable directly after the destination file has been opened with 'wb' and ANY data has been written. It will be empty afterwards

vladimir-v-diaz:
Thanks for these pull requests (including #367) to force downloaded metadata to be written to disk (mentioned in issue #366).

Issue #366, opened 2016-08-26
file corruptions


julianhille:
In https://github.com/theupdateframework/tuf/blob/develop/tuf/util.py#L214 is a copy of the downloaded file instead of a move.
We should consider changing that because:
- This needs twice the disk storage while copying
- it takes longer as necessary
- could lead to file corruptions (under circumstances explained in detail)

Doing a copy is an essential bad idea here. 
If the file downloaded is a metadata file and is saved into a temporary file
the destination file is opened with mode 'wb' afterwards.
This means the very first write to it will empty the file out.
If we now unplug the power cable we are left with a broken repository which could only be repaired with backups or a copy of 'previous' directory to the current.
This is what happens even if we do it without a flush or fsync.
A move / rename would be atomic on that.

Even if the power cable is unplugged AFTER the files are closed we depend on the system / kernel to flush to disk.
This means we could also lose content of files if update was successful and unplug some time later.

At the first glance i could not find any more situations or source code where this may happen, but we should take a closer look.

This PR https://github.com/theupdateframework/tuf/pull/365 will fix the data loss on unplugging the power cable after an update.

julianhille:
We may should consider an empty metadata file or a file which is not json decodable as missing instead of fail completely.

julianhille:
Its not as easy to change the current copy to a real move, as this would not work without change the temp file implementation.
I thought i could set the delete Flag to false on the fly in the move function.
This would not work on windows machines as pythons tempfile lib evaluates this delete flag 
only on once on init for windows systems.
Haven't found a way to "cancel" the OS_TEMPORARY flag for the file handler on windows systems.

trishankatdatadog:
Hi @julianhille! Sorry, been a while, but we're housecleaning, and we're wondering if you'd still like to send a PR to fix this? Thanks very much for your help and time!

julianhille:
I think this has been fixed in the two previous mentioned PRs

trishankatdatadog:
@julianhille I see, but didn't you make two subsequent commits to your branch (https://github.com/conhealth/tuf/commit/3a20925e951056252eaafe68708e0f3997814cf0 and https://github.com/conhealth/tuf/commit/d2129c1030cd8e4a2e04686707d95f7b1483e22d) w/o upstreaming them?

ivanayov:
[...]

PR #367, opened 2016-08-29
Handles broken metadata files gracefully like they would be missing. …

julianhille:
…Should be part of #366


julianhille:
At least this triggers a redownload if current and previous is broken and its not the root.json

Issue #391, opened 2016-11-02
Consistent versions of root appear to be needlessly written multiple times 


vladimir-v-diaz:
write_metadata_file() was modified in PRs #379 and #383 to always write `root.json` and `<version>.root.json` if consistent snapshots = True.  Given these changes, we should refactor the following lines: https://github.com/theupdateframework/tuf/blob/0ce016b7b19a0fc15443eed657a90246ec5df0aa/tuf/repository_lib.py#L213-L224

Log files and a closer inspection of this code show that multiple attempts are made to write root.json.

Thanks given to @FelixWang1994 for bringing this issue to our attention.

vladimir-v-diaz:
This issue was fixed in https://github.com/theupdateframework/tuf/pull/392 and https://github.com/theupdateframework/tuf/pull/394

@FelixWang1994 
Thanks for working on this issue.

FelixWang1994:
Me pleasure!

PR #392, opened 2016-11-02
Fix Issue #391

FelixWang1994:
Whether "consisent_snapshot" is True or False has been considered and handled with in
https://github.com/theupdateframework/tuf/blob/0ce016b7b19a0fc15443eed657a90246ec5df0aa/tuf/repository_lib.py#L2009-L2093


FelixWang1994:
I have fixed Issue #391. 
However, after changing, I notice there may be some mistakes in https://github.com/theupdateframework/tuf/blob/0ce016b7b19a0fc15443eed657a90246ec5df0aa/tuf/repository_lib.py#L239-L260
We may encounter into this "else" branch because of two situations:
1、If 'rolename' is a delegated role, then this is fine.
2、If 'rolename' is a top_level role and partially signed is allowed, then there may be a problem. Firstly, why don't we check whether there are enough signatures here?(e.g. like the "should_write" function in the "if" branch above). Secondly, even there are enough signatures, we do not need to call "write_metadata_file" with explicitly set "consistent_snapshot" true. The reason is the same as why we open this issue.
I don't make any changes for this problem, because I think there may be some reason I haven't realize. Please check if you are free.

vladimir-v-diaz:
I expanded some comment blocks, and added code to write `<version>.root.json` to disk even if `consistent_snapshot` = False. See https://github.com/theupdateframework/tuf/commit/f5e2bec287a1f818d2807f804e33ce9528e939b0

Here are the files that I think are written in this PR:

If `consistent_snapshot` = True
`root.json`, `2.root.json`, `timestamp.json`, `2.timestamp.json`, ...

if `consistent_snapshot` = False
`root.json`, `timestamp.json`, ...

The problem with the above is that `2.root.json` should also be written if consistent_snapshot = False.

Here are the requirements for _generate_and_write_metadata():
(1) If consistent_snapshot = True, write `<version>.rolename.json`, `rolename.json`
(2) If consisent_snapshot = False, write `rolename.json`
(3) `root.json` should always be written
(4) `timestamp.json` should always be written
(5) `<version>.root.json` should always be written

Please verify my changes and let me what you think.  For the other issue you raised (partially written metadata), let's address it in a separate ticket.

FelixWang1994:
That's definitely what I'm going to correct! Thanks!

vladimir-v-diaz:
I made some fixes here: https://github.com/theupdateframework/tuf/commit/f5e2bec287a1f818d2807f804e33ce9528e939b0

vladimir-v-diaz:
I want to verify that you've reviewed https://github.com/theupdateframework/tuf/commit/f5e2bec287a1f818d2807f804e33ce9528e939b0 and that this PR is ready to merge.

FelixWang1994:
Yeah, I have reviewed and it is good for me. Thanks

PR #417, opened 2017-01-11
Fix import simple_settings in basic_client.py

jerrytrieu:
From commit: b793c81739c94dceb4bc0707d633e4da5cea07f8


vladimir-v-diaz:
Thanks for the fix!  I am going to merge this for now, but the `simple-settings` dependency will be removed shortly in an upcoming PR.

Issue #356, opened 2016-08-08
Notary 2015 NCC security audit review


titanous:
Notary commissioned [a security audit](https://github.com/docker/notary/blob/master/docs/resources/ncc_docker_notary_audit_2015_07_31.pdf) of their code last year. Even though it was not of the Python code or specification, there are some items that are worth checking in the reference code and addressing in the specification.
1. _Signature Algorithm Not Matched to Key_ – I believe the signature `method` should be removed from the specification and each key type should only have one associated signature method. This mitigates this potential implementation bug completely.
2. _Compromise of Authority Keys Allows Extended Freeze/Compromise_ – This is an implementation bug. The spec is very clear that the root expiry needs to be checked.
3. _Temporary Access to Timestamp, Snapshot, or Targets Key Enables Persistent Lockout_ – The report makes this sound like an implementation bug, however the spec currently says "Clients MUST NOT replace a metadata file with a version number less than the one currently trusted." It seems like the scenario of an attacker bricking updates by bumping the version to the maximum integer should be considered.

The rest of the issues are specific to Notary and not relevant to the Python implementation or specification.

vladimir-v-diaz:
Thanks for bringing this to our attention.

3. Temporary Access to Timestamp, Snapshot, or Targets Key Enables Persistent Lockout
We've added the "Fast-forward" attack to the specification to specifically address this issue.  Another edit to the specification will be made to clearly outline how one should recover from a fast-forward attack.  https://github.com/theupdateframework/tuf/pull/342

2. Compromise of Authority Keys Allows Extended Freeze/Compromise
We found a similar scenario in which expiration of unchanged metadata files may not be detected.
It was fixed here: https://github.com/theupdateframework/tuf/issues/322

1. Signature Algorithm Not Matched to Key
We'll take a closer look at this.

SantiagoTorres:
I just ran into this issue. Is this worth reviewing?

mnm678:
There was a lot of discussion about the Signature Algorithm Not Matched to Key issue in #425 which was resolved with TAP 9.

trishankatdatadog:
Yeah, didn't we already fix this?

SantiagoTorres:
We did, I will close this issue now.

PR #421, opened 2017-01-27
Signature Algorithm Not Matched to Key -- Issue #356

FelixWang1994:
Solve the first problem for issue #356


vladimir-v-diaz:
Thanks for the pull request, Felix.  We have since decided to remove the signature method from the "sig" field.

FelixWang1994:
Hi Vlad,

You're welcome. Always glad to contribute to TUF.

Best,
Kejian Wang.

Issue #414, opened 2016-12-22
Fast Forward Attack Recovery Necessary (Mercury)


awwad:
#### TLDR
The Mercury paper noted that a fast-forward attack is possible against a TUF repository if the timestamp and snapshot role keys (the typically online keys) are compromised, due to the way that replay attack protection works. These fast-forward attacks should be resolved by resetting the highest-seen-valid-version-number cache on the client side when a change has been made in root.json to the snapshot role after a compromise is resolved. This is not currently done in TUF.

#### Replay attack background:
Replay attacks are prevented by having a client refuse to accept versions of a given metadata role file smaller than the highest that they have ever seen listed for that role in a valid snapshot.json metadata file.)

#### The Fast Forward Attack:
In this "fast forward" attack, the attacker employs compromised snapshot and timestamp keys to sign a snapshot metadata file that sets the version numbers listed in the snapshot metadata above the version number of the most recent version of the metadata files, or perhaps at the maximum version value. This results in a denial of service: clients who accept this snapshot metadata will not accept any metadata files with versions less than these arbitrarily inflated version numbers. (e.g. the malicious snapshot might list the version for root as 2048 when the most recent root version is actually 16.)

#### Recovery:
Once the compromised server is recovered, it is necessary to change the timestamp and snapshot keys since the old keys have been exposed. At this point, in order to allow clients who accepted the malicious snapshot.json file to update metadata again (to recover from the fast forward attack), it is necessary to reset the version numbers saved by the updater as the highest-seen-valid-version-number. (This is an attribute of the `Repository` object, `versioninfo`.)

Note that this must be done *after* signature validation of a new root file, of course, so that attackers without root keys can't cause snapshot reset to enable replay attacks

#### Code Delta
This change should be made to [tuf.client.updater.Updater::_update_metadata(), within the `if` block at the end of the method](https://github.com/theupdateframework/tuf/blob/5ba82fe9ddddc04f9ba094d4744e62733f64fae0/tuf/client/updater.py#L1403-L1405). There, it should compare canonicalized versions of `self.metadata['previous']['root']['roles']['snapshot']` and `self.metadata['current']['root']['roles']['snapshot']`. If they are not equal, we should reset the `versioninfo` cache.

#### Drawbacks
[...]

PR #423, opened 2017-01-27
WIP/ENH: Add fast forward attack tests

SantiagoTorres:
I've added this initial test to try to replicate a fast forward attack. For some reason, the updater doesn't fail. The process is the following:

1.- The repository writes the original metadata (snapshot_version -> 2)
2.- The client obtains that piece of metadata
3.- The repository writes a *fast-forwarded* version of the metadata (snapshot version -> 100)
4.- The client obtains that metadata by syncing (the client loads version 100)
5.- The repository writes a non-forwarded version of the metadata (snapshot_version -> 3)
6.- The client downloads it, and doesn't err on it (loaded snapshot version on the client is now 102)

Is this intended behavior? If so, any suggested way to test the FF attack? should I set this to UINT_MAX?

Ping @vladimir-v-diaz @awwad 

I'll update this as code is written/fixed.


vladimir-v-diaz:
Issue #414 is to implement the ability for a client to recover from a fast
forward attack.  If this unit test is for that, then it is expected to
trick the current implementation (because it isn't implemented yet). If the
client is accepting a version lower than what it currently trusts, then
that's a bug / security issue.  The client should block rollback attacks.

awwad:
So we've figured out why it doesn't raise an exception. Note, though, that the client has not actually updated to the old metadata. In other words, metadata on disk remains correct, but an error is not raised, when it should be.

In your test, this happens:

1. Repo writes version 1 of timestamp and version 1 of snapshot.
2. Client downloads and validates them successfully.
3. Repo writes version 2 of timestamp and version 101 of snapshot. (writeall actually increments the 100 you stick in repo.snapshot.version)
4. Client downloads and validates them successfully.
5. Repo writes version 3 of timestamp and version 3 of snapshot (writeall actually increments the 2 you stick in repo.snapshot.version)
6. Client downloads and validates timestamp successfully, then looks at its indicated version of snapshot and sees that it's NOT GREATER THAN its existing snapshot version, and doesn't download the new snapshot.

The thing I would like to have happen differently is that at step 6 above, the client should raise an exception that explains that any one of the following three things is occurring (and it can't tell which):
1. The mirror contacted has outdated data, which is okay. We haven't updated, therefore.
2. We are contacting a server that is attempting a replay attack.
3. We have previously been fast-forward attacked, and the server we're contacting is either not aware of this or just hasn't issued new root metadata that would allow any affected clients to recover. (Or, in our case at this very moment, such code for that recovery doesn't exist. (: )

vladimir-v-diaz:
Some history...
Timestamp previously specified only the hash(es) and length of Snapshot, and the client knew to download a new Snapshot file if it's hash(es) (listed in Timestamp) changed.  We now list the version number of Snapshot in Timestamp, and therefore can potentially detect replayed metadata in advance.

Something we'd need to decide...
Should we:
(1) throw an exception before we actually download replayed metadata (and then reject), or
(2) check for an updated Snapshot by inspecting its hash(es) in Timestamp and raise a ReplayedMetadataError when we retrieve it.

There might be some merit to approach (2), because we check for version numbers in one part of the code (after we download metadata and not also we we inspect referenced metadata) and ReplayedMetadataError is raised consistently for all roles.

For example, if we were to simulate a replay attack with the Timestamp role, the code would raise ReplayedMetadataError instead of an exception that says that an outdated mirror, or a Fast Forward or Replay attack might be the problem.

JustinCappos:
How do we know the metadata is replayed in case 1?  What info is there in
the timestamp file that says this?

Justin

vladimir-v-diaz:
timestamp.json includes the version number of snapshot.json
```Javascript
"snapshot.json": {
    "hashes": {
     "sha256": "c15b14217a3ad50ae2c136109983b7269d3bda42c4e1e530e30e7b5854b53a11"
    },
    "length": 678,
    "version": 1
}
```
https://github.com/theupdateframework/tuf/blob/develop/tests/repository_data/repository/metadata/timestamp.json#L13-L18

JustinCappos:
So without downloading the snapshot.json, all you know is that the
timestamp *says* the snapshot has an old version.  I suppose if you
download the snapshot.json, you can better detect what is happening in some
cases (such as when the files are served by different parties...

jku:
The issue is valid (and after #1444 I intend to take a stab at it as well) but this specific PR has not seen action in 4 years: closing, please reopen if needed

PR #433, opened 2017-02-27
Update specification

vladimir-v-diaz:
* Add `compression_algorithms`, `terminating`, and `consistent_snapshot` attributes
* Update text describing the format of PATHPATTERN
* Update text for handling overlapping targets between delegations
* Remove notion of a full rolename

Thanks is given to @ecordell (Evan Cordell) and the CoreOS team for their contribution to chained Root files.

### DIFF
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -773,21 +773,20 @@ Version 1.0 (Draft)
    "targets/foo.txt".  Likewise, path pattern "foo-version-?.tgz" matches
    foo-version-2.tgz", but not "foo-version-alpha.tgz".
 
-   We are currently investigating a few "priority tag" schemes to resolve
-   conflicts between delegated roles that share responsibility for overlapping
-   target paths. One of the simplest of such schemes is for the client to
-   consider metadata in order of appearance of delegations; we treat the order
-   of delegations such that the first delegation is trusted more than the
-   second one, the second delegation is trusted more than the third one, and so
-   on. The metadata of the first delegation will override that of the second
-   delegation, the metadata of the second delegation will override that of the
-   third delegation, and so on. In order to accommodate this scheme, the
-   "roles" key in the DELEGATIONS object above points to an array, instead of a
-   hash table, of delegated roles.
+   Several schemes exist to resolve conflicts between delegated roles that
+   share responsibility for overlapping target paths. One of the simplest of
+   such schemes is for the client to consider metadata in order of appearance
+   of delegations; we treat the order of delegations such that the first
+   delegation is trusted more than the second one, the second delegation is
+   trusted more than the third one, and so on. The metadata of the first
+   delegation will override that of the second delegation, the metadata of the
+   second delegation will override that of the third delegation, and so on. In
+   order to accommodate this scheme, the "roles" key in the DELEGATIONS object
+   above points to an array, instead of a hash table, of delegated roles.
 
-   Another priority tag scheme would have the clients prefer the delegated role
-   with the latest metadata for a conflicting target path. Similar ideas were
-   explored in the Stork package manager (University of Arizona Tech Report
+   Another scheme would have the clients prefer the delegated role with the
+   latest metadata for a conflicting target path. Similar ideas were explored
+   in the Stork package manager (University of Arizona Tech Report
    08-04)[https://isis.poly.edu/~jcappos/papers/cappos_stork_dissertation_08.pdf].
 
    The metadata files for delegated target roles has the same format as the
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -784,7 +784,7 @@ Version 1.0 (Draft)
    shell-style wildcards.  For example, the path pattern "targets/*.tgz" would
    match file paths "targets/foo.tgz" and "targets/bar.tgz", but not
    "targets/foo.txt".  Likewise, path pattern "foo-version-?.tgz" matches
-   foo-version-2.tgz", but not "foo-version-alpha.tgz".
+   foo-version-2.tgz" and "foo-version-a.tgz", but not "foo-version-alpha.tgz".
 
    Several schemes exist to resolve conflicts between delegated roles that
    share responsibility for overlapping target paths. One of the simplest of
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -742,12 +742,27 @@ Version 1.0 (Draft)
            "keyids" : [ KEYID, ... ] ,
            "threshold" : THRESHOLD,
            ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
-            "paths" : [ PATHPATTERN, ... ])
+            "paths" : [ PATHPATTERN, ... ]),
+           "terminating": TERMINATING,
        }, ... ]
      }
 
    ROLENAME is the full name of the delegated role.  For example,
-   "targets/projects"
+   "targets/projects".
+
+   TERMINATING is a boolean indicating whether subsequent delegations should be
+   considered.
+
+   As explained in the Diplomat paper
+   (https://github.com/theupdateframework/tuf/blob/develop/docs/papers/protect-community-repositories-nsdi2016.pdf),
+   terminating delegations instruct the client not to consider future trust
+   statements that match the delegation's pattern. This stops the delegation
+   processing once this delegation (and its descendants) have been processed.
+   (Handling this case is conceptually similar to the use of the cut operator
+   in Prolog to stop computation, except that TUF uses this technique for
+   security instead of efficiency.) A terminating delegation for a package
+   causes any further statements about a package that are not made by the
+   delegated party or its descendants to be ignored.
 
    In order to discuss target paths, a role MUST specify only one of the
    "path_hash_prefixes" or "paths" attributes, each of which we discuss next.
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
@trishankkarthik Can you please review this pull request for correctness?  There is more work to be done on the specification, but I'd like to update it gradually.  Smaller pull requests is also a goal.  I will begin by adding missing features introduced by Diplomat, Mercury, etc., and will later edit the entire thing to assist adopters (the current text can be improved to remove ambiguities).

trishankkarthik:
@vladimir-v-diaz Sure thing, yeah, as soon as I get some downtime!

docs/tuf-spec.txt
```diff
@@ -530,6 +532,14 @@ Version 1.0 (Draft)
            , ... }
      }
 
+   COMPRESSION_ALGORITHM specifies one of the compression algorithms supported
```
trishankkarthik: Why are the compression algorithms specified here? What are the benefits?

docs/tuf-spec.txt
```diff
@@ -731,12 +742,27 @@ Version 1.0 (Draft)
...
+            "paths" : [ PATHPATTERN, ... ]),
+           "terminating": TERMINATING,
        }, ... ]
      }
-     
-   ROLENAME is the full name of the delegated role.  For example,
-   "targets/projects"
+
+   ROLENAME is the name of the delegated role.  For example,
+   "projects".
```
trishankkarthik: So does this mean that a ROLENAME *must* only be alphanumeric? How does this new ROLENAME format change how metadata files are placed on the repository? Is this format change backwards-compatible with old clients?

docs/tuf-spec.txt
```diff
@@ -731,12 +742,27 @@ Version 1.0 (Draft)
...
+
+   TERMINATING is a boolean indicating whether subsequent delegations should be
+   considered.
+
+   As explained in the Diplomat paper
+   (https://github.com/theupdateframework/tuf/blob/develop/docs/papers/protect-community-repositories-nsdi2016.pdf),
+   terminating delegations instruct the client not to consider future trust
+   statements that match the delegation's pattern. This stops the delegation
+   processing once this delegation (and its descendants) have been processed.
+   (Handling this case is conceptually similar to the use of the cut operator
```
trishankkarthik: I think we can safely remove the parenthetical note, most people won't care :)

docs/tuf-spec.txt
```diff
@@ -754,30 +780,28 @@ Version 1.0 (Draft)
...
-   Another priority tag scheme would have the clients prefer the delegated role
-   with the latest metadata for a conflicting target path. Similar ideas were
-   explored in the Stork package manager (University of Arizona Tech Report
+   the target file.  PATHPATTERN can include shell-style wildcards and supports
+   the Unix filename pattern matching convention.  Its format may either
+   indicate a path to a single file, or to multiple paths with the use of
+   shell-style wildcards.  For example, the path pattern "targets/*.tgz" would
+   match file paths "targets/foo.tgz" and "targets/bar.tgz", but not
+   "targets/foo.txt".  Likewise, path pattern "foo-version-?.tgz" matches
+   foo-version-2.tgz", but not "foo-version-alpha.tgz".
```
trishankkarthik: One more matching example here would help. Right now, it looks as if "?" matches only "numbers".

docs/tuf-spec.txt
```diff
@@ -754,30 +780,28 @@ Version 1.0 (Draft)
...
+   such schemes is for the client to consider metadata in order of appearance
+   of delegations; we treat the order of delegations such that the first
+   delegation is trusted more than the second one, the second delegation is
+   trusted more than the third one, and so on. The metadata of the first
+   delegation will override that of the second delegation, the metadata of the
+   second delegation will override that of the third delegation, and so on. In
+   order to accommodate this scheme, the "roles" key in the DELEGATIONS object
+   above points to an array, instead of a hash table, of delegated roles.
+
+   Another scheme would have the clients prefer the delegated role with the
```
trishankkarthik: I think we can remove this paragraph. It's an interesting, historical footnote, but has nothing to do with how TUF works now.

docs/tuf-spec.txt
```diff
@@ -754,30 +780,28 @@ Version 1.0 (Draft)
...
-   explored in the Stork package manager (University of Arizona Tech Report
+   the target file.  PATHPATTERN can include shell-style wildcards and supports
+   the Unix filename pattern matching convention.  Its format may either
+   indicate a path to a single file, or to multiple paths with the use of
+   shell-style wildcards.  For example, the path pattern "targets/*.tgz" would
+   match file paths "targets/foo.tgz" and "targets/bar.tgz", but not
+   "targets/foo.txt".  Likewise, path pattern "foo-version-?.tgz" matches
+   foo-version-2.tgz", but not "foo-version-alpha.tgz".
+
+   Several schemes exist to resolve conflicts between delegated roles that
```
trishankkarthik: Could we please remove the mention of "several schemes", and simply note that we use prioritized delegations ala Diplomat to resolve conflicts? Thanks!

docs/tuf-spec.txt
```diff
@@ -1004,8 +1028,8 @@ Version 1.0 (Draft)
    required, the root.json file is versioned and accessible by version number,
    e.g. 3.root.json. Clients update the set of trusted root keys by requesting
    the current root.json and all previous root.json versions, until one is
-   found that has been signed by keys the client already trusts. This is to 
-   ensure that outdated clients remain able to update, without requiring all 
+   found that has been signed by keys the client already trusts. This is to
```
trishankkarthik: Very good change! If I'm not mistaken, CoreOS proposed this change, no? If so, let's credit them here?

docs/tuf-spec.txt
vladimir-v-diaz: Listing the compression algorithms used by the repository removes the need to list all compressed versions of metadata in Snapshot, reducing its size and minimizing code complexity.  Should the benefits  of listing the compression algorithm(s) be stated in the specification, or are you simply curious for the change?

There were a few other reasons, which are captured in an email thread with David Lawrence @endophage (see email thread with title, "role names").  I'll quote some part of his request to remove compressed metadata in Snapshot, etc.


> I feel very strongly that data compression should be implemented at layer 6 and TUF should not be validating compressed files. As is, a malicious publisher could intentionally publish decompression bomb. The TUF user would download and validate the hash and size of the compressed file but be vulnerable when they decompress. By implementing compression at layer 6, the client will decompress the stream as it arrives and never read more than Length bytes from the incoming stream.
> 
> If we were talking about GBs of data there would be a strong argument for compression on disk, but TUF data is (and necessarily so) in the order of KB and the storage gains from compression are minimal. Compression at layer 6 still provides the benefit of reduced network transfer.
> 
> ...
> 
> Overall, I think consistently not using file extensions when referencing TUF metadata within the metadata files, and specifying any compression is done at layer 6 overall is a security improvement on 2 fronts. Firstly it makes the metadata more consistent, reducing code complexity, which makes auditing easier and chance of bugs lower. Secondly, it better protects the user against decompression bombs. Note that in the Python implementation the code attempts to read the entire gzipped file and as such is vulnerable to a decompression bomb: utils.py:934: fileobject = six.StringIO(gzip.open(filepath).read().decode('utf-8’))

docs/tuf-spec.txt
vladimir-v-diaz: Thanks is now given (in the PR summary) to Evan and the CoreOS team.  In a separate edit to the specification, a "contributions" section on features that have been contributed by others will be added to the specification.

I think thanks is also given to the author(s) in the pull request that implements the feature.

docs/tuf-spec.txt
vladimir-v-diaz: [Section 3.1.2.1](https://github.com/theupdateframework/tuf/blob/3532fb800533f64484092fffa97a376ecc130c73/docs/tuf-spec.txt#L412-L427) explains how delegated metadata is placed on the repository.

This change is certainly backwards-incompatible, but so are many of the changes in this version 1.0 of the specification (e.g., terminating delegations, version numbers in Snapshot, etc.)

I'm not sure that we can set restrictions on the rolename, which must ultimately match the filename of the role as saved to the underlying file system.  Should we state in the specification that we do not place restrictions on rolenames, and supported characters in rolenames are left to adopters?

vladimir-v-diaz:
@trishankkarthik I have gone through all of your comments.  Please give the pull request, or my tacked-on commits, another pass when you have the time.

JustinCappos:
I'm slightly confused about why this change is showing up here.  Wouldn't
this require a TAP?  Was this added prior to 1.0?

docs/tuf-spec.txt
trishankkarthik: Like @JustinCappos, I'm also not sure that this change is straightforward.

This might have implications for snapshot metadata. There are use cases where, besides version numbers, the hashes of targets and delegated targets metadata files might be listed in the snapshot metadata. Do we include the hashes of both the compressed and uncompressed metadata files in there?

vladimir-v-diaz:
Trishank and I went over my comments.  We need to still need to verify whether
(1) the compression algorithms can be excluded from the Snapshot file
(2) restrictions should be set for rolenames.  For example, exclude unicode

vladimir-v-diaz:
> I'm slightly confused about why this change is showing up here. Wouldn't
> this require a TAP? Was this added prior to 1.0?

The change to exclude compressed version of metadata from Snapshot was discussed (via email with the Docker folks) and implemented before the TAP process was adopted.

vladimir-v-diaz:
> Trishank and I went over my comments. We need to verify whether
> (1) the compression algorithms can be excluded from the Snapshot file
> (2) restrictions should be set for rolenames. For example, exclude unicode

I have excluded the `compression algorithm` field from Root.json, and will leave open an issue for (2).  Merging!

Issue #436, opened 2017-04-18
Update example metadata


trishankkarthik:
We should update the example metadata to, for example:

- [ ] reflect changes to the snapshot metadata (Mercury)
- [ ] include the map file (Trident)
- [ ] include multi-role delegations in targets metadata (Trident)
- [ ] include examples of prioritized and / or terminating delegations (Diplomat)

joshuagl:
Looks like this was fixed in PR #496? If so this issue should be closed.

SantiagoTorres:
It appears to me that you are correct, @joshuagl. Thanks for pointing it out!

PR #430, opened 2017-02-15
Implement TAP 4 (the map file)

vladimir-v-diaz:
This is an implementation of [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md) (the map file).  A MultiRepoUpdater class is added that reads `map.json`, determines which repository URL to use for a requested target file, and returns the updater object and targetinfo for the requested target file.  The client is free to use the updater object to download the target file corresponding to the targetinfo.  An example is provided below:

```Python
multi_repo_updater = tuf.client.updater.MultiRepoUpdater("path/to/map.json")
targetinfo, my_updater = multi_repo_updater.get_one_valid_targetinfo("foo.tgz")

# save the `foo.tgz` target file to "path/to/destination".
my_updater.download_target(targetinfo, "path/to/destination")
```

The client can also access the updater objects for each repository...
```Python
pypi_updater = multi_repo_updater.repository_names_to_updaters['pypi']
pypi_updater.download_target(...)
```

@awwad PIease have a look at this PR and let me know if it supports everything needed to create Uptane repositories.  The repository tools behave as they currently do, but they now support the creation of multiple repositories within the same Python process.

PR #437, opened 2017-04-18
Update example metadata.  Fix issue #436.

vladimir-v-diaz:
The metadata that is used by the unit and integration tests (available [here](https://github.com/theupdateframework/tuf/tree/develop/tests/repository_data)) are the most up-to-date.  This pull request moves the relevant metadata, key files, target files, etc. to the [examples](https://github.com/theupdateframework/tuf/tree/develop/examples) directory to address issue #436.

Please note that the Trident modifications have not yet been merged into the "develop" branch.  An implementation of the map file is awaiting a review in PR #430, and support for multi-role delegations is available in Uptane, however, it is based on an old TAP/design that will need changing before being merged into TUF.

This pull request includes metadata that supports the specification and design goals of Mercury and Diplomat.


vladimir-v-diaz:
The example metadata listed in the specification should be up-to-date as well (excluding Trident modifications).
https://github.com/theupdateframework/tuf/blob/40aaf93f5a39f659cc312b66624f2cbff59b8582/docs/tuf-spec.txt#L658-L689

vladimir-v-diaz:
ping @awwad

awwad:
re the linked metadata example in the spec: there's a capitalized role name and it includes root's hash in snapshot. The former should be changed either here or in #472 (after #472 is rebased). When are we changing the latter?

vladimir-v-diaz:
I might make a few more changes to the format of metadata.  Perhaps I
should rebase this once things have settled down.  Another option is to
remove this example metadata and just have one place for it in METADATA.md.

I imagine that it will confuse people to have multiple examples.

examples/repository/metadata.staged/snapshot.json
```diff
@@ -1,40 +1,32 @@
...
+   "method": "ed25519",
+   "sig": "01bc8667e2afbceef3df6ec6782088c77e59bdb5a29b75634200528bdedc8af9694e38a672fa2cd00051fb25ac12d3079bbc48f424783d23fb532b7508adf40f"
   }
- ], 
+ ],
  "signed": {
-  "_type": "Snapshot", 
-  "expires": "2030-01-01T00:00:00Z", 
+  "_type": "Snapshot",
+  "expires": "2030-01-01T00:00:00Z",
```
awwad: Trailing whitespace here makes two lines that did not change look like they have changed. There's more like this above and below, too. Was this hand-edited or written out by repository_tool?

awwad:
Will number these.
Call the two old questions/points 1 & 2:
1. As mentioned, the metadata examples use capitalized role names, both in the spec and the examples directory. The former should be changed either here or in #472 (after #472 is rebased).

2. Root's hash is included in snapshot. Where are we changing that?

Here are some more:
3. In your PR description, you say that you're _moving_ the examples from the test data folder to the examples folder, but it looks like you're adding new stuff to examples/ and leaving the test data in test_data. Is that intended?

4. You made some of the ed25519 keys RSA keys. No particular reason, I assume? (just what you happened to generate? or is there a problem?)

5. keyid_hash_algorithms was added to the public keys. I remember talking about expecting this to show up in public keys, but was this change intended here? Did we change code that generated keys, or is this an artifact of the switch to RSA keys? (In which case we should make this consistent across key types.)

6. There's some trailing whitespace in some of the metadata files that makes me wonder if it's hand edited and makes several lines look like they've changed when they haven't. Are the signatures correct? Did whitespace somehow get added by repository_tool?

examples/repository/metadata.staged/snapshot.json
vladimir-v-diaz: I have it set up where my editor removes trailing whitespace.  This might be the cause of this issue.

awwad: Oh, geez. I looked at it backwards and processed it as whitespace added, not removed. Well, the other question stands, I suppose: are the signatures valid?

vladimir-v-diaz:
@awwad 
Do you think we should have only one set of example metadata (available [here](https://github.com/theupdateframework/tuf/blob/develop/METADATA.md))?  The example metadata in METADATA.md is actually the metadata used in the unit and integration tests, so they are the most up to date.  It also avoids the problem of having to edit two sets of example metadata when a change is made to the specification.

> There's some trailing whitespace in some of the metadata files that makes me wonder if it's hand edited and makes several lines look like they've changed when they haven't. Are the signatures correct? Did whitespace somehow get added by repository_tool?

My editor likely stripped the trailing whitespace.  The signatures might not be affected because metadata is always canonicalized.  However, the hashes might...

awwad:
There's sample or test metadata in several places:
1. tuf spec (and perhaps other documentation)
2. examples/ root directory
3. tests/test_data directory

You could combine 2 and 3. Uptane has sample metadata in a root-level samples/ directory and is shifting to using that data for testing purposes (instead of data kept in tests/test_data), but this may prove to be too much clutter in the samples/ directory... so I don't really have a good answer for you. Yes, I think that having example data in too many places is bad. I would lean toward combining 2 and 3, but I can't say that with a lot of conviction.

awwad:
Are you waiting for any further comments here?

vladimir-v-diaz:
No, I'm not waiting for further comments.  I'll likely work on this soon to combine 1 and 2.

vladimir-v-diaz:
Items (2) and (3) [here](https://github.com/theupdateframework/tuf/pull/437#issuecomment-317773194) were combined in https://github.com/theupdateframework/tuf/pull/496.  Closing this pull request since it's been fixed in #496.

Issue #438, opened 2017-04-24
Hashing algorithm to generate hashed bin prefix should be configurable


vladimir-v-diaz:
An attempt was made in the past to make hashing algorithms configurable. Although at the time it didn't seem doable, due to prepended sha512 digests and filename length restrictions on some systems, a later change to the specification allowed for configurable hashing algorithms in all of TUF. The change was that version numbers are now prepended to all metadata filenames, rather than digests.

After version numbers were added to filenames, Root and Snapshot were the only metadata required to have digests prepended, but they too have now been modified to only prepend version numbers. In addition, multiple hashing algorithms are supported for KEYIDs

Presently, I think the hashing algorithm for all of TUF can be made configurable.  The only place where it is still fixed appears to be with hashed bin prefixes.

See https://github.com/theupdateframework/tuf/blob/40aaf93f5a39f659cc312b66624f2cbff59b8582/tuf/repository_tool.py#L2803-L2810

vladimir-v-diaz:
This issue was fixed in https://github.com/theupdateframework/tuf/pull/439.

PR #439, opened 2017-04-25
Add configurable setting for hashing algorithm used in hashed bin delegations

vladimir-v-diaz:
Fix for issue #438.

### DIFF
--- a/tuf/repository_tool.py
+++ b/tuf/repository_tool.py
@@ -69,9 +69,8 @@ logger = logging.getLogger('tuf.repository_tool')
 tuf.log.add_console_handler()
 tuf.log.set_console_log_level(logging.INFO)
 
-# The algorithm used by the repository to generate the digests of the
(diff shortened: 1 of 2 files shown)


vladimir-v-diaz:
ping @awwad

awwad:
So, it looks like this is specific to hashed bin prefixes, but the comments make it sound like it's a general setting that affects hashing done across TUF. Am I misunderstanding something?

vladimir-v-diaz:
You are correct!  I updated the comment/description to indicate that this setting is now only used to calculate path hash prefixes.

awwad:
Do you think the name of the setting itself should also be more specific (HASH_ALGO_FOR_PATH_HASH_PREFIX or something), or is that getting too paranoid and wordy?

awwad:
No, I think you're fine. This makes sense to me.

PR #440, opened 2017-04-26
Expand on the TUF client update workflow, per popular demand.

trishankkarthik:
Besides our reference implementation, I do not believe that the TUF specification describes the precise steps that an alternative, compliant implementation should follow to download and verify a target. This PR is a humble attempt at solving this problem.

There are still some details are missing, such as:
- [X] Downloading and verifying the complete chain of root metadata files. (Maybe @ecordell or @heartsucker could help here?)
- [x] Merge reading consistent snapshots with the client update workflow.
- [ ] Using [the map file](https://github.com/theupdateframework/taps/blob/JustinCappos-patch-2/tap4.md), but the [pseudocode](https://gist.github.com/trishankkarthik/e5d8134a4052f712b7416c76787b077b#file-tap4-py-L7-L46) is available...

Hopefully, I didn't miss something important, or, worse, get something wrong. (I'll blame it on travel lag.) Please take it from here!

### DIFF
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -932,6 +932,16 @@ Version 1.0 (Draft)
 
 5.1. The client application
 
+    0. **Load the previous root metadata file.** We assume that a good, trusted
+    copy of this file was shipped with the package manager / software updater
+    using an out-of-band process.
+
+    0.1. **Check signatures.** The previous root metadata file MUST have been
+    signed by a threshold of keys specified in the previous root metadata file.
+
+    0.2. **Check for a freeze attack.** The latest known time should be lower
+    than the expiration timestamp in this metadata file.
+
     1. **Download the root metadata file**, up to some X number of bytes
     (because the size is unknown). The value for X is set by the authors of the
     application using TUF. For example, X may be tens of kilobytes.
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -1057,9 +1057,12 @@ Version 1.0 (Draft)
     5.1. If there is no targets metadata about this target, then report that
     there is no such target.
 
-    5.2. Otherwise, download the target (exactly the number of bytes specified
-    in the targets metadata), and verify that its hashes match the targets
-    metadata.
+    5.2. Otherwise, download the target (up to the number of bytes specified in
+    the targets metadata), and verify that its hashes match the targets
+    metadata. (We download up to this number of bytes, because in some cases,
+    the exact number is unknown. This may happen, for example, if an external
+    program is used to compute the root hash of a tree of targets files, and
+    this program does not provide the total size of all of these files.)
 
    Note: If at any point in the above procedure there is a problem (i.e., if
    unexpired, signed, valid metadata cannot be retrieved from the repository),
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -1,6 +1,6 @@
                       The Update Framework Specification
 
-7 October 2016
+18 May 2017
 Version 1.0 (Draft)
 
 1. Introduction
@@ -8,6 +8,10 @@ Version 1.0 (Draft)
 1.1. Scope
 
    This document describes a framework for securing software update systems.
+
+   The keywords "MUST," "MUST NOT," "REQUIRED," "SHALL," "SHALL NOT," "SHOULD,"
+   "SHOULD NOT," "RECOMMENDED," "MAY," and "OPTIONAL" in this document are to be
+   interpreted as described in RFC 2119.
 
 1.2. Motivation
 
@@ -74,7 +78,7 @@ Version 1.0 (Draft)
 
    The framework will not have the responsibility of deciding on the correct
    course of action in all error situations, such as those that can occur when
-   certain attacks are being performed.  Instead, the framework will provide 
+   certain attacks are being performed.  Instead, the framework will provide
    the software update system the relevant information about any errors that
    require security decisions which are situation-specific.  How those errors
    are handled is up to the software update system.
@@ -134,31 +138,31 @@ Version 1.0 (Draft)
    the client system. That is, an attacker can provide arbitrary files in
    response to download requests and the files will not be detected as
    illegitimate.
-   
+
    Endless data attacks.  Attackers should not be able to respond to client
    requests with huge amounts of data (extremely large files) that interfere
    with the client's system.
-   
+
    Extraneous dependencies attacks.  Attackers should not be able to cause
    clients to download or install software dependencies that are not the
    intended dependencies.
-   
+
    Fast-forward attacks.  An attacker arbitrarily increases the version numbers
-   of project metadata files in the snapshot metadata well beyond the current 
+   of project metadata files in the snapshot metadata well beyond the current
    value, thus tricking a software update system into thinking any subsequent
    updates are trying to rollback the package to a previous, out-of-date version.
    In some situations, such as those where there is a maximum possible version
    number, the perpetrator could use a number so high that the system would
    never be able to match it with the one in the snapshot metadata, and thus
    new updates could never be downloaded.
-   
+
    Indefinite freeze attacks.  Attackers should not be able to respond to
    client requests with the same, outdated metadata without the client being
    aware of the problem.
-   
+
    Malicious mirrors preventing updates.  Repository mirrors should be unable
    to prevent updates from good mirrors.
-   
+
    Mix-and-match attacks.  Attackers should not be able to trick clients into
    using a combination of metadata that never existed together on the
    repository at the same time.
@@ -170,7 +174,7 @@ Version 1.0 (Draft)
    Slow retrieval attacks.  Attackers should not be able to prevent clients
    from being aware of interference with receiving updates by responding to
    client requests so slowly that automated updates never complete.
-   
+
    Vulnerability to key compromises. An attacker who is able to compromise a
    single key or less than a given threshold of keys can compromise clients.
    This includes relying on a single online key (such as only being protected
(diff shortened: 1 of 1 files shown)


docs/tuf-spec.txt
```diff
@@ -928,24 +928,113 @@ Version 1.0 (Draft)
 
 5.1. The client application
 
-   1. The client application first instructs TUF to check for updates.
+    1. **Download the root metadata file**, up to some predetermined number of
```
JustinCappos: Please clarify what "predetermined" means here.

docs/tuf-spec.txt
```diff
@@ -928,24 +928,113 @@ Version 1.0 (Draft)
...
-   5. TUF provides the software update system with a list of available files
-      according to targets.json.
+    1.4. **If the the timestamp and / or snapshot keys have been rotated, then
+    delete the previous timestamp and snapshot metadata files.** This is done
+    in order to recover from fast-forward attacks after the repository has been
+    compromised and recovered. A _fast-forward attack_ happens when attackers
+    arbitrarily increase the version numbers of: (1) the timestamp metadata,
+    (2) the snapshot metadata, and / or (3) the targets, or a delegated
+    targets, metadata file in the snapshot metadata. Please see the Mercury
+    paper for more details (to be published soon).
```
JustinCappos: You can probably point to where this will be:

https://ssl.engineering.nyu.edu/papers/kuppusamy_usenix_17.pdf

For the time being, feel free to put the SUBMITTED there, but please label it DRAFT

docs/tuf-spec.txt
```diff
@@ -928,24 +928,113 @@ Version 1.0 (Draft)
...
 
-   6. The software update system instructs TUF to download a specific target file.
+    2. **Download the timestamp metadata file**, up to some predetermined
+    number of bytes (because the size is unknown.)
 
-   7. TUF downloads and verifies the file and then makes the file available to the
-      software update system.
+    2.1. **Check signatures.** It must have been signed by a threshold of keys
```
JustinCappos: Please avoid 'It', especially when it crosses points, etc.

docs/tuf-spec.txt
```diff
@@ -928,24 +928,113 @@ Version 1.0 (Draft)
...
+    4.5.2.3. Otherwise, if it is a non-terminating delegation, continue
+    processing the next delegation, if any. Stop the search, and jump to step 5
+    as soon as a delegation returns a result.
+
+    5. **Verify the desired target against its targets metadata.**
+
+    5.1. If there is no targets metadata about this target, then report that
+    there is no such target.
+
+    5.2. Otherwise, download the target (up to the number of bytes specified in
```
JustinCappos: It's worth explaining why this is 'up to' instead of 'exactly'

vladimir-v-diaz:
Should we postpone the specification edits for the map file changes?  The PR that implements the map file (TAP 4) hasn't yet been reviewed/merged with the master/develop branch.  We can also update the specification in pull request #430.

JustinCappos:
You could, but realistically, TAP 4 is moving quickly toward acceptance.
We will need to do TAP 7 first, but I hope we will have all approved in the
next two months.

docs/tuf-spec.txt
```diff
@@ -928,24 +928,120 @@ Version 1.0 (Draft)
 
 5.1. The client application
 
-   1. The client application first instructs TUF to check for updates.
+    1. **Download the root metadata file**, up to some X number of bytes
+    (because the size is unknown). The value for X is set by the authors of the
+    application using TUF. For example, X may be tens of kilobytes.
 
-   2. TUF downloads and verifies timestamp.json.
+    1.1. **Check signatures.** The root metadata file must have been signed by
```
heartsucker: I think you should mention key pinning over TOFU for the initial root.json.

trishankkarthik: Good idea. Where is it best in the specification to discuss this?

trishankkarthik: Also, where in the specification do you think we should use "SHOULD" or "MUST" (ala RFC 2119)?

heartsucker: For example:

> The latest known time should be lower than the expiration timestamp

should say

> The latest known time MUST be lower than the expiration timestamp

heartsucker: Also

> The version number of the previous root metadata file must be less than or equal

Should have the "must" in all caps

heartsucker: As for where to mention pinning vs. TOFU, I think there should be a section before 5.1 maybe called "client design" that mentions this, and then step 1 could specifically say something like "If a client was initialized with a pinned set of root keys, these keys and only these keys MUST be used to validate the original root.json."

heartsucker:
I have another question on this. I seem to have a chicken and egg problem.

Imagine a repo at time T1 with all metadata at versions 1. A client connects and downloads all metadata. At time T2, someone swaps the root metadata keys and all role keys creating all new metadata at versions 2. A client connects at T3 and starts by downloading the timestamp.json. This would be untrusted because the client doesn't know about the version 2 keys.

I think this could include the step saying "if timestamp.json is untrusted, the client downloads root.json an rebuilds the chain of trust, returns to the start and attempts to download timestamp.json again repeating the loop."

ecordell:
@heartsucker The first step of the update process should always be updating the root chain, so that trusted keys are up to date. I think the only change needed here is to clarify that "downloading the root" can mean "downloading n roots, until one is reached that is signed by trusted keys"

trishankkarthik:
Very good question, @heartsucker, about not being able to verify timestamp and / or snapshot metadata due to change of keys, and thanks for the response, @ecordell !

Indeed, due to our findings from [the submitted Mercury draft](https://ssl.engineering.nyu.edu/papers/kuppusamy_usenix_17.pdf), we have made updating the root metadata the first and foremost step, so that there are no problems verifying timestamp and / or snapshot metadata from a change of keys (especially after the repository has been compromised, and recovered).

docs/tuf-spec.txt
trishankkarthik: Thanks very much, @heartsucker, will incorporate these suggestions!

vladimir-v-diaz:
@trishankkarthik I think this PR is good enough to merge at the moment.  We can address the other points raised in the comments in separate pull requests.  Please go ahead and make
any last remaining edits before we merge.

Documenting the other needed changes to specification...

* The map file (TAP 4) hasn't been finalized yet.
* Downloading and verifying chain of root metadata
* More detailed outline of root migration (see https://github.com/theupdateframework/tuf/pull/444 and https://github.com/theupdateframework/tuf/pull/445)

trishankkarthik:
@vladimir-v-diaz @JustinCappos I agree we should merge. But first, I think we should:

1. Merge Section 7.2 (reading consistent snapshots) with Section 5.1 (client update workflow). It's confusing to have multiple sections about the same thing.

2. Remove some redundant notes at the end of Section 5.1.

trishankkarthik:
Okay, I merged reading consistent snapshots with the client update workflow. I also added some other details. Someone else please review to make sure I didn't make mistakes from burning the midnight oil.

@vladimir-v-diaz We should fix Section 7.1 (writing consistent snapshots). It seems to have some now inaccurate information. For example, aren't all targets metadata files written as FILENAME.VERSION.EXT instead of FILENAME.DIGEST.EXT? Isn't only the snapshot metadata file written as FILENAME.DIGEST.EXT? The root and timestamp metadata files are written as both FILENAME.EXT and FILENAME.DIGEST.EXT, right?

Also, have we decided whether it's (DIGEST|VERSION).FILENAME.EXT or FILENAME.(DIGEST|VERSION).EXT? Thanks!

trishankkarthik:
Okay, I just added that the expiration of the previous root metadata file does not matter, because we will attempt to update it in the next step.

docs/tuf-spec.txt
trishankkarthik: Done!

vladimir-v-diaz:
@trishankkarthik 

The [code](https://github.com/theupdateframework/tuf/blob/70cf57ad85330d6af8a99bda1f3db2660084ef90/tuf/repository_lib.py#L555-L578) expects metadata filenames to be in `<version_number>.rolename.<ext>`

I think that's the format we last settled on.  There were several reasons, one of them being that longn digests might exceed the filename limit set by certain filesystems.  However, according to Hannes, TAP 8 requires that filenames include the digest.  We'll see once the TAP is edited with more detail.

trishankkarthik:
@vladimir-v-diaz Got it. Please feel free to make the change!

docs/tuf-spec.txt
```diff
@@ -1108,10 +1108,19 @@ Version 1.0 (Draft)
...
-   found that has been signed by keys the client already trusts. This is to
-   ensure that outdated clients remain able to update, without requiring all
-   previous root keys to be kept to sign new root.json metadata.
-
+   found that has been signed by a threshold of keys that the client already
+   trusts. This is to ensure that outdated clients remain able to update,
+   without requiring all previous root keys to be kept to sign new root.json
+   metadata.
+   
+   In the event that the keys being updated are root keys, it is important to
```
trishankkarthik: @vladimir-v-diaz I think this is great! Should we say a bit more about this in Step 1 of Section 5.1?

vladimir-v-diaz: I think we should.  Do we also explain how to recover from a fast-forward attack?

Note: I just reviewed section 7.1 and it seems accurate to me, at least in regard to filenames.  Is there a sentence/statement in particular that is inaccurate?

vladimir-v-diaz: Okay, we do mention fast-forward attacks in the revised 1.5 section!

vladimir-v-diaz:
I think we might be able to reuse some of @heartsucker's outline of root key migrations [here](https://github.com/theupdateframework/tuf/pull/445/files).

trishankkarthik:
Okay, so I just worked with @vladimir-v-diaz to update the workflow. Specifically, I added details on how to update to the latest root metadata file (thanks to @ecordell and @heartsucker).

@JustinCappos Please review the entire workflow to make sure there are no problems?

PR #443, opened 2017-05-16
Fixed markdown errors

dingwilson:
GitHub is finicky with their markdown. This PR fixes the markdown issues.


vladimir-v-diaz:
Thanks for the fix!

Interesting.  Did GitHub change the markdown parser?  I recall these headings displaying correctly in the past.

awwad:
Perhaps in March: https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown

PR #444, opened 2017-05-18
Clarify procedure for updating to new root.json

vladimir-v-diaz:
Clients should validate new root.json according to the threshold and keys set by its previous version.

See @heartsucker's comment [here](https://github.com/heartsucker/rust-tuf/issues/42#issuecomment-302436972).

### DIFF
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -1014,9 +1014,9 @@ Version 1.0 (Draft)
    root keys in the previous version of root.json, up to the threshold listed
    for root in the previous version of root.json. If this is not the case,
    clients will (correctly) not validate the new root.json file.  For example,
-   if there is 1.root.json that has threshold 2 and 2.root.json that has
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
@heartsucker Let us know if this PR captures the case you raised in issue # rust-tuf/32

vladimir-v-diaz:
@awwad I added your and heartsucker's suggestions.  What do you think of the changes?

awwad:
I think it's good, with the typo at the end fixed. You don't think my wording was wordy / unclear, then, I gather.

awwad:
Oh - btw, contingent on the instructions for the client updater procedure being added as well, which we talked about. (Something like what @heartsucker has in #444). I guess @trishankkarthik is doing that?

vladimir-v-diaz:
Yup, a detailed client update workflow is being added to the specification in pull request #440.

PR #445, opened 2017-05-18
added root key migration details

heartsucker:
As per the discussion here: https://github.com/heartsucker/rust-tuf/issues/42

### DIFF
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -1008,6 +1008,23 @@ Version 1.0 (Draft)
    ensure that outdated clients remain able to update, without requiring all 
    previous root keys to be kept to sign new root.json metadata.
 
+   The exact steps to perform the root.json migration should be as follows:
+     
+   1. The client starts at version N of root.json
+
+   2. The client downloads version N + 1 of root.json
+
+   3. The client uses the keys and threshold defined in version N to
+      validate the signatures of N + 1
+
+   4. The client flushes all knowledge of root.json verion N then uses the
+      updated keys and threshold to validate the signatures of N + 1
+
+   5. Repeat until the client has reached the latest version of root.json
+
+   That is, if there is 1.root.json that has threshold 2 and 2.root.json that has
+   threshold 3, 2.root.json MUST be signed by at least 2 keys defined in
+   1.root.json and at least 3 keys defined in 3.root.json.
 
    To replace a delegated developer key, the role that delegated to that key
    just replaces that key with another in the signed metadata where the


heartsucker:
Duplicates #444

docs/tuf-spec.txt
```diff
@@ -1008,6 +1008,23 @@ Version 1.0 (Draft)
...
+      validate the signatures of N + 1
+
+   4. The client flushes all knowledge of root.json verion N then uses the
+      updated keys and threshold to validate the signatures of N + 1
+
+   5. Repeat until the client has reached the latest version of root.json
+
+   That is, if there is 1.root.json that has threshold 2 and 2.root.json that has
+   threshold 3, 2.root.json MUST be signed by at least 2 keys defined in
+   1.root.json and at least 3 keys defined in 3.root.json.
```
awwad: I think this example is a good one to include.

docs/tuf-spec.txt
```diff
@@ -1008,6 +1008,23 @@ Version 1.0 (Draft)
...
+     
+   1. The client starts at version N of root.json
+
+   2. The client downloads version N + 1 of root.json
+
+   3. The client uses the keys and threshold defined in version N to
+      validate the signatures of N + 1
+
+   4. The client flushes all knowledge of root.json verion N then uses the
+      updated keys and threshold to validate the signatures of N + 1
```
awwad: (Should this be N + 2?)

Either way, I this this is probably slightly wrong, and may leave you in a state where you have no trusted root metadata.

I would say that 4 should be removed and replaced by:
4. The client uses the keys and threshold defined in version N + 1 to validate the signatures on N + 1.

4.5. If both validations have succeeded, _then_ flush all knowledge of root.json version N and heretofore use N + 1's keys and thresholds to validate further versions.

awwad:
Oh - btw, contingent on the instructions for the client updater procedure being added as well, which we talked about. (Something like what @heartsucker has in #444). I guess @trishankkarthik is doing that?

docs/tuf-spec.txt
vladimir-v-diaz: I think heartsucker's use of N + 1 in (4) is correct, it matches your (4).  Nevertheless, I do think version N of root.json should only be flushed once N + 1 has been validated against itself and the previous version's keys and threshold.

heartsucker: @vladimir-v-diaz You're right. I should say the flush happens *after* the validation.

vladimir-v-diaz:
@heartsucker 
Thanks for the contributions!  Parts of this pull request (#445) were incorporated into pull requests #440 and #444 that have now been merged.

PR #448, opened 2017-05-24
Fix section 3, where the hash is specified rather than the

vladimir-v-diaz:
version number of the snapshot file.


trishankkarthik:
LGTM!

PR #455, opened 2017-06-02
Turn TUF spec into a well formatted RFC

heartsucker:
This PR is so, so, so not ready for a merge, but I wanted to show you what I've been working on today to get input.

My motivation for doing this is that a huge amount of the spec thus far isn't actually "spec" but discussion. Such discussion should be moved to a separate document. Second, there was a lot of inconsistency in how things were phrased, and a seemingly large amount of mushiness on what needs to be done (i.e., is it a MUST, MAY, or SHOULD?). Third, this is probably going to be turned into an RFC eventually, so here's a head start on that. Fourth, oh god it's so much more beautiful this way.

Also, sorry to all y'all authors of this that I didn't cite correctly. I figure that's something you'd want to put in yourselves.

screenshot_2017-06-02_16-32-30
screenshot_2017-06-02_16-32-56
screenshot_2017-06-02_16-33-15

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,0 +1,4 @@
+.refcache/
+tuf-rfc.xml
+tuf-rfc.html
+tuf-rfc.txt
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,5 +1,5 @@
-.PHONY: clean help html kramdown xml2rfc
-.DEFAULT_GOAL := xml2rfc
+.PHONY: clean help html kramdown open xml2rfc
+.DEFAULT_GOAL := help
 
 OPEN=$(word 1, $(wildcard /usr/bin/xdg-open /usr/bin/open /bin/echo))
 MKD := tuf-rfc.md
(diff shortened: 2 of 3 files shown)


heartsucker:
Also, furthering my motivation on this. There were both contradictions as well as significant amounts of undefined behavior. I have mostly reconciled them, but there are a number of TODOs where the behavior was not clear enough for me to feel confident in asserting what was meant by the previous spec.

vladimir-v-diaz:
> Also, furthering my motivation on this. There were both contradictions as well as significant amounts of undefined behavior. I have mostly reconciled them, but there are a number of TODOs where the behavior was not clear enough for me to feel confident in asserting what was meant by the previous spec.

Should there be a separate pull request for these issues?  One pull request that resolves the undefined behavior and contradictions, and another one just for the RFC-formatted text?

heartsucker:
I think that would lead to the case of trying to maintain (even temporarily) two sets of documentation, and if `tuf-spec.txt` wasn't removed, we've have two semi-authoritative documents, and I feel like it's better to cut and switch hard all at once.

It's also been hard for us here to read the current spec as it is because there is (in my opinion) too much discussion in the spec for engineers who just want to focus on the "how" of implementing TUF. I find that things like "Detect a Freeze Attack" are somewhat noisy phrases that add cognitive load and should wholly be replaced by simple directives like "ensure `expires` is after the current time". I think explanatory phrases like this belong in an ancillary design document, not a specification.

Think it would be best to take this PR and say:

1. Does it match the old spec (or improve where it must)?
2. Is it less ambiguous than the old spec?

And if yes to both, we make the switch.

I think keeping a single source for the specification will make it easier on implementors. Unless there's a compelling case to maintain the `tuf-spec.txt` that I'm not seeing, I think this document will be more useful than the document it replaces.

heartsucker:
Noting an additional case of lack of clarity on the TUF spec as evidence that this change might be a good thing. https://github.com/theupdateframework/taps/issues/36#issuecomment-308178031

vladimir-v-diaz:
As discussed via email, we decided to use Markdown to format the specification.  See pull request #477.

PR #461, opened 2017-06-19
What should happen when Root keys are compromised?

vladimir-v-diaz:
The question of what should happen when a threshold <=  the number of Root keys is compromised was asked here: https://github.com/heartsucker/rust-tuf/issues/64#issuecomment-308841123.  This pull request attempts to address this question.

### DIFF
--- a/docs/tuf-spec.txt
+++ b/docs/tuf-spec.txt
@@ -271,10 +271,15 @@ Version 1.0 (Draft)
    kept offline.  If less than a threshold of Root keys are compromised, the
    repository should revoke trust on the compromised keys.  This can be
    accomplished with a normal rotation of root keys, covered in section 6.1
-   (key management and migration). If a threshold of root keys is compromised,
+   (Key management and migration). If a threshold of root keys is compromised,
    the Root keys should be updated out-of-band, however, the threshold should
-   be chosen so that this is extremely unlikely.
-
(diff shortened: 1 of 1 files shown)


docs/tuf-spec.txt
```diff
@@ -268,7 +268,12 @@ Version 1.0 (Draft)
    configured repository.
 
    The root role's private keys must be kept very secure and thus should be
-   kept offline.
+   kept offline.  If less than a threshold of Root keys are compromised, the
+   repository should revoke trust on the compromised keys.  This can be
+   accomplished with a normal rotation of root keys, covered in section 6.1
+   (key management and migration). If a threshold of root keys is compromised,
+   the Root keys should be updated out-of-band, however, the threshold should
```
trishankkarthik: Great clarification! :+1: 

I think we should add the point that, in the unlikely event where a threshold of root role keys have been compromised, then you have much bigger problems to worry about than key revocation. Here, it's safest to assume that attackers have installed malware and taken over affected machines. Now, key revocation and replacement should be assumed to be impossible, at least not without destroying the remote device and rebuilding everything from scratch.

What do you guys think?

vladimir-v-diaz:
I added more text to explain why it is nearly impossible to recover from a complete compromise of the Root keys.  Please provide feedback or approve the pull request so we can get it merged.

Issue #465, opened 2017-07-06
Update spec to not include hashes/size in snapshot metadata


heartsucker:
As per the discussion here: https://github.com/heartsucker/rust-tuf/issues/104#issuecomment-313496672

vladimir-v-diaz:
We need to also determine if it is necessary to list the hash and length of Snapshot in Timestamp.
See https://github.com/theupdateframework/tuf/blob/904fa9b8df8ab8c632a210a2b05fd741e366788a/docs/tuf-spec.txt#L909-L913

JustinCappos:
The hash and length make sense to me.  Why wouldn't we have the hash and
length here?

The version of the snapshot file makes less sense, but I don't see a major
problem with it being there.

vladimir-v-diaz:
Fixed in https://github.com/theupdateframework/tuf/pull/466 and implemented in https://github.com/theupdateframework/tuf/pull/469.

PR #466, opened 2017-07-06
Fix for issue #465

vladimir-v-diaz:
Addresses issue #465, where the hash and length of Root are no longer listed in Snapshot.

@JustinCappos Please verify that it is safe to no longer list the hash of Root in Snapshot.


JustinCappos:
I've thought this over.  This change makes sense to me.

PR #479, opened 2017-08-03
Remove tuf-spec.txt

vladimir-v-diaz:
`tuf-spec.md` is the only specification file that we should need to edit, not two.


heartsucker:
I'm a bit late to the party here, but it might be worth putting a symlink from `tuf-spec.txt` to `tuf-spec.md` so all the hyperlinks out in the world don't break.

vladimir-v-diaz:
@heartsucker 
If someone were to visit tuf-spec.txt on GitHub, they would see the following:
https://github.com/vladimir-v-diaz/tuf/blob/develop/docs/tuf-spec.txt

Should I instead store a link to the full path on GitHub instead?  How do people normally handle this case?

heartsucker:
Hmm. I'm not sure what people usually do. It says "Symbolic link" and makes it somewhat obvious it moved. I guess people usually host docs on something like readthedocs.io or docs.rs, which handle this a little better.

vladimir-v-diaz:
I restored tuf-spec.txt for the time being.  Since the specification shouldn't change too often, I'll try to maintain both and see if it becomes a problem.

SantiagoTorres:
Perhaps you could leave tuf-spec.txt with a disclaimer and a mention it has been moved? probably with time you can just delete it.

vladimir-v-diaz:
We link to it from papers, so I'm not sure if it should ever be deleted.  A disclaimer was my first thought, but then I thought that some people might actually prefer to view the text version.

I/we will try to just copy/paste over changes for now, and see how it goes.

JustinCappos:
I'd prefer either Eric's or Santiago's suggestion.  I think those would
work better.

Justin

vladimir-v-diaz:
Okay, here's what visitors will see when they visit tuf-spec.txt:
https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt

JustinCappos:
Super, thanks!

Issue #355, opened 2016-08-08
Add ECDSA to spec


titanous:
While I'd love to only use Ed25519 for everything, there is a lack of support in cryptographic smartcards/HSMs for Ed25519. Many devices do however support ECDSA, including the Yubikey 4.

I think the specification should include support for ECDSA keys and signatures. The most common will be the NIST P-256 curve signing SHA-256 hashes.

I propose the addition of a new key type and signature method: `ecdsa-sha2-nistp256` (uses SHA-256 for hashing). This is [used in OpenSSH](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL) and specified in [RFC 5656](https://tools.ietf.org/html/rfc5656). I don't think there is any need for signature algorithm agility here, so the key type and signature method would be the same for simplicity.

titanous:
We implemented this in flynn/go-tuf#98. Any thoughts before it gets rolled out into the wild?

vladimir-v-diaz:
We are generally open to supporting new key types. This looks good to me.  I'll go ahead and revise the specification.  The Python implementation will also be updated to support this key type,

vladimir-v-diaz:
PR to support ECDSA keys and signatures (in the reference implementation) submitted here: https://github.com/secure-systems-lab/ssl_crypto/pull/4

Note: crypto-related modules have been moved to a separate GitHub repository.

vladimir-v-diaz:
ECDSA has been added to the specification in https://github.com/theupdateframework/tuf/pull/498 and implemented in https://github.com/secure-systems-lab/securesystemslib/pull/18.

PR #498, opened 2017-10-11
Add ecdsa-sha2-nistp256 to specification

vladimir-v-diaz:
**Fixes issue #**:
#355  

Add ECDSA (ecdsa-sha2-nistp256) to specification, as requested in issue #355.  ECDSA has been
implemented in https://github.com/secure-systems-lab/securesystemslib/pull/18.

requirements**:

### DIFF
--- a/docs/tuf-spec.md
+++ b/docs/tuf-spec.md
@@ -1,6 +1,6 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **2 June 2017**
+Last modified: **11 October 2017**
 
 Version: **1.0 (Draft)**
 
--- a/docs/tuf-spec.md
+++ b/docs/tuf-spec.md
@@ -470,15 +470,20 @@ Version: **1.0 (Draft)**
    although TUF is not restricted to any particular signature scheme,
    key type, or cryptographic library:
 
-       "RSASSA-PSS-SHA256" : RSA Probabilistic signature scheme with appendix.
-                      The underlying hash function is SHA256.
+       "rsassa-pss-sha256" : RSA Probabilistic signature scheme with appendix.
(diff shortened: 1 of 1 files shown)


docs/tuf-spec.md
```diff
@@ -491,9 +496,9 @@ Version: **1.0 (Draft)**
    used to sign documents.  The type determines the interpretation of
    KEYVAL.
```
awwad: How does the type determine the interpretation of KEYVAL? Type here implies ecdsa/rsa/ed25519, not public/private. Is public/private meant?

docs/tuf-spec.md
```diff
@@ -514,6 +519,15 @@ Version: **1.0 (Draft)**
...
 
+   The 'ecdsa' format is:
+
+        { "keytype" : "ecdsa-sha2-nistp256",
+          "scheme" : "ecdsa-sha2-nistp256",
+          "keyval" : { "public" : PUBLIC}
+        }
+
+   where PUBLIC is in PEM format and a string.
+
```
awwad: I don't see anywhere in the spec where private key formats are explained. Perhaps you should add a note after these examples or something, saying that for a private key, keyval will also contain "private": PRIVATE.

vladimir-v-diaz: We previously included a "private" attribute in KEYVAL, but later decided to remove altogether from metadata.  Since public metadata is never expected to contain a private key, people argued that the "private" attribute shouldn't exist.

Although the repository tool stores private keys in KEYVAL for convenience, it is not required by the specification.  Taking into account that adopters don't have to worry about producing metadata with a "private" element, should we bother explaining it in the specification?

PR #547, opened 2017-12-04
Adjust dependencies

awwad:
Update and reorder dependencies, and add pyasn1 dependency.

pyasn1 was considered optional for this fork, but the fact that this
branch depends on pyasn1 being pinned at 0.2.2 (this is fixed in
branch `refine_asn1_code_tuf`) means that the dependency should
really be listed in the requirements files.

Also make sure that `pip install -e .` comes *after* the listed
dependencies.


awwad:
Whoops, wrong repo. :P

PR #582, opened 2018-01-11
Read and write keys from memory: Fix issue#571

### LINKED ISSUE #571: Option to generate key in RAM instead of on disk?

Presently, there is only an option to generate a key on disk (e.g., using the ```tuf.repository_tool.generate_and_write_xxx_keypair("filename")``` methods). Sometimes, it is useful to generate a key in RAM only, so that the key is never written to persistent / volatile storage. This is useful in scenarios where we want to use a throwaway key (e.g., for the top-level targets role in [PEP 458](https://www.python.org/dev/peps/pep-0458/)).

**Current behavior**:

There are only methods to write and read keys from disk.

**Expected behavior**:

There should be methods to write and read keys from both volatile / nonvolatile storage.

vladimir-v-diaz:
**Fixes issue #**:

Closes #571.

This pull request adds wrapper functions to [securesystemslib](https://github.com/secure-systems-lab/securesystemslib) routines that can write and read keys from memory.  The repository tool only provided wrapper functions to read and write from disk.  For example, `import_ed25519_privatekey_from_file(filepath, password)` and `generate_and_write_ed25519_keypair(filepath, password=None)`

Wrapper functions that are being added to the repository tool....

```Python
from tuf.repository_tool import *

generate_ed25519_key()
generate_ecdsa_key()
generate_rsa_key()

import_ecdsakey_from_pem(pem)
import_rsakey_from_pem(pem)
```

This pull request also adds wrapper functions for missing ECDSA functions.
```Python
generate_and_write_ecdsa_keypair(filepath, password=None)
import_ecdsa_privatekey_from_file(filepath, password)
import_ecdsa_publickey_from_file(filepath)
```

requirements**:

Signed-off-by: Vladimir Diaz \<vladimir.v.diaz@gmail.com>

### DIFF
--- a/tuf/repository_tool.py
+++ b/tuf/repository_tool.py
@@ -3174,6 +3174,8 @@ def import_ed25519_publickey_from_file(filepath):
 def import_ed25519_privatekey_from_file(filepath, password):
   return repo_lib.import_ed25519_privatekey_from_file(filepath, password)
 
+# NOTE: securesystemslib cannot presently import an Ed25519 key from PEM.
+
 def generate_and_write_rsa_keypair(filepath, bits, password):
   return repo_lib.generate_and_write_rsa_keypair(filepath, bits, password)
 
@@ -3185,6 +3187,9 @@ def import_rsa_publickey_from_file(filepath):
 
 def import_rsa_privatekey_from_file(filepath, password):
   return repo_lib.import_rsa_privatekey_from_file(filepath, password)
+
+def import_rsakey_from_pem(pem):
+  return securesystemslib.keys.import_rsakey_from_pem(pem)
 
 def generate_and_write_ecdsa_keypair(filepath, password=None):
   return securesystemslib.interface.generate_and_write_ecdsa_keypair(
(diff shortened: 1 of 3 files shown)


tuf/repository_tool.py
```diff
@@ -3157,35 +3157,65 @@ def append_signature(signature, metadata_filepath):
...
+
+def generate_and_write_rsa_keypair(filepath, bits, password):
+  return repo_lib.generate_and_write_rsa_keypair(filepath, bits, password)
+
+def generate_rsa_key():
+  return securesystemslib.keys.generate_rsa_key()
+
+def import_rsa_publickey_from_file(filepath):
+  return repo_lib.import_rsa_publickey_from_file(filepath)
+
```
lukpueh: Is there a reason why some of these calls are passed through an equivalent wrapper in `repository_lib`? I suggest you directly call `securesystemslib` (as you already do in some functions) and shorten `repository_lib`, which would make maintenance easier. Or does that change the API too much?

vladimir-v-diaz:
@lukpueh 
I'll open an issue to refactor `repository_lib.py`, and added a commit to fix issue #591.

PR #590, opened 2018-01-16
WIP: Fixes #589. Allows 2+ roles to delegate to same role

### LINKED ISSUE #589: repository_tool incorrectly prevents distinct delegations to same role
The repository tool's `delegate` function will currently mistakenly prevent you from delegating to the same role from two different roles. Such delegations are expected to be allowed per prior conversation, not forbidden in the spec, and -- except for the error raised below -- seem to be supported by the code.

Simple scenario:
A delegates to C, with some delegation settings (path, keys expected, threshold, termination, etc)
B delegates to C, with possibly different delegation settings

The check raising this error is likely a holdover from the days when role and delegation were more or less synonymous and roles existed in a tree instead of a more general graph.

### Current behavior
```Python
>>> role_a.delegate('role_c', [dk1_pub, dk2_pub, dk3_pub], ["a*"], threshold=2)
>>> role_b.delegate('role_c', [dk1_pub, dk2_pub], ["b*"], threshold=1)
  File "<stdin>", line 1, in <module>
  File "/Users/s/w/tuf_clean/tuf/repository_tool.py", line 2227, in delegate
    raise securesystemslib.exceptions.Error(repr(rolename) + ' already'
securesystemslib.exceptions.Error: 'role_c' already delegated.
```

### Expected behavior
No error should be raised; the second delegation to the same role functions independently.
```Python
>>> role_a.delegate('role_c', [dk1_pub, dk2_pub, dk3_pub], ["a*"], threshold=2)
>>> role_b.delegate('role_c', [dk1_pub, dk2_pub], ["b*"], threshold=1)
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
```

### Fix
This is easily solved by removing the lines that raise the error. It should be verified that no damage is done, however (i.e. that the code does support this edge case correctly) so testing this will take a bit of effort and require updater testing.

awwad:
Removes an incorrect check that prevents delegating to a role if any role has previously delegated to that role. (Previously, if X delegated to A, then Y was prevented from delegating to A.)

Adds a few tests.

See #589 for more details. (Note that the central fix, a removal of a few lines, has been merged already in #638, and so has been removed from this commit history.)

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -140,6 +140,11 @@ For example, to delegate trust of `/foo*.gz` packages to the `foo` role:
 $ repo.py --delegate "/foo*.tgz" --delegatee foo --pubkeys ./keystore/foo.pub
 ```
 
+Note that multiple roles can delegate to the same role; the delegation graph
+does not have to be a tree. Roles A and B might independently delegate paths to
+the same role, C.
+Cycles are detected and cut short in the depth first traversal the updater
+performs to search for metadata about a target.
 
 
 ## Revoke trust ##
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -463,7 +463,7 @@ specification and [METADATA.md](METADATA.md) for a detailed example.
 All of the target files available on the software repository created so far
 have been added to one role (the top-level Targets role).  However, what if
 multiple developers are responsible for the files of a project?  What if
-responsiblity separation is desired?  Performing a delegation, where one role
+responsibility separation is desired?  Performing a delegation, where one role
 delegates trust of some paths to another role, is an option for integrators
 that require additional roles on top of the top-level roles available by
 default.
@@ -508,6 +508,13 @@ Dirty roles: ['timestamp', 'snapshot', 'targets', 'unclaimed']
 # and "timestamp".
 >>> repository.writeall()
 ```
+
+Note that multiple roles can delegate to the same role; the delegation graph
+does not have to be a tree. Roles A and B might independently delegate paths to
+the same role, C.
+Cycles are detected and cut short in the depth first traversal the updater
+performs to search for metadata about a target.
+
 
 #### Revoke Delegated Role ####
 ```python
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1189,6 +1189,11 @@ class TestTargets(unittest.TestCase):
     self.assertEqual(self.targets_object.get_delegated_rolenames(),
                      ['tuf'])
 
+    # Try to delegate to a role that has already been delegated. Expect success.
+    self.targets_object.delegate(rolename, public_keys, paths, threshold,
+        terminating=False, list_of_targets=list_of_targets,
+        path_hash_prefixes=path_hash_prefixes)
+
     # Test for targets that do not exist under the targets directory.
     self.targets_object.revoke(rolename)
     self.assertRaises(securesystemslib.exceptions.Error,
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1147,9 +1147,9 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     targets_directory = os.path.join(self.repository_directory, 'targets')
 
     # Patterns for delegating the directories
-    foo_pattern = '/foo/*'
-    bar_pattern = '/foo/bar/*'
-    baz_pattern = '/foo/baz/*'
+    foo_pattern = 'foo/*'
+    bar_pattern = 'foo/bar/*'
+    baz_pattern = 'foo/baz/*'
 
     os.makedirs(os.path.join(targets_directory, 'foo', 'bar'))
     os.makedirs(os.path.join(targets_directory, 'foo', 'baz'))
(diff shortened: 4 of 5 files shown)


vladimir-v-diaz:
Can you please edit the corresponding test case?https://github.com/theupdateframework/tuf/blob/aa2ab218f22d8682e03c992ea98f88efd155cffd/tests/test_repository_tool.py#L1192-L1196

tuf/repository_tool.py
```diff
@@ -2222,11 +2222,6 @@ def delegate(self, rolename, public_keys, paths, threshold=1,
     if path_hash_prefixes is not None:
       securesystemslib.formats.PATH_HASH_PREFIXES_SCHEMA.check_match(path_hash_prefixes)
 
-    # Check if 'rolename' is not already a delegation.
```
vladimir-v-diaz: It might be good to mention in the docstring, or in a comment, that 2+ roles can delegate to a single role.
https://github.com/awwad/tuf/blob/2e590485e540b822117de76d1ec949731720ba6f/tuf/repository_tool.py#L2134-L2144

awwad:
Will do

tuf/repository_tool.py
awwad: Actually, this doesn't feel natural. There's no reason to suspect that 2+ roles couldn't delegate to a single role in the code, docstring, or comments. I'll take a look at the documentation - maybe there's somewhere to note that there.

awwad:
Rebased to remove conflicts tidily.

awwad:
Fixed the test case to instead expect success, but we have to add some more detailed testing that covers these delegation scenarios, and until we do, this is still WIP.

awwad:
Some tests to add before this PR is OK:

### Scenario 1:
Targets delegates `"*"` to A, which delegates `"a*"` to C
Targets delegates `"*"` to B, which delegates `"b*"` to C
All roles are signed by all keys expected of them. (e.g. C is signed by a threshold of all keys that A and B expect of it for the delegations listed)
C lists targets `a1.zip` and `b1.zip`.

Tests needed:
- [x] all delegations above must succeed
- [ ] `get_one_valid_targetinfo(b1.zip)` should successfully fetch the fileinfo C listed for `b1.zip`
- [ ] `get_one_valid_targetinfo(a1.zip)` should successfully fetch the fileinfo C listed for `a1.zip`

### Scenario 2:
Same as Scenario 1, except that role C is NOT signed by the keys expected of it by role A (but still is signed by the keys expected of it by role B). (This is important to avoid a particular "attack" whose explanation I'm sitting on and will get to in a fresh issue later.)

Tests needed:
- [ ] `get_one_valid_targetinfo(b1.zip)` should successfully fetch the fileinfo C listed for `b1.zip`
- [ ] `get_one_valid_targetinfo(a1.zip)` should of course fail to fetch the fileinfo C listed for `a1.zip`

I may add some tests to this when I go back over my notes later.

awwad:
Will rebase and conflict resolve this due to merging of #638, which performs part of what this pull request was intended to do. This PR will then add the tests I recommended.

awwad:
Rebased. The check I removed here, which you also removed in the later-merged #638, is no longer in the commit history here. I've changed the docs changes as well: while the changes were previously in tuf/README.md, that has been split into docs/CLI.md and docs/TUTORIAL.md, and so the same notes as were added before to tuf/README.md have now been added to those two documents. You may want to take a peek at the changed docs/CLI.md and make sure the extra note isn't too disruptive.

vladimir-v-diaz:
I don't find the changes to `CLI.md` too disruptive.

vladimir-v-diaz:
@awwad 
I want to validate the correctness of the API redesign against this pull request, if it makes sense.

Has the task/test in scenario 2 (https://github.com/theupdateframework/tuf/pull/590#issuecomment-359841774) been implemented in this pull request?  I see that Travis is failing, but that is expected, correct?

awwad:
The checks in the comment capture the current state -- the last two tests in the list still need to be added. There are more tests I'd want for the API redesign, I think, though I'd have to make sure I understand what you mean by that: is the redesign the fix for #660? (For example, is the expectation that the redesign will result in validation of a delegated role to always consult the appropriate delegating role to determine the keys to expect?)

vladimir-v-diaz:
> is the redesign the fix for #660? (For example, is the expectation that the redesign will result in validation of a delegated role to always consult the appropriate delegating role to determine the keys to expect?)

That's correct!  The redesign of the API is outlined in https://github.com/theupdateframework/tuf/issues/574#issuecomment-358115382 and being worked on in https://github.com/vladimir-v-diaz/tuf/tree/refactor_low_level_api.

vladimir-v-diaz:
@awwad Is it okay if I finish the last two tests in https://github.com/theupdateframework/tuf/pull/590#issuecomment-359841774?  I'd like to use this pull request to test https://github.com/theupdateframework/tuf/pull/757.

I can let you know when pull request #757 is ready for a review.

awwad:
Certainly

joshuagl:
Due to imminent refactor efforts this code is unlikely to merge.

PR #619, opened 2018-02-02
Create LICENSE

dankohn:
Can we please add the license here so that the GitHub API will return the correct info for the repo.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
@dankohn 
Does it work if we add `LICENSE-APACHE.txt` and `LICENSE-MIT.txt` to the root directory?  The TUF project is dual-licensed.

vladimir-v-diaz:
Closing in favor of https://github.com/theupdateframework/tuf/pull/621, where we also update affected modules, docs, etc.

@dankohn 
Thanks for bringing this to our attention.  The LICENSE should now work with [licensee](https://github.com/benbalter/licensee) and display on GitHub's repository overview.

PR #667, opened 2018-03-29
Load latest consistent metadata from disk

vladimir-v-diaz:
**Fixes issue #**:

The issue tracker does not have an issue for this task.

This pull request ensures that `repository_tool.load_repository()` loads the latest consistent metadata from disk.  The latest consistent metadata for some role X should be loaded from disk even if a different, lower version of X is listed in Snapshot.  This behavior is needed because roles may independently add/edit metadata without updating Snapshot.

For example, suppose we have a collaborative repo where the Root or some delegated file is modified.  The person responsible for the metadata file signs it with keys that only they control.  The written metadata file now differs from what is specified in the "stale" Snapshot file.  To refresh the repo (new versions of Timestamp and Snapshot must be written), the repository maintainers must load the latest consistent metadata on disk and not what's in the current Snapshot, otherwise the newly updated Root or delegated file is ignored.  Once the latest consistent metadata is loaded, the stale Snapshot file can be re-signed with the latest versions of metadata available on disk.

requirements**:

Signed-off-by: Vladimir Diaz \<vladimir.v.diaz@gmail.com>

### DIFF
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -1491,7 +1491,7 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
   # files found there.  This information is stored in the 'meta' field of
   # 'snapshot.json'.
 
-  for metadata_filename in os.listdir(metadata_directory):
+  for metadata_filename in sorted(os.listdir(metadata_directory), reverse=True):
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
Thanks for the fix, @vladimir-v-diaz !

Issue #736, opened 2018-05-30
Bad assumption in roledb leads to persistent freeze attack


awwad:
## Immediate Issue Summary
If the client concludes that its snapshot version is out of date (i.e. it has validated a new timestamp version that lists a snapshot version newer than the one it currently has and trusts), it tries to fetch the new snapshot version. If that fails (can't get snapshot or snapshot can't be validated), then the previously trusted known-to-be-outdated snapshot metadata is deleted, both on disk and in memory. This is correct behavior per the spec. Unfortunately, the in-memory deletion of the snapshot metadata stored in roledb results in an error on any further attempts to update snapshot (see below).

Credit goes to Allen Cain (@eacain) for finding this issue.

## Diagnosis
There is a bug in TUF in the functioning of tuf.sig.verify / tuf.roledb. In this bug, the keys expected to sign for role X (snapshot in this case) are stored in the roledb entry for role X. (In this case, tuf.sig.verify expects to find the keys expected of valid snapshot metadata in roledb's entry for snapshot, not root.) While this information is still updated based only on validated root metadata, this implementation remains wrong for a few reasons. The issue here is that role X may be deleted if a new version fails to validate and it is known that the current version is outdated (as correctly occurs in this case); while that's fine, the implementation cannot continue to expect to find the keys in the deleted entry. These keys come from the delegating role (root in this case), and they should be checked against information indexed by the delegating role, not indexed against the delegated role. This behavior is also related to a known issue (#660) that can cause difficulties if multiple roles delegate to the same delegated Targets role. Other solutions to the issue identified here include changing the role removal code to leave the expected keys cached there from the delegating role; this, however is not advisable as it is somewhat counter-intuitive and might lead to flawed implementation or odd edge cases (similar to this one).

In short, the information should be indexed by delegating role, not by delegated role, solving both of these issues. This issue derives from issue #660, so see that for the broader explanation of the more general issue.

vladimir-v-diaz:
Are you able to reproduce the persistent freeze attack in the latest code?  We should have an integration test for this attack, so I'd be curious to know why it's not encountering this issue.

After reviewing the code, it appears that the role db is rebuilt (including the keys for Snapshot, Timestamp, Targets, and Root) when it validates and installs new Root metadata.  The relevant lines of code:

https://github.com/theupdateframework/tuf/blob/76ca2e83643fa01cf713e194553adb8a38113a83/tuf/client/updater.py#L1776-L1780


[...]

PR #737, opened 2018-05-30
WIP: Fix 736: Rebuild roledb in refresh() even if root unchanged

awwad:
In order to avoid freezes if role info is deleted due to prior
validation failures, make sure that roledb is rebuilt during
client tuf.client.updater.refresh(), even if root did not need
to be updated.

*WIP*: I may want to add some tests and wrap a line.

See Issue 736 for more details.

**Fixes issue #736**

requirements**:


vladimir-v-diaz:
The relocation of the call to `_rebuild_key_and_role_db()` looks good to me.  Another thing that we should do is document, in the relevant docstrings, the assumption that the client has the latest info for the top-level roles before attempting to update any specific metadata.

Note: Travis failed, so we probably need to update 1 or 2 test conditions.

awwad:
Sure

vladimir-v-diaz:
@awwad I went ahead and (1) added a test condition for the persistent freeze attack (2) wrapped the line  in `refresh()` that sets consistent snapshot (3) revised a bunch of comments and docstring to document the new behavior and expectations and (4) fixed the failing unit tests.

I think the pull request now just needs to re-sync with the develop branch.

Thanks for reporting the issue and submitting the initial fix!

tests/test_indefinite_freeze_attack.py
```diff
@@ -401,5 +401,101 @@ def test_with_tuf(self):
...
 
+
+
+    # Test 3 Begin:
+    #
+    # Serve the client expired Snapshot.  The client should reject the given,
+    # expired Snapshot and the locally trusted one, which should now be out of
+    # date.  After the attack, attempt to re-issue a valid Snapshot to verify
+    # that the client is able to recover from the indefinite freeze attack via
+    # the snapshot metadata.
```
awwad: I think it will be a bit confusing to future readers to refer to this as an indefinite freeze attack here without context (since there is nothing that obviously represents an attack in the test below). Perhaps:

```Python
# Serve the client expired Snapshot.  The client should reject the given,
# expired Snapshot and the locally trusted one, which should now be out of
# date.
# After the attack, attempt to re-issue a valid Snapshot to verify that
# the client is still able to update. (A bug previously caused snapshot
# expiration or replay to result in an indefinite freeze.)
```

You could also point to the particular issue, #736, in the comment, though that's probably less important. If that looks good, I can add the commit.

vladimir-v-diaz:
@trishankatdatadog 
Heads up: This is an important change that addresses a persistent freeze attack.  I plant to add it to the next stable release and v0.12.dev1 for testing.

tests/test_indefinite_freeze_attack.py
```diff
@@ -401,5 +401,101 @@ def test_with_tuf(self):
...
+    shutil.copytree(os.path.join(self.repository_directory, 'metadata.staged'),
+                    os.path.join(self.repository_directory, 'metadata'))
+
+    self.repository_updater.refresh()
+    self.assertTrue('snapshot' in self.repository_updater.metadata['current'])
+    self.assertEqual(sorted(['root', 'targets', 'timestamp', 'snapshot']),
+        sorted(self.repository_updater.metadata['current']))
+
+
+
```
awwad: Test looks good.

tests/test_multiple_repositories_integration.py
```diff
@@ -209,11 +209,16 @@ def test_update(self):
...
         sorted(tuf.roledb.get_rolenames('test_repository2')))
 
+    # Note: refresh() resets the known metadata and updates the latest
+    # top-level metadata.
     self.repository_updater.refresh()
 
-    self.assertEqual(sorted(['role1', 'root', 'snapshot', 'targets', 'timestamp']),
+    self.assertEqual(sorted(['root', 'snapshot', 'targets', 'timestamp']),
         sorted(tuf.roledb.get_rolenames('test_repository1')))
-    self.assertEqual(sorted(['role1', 'root', 'snapshot', 'targets', 'timestamp']),
```
awwad: I'm confused by this change. How did this result from the PR?

tuf/client/updater.py
```diff
@@ -2584,11 +2594,11 @@ def targets_of_role(self, rolename='targets'):
     # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
     securesystemslib.formats.RELPATH_SCHEMA.check_match(rolename)
 
+    self._refresh_targets_metadata(refresh_all_delegated_roles=True)
+
```
awwad: Why is refresh_all_delegated_roles necessary? This could be quite a few roles.


Next point is a bit complex:
`targets_of_role` remains a strange function to me if called on anything other than the Targets role itself. If the indicated role is a delegated role, it just isn't clear what it means to obtain trustworthy target info for whatever files listed by a trustworthy version of that role. (Whose word are we taking on the latter?)
Perhaps the best we can do is to: obtain some not-validated or partially-validated version of the delegated role (because delegation pathways depend on file paths and we haven't specified one), get the list of targets specified by that not-validated role file, and then properly obtain trustworthy target info for those file paths (whether or not they are ultimately provided by the same role?).

This more complex point plays into the correct way to run targets_of_role on a delegated role.

tests/test_multiple_repositories_integration.py
vladimir-v-diaz: refresh() rebuilds the roledb and only refreshes top-level metadata.  See line 212.

trishankatdatadog:
Does this break backwards-compatibility?

tests/test_indefinite_freeze_attack.py
vladimir-v-diaz: I like the comment that you added, but we should also point to #736 so the reader can understand how the bug was caused.

I'd appreciate you adding the commit, if you are not busy.

tuf/client/updater.py
vladimir-v-diaz: My thinking is that `all_targets()` and `targets_of_roles()` should be for  clients that do not know ahead of time which targets are available on the repo, or provided by a particular role.

Community repos will certainly use get_one_valid_targetinfo(), but we might want to provide all_targets() and targets_of_roles() for simpler use cases / adopters.

vladimir-v-diaz: After an offline discussion, @awwad and I came to the conclusion that it's better to not support `all_targets()` and `targets_of_role()`.  We'll open a separate issue to remove these two functions.

awwad:
No, it can't think of how it would break backward compatibility. (If it causes a problem, of course, LMK.)

BTW, @trishankatdatadog, do you use targets_of_role or all_targets?

trishankatdatadog:
@awwad No, we use `updated_targets()` https://github.com/DataDog/pip/blob/4b191d586172bc7dee665599f819ef05b45b40f1/src/pip/_internal/download.py#L136-L154

awwad:
(And from there, it looks like you get target info via get_one_valid_targetinfo)

Okay, good. (:

trishankatdatadog:
Thanks for checking and verifying, @awwad, much appreciated 🙇

trishankatdatadog:
Hi, any idea when v0.12.dev1 will be released on PyPI for testing? Thanks!

vladimir-v-diaz:
I can do it sometime today, will keep you posted here.

vladimir-v-diaz:
@trishankatdatadog 
update: v0.12.dev1 was released to PyPI for testing.

trishankatdatadog:
I'll test it, thanks!

PR #770, opened 2018-08-14
(WIP) Update dependencies 2018 08

awwad:
Replaces PR #769.

This updates the project's dependencies. See #769 for changelogs.

Currently, this results in some linter warnings. I'll comb through those.


awwad:
pylint 1.9.3 -> 2.0.0 jump results in some linter warnings

awwad:
A try-except-raise checker was added to pylint in 2.0.0 (originally in [this commit](https://github.com/PyCQA/pylint/pull/1939)), raising a warning (error downgraded to warning in 2.1) if there's an immediate re-raise in an except, outside of some special cases.

Here's one of the offending constructions (or see it in context [here](https://github.com/theupdateframework/tuf/blob/0ca83355e8eae0e6e7563266629f0111da5394a9/tuf/sig.py#L197-L199))
```
        try:
          # ....
        # Unknown role, re-raise exception.
        except tuf.exceptions.UnknownRoleError:
          raise
        ## Note: no further except or else clauses!
```

It's not clear to me why this construction is useful, beyond the guess that it might indicate to a reader that an exception of a certain type is expected to occur here in certain situations.... That alone doesn't seem worth the extra bit of code and the wrinkle in the exception stack, though.

**Is anyone aware of a good reason to keep a construction like the above?**
@alyptik @SantiagoTorres 
(Note that if there are further except clauses, there are reasons to have an immediate re-raise -- see [pylint 2323](https://github.com/PyCQA/pylint/issues/2323) -- the case of excluding a subclass re-raise from a broader exception handling clause. Incidentally, this would still be something to avoid, but it does make some sense....)

awwad:
PR #771 fixes the code construction issues pylint > 2.0.0 finds. Since this PR does not introduce any of those issues, it doesn't need to wait to be merged until after 771.

awwad:
As best I can tell, updating these dependencies doesn't break anything, so I could use a review at this point and I'll add sign-off and merge without the linter passing (see above comments).

awwad:
@trishankatdatadog, can you tell me if any of these dependency changes would be an issue for you?

trishankatdatadog:
Updating `cryptography` shouldn't bother us. Thanks for checking!

PR #773, opened 2018-08-20
Use PEP 508 to make dev-requirements compatible with Python 2 & 3

awwad:
Use PEP 508 to make dev-requirements compatible with Python 2 & 3
by specifying the appropriate versions of pylint and astroid
for Python2. pylint and astroid dropped Python2 compatibility
for their latest major revisions, so dev-requirements now pins
the latest compatible revision for Python2 if installing in a
Python2 environment.

**Fixes issue #**: No issue previously identified

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,7 +6,8 @@
 #
 -e .
 asn1crypto==0.24.0
-astroid==2.0.4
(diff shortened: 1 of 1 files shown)


awwad:
Note that we don't test the use of dev-requirements.txt in any automated way.

PR #775, opened 2018-08-20
Fix TUTORIAL.md and add regression testing for future issues

awwad:
Add a regression test for the tutorial instructions. Running test_tutorial.py attempts the commands replicated from TUTORIAL.md. This should help us avoid breaking the tutorial with future changes without noticing by having automated testing run the tutorial and produce helpful output.

It would be wise to add a similar test for the client use instructions.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,7 +25,7 @@ init:
 
 install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
-  - python -m pip install -U pip
+  - python -m pip install -U pip setuptools
   - pip install -e .
   - pip install securesystemslib[crypto,pynacl]
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,6 +28,7 @@ install:
   - python -m pip install -U pip setuptools
   - pip install -e .
   - pip install securesystemslib[crypto,pynacl]
+  - if %PYTHON_VERSION%==2.7 pip install mock
 
 build: false
 
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -115,7 +115,8 @@ The following four key files should now exist:
 
 If a filepath is not given, the KEYID of the generated key is used as the
 filename.  The key files are written to the current working directory.
-```
+```python
+# Continuing from the previous section . . .
 >>> generate_and_write_rsa_keypair()
 Enter a password for the encrypted RSA key (/path/to/b5b8de8aeda674bce948fbe82cab07e309d6775fc0ec299199d16746dc2bd54c):
 Confirm:
@@ -334,6 +335,11 @@ repository, the `add_targets()` method of a Targets role can be called to add
 the target filepaths to metadata.
 
 ```python
+# Continuing from the previous section . . .
+
+# NOTE: If you exited the Python interactive interpreter above you need to
+# re-import the repository_tool-functions and re-load the repository and
+# signing keys.
 >>> from tuf.repository_tool import *
 
 # The 'os' module is needed to gather file attributes, which will be included
@@ -387,8 +393,10 @@ metadata.  `snapshot.json` keys must be loaded and its metadata signed because
 `timestamp.json` role must also be signed.
 
 ```Python
-# The private key of the updated targets metadata must be loaded before it can
-# be signed and written (Note the load_repository() call above).
+# Continuing from the previous section . . .
+
+# The private key of the updated targets metadata must be re-loaded before it
+# can be signed and written (Note the load_repository() call above).
 >>> private_targets_key = import_rsa_privatekey_from_file('targets_key')
 Enter a password for the encrypted RSA key (/path/to/targets_key):
 
@@ -600,7 +608,10 @@ delegate_hashed_bins(list_of_targets, keys_of_hashed_bins, number_of_bins)
 We next provide a complete example of retrieving target paths to add to hashed
 bins, performing the hashed bin delegations, signing them, and delegating paths
 to some role.
+
 ```Python
+# Continuing from the previous section . . .
+
 # Get a list of target paths for the hashed bins.
 >>> targets = \
   repository.get_filepaths_in_directory('repository/targets/myproject', recursive_walk=True)
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -521,7 +521,7 @@ Enter a password for the encrypted RSA key (/path/to/unclaimed_key):
 <!--
 TODO: Integrate section with an updated delegation tutorial.
 As it is now, it just messes up the state of the repository, i.e. marks
-"unclaimed" as dirty, although there is nothing newto write.
+"unclaimed" as dirty, although there is nothing new to write.
 
 #### Revoke Delegated Role ####
 ```python
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -647,11 +647,24 @@ to some role.
 >>> targets = repository.get_filepaths_in_directory(
 ...     'repository/targets/myproject', recursive_walk=True)
 
+# Delegate trust to 32 hashed bin roles. Each role is responsible for the set
+# of target files, determined by the path hash prefix. TUF evenly distributes
+# hexadecimal ranges over the chosen number of bins (see output).
+# To initialize the bins we use one key, which TUF warns us about (see output).
+# However, we can assign separate keys to each bin, with the method used in
+# previous sections, accessing a bin by its hash prefix range name, e.g.:
+# "repository.targets('00-07').add_verification_key('public_00-07_key')".
 >>> repository.targets('unclaimed').delegate_hashed_bins(
 ...     targets, [public_unclaimed_key], 32)
+Creating hashed bin delegations.
+1 total targets.
+32 hashed bins.
+256 total hash prefixes.
+Each bin ranges over 8 hash prefixes.
+Adding a verification key that has already been used. [repeated 32x]
 
-# delegated_hashed_bins() only assigns the public key(s) of the hashed bins, so
-# the private keys may be manually loaded as follows:
+# The hashed bin roles can also be accessed by iterating the "delegations"
+# property of the delegating role, which we do here to load the signing key.
 >>> for delegation in repository.targets('unclaimed').delegations:
 ...   delegation.load_signing_key(private_unclaimed_key)
 
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -581,14 +581,24 @@ target file names specified in metadata do not contain digests in their names.)
 The repository maintainer is responsible for the duration of multiple versions
 of metadata and target files available on a repository.  Generating consistent
 metadata and target files on the repository is enabled by setting the
-`consistent_snapshot` argument of writeall() or write(). Note that changing the
-consistent_snapshot setting involves writing a new version of root.
+`consistent_snapshot` argument of `writeall()` or `write()` . Note that
+changing the consistent_snapshot setting involves writing a new version of
+root.
+
+<!--
+TODO: Integrate section with an updated consistent snapshot tutorial.
+As it is now, it just messes up the state of the repository, i.e. marks
+"root" as dirty, although all other metadata needs to be re-written with
+<VERSION> prefix and target files need to be re-written with <HASH> prefix in
+their filenames.
+
 ```Python
     # ----- Tutorial Section: Consistent Snapshots
 >>> repository.root.load_signing_key(private_root_key)
 >>> repository.root.load_signing_key(private_root_key2)
 >>> repository.writeall(consistent_snapshot=True)
 ```
+-->
 
 ## Delegate to Hashed Bins ##
 Why use hashed bin delegations?
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -227,12 +227,6 @@ top-level roles, including itself.
 >>> repository.root.load_signing_key(private_root_key)
 >>> repository.root.load_signing_key(private_root_key2)
 
-# Print the roles that are "dirty" (i.e., that have changed and have not yet
-# been written to disk.  Root should be dirty because verification keys were
-# added, signing keys loaded, and a threshold added)
->>> repository.dirty_roles()
-Dirty roles: ['root']
-
 # repository.status() shows missing verification and signing keys for the
 # top-level roles, and whether signatures can be created (also see #955).
 # This output shows that so far only the "root" role meets the key threshold and
@@ -297,8 +291,8 @@ Enter a password for the encrypted RSA key (/path/to/timestamp_key):
 # week), timestamp(1 day).
 >>> repository.timestamp.expiration = datetime.datetime(2080, 10, 28, 12, 8)
 
->>> repository.dirty_roles()
-Dirty roles: ['snapshot', 'targets', 'timestamp']
+# Mark roles for metadata update (see #964, #958)
+>>> repository.mark_dirty(['root', 'snapshot', 'targets', 'timestamp'])
 
 # Write all metadata to "repository/metadata.staged/"
 >>> repository.writeall()
@@ -414,8 +408,8 @@ Enter a password for the encrypted RSA key (/path/to/snapshot_key):
 Enter a password for the encrypted RSA key (/path/to/timestamp_key):
 >>> repository.timestamp.load_signing_key(private_timestamp_key)
 
->>> repository.dirty_roles()
-Dirty roles: ['root', 'snapshot', 'targets', 'timestamp']
+# Mark roles for metadata update (see #964, #958)
+>>> repository.mark_dirty(['snapshot', 'targets', 'timestamp'])
 
 # Generate new versions of the modified top-level metadata (targets, snapshot,
 # and timestamp).
@@ -433,11 +427,10 @@ new metadata to disk.
 # Remove a target file listed in the "targets" metadata.  The target file is
 # not actually deleted from the file system.
 >>> repository.targets.remove_target('myproject/file4.txt')
->>> repository.dirty_roles()
-Dirty roles: ['targets']
 
-# Mark roles as dirty that have not changed but need to be updated (see #958)
->>> repository.mark_dirty(['snapshot', 'timestamp'])
+# Mark roles for metadata update (see #964, #958)
+>>> repository.mark_dirty(['snapshot', 'targets', 'timestamp'])
+
 >>> repository.writeall()
 ```
 
@@ -519,18 +512,16 @@ Enter a password for the encrypted RSA key (/path/to/unclaimed_key):
 
 >>> repository.targets("unclaimed").load_signing_key(private_unclaimed_key)
 
->>> repository.dirty_roles()
-Dirty roles: ['targets', 'unclaimed']
+# Mark roles for metadata update (see #964, #958)
+>>> repository.mark_dirty(['snapshot', 'targets','timestamp', 'unclaimed'])
 
-# Mark roles as dirty that have not changed but need to be updated (see #958)
->>> repository.mark_dirty(["snapshot", "timestamp"])
 >>> repository.writeall()
 ```
 
 <!--
 TODO: Integrate section with an updated delegation tutorial.
 As it is now, it just messes up the state of the repository, i.e. marks
-"unclaimed" as dirty, although there is nothing new to write.
+"unclaimed" as dirty, although there is nothing newto write.
 
 #### Revoke Delegated Role ####
 ```python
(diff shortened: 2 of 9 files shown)


tests/test_tutorial.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+
+def clean_test_environment():
+  """
+  Delete temporary files and directories from this test (or with the same name
+  as those created by this test...).
+  """
+  for directory in ['repository', 'my_repo', 'client',
+      'repository/targets/my_project']:
```
lukpueh: `repository/targets/my_project` is already deleted in the first iteration of this loop, i.e. `shutil.rmtree('repository')`

tests/test_tutorial.py
```diff
@@ -0,0 +1,413 @@
...
+    if os.path.exists(directory):
+      shutil.rmtree(directory)
+
+  for fname in [
+        os.path.join('repository', 'targets', 'file1.txt'),
+        os.path.join('repository', 'targets', 'file2.txt'),
+        os.path.join('repository', 'targets', 'file3.txt'),
```
lukpueh: You're already deleting the `repository` dir above.

tests/test_tutorial.py
```diff
@@ -0,0 +1,413 @@
...
+        'targets_key.pub',
+        'snapshot_key',
+        'snapshot_key.pub',
+        'timestamp_key',
+        'timestamp_key.pub',
+        'unclaimed_key',
+        'unclaimed_key.pub']:
+    if os.path.exists(fname):
+      os.remove(fname)
+
```
lukpueh: On a more general note, would it make sense to create and change into a temporary test directory in `TestTutorial.setUp` and change back and remove the temporary directory in `TestTutorial.tearDown`?

awwad:
It's a good thought.  I'll have to try it out to see how practical it proves to be.  If it makes the tutorial harder to read, that's a dealbreaker.

lukpueh:
Taking over @awwad's work, I addressed my own [inline review comments from above](https://github.com/theupdateframework/tuf/pull/775#pullrequestreview-172429283) and most of the suggestions I made in https://github.com/theupdateframework/tuf/issues/808#issuecomment-487596129.

The tutorial is now fully copy-pastable (runnable) and should generally not leave the repository in an inconsistent state.

Remaining issues as described in #808 should be fixed in separate PRs.

I suggest reviewers to: 
- read #775 (to not restate known problems),
- try out the tutorial (in an interactive Python interpreter),
- focus on my diff, i.e. https://github.com/theupdateframework/tuf/compare/94055c5c65ff6cb127bb183f8b40af245806aafb...add_tutorial_regression_test (@awwad's commits I already reviewed).

Please read commit messages for details!

docs/TUTORIAL.md
```diff
@@ -230,25 +219,30 @@ top-level roles, including itself.
...
+# added, signing keys loaded, and a threshold added)
 >>> repository.dirty_roles()
 Dirty roles: ['root']
 
-# The status() function also prints the next role that needs editing.  In this
-# example, the 'targets' role needs editing next, since the root role is now
-# fully valid.
+# `repository.status()` shows missing verification and signing keys for the
+# top-level roles, and whether signatures can be created (also see #955).
+# Thi output shows that so far only the "root" role meets the key threshold and
```
joshuagl:
```suggestion
# This output shows that so far only the "root" role meets the key threshold and
```
Fix typo

docs/TUTORIAL.md
```diff
@@ -230,25 +219,30 @@ top-level roles, including itself.
...
+# Print the roles that are "dirty" (i.e., that have changed and have not yet
+# been written to disk.  Root should be dirty because verification keys were
+# added, signing keys loaded, and a threshold added)
 >>> repository.dirty_roles()
 Dirty roles: ['root']
 
-# The status() function also prints the next role that needs editing.  In this
-# example, the 'targets' role needs editing next, since the root role is now
-# fully valid.
+# `repository.status()` shows missing verification and signing keys for the
```
joshuagl:
```suggestion
# repository.status() shows missing verification and signing keys for the
```
Drop the code formatting, as other functions in comments aren't using it and it doesn't render correctly

docs/TUTORIAL.md
```diff
@@ -230,25 +219,30 @@ top-level roles, including itself.
...
+# Thi output shows that so far only the "root" role meets the key threshold and
+# can successfully sign its metadata.
 >>> repository.status()
 'targets' role contains 0 / 1 public keys.
+'snapshot' role contains 0 / 1 public keys.
+'timestamp' role contains 0 / 1 public keys.
+'root' role contains 2 / 2 signatures.
+'targets' role contains 0 / 1 signatures.
 
 # In the next section, update the other top-level roles and create a repository
```
joshuagl:
```suggestion
# In the next section we update the other top-level roles and create a repository
```

docs/TUTORIAL.md
```diff
@@ -376,17 +377,12 @@ the target filepaths to metadata.
...
 >>> repository.targets.add_targets(list_of_targets)
 
-# Note that you can also add targets to existing delegated targets roles,
-# accessing them this way:
->>> repository.targets('<delegated rolename>').add_target(...)
->>> repository.targets('<delegated rolename>').add_targets(...)
-
 # Individual target files may also be added to roles, including custom data
 # about the target.  In the example below, file permissions of the target
 # (octal number specifying file access for owner, group, others (e.g., 0755) is
```
joshuagl:
```suggestion
# (octal number specifying file access for owner, group, others e.g., 0755) is
```
We open more braces than we close, so I removed the nesting.

docs/TUTORIAL.md
```diff
@@ -399,26 +395,27 @@ metadata.  `snapshot.json` keys must be loaded and its metadata signed because
...
->>> private_timestamp_key = import_rsa_privatekey_from_file("keystore/timestamp_key")
-Enter a password for the encrypted RSA key (/path/to/keystore/timestamp_key):
+>>> private_timestamp_key = import_rsa_privatekey_from_file('timestamp_key')
+Enter a password for the encrypted RSA key (/path/to/timestamp_key):
 >>> repository.timestamp.load_signing_key(private_timestamp_key)
 
-# Which roles are dirty?
 >>> repository.dirty_roles()
-Dirty roles: ['timestamp', 'snapshot', 'targets']
+Dirty roles: ['root', 'snapshot', 'targets', 'timestamp']
```
joshuagl: At this point, I only have:
```
Dirty roles: ['targets']
```
I believe I ran through the tutorial verbatim, apart from that I didn't exit ipython so I didn't have to re-import and reload the keys.

tests/test_tutorial.py
```diff
@@ -0,0 +1,396 @@
...
+    append_signature(
+        signature,
+        os.path.join('repository', 'metadata.staged', 'timestamp.json'))
+
+
+
+    # ----- Tutorial Section: Delegations
+    generate_and_write_rsa_keypair(
+        'unclaimed_key', bits=2048, password='password')
+    public_unclaimed_key = import_rsa_publickey_from_file('unclaimed_key.pub')
```
joshuagl, line 315: A very minor nit, but the tutorial isn't consistent in its use of `'` vs `"` for strings. I suspect it highlights the new content Lukas added.

tests/test_tutorial.py
```diff
@@ -0,0 +1,396 @@
...
+    repository.targets.delegate(
+        'unclaimed', [public_unclaimed_key], ['myproject/*.txt'])
```
joshuagl, line 317: This emits:
> 'myproject/*.txt' is not located in the repository's targets directory: '/home/joshuagl/tuf-tut/repository/targets'

I think this is the first point in the tutorial we haven't warned the user to expect output.

tests/test_tutorial.py
```diff
@@ -0,0 +1,396 @@
...
+        os.path.join('repository', 'metadata'))
+
+
+    # ----- Tutorial Section: Delegate to Hashed Bins
+
+    targets = repository.get_filepaths_in_directory(
+        os.path.join('repository', 'targets', 'myproject'), recursive_walk=True)
+
+    repository.targets('unclaimed').delegate_hashed_bins(
+        targets, [public_unclaimed_key], 32)
```
joshuagl: Should we make a remark about the output here? I get:

```
Creating hashed bin delegations.
1 total targets.
32 hashed bins.
256 total hash prefixes.
Each bin ranges over 8 hash prefixes.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
Adding a verification key that has already been used.
```

docs/TUTORIAL.md
```diff
@@ -608,19 +630,34 @@ delegate_hashed_bins(list_of_targets, keys_of_hashed_bins, number_of_bins)
...
 # the private keys may be manually loaded as follows:
 >>> for delegation in repository.targets('unclaimed').delegations:
 ...   delegation.load_signing_key(private_unclaimed_key)
 
-# Delegated roles can be restricted to particular paths with add_restricted_paths().
->>> repository.targets('unclaimed').add_restricted_paths('repository/targets/myproject/*', 'django')
+
+>>> repository.dirty_roles()
+Dirty roles: ['00-07', '08-0f', '10-17', '18-1f', '20-27', '28-2f', '30-37', '38-3f', '40-47', '48-4f', '50-57', '58-5f', '60-67', '68-6f', '70-77', '78-7f', '80-87', '88-8f', '90-7', '98-9f', 'a0-a7', 'a8-af', 'b0-b7', 'b8-bf', 'c0-c7', 'c8-cf', 'd0-d7', 'd8-df', 'e0-e7', 'e8-ef', 'f0-f7', 'f8-ff', 'unclaimed']
+
```
joshuagl: I get the same roles but sorted differently, which I'd expect knowing that various Python data structures have non-deterministic ordering. Does it make sense to make these deterministic (i.e. `sorted()` this list)?

joshuagl: Patch to sort the list before printing in `dirty_roles()` submitted in #961

joshuagl: #961 was merged 🎉

lukpueh:
Thank you very much for the review, @joshuagl. Let me address/reply to your excellent observations!

appveyor.yml
```diff
@@ -21,7 +21,7 @@ init:
 
 install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
-  - python -m pip install -U pip
+  - python -m pip install -U pip setuptools
```
joshuagl, line 28: I think we need to explicitly install mock for Python 2.7 on appveyor?

lukpueh: True. Let me fix this.

docs/TUTORIAL.md
lukpueh: I remember a discussion with @awwad, where we concluded that backticks  might help (GitHub/Markdown-experienced) people to parse code statements/expressions in code documentation, even it's not rendered as Markdown.

But you're right, consistency is more important. Applying your suggestion...

docs/TUTORIAL.md
lukpueh: Yes, the tuf repository code has a bit of a problem with state-keeping. `dirty_roles()` shows roles that have been changed in memory, e.g. by loading a private key for a role (`load_signing_key()`), which is necessary after loading an existing repository (`load_repository()`). This does not mean, however, that the according metadata on disk actually needs to be updated, although this is exactly what's done with dirty roles when calling `writeall()`.

This clearly needs a revision. But probably not as part of the current PR.

A related issue, that is `writeall()` ignoring roles that were not explicitly changed and thus not marked as dirty, even though they should be updated in dependency of other roles is documented in #958.

joshuagl: Thanks for the explanation @lukpueh. I wasn't sure why the output I saw differed from that shown in the tutorial, so wanted to point it out. I'm happy with this not being addressed in the current PR, should we capture the issue around state-keeping so we don't forget?

lukpueh: I wonder if the best quick fix for the tutorial would be to just remove the calls to `dirty_roles()` and instead add a call to `mark_dirty()` passing all the roles that need to be written in a subsequent call to `writeall()`. This guarantees that the repository on disk is always in a consistent state, even if the interpreter was exited.

What do you think, @joshuagl?

tests/test_tutorial.py
lukpueh: Yes, you are absolutely right! I pointed this out in https://github.com/theupdateframework/tuf/issues/808#issuecomment-487596129 (3rd item), which I suggested in https://github.com/theupdateframework/tuf/pull/775#issuecomment-556052195 should be fixed along with other remaining issues in separate PRs.

Apologies for the many sources of truth here. I can see how it is hard for a reviewer to keep track of all the things that need to be fixed. Thanks for bearing with me! :)

docs/TUTORIAL.md
joshuagl: I really like this idea. I think it demonstrates how things fit together and helps make some of the workflow clearer.

tests/test_tutorial.py
joshuagl: Apologies, I read through #775 but forgot to read through #808 before reviewing. Glad we have this tracked, let's ignore for this PR.

docs/TUTORIAL.md
lukpueh: This is odd. I had already added this change in ac01033 specifically for that reason, albeit one level deeper.  Are you sure you got a different sorting?

joshuagl: Oh no, I was running the steps from the `develop` branch, not a branch with the changes from this PR 🤦‍♂

We should revert/undo ff21b36. Apologies @lukpueh!

tests/test_tutorial.py
lukpueh: I briefly thought about it but left it to be fixed along with other issues in the tutorial. 

Moreover, I have the suspicion that this is a bogus warning. It seems to literally check for a file called "myproject/*.txt" ignoring the fact that this is a path pattern.

What to you think about creating an issue, and adding a comment in the tutorial snippet that asks the reader to ignore that output, referring to that issue?

joshuagl: That sounds reasonable. Let's be sure to have the issue mention that the tutorial needs updating once the issue is resolved.

tests/test_tutorial.py
lukpueh: Yes! It is another instance of me thinking this should be addressed. More specifically with the 2nd to last item in https://github.com/theupdateframework/tuf/issues/808#issuecomment-487596129.

But adding a comment there that briefly explains the output does not hurt.

docs/TUTORIAL.md
lukpueh: No worries at all! :) What do you think about cherry-picking 282c9aa98b52d654f9610726a272c9c8bf34d2d1 into this PR and reverting #961?

joshuagl: Sounds good to me

tests/test_tutorial.py
lukpueh: My suspicion was wrong. The warning does not stem from an invalid file check but rather an unexpected file path prefix check. I documented the issue and a proposed fix in https://github.com/theupdateframework/tuf/issues/963.

docs/TUTORIAL.md
```diff
@@ -519,18 +508,16 @@ Enter a password for the encrypted RSA key (/path/to/unclaimed_key):
...
-# Mark roles as dirty that have not changed but need to be updated (see #958)
->>> repository.mark_dirty(["snapshot", "timestamp"])
 >>> repository.writeall()
 ```
 
 <!--
 TODO: Integrate section with an updated delegation tutorial.
 As it is now, it just messes up the state of the repository, i.e. marks
-"unclaimed" as dirty, although there is nothing new to write.
+"unclaimed" as dirty, although there is nothing newto write.
```
joshuagl: Accidental space deletion here?

docs/TUTORIAL.md
```diff
@@ -414,8 +408,8 @@ Enter a password for the encrypted RSA key (/path/to/snapshot_key):
 Enter a password for the encrypted RSA key (/path/to/timestamp_key):
 >>> repository.timestamp.load_signing_key(private_timestamp_key)
 
->>> repository.dirty_roles()
-Dirty roles: ['root', 'snapshot', 'targets', 'timestamp']
+# Mark roles for metadata update (see #XXX, #958)
+>>> repository.mark_dirty(['snapshot', 'targets', 'timestamp'])
```
joshuagl, line 412: Should root be marked dirty here also? It was in the output of `dirty_roles()` above.

docs/TUTORIAL.md
lukpueh: Oops, yes.

docs/TUTORIAL.md
lukpueh: No. The output was wrong before in TUTORIAL.md (test_tutorial.py had it right). At this point in the tutorial, root does not need to be re-written.

appveyor.yml
lukpueh: @joshuagl, do you happen to know how to install the dependency only for Python 2.7 builds on appveyor?

docs/TUTORIAL.md
lukpueh: But please double check! :)

appveyor.yml
joshuagl: Not off the top of my head, but I think we should be able to use `for.matrix.only/.except`:
https://www.appveyor.com/docs/build-configuration/#specializing-matrix-job-configuration

lukpueh: Thanks. Maybe 932c025 works too.

trishankatdatadog:
Looks like @joshuagl and @lukpueh have got this, I will let you guys handle it :)

appveyor.yml
```diff
@@ -24,6 +24,7 @@ install:
   - python -m pip install -U pip setuptools
   - pip install -e .
   - pip install securesystemslib[crypto,pynacl]
+  - if %PYTHON_VERSION%==2.7 pip install mock
```
joshuagl, line 31: This does the trick, nice work @lukpueh.

lukpueh:
@joshuagl, do you think I addressed your comments? @mnm678, would you mind leaving a review too? I'd really like to merge this ASAP.

docs/TUTORIAL.md
```diff
@@ -138,12 +127,12 @@ Confirm:
...
 
 # Import an existing public key.
->>> public_root_key = import_rsa_publickey_from_file("keystore/root_key.pub")
+>>> public_root_key = import_rsa_publickey_from_file("root_key.pub")
 
 # Import an existing private key.  Importing a private key requires a password,
 # whereas importing a public key does not.
->>> private_root_key = import_rsa_privatekey_from_file("keystore/root_key")
-Enter a password for the encrypted RSA key (/path/to/keystore/root_key):
+>>> private_root_key = import_rsa_privatekey_from_file("root_key")
```
mnm678, line 134: It might be a good idea to remind the reader that the password is 'password'

docs/TUTORIAL.md
```diff
@@ -138,12 +127,12 @@ Confirm:
...
+Enter a password for the encrypted RSA key (/path/to/root_key):
 ```
 
 `import_rsa_privatekey_from_file()` raises a
```
mnm678, line 138: I got a stack trace as well as this error, is that the correct behavior?

docs/TUTORIAL.md
```diff
@@ -433,34 +424,41 @@ new metadata to disk.
 ```python
 # Continuing from the previous section . . .
 
-# Remove a target file listed in the "targets" metadata.  The target file is
-# not actually deleted from the file system.
->>> repository.targets.remove_target("repository/targets/file3.txt")
```
mnm678, line 438: is this line still needed? As currently written I don't think any files are removed

lukpueh: Thanks for spotting this, @mnm678! I accidentally removed this in a recent commit. It's back now.

docs/TUTORIAL.md
lukpueh: Yes, this is correct behavior. I doubt though that it is useful to show the error or stack trace here in the tutorial. 

Item 7 in https://github.com/theupdateframework/tuf/issues/808#issuecomment-487596129 suggests to remove it. But as with the other open issues in #808 I'd like to do this in a separate PR.

docs/TUTORIAL.md
lukpueh: Good idea. I'll add this to #808, to be addressed in a separate PR.

lukpueh:
Python 3.4 builds fail on Travis and Appveyor, due to a recent update of the transitive dependency `colorama`. I suggest to drop 3.4 support/builds (and independently make `colorama` optional, see https://github.com/secure-systems-lab/securesystemslib/issues/179).

lukpueh:
I just rebased on top of develop, which does not test on 3.4 anymore, but instead tests on 3.7 and 3.8. Travis builds pass. So I guess it's time to merge.

Thanks for the initial work, @awwad, and for the great reviews, @joshuagl and @mnm678! 

Remaining issues shall be fixed with #808.

PR #779, opened 2018-08-30
(WIP) ASN.1: Initial commit of asn1 metadata definitions

awwad:
Currently, the TUF reference implementation uses only a JSON-compatible metadata format defined in the [TUF specification](https://github.com/theupdateframework/specification).

It's my intention to add ASN.1 metadata support to the TUF reference implementation (decoding, encoding, signing over, verifying signatures over). If necessary, this can be made into an optional feature or even an external package. Either way, the reference implementation would profit from having an easy means of using ASN.1 metadata.

In this PR, for conversation purposes, is a draft of the metadata definitions (in ASN.1 and in a pyasn1-compatible module).


awwad:
@alyptik Any thoughts on these? (Choices for types, structure)

tuf/encoding/asn1_metadata_definitions.py
```diff
@@ -0,0 +1,205 @@
...
+
+  - add MAX value
+
+  - when providing a default value, rather than a subtyped object constrained to
+     only have one value, provide a specific value (both seem to work, but I
+     think this has the right semantics)
+     e.g. use:   DefaultedNamedType('terminating', univ.Boolean(0))
+     instead of: DefaultedNamedType('terminating', univ.Boolean().subtype(value=0))
+
+  - Move superclass value overrides into bodies of classes
```
alyptik: Very good modification, ACK

tuf/encoding/asn1_metadata_definitions.py
```diff
@@ -0,0 +1,205 @@
...
+# import pyasn1.type.namedval as namedval
+# import pyasn1.type.tag as tag
+from pyasn1.type.namedtype import NamedType, NamedTypes, DefaultedNamedType
+
+# Maximum integer value when bounding integer values below.
+# It does not seem possible to place a minimum without placing a maximum.
+# For thresholds, versions, timestamps, etc., we want only non-negative
+# integers, so we're stuck setting a maximum (or not placing any constraints
+# at all in the ASN.1 metadata definition itself, would would also be fine).
+MAX = 2**32-1
```
alyptik: Is this inclusive or exclusive?

tuf/encoding/asn1_metadata_definitions.py
```diff
@@ -0,0 +1,205 @@
...
+      NamedType('threshold', Threshold()))
+
+
+
+
+## Types used only in Timestamp metadata
+class HashOfSnapshot(univ.Sequence):
+  componentType = NamedTypes(
+      NamedType('filename', FileName()),
+      NamedType('num-hashes', Length()),
```
alyptik: Is there a reason Length() is reused for num-keys, num-keyids, and num-hashes? To clarify, is it possible to have a more descriptive name?

tuf/encoding/metadata_definitions.asn1
```diff
@@ -0,0 +1,302 @@
...
+This file is organized into five sections:
+   Common Definitions
+   Root Role Definitions
+   Timestamp Role Definitions
+   Snapshot Role Definitions
+   Targets Role Definitions
+
+
+A note for implementers of TUF-specification-conformant systems:
+There are constraints and optimizations that you can bake into your ASN.1
```
alyptik: Was this meant to be "These are..."? If so, a better wording might be "The following are..."

tuf/encoding/metadata_definitions.asn1
```diff
@@ -0,0 +1,302 @@
...
+ to compilation issues), so the only change is the removal of
+ StrictFilename, which was functionally equivalent to Filename in any event.
+ I'm also removing length constraints and enumerations to keep this simple
+ for now. Implementers will impose their own length constraints in their
+ ASN.1 definitions (or, if they are able, at a higher level) and perform
+ their own optimizations.
+
+
+Questions:
+ - Would it be better to use fewer empty wrapped classes? e.g. is it better to
```
alyptik: I'm personally not a fan of abstraction which isn't used for some form of data hiding. My vote would be towards yes, less empty classes.

tuf/encoding/metadata_definitions.asn1
```diff
@@ -0,0 +1,302 @@
...
+ - TopLevelDelegation may be a confusing name, since we usually refer to
+   targets-level delegations when we talk about delegations; however, note that
+   the info in TopLevelDelegation is information about root's delegation to the
+   roles, not just role info. keys, threshold, etc. are pieces of a delegation,
+   and this specific confusion around roles vs delegations has been very
+   damaging in the past. (See TUF Issue #660)
+
+ - I've tried to minimize the differences between this ASN.1 definition set and
+   the TUF spec (which uses JSON). Consider each:
+    - No dictionary structures (not clear it's possible?); instead, former
```
alyptik: This is a perfectly fine way to store dictionaries inside lists, ACK

tuf/encoding/asn1_metadata_definitions.py
awwad: It's inclusive, which is a little odd, but that's why this is -1.

tuf/encoding/asn1_metadata_definitions.py
awwad: I guess we could call it SequenceLength. Do you think that's better? Or do you think it's clearer if I just exclude it and use an integer primitive directly in each of these places?

tuf/encoding/metadata_definitions.asn1
awwad: No, it's intended to be "There are", i.e. "There exist". I'll reword it to make it clearer.

joshuagl:
We plan to include this functionality in a future release of tuf (or securesystemslib) once we have completed refactoring efforts.

Issue #209, opened 2014-04-22
Should we replace urllib(2) with requests?


trishankkarthik:
Right now, we have some code copied over from pip to ensure that urllib(2) will verify SSL certificates and so on.

I think we are better off replacing the lower-level urllib(2) code with the [requests](http://docs.python-requests.org/en/latest/) library instead. Even pip does so now.

However, we need to check whether requests will allow us to control things like watching for a slow retrieval attack and so on.

lukpueh:
`requests` was introduced with #781. Closing here.

Issue #204, opened 2014-04-11
Socket timeout default value pretty onerous


trishankkarthik:
Don't know what I was thinking when I [set the default socket timeout value to be 1 second](https://github.com/theupdateframework/tuf/blob/2c55b94a8bfabf61404df3af5a7a04bc91191a1a/tuf/conf.py#L51-L52), but it turns out to be pretty impractical in practice (e.g. client fails to connect via WiFi to server on the same LAN). Minor issue, but should reconsider.

lukpueh:
`SOCKET_TIMEOUT` has in the meanwhile been tweaked several times (see `git log -L :SOCKET_TIMEOUT:tuf/settings.py`). Currently it's at 4 seconds. And there is a pending PR that splits it into `SOCKET_CONNECT_TIMEOUT = 10` and `SOCKET_INTERBYTE_TIMEOUT = 4`. I think we can close this issue.

PR #781, opened 2018-08-30
Update TUF to handle HTTPS proxies

trishankatdatadog:
**Fixes issue #**:

You will see that tuf fails to work via the HTTPS proxy, whereas the `requests` library does.

The problem is this [well-intentioned but outdated code](https://github.com/theupdateframework/tuf/blob/98622521a3a6632d6e44f01f3158d308d0416296/tuf/download.py#L739).

As the comment hints, I had borrowed that VerifiedHTTPSConnection code from pip back then, because I knew that python2 back then was not verifying hostnames on SSL certs. Unfortunately, this code is now severely outdated.

The tuf.download module now uses the `requests` library. I have manually tested that it handles HTTPS proxies.

Changes:
- Install `requests` and its dependencies by default.
- Change the networking code to use `requests` instead of custom, outdated networking code.
- Remove option to specify custom SSL certs via a TUF setting. Instead, we depend on clients exporting the `REQUESTS_CA_BUNDLE` environment variable to `requests`.
- Remove `tuf.settings.SLOW_START_GRACE_PERIOD`, as this no longer seems terribly useful, because `requests` allow for setting connect and / or read timeouts.
- We no longer explicitly check for supported schemes (e.g., `http` and `https` only). This was originally done to prevent potentially unsafe requests to files on disk, but `requests` does not handle opening local files anyway.

TODO:
- [ ] Unfortunately, a few slow retrieval attack tests are failing. I'm too tired right now to figure out what is going on.
- [ ] Use a separate `requests.Session` per repository mirror to avoid sharing state, and accidentally falling for unknown / unforeseen security issues.
- [ ] Add unit tests to see whether: (1) hostnames are verified on SSL certs, and (2) HTTPS proxies are handled transparently
- [ ] Add coverage tests, if necessary

requirements**:


trishankatdatadog:
Cc @JustinCappos @SantiagoTorres @awwad 

Need your eyes to carefully review, edit, and merge! 👀

tuf/download.py
```diff
@@ -35,29 +35,36 @@
...
+# requests to the same host, the underlying TCP connection will be reused,
+# which can result in a significant performance increase (see HTTP persistent
+# connection)."
+#
+# FIXME: There are very likely security implications to sharing the same
+# session. Ideally, tuf.client.updater would use a separate Session for every
+# mirror, and it would pass the Session to be used in this module, but that
+# requires a much more invasive change.
+#
+_session = requests.Session()
```
JustinCappos: Key to fix before merging.  Using a separate session for each hostname part of the URL should be easy.  Perhaps make this a dict?

tuf/download.py
```diff
@@ -172,20 +163,6 @@ def unsafe_download(url, required_length):
   securesystemslib.formats.URL_SCHEMA.check_match(url)
   securesystemslib.formats.LENGTH_SCHEMA.check_match(required_length)
 
-  # Ensure 'url' specifies one of the URI schemes in
```
JustinCappos: We probably should retain a comment explaining that this is where the http / https prefix from the settings.py file is checked (assuming that is the case) and that an ???Exception may be thrown as a result...

tuf/download.py
```diff
@@ -257,14 +231,34 @@ def _download_file(url, required_length, STRICT_REQUIRED_LENGTH=True):
...
-    # this kind can be minimized by adjusting the socket timeout in
-    # settings.py.
-    connection = _open_connection(url)
+    # Attach some default headers to every Session.
+    _session.headers.update({
+        # Tell the server not to compress or modify anything.
+        # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#Directives
+        'Accept-Encoding': 'identity',
+        # The TUF user agent.
+        # TODO: Attach version number, which cannot be easily found right now.
```
JustinCappos: What does this mean?  Do you mean that TUF's communication protocol should have a version of its own so we might use different versions at different times?  Do we want the version of TUF to be here instead?

tests/test_slow_retrieval_attack.py
```diff
@@ -157,10 +157,9 @@ def setUp(self):
     # The slow retrieval server, in mode 2 (1 byte per second), will only
     # sleep for a  total of (target file size) seconds.  Add a target file
     # that contains sufficient number of bytes to trigger a slow retrieval
-    # error.  "sufficient number of bytes" assumed to be
-    # >> 'tuf.settings.SLOW_START_GRACE_PERIOD' bytes.
+    # error.  "sufficient number of bytes" assumed to be 10x more.
     extra_bytes = 8
-    total_bytes = tuf.settings.SLOW_START_GRACE_PERIOD + extra_bytes
+    total_bytes = 100 * extra_bytes
```
SantiagoTorres: I'd be happier if re-used that variable or assigned a constant for this `100`

tuf/download.py
SantiagoTorres: I think it's just overwriting the User-Agent to something that's not `requests` or `python urrlibX`. It'd be cleaner (and easier to analyze traffic) if we used something like `TUF vX`

tuf/settings.py
```diff
@@ -46,11 +46,6 @@
 # /tmp/repositories/django_repo/metadata/current/root.METADATA_EXTENSION
 repositories_directory = None
 
-# A PEM (RFC 1422) file where you may find SSL certificate authorities
-# https://en.wikipedia.org/wiki/Certificate_authority
-# http://docs.python.org/2/library/ssl.html#certificates
-ssl_certificates = None
```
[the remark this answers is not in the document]
SantiagoTorres: This used to be "use the default bundle that urllib resolves", instead we are going to use the environment variable that requests uses: `REQUESTS_CA_BUNDLE` that's [here](https://github.com/theupdateframework/tuf/pull/781/files#diff-e3c1e2065ceb55ca960d7bc2bd8de44cR220)

tuf/settings.py
```diff
@@ -89,16 +84,6 @@
 # avoid being considered as a slow retrieval attack.
 MIN_AVERAGE_DOWNLOAD_SPEED = 50 #bytes/second
 
-# The time (in seconds) we ignore a server with a slow initial retrieval speed.
-SLOW_START_GRACE_PERIOD = 0.1 #seconds
```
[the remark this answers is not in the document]
SantiagoTorres: It is gone now. I think requests has a built-in start grace period.

tuf/download.py
```diff
@@ -399,122 +380,27 @@ def _download_fixed_amount_of_data(connection, temp_file, required_length):
     raise
 
   else:
-    # This else block returns and skips closing the connection in the finally
-    # block, so close the connection here.
-    connection.close()
+    # This else block returns and skips closing the response in the finally
+    # block, so close the response here.
+    response.close()
```
SantiagoTorres: Not related to this PR, but this `finally` clause with the response.close is then useless. I don't think `finally` works like this at all...

tuf/download.py
SantiagoTorres: We definitely need something defined here, and add it to part of our release process...

tests/test_download.py
```diff
@@ -168,44 +170,29 @@ def test_download_url_to_tempfileobj_and_urls(self):
     self.assertRaises(securesystemslib.exceptions.FormatError,
                       download_file, None, self.target_data_length)
 
-    self.assertRaises(securesystemslib.exceptions.FormatError,
+    self.assertRaises(requests.exceptions.MissingSchema,
```
SantiagoTorres: Again, not related to this PR, but moving to `assertRaises` as a context processor would make this code more readable...

tuf/download.py
awwad: Santiago, that was fixed in another PR and will disappear here once it's rebased and the conflict is resolved (shortly).

awwad:
I'll finish looking through it to make sure I haven't missed anything subtle and then fix the tests.

trishankatdatadog:
@awwad AFAICT with `tox` at least, there were 2 test_slow_retrieval_attack tests failing for sure. There was 1 nondeterministic failure which made 1 map file test error out, which was weird. Thanks for fixing the tests, I'll address the rest of the comments!

tuf/exceptions.py
```diff
@@ -173,7 +173,7 @@ def __init__(self, expected_length, observed_length):
 
   def __str__(self):
     return 'Observed length (' + repr(self.observed_length)+\
-           ') <= expected length (' + repr(self.expected_length) + ').'
+           ') != expected length (' + repr(self.expected_length) + ').'
```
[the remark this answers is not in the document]
trishankatdatadog: The reason why I changed it` from `<=` to `!= is because if there is a bug in the way the download function works, nothing actually prevents it from downloading more than it should. This happened to me, and I found the error message very misleading (e.g., `9 <= 5`).

tuf/settings.py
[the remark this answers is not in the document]
trishankatdatadog: Correct. I figured we don't need two ways to configure the same thing right now.

tuf/settings.py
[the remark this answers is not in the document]
trishankatdatadog: Correct, `requests` has a `connect()` [timeout](http://docs.python-requests.org/en/master/user/advanced/#timeouts).

tuf/settings.py
```diff
@@ -80,7 +75,7 @@
 DEFAULT_TARGETS_REQUIRED_LENGTH = 5000000 #bytes
 
 # Set a timeout value in seconds (float) for non-blocking socket operations.
-SOCKET_TIMEOUT = 4 #seconds
+SOCKET_TIMEOUT = 5 #seconds
```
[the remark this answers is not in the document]
trishankatdatadog: Because network connections can be quite lossy in the real world. We found a customer who was timing out over HTTP, but that was more likely due to the stringest `SLOW_START_GRACE_PERIOD` of 0.1 seconds. This latter setting is gone now, because `requests` handles both `connect()` and `read()` timeouts.

tuf/download.py
trishankatdatadog: Each hostname might be a bit overkill --- I was thinking a separate session per repository mirror, but better to be safe than sorry. Let me work on this.

tuf/download.py
trishankatdatadog: Correct. Sometimes you want to exactly which piece of software was downloading from your server. To be fair, it should be something like `tuf-{tuf-version-number}-over-requests-{requests-version-number}`.

trishankatdatadog:
I have a PR for the sessions issue, but I can't push it now for some network reason, and I gotta run now. See you guys Monday!

setup.py
```diff
@@ -75,12 +75,26 @@
...
+
+
 with open('README.md') as file_object:
   long_description = file_object.read()
 
+
 setup(
   name = 'tuf',
-  version = '0.11.1',
+  version = find_version("tuf", "__init__.py"),
```
awwad: This isn't going to work: read isn't defined. This is code from [here](https://packaging.python.org/guides/single-sourcing-package-version/) and is a little more complicated than is ideal. It'll also match comment lines if they exist. Single-sourcing version number isn't necessary for this pull request, but if I was going to do it, I'd probably add a VERSION file and have `tuf/__init__.py` and `setup.py` each read that in. There could be problems with that, too. I'm going to punt on this and keep the version in two places and we can fix that less urgently. (Also, the user agent reporting a version seems less critical in any case than the rest of the PR.)

awwad:
SlowRetrievalError is essentially part of the API (see [ATTACKS.md](https://github.com/theupdateframework/tuf/blob/develop/tuf/ATTACKS.md)), so I'll adjust this a bit to still raise that when we catch urllib3.exceptions.ReadTimeoutError().

trishankatdatadog:
I am still having some strange network issues in my VM trying to push to / pull from GitHub. Let me send my updated PR as soon as I figure out what's going on.

awwad:
Keep an eye out for any history differences? It wouldn't look like a network issue, but I did have to strip one of your commits (self-merge -- merging remote branch into local branch and then pushing) from the commit history.

Meanwhile, I'm looking at why bad hash errors are occurring in one of the tests where I don't expect them to occur. (Slow retrieval should cause an error that indicates that retrieval was slow/disrupted, not a bad hash.) The download code is a little kinky and probably could use some tidying up.

tests/test_slow_retrieval_attack.py
trishankatdatadog: @awwad, could you help me fix this? I think this might be related to the problems you've been seeing with bad hashes, and so on. Thanks in advance!

trishankatdatadog:
Okay, guys, I've added code to use a different session per hostname. Let me know what you think, and I'm happy to help with anything else!

P.S. There is not so much time pressure on this anymore, but the sooner we can fix this, the better :)

awwad:
So, unlike the code without this PR merged, this branch downloads a target file all in the code that claims to make a connection. This makes the slow retrieval, endless data, etc. code moot (and also invalidates all the docstrings -- e.g. `_download_fixed_amount_of_data` is called on the response provided by requests, slowly reading out data from the request already retrieved from the server, and claims to be "where the download really happens").

Using stream=True in the `requests.get()` call might fix this. I'll tinker more.

trishankatdatadog:
We already use `request.get(stream=True)`. Not sure what is going on to break the tests, I'm guessing some fragile assumptions were made...

awwad:
Fixed some bad assumptions in the tests that resulted -- when the test failed -- in misleading errors. Also raised ReadTimeoutError (from urllib) as an expected tuf.exceptions.SlowRetrievalError.

The test fails, however, because a slow retrieval attack is not prevented. Streaming one byte per second for 400 (or, previously, 800) seconds is permitted to complete. I'll tinker and figure out how to prevent slow retrieval attacks with the requests calls.

trishankatdatadog:
I wonder what's going on. Have you looked at the [doc]](http://docs.python-requests.org/en/master/user/advanced/#timeouts), Sebastien?

awwad:
Yes. The current settings include a chunk size of 40k and a 5s timeout that is now interpreted by requests as a maximum gap between bytes (and also the timeout of the connect interval). This means that a transfer of 40k could take 50+ hours and still pass. We could tinker with those settings`*`, but what we really want is to interrupt the <=40k transfer if it's taking too long. I'm not sure if there's an option like that somewhere in requests.

`*` Barring a total-time-spent timeout for each chunk, I guess we could reduce the expected interval between bytes below one second (🙁), but we could also very substantially reduce the size of chunks and allow the existing means of measuring the rate of transfer to decide to raise SlowRetrievalError. I assume that'll lead to a slowdown. If we can't get a different behavior from requests, then max acceptable delay will be given by inter-byte timeout * chunk size. We can pick a sensible value (1 minute per chunk and a chunk size of 10kb?) and set tuf settings accordingly and add comments by the settings explaining their significance (and similar comments in the rate testing code).

trishankatdatadog:
I think I sort of understand what's going on. Would you mind terribly much showing what you mean in the updated unit tests? Thanks in advance!

awwad:
Not clear on the question, so in case you mean explaining the stack as it is tested, then:

```
updater.download_target() provides fpath len hashes to:
  -> updater._get_target_file() provides verifyfunc, fpath, len, safely=True to
    -> updater._get_file() provides mirrors and len to:
      -> tuf.download.safe_download provides len to:
        -> tuf.download._download_file(), which calls:      # the part we care about now
          -> tuf.download._download_fixed_amount_of_data()  # the part we care about now
        then tuf.download._check_downloaded_length
    then updater._get_file() continues by calling:
      -> verifyfunc (which is passed as verify_target_file, which calls updater._hard_check_file_length() and updater._check_hashes())
```

tuf.settings.SOCKET_TIMEOUT (5s) is used by `tuf.download._download_file()` to initialize a request object with a "timeout" that sets the aforementioned maximum time before the first byte and between further bytes. The request object is passed down to `_download_fixed_amount_of_data()`, which uses request.raw.read() with a chunk size of 40k (checks tuf.settings.CHUNK_SIZE) to read 40k at a time. Requests will time out if the first byte takes 5s to come, or any of the gaps between further bytes exceed 5s. After this first chunk is downloaded, `_download_fixed_amount_of_data()` checks to see if it's done. (It is; the test file provides 400b which is less than the chunk size of 40k, so it's done after one chunk.) 
 The call returns the data after 400+s and the test fails because no slow retrieval error was raised.

If there had been more than one chunk (if target file size > 40k), `_download_fixed_amount_of_data()` would, before continuing with another request.raw.read call for the second chunk, would check the amount of data requests provided and the time that passed during the call to requests, against `tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED`. This check would fail and an error would be raised, after the 400s for the first chunk. (If the file size were, say, 40k, this would have taken ~ 40,000 * delay between each byte).

### Change Required:
The change that's called for (I'll make one or the other in a bit assuming I'm not off my rocker) is to either reduce chunk size and continue to use request.raw.read() or, instead of using request.raw.read(), call request.iter_content() with a much smaller chunk size (I think I prefer this option? Is there a reason raw.read() is preferable?), say 1k, resulting in the largest possible delay that can be caused by an attacker delivering at speeds less than that acceptable (tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED) being something like 83s (1k * tuf.settings.SOCKET_TIMEOUT).

awwad:
I feel like Vlad toyed with requests and elected not to use it? I don't really recall at this point -- maybe we used to use it and moved away.... This (no total timeout option for an individual read) would be a decent reason for that. Maybe there's a way to do this via urllib3? (don't see that yet, but haven't really looked properly).

trishankatdatadog:
The problem with `response.iter_content()` is that it's a _generator_ that uses a fixed-size chunk in every loop. That means you can control downloading _exactly_ the number of bytes you want, and avoid endless data attacks (admittedly small with a small chunk size, but still).

trishankatdatadog:
I strongly recommend using `requests` as a backbone, because custom networking code is prone to problems that have been solved in production-tested libraries such as `requests`.

trishankatdatadog:
Having said all this, the way we used to handle slow retrieval errors appear to be broken using `requests`, and so we need to think of satisfactory solutions. Your idea of using a small chunk size seems feasible, but it does result in a lot of network calls. Let's think a bit more, and see.

awwad:
I agree that we should use something well-established and common, like requests. I also don't really have a feel for the alternatives.

Sorry, I'm not clear on the problem with iter_content yet: you just mean that we'd be using a generator (which I guess we try to avoid)? (Generator use would be fairly transparent, but I guess the stack trace could be wonky?)

The change would be in download.py:_download_fixed_amount_of_data:
* Instead of while True, we'd use for data in request.iter_content(smaller_chunk_or_subchunk_length)
* We might change the math on read amount a little bit

Note: I'm only really recommending it because it's the intended requests interface for this sort of thing and so spares us the use of a lower-level urllib3 object. I don't actually know if it's much more efficient or anything like that.

trishankatdatadog:
Ah, the problem with `iter_content(chunk_size)` is this. Imagine the following for loop. You want to download 6 bytes, and the chunk size is 4 bytes. So you make 2 for loops, which means you read a maximum of 8 bytes. But we wanted exactly 6 bytes, and no more. But the way `iter_content(chunk_size)` works, it's a fixed-size generator you setup once, and you keep iterating over it to get exactly `chunk_size` bytes. Does this make sense?

awwad:
Oh, I see! Yes, very good point. We do promise not to read past the expected length.

awwad:
Okay, so pending other options, we're back to request.raw.read(), but with extremely small chunk size 😞.

trishankatdatadog:
My concern is that using a small chunk size would be quite inefficient in terms of increasing the number of network calls. I wonder whether we should use an adaptive chunk size (choosing something proportional to the file size)? Let me think whether there are better alternatives.

trishankatdatadog:
I have an idea. So this is what `requests` is doing:

```Specifically, it’s the number of seconds that the client will wait between bytes sent from the server. In 99.9% of cases, this is the time before the server sends the first byte.```

Let's say the maximum we're willing to tolerate _in total_ for a whole chunk is `tuf.settings.SOCKET_TIMEOUT`, which just so happens to be 5s. Let's say the chunk size, `tuf.settings.CHUNK_SIZE`, is 4k. So why not tell requests to timeout at `CHUNK_SIZE/SOCKET_TIMEOUT` (something like 0.001220703125s in this case)? 

This may be too stringent for a `connect()` timeout, but we can set that separately.

**Comment by awwad** [comment]
That would mean that the server isn't allowed to have a 0.001220703125s gap between their sends if they don't send the whole chunk of data at once. Is that reasonable? I'm skeptical.
(Yeah, the timeout before the first byte comes can be specified separately and isn't much of an issue.)

**Comment by trishankatdatadog** [comment]
Yes, two concerns: (1) I'm not even sure such a granular read() timeout is
possible, and (2) it may be too stringent.

**Comment by vladimir-v-diaz** [comment]
Requests is vulnerable to a slow retrieval attack, at least when I last checked.  The timeout option on connection and read() attempts still allows an attacker to feed a single byte before the timeout, thus allowing them to prolong a download indefinitely.

What is needed is non-blocking I/O.  The old code enabled non-blocking I/O on the request objects, and kept track of the total elapsed time to detect a slow retrieval attack.  Unfortunately, Requests does not support any kind of non-blocking I/O.

You might want to look into [GRequests](https://github.com/kennethreitz/grequests)(asynchonous requests) to do something similar to the old code.

**Comment by alyptik** [comment]
For what it's worth, GThreads itself recommends instead using [requests/requests-threads](requests/requests-threads) or [ross/requests-futures](ross/requests-futures), and after looking through the code of all three I am of the opinion that [ross/requests-futures](ross/requests-futures) looks the most mature; it also advertises needing "minimal API changes" compared to using GEvent or Twisted which seems like a definite plus.

**Comment by trishankatdatadog** [comment]
@alyptik I like this. Can we account for slow retrieval attacks using one of these async `requests` extensions?

**Comment by SantiagoTorres** [comment]
> Requests is vulnerable to a slow retrieval attack, at least when I last checked. The timeout option on connection and read() attempts still allows an attacker to feed a single byte before the timeout, thus allowing them to prolong a download indefinitely.

Thanks @vladimir-v-diaz , I had a vague recollection about this being the original showstopper for moving to requests.

> @alyptik I like this. Can we account for slow retrieval attacks using one of these async requests extensions?

I'm sorry for the unpopular opinion, but I'm starting to wonder if this timeout detection using the transport layer is an issue in and by itself. I'm remembering the discussion on [tap 10](https://github.com/theupdateframework/taps/blob/master/tap10.md) and the argument seems to be overlapping.

Should we be considering a non-transport solution instead? I know it's probably harder to implement, but it definitely will keep both layers separated.

**Comment by awwad** [comment]
@SantiagoTorres I think you're probably right.

NB, I don't think we've discussed here adequate reason to suppose that something transport level _wouldn't_ serve if the wall-clock timeout feature _were_ present in requests, say. I wouldn't be surprised, though, from perusing the issues in these libraries, to find out that there were further problems, so I think I agree.

How much higher level are we talking, though?
- Would something like having a subprocess that attempts to do the download and can be killed if it doesn't yield results at an expected rate fit what you're talking about? (Still kinda transport level, but would kill a session)
- Would you imagine having a subprocess that attempts to do the download and can be killed if its total time exceeds the total duration expected? (more fully session level, would allow longer DoS attacks for larger files due to lack of visibility into the flow of data)

Or something higher yet?

**Comment by trishankatdatadog** [comment]
Yeah, I'll let you guys duke this out. In my mind, slow retrieval attacks are not a major concern in production environments we control, especially those using TLS. I wouldn't lose any sleep if we don't solve this problem right now. In a world where everyone is being encouraged to move to TLS, I'm not sure how worried we should be about this problem, but feel free to correct me.

**Comment by SantiagoTorres** [comment]
Probably @JustinCappos 's input on this would be helpful?

**Comment by JustinCappos** [comment]
I'd say that an approach that is more along the lines of what
@SantiagoTorres suggested about separation makes the most sense to me.  I'd
love to have another level of detail before really committing though.

**Comment by trishankatdatadog** [comment]
Hi guys, I don't mean to press you, but any resolution on this? We need a fix for this pretty soon for testing...

**Comment by trishankatdatadog** [comment]
As I said earlier, I do not think slow retrieval attacks are the most pressing issue to address in an age where TLS is being pushed everywhere. What is much more important is for TUF to work transparently over HTTPS proxies. As such, I think it's all right to punt this issue right now, and fix it later.

**Comment by awwad** [comment]
If it's consistent with what @SantiagoTorres meant and it makes sense to @JustinCappos, I can make a few edits to basically put the lowest-level download code in download.py into a subprocess that will get killed by the main process if it doesn't return in a reasonable time based on the minimum expected download rate and target file size.

**Comment by trishankatdatadog** [comment]
Sorry, but using subprocesses (takes more system resources) seems more clunky and error-prone, and does not seem to be worth the effort to me. I do not think this is a good idea.

**Comment by SantiagoTorres** [comment]
well, there's two things going on here:

1. The current impl doesn't work with transparent HTTPS proxies
2. The new impl doesn't really protect against slow-retrieval att

One of them is more pressing than another regarding the integration. Could we probably ticketize one and merge/close this then? I don't want to make that call though.

**Comment by awwad** [comment]
Yeah, as to the question of whether or not it's okay for us to release a TUF version that gives up on protecting against slow retrieval attacks, I leave that to @JustinCappos.

In terms of options for handling both slow retrieval and the proxies, in summary:
- 👎❓Grequests might work (gevent + requests), but it's much less well established and the repo itself suggests not using it (and I'm not actually sure yet if it would be less clunky or result in fewer network requests).
- 👎❓I don't see options yet in requests-futures or requests-threads that would provide the timeout we're looking for (so I'm not actually sure they're better than using a dramatically smaller chunk size with requests yet in terms of slow retrieval protection).
- 👎 Windows doesn't support SIGALARM, so I can't just use a simple timeout decorator.
- 👎 Python doesn't have a clean way to kill threads.
- 👎I could do some renovations with the old code, but I would still prefer to use a good library.
I'm not sure how to get out of this with the tools provided without using a subprocess, though that would be clunky, yes. (Error prone, I'm not sure -- maybe? It might also interfere with the readability of the code.)

**Comment by trishankatdatadog** [comment]
@SantiagoTorres @awwad Yeah, the HTTPS proxy issue is the pressing one to solve, but I'm afraid the two issues are a bit linked. I prefer that TUF uses a battle-tested 3rd party library like `requests` in order to solve this issue. If that means giving up protection against slow retrieval attacks (really, most likely due to MitM attack against HTTP-only connections), then I'm willing to live with that, but that's just me.

**Comment by JustinCappos** [comment]
Why would we need to kill threads in the case of a slow retrieval? If we
use a multi-threaded library, can't we handle this with socket timeouts?

**Comment by SantiagoTorres** [comment]
I did some quick digging, the timeout stuff is being handled by `urllib3`, which is in turn used by `requests`. After reading [their docs](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.timeout) I realized that they suggest using a threaded approach using a watchdog (like @awwad suggested).

> If your goal is to cut off any request after a set amount of wall clock time, consider having a second “watcher” thread to cut off a slow request.

Unless we have access to the socket object, which not even urllib3 is happy with sharing, I don't think we can just overwrite the timeouts on it.

**Comment by JustinCappos** [comment]
I understand.  I also think subprocess is very undesirable.

I'd prefer a solution that uses non-blocking sockets with a timeout so that
we can correctly recover.  Do none of the other popular libraries provide
this?

**Comment by JustinCappos** [comment]
Put differently, I know we do not want to open the pandora's box of fixing
our own implementation to handle proxies.  Given that, the options seem to
be:

1) Move to a library to that handles HTTPS and also protects against slow
retrieval
2) Be vulnerable to slow retrieval
3) Don't solve the HTTPS proxy issue

In my mind, option 2 is better than 3.  Option 1 is clearly the best, even
if we have to upstream some slow retrieval protections.  However, this
doesn't seem practical w/ requests.

So, as I see it, we can go for option 2 now, but have a fairly high
priority ticket for moving to option 1.

Thoughts?

**Comment by trishankatdatadog** [comment]
Agreed 💯 

It's unfortunate that we couldn't go to (1) right away, but I believe (2) is a decent compromise for the short-term.

Thanks for your understanding!

**Comment by awwad** [comment]
@JustinCappos Sure, I'll:
- [ ] make some quick changes to testing in this PR (primarily an expected-fail on the slow retrieval test)
- [ ] confirm that other comments are dealt with
- [ ] merge this PR
- [ ] create an issue for the resulting slow retrieval issue and tinker with options there

I gave it another shot with requests (without subprocess), but no joy. I spawned a thread for the requests read(). I had the main thread decide time had passed and try to close the whole connection pool a few ways... but the read always continues. I can't seem to reach any deeper in a reasonable way. I'll try more threading options later.

File: tuf/download.py [hunk 12]
```diff
@@ -257,14 +231,57 @@ def _download_file(url, required_length, STRICT_REQUIRED_LENGTH=True):
...
 
   try:
-    # Open the connection to the remote file.  _open_connection() can raise
-    # socket connection exceptions (such as SSLError).  Connection errors of
-    # this kind can be minimized by adjusting the socket timeout in
-    # settings.py.
-    connection = _open_connection(url)
+    # Use a different requests.Session per hostname to reuse connections while
+    # minimizing subtle security issues.
+    hostname = six.moves.urllib.parse.urlparse(url).hostname
```
awwad: Calls to fetch URLs using different protocols (http vs https) and different ports will share the same session. I don't know if that will result in some subtle security issue.

tuf/download.py
```diff
@@ -257,14 +231,57 @@ def _download_file(url, required_length, STRICT_REQUIRED_LENGTH=True):
...
+    # We will always manually close Responses, so no need for a context
+    # manager.
+    #
+    # Always set the connect + read timeout:
+    # http://docs.python-requests.org/en/master/user/advanced/#timeouts
+    response = session.get(url, stream=True,
+                           timeout=tuf.settings.SOCKET_TIMEOUT)
+
+    # Check response status.
+    response.raise_for_status()
```
awwad: Not style compatible, so fixed:
 - indent size: 4 if continuing line else 2
 - line arrangement per PEP 8

tests/test_download.py
awwad: The code no longer raises MissingSchema, since it now parses URLs to extract the hostname. It instead raises securesystemslib.exceptions.FormatError, so we'd have to test for that instead. That said, it's probably best to use a different error type, since while this is, true enough, a formatting error, FormatError is customarily reserved for the automatic detection based on schemas in formats.py (using `<SCHEMA>.check_match()`)

awwad:
Rebased, signing off on TKK's commits so DCO signoff check stops failing. Removed a now-unnecessary import to make pylint happy.

tuf/download.py
trishankatdatadog: That's a good point. Some cookies are [supposed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Secure_and_HttpOnly_cookies) to be over HTTPS only. Feel free to index sessions by "{scheme}+{hostname}" in there if you wish. That would basically be `netloc` IIRC.

awwad: Resolved that way, with another commit

awwad:
I added a test for SSL hostname verification and untrusted certs.

Assuming all this passes now, I think I'm ready to merge. There have been some changes, so I could use a final eyeball on it in tandem with my own last check. @trishankatdatadog?

(I also added an extra bit of commented-out remote testing, in case someone wants to do some diagnosis, say. No other TUF testing fails on a machine with no net access, so that's why this one is commented out.)

Note that there is not testing to ensure that the behavior using a transparent HTTPS proxy is correct. That might be a little involved. I gather that the new behavior works for you, Trishank?

trishankatdatadog:
@awwad Thanks for all your hard and speedy work!

I'm a bit too busy at the moment, but I'll take a look when some time frees up...

tests/test_download.py
```diff
@@ -168,81 +170,148 @@ def test_download_url_to_tempfileobj_and_urls(self):
...
-    fake_cacert = self.make_temp_data_file()
 
-    with open(fake_cacert, 'wt') as file_object:
-      file_object.write('fake cacert')
 
-    tuf.settings.ssl_certificates = fake_cacert
-    tuf.download._get_opener('https')
-    tuf.settings.ssl_certificates = None
+  # This test uses sites on the internet, requiring a net connection to succeed.
+  # Since this is the only such test in TUF, I'm not going to enable it... but
```
trishankatdatadog: I think we should turn this on, they are very useful tests, and these days it's reasonable to assume tests can reach out to the Internet...

trishankatdatadog:
@awwad I'm sure you're aware of this, but why are the tests failing on the cloud?

trishankatdatadog:
I think it's important to add an explicit unit test to see whether TUF handles HTTPS proxies, which is the whole point of this PR. Let me see how difficult that would be with `mitmproxy`...

trishankatdatadog:
@awwad Sorry, it doesn't look like I have time for this today. Would you mind terribly much helping me with this, please? There is some nice code we might be able to borrow from [here...](https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/test_connections.py)

awwad:
Tinkering. @trishankatdatadog: you said "The tuf.download module now uses the requests library. I have manually tested that it handles HTTPS proxies." When you run the TUF tests with your proxy setup (on this branch), do you get any failures?

Requests is raising slightly different errors for me than the test code expects if something goes wrong when using a proxy, meaning that I have to update some of the tests to allow for additional error classes -- but I thought I'd check first....

trishankatdatadog:
@awwad I haven't tried running tuf tests against the proxy. How would you run this, please?

awwad:
The error profile of TUF may be different through a proxy, and that merits some attention and test construction, too, but I don't think that has to be part of this PR; I can raise an issue for that afterwards, too.

awwad:
In answer to your question on how you'd run the TUF tests through the proxy, though:

Assuming you're running a local proxy, make sure not to exclude localhost / 127.0.0.1 from it. You can then set environment variables, HTTP_PROXY and HTTPS_PROXY -- which requests respects -- before running tests directly, e.g. `cd tests; python test_download.py`. Note that you probably can't just set the env variables and run tox, as tox will set up its own environment. You could edit the tox config.

(If you're not running a local proxy, it wouldn't be easy to arrange for the TUF tests to go through a network's forced proxy, since the tests are all local.)

trishankatdatadog:
Okay, I'll take a look at this when I find some downtime (at an event now). Thanks!

awwad:
K -- no worries either way: it was just an opportunistic ask.

The primary two things you should let me know are 1) if this becomes pressing again or 2) if the PR doesn't provide for what you need.

In the case of it being pressing again, the PR can be merged in its current state, and I can just continue working on testing in another PR. LMK if we get to that point.

trishankatdatadog:
Ah, I see, thanks for asking the question. It would be nice to be able to test this with customers, but I think it is also important to make sure this feature does not regress. So could we please add the tests against a proxy?

awwad:
Yup, working on it.

trishankatdatadog:
Thanks for all your hard work, @awwad !

awwad:
@trishankatdatadog 

So, I have testing (not in this branch) that tries HTTP proxies that do and don't support the HTTP CONNECT verb (which can ferry HTTPS connections, acting basically as a TCP tunnel). The tests I have there work fine. They require more than I'd like to add as dependencies, but I'll work on that.

The test I'm still working on is trying to use a transparent HTTPS proxy that has its own certificate that is trusted by the client. **That's the scenario you're interested in, right?** I think I'm going to try using code from [here](https://github.com/inaz2/proxy2) for this.

trishankatdatadog:
Yes, sir, the second part is the blocker for now.

The first part will be important for other customers with that kind of setup...

awwad:
Fixes #209 
Fixes #204

awwad:
Diagnosis on last build error: Travis-CI ran an environment with Python2.7.3. SSLContext had no create_default_context until Python2.7.9. Adapted the test proxy code slightly based on Python version. For more details, see the commit message on the commit `Test: changed target server cert checking in test proxy script`. Once tests pass, I'll rebase the debug commits out.

tests/test_download.py
```diff
@@ -168,81 +170,148 @@ def test_download_url_to_tempfileobj_and_urls(self):
...
+  #   Use some known URLs on the net to ensure that TUF download checks SSL
+  #   certificates appropriately.
+  #   """
+  #   # We should never get as far as the target file download itself, so the
+  #   # length we pass to safe_download and unsafe_download shouldn't matter.
+  #   irrelevant_length = 10
+  #
+  #   for bad_url in [
+  #       'https://expired.badssl.com/', # expired certificate
+  #       'https://wrong.host.badssl.com/', ]: # hostname verification fail
```
awwad: I added a local test for the wrong host scenario, and I've just now added a local test for the expired cert scenario. That should suffice. It's still nice to have the online tests, but I'd still rather keep them off and leave them as diagnostic tools as long as they aren't much more effective; net connection issues causing test failures aren't worth it yet.

tests/test_download.py
awwad: I added a local test for the wrong host scenario, and I've just now added a local test for the expired cert scenario. That should suffice. It's still nice to have the online tests, but I'd still rather keep them off and leave them as diagnostic tools as long as they aren't much more effective; net connection issues causing test failures aren't worth it yet.

tests/aggregate_tests.py
```diff
@@ -45,11 +45,29 @@
...
+# such listed constraints don't match the python version currently running.
+VERSION_SPECIFIC_TESTS = {
+    'test_proxy_use': {'major': 2, 'minor': 7}} # Run test only if Python2.7
+
 # Remove '.py' from each filename to allow loadTestsFromNames() (called below)
 # to properly load the file as a module.
 tests_without_extension = []
 for test in tests_list:
   test = test[:-3]
+  if test in VERSION_SPECIFIC_TESTS:
```
aaaaalbert: Since we're dealing with special cases here, this logic must be made more robust. Particularly, the specification check should **fail harshly** if the desired spec makes no sense, e.g. has a `minor` but no `major`, or neither (see test cases below). So, roughly:
```python
testdict = VERSION_SPECIFIC_TESTS[test]
major, minor = testdict.get("major"), testdict.get("minor")
if not major:
  ERROROUT("A version-specific test that requires no major version makes no sense at all!")
  # XXX Do not continue here! Must fix!
if major != sys.version_info.major:
  NOTIFY("Can't run this test on that Python major version")
  continue
if minor != sys.version_info.minor:
  NOTIFY("Can't run this test on that Python minor version")
  continue
# Otherwise, we're good to go!
```

aaaaalbert: The promised test cases are all generated and tested with this snippet. Anytime it ends up in `Clause 3`, take a close look at the result. For example, `None,None (2, 7) Clause 3` means that an entry in the `VERSION_SPECIFIC_TESTS` dict that did not specify `major` and `minor` at all (!!!) ended up getting run on (a simulated) Python 2.7.

```python
import sys
print("Actual Python version is", sys.version_info)
del sys

# Mock `sys.version_info` so we can use the production code (mostly) unchanged
class sys():
  class version_info():
    major = None
    minor = None

VERSION_SPECIFIC_TESTS = {}
sys_versions = []

# Initialize version specific test dict and system versions list
for major in [None, 2, 3]:
  for minor in [None, 5, 6, 7]:
    name = str(major) + "," + str(minor)
    VERSION_SPECIFIC_TESTS[name] = {}
    if major:
      VERSION_SPECIFIC_TESTS[name]["major"] = major
    if minor:
      VERSION_SPECIFIC_TESTS[name]["minor"] = minor
    if major and minor:
      sys_versions.append((major, minor))


def run_test(test, version):
  sys.version_info.major, sys.version_info.minor = version
  for junk in [1]:
    # Start excerpt: https://github.com/theupdateframework/tuf/pull/781/files#diff-6ec6c904b1b852303ff24f29e2058ad8R64
    if test in VERSION_SPECIFIC_TESTS:
      if 'major' in VERSION_SPECIFIC_TESTS[test] \
          and sys.version_info.major != VERSION_SPECIFIC_TESTS[test]['major']:
        print(test, version, "Clause 1")
        continue
      elif 'minor' in VERSION_SPECIFIC_TESTS[test] \
          and sys.version_info.minor != VERSION_SPECIFIC_TESTS[test]['minor']:
        print(test, version, "Clause 2")
        continue
      print(test, version, "Clause 3")
      # End excerpt: https://github.com/theupdateframework/tuf/pull/781/files#diff-6ec6c904b1b852303ff24f29e2058ad8R71



print("Version specific tests:", VERSION_SPECIFIC_TESTS)
print()
print("test   OSver   clause")

for version in sys_versions:
  for test in VERSION_SPECIFIC_TESTS:
    run_test(test, version)
```

tests/aggregate_tests.py
```diff
@@ -45,11 +45,29 @@
...
+# If minor is included, limit version to that listed here. Skip the test if any
+# such listed constraints don't match the python version currently running.
+VERSION_SPECIFIC_TESTS = {
+    'test_proxy_use': {'major': 2, 'minor': 7}} # Run test only if Python2.7
+
 # Remove '.py' from each filename to allow loadTestsFromNames() (called below)
 # to properly load the file as a module.
 tests_without_extension = []
 for test in tests_list:
   test = test[:-3]
```
aaaaalbert: What about `os.path.splitext` instead?

```
>>> os.path.splitext("Willkommen.alter.Freund")
('Willkommen.alter', '.Freund')
```

tests/test_download.py
aaaaalbert: (You could make turning on the tests easier though --- wrap them in `"""` instead of per-line `# ` prefixing.)

tests/test_download.py
```diff
@@ -168,81 +172,188 @@ def test_download_url_to_tempfileobj_and_urls(self):
...
+    # bad:     An otherwise-valid cert with the wrong hostname. The good certs
+    #          list "localhost", but this lists "notmyhostname".
+    # expired: An otherwise-valid cert but which is expired (no valid dates
+    #          exist, fwiw: startdate > enddate).
+    good_cert_fname = os.path.join('ssl_certs', 'ssl_cert.crt')
+    good2_cert_fname = os.path.join('ssl_certs', 'ssl_cert_2.crt')
+    bad_cert_fname = os.path.join('ssl_certs', 'ssl_cert_wronghost.crt')
+    expired_cert_fname = os.path.join('ssl_certs', 'ssl_cert_expired.crt')
+
+    # Launch three https servers (serve files in the current dir).
```
aaaaalbert: `s/three/four/`, and please explain why there are two good ones.

tests/test_download.py
```diff
@@ -168,81 +172,188 @@ def test_download_url_to_tempfileobj_and_urls(self):
...
+        download.unsafe_download(expired_https_url, target_data_length)
+
+      # Try connecting to the server processes with the good certs while
+      # trusting the appropriate good certs. Expect success.
+      # Note: running these OK downloads at the top of this try section causes
+      #       a failure in a previous assertion: retrieving the same good URL
+      #       again after no longer "trusting" the good certfile still succeeds
+      #       if we had previously succeeded in retrieving that same URL while
+      #       still trusting the good cert. Perhaps it's a caching issue....?
+      #       I'm not especially concerned yet, but take note for later....
```
aaaaalbert: It's probably worth checking if `requests` does what you expect it to (i.e. reload the CA bundle from the environment variable). A [quick glance suggests](https://github.com/requests/requests/blob/ff0c325014f817095de35013d385e137b111d6e8/requests/sessions.py#L698) that a new `Session` must be created for a reload to happen. (Are you sure that the explanations for the expected failures above are well-grounded? Are the bad etc. certs ever seen by `requests`?)

tests/test_proxy_use.py
```diff
@@ -0,0 +1,366 @@
...
+    #   server certs to accept in its HTTPS connection to the target server.
+    #   This is only relevant if the proxy is in intercept mode.
+    cls.https_proxy_port = cls.http_port + 3
+    cls.https_proxy_proc = subprocess.Popen(
+        ['python', 'proxy_server.py', str(cls.https_proxy_port), 'intercept',
+        os.path.join('ssl_certs', 'ssl_cert.crt')],
+        stderr=subprocess.PIPE)
+    # Note that the https proxy server's address uses https://, regardless of
+    # the type of connection used with the target server.
+    cls.https_proxy_addr = 'https://127.0.0.1:' + str(cls.https_proxy_port)
```
aaaaalbert: This uses "127.0.0.1", i.e. one of the IPv4 addresses of `localhost`. Why was IPv4 chosen here over the symbolic name? Will the code work with IPv6 too?

tests/test_proxy_use.py
```diff
@@ -0,0 +1,366 @@
...
+    # and its url on the server.
+    current_dir = os.getcwd()
+    target_filepath = self.make_temp_data_file(directory=current_dir)
+    rel_target_filepath = os.path.basename(target_filepath)
+
+    with open(target_filepath, 'r') as target_file_object:
+      self.target_data_length = len(target_file_object.read())
+
+    self.url = \
+        'http://localhost:' + str(self.http_port) + '/' + rel_target_filepath
```
aaaaalbert: Now `localhost`? ;-)

tests/test_proxy_use.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+
+  def test_baseline_no_proxy(self):
+    """
+    Test a length-validating TUF download of a file through a proxy. Use an
+    HTTP proxy, and perform an HTTP connection with the final server.
+    """
+
+    logger.info('Trying http download with no proxy: ' + self.url)
```
aaaaalbert: If there isn't a striking reason for not capitalizing HTTP, I would suggest you do throughout.

tests/test_proxy_use.py
```diff
@@ -0,0 +1,366 @@
...
+
+    logger.info('Trying http download via http proxy: ' + self.url)
+    download.safe_download(self.url, self.target_data_length)
+    download.unsafe_download(self.url, self.target_data_length)
+
+
+
+
+
+  def test_httpS_dl_via_smart_http_proxy(self):
```
aaaaalbert: (I think I get the idea of capitalization ~ highlighting, but this looks terrible to me.)

tests/test_proxy_use.py
```diff
@@ -0,0 +1,366 @@
...
+    with new saved values.
+    """
+
+    if key in self.old_env_values:
+      # Do not save the current value. We already saved an older value, and
+      # the original one is the one we'll restore to, not whatever we most
+      # recently overwrote it with.
+      pass
+
+    elif key in os.environ:
```
aaaaalbert: Perhaps remove the `elif`/`else` and do `self.old_env_values[key] = os.environ.get(key, None)`.

tests/test_proxy_use.py
```diff
@@ -0,0 +1,366 @@
...
+        'variable does not appear in the list of values to restore. ' \
+        'Please make sure to use _set_env_value().'
+
+    if self.old_env_values[key] is None:
+      # If it was not previously set, try to unset it.
+      # If the platform provides a way to unset environment variables,
+      # del os.environ[key] should unset the variable. Otherwise, we'll just
+      # have to settle for setting it to an empty string.
+      # See os.environ in:
+      #    https://docs.python.org/2/library/os.html#process-parameters)
```
aaaaalbert: Remove trailing `)`

tests/test_proxy_use.py
aaaaalbert: (BTW, what about `os.execve`ing things and supplying a modified environment directly, instead of saving/restoring the current one?)

tests/aggregate_tests.py
```diff
@@ -45,11 +45,29 @@
...
+  if test in VERSION_SPECIFIC_TESTS:
+    if 'major' in VERSION_SPECIFIC_TESTS[test] \
+        and sys.version_info.major != VERSION_SPECIFIC_TESTS[test]['major']:
```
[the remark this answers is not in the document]
awwad: Good point. We no longer support 2.6, though.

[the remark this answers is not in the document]
aaaaalbert: Fine by me, just something I noticed :-)

tests/test_proxy_use.py
awwad: The SSL certificates specify localhost (including one that's inconvenient to replace). It might be better to replace the 127.0.0.1 in the proxy server addresses with locahost, though.

tests/test_download.py
awwad: Thanks for the catch. There's an explanation 13 lines up, [here]( https://github.com/theupdateframework/tuf/pull/781/files/b163caa29b37f2794201cbec43be2f6697ade745#diff-e3c1e2065ceb55ca960d7bc2bd8de44cR240). I'll adjust this, though, for clarity.

tests/test_download.py
awwad: Good tip. Has to be `'''`, since the docstring inside uses """.

tests/test_proxy_use.py
awwad: Your suggestion reduced 6 lines to 2. Thanks!
As for execve: seems like a messy option. I'm not sure I'd want to fully replace the environment variables (instead of tweaking them) and I'm not sure I want to close the existing process and transition to a new one (temp files?). Popen execve? Not sure. I think this is probably easier and more acceptable.

tuf/exceptions.py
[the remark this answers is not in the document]
awwad: Behavior here has not changed. It should not be possible to get a length greater than expected due to the way the download code was written (and now as well, I think). If the code is changed and it becomes possible to end up with more, we can change the error message, but '<' currently communicates an expectation that has some value.

tests/test_proxy_use.py
awwad: I don't think the included test proxy will work in IPv6 as configured either way, no.

trishankatdatadog:
Hi Sebastien,

Sorry, don't mean to press you, but could we please get a dev release at least ASAP? We have the Agent 6.6 release frozen at the end of next week, and I'd like to test this ASAP, so that we can update TUF in the Agent next week if all goes well. Thanks very much in advance!

Cc: @nmuesch @olivielpeau

awwad:
No problem. I'll merge sometime today.

tests/test_proxy_use.py
awwad: The proxy tests run an IPv4 proxy and I think will fail with IPv6. It's probably true that the symbolic name should be used. I don't recall why I made this 127.0.0.1, but switching to localhost (both server and client side) seems to break things for some reason.

trishankatdatadog:
@nmuesch Can we get your customer to install the latest stable Agent, and then update TUF manually to see if that fixes the problem?

tests/test_download.py
awwad: Thanks! I think I saw a bit of behavior supporting this notion, but there was enough varied testing that I don't think the tests weren't succeeding at what they were doing. I'm curious, but this was easily solved by purging the sessions when the certificate expectations changed, so I've removed the possibility for this issue and removed a related comment now that it's gone.

awwad:
Merged and an alpha released produced: https://github.com/theupdateframework/tuf/releases/tag/v0.11.2-alpha

awwad:
@trishankatdatadog

trishankatdatadog:
@awwad Could we please get a `tuf-0.11.2.dev1` release on PyPI? Otherwise, it's hard for us to beta test this. The dev release ensures users cannot install it by accident (need to pass --pre flag to pip). Thanks!

awwad:
https://pypi.org/project/tuf/0.11.2.dev1/
There's some confusion in the way we have dev version numbers listed, so mind the version.

trishankatdatadog:
Thanks @awwad !

PR #800, opened 2018-11-06
Refactored updater to support different types of targets and ways of downloading data

renatav:
Our use case has two main differences compared to the current implementation of the updater:

All metadata files are stored in a git repository. This only means that we want to change the way of downloading the metadata files. A user of our system would have a local clone of this git repository and updating metadata would mean fetching and merging the newest changes. Also, we do not need to have previous and current directories as we can just see the content of a file at a particular revision. Additionally, we do not need to read data from physical files, we use git show.

In our case targets are not files, but commits in git repositories, seeing that our goal is to protect these repositories. So, updating targets means updating git repositories. Validation needs to ensure that history of a repository was not modified by an attacker. We need to verify that the commits are in accordance with the metadata. We do not need to check hashes or lengths of individual files.

Other than the mentioned changes, we want to closely follow the TUF specification.

What this refactoring focuses on is separating validation of metadata (and operations like traversing target roles) from loading and downloading metadata files and targets. Once the metadata is loaded/updated and is stored in memory, the same checks can be executed, regardless of how the previous actions were carried out. In order to provide the needed flexibility, all code for loading and downloading metadata was moved to a separate class. Just like there is an implementation of that class which loads metadata from the file system and downloads these files from a server, there can be another implementation where git repositories are used. These classes are referred to as metadata handlers. When instantiating the updater, one only needs to specify which class of the metadata handler should be used. Similarly, everything related to downloading targets and validating the downloaded was moved to another class. This class is referred to as the targets handler. We could have another class which would handle updating git repositories.

So, this change would enable redefining certain aspects of the update process, while still relying on TUF to handle important security checks. Git metadata/targets handlers are empty, as we only want to show how this would work. We do not want to suggest merging these classes into the official TUF repository. We believe that this refactoring would generally be useful and that other TUF users could modify the update process to fit their needs.

This refactoring would not change of the updater's features. The updater's tests should all still be passing. The only difference is that, due to some of the methods being moved to the handler, the tests need to call the metadata/target handler's methods.


renatav:
Sorry, wanted to create this inside our fork. Please ignore. Had the wrong  tab open :(

Issue #660, opened 2018-03-23
Roles and Delegations are still confused in parts of the implementation


awwad:
Several issues are related to this, including at least: #589, #646, #658, [spec issue 19](https://github.com/theupdateframework/specification/issues/19).

### Summary
The reference implementation continues to try to provide a 1-to-1 mapping of roles to keyids-the-role-should-be-signed-by-in-order-to-be-valid. This is not correct: the same role may need to be validated expecting different sets of keys, based on how the role was reached in the depth-first search while looking for target information.

The bottom line here is that **code should be rewritten in places like roledb, to avoid things like get_role_keyids(), or roleinfo['threshold'] or roleinfo['keyids'], etc.**

### Detailed explanation
I think this all comes from a deeper issue:

Per the design of TUF, roles and delegations should never be regarded as the same thing. While in the last few years, there have emerged a few additional reasons that this separation is important (TAPs 3's multi-role delegations and TAP 5's alternative repository roots, for example), my understanding of the design history is that there was not originally an expectation of a one-to-one mapping of roles to delegations pointing to them.

So let's be pedantic:
- A role is not the same thing as a delegation: a delegation is a relationship between two roles.
- A delegation is a property of the delegating role, not the delegated-to role.
- Therefore, for delegated roles, key IDs and thresholds are properties of a delegation, not properties of the delegated role.
- You cannot simply map delegated roles to the keyids that validate them: you must map keys to delegations (that is, delegator + delegatee, if we're assuming no parallel edges in the graph (which, btw, **are we?**)).

The graph of delegations used to be guaranteed -- in the reference implementation -- to be a simple directed graph where Root is a node with an indegree of 0 and all other nodes have an indegree of 1. I.e.: No roles delegate to Root. Root delegates to Targets. Targets can delegate to various roles, but two or more roles never delegate to the same role. The design intention, however, did not guarantee maximum indegree of 1: in the design, delegations could be promiscuous: a single role can be delegated to by multiple roles.

A result of this confusion is that roledb still represents roles as flat in certain ways, when it should not assume this. This flatness resulted in #589, #658, probably #646, and presumably other forgotten issues.

Take the function roledb.get_role_keyids() as an example: `get_role_keyids` takes only the name of a role (and a repository name), purporting to return the keys that that role should be signed by. The trouble, though, is that different roles delegating to that role can expect different keys and validate or reject role C's target information based on those different expectations.

### Specification / Avoiding Future Issues
[...]

PR #811, opened 2018-12-07
Avoid unnecessary delegated role downloads

awwad:
Using deprecated function [`Updater.targets_of_role()`](https://github.com/theupdateframework/tuf/blob/d910c0865a83fda735b4cdd595c711e69c837974/tuf/client/updater.py#L2486) to fetch the targets listed by the top-level Targets role previously led to the unneeded download of all delegated targets roles.  This PR fixes that specifically for calls to `targets_of_role('targets')` (which is the only well-defined and secure use of the deprecated function `targets_of_role`).

All delegated targets continue to be downloaded if `targets_of_role` is called on a delegated targets role.  This provides something like what we think users who might do this expect (It's not clear there are any people doing this.), refreshing all delegated targets roles to avoid use of outdated delegation information, and essentially hoping that the most-recently-validated delegation to the role of interest is what the user would expect that role to have been validated by.....

`targets_of_role` will be removed or replaced, so this is a fix for the sole reasonable use case until then.

The PR also adds some explanatory comments and warnings to the code.

### DIFF
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -2349,6 +2349,12 @@ class Updater(object):
   def all_targets(self):
     """
     <Purpose>
+
+      NOTE: This function is deprecated.  Its behavior with regard to which
+      delegating Targets roles are trusted to determine how to validate a
+      delegated Targets role is NOT WELL DEFINED.  Please transition to use of
+      get_one_valid_targetinfo()!
+
       Get a list of the target information for all the trusted targets on the
       repository.  This list also includes all the targets of delegated roles.
       Targets of the list returned are ordered according the trusted order of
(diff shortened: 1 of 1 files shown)


awwad:
I'm planning on rewriting the function to provide target info from a given role only from what is verified and in memory (rather than trying to refresh and verify the metadata).  If there is no verified metadata for the given role, the function would raise an error indicating that.  Basically, it becomes a convenience function that extracts the info from what is already known and verified about the role.  Use for role 'targets' would then just be:
```
refresh()
targets_of_role('targets')
```

As for delegated roles, if the user wants to skip the delegation tree and jump straight to a role (instead of using the proper method to walk down the delegation tree starting at the 'targets' role), it would be the user's job to get the metadata verified through the appropriate path.

(The broader problem of [the reference implementation sometimes treating roles and delegations as interchangeable](https://github.com/theupdateframework/tuf/issues/660) needs to be addressed.)

trishankatdatadog:
Yes, the key, role, and delegation management in the reference implementation is a bit of a mess, I'm afraid. I had advocated using a more object-oriented approach a long time ago, which I think would have led to a cleaner, more intuitive implementation.

awwad:
We can do it with or without OO, and I think the role-vs-delegation trap was there either way.  Based on lab principles, we'll try to keep OO to a minimum.

PR #812, opened 2018-12-07
Split download timeout into connection and inter-byte timeouts

awwad:
Requests accepts two kinds of timeouts: connection timeouts and
inter-byte ("read", "gap") timeouts.  Separate these so that we
can use sensible values for each.

See http://docs.python-requests.org/en/master/user/advanced/#timeouts


joshuagl:
Thank you for the submission Sebastien. This somewhat breaks the existing interface and is unlikely to be propagated into our imminent refactor efforts, therefore we're closing this PR.

PR #815, opened 2018-12-21
Making the process of updating metadata a little bit more flexible

renatav:

`client/updater.py` was slightly refactored in order to make the process of updating metadata files a bit more flexible. This flexibility is only related to how the files are downloaded and does not impact validation. Finding mirrors and the actual download were moved to a so-called handler class.
With that change, it is possible to implement the majority of the updater which would use a git repository to store metadata files. An example of that implementation can be seen [here](https://github.com/openlawlibrary/taf/blob/master/oll/updater/handlers.py).

requirements**:

### DIFF
--- a/tuf/client/handlers.py
+++ b/tuf/client/handlers.py
@@ -1,0 +1,158 @@
+import tuf
+import logging
+import tuf.exceptions
+
+logger = logging.getLogger('tuf.client.updater')
+
+
+class MetadataUpdater(object):
+  """
+  <Purpose>
+    Provide a way to redefine certain parts of the process of updating metadata.
+    To be more specific, this class should enable redefinition of how metadata
+    is downloaded.
+
+
+  <Arguments>
+    mirrors:
+      A dictionary holding repository mirror information, conformant to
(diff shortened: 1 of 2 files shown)


tuf/developer_tool.py
```diff
@@ -391,6 +391,7 @@ def status(self):
         self._log_status(self.project_name, signable, self.repository_name)
 
       except securesystemslib.exceptions.Error as e:
+        # pylint: disable=unsubscriptable-object
```
lukpueh: OT: This seems to be the reason why the current [TUF@develop build is failing](https://travis-ci.org/theupdateframework/tuf). IMO it should be fixed in a separate PR. Such a fix should probably also evaluate, whether pylint has a point here.

awwad: https://github.com/theupdateframework/tuf/pull/839 resolves this issue.  (Sorry, folks.)

joshuagl:
Thank you for the contribution. Due to imminent refactor efforts this code is unlikely to merge, but we would very much welcome your input and engagement on our refactor plans.

PR #819, opened 2019-01-08
Reduce test spam by using unittest output buffering

awwad:
After merge, stdout should only appear if a test has failed.  This functionality is provided by `unittest.main` argument `buffer=True`.  This functions like the `--buffer` command line argument listed [here](https://docs.python.org/2/library/unittest.html#command-line-options).  std out is discarded if a test succeeds.


awwad:
Closing in favor of #825.

PR #825, opened 2019-02-19
Reduce test spam by using unittest output buffering

awwad:
Thanks go to @lukpueh for this helpful tip.

This is a succinct replacement for PR #819.

After merge, stdout from the code being tested should only appear if the test has failed or errored out.

This functionality is provided by `unittest.TextTestRunner` argument `buffer=True`.
This functions like the `--buffer` command line argument listed [here](https://docs.python.org/3/library/unittest.html#command-line-options)

### DIFF
--- a/tests/aggregate_tests.py
+++ b/tests/aggregate_tests.py
@@ -91,7 +91,8 @@ random.shuffle(test_modules_to_run)
 
 if __name__ == '__main__':
   suite = unittest.TestLoader().loadTestsFromNames(test_modules_to_run)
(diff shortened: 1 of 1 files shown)


awwad:
I looked over it, and I ran it again and it passed.  Usually, the cause of something like the error there is an inadequate delay to let the http/s server or (in this case) proxy start up.   In this case, though, the delay of 4s seems like plenty.  I've also seen rare port collisions, but that's already corrected for here (unless multiple test modules are running in parallel for some reason, perhaps).  The proxy may not have been ready, or some other issue may have occurred with the proxy server startup.  I'm going to tolerate low rates of test failures if they occur due to a test proxy connection issue.

Issue #171, opened 2014-02-08
Documentation: private keys must be exposed?


nyarly:
Per https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt a key is (effectively)

```
 { "keytype" : "rsa",
        "keyval" : { "public" : PUBLIC,
                     "private" : PRIVATE }
      }
```

The KEYID is its SHA-256 digest.

A client must verify the calculation of a KEYID.

How can a client do that without the private key of the role in question?

PR #836, opened 2019-03-15
Destroy unnecessary MetaFile classes in formats.py

awwad:
This PR is related to two efforts:
- The consolidation of metadata definitions and destruction of conflicting or redundant internal metadata representations, related to work being done for #660
- [Uptane PR #171 related to timeserver key rotation](https://github.com/uptane/uptane/pull/171)

The primary effect is to replace six classes and their associated methods and some related external functions with one, simpler, general purpose function in formats.py, `build_dict_conforming_to_schema`.  These functions are involved in the production of metadata.

This PR also:
- corrects a mistake in TIMESTAMP_SCHEMA, the exclusion of spec_version (since writing timestamp metadata previously did not properly respect this schema, written metadata still included spec_version)
- **Adds a notion of the version of the TUF specification that this reference implementation intends and is expected to support -- tuf.SPECIFICATION_VERSION**

### Details
Kills classes MetaFile, RootFile, TargetsFile, MirrorsFile, SnapshotFile, and TimestampFile.  They each had an unused from_ method and a used make_ method.  They were all additional, unnecessary representations of the same metadata, and it is very important that metadata formats be defined once in the reference implementation, in the schemas that are already used more broadly, in formats.py.

Replaces the classes, their methods, some related external functions, and some associated variables with a single short function called build_dict_conforming_to_schema that takes keyword arguments and builds a dictionary, then checks to make sure that the result conforms to the given schema.

This commit shifts repository_lib from use of the old classes to the new function.

In later commits, we should use this function more broadly, since it can be of use in all schema construction.

There are several TODOs added to the code, mostly for post-#660 tasks.


awwad:
Rebased after #844

awwad:
See sister PR and review comments [here}(https://github.com/awwad/tuf/pull/23) in Uptane's TUF fork.

awwad:
I think I've addressed your comments on the sister (upTUF) PR, @lukpueh.  I've also expanded this PR just a little bit.  Should be a quick re-review, I hope. :)

tuf/formats.py
```diff
@@ -474,227 +467,102 @@ def make_signable(role_schema):
...
+      # Currently, the list is limited to ['_type', 'spec_version'].
 
+      if key == '_type' and isinstance(element_type, SCHEMA.String):
+        # A SCHEMA.String stores its expected value in _string, so use that.
+        dictionary[key] = element_type._string #pylint: disable=protected-access
 
+      elif (key == 'spec_version' and
+          element_type == SPECIFICATION_VERSION_SCHEMA):
+        # If not provided, use the specification version in tuf/__init__.py
+        dictionary[key] = tuf.SPECIFICATION_VERSION
```
lukpueh: I know, I [suggested this](https://github.com/awwad/tuf/pull/23#discussion_r269027063), but now that I look at it I think it's pretty specific, at least for the function name `build_dict_conforming_to_schema`. Do you think there is a better name? If not, disregard this comment.

tuf/repository_lib.py
```diff
@@ -1442,10 +1456,32 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
+        version=version,
+        expires=expiration_date,
+        targets=filedict,
+        delegations=delegations)
+  else:
+    return tuf.formats.build_dict_conforming_to_schema(
+        tuf.formats.TARGETS_SCHEMA,
+        version=version,
+        expires=expiration_date,
+        targets=filedict)
```
lukpueh: So I couldn't sell the [`kwargs` alternative to you](https://github.com/awwad/tuf/pull/23#discussion_r269024916)? :)

tuf/formats.py
```diff
@@ -356,6 +348,7 @@
 TIMESTAMP_SCHEMA = SCHEMA.Object(
   object_name = 'TIMESTAMP_SCHEMA',
   _type = SCHEMA.String('timestamp'),
+  spec_version = SPECIFICATION_VERSION_SCHEMA,
```
lukpueh: Does this require new tests?

tuf/repository_lib.py
awwad: Whatcha mean?  The extras bit (e.g. delegations) in the calls below this one?

awwad: The function uses kwargs and returns deepcopy(kwargs) (after some modifications).

tuf/formats.py
```diff
@@ -474,227 +467,102 @@ def make_signable(role_schema):
...
+  # This includes:
+  #   _type:        <securesystemslib.schema.String object>
+  #   spec_version: SPECIFICATION_VERSION_SCHEMA
+  #
+  # (Please note that _required is slightly misleading, as it includes both
+  #  required and optional elements. It should probably be called _components.)
+  #
+  for schema_element in schema._required: #pylint: disable=protected-access
+    key = schema_element[0]
+    element_type = schema_element[1]
```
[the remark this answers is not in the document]
awwad: 1. Re. the additional typing requirements: I forgot to edit the type check after also requiring `_required`.  You're right: I should switch to a type check now.

2. I'll add a comment noting that this is intended for Object-type schemas (which are for the most part the ones for which it's of interest).

3. On the 2-tuples: quite right, of course -- dunno why I didn't unpack that way.  Fixing.

tuf/formats.py
awwad: This in itself, no.  The spec version was simply missing in the schema for Timestamp.  It was already being written in metadata despite the fact that it was missing in the schema, because this schema was not really used to build the metadata before.  Now the schema is used to build metadata.

I want to add more validation that uses these schemas in updater, and to add some additional testing for formats.py related to the schemas more generally, but that's not really related to this change, and it'll have to wait.

tuf/formats.py
awwad: The function is still intended to be a general purpose (Object-type) schema builder, so I would rather remove this convenience functionality than change that scope.  There may be a better name, but I don't want to narrow its purpose to Role metadata construction, if that's what you mean.

tuf/repository_lib.py
awwad: Gonna merge and use this in some other work, now that the rest has been addressed, but if I missed something, please LMK.  I'll keep the extras-style optional arguments construction in mind if there are other places calls like this start to need to happen.

awwad:
Thanks for the review, @lukpueh!

tuf/formats.py
```diff
@@ -474,227 +467,106 @@ def make_signable(role_schema):
...
-    # Raise securesystemslib.exceptions.FormatError if not.
-    tuf.formats.TARGETS_SCHEMA.check_match(targets_metadata)
+  """
 
-    version = targets_metadata['version']
-    expires = targets_metadata['expires']
-    filedict = targets_metadata.get('targets')
-    delegations = targets_metadata.get('delegations')
+  # Check the schema argument type (must provide check_match and _required).
+  if not isinstance(schema, SCHEMA.Schema):
```
lukpueh: s/Schema/Object/

tuf/formats.py
```diff
@@ -474,227 +467,106 @@ def make_signable(role_schema):
...
+    raise ValueError(
+        'The first argument must be a schema.Schema object, but is not. '
```
lukpueh: s/Schema/Object/

tuf/formats.py
lukpueh: I think it's fine as it is.

tuf/repository_lib.py
lukpueh: Oh yes, I was referring to passing something like an `extra_kwargs` that may or may not contain the `delegations` (sorry for being unclear).

PR #843, opened 2019-03-28
Fix try-except-else in test_updater.py

awwad:
There were some silly holes in the try-except clauses in some tests expecting NoWorkingMirrorError, and some other errors.  The cases would have passed if no errors were raised....  This patches those holes in logic.

Merge #842 first.  That included two of these.....


awwad:
Rebased after merge of #842.

Issue #808, opened 2018-11-27
Documentation Overhaul


awwad:
User documentation has to be reorganized and updated.

The main `README.md` links to `Getting_Started.rst`, which among other things links to:
- Quickstart
- CLI
- CLI usage examples
- Tutorial

The problems:
- Each of these is actually a tutorial.  They cover CLI client, CLI repo, and advanced repo (repository_tool).
- The CLI client is not covered very well.
- Tutorial.py is actually a detailed repository creation tutorial, so it should be renamed something like 'Advanced Repository Tutorial'
- Advanced client is not covered here (`client/updater.py`).  There are some links and references in Tutorial.py about this, but that's not where they belong since that's really a repository tutorial, and they are wrong / outdated anyway.  An unlinked-to tuf/client/README.md provides some instructions related to updater.py, but they need to be reviewed and linked to from Getting Started.
- Tutorial.py's client section is bad / out of date:
  - It talks about a repository example created in tuf/README.md, but that file has no real contents, so we **need to delete it**.
  - It links to a document (`tuf/client_setup_and_repository_example.md`) that talks about using `basic_client.py`, which no longer exists.  That document should also be deleted.  Any valuable copy should be moved to tuf/client/README.md.
  - Client Setup and Test TUF Locally sections should be merged and clarified.  ("To test the repository we've just created, we need to connect to it...")

My current opinion is that we should move to having three tutorials instead of the many docs above:
- QUICKSTART, talking about the CLI client and repo, with these additions:
  - a note that these quickstart instructions will use the basic command-line interface, and that much more flexibility is available in "these advanced tutorials for repository_tool.py and updater.py", with links (to the docs that are currently `TUTORIAL.md` and `tuf/client/README.md`).
  - a bit more about the CLI client than it currently has
- Advanced Repository Tutorial -- largely, the contents of TUTORIAL.md
- Advanced Client Tutorial -- what's currently in tuf/client/README.md, updated and expanded a bit

We should also probably move those three docs to docs directory.

Miscellaneous other corrections:
- CLI client instructions should point out that it is necessary to run `repo.py --init` first, since that's actually what sets up the client directory....
- There's at least one missing cd-to-client-directory instruction in the CLI client instructions.

awwad:
Note that Lukas has suggested a [doctest-style validation of the tutorials](https://github.com/theupdateframework/tuf/pull/775#pullrequestreview-172429283), in response to a PR in which I added some regression tests for TUTORIAL.md.

[...]

PR #849, opened 2019-04-04
Reframe CLI, revise quickstart, and reorganize tutorials

awwad:
This PR:
- correctly frames the CLI's current state as a tutorial toy.
- provides a friendlier quickstart that puts what it's doing into perspective and guides you to next steps.
- provides a better sense of what each tutorial/quickstart doc is for.
- makes the getting started page slightly more friendly.
- combines the two CLI docs, `CLI.md` and `CLI_EXAMPLES.md`
- clarifies some installation instructions
- other minor tweaks to the quickstart and CLI docs

The larger task this PR is part of is Issue #808.


awwad:
Some more tasks:
- [ ] `CLI.md` and `CLI_EXAMPLES.md` are really doing the same thing and should probably be merged into a single file: `CLI_EXAMPLES.md`.
- [ ] adjust the mention of PyNaCl in TUTORIAL.md to refer to the replacement library

docs/QUICKSTART.md
```diff
@@ -1,21 +1,44 @@
...
-The CLI requires a few dependencies and C extensions that can be installed with
-`pip install securesystemslib[crypto,pynacl]`.
+In this quickstart tutorial, we'll use the basic TUF command-line interface
+(CLI), which includes the `repo.py` script and the `client.py` script, to set
+up a repository with an update and metadata about that update, then download
+and verify that update as a client.
+
+Unlike the underlying TUF modules that the CLI uses, the CLI itself is a bit
+bare-bones.  Using the CLI is the easiest way to familiarize yourself with
+how TUF works, however.  It will serve as a very basic update system and use
```
lukpueh: ...? I think this is missing the end of the sentence.

docs/QUICKSTART.md
```diff
@@ -1,21 +1,44 @@
...
 
 ----
-The following is a basic workflow in four steps:
 
-**Step (1)** - Initialize a repo.  The `tufrepo`, `tufkeystore`, and
-`tufclient` directories are created in the current working directory.
+**Step (0)** - Make sure TUF is installed
+
+See the [installation instructions for TUF](docs/INSTALLATION.rst).
```
lukpueh: We are in the same directory `INSTALLATION.rst`. Remove `docs` form relative path or use absolute path.

lukpueh: Also, would it be bad to just provide the snippet here? So that quickstarters don't have to go through INSTALLATION.rst?
```
pip install tuf securesystemslib[crypto,pynacl]
```

docs/QUICKSTART.md
```diff
@@ -38,21 +61,38 @@ tufrepo/
...
+# or, if you are using Python2:
 $ python -m SimpleHTTPServer 8001
 
-or with Python 3...
-$ python3 -m http.server 8001
 ```
 
-**Step (4)** - Fetch a target file from the repo.  The client downloads
-any required metadata and the requested target file.
+**Step (4)** - Obtain and verify the `testfile` update on a client.
```
lukpueh, line 90: minuscule nit-pick: You end some heading sentences with a period and other not.

docs/GETTING_STARTED.rst
```diff
@@ -1,9 +1,8 @@
...
 ---------------
 
+- `Overview of TUF <OVERVIEW.rst>`_
 - `Installation <INSTALLATION.rst>`_
-- `Contributors <CONTRIBUTORS.rst>`_
-- `Quickstart <QUICKSTART.md>`_
-- `CLI <CLI.md>`_
-- `CLI Usage Examples <CLI_EXAMPLES.md>`_
-- `Tutorial <TUTORIAL.md>`_
+- Beginner Tutorials (using the basic command-line interface): `Quickstart <QUICKSTART.md>`_; `CLI Tutorial <CLI.md>`_; `CLI Further Examples <CLI_EXAMPLES.md>`_
```
lukpueh: The long line looks a little daunting now. I would leave QUICKSTART.md separate and see if the cli tutorials could be combined, or restructured? From a quick glance, CLI.md looks like a documentation of the interface, whereas CLI_EXAMPLES.md looks like a slightly more in-depth version of QUICKSTART.md.

docs/GETTING_STARTED.rst
```diff
@@ -1,9 +1,8 @@
...
+- `Advanced Tutorial <TUTORIAL.md>`_
```
lukpueh, line 9: If you call it advanced tutorial here, which I think is a good idea, you maybe also want to call it that in the document heading?

docs/QUICKSTART.md
awwad: They do need to go through the installation instructions to install TUF.  Prior text was a bit misleading in that it suggested that they only needed to run the ssl install.  In any case, that step is listed two lines down, to emphasize it, since it would otherwise be optional.

docs/QUICKSTART.md
awwad: Thanks. ^_^

docs/QUICKSTART.md
lukpueh: But isn't the line I'm suggesting all that's required for the quick start?

docs/QUICKSTART.md
awwad: Fixed

docs/GETTING_STARTED.rst
awwad: Fixed, and I've now combined the two CLI documents.

docs/GETTING_STARTED.rst
awwad: Fixed there and elsewhere, thanks.

docs/QUICKSTART.md
awwad: No, you need TUF installed.

awwad: Hopefully the newer text is less likely to lead to that conclusion.

awwad: 😞 I kept reading your suggested line here as the same as the existing line two lines below, and as the last line in INSTALLATION.md  -- I missed that sneaky "tuf" inclusion.  Sorry. 😄

I also managed to miss that the "external dependencies" section I was assuming that the [installation instructions](INSTALLATION.md) still included is now only in [`CONTRIBUTORS.rst`](https://github.com/theupdateframework/tuf/blob/e82ce352b42cd7012571657fe4a1a70d63368751/docs/CONTRIBUTORS.rst#external-dependencies).  That's where lines like `apt-get install build-essential libssl-dev libffi-dev python-dev` live....  I'm not sure why that makes sense.  I need to try installation out without all those on a vm.

awwad: Simplified.  Installing pip produces everything you need already on a basic Ubuntu platform, so I didn't delve into python-dev or libssl-dev, etc.  Just provided the single command in the QUICKSTART.

trishankatdatadog:
@awwad It's because `cryptography` manylinux1 wheel bundles its own copy of OpenSSL, unless you use something glib-c incompatible, like Alpine

docs/QUICKSTART.md
SantiagoTorres: >  python-dev or libssl-dev

To contextualize, you need this to build the native support for the openssl backend. I think it's better to keep this from becoming a transient dependency as that may change  at some point in the future...

awwad: I'm not clear on your suggestion, sorry:  are you saying that we should avoid _implicitly_ requiring that `libssl-dev` or similar packages be installed, or that we should avoid advising it explicitly since it may not be necessary?  (Or something else?)

SantiagoTorres: oh, the opposite. I understood that you were advising to remove these deps because something else is pulling them in. I'd rather have them explicitly just in case this is not the default in the future.

awwad: Currently, those requirements are now only in the collaborator docs, not in the installation docs.  I was wondering if I should move them back into the installation docs.  You're suggesting that we include them, which is good to hear, so I'll pull them back in to the natural place to put them (the installation instructions).  I'll frame them as as-necessary installations, I guess, to try to keep things simple.

That's beyond the scope of this PR, so I'll just do that in the next one, but for this PR, I'll split the difference in `QUICKSTART.md` in anticipation of that change....

PR #854, opened 2019-04-12
Version Management

mnm678:
that is provided, the better.

**Fixes issue #**:
Reference implementation for Version Management TAP (https://github.com/theupdateframework/taps/pull/107/files)

This pull request adds a comparison of minor spec_versions (in format major.minor.fix) during an update and reports an out of date client to that client. Minor versions will not have breaking changes, so the update may continue.

requirements**:

### DIFF
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -128,7 +128,7 @@ import time
 import fnmatch
 import copy
 import warnings
-import distutils.version
+import pkg_resources
 
 import tuf
 import tuf.download
@@ -1505,7 +1505,7 @@ class Updater(object):
                 'metadata lists version number: ' + str(metadata_spec_version))
 
           #report to user if minor versions do not match, continue with update
-          if distutils.version.StrictVersion(metadata_spec_version) != distutils.version.StrictVersion(tuf.SPECIFICATION_VERSION):
+          if pkg_resources.parse_version(metadata_spec_version) != pkg_resources.parse_version(tuf.SPECIFICATION_VERSION):
             logger.info("Downloaded metadata that specifies a higher minor " +
                 "spec_version. This code has version " + str(tuf.SPECIFICATION_VERSION) +
                 "and the metadata lists version number " + str(metadata_spec_version) +
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -128,7 +128,6 @@ import time
 import fnmatch
 import copy
 import warnings
-import pkg_resources
 
 import tuf
 import tuf.download
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -128,6 +128,7 @@ import time
 import fnmatch
 import copy
 import warnings
+import distutils.version
 
 import tuf
 import tuf.download
@@ -1496,20 +1497,6 @@ class Updater(object):
           metadata_spec_major_version = int(metadata_spec_version.split('.')[0])
           code_spec_major_version = int(tuf.SPECIFICATION_VERSION.split('.')[0])
           
-          metadata_spec_minor_version = 0
-          metadata_spec_fix_version = 0
-          code_spec_minor_version = 0
-          code_spec_fix_version = 0
-
-          if(len(metadata_spec_version.split('.')) > 1):
-            metadata_spec_minor_version = int(metadata_spec_version.split('.')[1])
-          if(len(metadata_spec_version.split('.')) > 2):
-            metadata_spec_fix_version = int(metadata_spec_version.split('.')[2])
-          if(len(tuf.SPECIFICATION_VERSION.split('.')) > 1):
-            code_spec_minor_version = int(tuf.SPECIFICATION_VERSION.split('.')[1])
-          if(len(tuf.SPECIFICATION_VERSION.split('.')) > 2):
-            code_spec_fix_version = int(tuf.SPECIFICATION_VERSION.split('.')[2])
-
           if metadata_spec_major_version != code_spec_major_version:
             raise tuf.exceptions.UnsupportedSpecificationError(
                 'Downloaded metadata that specifies an unsupported '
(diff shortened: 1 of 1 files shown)


lukpueh (commented):
It might be worth taking a look at version parsing libraries? [`distutils.version`](https://docs.python.org/3/distutils/apiref.html#module-distutils.version) is in the standard library and should do the trick.

Or the third-party [`packaging.version`](https://packaging.pypa.io/en/latest/version/#version-handling), if [PEP-440](https://www.python.org/dev/peps/pep-0440/) compliance is required.

trishankatdatadog:
@lukpueh https://github.com/DataDog/integrations-core/blob/23ebd01428da80b85ad59d5c7e23be57a1970cee/datadog_checks_downloader/datadog_checks/downloader/download.py#L298-L299

awwad:
What's the goal here?  Existing code on the develop branch already does this comparison.  This adds some temp variables for minor and fix, only if they exist, and then checks them (whether or not they existed).

mnm678:
I updated it to use parse_version which eliminates the temp variables and makes it a bit cleaner.

The goal is to report to the user if the minor versions do not match. The update can still continue if the major versions are the same, but this allows users to know when their client is out of date. This is a small change to make the reference implementation consistent with a proposed TAP (https://github.com/theupdateframework/taps/pull/107/files)

tuf/client/updater.py
```diff
@@ -1495,14 +1496,21 @@ def _get_metadata_file(self, metadata_role, remote_filename,
           metadata_spec_version = metadata_signable['signed']['spec_version']
           metadata_spec_major_version = int(metadata_spec_version.split('.')[0])
           code_spec_major_version = int(tuf.SPECIFICATION_VERSION.split('.')[0])
-
+
```
lukpueh: Please remove trailing whitespace

lukpueh: OT: We might want to activate a corresponding lint policy (see [in-toto's pylintrc](https://github.com/in-toto/in-toto/blob/develop/pylintrc#L59)).

tuf/client/updater.py
```diff
@@ -1495,14 +1496,21 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
                 repr(code_spec_major_version) + '; however, the obtained '
                 'metadata lists version number: ' + str(metadata_spec_version))
 
+          #report to user if minor versions do not match, continue with update
+          if pkg_resources.parse_version(metadata_spec_version) != pkg_resources.parse_version(tuf.SPECIFICATION_VERSION):
+            logger.info("Downloaded metadata that specifies a higher minor " +
+                "spec_version. This code has version " + str(tuf.SPECIFICATION_VERSION) +
+                "and the metadata lists version number " + str(metadata_spec_version) +
+                ". The update will continue as the major versions match.")
+
```
lukpueh, line 1519: Please try to [wrap lines at character 79](https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length).

lukpueh: OT: This is actually something we should have the linter check as well.

tuf/client/updater.py
```diff
@@ -1495,14 +1496,21 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+          
           if metadata_spec_major_version != code_spec_major_version:
             raise tuf.exceptions.UnsupportedSpecificationError(
                 'Downloaded metadata that specifies an unsupported '
                 'spec_version.  This code supports major version number: ' +
                 repr(code_spec_major_version) + '; however, the obtained '
                 'metadata lists version number: ' + str(metadata_spec_version))
 
+          #report to user if minor versions do not match, continue with update
+          if pkg_resources.parse_version(metadata_spec_version) != pkg_resources.parse_version(tuf.SPECIFICATION_VERSION):
```
lukpueh: If you compare the full versions you don't need to parse them. But if I understand correctly, you don't want to compare the full version, just the minor version.  The major version you already compare above and patch/fix version you don't care about.

Also, your log message says that one version (part) is higher than the other, so you will need to compare with `>` operator. 

I suggest you parse both `metadata_spec_version` and `tuf.SPECIFICATION_VERSION` only once and then use the major parts for the first (existing) check and the minor parts for the second (your new) check.

lukpueh (changes requested):
Thanks for the patch. See my comments inline.

Please also consider reading our [dev guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/dev-workflow.md), especially the parts about uniform descriptive commit messages. They are very helpful for reviewers. :)

Thanks!

tuf/client/updater.py
```diff
@@ -1493,16 +1493,30 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+          code_spec_minor_version = int(code_spec_version_split[1])
+ 
           if metadata_spec_major_version != code_spec_major_version:
             raise tuf.exceptions.UnsupportedSpecificationError(
                 'Downloaded metadata that specifies an unsupported '
                 'spec_version.  This code supports major version number: ' +
                 repr(code_spec_major_version) + '; however, the obtained '
                 'metadata lists version number: ' + str(metadata_spec_version))
 
+          #report to user if minor versions do not match, continue with update
```
lukpueh, line 1511: The comment suggests `!=` but the code and the log message `>`. Which one is it?

tuf/client/updater.py
```diff
@@ -1493,16 +1493,30 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+          if metadata_spec_minor_version > code_spec_minor_version:
+            logger.info("Downloaded metadata that specifies a higher minor " +
+                "spec_version. This code has version " + 
+                str(tuf.SPECIFICATION_VERSION) + 
+                "and the metadata lists version number " +
```
lukpueh: I think you are missing a whitespace before `and`.

lukpueh (commented):
Thanks or addressing my comments. I have two more minor concerns. Also you have trailing whitespace in  three lines, consider taking a look at https://github.com/secure-systems-lab/code-style-guidelines/issues/5, which describes how to detect them.

tuf/client/updater.py
mnm678: I updated the operator and message to report any mismatch in the minor version.

lukpueh (approved):
LGTM! Please run the necessary git commands (e.g. as [suggested by probot](https://github.com/theupdateframework/tuf/pull/854/checks?check_run_id=144156013)) to "sign them off" retroactively.

PR #856, opened 2019-04-16
WIP: minor: fix bug in test_developer_tool

awwad:
The PR corrects a hole in one test's logic in `test_developer_tool`.  It's another instance of a test not failing if an expected error is not raised, due to flawed try/except/else construction.

### DIFF
--- a/tests/test_developer_tool.py
+++ b/tests/test_developer_tool.py
@@ -173,6 +173,9 @@ class TestProject(unittest.TestCase):
     except (OSError, tuf.exceptions.RepositoryError):
       pass
 
+    else:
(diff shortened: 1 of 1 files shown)


lukpueh:
The PR does indeed correct a hole in the developer tool unit test, with the consequence that the test now fails. We need to find out if the full test is still valid and if so fix the  developer tool.

jku:
Documentation:

> The 'metadata_directory' and 'targets_directory'  directories are created if they do not exist.

This is true: former is created with os.makedirs(), latter with os.mkdir() so the test as it's written should not expect failure.

jku:
I'm closing this, the issue may be valid but this won't be merged with an unexpected test failure

Issue #857, opened 2019-04-17
NoWorkingMirrorError swallows error messages


therve:
**Description of issue**:

When tuf raises a NoWorkingMirrorError, it displays the list of URLs with the corresponding exception classes, but the error message of each exception is lost.

**Current behavior**:

I got the following error in some logs:

```
tuf.exceptions.NoWorkingMirrorError: No working mirror was found:
  'xxx': ReplayedMetadataError()
```

You can reproduce the issue easily with the following script:

```
import tuf.exceptions

error = tuf.exceptions.ReplayedMetadataError('myrole', '1.2.3', '1.2.1')
print(str(error))
mirror_error = tuf.exceptions.NoWorkingMirrorError({'myurl': error})
print(str(mirror_error))
````

Which outputs:

```
Downloaded 'myrole' is older ('1.2.3') than the version currently installed ('1.2.1').
No working mirror was found:
  '': ReplayedMetadataError()
```

**Expected behavior**:

I want to get the actual content of ReplayedMetadataError.

**Comment by adityasaky** [comment]
Hello!

This seems like a very straightforward fix:

https://github.com/theupdateframework/tuf/blob/develop/tuf/exceptions.py#L247

`repr(mirror_error) -> str(mirror_error)`

Which gives me:

```>>> import tuf.exceptions
>>> error = tuf.exceptions.ReplayedMetadataError('myrole', '1.2.3', '1.2.1')
>>> print(str(error))
Downloaded 'myrole' is older ('1.2.3') than the version currently installed ('1.2.1').
>>> mirror_error = tuf.exceptions.NoWorkingMirrorError({'myurl': error})
>>> print(str(mirror_error))
No working mirror was found:
  '': Downloaded 'myrole' is older ('1.2.3') than the version currently installed ('1.2.1').
```

trishankatdatadog:
Ah, @therve is one of us at Datadog, thanks for helping out :)

adityasaky:
No problem!

awwad:
That actually won't fix the missing information in the logs, in part because `updater`'s loggers don't call `str()` on `NoWorkingMirrorError`, so changing `NoWorkingMirrorError.__str__()` won't fix it.  Example logging [here](https://github.com/theupdateframework/tuf/blob/b005002fef3d6fa679ed77878b248b7d65876083/tuf/client/updater.py#L1558-L1559).

[...]

PR #858, opened 2019-04-17
Display sub error messages for NoWorkingMirrorError

adityasaky:
that is provided, the better.

**Fixes issue #**: #857 

Use `str()` instead of `repr()` when generating error messages for `NoWorkingMirrorError`.


awwad:
Won't suffice, see my comment on #857.

PR #859, opened 2019-04-19
Fix #857: allow logging NoWorkingMirrorError to include details

### LINKED ISSUE #857: NoWorkingMirrorError swallows error messages

awwad:
Fixes issue #857, allowing logging of `NoWorkingMirrorError` to include details for contained errors that use calculated error messages (like `ReplayedMetadataError`).

The PR does this by adding `__repr__()` functions to the contained errors.

While it's at it, this PR additionally:
- adds `__repr__()` funcs to some other exceptions that have `__str__()` funcs
- fixes some minor style issues
- for reviewers, adds commented-out direct-reproduction code
  to the `__repr__()` funcs (probably will be cut before merging)

### DIFF
--- a/tuf/exceptions.py
+++ b/tuf/exceptions.py
@@ -60,8 +60,14 @@ class InvalidMetadataJSONError(FormatError):
     self.exception = exception
 
   def __str__(self):
+    return repr(self)
+
+  def __repr__(self):
     # Show the original exception.
-    return repr(self.exception)
+    return self.__class__.__name__ + ' : wraps error: ' + repr(self.exception)
+
+    # # Directly instance-reproducing:
+    # return self.__class__.__name__ + '(' + repr(self.exception) + ')'
 
 
 class UnsupportedAlgorithmError(Error):
@@ -69,7 +75,7 @@ class UnsupportedAlgorithmError(Error):
 
 
 class BadHashError(Error):
-  """Indicate an error while checking the value a hash object."""
+  """Indicate an error while checking the value of a hash object."""
 
   def __init__(self, expected_hash, observed_hash):
     super(BadHashError, self).__init__()
@@ -78,8 +84,19 @@ class BadHashError(Error):
     self.observed_hash = observed_hash
 
   def __str__(self):
-    return 'Observed hash (' + repr(self.observed_hash)+\
-           ') != expected hash (' + repr(self.expected_hash)+')'
+    return (
+        'Observed hash (' + repr(self.observed_hash) + ') != expected hash (' +
+        repr(self.expected_hash) + ')')
+
+  def __repr__(self):
+    return self.__class__.__name__ + ' : ' + str(self)
+
+    # # Directly instance-reproducing:
+    # return (
+    #     self.__class__.__name__ + '(' + repr(self.expected_hash) + ', ' +
+    #     repr(self.observed_hash) + ')')
+
+
 
 
 class BadVersionNumberError(Error):
@@ -122,9 +139,19 @@ class ReplayedMetadataError(RepositoryError):
 
 
   def __str__(self):
-    return 'Downloaded ' + repr(self.metadata_role)+' is older ('+\
-           repr(self.previous_version) + ') than the version currently '+\
-           'installed (' + repr(self.current_version) + ').'
+    return (
+        'Downloaded ' + repr(self.metadata_role) + ' is older (' +
+        repr(self.previous_version) + ') than the version currently '
+        'installed (' + repr(self.current_version) + ').')
+
+  def __repr__(self):
+    return self.__class__.__name__ + ' : ' + str(self)
+
+    # # Directly instance-reproducing:
+    # return (
+    #     self.__class__.__name__ + '(' + repr(self.metadata_role) + ', ' +
+    #     repr(self.previous_version) + ', ' + repr(self.current_version) + ')')
+
 
 
 class CryptoError(Error):
@@ -140,7 +167,14 @@ class BadSignatureError(CryptoError):
     self.metadata_role_name = metadata_role_name
 
   def __str__(self):
-    return repr(self.metadata_role_name) + ' metadata has bad signature.'
+    return repr(self.metadata_role_name) + ' metadata has a bad signature.'
+
+  def __repr__(self):
+    return self.__class__.__name__ + ' : ' + str(self)
+
+    # # Directly instance-reproducing:
+    # return (
+    #     self.__class__.__name__ + '(' + repr(self.metadata_role_name) + ')')
 
 
 class UnknownMethodError(CryptoError):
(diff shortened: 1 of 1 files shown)


awwad:
👍
A `tuf.log` excerpt from running the tests on this branch shows `ReplayedMetadataError` details in a `NoWorkingMirrorError` log line.

```
[2019-04-19 17:39:56,710 UTC] [tuf.client.updater] [ERROR] [_get_metadata_file:1559@updater.py]
Failed to update 'timestamp.json' from all mirrors: {'http://localhost:33283/tmpj3hx5_qv/TestReplayAttack__u7ayn0g/repository/metadata/timestamp.json': ReplayedMetadataError : Downloaded 'timestamp' is older (1) than the version currently installed (2).}
```

adityasaky (approved):

tuf/exceptions.py
```diff
@@ -60,16 +60,22 @@ def __init__(self, exception):
...
   def __str__(self):
+    return repr(self)
+
+  def __repr__(self):
     # Show the original exception.
-    return repr(self.exception)
+    return self.__class__.__name__ + ' : wraps error: ' + repr(self.exception)
+
+    # # Directly instance-reproducing:
+    # return self.__class__.__name__ + '(' + repr(self.exception) + ')'
```
lukpueh, line 70: IIUC you could also remove custom `__init__`, `__str__` and `__repr__` here. Just inheriting from `Exception` (or a child class thereof) should do the trick, e.g.:

```python
>>> class MyInvalidMetadataJSONError(Exception):
...     pass
...
>>> e = MyInvalidMetadataJSONError(Exception("the real error"))
>>> repr(e)
"MyInvalidMetadataJSONError(Exception('the real error'))"
>>> str(e)
'the real error'
>>> raise e
  File "<input>", line 1, in <module>
    raise e
MyInvalidMetadataJSONError: the real error
```

tuf/exceptions.py
```diff
@@ -78,8 +84,19 @@ def __init__(self, expected_hash, observed_hash):
...
+        'Observed hash (' + repr(self.observed_hash) + ') != expected hash (' +
+        repr(self.expected_hash) + ')')
+
+  def __repr__(self):
+    return self.__class__.__name__ + ' : ' + str(self)
+
+    # # Directly instance-reproducing:
+    # return (
+    #     self.__class__.__name__ + '(' + repr(self.expected_hash) + ', ' +
+    #     repr(self.observed_hash) + ')')
```
lukpueh, line 97: I think the "directly instance-reproducing" `__repr__` is preferable. See above example and the [corresponding docs](https://docs.python.org/3.7/reference/datamodel.html#object.__repr__). (Same goes for the rest of the PR)

tuf/exceptions.py
```diff
@@ -165,8 +199,18 @@ def __init__(self, expected_length, observed_length):
     self.observed_length = observed_length #bytes
 
   def __str__(self):
-    return 'Observed length (' + repr(self.observed_length) + \
-        ') < expected length (' + repr(self.expected_length) + ').'
+    return (
+        'Observed length (' + repr(self.observed_length) +
+        ') < expected length (' + repr(self.expected_length) + ').')
```
lukpueh, line 204: OT: Would it be worth mentioning somewhere (e.g. in the docstring) that `observed_length` is expected to be less than `expected_length`?

tuf/exceptions.py
```diff
@@ -214,6 +266,14 @@ def __init__(self, message, signable):
...
     return self.exception_message
 
+  def __repr__(self):
+    return self.__class__.__name__ + ' : ' + str(self)
+
+    # # Directly instance-reproducing:
+    # return (
+    #     self.__class__.__name__ + '(' + repr(self.exception_message) + ', ' +
+    #     repr(self.signable) + ')')
+
```
lukpueh, line 276: If `signable` is no longer relevant, this could also be shortned to:
```python
class UnsignedMetadataError(Error):
   """..."""
```
Since the custom implementation does exactly what the default would do.

tuf/exceptions.py
```diff
@@ -248,6 +308,14 @@ def __str__(self):
```
lukpueh, line 308: OT: Is it really necessary to `urlparse` this? The unparsed string is used as fallback in case of a (blanket!) exception anyway. 

Also note that there are scenarios, where `urlparse` does not raise an exception but the used `netloc` field is still empty (see e.g. https://github.com/theupdateframework/tuf/issues/857#issue-434256120).

tuf/exceptions.py
lukpueh: ... the default IMHO also looks better than the custom, e.g.:

```python
>>> raise tuf.exceptions.InvalidMetadataJSONError(Exception("the real error"))
  File "<input>", line 1, in <module>
    raise tuf.exceptions.InvalidMetadataJSONError(Exception("the real error"))
tuf.exceptions.InvalidMetadataJSONError: InvalidMetadataJSONError : wraps error: Exception('the real error')
```

lukpueh (commented):
Looks good overall. I left a couple of suggestions, nothing blocking though.

Also, here's a snippet to `__repr__` and `__str__` the exceptions:

```python
import tuf.exceptions as e
errors = [
  e.InvalidMetadataJSONError(Exception("bad")),
  e.BadHashError("dead", "beef"),
  e.ReplayedMetadataError("ROOT", "v1", "v2"),
  e.BadSignatureError("ROOT"),
  e.DownloadLengthMismatchError(2048, 1024),
  e.SlowRetrievalError("1024"),
  e.UnsignedMetadataError("bad sigs", "SIGNABLE"),
  e.NoWorkingMirrorError({"https://pypi.org/": Exception("bad mirror")}),
]

for error in errors:
  repr(error)
  str(error)
```

awwad:
Didn't catch the recent review before merging via cmd line.  (I guess the page I looked at hadn't refreshed in a few hours?)  I'll add a PR to address the comments.

tuf/exceptions.py
awwad: It's a lot harder to read in the logs.

PR #878, opened 2019-05-25
check if absolute path when add target into repo

lixuefeng2:
Signed-off-by: lixuefeng (Cloud) <li.xuefeng@h3c.com>

repo.py --add doesn't support absolute path.  I add check for it in codes.
Actually, relative path with ..(upper folder) has also problem. 
The document should add comment to forbid upper folder.

### DIFF
--- a/tuf/scripts/repo.py
+++ b/tuf/scripts/repo.py
@@ -701,7 +701,10 @@ def add_target_to_repo(parsed_arguments, target_path, repo_targets_path,
 
   if not os.path.exists(target_path):
     logger.debug(repr(target_path) + ' does not exist.  Skipping.')
-
+    
+  elif os.path.isabs(target_path):
+    logger.warn('omit this file:' + target_path + ', the target '
+        'file should be in ' + parsed_arguments.path)  
   else:
     securesystemslib.util.ensure_parent_dir(
         os.path.join(repo_targets_path, target_path))


lixuefeng2:
@Sebastien Awwad, It seems the CI has problems? The codes should be all right.

tuf/scripts/repo.py
```diff
@@ -701,7 +701,10 @@ def add_target_to_repo(parsed_arguments, target_path, repo_targets_path,
 
   if not os.path.exists(target_path):
     logger.debug(repr(target_path) + ' does not exist.  Skipping.')
-
+    
+  elif os.path.isabs(target_path):
+    logger.warn('omit this file:' + target_path + ', the target '
```
lukpueh, line 706: `logger.warn` is deprecated, use `logger.warning` instead (see linter message)

lukpueh (commented):
Thanks for pointing out this issue and for providing a patch, @lixuefeng2! 

I think the main problem here is that `repo.py` wants to preserve the full relative path of each target file (`target_path`)  when copying it to the tuf repo directory (`repo_targets_path`):

https://github.com/theupdateframework/tuf/blob/2c9e3dfacaeb9c2092c81d74aec95905f81db296/tuf/scripts/repo.py#L709-L711

As you pointed out, this breaks if `target_path` is an absolute path (see `os.path.join`) or has up-level references (see `securesystemslib.util.ensure_parent_dir`).

I suggest to either,
(1) first normalize the path, and only consider it (i.e. copy to repo, add to metadata) if it exists and is not absolute, i.e.
`foo/../bar` --> `foo/bar` (if exists) --> `<REPO_DIR>/targets/foo/bar`, or

(2) allow any file (or directory) that exists, but don't preserve the full path when copying, i.e. 
  `/foo/bar` --> `<REPO_DIR>/targets/bar`, and
  `../baz` --> `<REPO_DIR>/targets/baz`

I think the second approach is less surprising to the user. What do others think? @awwad? @lixuefeng2?

Btw. it might be worth to revise the other file/path related operations in `repo.py` for consistency, and also update the argparse help messages.

lukpueh:
Closing due to inactivity. Re-visit with #811.

PR #879, opened 2019-05-25
add comment for step2

lixuefeng2:


docs/QUICKSTART.md
```diff
@@ -41,9 +41,10 @@ to the client in the `tufclient` directory.
 
 We'll create a target file that will later be delivered as an update to clients.
 Metadata about that file will be created and signed, and added to the
-repository's metadata.
+repository's metadata.
```
lukpueh: Please remove trailing whitespace.

docs/QUICKSTART.md
```diff
@@ -41,9 +41,10 @@ to the client in the `tufclient` directory.
 
 We'll create a target file that will later be delivered as an update to clients.
 Metadata about that file will be created and signed, and added to the
-repository's metadata.
+repository's metadata. 
 
 ```Bash
+# be in the same directory as step 1
```
lukpueh, line 47: If we add this here, we should add a similar first line to the snippets below for consistency. What about `# Continued from above`?

lukpueh (changes requested):
Thanks for the PR, @lixuefeng2! I appreciate the suggestion, but I wonder if it is really necessary. To me the tutorial reads as one train of thought, where it is implicit that the snippets are based on each other. I vote for not merging. @awwad, what do you think?

If we do, please consider my comments below.

docs/QUICKSTART.md
lixuefeng2: @lukpueh done.
I tried to test TUF in my project. I encountered this problem(I changed to data directory and I thought TUF shall remember the repo folder). It may be a trap for beginners like me:). So in my opinion it's necessary. You are expert of it, but not everyone is expert. People may think it's hard to use or it has bugs..

lukpueh:
Closing due to inactivity and incompleteness. https://github.com/theupdateframework/tuf/issues/808#issuecomment-487596129, has a similar request:
> - Let's make all snippets # Continuing from the previous section . . ..

PR #880, opened 2019-05-25
just for format

lixuefeng2:


lukpueh (changes requested):
Thanks, @lixuefeng2. Please remove d66a9b3e557943002413c15dde854eafdb59781e from this PR. You have also submitted it as part of #879.

lukpueh:
Merged locally (after removing stray commit). Closing here.

PR #885, opened 2019-06-04
Do not immediately verify latest root when rotating

trishankatdatadog:
**Fixes issue #**:

N/A

When rotating the root (say from v2 to v3, then v3 to v4), the reference implementation currently immediately _verifies_ signatures on the latest root (say, v4). This will fail when v4 is signed using v3's but not v2's keys. This PR changes the implementation so that signatures on the latest root is _not_ immediately verified.

requirements**:

### DIFF
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1128,7 +1128,7 @@ class Updater(object):
     # Retrieve the latest, remote root.json *WITHOUT* verifying it just yet.
     # We will verify it soon if all goes well.
     latest_root_metadata_file = self._get_file(
-        'root.json', lambda f: f, 'meta', DEFAULT_ROOT_UPPERLENGTH,
+        'root.json', lambda f: True, 'meta', DEFAULT_ROOT_UPPERLENGTH,
         download_safely=False)
 
     latest_root_metadata = securesystemslib.util.load_json_string(
@@ -1149,6 +1149,16 @@ class Updater(object):
       # _update_metadata().
       self.consistent_snapshot = True
       self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH, version=version)
+      # Ensure that the role and key information of the top-level roles is the
+      # latest.  We do this whether or not Root needed to be updated, in order
+      # to ensure that, e.g., the entries in roledb for top-level roles are
+      # populated with expected keyid info so that roles can be validated.  In
+      # certain circumstances, top-level metadata might be missing because it
+      # was marked obsolete and deleted after a failed attempt, and thus we
+      # should refresh them here as a protective measure.  See Issue #736.
+      self._rebuild_key_and_role_db()
+      self.consistent_snapshot = \
+          self.metadata['current']['root']['consistent_snapshot']
 
 
 
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1125,9 +1125,11 @@ class Updater(object):
       None.
     """
 
-    # Retrieve the latest, remote root.json.
-    latest_root_metadata_file = self._get_metadata_file(
-        'root', 'root.json', DEFAULT_ROOT_UPPERLENGTH, None)
+    # Retrieve the latest, remote root.json *WITHOUT* verifying it just yet.
+    # We will verify it soon if all goes well.
+    latest_root_metadata_file = self._get_file(
+        'root.json', lambda f: f, 'meta', DEFAULT_ROOT_UPPERLENGTH,
+        download_safely=False)
 
     latest_root_metadata = securesystemslib.util.load_json_string(
         latest_root_metadata_file.read().decode('utf-8'))
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1126,24 +1126,28 @@ class Updater(object):
       None.
     """
 
-    # Set the next version of root to download.
-    next_version = current_root_metadata['version'] + 1
-
     # Temporarily set consistent snapshot. Will be updated to whatever is set
     # in the latest root.json after running through the intermediates with
     # _update_metadata().
     self.consistent_snapshot = True
 
-    # Following the spec, try downloading the N+1th root for as long as we can.
-    while True:
-      # Try downloading the next root.
+    # Following the spec, try downloading the N+1th root for a certain maximum
+    # number of times.
+    lower_bound = current_root_metadata['version'] + 1
+    upper_bound = lower_bound + tuf.settings.MAX_NUMBER_ROOT_ROTATIONS
+
+    # Try downloading the next root.
+    for next_version in range(lower_bound, upper_bound):
       try:
         # Thoroughly verify it.
         self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH,
-          version=next_version)
+            version=next_version)
       # The first time we run into any HTTP error, break out of loop.
-      except requests.exceptions.HTTPError:
-        logging.traceback('Failed to download root version '+str(next_version))
+      except (requests.exceptions.HTTPError,
+              tuf.exceptions.NoWorkingMirrorError) as exception:
+        # Calling this function should give us a detailed stack trace including
+        # an HTTP error code, if any.
+        logging.exception('Failed to download root version '+str(next_version))
         break
 
       # Ensure that the role and key information of the top-level roles is the
@@ -1155,12 +1159,9 @@ class Updater(object):
       # should refresh them here as a protective measure.  See Issue #736.
       self._rebuild_key_and_role_db()
 
-      # Set the next version of root to download.
-      next_version += 1
-
     # Set our consistent snapshot property to what the latest root has said.
     self.consistent_snapshot = \
-      self.metadata['current']['root']['consistent_snapshot']
+        self.metadata['current']['root']['consistent_snapshot']
 
 
 
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -145,7 +145,6 @@ import securesystemslib.keys
 import securesystemslib.util
 import six
 import iso8601
-import requests.exceptions
 
 # The Timestamp role does not have signed metadata about it; otherwise we
 # would need an infinite regress of metadata. Therefore, we use some
@@ -1142,12 +1141,22 @@ class Updater(object):
         # Thoroughly verify it.
         self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH,
             version=next_version)
-      # The first time we run into any HTTP error, break out of loop.
-      except (requests.exceptions.HTTPError,
-              tuf.exceptions.NoWorkingMirrorError) as exception:
-        # Calling this function should give us a detailed stack trace including
-        # an HTTP error code, if any.
-        logging.exception('Failed to download root version '+str(next_version))
+      # When we run into HTTP 403 /404 error from ALL mirrors, break out of
+      # loop, because the next root metadata file is most likely missing.
+      except tuf.exceptions.NoWorkingMirrorError as exception:
+        for mirror_error in exception.mirror_errors.values():
+          # Otherwise, reraise the error, because it is not a simple HTTP
+          # error.
+          if not isinstance(mirror_error, six.moves.urllib.error.HTTPError) \
+             or mirror_error.code not in {403, 404}:
+            raise
+          else:
+            # Calling this function should give us a detailed stack trace
+            # including an HTTP error code, if any.
+            logging.exception('HTTP error for root version '+str(next_version))
+        # If we are here, then we ran into only 403 / 404 errors, which are
+        # good reasons to suspect that the next root metadata file does not
+        # exist.
         break
 
       # Ensure that the role and key information of the top-level roles is the
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -145,6 +145,7 @@ import securesystemslib.keys
 import securesystemslib.util
 import six
 import iso8601
+import requests.exceptions
 
 # The Timestamp role does not have signed metadata about it; otherwise we
 # would need an infinite regress of metadata. Therefore, we use some
@@ -1125,6 +1126,16 @@ class Updater(object):
       None.
     """
 
+    def neither_403_nor_404(mirror_error):
+      if isinstance(mirror_error, six.moves.urllib.error.HTTPError):
+        if mirror_error.code in {403, 404}:
+          return False
+      elif isinstance(mirror_error, requests.exceptions.HTTPError):
+        if mirror_error.response.status_code in {403, 404}:
+          return False
+      else:
+        return True
+
     # Temporarily set consistent snapshot. Will be updated to whatever is set
     # in the latest root.json after running through the intermediates with
     # _update_metadata().
@@ -1147,8 +1158,8 @@ class Updater(object):
         for mirror_error in exception.mirror_errors.values():
           # Otherwise, reraise the error, because it is not a simple HTTP
           # error.
-          if not isinstance(mirror_error, six.moves.urllib.error.HTTPError) \
-             or mirror_error.code not in {403, 404}:
+          if neither_403_nor_404(mirror_error):
+            logging.exception('Misc error for root version '+str(next_version))
             raise
           else:
             # Calling this function should give us a detailed stack trace
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1148,7 +1148,7 @@ class Updater(object):
         # Thoroughly verify it.
         self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH,
             version=next_version)
-      # When we run into HTTP 403 /404 error from ALL mirrors, break out of
+      # When we run into HTTP 403/404 error from ALL mirrors, break out of
       # loop, because the next root metadata file is most likely missing.
       except tuf.exceptions.NoWorkingMirrorError as exception:
         for mirror_error in exception.mirror_errors.values():
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1127,14 +1127,14 @@ class Updater(object):
     """
 
     def neither_403_nor_404(mirror_error):
+      WHITELIST = {403, 404}
       if isinstance(mirror_error, six.moves.urllib.error.HTTPError):
-        if mirror_error.code in {403, 404}:
+        if mirror_error.code in WHITELIST:
           return False
       elif isinstance(mirror_error, requests.exceptions.HTTPError):
-        if mirror_error.response.status_code in {403, 404}:
+        if mirror_error.response.status_code in WHITELIST:
           return False
-      else:
-        return True
+      return True
 
     # Temporarily set consistent snapshot. Will be updated to whatever is set
     # in the latest root.json after running through the intermediates with
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1128,7 +1128,7 @@ class Updater(object):
     # Retrieve the latest, remote root.json *WITHOUT* verifying it just yet.
     # We will verify it soon if all goes well.
     latest_root_metadata_file = self._get_file(
-        'root.json', lambda f: True, 'meta', DEFAULT_ROOT_UPPERLENGTH,
+        'root.json', lambda f: None, 'meta', DEFAULT_ROOT_UPPERLENGTH,
         download_safely=False)
 
     latest_root_metadata = securesystemslib.util.load_json_string(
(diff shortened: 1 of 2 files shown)


JustinCappos:
@tanishqjasoria Can you also take a look?

trishankatdatadog:
Yeah, I should say that someone should triple-check this to make we are not skipping some important security check. But at that point, all we are trying to do is get the latest version number of the root.

Cc @lukpueh

tanishqjasoria:
I guess there might be a problem. Form switching to _get_file() from _get_metadata_file(), the root.json we get is not verified.
Then, we are checking if the root metadata is changed or not by comparing the next_version and latest_version and the latest_version is extracted from the unverified root.json file. 
So if the unverified root.json file points that the latest_version is same as the current_version, then the root metadata would not be updated even if there is an update available and we would continue using the old root metadata.

trishankatdatadog:
@tanishqjasoria Good observation, but this would be true even if with a signed and verified "latest" root metadata file: attackers who control the repository can always perform a freeze attack (limited by the expiration of the previous root, which could be as long as years).

tanishqjasoria:
@trishankatdatadog I missed that!

tuf/client/updater.py
```diff
@@ -1152,6 +1154,16 @@ def _update_root_metadata(self, current_root_metadata):
...
       self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH, version=version)
+      # Ensure that the role and key information of the top-level roles is the
+      # latest.  We do this whether or not Root needed to be updated, in order
+      # to ensure that, e.g., the entries in roledb for top-level roles are
+      # populated with expected keyid info so that roles can be validated.  In
+      # certain circumstances, top-level metadata might be missing because it
+      # was marked obsolete and deleted after a failed attempt, and thus we
+      # should refresh them here as a protective measure.  See Issue #736.
+      self._rebuild_key_and_role_db()
+      self.consistent_snapshot = \
```
trishankatdatadog: While I tested that this code works, we need to figure out how to set `self.consistent_snapshot` in every iteration: clearly what we have is not optimal.

tuf/client/updater.py
```diff
@@ -1130,9 +1130,11 @@ def _update_root_metadata(self, current_root_metadata):
       None.
     """
 
-    # Retrieve the latest, remote root.json.
-    latest_root_metadata_file = self._get_metadata_file(
-        'root', 'root.json', DEFAULT_ROOT_UPPERLENGTH, None)
+    # Retrieve the latest, remote root.json *WITHOUT* verifying it just yet.
+    # We will verify it soon if all goes well.
+    latest_root_metadata_file = self._get_file(
+        'root.json', lambda f: True, 'meta', DEFAULT_ROOT_UPPERLENGTH,
```
lukpueh: Very very minor nit: I'd change the second argument to `lambda f: None` as the expected function does not use a return value (definitely not a blocker).

tuf/client/updater.py
```diff
@@ -1152,6 +1154,16 @@ def _update_root_metadata(self, current_root_metadata):
...
+      # Ensure that the role and key information of the top-level roles is the
+      # latest.  We do this whether or not Root needed to be updated, in order
+      # to ensure that, e.g., the entries in roledb for top-level roles are
+      # populated with expected keyid info so that roles can be validated.  In
+      # certain circumstances, top-level metadata might be missing because it
+      # was marked obsolete and deleted after a failed attempt, and thus we
+      # should refresh them here as a protective measure.  See Issue #736.
+      self._rebuild_key_and_role_db()
+      self.consistent_snapshot = \
+          self.metadata['current']['root']['consistent_snapshot']
```
lukpueh: Comment in L1152-1154 claims that `self.consistent_snapshot` is set with `_update_metadata()`.

https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1152-L1154

 Why set it here explicitly? Is the comment above inaccurate?

tuf/client/updater.py
```diff
@@ -1130,9 +1130,11 @@ def _update_root_metadata(self, current_root_metadata):
       None.
     """
 
-    # Retrieve the latest, remote root.json.
-    latest_root_metadata_file = self._get_metadata_file(
-        'root', 'root.json', DEFAULT_ROOT_UPPERLENGTH, None)
+    # Retrieve the latest, remote root.json *WITHOUT* verifying it just yet.
+    # We will verify it soon if all goes well.
+    latest_root_metadata_file = self._get_file(
```
lukpueh: By moving from `_get_metadata_file` to `_get_file` you are "bypassing" the following checks on the downloaded root metadata file:
-  is [`spec_version` field](https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1519-L1520) properly formatted,
- is [major `spec_version` supported](https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1513-L1516),
- if there is an existing root file, is [current metadata version lower](https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1548-L1550) than of downloaded root file, and
- via [`_verify_uncrompressed_metadata_file`](https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1555):
  - is [metadata gernally properly formatted](https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1428),
  - is  [metadata  not expired](https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1431), and
  - does [metadata have good signatures](https://github.com/theupdateframework/tuf/blob/c75a2b7fe1eaabc88f82a5f0bdfc8bd25cab7ab1/tuf/client/updater.py#L1438-L1442).

Is this intended? I don't see where these checks are made up for. Is the chain of trusted root files, as described in the docstring of `_update_root_metadata`, established somewhere else?

lukpueh (commented):
Thanks for the patch, @trishankatdatadog! I am not sure I understand the full consequences of your changes. Could you provide answers to my questions below?

tuf/client/updater.py
trishankatdatadog: Done

tuf/client/updater.py
trishankatdatadog: IDK, it's possible.

tuf/client/updater.py
trishankatdatadog: Certainly we don't want to check for expiry and good signatures right now, because let's say you're at v2, you don't have v3 yet, and the latest is v4. You just want to find out what the latest version is (v4), so that you rotate from v2->v3, then v3->v4, performing the metadata expiry and signatures on each rotation.

There might be some other useful checks we can do, such as checking `spec_version`, but the thing we want to _first_ is to get the _latest_ root _version number_. Does this make sense?

tuf/client/updater.py
```diff
@@ -1146,12 +1148,26 @@ def _update_root_metadata(self, current_root_metadata):
...
+    # in the latest root.json after running through the intermediates with
+    # _update_metadata().
+    self.consistent_snapshot = True
+
     for version in range(next_version, latest_version + 1):
-      # Temporarily set consistent snapshot. Will be updated to whatever is set
-      # in the latest root.json after running through the intermediates with
-      # _update_metadata().
-      self.consistent_snapshot = True
       self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH, version=version)
```
lukpueh: I somehow missed this line in my first review. It downloads all the root files again using `_get_metadata_file`, with all its verification routines I pointed out in my first review.

lukpueh (commented):
I somehow missed the call to `_update_metadata` in the top line of below for-loop. It downloads all the root files again using `_get_metadata_file` exactly how it used to fetch `latest_root_metadata_file`, before your patch.  That means, all the checks that I listed in my last review are performed eventually. 

So I'd say, LGTM! :)

Do you have capacities to add a test that has at least one intermediate root and where the latest root is signed with keys that are not available in the beginning?


On a side-note, I re-read the [relevant part of the spec](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#the-client-application) and it seems a bit different from what we do here. As per the spec the client downloads root files with incrementing version numbers until the server *404*s. So in the spec there is no initial download of the latest root.

And on  another side-note, did you see @vladimir-v-diaz's TODO comment in the `refresh`, where _update_root_metadata is called?

https://github.com/theupdateframework/tuf/blob/cf6fdec5101fec5e2402dc8d073fa17ace1fc25f/tuf/client/updater.py#L1079-L1082

trishankatdatadog:
@lukpueh Following our discussion, I have adjusted the code to [follow the spec](https://github.com/theupdateframework/specification/blame/bc97d9660bb6b9234fa3d41140978344ce698b84/tuf-spec.md#L1008). Please review.

To answer your questions:
1. Could someone else please help add the tests? Maybe an intern...
2. That TODO requires a wider discussion on our mailing list. @JustinCappos

trishankatdatadog:
Our spec should also be updated to say: once you have turned on consistent snapshots, you should always also write versioned besides unversioned root metadata files so that outdated clients can updated to your latest root. @JustinCappos 

Update: https://github.com/theupdateframework/specification/pull/52

tuf/client/updater.py
```diff
@@ -1130,28 +1131,41 @@ def _update_root_metadata(self, current_root_metadata):
...
+    # _update_metadata().
+    self.consistent_snapshot = True
 
+    # Following the spec, try downloading the N+1th root for as long as we can.
+    while True:
+      # Try downloading the next root.
+      try:
+        # Thoroughly verify it.
+        self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH,
+          version=next_version)
```
lukpueh: Minor nit: We use double-indentation for line continuation.

tuf/client/updater.py
```diff
@@ -1130,28 +1131,41 @@ def _update_root_metadata(self, current_root_metadata):
...
+      # was marked obsolete and deleted after a failed attempt, and thus we
+      # should refresh them here as a protective measure.  See Issue #736.
+      self._rebuild_key_and_role_db()
+
+      # Set the next version of root to download.
+      next_version += 1
+
+    # Set our consistent snapshot property to what the latest root has said.
+    self.consistent_snapshot = \
+      self.metadata['current']['root']['consistent_snapshot']
```
lukpueh: Minor nit: We use double-indentation for line continuation.

tuf/client/updater.py
```diff
@@ -1130,28 +1131,41 @@ def _update_root_metadata(self, current_root_metadata):
...
 
+    # Following the spec, try downloading the N+1th root for as long as we can.
+    while True:
+      # Try downloading the next root.
+      try:
+        # Thoroughly verify it.
+        self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH,
+          version=next_version)
+      # The first time we run into any HTTP error, break out of loop.
+      except requests.exceptions.HTTPError:
```
lukpueh: Although the spec does say to loop until the *"file is not available"*  (see last sentence in step 1.2. of [5. Detailed Workflows](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#the-client-application)), I wonder if there should be some case handling depending on the http status code. E.g. 
4xx -> no newer root
5xx -> no newer root, but also, the server seems to have gone south

In any case, I'd at least add the status code (exception message) to your logger call.

tuf/client/updater.py
```diff
@@ -1130,32 +1131,26 @@ def _update_root_metadata(self, current_root_metadata):
...
 
     # Temporarily set consistent snapshot. Will be updated to whatever is set
     # in the latest root.json after running through the intermediates with
     # _update_metadata().
     self.consistent_snapshot = True
 
-    for version in range(next_version, latest_version + 1):
-      self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH, version=version)
+    # Following the spec, try downloading the N+1th root for as long as we can.
+    while True:
```
lukpueh: Should there be an upper bound for iterations similar to `DEFAULT_ROOT_UPPERLENGTH`?

tuf/client/updater.py
```diff
@@ -1130,28 +1131,41 @@ def _update_root_metadata(self, current_root_metadata):
...
 
-    latest_root_metadata = securesystemslib.util.load_json_string(
-        latest_root_metadata_file.read().decode('utf-8'))
+    # Temporarily set consistent snapshot. Will be updated to whatever is set
+    # in the latest root.json after running through the intermediates with
+    # _update_metadata().
+    self.consistent_snapshot = True
 
+    # Following the spec, try downloading the N+1th root for as long as we can.
+    while True:
```
lukpueh: Should there be an upper bound for iterations similar to `DEFAULT_ROOT_UPPERLENGTH`?

tuf/client/updater.py
```diff
@@ -1130,28 +1131,41 @@ def _update_root_metadata(self, current_root_metadata):
...
+      # Try downloading the next root.
+      try:
+        # Thoroughly verify it.
+        self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH,
+          version=next_version)
+      # The first time we run into any HTTP error, break out of loop.
+      except requests.exceptions.HTTPError:
+        logging.traceback('Failed to download root version '+str(next_version))
```
lukpueh: `logging.traceback` is a module, not a function

tuf/client/updater.py
lukpueh: `requests.exceptions.HTTPError` is not the exception you are looking for. That's also why all tests fail. If `_update_metadata` should raise a `tuf.exceptions.NoWorkingMirrorError` if no mirror served the corresponding file.

lukpueh (changes requested):
Thanks for your update, @trishankatdatadog. It does look more like the spec now. However, the code does not work yet. @JustinCappos, do we have an intern who can take over, i.e. fix the things I noted below and add some tests?

trishankatdatadog:
@lukpueh Sorry, I wrote on one machine, and tested on another, so there were some mistakes in translation.

Does this work better for you?

tuf/client/updater.py
trishankatdatadog: Ok, I tested that AWS returns HTTP 403 on an S3 bucket / CloudFront distribution which does not permit listing of files. I also added `tuf.exceptions.NoWorkingMirrorError` based on your comment.

tuf/client/updater.py
```diff
@@ -1130,28 +1131,62 @@ def _update_root_metadata(self, current_root_metadata):
...
 
-    # Retrieve the latest, remote root.json.
-    latest_root_metadata_file = self._get_metadata_file(
-        'root', 'root.json', DEFAULT_ROOT_UPPERLENGTH, None)
+    def neither_403_nor_404(mirror_error):
+      WHITELIST = {403, 404}
+      if isinstance(mirror_error, six.moves.urllib.error.HTTPError):
+        if mirror_error.code in WHITELIST:
+          return False
+      elif isinstance(mirror_error, requests.exceptions.HTTPError):
```
lukpueh: Do you expect `requests.exceptions.HTTPError` here? I think you should be fine with  `six.moves.urllib.error.HTTPError`.

lukpueh (commented):
Thanks for the updates! I'm unsure whether you need `request.exceptions`. Otherwise it looks good.

tuf/client/updater.py
trishankatdatadog: Apparently we do. For some reason I haven't figured out yet, I see the first exception on my local tox, but the second one of the Travis tox. Makes zero sense, but that's what seems to be happening. I'll try removing either one and check on both environments.

lukpueh: I just cannot reproduce what you're saying. I ran Travis on [a commit that re-raises the exception if it is an instance of `requests.exceptions.HTTPError`](https://github.com/lukpueh/tuf/blob/70d9146909019ebe523bfb199d3a6f2d0056a52f/tuf/client/updater.py#L1139-L1141), but the exception is never raised.

If you scroll down to the end of the corresponding Travis build logs and take a look at the coverage summary you can see how the corresponding lines ([L1140 - L1141](https://github.com/lukpueh/tuf/blob/70d9146909019ebe523bfb199d3a6f2d0056a52f/tuf/client/updater.py#L1140-L1141)) never get touched (Travis logs for [Py2](https://travis-ci.com/lukpueh/tuf/jobs/210722135), [Py3.4](https://travis-ci.com/lukpueh/tuf/jobs/210722136) and [Py3.6](https://travis-ci.com/lukpueh/tuf/jobs/210722137)).

lukpueh: Btw. I don't understand what's going on in the coverage reports in your latest Travis builds. It shows missing branches that correspond to docstring lines. Something seems to be off there.

lukpueh: It doesn't seem likely, but could this somehow be related to TUF's Travis builds haven't yet been [migrated to `travis-ci.com`](https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/)?

lukpueh: Okay, @trishankatdatadog, hold on. I know what happened, at least why I couldn't reproduce the `requests.exceptions.HTTPError` neither locally nor on Travis, and why you only saw them on Travis.

Your Travis builds ran over the PR, i.e. your changes on top of the current upstream `develop` branch. While locally, I suppose, you ran the tests on your `trishankatdatadog:trishankatdatadog/correctly-rotate-root` branch, which is on top of a pretty outdated `develop` branch. I'm sure that if you rebase onto `upstream/develop`, you'll get the same results locally. At least that (an outdated base develop) was the reason why I couldn't reproduce the errors.

lukpueh: So it turns out that we can scrap the first instance check, i.e. for `six.moves.urllib.error.HTTPError`. 

`six.moves.urllib.request` was replaced by `requests` in b9bc8602c2bdc31f09b241f3612f7fda6d601ea9. But those changes aren't part of the `trishankatdatadog:trishankatdatadog/correctly-rotate-root` branch, hence the different results locally and on Travis.

trishankatdatadog: Great job, @lukpueh, done, thanks!

lukpueh:
Thanks for the update, @trishankatdatadog! This looks sound now. I'd still like to see tests before we merge. @JustinCappos, do we have a student you might want to volunteer? If not, I'll add some tests.

trishankatdatadog:
This is very important to Datadog for our next update. Could we get this fixed ASAP? Thanks.

lukpueh:
FYI. I am working on tests...

tuf/client/updater.py
```diff
@@ -1125,28 +1126,58 @@ def _update_root_metadata(self, current_root_metadata):
...
+    lower_bound = current_root_metadata['version'] + 1
+    upper_bound = lower_bound + tuf.settings.MAX_NUMBER_ROOT_ROTATIONS
+
+    # Try downloading the next root.
+    for next_version in range(lower_bound, upper_bound):
+      try:
+        # Thoroughly verify it.
+        self._update_metadata('root', DEFAULT_ROOT_UPPERLENGTH,
+            version=next_version)
+      # When we run into HTTP 403 /404 error from ALL mirrors, break out of
```
trishankatdatadog:
```suggestion
      # When we run into HTTP 403/404 error from ALL mirrors, break out of
```

trishankatdatadog:
Thanks for helping to write the tests, @lukpueh...

lukpueh:
And here they are: https://github.com/trishankatdatadog/tuf/compare/b170601...lukpueh:correctly-rotate-root-test

@trishankatdatadog, could you please sign off your most recent commit? And would you mind adding my three tests-adding commits linked above on top of your branch? Then we can review the tests here on the same PR.

trishankatdatadog:
@lukpueh I had to manually set DCO to pass, should be OK (only one commit done directly from GitHub didn't have sign off). Great test, BTW. Let me know if you need anything else!

lukpueh:
Merged manually via https://github.com/theupdateframework/tuf/commit/3d342e648fbacdf43a13d7ba8886aaaf07334af7. Tests will be added with https://github.com/theupdateframework/tuf/pull/930.

lukpueh:
Thanks for the fix, @trishankatdatadog!! 💯

PR #855, opened 2019-04-16
Adjust to removals from securesystemslib

awwad:
Removal of securesystemslib exceptions that are TUF-specific occurs in [securesystemslib PR #165](https://github.com/secure-systems-lab/securesystemslib/pull/165).

This PR adapts to those changes.  Exceptions that are specific to TUF should be in TUF and not in securesystemslib.  This commit uses those already-existing TUF exceptions instead of pointing to securesystemslib exceptions that will be removed.

For example, securesystemslib has no notion of repositories, so it's clearly wrong to have a `RepositoryError` in securesystemslib and clearly wrong for TUF to use `securesystemslib.exceptions.RepositoryError`.

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #912, opened 2019-09-13
Use TUF specific formats as they have been removed from securesystemslib

joshuagl:

Use TUF specific RELPATH_SCHEMA and RELPATHS_SCHEMA formats as they are removed from securesystemslib in secure-systems-lab/securesystemslib#165

Without this change we can't use tuf with the master branch of securesystemslib

Aside: is it tuf or TUF? I've seen both used and wasn't sure which to use in my commit message.

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -71,9 +71,9 @@ class TestFormats(unittest.TestCase):
 
       'SCHEME_SCHEMA': (securesystemslib.formats.SCHEME_SCHEMA, 'rsassa-pss-sha256'),
 
-      'RELPATH_SCHEMA': (securesystemslib.formats.RELPATH_SCHEMA, 'metadata/root/'),
+      'RELPATH_SCHEMA': (tuf.formats.RELPATH_SCHEMA, 'metadata/root/'),
 
-      'RELPATHS_SCHEMA': (securesystemslib.formats.RELPATHS_SCHEMA,
+      'RELPATHS_SCHEMA': (tuf.formats.RELPATHS_SCHEMA,
                           ['targets/role1/', 'targets/role2/']),
 
       'PATH_SCHEMA': (securesystemslib.formats.PATH_SCHEMA, '/home/someuser/'),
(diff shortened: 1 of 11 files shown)


awwad:
Sure.  There's also [PATH_SCHEMA](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/formats.py#L128) still in SSL, which you could use.  There's no meaningful distinction in the codebase between relative and absolute paths, and treatment is identical -- validation and schema are/were the same.

There are some other changes that are probably necessary (unless the securesystemslib PR had some additional changes I haven't peeked at), like these:
https://github.com/theupdateframework/tuf/pull/855

joshuagl:
There were indeed some required changes missing, I've just pushed an updated patch which includes a complete set of changes to use schemas which are already defined in tuf.formats

This change shuffles around tuf.formats to ensure all schemas are defined before they are used, however it may be desirable to be more intentional about the order of schema declarations in this file?

joshuagl:
The second patch 4e6942f8465122b1cd5dc72c931cad48629ef3ac results in breakage because we're in a situation where we're encoding data before passing to ssl and yet the older version of ssl is attempting to encode the data internally also. 

I'd welcome thoughts on whether to try and be compatible with old/current ssl and master/next-version ssl?

It will be easier to be confident of these changes once #915 has landed and we can verify this change against the full Tox test matrix.

This change should land after #855, the combination of both PRs should result in successful builds of tuf master against ssl master.

lukpueh:
@joshuagl, I left a [comment on #913](https://github.com/theupdateframework/tuf/issues/913#issuecomment-532124275) that also has some thoughts regarding compatibility with old/current ssl and master/next-version ssl.

joshuagl:
Thanks for the feedback @lukpueh , I'm changing this PR back to just addressing the issue of formats removed from ssl and added to tuf. This doesn't break compatibility with the existing ssl dependency so should be good to merge.

lukpueh (approved):
Thanks for the patch, @joshuagl! LGTM, i.e. a couple of module path updates to account for secure-systems-lab/securesystemslib#165 and thus necessary re-orderings in `tuf.formats.py`.


On a side note, and to answer your question about TUF vs. tuf. I think I tend to use the former when talking about the project, as in _"The Update Framework (**TUF**)_, and the latter when talking about the package, e.g. _"pip install **tuf**"_.

PR #918, opened 2019-09-17
Remove uses of securesystemslib.util.TempFile

joshuagl:
### Note: this change breaks compatibility with ssl v0.11.3 and thus should not be merged until a new release of ssl has been made and is available on PyPI.

**Description of the changes being introduced by the pull request**: remove use of `securesystemslib.util.TempFile` in favour of `tempfile.TemporaryFile` from the Python standard library. SSL's `TempFile` included features which are no longer used by TUF or in-toto and had API which was inconsistent with normal Python file objects, and thus was removed.

requirements**:


lukpueh:
Will be merged with #934.

PR #921, opened 2019-09-20
Bump version 0.12.0

lukpueh:
**TODO**
1. Wait for securesystemslib 0.12.0 to be released on PyPI
1. Merge #917 and #918 to adapt to breaking changes in securesystemslib
1. Rebase this release PR onto develop
1. Release tuf 0.12.0-dev0
*Alternatively, and if preferred, release as 1.0.0-dev0 instead (requires changing the version numbers in `CHANGELOG.md`, `setup.py` and `__init__.py`)*

PR #928, opened 2019-10-03
Add disclaimer about limited slow retrieval attack protection

lukpueh:
**Fixes issue #**:
None. Related to #781.

Since https://github.com/theupdateframework/tuf/pull/781 we only provide limited protection against slow retrieval attacks. So far this has only been discussed in above issue and hinted at by a disabled test and a code comment in that test.

This change adds a corresponding disclaimer to a more prominent place, i.e. the list of attacks in SECURITY.md.

requirements**:

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -20,7 +20,8 @@ snapshot metadata, and thus new updates could never be downloaded.
 
 * **Endless data attacks**. An attacker responds to a file download request with an endless stream of data, causing harm to clients (e.g. a disk partition filling up or memory exhaustion). 
 
-* **Slow retrieval attacks**. An attacker responds to clients with a very slow stream of data that essentially results in the client never continuing the update process. 
+* **~~Slow retrieval attacks~~**. An attacker responds to clients with a very slow stream of data that essentially results in the client never continuing the update process.\
+**_NOTE: Due to limitations in a 3rd-party HTTP library, the TUF reference implementation currently provides only limited protection against slow retrieval attacks (see [tuf#932](https://github.com/theupdateframework/tuf/issues/932)). We plan to fix this in a future release._**
 
 * **Extraneous dependencies attacks**. An attacker indicates to clients that in order to install the software they wanted, they also need to install unrelated software. This unrelated software can be from a trusted source but may have known vulnerabilities that are exploitable by the attacker. 
 


lukpueh:
We may also want to exclude (or add a comment to) the section about slow retrieval attacks in the tutorial-like ATTACKS.md document.

docs/SECURITY.md
```diff
@@ -20,7 +20,8 @@ snapshot metadata, and thus new updates could never be downloaded.
 
 * **Endless data attacks**. An attacker responds to a file download request with an endless stream of data, causing harm to clients (e.g. a disk partition filling up or memory exhaustion). 
 
-* **Slow retrieval attacks**. An attacker responds to clients with a very slow stream of data that essentially results in the client never continuing the update process. 
+* **~~Slow retrieval attacks~~**. An attacker responds to clients with a very slow stream of data that essentially results in the client never continuing the update process.\
+**_NOTE: The TUF reference implementation currently provides only limited protection against slow retrieval attacks (see [tuf#781](https://github.com/theupdateframework/tuf/pull/781))._**
```
JustinCappos: Should we be trying to fix this upstream in request?  Is that sensible?  If so, should we link to that discussion instead?

The issue we refer to here has a lot going on, I'd prefer we had something clearer to point to.

trishankatdatadog:
```suggestion
**_NOTE: Due to limitations in a 3rd-party HTTP library, the TUF reference implementation currently provides only limited protection against slow retrieval attacks (see [tuf#781](https://github.com/theupdateframework/tuf/pull/781)). We plan to fix this in a future release._**
```

trishankatdatadog (changes requested):

trishankatdatadog:
Thanks, @lukpueh! Just want to make it clearer that this limitation is not due to our code...

docs/SECURITY.md
trishankatdatadog: Yeah, I guess we should at least file an issue that points to slow loris attacks.

lukpueh:
I created an issue #932 (downstream for the time being) that summarizes key points from #781 and updated the disclaimer in this PR. Please review/approve/merge.

 I think this is the last thing we wanted to get in before bumping tuf to 0.12.0 (https://github.com/theupdateframework/tuf/pull/921).

mnm678 (approved):
LGTM

JustinCappos:
Is the failing test a race condition / Heisenbug?  I would merge otherwise

lukpueh:
> Is the failing test a race condition / Heisenbug? I would merge otherwise

It's a timeout issue, which seems to happen fairly often for tests on appveyor/windows (less often on travis/linux too). In any case, the fail here is impossibly related to the change in this PR. I say it's safe to merge.

Issue #965, opened 2019-12-11
test_key_revocation_integration.TestKeyRevocation fails on windows with Python 2.7


joshuagl:
Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.

**Description of issue or feature request**:

The full error log can be viewed on [AppVeyor](https://ci.appveyor.com/project/lukpueh/tuf/builds/29255410/job/9yiyp3m58mma66o9), I have pasted the section containing errors below:

```
....................E
Stderr:
DEBUG:tuf.client.updater:Adding roles delegated from u'targets'.
DEBUG:tuf.client.updater:Adding delegated role: role1.
INFO:tuf.download:Downloading: u'http://localhost:43996/tmp_hppez/TestKeyRevocation_vtpw4q/repository/metadata/2.root.json'
DEBUG:tuf.download:url: http://localhost:43996/tmp_hppez/TestKeyRevocation_vtpw4q/repository/metadata/2.root.json
DEBUG:tuf.download:session index: http+localhost
DEBUG:tuf.download:Reusing session for http+localhost
ERROR:tuf.download:Could not download URL: u'http://localhost:43996/tmp_hppez/TestKeyRevocation_vtpw4q/repository/metadata/2.root.json'
Traceback (most recent call last):
  File "c:\projects\tuf\tuf\download.py", line 264, in _download_file
    response.raise_for_status()
  File "C:\Python27\lib\site-packages\requests\models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 404 Client Error: File not found for url: http://localhost:43996/tmp_hppez/TestKeyRevocation_vtpw4q/repository/metadata/2.root.json
ERROR:tuf.client.updater:Update failed from http://localhost:43996/tmp_hppez/TestKeyRevocation_vtpw4q/repository/metadata/2.root.json.
Traceback (most recent call last):
  File "c:\projects\tuf\tuf\client\updater.py", line 1507, in _get_metadata_file
    upperbound_filelength)
  File "c:\projects\tuf\tuf\download.py", line 150, in unsafe_download
    return _download_file(url, required_length, STRICT_REQUIRED_LENGTH=False)
  File "c:\projects\tuf\tuf\download.py", line 264, in _download_file
    response.raise_for_status()
  File "C:\Python27\lib\site-packages\requests\models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
[...]

PR #985, opened 2020-02-24
Fix failing AppVeyor Python2.7 tests

lukpueh:
**Fixes issue #**:
#965 

Since #885 the tests in TestUpdater and TestKeyRevocation fail on Appveyor Python 2.7 builds. After some live debugging, it turns out that the tests fail due to the extra amount of http requests to the simple http server (see tests/simple_server.py) that were added in #885.

The simple server runs in a subprocess and is re-used for the entire TestCase. After a certain amount of requests it becomes unresponsive. Note that neither the subprocess exits (ps -W), nor does the port get closed (netstat -a). It just doesn't serve the request, making it time out and fail the test.

The following script can be used to reproduce the issue (run in tests directory):

```python
import subprocess
import requests
import random

counter = 0

port = random.randint(30000, 45000)
command = ['python', 'simple_server.py', str(port)]
server_process = subprocess.Popen(command, stderr=subprocess.PIPE)
url = 'http://localhost:'+str(port) + '/'

sess = requests.Session()

try:
  while True:
    sess.get(url, timeout=3)
    counter +=1

finally:
  print(counter)
  server_process.kill()
```

It fails repeatedly on the 69th request, but only if `stderr=subprocess.PIPE` is passed to Popen. Given that for each request the simple server writes about ~60 characters to stderr, e.g.

```
127.0.0.1 - - [24/Feb/2020 12:01:23] "GET / HTTP/1.1" 200 -
```
... it looks a lot like a full pipe buffer of size 4096. Note that the `bufsize` argument to Popen does not change anything.

As a simple workaround we silence the test server on Windows/Python2 to not fill the buffer.

requirements**:


joshuagl:
Great job tracking the sources of failure down @lukpueh! These changes LGTM.

lukpueh:
Thanks for the feedback, @joshuagl! I say we merge this.

Issue #933, opened 2019-10-11
Remove root from snapshot 


lukpueh:
**Description of issue or feature request**:
https://github.com/theupdateframework/specification/pull/40 updates the specification to remove the `"root.json"` entry form the `"meta"` dictionary in `"snapshot.json"`.

The reference implementation needs to be adopted accordingly and comprehensively, i.e. in code, test data, docs, etc.

**Current behavior**:
Code, test data, docs (etc.) list root.json in snapshot.json.

**Expected behavior**:
Update where necessary to reflect the spec's definition of snapshot.json.

*Hints:*
- The commit message in af1a21b05afa77a9c465637ca8fccc55a224c8c9 has some info about how to re-generate test metadata.
- https://github.com/theupdateframework/specification/issues/31#issuecomment-487055404 has some info about docs.

trishankatdatadog:
Thanks for noting this @lukpueh

ChamanAgrawal:
Hi!! I am new to TUF and want to contribute. For this issue, I started looking at the code base and I think, changes are required in SNAPSHOT_SCHEMA. 
I was thinking of just filtering the root.json from snapshot.meta the same way as timestamp.json would have been in the existing code. So I started looking for code that generates the metadata files.
I found that snapshot metadata is constructed [here](https://github.com/theupdateframework/tuf/blob/develop/tuf/repository_lib.py#L1358)  and this function returns the complete metadata json(dictionary) corresponding to the given schema using the function ``build_dict_conforming_to_schema()`` ([here](https://github.com/theupdateframework/tuf/blob/develop/tuf/formats.py#L482)). 
But I am stuck here as I am not able to find where the function ``build_dict_conforming_to_schema()`` filters the timestamp.json for snapshot.meta
Can anyone please help me with this? Also, if my approach is incorrect, please let me know. Thanks!!

lukpueh:
Thanks for picking up the issue, @ChamanAgrawal! :)

> But I am stuck here as I am not able to find where the function build_dict_conforming_to_schema() filters the timestamp.json for snapshot.meta

Not sure I understand the question. 
You have already found `generate_snapshot_metadata` (to me this seems like the harder part). Now you need to find the lines in that function that assemble the dictionary that will go into `"meta"` and remove the one line that adds the `"root.json"` entry.

[...]

PR #988, opened 2020-02-27
Remove root from snapshot

joshuagl:
**Fixes issue #**: #933 

Per the updated spec in theupdateframework/specification#31 `root.json` should no longer be listed in `snapshot.json`.

* Remove the listing of `root.json` from `snapshot.json` and update the tests
* Update the docs/METADATA.md entry on `snapshot.json` 
* Update docstrings and code comments that refer to the order of operations for top-level roles to reflect the workflow in the latest spec
* Modify the test repository generator to remove a warning, update a comment and pass relative paths to the target files in `add_target()` calls
* Regenerate the repository_data for the `snapshot.json` change

requirements**:


tests/repository_data/generate.py
```diff
@@ -119,12 +119,12 @@
 # about the target (i.e., file permissions in octal format.)
 octal_file_permissions = oct(os.stat(target1_filepath).st_mode)[4:]
 file_permissions = {'file_permissions': octal_file_permissions}
-repository.targets.add_target(target1_filepath, file_permissions)
-repository.targets.add_target(target2_filepath)
+repository.targets.add_target(os.path.basename(target1_filepath), file_permissions)
+repository.targets.add_target(os.path.basename(target2_filepath))
```
joshuagl, line 129: As @lukpueh's hint in #933 was to make changes like this in order to re-generate the test repository data I wondered why we aren't making this change to the `generate.py` script, rather than hacking it each time?

I will dig into this and try to better understand it.

joshuagl, line 129: As @lukpueh's hint in #933 was to make changes like this in order to re-generate the test repository data I wondered why we aren't making this change to the `generate.py` script, rather than hacking it each time?

I will dig into this and try to better understand it.

tests/test_updater.py
```diff
@@ -886,7 +886,8 @@ def test_3__update_metadata_if_changed(self):
     self.repository_updater._update_metadata('timestamp', DEFAULT_TIMESTAMP_FILELENGTH)
     self.repository_updater._update_metadata_if_changed('snapshot', 'timestamp')
     self.repository_updater._update_metadata_if_changed('targets')
-    self.repository_updater._update_metadata_if_changed('root')
+    # TODO: should this still work?
+    # self.repository_updater._update_metadata_if_changed('root')
```
joshuagl: I need to read through the spec and code to figure out whether this should still work, it throws:
```
  File "/home/joshuagl/Projects/tuf/tests/test_updater.py", line 890, in test_3__update_metadata_if_changed
    self.repository_updater._update_metadata_if_changed('root')
  File "/home/joshuagl/Projects/tuf/tuf/client/updater.py", line 1915, in _update_metadata_if_changed
    ['meta'][metadata_filename]
KeyError: 'root.json'
```

since the removal of `root.json` from `snapshot.json`, which I believe makes sense, but I have yet to verify (hence draft PR).

joshuagl (commented):
The PR is marked as draft as I'm sharing before I've finished the changes.

tests/test_updater.py
joshuagl: Removal of this test is correct, the `root` metadata is no longer referenced by `snapshot`, hence the traceback above. Furthermore, the root role is updated separately and verified based on the existing root metadata, following the detailed client workflow logic in the spec, in the `_update_root_metadata()` function.

joshuagl:
I believe this PR is ready for review

tests/repository_data/client/test_repository1/metadata/current/targets.json
```diff
@@ -40,7 +40,7 @@
   "targets": {
    "file1.txt": {
     "custom": {
-     "file_permissions": "0644"
+     "file_permissions": "0664"
```
mnm678: Why did the write bit get added to these?

trishankatdatadog: Good catch, Marina! Also, do we discuss anywhere why we set file permissions here at all? Is it show what the "custom" attribute can do? If so, we really ought to document it somewhere.

tests/repository_data/client/test_repository1/metadata/current/targets.json
```diff
@@ -28,7 +28,7 @@
      ],
      "name": "role1",
      "paths": [
-      "file3.txt"
+      "/file3.txt"
```
trishankatdatadog: I thought we removed adding a leading path separator to targets paths?

trishankatdatadog:
Other than @mnm678 and my comments, this PR LGTM!

tests/repository_data/client/test_repository1/metadata/current/targets.json
joshuagl: > Why did the write bit get added to these?

Nice catch!

664 is the default file permissions for user-created files on Fedora (my host OS) and `generate.py` does not explicitly set permissions on the target files at creation time. Presumably this differs from wherever the original files were generated.

I can edit the patch to drop this change?

tests/repository_data/client/test_repository1/metadata/current/targets.json
joshuagl: Good eye, this was caused by changing:
`repository.targets.delegate('role1', [delegation_public], [target3_filepath])`
to
`repository.targets.delegate('role1', [delegation_public], [], list_of_targets=[os.path.basename(target3_filepath)])`

I made that change because I was trying to do away with the incorrect warning from `generate.py` that:
> 'file3.txt' is not located in the repository's targets directory: '/home/joshuagl/Projects/tuf/tests/repository_data/repository/targets'

I misunderstood the documentation for `delegate()` on first read, therefore I think it makes sense to undo the change to the `delegate()` call in the generate tool and regenerate the repository_data for this PR.

Then separate from this PR I can create a new Issue to clarify the docs and remove the incorrect warning in `repository_tool.targets.delegate()`

tests/repository_data/client/test_repository1/metadata/current/targets.json
joshuagl: Default file permissions on Ubuntu and Debian are 644 (umask `0022` vs umask `0002` on Fedora), which I believe would account for the discrepancy. I'll edit the patch to drop this change.

tests/repository_data/client/test_repository1/metadata/current/targets.json
joshuagl: In fact, I think the warning isn't even incorrect. At the time that we call `delegate()` in `generate.py` the path being delegated is not yet listed in the targets metadata because we are delegating the path before we have added a target file (with `add_target()`) at that path.

This was all misunderstanding on my part, thanks for the diligent review @trishankatdatadog

tests/repository_data/client/test_repository1/metadata/current/targets.json
joshuagl: I added an extra patch efae583 to set the permissions on `file1.txt` so that we don't generate different data on different Linux host OS.

trishankatdatadog:
@joshuagl Yeah, unfortunately the tuf logger is very noisy, and sometimes misleading...

docs/METADATA.md
```diff
@@ -64,7 +64,7 @@ of delegated Targets metadata and [example](https://raw.githubusercontent.com/th
 
 Signed by: Snapshot role.
 
-The snapshot.json metadata file lists hashes and sizes of all metadata files other than timestamp.json. This file ensures that clients will see a consistent view of the files on the repository. That is, metadata files (and thus target file) that existed on the repository at different times cannot be combined and presented to clients by an attacker.
+The snapshot.json metadata file lists the version, and optionally the file hashes and sizes, of all metadata files other than root.json and timestamp.json. This file ensures that clients will see a consistent view of the files on the repository. That is, metadata files (and thus target file) that existed on the repository at different times cannot be combined and presented to clients by an attacker.
```
lukpueh: Minor nit: The *"all metadata files other"* wording is absolutely correct, but wouldn't it be more helpful to tell the reader what files to include, i.e. *"top-level targets metadata and all delegated targets metadata"* (instead of what files not to include)?
What do others think? This is definitely not a blocker for this PR. 

I proposed a similar change in the spec (see https://github.com/theupdateframework/specification/commit/8dc300f3d24298d3ccc3d154cd6221ed3f22843b and the newer https://github.com/theupdateframework/specification/commit/a5848e758a7aab27cb7143ecec287251a4f6d382)

tests/repository_data/generate.py
```diff
@@ -119,12 +124,12 @@
...
 
 repository.targets.delegate('role1', [delegation_public],
     [os.path.basename(target3_filepath)])
-repository.targets('role1').add_target(target3_filepath)
+repository.targets('role1').add_target(os.path.basename(target3_filepath))
 repository.targets('role1').load_signing_key(delegation_private)
 
 repository.targets('role1').delegate('role2', [delegation_public], [])
```
lukpueh, line 135: Thanks for taking the time to actually fixing this. It makes more sense than adding a dirty patch to a commit message. :)

tuf/client/updater.py
```diff
@@ -1836,21 +1831,23 @@ def _update_metadata_if_changed(self, metadata_role,
...
     <Purpose>
       Non-public method that updates the metadata for 'metadata_role' if it has
-      changed.  With the exception of the 'timestamp' role, all the top-level
+      changed.  All top-level roles other than the 'timestamp' and 'root'
       roles are updated by this method.  The 'timestamp' role is always
       downloaded from a mirror without first checking if it has been updated;
       it is updated in refresh() by calling _update_metadata('timestamp').
+      The 'root' role is always updated first and verified based on the trusted
+      root metadata file the client already has a copy of; it is updated in
+      refresh() by calling _update_root_metadata().
```
lukpueh, line 1840: Thanks for making the effort of updating the docstrings. If I understand this function correctly, then it should only be called for "referenced metadata" (On an unrelated side-note: I think the parameter should be called `referencing_metadata`). 

This also means that that the function should no longer be called for root metadata as it is not referenced by any other metadata.  Would you mind removing the `elif metadata_role == 'root':`-logic and `TODO: If 'metadata_role' is root `-comment in the function body?

docs/METADATA.md
joshuagl: That would be much more helpful indeed. I've force-pushed an update to this patch which changes this to:
> The snapshot.json metadata file lists the version, and optionally the file hashes and sizes, of the top-level targets metadata and all delegated targets metadata.

lukpueh (changes requested):
Excellent work and exemplary commit discipline, @joshuagl! I have one tiny request (see inline comment) otherwise it looks great to me.

By the way, I also tested your updates with the old metadata and tests still pass, which is great!

tuf/client/updater.py
joshuagl: I've pushed an additional patch to address the latter part of this comment

joshuagl:
Thank you for the diligent and exhaustive review, as always, @lukpueh ! Contributing to tuf is always a pleasure.

> By the way, I also tested your updates with the old metadata and tests still pass, which is great!

I should have thought to test that, happy to read it! 😄

lukpueh (approved):

Issue #964, opened 2019-12-02
`Repository.writeall()` may not need to write all dirty roles


lukpueh:
**Description of issue or feature request**:

TUF marks roles as dirty if they are changed in roledb. This information is then used by `writeall` to update role metadata on disk. But not all changes in roledb require updating metadata on disk. E.g. loading a signing key does not. 

**Current behavior**:
At least one function, i.e. `load_signing_key` marks a role as dirty, although it shouldn't update the corresponding metadata on disk.

**Expected behavior**:
Reliably track the state of the repository in memory and on disk to know if role metadata needs to be updated. Also see #958.

lukpueh:
When fixing this issue, we may remove the explicit `mark_dirty()`-calls in the tutorial code snippets (and the comments that point to this issue and to #958). Ideally `writeall()` will know by itself which metadata to update.

jku:
Closing this issue as it was filed against (what is now known as) the legacy codebase: issue seems to not be relevant anymore. Please re-open or file a new issue if you feel that the issue is revelant to current python-tuf.

### More details

Current source code (and upcoming 1.0 release) only contains the modern components
 * a low-level Metadata API (`tuf.api`) and 
 * `tuf.ngclient` that implements the client workflow, 

Legacy components (e.g. tuf.client, tuf.repository_tool, tuf.repository_lib as well as the repo and client scripts) are no longer included. See [announcement](https://github.com/theupdateframework/python-tuf/blob/develop/docs/1.0.0-ANNOUNCEMENT.md) and [API reference](https://theupdateframework.readthedocs.io/en/latest/api/api-reference.html) for more details.

Issue #958, opened 2019-11-13
`Repository.writeall()` does not "write all" and may leave repo in inconsistent state 


lukpueh:
**Description of issue or feature request**:

The [name and docstring](https://github.com/theupdateframework/tuf/blob/42c8d116d9f842a8eb111a8d7289ebf785fd9b88/tuf/repository_tool.py#L220-L226) of `Repository.writeall()`-classmethod and its [commented use in the tuf tutorial](https://github.com/theupdateframework/tuf/blame/42c8d116d9f842a8eb111a8d7289ebf785fd9b88/docs/TUTORIAL.md#L440-L444), suggest that it writes all the metadata that needs to be written upon a metadata change, when in practice it only writes metadata that is marked as dirty.

This may leave the repository metadata in an inconsistent state, as is the case in above linked tutorial part.

**Current behavior**:

`writeall` only writes metadata that is marked as dirty, ignoring metadata interdependencies, i.e. timestamp needs to be re-written if snapshot changes,  snapshot needs to be rewritten if targets or delegated targets changes, which may leave a repository in an inconsistent state.

**Expected behavior**:
Two options are conceivable:

- Update documentation but above all the tutorial to clarify that `writeall` might require a prior call to `mark_dirty` (I will add a quick fix in #775)

- Update `writeall` to indeed write all the metadata that is required to be written upon any metadata change (preferred solution).

JustinCappos:
I agree with the second option.  Let's make writeall actually write all!

lukpueh:
Thanks for you comment, @JustinCappos! Can we briefly together look at the metadata inter-dependencies to make sure that `writeall` will indeed write all?

As far as I can see the following inter-dependencies exist:
- a change of keys in a **delegating role**, i.e. root or any delegating targets role, requires to bump the version and renew the signature of the **delegated role**, whose key was changed:
  - (change root keys in **root** --> bump version and re-sign **root**)
  - change timestamp key in **root** --> bump version and re-sign **timestamp**
  - change snapshot key in **root** --> bump version and re-sign **snapshot**
  - change targets key in **root** --> bump version and re-sign **targets**
  - change delegated targets key in **delegating targets** --> bump version and re-sign **delegated targets**
- any change in **targets or delegated targets** (including above) requires to update the relevant targets version number (and optionally its length and hashes) in **snapshot**, bump snapshot's own version number, and renew snapshot's signature
[...]

PR #989, opened 2020-03-03
Use __name__ for loggers, per convention

joshuagl:
**Fixes issue #**: N/A

* As in secure-systems-lab/securesystemslib#212 replace hard-coded logger names with the conventional pattern `logging.getLogger(__name__)`.

requirements**:


tests/test_log.py
```diff
@@ -34,7 +34,10 @@
 
 from six.moves import reload_module
 
-
+# We explicitly create a logger which is a child of the tuf hierarchy,
+# instead of using the standard getLogger(__name__) pattern, because the
+# tests are not part of the tuf hierarchy and we are testing functionality
+# of the tuf package explicitly enabled on the tuf hierarchy
 logger = logging.getLogger('tuf.test_log')
```
joshuagl: We could avoid this if we move initialising the top-level logger into `tuf/__init__.py`, but as we will still want this module and its API to be able to configure the logger hierarchy it didn't seem particularly beneficial to make that change.

tuf/log.py
```diff
@@ -113,6 +113,9 @@
 file_handler = None
 
 # Set the logger and its settings.
+# Note: we're configuring the top-level hierarchy for the tuf package,
+# therefore we explicitly request the 'tuf' logger, rather than following
+# the standard pattern of logging.getLogger(__name__)
```
joshuagl, line 118: We could avoid this if we move initialising the top-level logger into `tuf/__init__.py`, but as we will still want this module and its API to be able to configure the logger hierarchy it didn't seem particularly beneficial to make that change.

lukpueh:
Hooray for passing AppVeyor tests (see #985). :)

tests/test_log.py
trishankatdatadog, line 41:
```suggestion
logger = logging.getLogger(__name__)
```

trishankatdatadog (changes requested):
Thanks for this much less brittle code, Joshua! 🎉

tests/test_log.py
joshuagl: Per the comment above I explicitly didn't change to `__name__` here because that gives us a logger names "test_log", which isn't part of the "tuf" logger hierarchy and therefore we can not test the logging functionality in `tuf.log` which only affects the "tuf" logger and its children.

trishankatdatadog: Got it, somehow missed that! Curious, is this necessary for the tests to work?

joshuagl: It is necessary for the tests to work.

Functions in `tuf.log` change logger configuration settings on the "tuf" logger and its children. Therefore if we create a logger in the test named "test_log" (via `getLogger(__name__)`) the `tuf.log` calls don't affect the logger we created (as it's not a member of the "tuf" logger hierarchy) and the tests fail:
```
FAIL: test_set_log_level (test_log.TestLog)
  File "/home/joshuagl/Projects/tuf/tests/test_log.py", line 62, in test_set_log_level
    self.assertTrue(logger.isEnabledFor(logging.DEBUG))
AssertionError: False is not true
```

woodruffw (approved):
Thanks for this! Should make logging control in the context of Warehouse a good deal easier 😄

tuf/repository_tool.py
```diff
@@ -220,7 +220,8 @@ def __init__(self, repository_directory, metadata_directory,
   def writeall(self, consistent_snapshot=False):
     """
     <Purpose>
-      Write all the JSON Metadata objects to their corresponding files.
+      Write all the JSON Metadata objects to their corresponding files for
+      roles which have changed.
```
lukpueh, line 224: Did this slip in by accident? I'm okay with leaving it there. The rest of the PR is very clear and this is a legitimate clarification of `writeall`'s purpose (see #958 and #964).

lukpueh (approved):
LGTM. Many thanks for taking on this dull but very valuable task! :)

tuf/repository_tool.py
joshuagl: Yes, it did. Whoops.

PR #991, opened 2020-03-05
More flexible file hashing support

joshuagl:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #**:

**Description of the changes being introduced by the pull request**:
* Don't use securesystemslib.settings.HASH_ALGORITHMS for file hashing – doing so is poor cohesion and overloading a setting meant for key hashing
* Add a new tuf.settings.FILE_HASHING_ALGORITHM for file hashing
* Use tuf.settings.FILE_HASHING_ALGORITHM for hashes in timestamp and targets metadata

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

Issue #468, opened 2017-07-10
Ensure Root's hashes and length are no longer listed in Snapshot


vladimir-v-diaz:
The specification was recently edited to longer list the hash and length of the Root file in Snapshot:
https://github.com/theupdateframework/tuf/pull/466

The reference implementation should be updated to incorporate the change.

vladimir-v-diaz:
Fixed in https://github.com/theupdateframework/tuf/pull/469.

Issue #848, opened 2019-04-04
Documenting or creating a TAP to describe keyid_hash_algorithms


erickt:
Hello again,

Would you consider documenting, or possibly writing a TAP, to describe the `keyid_hash_algorithms` field in a key definition? As I upgrade go-tuf (and eventually rust-tuf) to the TUF 1.0 Draft, I am also updating it to be compatible with the metadata produced and consumed by this project. In order to do so, I need to also produce a compatible `keyid_hash_algorithms` field the metadata I produce. However, I'm not completely sure what I am supposed to do with this field when I receive it from python-tuf.

It appears this field originally came from https://github.com/secure-systems-lab/securesystemslib/pull/37. As best as I can tell, it is present so that a client knows which algorithm was used to compute a keyid, so it can verify that a keyid is correct. Is this right, and are there other ways the `keyid_hash_algorithms` is used?

awwad:
((Background note for others: KeyIDs matter in TUF.  It is technically possible, I think, to construct a slower client for the reference implementation that could verify metadata without knowing how keyids were calculated, without substantial security issues.  You could do this by -- to take signature checking as example -- trying to verify a signature using all authorized keys rather than just the one with a matching keyid the signature lists....  I'm not recommending that.))

There's been some debate (#442 and #434, for example), about keyid_hash_algorithms and keyid calculation in general.  Personally, I don't think the keyid_hash_algorithms field is of use, and I would like to remove it from keys, as I think it's probably sensible to assume across an implementation and deployment that all keyid calculations employ a specific hash function.


As for your question, there are two answers:

1. The hypothetical feature keyid_hash_algorithms is supposed to provide is the ability to limit the client's calculation of possible keyids for a key to only the algorithms listed there, to reduce collision space for keys or exclude, say, outdated hash algorithms from keyid calculations for that key.  You could have keys that expect to only be hashed to produce keyids using SHA512 or something, I guess.  (But really, if a hash algorithm is unacceptable, it should be unacceptable for all keys in the implementation, IMHO.)

2. The current code in the reference implementation includes keyid_hash_algorithms in the keyid calculation, which means that:
   - changing the values in that field for a key changes the key's keyid
   - removing the field changes all keys' keyids, which is a bit of a backward compatibility headache


Sorry for the headache. ^_^  Does that clear things up enough to ensure compatibility?

awwad:
Oh, sorry, I didn't answer the TAP question:

[...]

PR #993, opened 2020-03-06
Support blake algorithms for file hashing

joshuagl:
**Fixes issue #**: N/A

* Don't use securesystemslib.settings.HASH_ALGORITHMS for file hashing – doing so is poor cohesion and overloading a setting meant for key hashing
* Add a new tuf.settings.FILE_HASHING_ALGORITHM for file hashing
* Use tuf.settings.FILE_HASHING_ALGORITHM for hashes in timestamp and targets metadata

requirements**:

### DIFF
--- a/tuf/settings.py
+++ b/tuf/settings.py
@@ -102,6 +102,9 @@ CONSISTENT_METHOD = 'copy'
 # the securesystemslib external library.
 DEFAULT_HASH_ALGORITHM = 'sha256'
 
+# The hashing algorithms used to compute file hashes
+FILE_HASH_ALGORITHMS = ['sha256', 'sha512']
+
 # The client's update procedure (contained within a while-loop) can potentially
 # hog the CPU.  The following setting can be used to force the update sequence
 # to suspend execution for a specified amount of time.  See
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -1531,7 +1531,8 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
 
   # Retrieve the versioninfo of the Snapshot metadata file.
   snapshot_fileinfo = {}
-  length, hashes = securesystemslib.util.get_file_details(snapshot_filename)
+  length, hashes = securesystemslib.util.get_file_details(snapshot_filename,
+      tuf.settings.FILE_HASH_ALGORITHMS)
   snapshot_version = get_metadata_versioninfo('snapshot', repository_name)
   snapshot_fileinfo[SNAPSHOT_FILENAME] = \
     tuf.formats.make_fileinfo(length, hashes, version=snapshot_version['version'])
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -974,7 +974,7 @@ def get_metadata_fileinfo(filename, custom=None):
   # file information, such as the file's author, version/revision
   # numbers, etc.
   filesize, filehashes = securesystemslib.util.get_file_details(filename,
-      securesystemslib.settings.HASH_ALGORITHMS)
+      tuf.settings.FILE_HASH_ALGORITHMS)
 
   return tuf.formats.make_fileinfo(filesize, filehashes, custom=custom)
 


joshuagl:
(I closed PR #991 because I didn't want that branch name to live forever in the git history, sorry for the noise)
cc @mnm678 who was assigned to review #991 by @JustinCappos 

*Note*: this change has been tested with the simple script in this GitHub gist: [test_blake.py](https://gist.github.com/joshuagl/bb2e5a372988bbc05be4a382bb4b9c59) which requires the corresponding change in secure-systems-lab/securesystemslib#218

@lukpueh shared a list of places to check for use of hashing algorithms:
* targets (for targets metadata) – addressed in this PR with 
1f0cb9c
* targets (for consistent snapshot prefixes) – uses the already computed hashes in the fileinfo dict
* targets (for hashed bins) – assigns the value of tuf.settings.DEFAULT_HASH_ALGORITHM to the module level constant `HASH_FUNCTION` and uses that for hashing
* targets metadata (for snapshot metadata) – I think this uses the already computed hashes in the fileinfo dict 
* snapshot metadata (for timestamp metadata) – addressed in this PR with 930d832
*  keys (for keyid_hash_algorithms) – uses `securesystemslib.settings.HASH_ALGORITHMS` and often includes an awkward dance where that setting is changed before calling `securesystemslib.keys.format_metadata_to_key()` and then reverted. This is a prime candidate for a refactoring, but doesn't preclude this PR being merged
* client updater  – needs work, for example `_get_target_hash()` hard-codes the use of 'sha256' hashing algorithm

In order to unblock @woodruffw's PEP 458 implementation could we land this and create an issue to address client updater in a follow-on PR?

lukpueh:
> In order to unblock @woodruffw's PEP 458 implementation could we land this and create an issue to address client updater in a follow-on PR?

@joshuagl: Have you already identified any issues in the client updater resulting from this PR?

joshuagl:
> > In order to unblock @woodruffw's PEP 458 implementation could we land this and create an issue to address client updater in a follow-on PR?
> 
> @joshuagl: Have you already identified any issues in the client updater resulting from this PR?

I have not. The only thing I've spotted so far is the related issue that `_get_target_hash()` hard-codes the hash algorithm to 'sha256', whereas I think it should be using `tuf.settings.DEFAULT_HASH_ALGORITHM`.

I have been able to make a simple modification to `tuf/scripts/repo.py` to create a repository which uses blake algorithms for filesystem hashes and successfully use `tuf/scripts/client.py` to fetch a target file from that repo.

tuf/settings.py
```diff
@@ -102,6 +102,9 @@
 # the securesystemslib external library.
 DEFAULT_HASH_ALGORITHM = 'sha256'
 
+# The hashing algorithms used to compute file hashes
+FILE_HASH_ALGORITHMS = ['sha256', 'sha512']
```
woodruffw, line 106: This doesn't apply directly to this PR, but could someone on the TUF team comment on best practices re: configuring `tuf.settings`?

After this PR, Warehouse will probably do something like this:

```python
import tuf.settings

# ...
tuf.settings.FILE_HASH_ALGORITHMS = ['blake2b']
```

is this the intended configuration pattern? It feels slightly wrong to modify a third-party module global like this, but it doesn't look like any other interfaces for configuring TUF are exposed.

mnm678:
This looks good as a short term fix, but in the long term I would prefer changing the requirements for keyid calculation as discussed here: https://github.com/theupdateframework/tuf/issues/848. As currently implemented, the client has to hard code the same hash algorithm for keyids as the metadata which does not leave much room for flexibility, especially because the keyid does not need to be globally unique.

joshuagl:
> This looks good as a short term fix

Thanks @mnm678. This is very much intended as a short-term fix to unblock PEP 458 implementation. @lukpueh mentioned #848 and some other issues around cohesion between securesystemslib and tuf when we discussed this quick fix. I'm absolutely interested in contributing to discussions and solutions towards long-term fixes too.

trishankatdatadog:
I agree with both of you. I think a good long-term fix might be specifying the targets hashing and keyid calculation (not necessarily hashing) algorithms in the `root` metadata?

mnm678:
I think an even simpler long term solution would be to keep the keyid calculation independent of the client. The client only needs to know that keyids are unique within a metadata file, not how they are determined. But we can have a meeting/discussion about this elsewhere.

tuf/settings.py
lukpueh: I agree it feels wrong, but I fear in some cases it is the only way to configure TUF. We should fix this at some point.

woodruffw (approved):

mnm678 (approved):

lukpueh:
Quick question, @joshuagl, regarding:
> targets (for consistent snapshot prefixes) – uses the already computed hashes in the fileinfo dict

Are you referring to the updater ... ?
https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/client/updater.py#L1375-L1377

(btw. the `pop` seems hazardous, but that's a different story)

I couldn't find any repo tooling that helps prefixing target files with their hash digests. 🤷‍♂

lukpueh:
> I couldn't find any repo tooling that helps prefixing target files with their hash digests. 🤷‍♂

Never mind, just found it:
https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_lib.py#L1319-L1324

lukpueh:
Regarding
> targets metadata (for snapshot metadata) – I think this uses the already computed hashes in the fileinfo dict

Rather looks like python-tuf does not include hashes (only version) in snapshot metadata. See [repository_lib.generate_snapshot_metadata](https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_lib.py#L1358).

lukpueh (approved):
LGTM.

joshuagl:
> Regarding
> 
> > targets metadata (for snapshot metadata) – I think this uses the already computed hashes in the fileinfo dict
> 
> Rather looks like python-tuf does not include hashes (only version) in snapshot metadata. See [repository_lib.generate_snapshot_metadata](https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_lib.py#L1358).

Seems like this could be worthy of an issue so that we can remember to add to python-tuf?

lukpueh:
> Seems like this could be worthy of an issue so that we can remember to add to python-tuf?

Good suggestion! Here we go https://github.com/theupdateframework/tuf/issues/996

For the interested reader: The fields were removed in #469 to fix #468 because of the spec changes in #466 motivated by #465. (I just love git and GitHub! :D)

Issue #1004, opened 2020-03-18
Take use_existing_fileinfo kwarg in write and writeall methods


lukpueh:
**Description of issue or feature request**:
Targets file hashes stored in targets metadata currently are generated ad-hoc when calling repository tool's `write` or `writeall` methods, requiring the corresponding targets file to be available at that time.

To make the tool more flexible (e.g. for PEP458) it should be possible to use previously (out-of-band) generated hashes that were already added to the roledb (see #1003).

**Current behavior**:
When calling repository tool's `write` or `writeall`, tuf generate hashes for all relevant target files in the roledb.

**Expected behavior**:
`write` and `writeall` should take an optional `use_existing_fileinfo` kwarg, which skips the hashing of target files in underlying functions (**) and instead uses the fileinfo in roledb (***), failing if not available.

(**) stacktrace is:
`write`/`writeall`  --> `repository_lib._generate_and_write_metadata` --> `repository_lib.generate_targets_metadata`

(***) Depends on roledb changes through #1003

lukpueh:
Fixed in #1007

Issue #1003, opened 2020-03-17
Take fileinfo kwarg in add_target and add_target_to_bin methods


lukpueh:
**Description of issue or feature request**:

The repository tool `add_target` and `add_target_to_bin` methods can be used to add target **paths** to the targets metadata in memory. The corresponding target file hashes are produced later e.g. on `write` or `writeall`.

It should be possible to add target paths using those functions, along with target file hashes that are generated out of band (this is required for PEP 458).

**Current behavior**:
`add_target` and `add_target_to_bin` don't take a `fileinfo` argument.

*Implementation hint*: `roledb` currently has a paths dictionary, where target paths are listed as dict keys and the optionally passed `custom` dict as value:

``` python
{"paths" : {"<TARGET PATH>": "custom": <CUSTOM DICT>}, ...}}
```

**Expected behavior**:

`add_target` and `add_target_to_bin` take an optional `fileinfo` argument that adheres to `tuf.formats.FILEINFO_SCHEMA`.

*Implementation hint*:  probably requires a change to `roledb` to store all `fileinfo` properties, e.g.:
``` python
{
"paths" : {
  "<TARGET PATH>": {
    "length": <LENGTH>,
    "hashes": <HASHES>,
    "version": <VERSION>,
    "custom": <CUSTOM DICT>
    }, ... 
  }
}
```

**Discussion**:

- `fileinfo` as returned by `tuf.formats.make_fileinfo` optionally includes the `custom` field (see above). It feels redundant to have `add_target` accept both `custom` and `fileinfo`. If we are fine with an interface change, I suggest to replace the former with the latter. Otherwise, we can have both, and fail if someone passes both `custom` and `fileinfo`

- Functions that take multiple target paths as list of string,  i.e. `add_targets`, `delegate`, `delegate_hashed_bins`, IMO don't need to provide this feature.

woodruffw:
To clarify this on the downstream side, this is *roughly* how I'd expect to consume this API (the `utils` calls are likely to change, that's just me scratchpadding):

```python
from warehouse.tuf import utils

@task(bind=True, ignore_result=True, acks_late=True)
def add_target(task, request, file):
    repository = utils.open_repository(request)
    fileinfo = utils.make_fileinfo(file)
    
    repository.add_target_to_bin(file.path, fileinfo=fileinfo)
    # signing, writing operations
```

lukpueh:
Fixed in #1007

Issue #995, opened 2020-03-09
`delegate_hashed_bins` must hash a target path as it appears in metadata


lukpueh:
**Description of issue or feature request**:
Fix in accordance with https://github.com/theupdateframework/tuf/issues/957 (and #963)!!

The repository tool's [`delegate_hashed_bins`](https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_tool.py#L2419) creates hash bins (delegated targets roles) and assigns passed target file paths based on their hash. Later it calls `delegate` for each bin passing the relevant paths. `delegate` then might left-strip the targets directory path from each targets path, changing the targets path in a way that it no longer is in the right bin.

**Current behavior**:
`delegate_hashed_bins` might hash target path based on a different path than the one that will appear in the (delegated) targets metadata.

**Expected behavior**:
`delegate_hashed_bins` must hash a target path as it appears in metadata.

lukpueh:
*On a related side-note:*

`delegated_hashed_bins` might generally benefit from some refactoring/clean up. E.g. it does not feel quite necessary to create a dictionary that has regularly incrementing indices as keys:
https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_tool.py#L2513-L2515

The refactor could be done while fixing above or separately.  Below is some semi-pseudo-code, that shows how I would distribute target paths in bins. Feel free to use as inspiration, enhancements/optimizations are welcome... :)

```python
def distribute_in_bins(target_paths, bin_count):
  prefix_length = len(f"{bin_count - 1:x}")
  prefix_count = 16 ** prefix_length
  bin_size = prefix_count // bin_count

  # Create targets object for each bin and assign it to a list, where the index
  # equals the lower bound of the hash prefix range served by the bin divided
  # by the bin size
  ordered_roles = []
  for idx in range(0, prefix_count, bin_size):
    name = f"{idx:0{prefix_length}x}-{idx+bin_size-1:0{prefix_length}x}"
    role = {"name": name, "target_paths": []}
    ordered_roles.append(role)

  # Use floor division to assign targets to their serving bin
  for target_path in target_paths:
    # hash_prefix = _get_hash(target_path)[:prefix_length]
    hash_prefix = target_path # <-- only for testing purpose (FIXME)
    ordered_roles[int(hash_prefix, 16) // bin_size]["target_paths"].append(target_path)

[...]

Issue #994, opened 2020-03-06
Refactor `add_target_to_bin` and `remove_target_from_bin`


lukpueh:
**Description of issue or feature request**:
The TUF repository tool provides a functionality to distribute a large number of target files over multiple delegated roles (see [`delegate_hashed_bins`](https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_tool.py#L2419)). 

To add a target to the corresponding bin, identified by the hash of the target path, or remove one from a bin the methods remove `add_target_to_bin` and `remove_target_from_bin` are provided. These methods are wrappers around `add_target` and `remove_target`, which don't call those methods directly but via a dispatcher helper `_locate_and_update_target_in_bin`, which also implements common functionality for the wrappers.

 These methods should be refactored for the reasons and as outlined below:

**Current behavior**:
-  The current architecture doesn't allow to pass through different arguments to the wrapped methods, e.g. the `custom` argument only supported by `add_target`:
https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_tool.py#L2672

- `_locate_and_update_target_in_bin` doesn't seem to be designed to dispatch to any other methods than `add_target` and `remove_target` to justify the dispatcher architecture

- `_locate_and_update_target_in_bin` replicates functionality from the wrapped methods, e.g. that the targets path exists as file in the current directory 
https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_tool.py#L2737-L2739
(see https://github.com/theupdateframework/tuf/issues/957 for reasons why that might not even be necessary).

- `_locate_and_update_target_in_bin` makes risky assumption, e.g. it infers the `hash_prefix_length` (i.e. `len(hex(number_of_bins - 1)[2:])`), by taking the length of the first element in the `path_hash_prefixes` field of the first delegation of the targets object on which it is called. 
https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_tool.py#L2719-L2733

- `_locate_and_update_target_in_bin` makes a `O(1)` search `O(n)`.

https://github.com/theupdateframework/tuf/blob/4fe29d138df02035eed9716657cfe0a76e17e178/tuf/repository_tool.py#L2749-L2756

**Expected behavior**:

[...]

PR #1007, opened 2020-03-26
Enhancements for hashed bin delegation

joshuagl:
**Fixes issue #**: #994, #995, #1003, #1004

__Note__: This is the first PR addressing changes required by the PEP 458 implementation.

* Make tests around hashed bins simpler and more robust
* Add various helper functions for common functionality around hashed bins:
  * `_get_hash()` to hash a filepath string
  * `_create_bin_name()` to create the string name of a delegated hash bin (mostly to consolidate the instances where we do non-trivial string formatting)
  * `_find_bin_for_hash()` to calculate the bin a hashed file string would be allocated to
* Refactor `add_target_to_bin()` and `remove_target_from_bin()` to remove the dispatcher and simplify the code as well as allowing `add_target_to_bin()` to take a custom argument, the same as `add_target()`
* Refactor `delegate_hashed_bins()` to simplify the code
* Ensure file paths are hashed as they appear in the targets metadata
* Enable passing a fileinfo to `add_target()` and `add_target_to_bin()`
* Add `use_existing_fileinfo` argument to `write()` and `writeall()` to allow writing metadata without access to targets files

requirements**:

cc @woodruffw

### DIFF
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1374,10 +1374,13 @@ class TestTargets(unittest.TestCase):
     def check_prefixes_match_range():
       roleinfo = tuf.roledb.get_roleinfo(self.targets_object.rolename,
           'test_repository')
+      have_prefixes = False
+
       for delegated_role in roleinfo['delegations']['roles']:
         if len(delegated_role['path_hash_prefixes']) > 0:
           rolename = delegated_role['name']
           prefixes = delegated_role['path_hash_prefixes']
+          have_prefixes = True
 
           if len(prefixes) > 1:
             prefix_range = "{}-{}".format(prefixes[0], prefixes[-1])
@@ -1386,6 +1389,8 @@ class TestTargets(unittest.TestCase):
 
           self.assertEqual(rolename, prefix_range)
 
+      # We expect at least one delegation with some path_hash_prefixes
+      self.assertTrue(have_prefixes)
 
 
     # Test delegate_hashed_bins() and verify that 16 hashed bins have
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -369,6 +369,108 @@ class TestRepository(unittest.TestCase):
     self.assertRaises(securesystemslib.exceptions.FormatError, repository.writeall, 3)
 
 
+  def test_writeall_no_files(self):
+    # Test writeall() when using pre-supplied fileinfo
+
+    repository_name = 'test_repository'
+    temporary_directory = tempfile.mkdtemp(dir=self.temporary_directory)
+    repository_directory = os.path.join(temporary_directory, 'repository')
+    targets_directory = os.path.join(repository_directory,
+                                     repo_tool.TARGETS_DIRECTORY_NAME)
+
+    repository = repo_tool.create_new_repository(repository_directory, repository_name)
+
+    # (1) Load the public and private keys of the top-level roles, and one
+    # delegated role.
+    keystore_directory = os.path.join('repository_data', 'keystore')
+
+    # Load the public keys.
+    root_pubkey_path = os.path.join(keystore_directory, 'root_key.pub')
+    targets_pubkey_path = os.path.join(keystore_directory, 'targets_key.pub')
+    snapshot_pubkey_path = os.path.join(keystore_directory, 'snapshot_key.pub')
+    timestamp_pubkey_path = os.path.join(keystore_directory, 'timestamp_key.pub')
+
+    root_pubkey = repo_tool.import_rsa_publickey_from_file(root_pubkey_path)
+    targets_pubkey = \
+      repo_tool.import_ed25519_publickey_from_file(targets_pubkey_path)
+    snapshot_pubkey = \
+      repo_tool.import_ed25519_publickey_from_file(snapshot_pubkey_path)
+    timestamp_pubkey = \
+      repo_tool.import_ed25519_publickey_from_file(timestamp_pubkey_path)
+
+    # Load the private keys.
+    root_privkey_path = os.path.join(keystore_directory, 'root_key')
+    targets_privkey_path = os.path.join(keystore_directory, 'targets_key')
+    snapshot_privkey_path = os.path.join(keystore_directory, 'snapshot_key')
+    timestamp_privkey_path = os.path.join(keystore_directory, 'timestamp_key')
+
+    root_privkey = \
+      repo_tool.import_rsa_privatekey_from_file(root_privkey_path, 'password')
+    targets_privkey = \
+      repo_tool.import_ed25519_privatekey_from_file(targets_privkey_path,
+                                                'password')
+    snapshot_privkey = \
+      repo_tool.import_ed25519_privatekey_from_file(snapshot_privkey_path,
+                                                'password')
+    timestamp_privkey = \
+      repo_tool.import_ed25519_privatekey_from_file(timestamp_privkey_path,
+                                                'password')
+
+
+    # (2) Add top-level verification keys.
+    repository.root.add_verification_key(root_pubkey)
+    repository.targets.add_verification_key(targets_pubkey)
+    repository.snapshot.add_verification_key(snapshot_pubkey)
+
+    # Verify that repository.writeall() fails for insufficient threshold
+    # of signatures (default threshold = 1).
+    self.assertRaises(tuf.exceptions.UnsignedMetadataError, repository.writeall)
+
+    repository.timestamp.add_verification_key(timestamp_pubkey)
+
+
+    # (3) Load top-level signing keys.
+    repository.status()
+    repository.root.load_signing_key(root_privkey)
+    repository.status()
+    repository.targets.load_signing_key(targets_privkey)
+    repository.status()
+    repository.snapshot.load_signing_key(snapshot_privkey)
+    repository.status()
+
+    # Verify that repository.writeall() fails for insufficient threshold
+    # of signatures (default threshold = 1).
+    self.assertRaises(tuf.exceptions.UnsignedMetadataError, repository.writeall)
+
+    repository.timestamp.load_signing_key(timestamp_privkey)
+
+    # Add target fileinfo
+    target1_hashes = {'sha256': 'c2986576f5fdfd43944e2b19e775453b96748ec4fe2638a6d2f32f1310967095'}
+    target2_hashes = {'sha256': '517c0ce943e7274a2431fa5751e17cfd5225accd23e479bfaad13007751e87ef'}
+    target1_fileinfo = tuf.formats.make_fileinfo(555, target1_hashes)
+    target2_fileinfo = tuf.formats.make_fileinfo(37, target2_hashes)
+    target1 = os.path.join(targets_directory, 'file1.txt')
+    target2 = os.path.join(targets_directory, 'file2.txt')
+    repository.targets.add_target(target1, fileinfo=target1_fileinfo)
+    repository.targets.add_target(target2, fileinfo=target2_fileinfo)
+
+    repository.writeall(use_existing_fileinfo=True)
+
+    # Verify that the expected metadata is written.
+    metadata_directory = os.path.join(repository_directory,
+                                      repo_tool.METADATA_STAGED_DIRECTORY_NAME)
+
+    for role in ['root.json', 'targets.json', 'snapshot.json', 'timestamp.json']:
+      role_filepath = os.path.join(metadata_directory, role)
+      role_signable = securesystemslib.util.load_json_file(role_filepath)
+
+      # Raise 'securesystemslib.exceptions.FormatError' if 'role_signable' is
+      # an invalid signable.
+      tuf.formats.check_signable_object_format(role_signable)
+
+      self.assertTrue(os.path.exists(role_filepath))
+
+
 
   def test_get_filepaths_in_directory(self):
     # Test normal case.
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1076,14 +1076,14 @@ class TestTargets(unittest.TestCase):
 
     self.assertEqual(len(self.targets_object.target_files), 2)
     self.assertTrue('file2.txt' in self.targets_object.target_files)
-    self.assertEqual(self.targets_object.target_files['file2.txt'],
+    self.assertEqual(self.targets_object.target_files['file2.txt']['custom'],
                      custom_file_permissions)
 
     # Attempt to replace target that has already been added.
     octal_file_permissions2 = oct(os.stat(target_filepath).st_mode)[4:]
     custom_file_permissions2 = {'file_permissions': octal_file_permissions}
     self.targets_object.add_target(target2_filepath, custom_file_permissions2)
-    self.assertEqual(self.targets_object.target_files['file2.txt'],
+    self.assertEqual(self.targets_object.target_files['file2.txt']['custom'],
     custom_file_permissions2)
 
     # Test improperly formatted arguments.
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1094,18 +1094,6 @@ class TestTargets(unittest.TestCase):
                       target_filepath, 3)
 
 
-    # Test invalid filepath argument (i.e., non-existent or invalid file.)
-    self.assertRaises(securesystemslib.exceptions.Error, self.targets_object.add_target,
-                      'non-existent.txt')
-
-    # Not under the repository's targets directory.
-    self.assertRaises(securesystemslib.exceptions.Error, self.targets_object.add_target,
-                      self.temporary_directory)
-
-    # Not a file (i.e., a valid path, but a directory.)
-    test_directory = os.path.join(self.targets_directory, 'test_directory')
-    os.mkdir(test_directory)
-    self.assertRaises(securesystemslib.exceptions.Error, self.targets_object.add_target, test_directory)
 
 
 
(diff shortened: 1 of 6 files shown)


tuf/repository_tool.py
```diff
@@ -2506,89 +2502,47 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
     logger.info(repr(len(list_of_targets)) + ' total targets.')
```
woodruffw: Nitpick: Could these separate `info` calls be merged into a single message? That would eliminate the chance of them being spread when integrated into a logging system with multiple parallel sources.

tuf/repository_tool.py
```diff
@@ -1953,31 +1961,28 @@ def add_target(self, filepath, custom=None):
...
-      # targets directory.
-      targets_directory_length = len(self._targets_directory) + 1
-      roleinfo = tuf.roledb.get_roleinfo(self._rolename, self._repository_name)
-      relative_path = filepath[targets_directory_length:].replace('\\', '/')
+    # Update the role's 'tuf.roledb.py' entry and avoid duplicates.  Make
+    # sure to exclude the path separator when calculating the length of the
+    # targets directory.
+    targets_directory_length = len(self._targets_directory) + 1
+    roleinfo = tuf.roledb.get_roleinfo(self._rolename, self._repository_name)
+    relative_path = filepath[targets_directory_length:].replace('\\', '/')
```
woodruffw, line 1984: Nit (can probably be ignored in this PR): It might be good to add (or use) a path canonicalization method instead of calling `replace('\\', '/')` throughout the codebase.

joshuagl: Agreed, this would be good cleanup.

@lukpueh any thoughts on whether this belongs in securesystemslib or in tuf? Perhaps as part of a new utils module?

tuf/repository_tool.py
joshuagl: Good idea, I've pushed that in a follow-on patch ed93fcc

joshuagl:
I've pushed a couple of extra patches to this branch, one to address a review nit from @woodruffw  and another to address #1004

tuf/repository_tool.py
lukpueh: Absolutely! https://github.com/theupdateframework/tuf/pull/1008#pullrequestreview-382660018 has some thoughts about path normalization. The TLDR is, if the caller passes a *bad* path,
- either fail,
- or warn and normalize

In both cases this should happen once at the boundary where a path is added to the system, and thus remove the need for many `replace('\\', '/')` throughout the code.

tests/test_repository_tool.py
```diff
@@ -1288,6 +1288,27 @@ def test_delegate_hashed_bins(self):
...
 
+
+    # A helper function to check that the range of prefixes the role is
+    # delegated for, specified in path_hash_prefixes, matches the range
+    # implied by the bin, or delegation role, name.
+    def check_prefixes_match_range():
+      roleinfo = tuf.roledb.get_roleinfo(self.targets_object.rolename,
+          'test_repository')
+      for delegated_role in roleinfo['delegations']['roles']:
+        if len(delegated_role['path_hash_prefixes']) > 0:
```
lukpueh, line 1380: Minor nit: This is a great check, it would a shame if it were silently omitted because of a hiccup in the roledb that makes so that we only get empty `delegated_role['path_hash_prefixes']` lists. Maybe worth a tiny TODO: comment?

tuf/repository_tool.py
```diff
@@ -2824,6 +2824,59 @@ def _get_hash(target_filepath):
...
+      The number of hashed_bins in use
+
+  <Returns>
+    The name of the hashed bin path_hash would be binned into.
+  """
+
+  prefix_len = len("{:x}".format(number_of_bins - 1))
+  prefix_count = 16 ** prefix_len
+
+  if prefix_count % number_of_bins != 0:
```
lukpueh: Minor nit: Maybe it's worth a comment that `x % y != 0` does not guarantee  that `y` is not a power of two for any `x` and `y`, but due to the relationship between `number_of_bins` and `prefix_count` it is true for them.

tuf/repository_tool.py
```diff
@@ -2669,107 +2693,20 @@ def remove_target_from_bin(self, target_filepath):
...
 
-    # Set 'delegation' if this Targets role has performed any delegations.
-    if len(roleinfo['delegations']['roles']):
-      delegation = roleinfo['delegations']['roles'][0]
+    path_hash = _get_hash(target_filepath)
+    bin_name = _find_bin_for_hash(path_hash, number_of_bins)
 
-    else:
+    # Ensure the Targets object has delegated to hashed bins
+    if not self._delegated_roles.get(bin_name, None):
```
lukpueh, line 2657: Would you mind also fixing the error message to what you use above? 
```
      raise securesystemslib.exceptions.Error(self.rolename + ' does not have'
          ' a delegated role ' + bin_name)
```

tuf/repository_tool.py
```diff
@@ -2486,14 +2486,13 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
...
-    # representation of hexadecimal values (so that they are not used in the
-    # calculation of the prefix length.) Example: number_of_bins = 32,
+    # 'number_of_bins'. Example: number_of_bins = 32,
     # total_hash_prefixes = 256, and each hashed bin is responsible for 8 hash
     # prefixes.  Hashed bin roles created = 00-07.json, 08-0f.json, ...,
     # f8-ff.json.
-    prefix_length =  len(hex(number_of_bins - 1)[2:])
+    prefix_length = len("{:x}".format(number_of_bins - 1))
     total_hash_prefixes = 16 ** prefix_length
+    bin_size = total_hash_prefixes // number_of_bins
```
lukpueh: Minor nit: Maybe above three lines together with the "power of two" check below are worth a helper, to be used here and in `_find_bin_for_hash()`.

tuf/repository_tool.py
```diff
@@ -2506,84 +2505,41 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
...
 
-    # Store the target paths that fall into each bin.  The digest of the target
-    # path, reduced to the first 'prefix_length' hex digits, is calculated to
-    # determine which 'bin_index' it should go.  we use xrange() here because
-    # there can be a large number of prefixes to process.
-    target_paths_in_bin = {}
-    for bin_index in six.moves.xrange(total_hash_prefixes):
-      target_paths_in_bin[bin_index] = []
+      role = {"name": name, "target_paths": [],
+          'target_hash_prefixes': target_hash_prefixes}
```
lukpueh: Minor nit: I think dict literals are easier to read when each item is on a separate line.

tuf/repository_tool.py
```diff
@@ -2551,10 +2566,17 @@ def add_target_to_bin(self, target_filepath, number_of_bins, custom=None):
       custom:
         An optional object providing additional information about the file.
 
+      fileinfo:
+        An optional fileinfo object, conforming to tuf.formats.FILEINFO_SCHEMA,
+        providing full information about the file.
+
```
lukpueh, line 2567: Given that we said that the only reason to have `custom` and `fileinfo`  is to support backwards compatibility (as the former can be passed within the latter) and `add_target_to_bin` didn't support this before, I think it is safe to not add `custom` here.

lukpueh: I understand that my suggestion comes at the cost of consistency between `add_target` and `add_target_to_bin`. I am fine either way.

tuf/repository_lib.py
```diff
@@ -1284,41 +1304,22 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
-    relative_targetpath = target
+      # Ensure all fileinfo entries in target_files have a non-empty hashes dict
+      if not fileinfo.get('hashes', None):
+        raise securesystemslib.exceptions.Error('use_existing_hashes option set'
+            ' but no hashes exist in roledb for ' + repr(target))
 
-    # Note: join() discards 'targets_directory' if 'target' contains a leading
-    # path separator (i.e., is treated as an absolute path).
-    target_path = os.path.join(targets_directory, target.lstrip(os.sep))
+      filedict[target.replace('\\', '/').lstrip('/')] = fileinfo
```
lukpueh, line 1322: Unrelated: I think this should have happened before (see https://github.com/theupdateframework/tuf/pull/1007#discussion_r399222066)

tuf/repository_lib.py
```diff
@@ -1284,41 +1304,22 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
-  for target, fileinfo in six.iteritems(target_files):
+  if use_existing_fileinfo:
+    for target, fileinfo in six.iteritems(target_files):
 
-    # The root-most folder of the targets directory should not be included in
-    # target paths listed in targets metadata.
-    # (e.g., 'targets/more_targets/somefile.txt' -> 'more_targets/somefile.txt')
-    relative_targetpath = target
+      # Ensure all fileinfo entries in target_files have a non-empty hashes dict
+      if not fileinfo.get('hashes', None):
```
lukpueh, line 1314: Important: This should also fail if 'length' is missing.

tuf/repository_lib.py
```diff
@@ -1352,6 +1353,58 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
+    # (e.g., 'targets/more_targets/somefile.txt' -> 'more_targets/somefile.txt')
+    relative_targetpath = target
+
+    # Note: join() discards 'targets_directory' if 'target' contains a leading
+    # path separator (i.e., is treated as an absolute path).
+    target_path = os.path.join(targets_directory, target.lstrip(os.sep))
+
+    # Ensure all target files listed in 'target_files' exist.  If just one of
+    # these files does not exist, raise an exception.
+    if not os.path.exists(target_path):
```
lukpueh, line 1387: Absolutely unrelated: This is the first of three `path.exists`/ `path.isfile` checks in the stack before we open that file (`get_metadata_fileinfo` has one and `securesystemslib.util.get_file_details` has one too). As good Pythonistas, we should ask for forgiveness instead.

lukpueh (changes requested):
This is impeccable work, @joshuagl. I only have one blocking comment, which is amazing for a PR of that size. That comment starts with "Important:". All the other comments are just sugar, so that you see I actually reviewed this. :P 

Many many thanks!

tests/test_repository_tool.py
joshuagl: Great point @lukpueh. 

When looking at the patch just now I realised that we only ever call this function when we know we have at least one delegation with `path_hash_prefixes`, so I'm going to add a flag which is set when we find a role with `len(delegated_role['path_hash_prefixes']) > 0` and assert that the flag is set at the end of the `for` loop. That was we should catch a hiccup in the roledb like you described.

joshuagl: Added the change I mentioned above as a follow-on patch f362544

tuf/repository_tool.py
joshuagl: That's a good point, I've updated this patch to include a comment

tuf/repository_tool.py
joshuagl: Good catch, thanks. I've updated the patch to make this consistent.

tuf/repository_tool.py
joshuagl: I've squashed in a change to make each item on a separate line _and_ use consistent quotation marks.

tuf/repository_tool.py
joshuagl: I agree. As we plan to deprecate the `custom` parameter it doesn't make sense to _add_ it to `add_target_to_bin`. I've added a follow-on patch in 5ea1f63, though maybe I should have squashed it.

tuf/repository_lib.py
joshuagl: Great catch, thank you. I've squashed a change to implement this into the patch.

tuf/repository_tool.py
joshuagl: The rest of the code uses all three of the variables (though one only in a comment iirc) and I'm not sure about adding a function that returns multiple values, so I've left this suggestion for now.

tuf/repository_lib.py
joshuagl: Hopefully it's OK to remove this as we consolidate path handling under #1008, or as a follow on to those changes?

tuf/repository_lib.py
joshuagl: We should absolutely move towards the Python idiom of [EAFP](https://docs.python.org/3/glossary.html#term-eafp). I think we can do that in future changes as we work to address #1009

joshuagl:
Thank you for the thorough review @lukpueh . I've addressed most of your comments, some in squashed commits and some in follow-on patches.

I have replied to all of your comments so that you can see where I've made changes.

tuf/repository_tool.py
trishankatdatadog: There should also be a comment that this particular LoC is _rescaling_ the index. I didn't really understand what was going on until I looked above to see what `number_of_bins` is.

lukpueh: I'm fine with leaving it as it is, although it is the same code  (with different names and order) in both cases.

`delegate_hashed_bins`:
https://github.com/theupdateframework/tuf/blob/5ea1f6398a27047a5a299c42b7428c6248d45e0e/tuf/repository_tool.py#L2504-L2513

`_find_bin_for_hash`:
https://github.com/theupdateframework/tuf/blob/5ea1f6398a27047a5a299c42b7428c6248d45e0e/tuf/repository_tool.py#L2768-L2777

And I personally see no problems with idioms like 

```python
pefix_length, prefix_count, bin_size = _get_bin_numbers(number_of_bins)
```

Also, and this was part of my motivation, we'd only need one place to properly comment on this math-magic (thanks for pointing that out, @trishankatdatadog).

Btw., @trishankatdatadog, I wouldn't say that this particular LoC rescales the index. Did you maybe mean this one...?
https://github.com/theupdateframework/tuf/blob/5ea1f6398a27047a5a299c42b7428c6248d45e0e/tuf/repository_tool.py#L2547

Yes, the general idea here is that we scale `0..total_hash_prefixes` to `0..number_of_bins`. But I wouldn't say that the particular LoC pointed out by you does this.

joshuagl: Convincing points. I will factor out this into a helper now.

joshuagl: Helper added, with lots of comments, in ec68416

trishankatdatadog: @lukpueh Yes, I meant that line, the one with the integer division

tests/test_repository_tool.py
```diff
@@ -369,6 +369,108 @@ def test_writeall(self):
     self.assertRaises(securesystemslib.exceptions.FormatError, repository.writeall, 3)
 
 
+  def test_writeall_no_files(self):
```
woodruffw, line 372: Just wanted to give kudos for adding a test case that closely mirrors the current work on Warehouse ❤️

tuf/repository_lib.py
woodruffw: Under #1008 or a follow on is okay for me!

tuf/repository_lib.py
```diff
@@ -1284,43 +1304,26 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
+        raise securesystemslib.exceptions.Error('use_existing_hashes option set'
+            ' but no hashes exist in roledb for ' + repr(target))
 
-    # Add 'custom' if it has been provided.  Custom data about the target is
-    # optional and will only be included in metadata (i.e., a 'custom' field in
-    # the target's fileinfo dictionary) if specified here.
-    custom_data = None
-    if len(custom):
-      custom_data = custom
+      if fileinfo.get('length', -1) < 0:
```
lukpueh, line 1318: Cool way of checking this. 👍

tuf/repository_tool.py
```diff
@@ -2626,12 +2586,24 @@ def add_target_to_bin(self, target_filepath):
...
+    path_hash = _get_hash(target_filepath)
+    bin_name = _find_bin_for_hash(path_hash, number_of_bins)
+
+    # Ensure the Targets object has delegated to hashed bins
+    if not self._delegated_roles.get(bin_name, None):
+      raise securesystemslib.exceptions.Error(self.rolename + ' does not have'
+          ' a delegated role ' + bin_name)
 
-    return self._locate_and_update_target_in_bin(target_filepath, 'add_target')
+    self._delegated_roles[bin_name].add_target(target_filepath, fileinfo)
```
lukpueh:
```suggestion
    self._delegated_roles[bin_name].add_target(target_filepath, fileinfo=fileinfo)
```
... otherwise `fileinfo` is passed as `custom` to `add_target`.

It would be great to have a test for `add_target_to_bin` that adds a `fileinfo`.

tuf/repository_tool.py
```diff
@@ -2669,135 +2645,150 @@ def remove_target_from_bin(self, target_filepath):
...
+
+    # TODO: check target_filepath is sane?
+
+    path_hash = _get_hash(target_filepath)
+    bin_name = _find_bin_for_hash(path_hash, number_of_bins)
+
+    # Ensure the Targets object has delegated to hashed bins
+    if not self._delegated_roles.get(bin_name, None):
+      raise securesystemslib.exceptions.Error(self.rolename + ' does not have'
+          ' a delegated roles ' + bin_name)
```
lukpueh:
```suggestion
          ' a delegated role ' + bin_name)
```

tuf/repository_tool.py
```diff
@@ -2669,135 +2645,150 @@ def remove_target_from_bin(self, target_filepath):
...
+  # hexadecimal digits needed by each hash prefix
+  prefix_length = len("{:x}".format(number_of_bins - 1))
+  # Calculate the total number of hash prefixes (e.g., 000 - FFF total values)
+  prefix_count = 16 ** prefix_length
+  # Determine how many prefixes to assign to each bin
+  bin_size = prefix_count // number_of_bins
+
+  # For simplicity, ensure that 'prefix_count' (16 ^ n) can be evenly
+  # distributed over 'number_of_bins' (must be 2 ^ n).  Each bin will contain
+  # (prefix_count / number_of_bins) hash prefixes.
```
lukpueh, line 2757: Ping https://github.com/theupdateframework/tuf/pull/1007#discussion_r399269492

lukpueh (changes requested):
Thanks for the updates! ❤️ Here are some more comments.

tuf/repository_lib.py
joshuagl: Feels a bit C-like, but I couldn't think of a cleaner way to do it.

tuf/repository_tool.py
joshuagl: Great catch Lukas. I've squashed this fix and added a test in 62e4364

tuf/repository_tool.py
joshuagl: Thanks for catching this. My original comment was lost when factoring out into the helper. I've updated that patch to add a comment as suggested.

lukpueh (approved):
🚀

Issue #963, opened 2019-11-29
Conflicting file path assertions in `delegate()` and `add_paths()`


lukpueh:
**Description of issue or feature request**:

Related to #957 

The `Targets.delegate()`method fails if one of the file paths or patterns passed via the `paths` argument is absolute, i.e. starts with a directory separator.  At the same time it warns if it the path does not have `self._targets_directory` as prefix, which very well may be an absolute path. 

https://github.com/theupdateframework/tuf/blob/409eef1a4872a88b610e81b375f36b07a8e0f996/tuf/repository_tool.py#L2286-L2294

The `Targets.add_paths()` methods emits a similar message (with debug level). Note that the method does not check whether the passed paths have a leading directory separator, although it should, in order to be consistent with above function.

https://github.com/theupdateframework/tuf/blob/409eef1a4872a88b610e81b375f36b07a8e0f996/tuf/repository_tool.py#L1868-L1872

**Current behavior**:
- `Targets.delegate()` and `Targets.add_paths()` emit warning/debug message if a passed path or path pattern is not prefixed with the targets directory base path, i.e. `self._targets_directory`.
- `Targets.add_paths` does not check if the passed paths are indeed relative.

**Expected behavior**:

Either,
- `Targets.delegate()` and `Targets.add_paths()` should not care about the prefix of the passed paths or path patterns.
- `Targets.add_paths` should probably assert that the passed paths are relative, to be consistent with the sibling the corresponding functionality of `delegate()`.

lukpueh:
When fixing this issue, don't forget to remove the comment in a tutorial code snippet that asks the reader to ignore a misleading warning about the location of a delegation path pattern (see 41d7bbde5dbeabb938b1cb90f4c7bc20bfdee725 of #775).

joshuagl:
The specification states:
> A TARGETPATH is a path to a file that is relative to a mirror's base URL of targets. To avoid surprising behavior when resolving paths, it is RECOMMENDED that a TARGETPATH uses the forward slash (/) as directory separator and does not start with a directory separator. The recommendation for TARGETPATH aligns with the "path-relative-URL string" definition in the WHATWG URL specification.

Therefore I think the latter expected behaviour is most consistent with the spec, i.e. we should change such that:
> `Targets.add_paths` should probably assert that the passed paths are relative, to be consistent with the sibling the corresponding functionality of delegate().

lukpueh:
[...]

Issue #957, opened 2019-11-12
Targets' `add_target(s)` methods do not correctly check passed paths


lukpueh:
**Description of issue or feature request**:

Judging from the docstrings and comments, `add_target` and `add_targets` are expected to update a `Targets` object with passed target paths, if they are relative to a targets directory base path (i.e. `self._targets_directory`).

However, they don't seem to take into account that Python's [`os.path.join`](https://docs.python.org/3/library/os.path.html#os.path.join) ignores all path component arguments prior to a component that looks like an absolute path. 

**Current behavior**:
* Works as expected (?) if passed target path is relative to a targets directory base path or absolute but falls into the targets directory:
```python
# repo._targets_directory == "/path/to/repo/targets"
repo.targets.add_target("foo.txt")
repo.targets.add_target("/path/to/repo/targets/bar.txt)
# Successfully adds "foo.txt" and "bar.txt" 
```
* Fails if the passed target path is not absolute and does not exist relative to  targets directory base path "/path/to/repo/targets".

```python
# repo._targets_directory == "/path/to/repo/targets"
repo.targets.add_target("repo/targets/foo.txt")
# Fails with "'/path/to/repo/targets/repo/targets/foo.txt' is not a valid file in the repository's targets directory"
```
* Fails if the passed target path is absolute and does not exist
```python
#repo._targets_directory == "/path/to/repo/targets"
repo.targets.add_target("/path/does/not/exist/foo.txt")
# Fails with "'/path/does/not/exist/foo.txt' is not a valid file in the repository's targets directory"
```
* Has unexpected behavior if the passed path is absolute and exists!

**Expected behavior**:
Revise intended behavior, update implementation accordingly.

lukpueh:
IIUC target files don't get hashed right away, i.e. on `add_target()`. This happens when generating and writing targets metadata to disk, e.g. via `writeall()`. So why not completely ignore the notion of files on disk until then?

`add_target` could just add any paths passed to the internal representation of targets metadata, checking that they are relative (and maybe don't contain ".."?).  And only later, when the corresponding files are actually needed, i.e. to generate hashes, should it fail gracefully if any of the paths can't be resolved into a file relative to the targets directory base path.

joshuagl:
A possible disadvantage of this proposal may be that it's harder to debug if the errors come at a seemingly unrelated call site. Path related failures when you add those paths in `add_target()` feels a lot more intuitive than path related failures at a later time.

[...]

PR #1008, opened 2020-03-26
Adopt a consistent behavior when adding targets and paths

sechkova:
**Fixes issue #** : #957, #963 

As described in the issues above, the methods of Targets class use inconsistent checks when adding target files and delegated paths. As a result of this plus the need to avoid access to the file system in certain cases, this pull request suggests:

- Adopt the strict behavior of accepting only relative paths (all absolute paths raise an error, even existing ones)
- Assume that relative paths are relative to the targets directory and targets directory is not included in the path/target name
- Perform the checks only on the path/target name, without accessing the file system
- If the added target file does not exist relative to the targets dir, an error is raised at a later stage, during the hash calculation in `write()` or `writeall()`
- If the path/pattern added to a delegated role does not exist relative to the targets dir, target file won't be matched during a client update (same as current behavior) 

Code changes:

- Add `_check_relpath()` method to Targets class 
- Modify all methods adding paths (`add_paths()`, `delegate()`) to utilise `_check_relpath()`
- Modify all methods adding targets (`add_target(s)()`, `delegate()`, `delegate_hashed_bins()`, `_locate_and_update_target_in_bin`) to utilise `_check_relpath()`

**Note**: Tests, delegation section in tutorial as well as `get_filepaths_in_directory()` helper also need updating but I am sharing the PR sooner as it includes changes needed by PEP 458 implementation. I will be updating them accordingly either in the PR or as a follow on it. 

requirements**:

### DIFF
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -352,15 +352,8 @@ the target filepaths to metadata.
 # in targets metadata.
 >>> repository = load_repository('repository')
 
-# get_filepaths_in_directory() returns a list of file paths in a directory. It
-# can also return files in sub-directories if 'recursive_walk' is True.
->>> list_of_targets = repository.get_filepaths_in_directory(
-...     "repository/targets/", recursive_walk=False, followlinks=True)
-
-# Note: Since we set the 'recursive_walk' argument to false, the 'myproject'
-# sub-directory is excluded from 'list_of_targets'.
->>> list_of_targets
-['/path/to/repository/targets/file2.txt', '/path/to/repository/targets/file1.txt', '/path/to/repository/targets/file3.txt']
+# Create a list of all targets in the directory.
+>>> list_of_targets = ['file1.txt', 'file2.txt', 'file3.txt']
 
 # Add the list of target paths to the metadata of the top-level Targets role.
 # Any target file paths that might already exist are NOT replaced, and
@@ -376,10 +369,12 @@ the target filepaths to metadata.
 # (octal number specifying file access for owner, group, others e.g., 0755) is
 # added alongside the default fileinfo.  All target objects in metadata include
 # the target's filepath, hash, and length.
+# Note: target path passed to add_target() method has to be relative
+# to the targets directory or an exception is raised.
 >>> target4_filepath = os.path.abspath("repository/targets/myproject/file4.txt")
 >>> octal_file_permissions = oct(os.stat(target4_filepath).st_mode)[4:]
 >>> custom_file_permissions = {'file_permissions': octal_file_permissions}
->>> repository.targets.add_target(target4_filepath, custom_file_permissions)
+>>> repository.targets.add_target('myproject/file4.txt', custom_file_permissions)
 ```
 
 The private keys of roles affected by the changes above must now be imported and
@@ -498,7 +493,6 @@ targets and generate signed metadata.
 
 # Make a delegation (delegate trust of 'myproject/*.txt' files) from "targets"
 # to "unclaimed", where "unclaimed" initially contains zero targets.
-# NOTE: Please ignore the warning about the path pattern's location (see #963)
 >>> repository.targets.delegate('unclaimed', [public_unclaimed_key], ['myproject/*.txt'])
 
 # Thereafter, we can access the delegated role by its name to e.g. add target
@@ -635,8 +629,7 @@ to some role.
 >>> repository.targets('unclaimed').remove_target("myproject/file4.txt")
 
 # Get a list of target paths for the hashed bins.
->>> targets = repository.get_filepaths_in_directory(
-...     'repository/targets/myproject', recursive_walk=True)
+>>> targets = ['myproject/file4.txt']
 
 # Delegate trust to 32 hashed bin roles. Each role is responsible for the set
 # of target files, determined by the path hash prefix. TUF evenly distributes
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -371,10 +371,11 @@ the target filepaths to metadata.
 # the target's filepath, hash, and length.
 # Note: target path passed to add_target() method has to be relative
 # to the targets directory or an exception is raised.
->>> target4_filepath = os.path.abspath("repository/targets/myproject/file4.txt")
->>> octal_file_permissions = oct(os.stat(target4_filepath).st_mode)[4:]
+>>> target4_filepath = 'myproject/file4.txt'
+>>> target4_abspath = os.path.abspath(os.path.join('repository', 'targets', target4_filepath))
+>>> octal_file_permissions = oct(os.stat(target4_abspath).st_mode)[4:]
 >>> custom_file_permissions = {'file_permissions': octal_file_permissions}
->>> repository.targets.add_target('myproject/file4.txt', custom_file_permissions)
+>>> repository.targets.add_target(target4_filepath, custom_file_permissions)
 ```
 
 The private keys of roles affected by the changes above must now be imported and
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1452,13 +1452,18 @@ class TestTargets(unittest.TestCase):
                       list_of_targets, public_keys, number_of_bins=3)
 
     # Invalid 'list_of_targets'.
-    # TODO
-    """
-    invalid_targets = ['/non-existent']
-    self.assertRaises(securesystemslib.exceptions.Error,
+    # A path or target starting with a directory separator
+    self.assertRaises(tuf.exceptions.InvalidNameError,
                       self.targets_object.delegate_hashed_bins,
-                      invalid_targets, public_keys, number_of_bins=16)
-    """
+                      ['/file1.txt'], public_keys,
+                      number_of_bins=2)
+
+    # A path or target using '\' as a directory separator
+    self.assertRaises(tuf.exceptions.InvalidNameError,
+                      self.targets_object.delegate_hashed_bins,
+                      ['subpath\\file1.txt'], public_keys,
+                      number_of_bins=2)
+
 
   def test_add_target_to_bin(self):
     # Test normal case.
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -213,14 +213,14 @@ class TestRepository(unittest.TestCase):
 
 
     # (4) Add target files.
-    target1 = os.path.join(targets_directory, 'file1.txt')
-    target2 = os.path.join(targets_directory, 'file2.txt')
-    target3 = os.path.join(targets_directory, 'file3.txt')
+    target1 = 'file1.txt'
+    target2 = 'file2.txt'
+    target3 = 'file3.txt'
     repository.targets.add_target(target1)
     repository.targets.add_target(target2)
 
     # (5) Perform delegation.
-    repository.targets.delegate('role1', [role1_pubkey], [os.path.basename(target3)])
+    repository.targets.delegate('role1', [role1_pubkey], [target3])
     repository.targets('role1').load_signing_key(role1_privkey)
 
     # (6) Write repository.
@@ -449,8 +449,8 @@ class TestRepository(unittest.TestCase):
     target2_hashes = {'sha256': '517c0ce943e7274a2431fa5751e17cfd5225accd23e479bfaad13007751e87ef'}
     target1_fileinfo = tuf.formats.make_fileinfo(555, target1_hashes)
     target2_fileinfo = tuf.formats.make_fileinfo(37, target2_hashes)
-    target1 = os.path.join(targets_directory, 'file1.txt')
-    target2 = os.path.join(targets_directory, 'file2.txt')
+    target1 = 'file1.txt'
+    target2 = 'file2.txt'
     repository.targets.add_target(target1, fileinfo=target1_fileinfo)
     repository.targets.add_target(target2, fileinfo=target2_fileinfo)
 
@@ -1064,11 +1064,10 @@ class TestTargets(unittest.TestCase):
     keystore_directory = os.path.join('repository_data', 'keystore')
     public_keypath = os.path.join(keystore_directory, 'snapshot_key.pub')
     public_key = repo_tool.import_ed25519_publickey_from_file(public_keypath)
-    target1_filepath = os.path.join(self.targets_directory, 'file1.txt')
 
     # Create Targets() object to be tested.
     targets_object = repo_tool.Targets(self.targets_directory)
-    targets_object.delegate('role1', [public_key], [os.path.basename(target1_filepath)])
+    targets_object.delegate('role1', [public_key], ['file1.txt'])
 
     self.assertTrue(isinstance(targets_object('role1'), repo_tool.Targets))
 
@@ -1087,18 +1086,16 @@ class TestTargets(unittest.TestCase):
     keystore_directory = os.path.join('repository_data', 'keystore')
     public_keypath = os.path.join(keystore_directory, 'snapshot_key.pub')
     public_key = repo_tool.import_ed25519_publickey_from_file(public_keypath)
-    target1_filepath = os.path.join(self.targets_directory, 'file1.txt')
-    target2_filepath = os.path.join(self.targets_directory, 'file2.txt')
 
     # Set needed arguments by delegate().
     public_keys = [public_key]
     threshold = 1
 
     self.targets_object.delegate('tuf', public_keys, [], threshold, False,
-        [target1_filepath], path_hash_prefixes=None)
+        ['file1.txt'], path_hash_prefixes=None)
 
     self.targets_object.delegate('warehouse', public_keys, [], threshold, False,
-        [target2_filepath], path_hash_prefixes=None)
+        ['file2.txt'], path_hash_prefixes=None)
 
     # Test that get_delegated_rolenames returns the expected delegations.
     expected_delegated_rolenames = ['targets/tuf/', 'targets/warehouse']
@@ -1112,11 +1109,11 @@ class TestTargets(unittest.TestCase):
     # Verify the targets object initially contains zero target files.
     self.assertEqual(self.targets_object.target_files, {})
 
-    target_filepath = os.path.join(self.targets_directory, 'file1.txt')
+    target_filepath = 'file1.txt'
     self.targets_object.add_target(target_filepath)
 
     self.assertEqual(len(self.targets_object.target_files), 1)
-    self.assertTrue('file1.txt' in self.targets_object.target_files)
+    self.assertTrue(target_filepath in self.targets_object.target_files)
 
 
 
@@ -1127,12 +1124,11 @@ class TestTargets(unittest.TestCase):
     keystore_directory = os.path.join('repository_data', 'keystore')
     public_keypath = os.path.join(keystore_directory, 'snapshot_key.pub')
     public_key = repo_tool.import_ed25519_publickey_from_file(public_keypath)
-    target1_filepath = os.path.join(self.targets_directory, 'file1.txt')
 
     # Set needed arguments by delegate().
     public_keys = [public_key]
     rolename = 'tuf'
-    paths = [os.path.basename(target1_filepath)]
+    paths = ['file1.txt']
     threshold = 1
 
     self.targets_object.delegate(rolename, public_keys, paths, threshold,
@@ -1160,42 +1156,53 @@ class TestTargets(unittest.TestCase):
     # Verify the targets object initially contains zero target files.
     self.assertEqual(self.targets_object.target_files, {})
 
-    target_filepath = os.path.join(self.targets_directory, 'file1.txt')
+    target_filepath = 'file1.txt'
     self.targets_object.add_target(target_filepath)
 
     self.assertEqual(len(self.targets_object.target_files), 1)
-    self.assertTrue('file1.txt' in self.targets_object.target_files)
+    self.assertTrue(target_filepath in self.targets_object.target_files)
 
     # Test the 'custom' parameter of add_target(), where additional information
     # may be specified for the target.
-    target2_filepath = os.path.join(self.targets_directory, 'file2.txt')
+    target2_filepath = 'file2.txt'
+    target2_fullpath = os.path.join(self.targets_directory, target2_filepath)
 
     # The file permission of the target (octal number specifying file access
     # for owner, group, others (e.g., 0755).
-    octal_file_permissions = oct(os.stat(target2_filepath).st_mode)[4:]
+    octal_file_permissions = oct(os.stat(target2_fullpath).st_mode)[4:]
     custom_file_permissions = {'file_permissions': octal_file_permissions}
     self.targets_object.add_target(target2_filepath, custom_file_permissions)
 
     self.assertEqual(len(self.targets_object.target_files), 2)
-    self.assertTrue('file2.txt' in self.targets_object.target_files)
+    self.assertTrue(target2_filepath in self.targets_object.target_files)
     self.assertEqual(self.targets_object.target_files['file2.txt']['custom'],
                      custom_file_permissions)
 
     # Attempt to replace target that has already been added.
-    octal_file_permissions2 = oct(os.stat(target_filepath).st_mode)[4:]
+    octal_file_permissions2 = oct(os.stat(target2_fullpath).st_mode)[4:]
     custom_file_permissions2 = {'file_permissions': octal_file_permissions}
     self.targets_object.add_target(target2_filepath, custom_file_permissions2)
-    self.assertEqual(self.targets_object.target_files['file2.txt']['custom'],
-    custom_file_permissions2)
+    self.assertEqual(self.targets_object.target_files[target2_filepath]['custom'],
+        custom_file_permissions2)
 
     # Test improperly formatted arguments.
-    self.assertRaises(securesystemslib.exceptions.FormatError, self.targets_object.add_target, 3)
-    self.assertRaises(securesystemslib.exceptions.FormatError, self.targets_object.add_target, 3,
-                      custom_file_permissions)
-    self.assertRaises(securesystemslib.exceptions.FormatError, self.targets_object.add_target,
-                      target_filepath, 3)
+    self.assertRaises(securesystemslib.exceptions.FormatError,
+        self.targets_object.add_target, 3)
+    self.assertRaises(securesystemslib.exceptions.FormatError,
+        self.targets_object.add_target, 3, custom_file_permissions)
+    self.assertRaises(securesystemslib.exceptions.FormatError,
+        self.targets_object.add_target, target_filepath, 3)
 
+    # A target path starting with a directory separator
+    self.assertRaises(tuf.exceptions.InvalidNameError,
+        self.targets_object.add_target, '/file1.txt')
 
+    # A target path using a backward slash as a separator
+    self.assertRaises(tuf.exceptions.InvalidNameError,
+        self.targets_object.add_target, 'subdir\\file1.txt')
+
+    # Should not access the file system to check for non-existent files
+    self.targets_object.add_target('non-existent')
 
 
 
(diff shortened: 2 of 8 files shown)


tuf/repository_tool.py
```diff
@@ -2772,6 +2755,61 @@ def _locate_and_update_target_in_bin(self, target_filepath, method_name):
...
+      Normalizes pathname.
+
+    <Returns>
+      The normazlied 'pathname'.
+    """
+
+    tuf.formats.RELPATH_SCHEMA.check_match(pathname)
+
+    if os.path.isabs(pathname):
+      raise tuf.exceptions.InvalidNameError(repr(pathname) + ' contains a leading'
```
lukpueh: Minor nit: Could you please wrap before "*leading*", so that we don't exceed 80 characters per line. (note to myself: we need to fix the linter configuration!)

tuf/repository_tool.py
```diff
@@ -2772,6 +2755,61 @@ def _locate_and_update_target_in_bin(self, target_filepath, method_name):
...
+
+    <Exceptions>
+      securesystemslib.exceptions.FormatError, if 'pathname' is improperly
+      formatted.
+
+      tuf.exceptions.InvalidNameError, if 'pathname' starts with a directory
+      separator.
+
+    <Side Effects>
+      Normalizes pathname.
```
lukpueh: Minor nit: I wouldn't call this a side effect, as it doesn't change any state outside of this function (as e.g. I/O or modifying some globals would do). I know that other function docstrings in tuf sometimes get this wrong. Please remove.

tuf/repository_tool.py
```diff
@@ -2772,6 +2755,61 @@ def _locate_and_update_target_in_bin(self, target_filepath, method_name):
           ' in any of the bins.')
 
 
+  def _check_relpath(self, pathname):
```
lukpueh: Minor: If we do normalize paths we should call this function something like  **_normalize_path**.

tuf/repository_tool.py
```diff
@@ -1937,47 +1939,38 @@ def add_target(self, filepath, custom=None):
...
+    if relative_path not in roleinfo['paths']:
+      logger.debug('Adding new target: ' + repr(relative_path))
+      roleinfo['paths'].update({relative_path: custom})
 
     else:
-      raise securesystemslib.exceptions.Error(repr(filepath) + ' is not'
-          ' a valid file in the repository\'s targets'
-          ' directory: ' + repr(self._targets_directory))
+      logger.debug('Replacing target: ' + repr(relative_path))
+      roleinfo['paths'].update({relative_path: custom})
```
lukpueh: Looks like the if/else is only needed for the debug message and we can put `roleinfo['paths'].update({relative_path: custom})` below it (only once).

tuf/repository_tool.py
```diff
@@ -1937,47 +1939,38 @@ def add_target(self, filepath, custom=None):
...
+
+    tuf.roledb.update_roleinfo(self._rolename, roleinfo,
+      repository_name=self._repository_name)
```
lukpueh: Minor nit: Please use double indentation (4 spaces) for line continuation (here and elsewhere too).

tuf/repository_tool.py
```diff
@@ -2772,6 +2755,61 @@ def _locate_and_update_target_in_bin(self, target_filepath, method_name):
...
+      formatted.
+
+      tuf.exceptions.InvalidNameError, if 'pathname' starts with a directory
+      separator.
+
+    <Side Effects>
+      Normalizes pathname.
+
+    <Returns>
+      The normazlied 'pathname'.
```
lukpueh: s/normazlied/normalized

tuf/repository_tool.py
```diff
@@ -2273,26 +2258,27 @@ def delegate(self, rolename, public_keys, paths, threshold=1,
...
-      if path.startswith(os.sep):
-        raise tuf.exceptions.Error('One of the given paths contains a leading'
-            ' path separator: ' + repr(path) + '.  All delegated paths should'
-            ' be relative to the repo\'s targets directory.')
-
-      if not path.startswith(self._targets_directory + os.sep):
-        logger.warning(repr(path) + ' is not located in the repository\'s'
-          ' targets directory: ' + repr(self._targets_directory))
+      # Check if the delegated paths or glob patterns are relative and
+      # normalize them. Paths' existense on the file system is not verified.
```
lukpueh: s/existense/existence

lukpueh (changes requested):
Thanks a ton for this beautiful PR, @sechkova.  💯 for clean and well-documented code and great commit-discipline. 

I have left a few very minor inline comments and would like to share and discuss some high-level thoughts about the `_check_relpath` function. Note that there is no need to fix inline comments before we have agreed on desired behavior.

Let's first agree on **what we want here**:
1. Remove file existence checks in `add_target*` and `add_paths` methods (`write`/`writeall` should  deal with that).  
2. Make sure that the paths that end-up in roledb have forward slash path separators and don't start with a leading forward slash. Although the spec only recommends this (see [spec#67](https://github.com/theupdateframework/specification/pull/67)) python-tuf seems to enforce this (see various `path.replace('\\', '/')`).
 
While you already fixed goal 1. (hooray ✅ ), I see a **few minor issue** in regards to goal 2.
1) `isabs`, `split(os.path)` and `normpath` all behave differently on Unix and Windows, so the warning/erroring conditions and normalization behavior vary depending on the combination of used platform and passed path.
2) I wonder if the mixed behavior of warning/erroring/normalizing won't confuse the caller.

Let's back up a little, I think there are two ways to **achieve what we want** (part 2 above).
1. expect the caller to pass the right path (`/` as separator, no leading `/`), no matter the platform, and raise an error if they don't:
    - Pro: clear behavior (no hidden normalizing magic)
    - Con: inconvenient for the caller, e.g. if they want to pass the return value of tuf's 
 `get_filepaths_in_directory` (returns abs paths), or if they are on windows and use (`listdir` or `walk`).
2.  take *anything* and normalize, i.e. convert separators to `/` and remove leading `/`.
    - Pro: convenient for the caller 
    - Con: unclear behavior (too much magic)

I strongly lean towards solution 1 (least surprise) plus providing normalizing utility functions for the caller's convenience (can be in a separate PR). What do others think?

Btw. in the future we might be even more sophisticated and have separate checker functions for target path, i.e. path-relative-scheme-less-URL, and path pattern, i.e. path-relative-scheme-less-URL but with glob pattern symbols. (see [spec#67](https://github.com/theupdateframework/specification/pull/67)).

lukpueh:
I'd love to hear what others think about the two approaches I suggested above... 
> 1. expect the caller to pass the right path (`/` as separator, no leading `/`), no matter the platform, and raise an error if they don't.
> 2. take anything and normalize, i.e. convert separators to `/` and remove leading `/`.

See https://github.com/theupdateframework/tuf/pull/1008#pullrequestreview-382660018 for more details.

(ping @mnm678,  @SantiagoTorres, @trishankatdatadog)

trishankatdatadog:
> I'd love to hear what others think about the two approaches I suggested above...
> 
> >

Approach #2 seems friendlier, although some debug/info/warning notices to the logger might help users debug unexpected issues.

mnm678:
I think #1 would be easier to implement on a wide range of systems. It adds a bit of work for the caller, but eliminates confusing errors when the path doesn't match a known format.

lukpueh:
Thanks for your 4 cents, @mnm678 and @trishankatdatadog. I think Marina and I outvote you in the favor of a strict approach, Trishank. :P I really feel like we should have a function that accepts exactly and only what the specification recommends.

Nonetheless, I do share the friendliness concern, which I think we can accommodate by providing normalization tooling, such as `get_target_paths_in_directory(path_to_dir)`, or `normalize_file_paths(list_of_file_paths)`.

What's your take on this @sechkova and @joshuagl?

joshuagl:
I'm in favour of the strict approach with supporting helper functions; so that we follow the principle of least surprise, match the specification and support client developers (with the helper functions).

sechkova:
Thanks a lot everyone for the helpful comments and suggestions!
I vote in favour of a strict approach + helpers too. Let me try and summarize to help moving the work forward. 

Related to path verification:
- expect the caller to pass the right path (/ as separator, no leading /), no matter the platform, and raise an error if the path (seen as a string) does not match
- do all file system access operations on the path (if any) on` write()`, `writeall(`) and raise an error in case of non-existing files etc ✅.

In terms of adding normalization functions, how do you feel about:
- adding `normalize_file_paths(list_of_file_paths)` which would work only on paths as strings (no file system access)
- keeping `get_target_paths_in_directory(path_to_dir)` which already performs checks on the file systems in case the caller needs such functionality + maybe adding a call to the above `normalize_file_paths`

@lukpueh Parts of the tutorial seem to be suggesting absolute paths, which will be rejected by the new approach. Do you think the normalization tooling + updating the tutorials about its usage is appropriate for a separate PR?

lukpueh:
> * expect the caller to pass the right path (/ as separator, no leading /), no matter the platform, and raise an error if the path (seen as a string) does not match

Yes, I think this is a good start.

(Later in a follow-up PR, we should consider to only accept [path-relative-scheme-less-URL](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string) strings for *target paths* (see [TARGETPATH in spec](https://github.com/theupdateframework/specification/blame/69b8456614796b57c6076d3894568a83b3f7ce1a/tuf-spec.md#L808-L815)), and something similar but with shell-style wildcards allowed for *path* patterns (see [PATHPATTERN in spec](https://github.com/theupdateframework/specification/blame/69b8456614796b57c6076d3894568a83b3f7ce1a/tuf-spec.md#L881-L891))).



> * do all file system access operations on the path (if any) on` write()`, `writeall(`) and raise an error in case of non-existing files etc ✅.

Yes!


> * adding `normalize_file_paths(list_of_file_paths)` which would work only on paths as strings (no file system access)

Yes! But we can do this in a follow-up PR. 

> * keeping `get_target_paths_in_directory(path_to_dir)` which already performs checks on the file systems in case the caller needs such functionality + maybe adding a call to the above `normalize_file_paths`

`get_target_paths_in_directory` doesn't exist yet. You probably mean `get_filepaths_in_directory`? I think we can leave that. And maybe add something like:
```python
 def get_normalized_target_paths_in_directory(path):
    return normalize_file_paths(get_filepaths_in_directory(path)):
```
We can also do that in in a follow-up PR.


> @lukpueh Parts of the tutorial seem to be suggesting absolute paths, which will be rejected by the new approach. Do you think the normalization tooling + updating the tutorials about its usage is appropriate for a separate PR?

Hm. I'd rather keep the tutorial usable and not fail the tests. Maybe we can add a quick intermediate fix that replaces the `get_filepaths_in_directory` with a hard-coded list of the paths it would return but in the format expected by the new add targets functions. What do you think?

sechkova:
> `get_target_paths_in_directory` doesn't exist yet. You probably mean `get_filepaths_in_directory`?

 Yes :eyes: 

> 
> > @lukpueh Parts of the tutorial seem to be suggesting absolute paths, which will be rejected by the new approach. Do you think the normalization tooling + updating the tutorials about its usage is appropriate for a separate PR?
> 
> Hm. I'd rather keep the tutorial usable and not fail the tests. Maybe we can add a quick intermediate fix that replaces the `get_filepaths_in_directory` with a hard-coded list of the paths it would return but in the format expected by the new add targets functions. What do you think?

Agree on keeping the tutorials consistent with the code and the tests working. A hard-coded list will do the job for now. I'll add it in the PR 👍

tuf/repository_tool.py
sechkova: Fixed in https://github.com/theupdateframework/tuf/pull/1008/commits/9eef0bd14370414bd649e9093ff8fbb500465301

tuf/repository_tool.py
sechkova: Fixed in https://github.com/theupdateframework/tuf/pull/1008/commits/9eef0bd14370414bd649e9093ff8fbb500465301

tuf/repository_tool.py
sechkova: Fixed in https://github.com/theupdateframework/tuf/pull/1008/commits/9eef0bd14370414bd649e9093ff8fbb500465301

tuf/repository_tool.py
sechkova: Fixed in 9eef0bd14370414bd649e9093ff8fbb500465301

tuf/repository_tool.py
sechkova: Fixed in https://github.com/theupdateframework/tuf/commit/9eef0bd14370414bd649e9093ff8fbb500465301

tuf/repository_tool.py
sechkova: Already fixed by merging #1007

tuf/repository_tool.py
sechkova: Already fixed by merging #1007

tests/test_repository_tool.py
```diff
@@ -1233,25 +1257,25 @@ def test_remove_target(self):
...
 
 
-    # Test invalid filepath argument (i.e., non-existent or invalid file.)
-    self.assertRaises(securesystemslib.exceptions.Error,
-                      self.targets_object.remove_target,
-                      '/non-existent.txt')
+    # # Test invalid filepath argument (i.e., non-existent or invalid file.)
+    # self.assertRaises(securesystemslib.exceptions.Error,
+    #                   self.targets_object.remove_target,
+    #                   '/non-existent.txt')
```
joshuagl: There's no need to leave this commented out code, could you remove it?

tests/test_repository_tool.py
```diff
@@ -1299,22 +1320,20 @@ def test_delegate(self):
...
+    self.targets_object.delegate(rolename, public_keys, ['non-existent'],
         threshold, terminating=False, list_of_targets=list_of_targets,
         path_hash_prefixes=path_hash_prefixes)
 
+    # Test for delegated targets that do not exist.
+    # An exception should not be raised for non-existent delegated paths,
+    # since at this point the file system should not be accessed yet
+    self.targets_object.delegate(rolename, public_keys, [], threshold,
+        terminating=False, list_of_targets=['non-existent.txt'],
+        path_hash_prefixes=path_hash_prefixes)
```
joshuagl: I'm not sure this test still makes sense?

tests/test_repository_tool.py
```diff
@@ -1349,10 +1368,20 @@ def test_delegate(self):
...
-        self.targets_object.delegate, rolename, public_keys, [], threshold,
-        ['/non-existent'], path_hash_prefixes)
+    # A path or target starting with a directory separator
+    self.assertRaises(tuf.exceptions.InvalidNameError,
+        self.targets_object.delegate, rolename, public_keys, ['/*'])
+    self.assertRaises(tuf.exceptions.InvalidNameError,
+        self.targets_object.delegate, rolename, public_keys, [],
+        list_of_targets=['/file1.txt'])
+
+    # A path or target starting with a directory separator
```
joshuagl: This comment doesn't seem right, aren't we testing Windows path separators here?

tests/test_repository_tool.py
```diff
@@ -1676,6 +1711,27 @@ def test_revoke(self):
...
+    # Test that non-exsitent path does not raise exception (_check_path works
+    # checks only the path string for compliance
+    self.targets_object._check_path('non-existent.txt')
+    self.targets_object._check_path('subdir/non-existent')
+
+    # Test improperly formatted rolename argument.
+    self.assertRaises(securesystemslib.exceptions.FormatError,
+        self.targets_object._check_path, 3)
+
+    # Test improperly formatted rolename argument.
```
joshuagl: Same here, is this comment correct?

tests/test_repository_tool.py
```diff
@@ -1676,6 +1711,27 @@ def test_revoke(self):
...
+    # Test that correct path does not raise exception: using '/' as a separator
+    # and does not start with a directory separator
+    self.targets_object._check_path('file1.txt')
+
+    # Test that non-exsitent path does not raise exception (_check_path works
+    # checks only the path string for compliance
+    self.targets_object._check_path('non-existent.txt')
+    self.targets_object._check_path('subdir/non-existent')
+
+    # Test improperly formatted rolename argument.
```
joshuagl: This comment doesn't seem to match the code, does it?

tests/test_tutorial.py
```diff
@@ -223,13 +223,10 @@ def test_tutorial(self):
...
 
     repository = load_repository('repository')
-    list_of_targets = repository.get_filepaths_in_directory(
-        os.path.join('repository', 'targets'), recursive_walk=False, followlinks=True)
 
-    self.assertListEqual(sorted(list_of_targets), [
-        os.path.abspath(os.path.join('repository', 'targets', 'file1.txt')),
-        os.path.abspath(os.path.join('repository', 'targets', 'file2.txt')),
-        os.path.abspath(os.path.join('repository', 'targets', 'file3.txt'))])
+    # List of targets is hardcoded since get_filepaths_in_directory()
```
joshuagl: Is this a TODO item to switch to using `get_filepaths_in_directory` + a helper at some point? The comment doesn't make a lot of sense once the diff showing the deleted call to `get_filepaths_in_directory` isn't visible. What do you think?

docs/TUTORIAL.md
```diff
@@ -376,10 +369,12 @@ the target filepaths to metadata.
 # (octal number specifying file access for owner, group, others e.g., 0755) is
 # added alongside the default fileinfo.  All target objects in metadata include
 # the target's filepath, hash, and length.
+# Note: target path passed to add_target() method has to be relative
+# to the targets directory or an exception is raised.
 >>> target4_filepath = os.path.abspath("repository/targets/myproject/file4.txt")
 >>> octal_file_permissions = oct(os.stat(target4_filepath).st_mode)[4:]
 >>> custom_file_permissions = {'file_permissions': octal_file_permissions}
->>> repository.targets.add_target(target4_filepath, custom_file_permissions)
+>>> repository.targets.add_target('myproject/file4.txt', custom_file_permissions)
```
joshuagl: Is it just me or does this pattern of assigning the full path to a variable and then passing the relative sub-path as a string look a little strange?

tests/test_tutorial.py
```diff
@@ -243,7 +240,7 @@ def test_tutorial(self):
         'repository', 'targets', 'myproject', 'file4.txt'))
     octal_file_permissions = oct(os.stat(target4_filepath).st_mode)[4:]
     custom_file_permissions = {'file_permissions': octal_file_permissions}
-    repository.targets.add_target(target4_filepath, custom_file_permissions)
+    repository.targets.add_target('myproject/file4.txt', custom_file_permissions)
```
joshuagl: Same as (several comments) above, does this pattern look a little strange?

tuf/repository_tool.py
```diff
@@ -2526,11 +2523,15 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
       ordered_roles.append(role)
 
     for target_path in list_of_targets:
+      # Check if the target path is relative and normalize it. File's existence
```
joshuagl: `_check_path()` no longer does any normalising, right? Could you update the comment here?

joshuagl (commented):
Thanks for the PR @sechkova, great work.

I've made a few minor inline comments, mostly just clarifying code comments in the tests.

tests/test_repository_tool.py
sechkova: Fixed in squashed commit https://github.com/theupdateframework/tuf/pull/1008/commits/3a13b7312a9148f7aac553affa371f3f26c69f11

tests/test_repository_tool.py
sechkova: I kept the 'non-existent' target checks in all related tests  just to make sure that the file system is not accessed at this point and no exception is raised.

tests/test_repository_tool.py
sechkova: Fixed in squashed commit https://github.com/theupdateframework/tuf/pull/1008/commits/3a13b7312a9148f7aac553affa371f3f26c69f11

tests/test_repository_tool.py
sechkova: Fixed in squashed commit https://github.com/theupdateframework/tuf/pull/1008/commits/3a13b7312a9148f7aac553affa371f3f26c69f11

tests/test_repository_tool.py
sechkova: Fixed in squashed commit https://github.com/theupdateframework/tuf/pull/1008/commits/3a13b7312a9148f7aac553affa371f3f26c69f11

tests/test_tutorial.py
sechkova: Agree, I tried to improve the comment and keep the TODO idem: https://github.com/theupdateframework/tuf/pull/1008/commits/6d8a84d9938d149e01c3bd0bfb52d056bb5b7b55

docs/TUTORIAL.md
sechkova: Agree, it looks awkward but it is also a good example why a normalising path helper is needed :) 
I tired to rewrite it in a more reasonable way: https://github.com/theupdateframework/tuf/pull/1008/commits/6d8a84d9938d149e01c3bd0bfb52d056bb5b7b55

tests/test_tutorial.py
sechkova: Same as comment above: https://github.com/theupdateframework/tuf/pull/1008/commits/6d8a84d9938d149e01c3bd0bfb52d056bb5b7b55

tuf/repository_tool.py
sechkova: Fixed during merge in https://github.com/theupdateframework/tuf/pull/1008/commits/ba8ffd218da8e44f1266a09b63c462b34f7f183d

sechkova:
Thank you @joshuagl for actually reading all comments sections!
Fixes for your comments and for the tests failing on Windows are added.

docs/TUTORIAL.md
```diff
@@ -376,10 +369,13 @@ the target filepaths to metadata.
...
->>> target4_filepath = os.path.abspath("repository/targets/myproject/file4.txt")
->>> octal_file_permissions = oct(os.stat(target4_filepath).st_mode)[4:]
+# Note: target path passed to add_target() method has to be relative
+# to the targets directory or an exception is raised.
+>>> target4_filepath = 'myproject/file4.txt'
+>>> target4_abspath = os.path.abspath(os.path.join('repository', 'targets', target4_filepath))
+>>> octal_file_permissions = oct(os.stat(target4_abspath).st_mode)[4:]
 >>> custom_file_permissions = {'file_permissions': octal_file_permissions}
->>> repository.targets.add_target(target4_filepath, custom_file_permissions)
+>>> repository.targets.add_target('target4_filepath', custom_file_permissions)
```
lukpueh: This should be the variable `target4_filepath` and not the string, akin to `test_tutorial.py`

*If we only had a tutorial test that actually tests the tutorial document  (see https://github.com/theupdateframework/tuf/pull/775#pullrequestreview-172429283 and https://github.com/theupdateframework/tuf/issues/808#issuecomment-487597020)  ... :)*

tuf/repository_tool.py
```diff
@@ -2754,6 +2755,44 @@ def delegations(self):
...
+      None.
+    """
+
+    tuf.formats.RELPATH_SCHEMA.check_match(pathname)
+
+    if '\\' in pathname:
+      raise tuf.exceptions.InvalidNameError('Path ' + repr(pathname)
+          + ' does not use the forward slash (/) as directory separator.')
+
+    if pathname.startswith(os.sep):
```
lukpueh: I'm pretty sure that this would allow a windows user to pass an absolute path that starts with "/", which we also don't want. I suggest to change to `if pathname.startswith('/'):`. Or am I missing something?

tests/test_repository_tool.py
```diff
@@ -1676,6 +1711,29 @@ def test_revoke(self):
...
+    self.targets_object._check_path('subdir/non-existent')
+
+    # Test improperly formatted pathname argument.
+    self.assertRaises(securesystemslib.exceptions.FormatError,
+        self.targets_object._check_path, 3)
+
+    # Test invalid pathname - starting with os separator
+    pathname = os.sep + 'file1.txt'
+    self.assertRaises(tuf.exceptions.InvalidNameError,
+        self.targets_object._check_path, pathname)
```
lukpueh: This test will raise for different reasons on *nix and windows, as a consequence we don't trigger the *starts with "/"* condition when testing on windows. (also see comment in `_check_path`)

lukpueh (changes requested):
Many many thanks for the updates, @sechkova! I have left three minor comments. Please address and we are good to go here. :)

docs/TUTORIAL.md
sechkova: Thanks for noticing, fix is squashed in https://github.com/theupdateframework/tuf/pull/1008/commits/87e1e11e99cbdc3d12a7e1eb0838b1e4baf6f63b

tests/test_repository_tool.py
sechkova: Agree, the commit where I added os.sep in the tests is reverted and instead `_check_path` uses `if pathname.startswith('/')` as suggested, thanks!
Also added an additional test case for paths starting with '\\', just in case for the future.
The new commit: https://github.com/theupdateframework/tuf/pull/1008/commits/e85a3b3d694336d6361ded7aaccd3f45e10fef0f

tuf/repository_tool.py
sechkova: See comment below

sechkova:
Rebased and new changes added

tuf/repository_tool.py
```diff
@@ -1854,6 +1854,8 @@ def add_paths(self, paths, child_rolename):
       securesystemslib.exceptions.Error, if 'child_rolename' has not been
       delegated yet.
 
+      tuf.exceptions.InvalidNameError, if 'pathname' does not match pattern.
```
lukpueh: Minor nit: Would you mind replacing `'pathname'` with what's actually being checked, that is here  *"any of the passed 'paths'"*. (same comment applies to other occurrences of this line).

tuf/repository_tool.py
```diff
@@ -1879,16 +1881,11 @@ def add_paths(self, paths, child_rolename):
...
-      if not path.startswith(targets_directory):
-        logger.debug(repr(path) + ' is not located in the'
-            ' repository\'s targets'
-            ' directory: ' + repr(self._targets_directory))
-
+      # Check if the delegated paths or glob patterns are relative and use
+      # forward slash as a separator or raise an exception. Paths' existence
+      # on the file system is not verified. If the path is incorrect,
+      # the targetfile won't be matched successfully during a client update.
+      self._check_path(path)
```
lukpueh, line 1885: It seems like we don't need the `relative_paths` variable anymore.

tuf/repository_tool.py
```diff
@@ -2037,21 +2040,17 @@ def add_targets(self, list_of_targets):
...
+    # filepaths here allows the freedom to add targets and parent restrictions
+    # in any order and minimize the number of times these checks are performed.
     for target in list_of_targets:
-      filepath = os.path.join(self._targets_directory, target)
-
-      relative_list_of_targets.append(
-          filepath[targets_directory_length + 1:].replace('\\', '/'))
-
+      self._check_path(target)
+      relative_list_of_targets.append(target)
```
lukpueh: Same as above in `add_paths`. Given that we don't do no normalizing anymore, we don't have to copy the target paths to a new list, i.e. `relative_list_of_targets`. Please remove.

lukpueh (commented):
Just stumbled across two more things that probably came from a rebase on some other work (plus a a minor docstring request). Would you mind fixing those? :)

sechkova:
Rebased and fixed in https://github.com/theupdateframework/tuf/pull/1008/commits/882df8e26d64d12a4e157f789fcf0ba2124de371 :)

lukpueh (approved):

lukpueh:
Hooray! 🎉  This does look good to me now. Thanks for your patience, @sechkova! I created tickets for the follow-up PRs we discussed above (#1018, #1019). Merging...

Issue #1005, opened 2020-03-19
Fix performance and metadata representation issues in roledb


lukpueh:
*Update on 2020-04-08:*
The particular performance issue described below as **Problem 1** has been resolved in #1012 by applying the below proposed **Quick fix (a)**, however, the underlying **Problem 2** remains.

 I suggest to keep this issue around for context, and proceed with below outlined  **Long-term fixes (c) and (d)**.

----------

As reported by @woodruffw (thanks!) and reproduced in the profiling experiment below, repository tool's `delegate_hashed_bins` method  has massive performance issues when delegating to many bins (e.g. 16K), for the following reasons:

**Problem 1:**
`delegate_hashed_bins` calls `delegate` for each new bin. `delegate` then adds the new [**delegated** role to roledb](https://github.com/theupdateframework/tuf/blob/c63b7f3bdcab48fad8dd0a8b1e828ae5d9af0a34/tuf/repository_tool.py#L2310-L2312), but also [adds delegation infos to the steadily growing **delegating** role in roledb](https://github.com/theupdateframework/tuf/blob/c63b7f3bdcab48fad8dd0a8b1e828ae5d9af0a34/tuf/repository_tool.py#L2315-L2337) in each iteration.

**Problem 2:**
Operations on `roledb` are extremely expensive, because they perform two `deepcopy`s of a `roleinfo` dictionary, like so:

```
1. roleinfo = roledb.get_roleinfo() # deepcopying roleinfo out of roledb
2. modify roledinfo
3. roledb.update_roleinfo(roleinfo) # deepcopying roleinfo back into roleb
```

Furthermore, `update_roleinfo` calls `tuf.formats.ROLEDB_SCHEMA.check_match` for the passed roleinfo, which recursively iterates over all its elements to check for some criteria.

-----

I propose the following fixes:

- **Quick fix (a)**:

Change `delegate_hashed_bins` to update the delegating role only once in the roledb, instead of once for every bin. This means that we can't call `delegate` as it is right now, but would need to replicate or, ideally, factor out common functionality.

- **Intermediate fix (b)**:

Stop deepcopying and instead update roledb by reference. The deepcopys are an unnecessary overhead throughout the codebase (see profiling experiment). This seems quite feasible in most cases, but it is also risky. It might be worth skipping the intermediate fix, and move from quick fix directly to the long-term fix.

- **Long-term fix (c)**:
[...]

PR #1012, opened 2020-04-02
Improve performance of `delegate_hashed_bins` when delegating to a large number of bins

joshuagl:
**Fixes issue #**: Quick fix for #1005 

* Factor out `delegate()` logic into various helpers:
  * `_keys_to_keydict()`
  * `_create_delegated_target()`
  * `_update_roledb_delegations()`
* Use those helpers directly in `delegated_hashed_bins()`, rather than calls to `delegate()`, most significantly queuing up changes to roledb and performing them in one operation with `_update_roledb_delegations()`

Based on the profiling done by @lukpueh after @woodruffw's report we gain a significant speed improvement when delegating to a large number of hashed bins (i.e. 16k):
before: `python3 profile_tuf_bins.py  4978.27s user 39.37s system 99% cpu 1:24:00.47 total`
after: `python3 profile_tuf_bins.py  21.09s user 10.61s system 96% cpu 32.979 total`

requirements**:

### DIFF
--- a/tuf/repository_tool.py
+++ b/tuf/repository_tool.py
@@ -2546,12 +2546,56 @@ class Targets(Metadata):
       hash_prefix = _get_hash(target_path.replace('\\', '/').lstrip('/'))[:prefix_length]
       ordered_roles[int(hash_prefix, 16) // bin_size]["target_paths"].append(target_path)
 
-    for bin_rolename in ordered_roles:
-      # Delegate from the "unclaimed" targets role to each 'bin_rolename'
-      self.delegate(bin_rolename['name'], keys_of_hashed_bins, [],
-          list_of_targets=bin_rolename['target_paths'],
-          path_hash_prefixes=bin_rolename['target_hash_prefixes'])
-      logger.debug('Delegated from ' + repr(self.rolename) + ' to ' + repr(bin_rolename))
+    keyids, keydict = _keys_to_keydict(keys_of_hashed_bins)
+
+    # A queue of roleinfo's that need to be updated in the roledb
+    delegated_roleinfos = []
+
+    for bin_role in ordered_roles:
+      # TODO: originally we just called self.delegate() for each item in this
+      # iteration. However, this is *extremely* slow when creating a large
+      # number of hashed bins, i.e. 16k as is recommended for PyPI usage in
+      # PEP 458: https://www.python.org/dev/peps/pep-0458/
+      # The source of the slowness is the interactions with the roledb, which
+      # causes several deep copies of roleinfo dictionaries:
+      # https://github.com/theupdateframework/tuf/issues/1005
+      # Once the underlying issues in #1005 are resolved, i.e. some combination
+      # of the intermediate and long-term fixes, we may simplify here by
+      # switching back to just calling self.delegate(), but until that time we
+      # queue roledb interactions and perform all updates to the roledb in one
+      # operation at the end of the iteration.
+
+      relative_paths = {}
+      targets_directory_length = len(self._targets_directory)
+      for path in bin_role['target_paths']:
+        relative_paths.update({path[targets_directory_length:]: {}})
+
+      # Delegate from the "unclaimed" targets role to each 'bin_role'
+      target = self._create_delegated_target(bin_role['name'], keyids, 1,
+          relative_paths)
+
+      roleinfo = {'name': bin_role['name'],
+                  'keyids': keyids,
+                  'threshold': 1,
+                  'terminating': False,
+                  'path_hash_prefixes': bin_role['target_hash_prefixes']}
+      delegated_roleinfos.append(roleinfo)
+
+      for key in keys_of_hashed_bins:
+        target.add_verification_key(key)
+
+      # Add the new delegation to the top-level 'targets' role object (i.e.,
+      # 'repository.targets()').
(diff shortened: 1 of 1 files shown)


tuf/repository_tool.py
```diff
@@ -2171,6 +2171,54 @@ def get_delegated_rolenames(self):
...
+    Create a new Targets object for the 'rolename' delegation.  An initial
+    expiration is set (3 months from the current time).
+    """
+
+    expiration = tuf.formats.unix_timestamp_to_datetime(
+        int(time.time() + TARGETS_EXPIRATION))
+    expiration = expiration.isoformat() + 'Z'
+
+    if not paths:
+      paths = {}
```
lukpueh: Minor suggestion: Both function that use this helper are guaranteed to pass a paths object that is a `dict`. I guess we could make the paths kwarg an arg and remove the two lines here (we'll have fewer things to revert :P).  What do you think?

tuf/repository_tool.py
```diff
@@ -2530,14 +2546,58 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
       hash_prefix = _get_hash(target_path.replace('\\', '/').lstrip('/'))[:prefix_length]
       ordered_roles[int(hash_prefix, 16) // bin_size]["target_paths"].append(target_path)
 
+    keyids, keydict = _keys_to_keydict(keys_of_hashed_bins)
+
+    # A queue of roleinfo's that need to be updated in the roledb
+    delegated_roleinfos = []
+
     for bin_rolename in ordered_roles:
```
lukpueh: Minor nit: I guess I approved this just two days ago, but now that I look at it again, I wonder if  it would make sense to call it `bin_role` instead of `bin_rolename`, given that it is a dict that i.a. contains a `name` field. `bin_rolename['name']` below reads a bit odd. What do you think?

tuf/repository_tool.py
```diff
@@ -2530,14 +2546,58 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
...
+      targets_directory_length = len(self._targets_directory)
+      for path in bin_rolename['target_paths']:
+        relative_paths.update({path[targets_directory_length:]: {}})
+
       # Delegate from the "unclaimed" targets role to each 'bin_rolename'
-      self.delegate(bin_rolename['name'], keys_of_hashed_bins, [],
-          list_of_targets=bin_rolename['target_paths'],
-          path_hash_prefixes=bin_rolename['target_hash_prefixes'])
+      target = self._create_delegated_target(bin_rolename['name'], keyids,
+          paths=relative_paths)
```
lukpueh: May I suggest to pass the same hardcoded threshold of `1` as below? This is what `delegate()` would have done.
https://github.com/theupdateframework/tuf/blob/32326975e040ad82e22d66ec27823a6122e49136/tuf/repository_tool.py#L2318
https://github.com/theupdateframework/tuf/blob/32326975e040ad82e22d66ec27823a6122e49136/tuf/repository_tool.py#L2336

As a consequence we could  remove the default for threshold in `_create_delegated_target`. What do you think?

lukpueh (approved):
Thanks for the changes, @joshuagl. The performance gain is uncanny! :)

I have a few minor suggestions nothing blocking though. Let me know if you want to change anything, otherwise we can merge this.

tuf/repository_tool.py
joshuagl: I'm always happy to remove lines of code, thanks for the suggestion. Squashed in to the first patch in 9be4634.

tuf/repository_tool.py
joshuagl: Agreed, it does look a bit odd. I've squashed the suggested change into the second patch in 6703b51

tuf/repository_tool.py
joshuagl: Squashed in to 6703b51

joshuagl:
> Thanks for the changes, @joshuagl. The performance gain is uncanny! :)

Thanks largely to your profiling 🙂 

> I have a few minor suggestions nothing blocking though. Let me know if you want to change anything, otherwise we can merge this.

All great suggestions, I've squashed them into the original patches. Thanks for the thoughtful review @lukpueh

lukpueh (approved):
Thanks for the quick fixes! Re-approving...

lukpueh:
Let's wait for appveyor to return and merge this. :)

tuf/repository_tool.py
```diff
@@ -2174,6 +2174,51 @@ def get_delegated_rolenames(self):
...
 
+  def _create_delegated_target(self, rolename, keyids, threshold, paths):
+    """
+    Create a new Targets object for the 'rolename' delegation.  An initial
+    expiration is set (3 months from the current time).
+    """
+
+    expiration = tuf.formats.unix_timestamp_to_datetime(
+        int(time.time() + TARGETS_EXPIRATION))
+    expiration = expiration.isoformat() + 'Z'
```
woodruffw, line 2185: Unrelated to these changes, but does anybody know why `unix_timestamp_to_datetime` isn't including the `Z` suffix by default? Feels like an API error if the times are intended to always be GMT 🙂

woodruffw (approved):
Thanks a ton!

PR #1014, opened 2020-04-02
Remove uses of keyid_hash_algorithms

mnm678:
This is a first step toward removing keyid_hash_algorithms from the reference implementation as discussed in #848. This pr removes all uses of this field during the client verification by using the keyids provided in the metadata instead of recalculating keyids using the keyid_hash_algorithms.

This pr requires changes to securesystemslib that are is a pull request at https://github.com/secure-systems-lab/securesystemslib/pull/225.

### DIFF
--- a/tests/test_keydb.py
+++ b/tests/test_keydb.py
@@ -389,10 +389,16 @@ class TestKeydb(unittest.TestCase):
 
     # Ensure only 'keyid2' was added to the keydb database.  'keyid' and
     # 'keyid3' should not be stored.
+    self.maxDiff = None
     self.assertEqual(rsakey2, tuf.keydb.get_key(keyid2))
-    self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid)
+
+    test_key = rsakey2
+    test_key['keyid'] = keyid
+    self.assertEqual(test_key, tuf.keydb.get_key(keyid))
     self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid3)
+
     rsakey3['keytype'] = 'rsa'
+    rsakey2['keyid'] = keyid2
 
 
 
--- a/tests/test_keydb.py
+++ b/tests/test_keydb.py
@@ -374,11 +374,6 @@ class TestKeydb(unittest.TestCase):
     rsakey3['keytype'] = 'bad_keytype'
     keydict[keyid3] = rsakey3
 
-    # New key with a duplicate keyid
-    #rsakey4 = KEYS[1]
-    #keyid4 = KEYS[3]['keyid']
-    #keydict[keyid4] = rsakey4
-
     version = 8
     expires = '1985-10-21T01:21:00Z'
 
@@ -404,7 +399,6 @@ class TestKeydb(unittest.TestCase):
     self.assertEqual(test_key, tuf.keydb.get_key(keyid))
 
     self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid3)
-    #self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid4)
 
     # reset values
     rsakey3['keytype'] = 'rsa'
--- a/tests/test_keydb.py
+++ b/tests/test_keydb.py
@@ -43,7 +43,7 @@ logger = logging.getLogger(__name__)
 
 # Generate the three keys to use in our test cases.
 KEYS = []
-for junk in range(3):
+for junk in range(4):
   rsa_key = securesystemslib.keys.generate_rsa_key(2048)
   rsa_key['keyid_hash_algorithms'] = securesystemslib.settings.HASH_ALGORITHMS
   KEYS.append(rsa_key)
@@ -365,6 +365,7 @@ class TestKeydb(unittest.TestCase):
     tuf.keydb.clear_keydb()
 
     # 'keyid' does not match 'rsakey2'.
+    # In this case, the key will be added to the keydb
     keydict[keyid] = rsakey2
 
     # Key with invalid keytype.
@@ -372,6 +373,12 @@ class TestKeydb(unittest.TestCase):
     keyid3 = KEYS[2]['keyid']
     rsakey3['keytype'] = 'bad_keytype'
     keydict[keyid3] = rsakey3
+
+    # New key with a duplicate keyid
+    #rsakey4 = KEYS[1]
+    #keyid4 = KEYS[3]['keyid']
+    #keydict[keyid4] = rsakey4
+
     version = 8
     expires = '1985-10-21T01:21:00Z'
 
@@ -387,16 +394,19 @@ class TestKeydb(unittest.TestCase):
 
     self.assertEqual(None, tuf.keydb.create_keydb_from_root_metadata(root_metadata))
 
-    # Ensure only 'keyid2' was added to the keydb database.  'keyid' and
-    # 'keyid3' should not be stored.
+    # Ensure only 'keyid2' and 'keyid' were added to the keydb database.
+    # 'keyid3' and 'keyid4' should not be stored.
     self.maxDiff = None
     self.assertEqual(rsakey2, tuf.keydb.get_key(keyid2))
 
     test_key = rsakey2
     test_key['keyid'] = keyid
     self.assertEqual(test_key, tuf.keydb.get_key(keyid))
-    self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid3)
 
+    self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid3)
+    #self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid4)
+
+    # reset values
     rsakey3['keytype'] = 'rsa'
     rsakey2['keyid'] = keyid2
 
--- a/tests/test_keydb.py
+++ b/tests/test_keydb.py
@@ -43,7 +43,7 @@ logger = logging.getLogger(__name__)
 
 # Generate the three keys to use in our test cases.
 KEYS = []
-for junk in range(4):
+for junk in range(3):
   rsa_key = securesystemslib.keys.generate_rsa_key(2048)
   rsa_key['keyid_hash_algorithms'] = securesystemslib.settings.HASH_ALGORITHMS
   KEYS.append(rsa_key)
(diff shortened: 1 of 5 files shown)


tests/test_keydb.py
```diff
@@ -389,10 +389,16 @@ def test_create_keydb_from_root_metadata(self):
 
     # Ensure only 'keyid2' was added to the keydb database.  'keyid' and
     # 'keyid3' should not be stored.
+    self.maxDiff = None
     self.assertEqual(rsakey2, tuf.keydb.get_key(keyid2))
-    self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid)
+
+    test_key = rsakey2
+    test_key['keyid'] = keyid
+    self.assertEqual(test_key, tuf.keydb.get_key(keyid))
```
jku, line 399: The test is a bit hard to follow here...  Does it really test what the comment above says it tests?

tests/test_keydb.py
```diff
@@ -365,13 +365,20 @@ def test_create_keydb_from_root_metadata(self):
...
     # Key with invalid keytype.
     rsakey3 = KEYS[2]
     keyid3 = KEYS[2]['keyid']
     rsakey3['keytype'] = 'bad_keytype'
     keydict[keyid3] = rsakey3
+
+    # New key with a duplicate keyid
+    #rsakey4 = KEYS[1]
+    #keyid4 = KEYS[3]['keyid']
+    #keydict[keyid4] = rsakey4
```
joshuagl: Should this test work? If not, what's preventing it from working? Should we be addressing it in this PR?

tests/test_keydb.py
```diff
@@ -387,12 +394,21 @@ def test_create_keydb_from_root_metadata(self):
...
+    self.maxDiff = None
     self.assertEqual(rsakey2, tuf.keydb.get_key(keyid2))
-    self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid)
+
+    test_key = rsakey2
+    test_key['keyid'] = keyid
+    self.assertEqual(test_key, tuf.keydb.get_key(keyid))
+
     self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid3)
+    #self.assertRaises(tuf.exceptions.UnknownKeyError, tuf.keydb.get_key, keyid4)
```
joshuagl: Can we enable this part of the test?

tuf/client/updater.py
```diff
@@ -951,18 +951,10 @@ def _import_delegations(self, parent_role):
...
-          hash_algorithms = securesystemslib.settings.HASH_ALGORITHMS
-          securesystemslib.settings.HASH_ALGORITHMS = keyinfo['keyid_hash_algorithms']
-          key, keyids = securesystemslib.keys.format_metadata_to_key(keyinfo)
-          securesystemslib.settings.HASH_ALGORITHMS = hash_algorithms
-
-          for key_id in keyids:
-            key['keyid'] = key_id
-            tuf.keydb.add_key(key, keyid=None, repository_name=self.repository_name)
+          key['keyid'] = keyid
+          tuf.keydb.add_key(key, keyid=None, repository_name=self.repository_name)
```
```suggestion
          tuf.keydb.add_key(key, repository_name=self.repository_name)
```
the `keyid` argument defaults to `None`, so let's not pass it here.

Aside: I can't find a single caller of `keydb.add_key()` which passes a value other than `None` for the keyid 😕

tuf/keydb.py
```diff
@@ -113,30 +113,22 @@ def create_keydb_from_root_metadata(root_metadata, repository_name='default'):
...
+
+      # Make sure to update key_dict['keyid'] to use one of the other valid
+      # keyids, otherwise add_key() will have no reference to it.
       try:
-        for keyid in keyids:
-          # Make sure to update key_dict['keyid'] to use one of the other valid
-          # keyids, otherwise add_key() will have no reference to it.
-          key_dict['keyid'] = keyid
-          add_key(key_dict, keyid=None, repository_name=repository_name)
+        key_dict['keyid'] = keyid
```
joshuagl: I think `key_dict['keyid']` should already be set to `keyid` here, because we pass it in as the `default_keyid` in `format_metadata_to_key`?

tuf/keydb.py
```diff
@@ -113,30 +113,22 @@ def create_keydb_from_root_metadata(root_metadata, repository_name='default'):
...
+        add_key(key_dict, keyid=None, repository_name=repository_name)
```
joshuagl:
```suggestion
        add_key(key_dict, repository_name=repository_name)
```
The default argument for `keyid` is already `None`, no need to pass that here.

tuf/repository_lib.py
```diff
@@ -700,10 +695,8 @@ def _load_top_level_metadata(repository, top_level_filenames, repository_name):
       # repository maintainer should have also been made aware of the duplicate
       # key when it was added.
       try:
-        for keyid in keyids: #pragma: no branch
-          key_object['keyid'] = keyid
-          tuf.keydb.add_key(key_object, keyid=None,
-              repository_name=repository_name)
+        key_object['keyid'] = keyid
```
joshuagl: Again, I think this is already set correctly by `format_metadata_to_key()`?

joshuagl (changes requested):
Thanks for the submission Marina! Am I correct in thinking the behavioural changes introduced by this PR are that:
* a client no longer recalculates keyids, and instead trusts the values in the metadata
* when loading an existing repository with repository_tool, keyids are not calculated and instead the keyids from the metadata are used

I'd like to see the additional test for a duplicate keyid enabled, or removed from this PR. I also pointed had some minor comments inline about not passing the default value for an argument and what appears to be unnecessarily assigning a value to a dictionary field.

__NOTE__: the required securesystemslib change has been merged and is available in release [0.15.0](https://github.com/secure-systems-lab/securesystemslib/releases/tag/v0.15.0v), which we already depend on in py-TUF.

joshuagl:
Note the DCO check is failing because none of the commits have a `Signed-off-by:` line, the [check details](https://github.com/theupdateframework/tuf/pull/1014/checks?check_run_id=895271283) suggest the following to address that:

> git rebase HEAD~8 --signoff
> git push --force-with-lease origin remove-keyid_hash_algorithms

tests/test_keydb.py
mnm678: As keyids are no longer calculated before use, the client will want a different way to verify that all keys applied to a threshold are unique. I think this may be better left to a separate pr (which I've documented in #1084) for 2 reasons:
* This is a larger change that will require creating a standardized representation for keys
* It is only an issue if the delegating metadata makes a mistake or in the case of multi-role delegations (which are not currently in the reference implementation). Also, this issue [already exists](https://github.com/theupdateframework/tuf/blob/develop/tuf/sig.py#L89) with the current implementation if multiple `keyid_hash_algorithms` are used 

That being said, I have started looking into this issue and could include the fix here if that would be more clear. Otherwise I'll remove this test for now.

mnm678:
> Thanks for the submission Marina! Am I correct in thinking the behavioural changes introduced by this PR are that:
> 
>     * a client no longer recalculates keyids, and instead trusts the values in the metadata
> 
>     * when loading an existing repository with repository_tool, keyids are not calculated and instead the keyids from the metadata are used

Thanks for the review! Yes, those are the main behavioural changes. These changes will help both with the TAP 12 implementation and with the removal of `keyid_hash_algorithms` from TUF metadata per #848.

mnm678:
The failing test looks like a timeout issue that should be unrelated to this pr.

tests/test_keydb.py
```diff
@@ -43,7 +43,7 @@
 
 # Generate the three keys to use in our test cases.
 KEYS = []
-for junk in range(3):
+for junk in range(4):
```
joshuagl:
```suggestion
for junk in range(3):
```
I think we don't need this change now that we've dropped the WIP test in 376590d4e82ecf93e40157a7178ed27e5bb254fe?

tests/test_keydb.py
```diff
@@ -387,12 +389,20 @@ def test_create_keydb_from_root_metadata(self):
 
     self.assertEqual(None, tuf.keydb.create_keydb_from_root_metadata(root_metadata))
 
-    # Ensure only 'keyid2' was added to the keydb database.  'keyid' and
-    # 'keyid3' should not be stored.
+    # Ensure only 'keyid2' and 'keyid' were added to the keydb database.
+    # 'keyid3' and 'keyid4' should not be stored.
```
joshuagl:
```suggestion
    # 'keyid3' should not be stored.
```

joshuagl (approved):
Some minor comments in-line, but otherwise LGTM.

trishankatdatadog (commented):
💯 

Should we have an accompany `securesystemslib` PR to remove `securesystemslib.settings.HASH_ALGORITHMS`?

mnm678:
> Should we have an accompany `securesystemslib` PR to remove `securesystemslib.settings.HASH_ALGORITHMS`?

Is this setting used by other securesystemslib users, especially intoto? @SantiagoTorres @lukpueh

tests/test_keydb.py
joshuagl: I'd 

> > Should we have an accompany `securesystemslib` PR to remove `securesystemslib.settings.HASH_ALGORITHMS`?
> 
> Is this setting used by other securesystemslib users, especially intoto? @SantiagoTorres @lukpueh

Doesn't appear in a quick `git grep` of in-toto.

joshuagl:
> > Should we have an accompany `securesystemslib` PR to remove `securesystemslib.settings.HASH_ALGORITHMS`?
> 
> Is this setting used by other securesystemslib users, especially intoto? @SantiagoTorres @lukpueh

Based on a quick `git grep` of in-toto, it doesn't appear to be used there.

lukpueh:
Hi @lorenzo-blasa! What exactly are you trying to achieve? Maybe taking a look at the corresponding PR (#1014) helps to better understand the commit (3c78d675189fc13163195e183feab5d1656ac5e3)?

lorenzo-blasa:
@lukpueh many thanks! That PR definitely helps. I'm currently working on having a TUF reference implementation written in JavaScript. The effort started last year and only last week I resumed.

So I was running the local JavaScript tests and comparing that with the version I had in Python and I found it was also failing but maybe because the unit tests were actually referring the TUF package installed instead of the local copy. 

Running ` pip install -r requirements-dev.txt` made it all work but it wasn't very clear what had changed as I didn't do a very good job at tracking the revision I was using as base.

It's all good now :)

trishankatdatadog:
@lorenzo-blasa A tuf.js has been a dream of mine for a while now... you might look at my toy reimplementation [tuf-on-a-plane](https://github.com/trishankatdatadog/tuf-on-a-plane) to get a few more clarifications or ideas. Thanks for your feedback!

lorenzo-blasa:
@trishankatdatadog thanks for sharing that! 

The intention I have with the JS implementation is to open source it as soon as possible, being complete and going through necessary approvals. So it's great to know there's more people interested in it.

lukpueh:
@lorenzo-blasa, tuf.js sounds exciting, please keep us posted!  And of course, feel free to ping us if you have any questions regarding the TUF spec or this reference implementation.

Issue #1009, opened 2020-03-30
Support abstract files and directories


lukpueh:
So far we built on the assumption that both target files and TUF metadata can be loaded from and written to the local filesystem. This, however, is no necessity. In a large-scale production environment (like e.g. Python warehouse, see PEP 458) the TUF repository management code (most notably `repository_tool` and its underlying `repository_lib`) can and is likely to run on a different node than where TUF **metadata files** or **target files** reside. To support distributed operation, TUF repository code needs to be updated as outlined below.

**metadata files**
----------------

- [x] Provide an abstract file interface that supports at least reading and writing files, creating directories and listing files in a directory (implement this in `securesystemslib`).

- [x] Provide a file service that implements the abstract file interface and performs said file operations on the local file system to be used below as default file backend (implement this in `securesystemslib`).

- [x] Update TUF repository code to create a new or load an existing TUF repository, to obtain hashes and sizes of metadata files, and to persist metadata files, all using a customizable file backend.

  (*repository_lib.generate_snapshot_metadata*, *repository_lib.write_metadata_file*, *repository_tool.create_new_repository*, *repository_tool.load_repository* (**))

- [x] Update *securesystemslib* code that is currently used by TUF repository code for file operations to support the use of a customizable file backend.

  (*util.get_file_details*, *util.load_json_file*, *util.persist_temp_file*, *hash.digest_filename*(**))

- [x] Revise file existence checks (*os.path.{isfile,isdir,exists}*) in TUF repository code, and depending on which seems less invasive, or generally better suited, either
  - switch to a file system agnostic [EAFP](https://docs.python.org/3/glossary.html#term-eafp)-style (e.g. catching IOError), or
  - allow TUF to perform these checks using a customizable file backend.

  (*repository_lib.generate_targets_metadata*, *repository_lib.write_metadata_file*, *repository_lib._check_directory*, *repository_lib._delete_obsolete_metadata*, *repository_lib._load_top_level_metadata*(**))

(**) (Non-exhaustive list of probably affected functions)

**target files** 
--------------
@joshuagl and @sechkova have submitted PRs that decouple abstract targets in TUF metadata from their physical equivalents on disk. This work includes:
- [x] removal of file existence checks in user functions that add target files to the internal TUF metadata store (#1008),
- [x] support for a `fileinfo` argument on add target user functions to pass out-of-band obtained hashes and sizes of files (#1007),
[...]

PR #1024, opened 2020-04-23
Port to securesystemslib with abstract files and directories (securesystemslib PR 232)

joshuagl:
that is provided, the better.

**Fixes issue #**: this PR encapsulates step three of #1009

* port to securesystemslib with abstract filesystem backend
  * Update TUF repository code to create a new or load an existing TUF repository, to obtain hashes and sizes of metadata files, and to persist metadata files, all using a customizable file backend
* Drop support for `tuf.settings.CONSISTENT_METHOD`

**NOTE**: this change shouldn't land until we've merged secure-systems-lab/securesystemslib#232 and made a release.

**UPDATE**: prior to this PR `_delete_obsolete_metadata()` would `os.walk()` the metadata directory, thus picking up any metadata files in subdirectories of the metadata directory. The change in this PR to use `list_folder()` from the `StorageBackendInterface` means that _should_ there be any sub-directories of the metadata directory which contain metadata, the metadata in those directories would not be removed by the changed version of `_delete_obsolete_metadata()`. 
This should not be a problem because, by default, all metadata files exist as children of the metadata directory, rather than in subdirectories.

requirements**:

### DIFF
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -598,7 +598,8 @@ def _load_top_level_metadata(repository, top_level_filenames, repository_name):
         repository_name=repository_name)
 
   except securesystemslib.exceptions.StorageError:
-    logger.debug('Cannot load the Timestamp  file: ' + repr(timestamp_filename))
+    raise tuf.exceptions.RepositoryError('Cannot load the Timestamp file: '
+        + repr(timestamp_filename))
 
   # Load 'snapshot.json'.  A consistent snapshot.json must be calculated if
   # 'consistent_snapshot' is True.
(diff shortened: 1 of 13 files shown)


joshuagl:
Now that we have a securesystemslib release with the `storage` module, this PR is ready for review.

tuf/repository_lib.py
```diff
@@ -56,6 +55,8 @@
 import iso8601
 import six
 
+import securesystemslib.storage
```
mnm678, line 58: Could this go up with the other ssl imports?

mnm678 (approved):
One nit, but otherwise LGTM

tuf/repository_lib.py
joshuagl: Good catch, thanks. I've fixed this and squashed the change into "Port to new securesystemslib w abstract filesystem" 0214302

tuf/repository_lib.py
```diff
@@ -608,8 +553,9 @@ def _load_top_level_metadata(repository, top_level_filenames, repository_name):
     tuf.roledb.update_roleinfo('timestamp', roleinfo, mark_role_as_dirty=False,
         repository_name=repository_name)
 
-  else:
-    logger.debug('Cannot load the Timestamp  file: ' + repr(timestamp_filename))
+  except securesystemslib.exceptions.StorageError:
+    raise tuf.exceptions.RepositoryError('Cannot load the Timestamp  file: '
```
woodruffw: Tiny nit (looks like it was carried over from the old string): double spacing between `Timestamp` and `file`.

woodruffw (approved):
LGTM!

tuf/repository_lib.py
joshuagl: Good catch, thanks. Corrected in a187377.

Issue #1010, opened 2020-04-02
Problems running tuf unit tests against local version of securesyslib


MVrachev:
**Description of issue or feature request**:
While I was working on issue https://github.com/secure-systems-lab/securesystemslib/issues/220 I wanted to approach it that way:
1. create and use editable venv in `tuf` 
2. install the development dependencies
3. make the necessary changes in `securesyslib`
4. adapt `tuf` with the changes made in `securesyslib`
5. submit a pr to `securesyslib`
6. submit a pr to `tuf`

On step 4 I ran the tests with the tox command I received  `ConnectionRefusedError: [Errno 111] Connection refused` errors when running the unit tests.
Then I stepped back, stashed all my changes and ran all of the tests again and the same happened.
I will paste more details from the logs below.

This error doesn't encounter when running the tests with `tox -e py37` for example.
It seems to me that I am missing something when creating my editable venv.

**Steps to reproduce the error**:

1. Clone `securesyslib` when the top of the master branch is:  https://github.com/secure-systems-lab/securesystemslib/commit/caf029ea61339dc5fa9beedf230ca1d026129169
2. Clone `tuf` when the top of the master branch is: https://github.com/theupdateframework/tuf/commit/e095112be7d57b26406469b767b7ef8561a8f6a9
3. In the tuf folder run: 
```
$ python3.6 -m venv . .editable_venv
$ source .editable_venv/bin/activate
```
4. From the securesyslib folder run `pip3.6 install -r requirements-dev.txt`
5. From the tuf folder run `pip3.6 install -r requirements-dev.txt`
6. Finally, from the tuf directory, without changing the source code of `securesyslib` or `tuf`, run:

```
(.editable_venv) $ cd ./tests
(.editable_venv) $ python3.6 aggregate_tests.py
```

and then you will see the following errors:
```
..................x..............................................................................E
Stderr:
DEBUG:tuf.client.updater:Adding roles delegated from 'targets'.
DEBUG:tuf.client.updater:Adding delegated role: role1.
DEBUG:tuf.client.updater:Adding roles delegated from 'targets'.
DEBUG:tuf.client.updater:Adding delegated role: role1.
DEBUG:tuf.client.updater:Interrogating repository: 'test_repository1'
DEBUG:tuf.client.updater:Interrogating repository: 'test_repository1'
DEBUG:tuf.client.updater:Found local directory for 'test_repository1'
DEBUG:tuf.client.updater:Interrogating repository: 'test_repository1'
[...]

PR #1029, opened 2020-05-05
Fix error "[Errno 111] Connection refused" and make logs more usable

MVrachev:
**Fixes issue #**: #1010 

When running the tests this error appears
"[Errno 111] Connection refused". After some digging Lukas
found another error "No module named tests.simple_server"
which is the root cause for error 111.

With the help of Joshua Lock, we found out that the simple_server.py
was not being found because of subprocess.Popen was being passed
a cwd kwarg which moved the current working directory
away from tuf/tests.
That's what I fix with the first commit.

In my second commit, I remove stderr=subprocess.PIPE argument from subprocess.Popen function calls and make the  QuietHTTPRequestHandler as the default HTTP handler in the `simple_server.py` script.
This change is inspired by the digging needed to discover the real problem causing issue 1010 summarized well in the @lukpueh comment https://github.com/theupdateframework/tuf/issues/1010#issuecomment-612064378.

requirements**:

### DIFF
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -71,7 +71,15 @@ class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
 if six.PY2 and platform.system() == 'Windows':
   handler = QuietHTTPRequestHandler
 else:
-  handler = SimpleHTTPRequestHandler
+  use_quiet_http_request_handler = True
+
+  if len(sys.argv) > 2:
+    use_quiet_http_request_handler = sys.argv[2]
+
+  if use_quiet_http_request_handler:
+    handler = QuietHTTPRequestHandler
+  else:
+    handler = SimpleHTTPRequestHandler
 
 httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
 
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -66,20 +66,19 @@ class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
 
 # NOTE: On Windows/Python2 tests that use this simple_server.py in a
 # subprocesses hang after a certain amount of requests (~68), if a PIPE is
-# passed as Popen's stderr argument. As a simple workaround we silence the
-# server on those Windows/Py2 to not fill the buffer.
-if six.PY2 and platform.system() == 'Windows':
+# passed as Popen's stderr argument. This problem doesn't emerge if
+# we silence the HTTP messages.
+# If you decide to receive the HTTP messages, then this bug
+# could reappear.
+use_quiet_http_request_handler = True
+
+if len(sys.argv) > 2:
+  use_quiet_http_request_handler = sys.argv[2]
+
+if use_quiet_http_request_handler:
   handler = QuietHTTPRequestHandler
 else:
-  use_quiet_http_request_handler = True
-
-  if len(sys.argv) > 2:
-    use_quiet_http_request_handler = sys.argv[2]
-
-  if use_quiet_http_request_handler:
-    handler = QuietHTTPRequestHandler
-  else:
-    handler = SimpleHTTPRequestHandler
+  handler = SimpleHTTPRequestHandler
 
 httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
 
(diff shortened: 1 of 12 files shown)


mnm678 (commented):
Thanks for the fix, this looks good to me.

@joshuagl will the calls to `os.getcwd()` be affected by the abstract file system work in #1024?

tests/simple_server.py
```diff
@@ -70,6 +70,13 @@ def log_request(self, code='-', size='-'):
 # server on those Windows/Py2 to not fill the buffer.
 if six.PY2 and platform.system() == 'Windows':
   handler = QuietHTTPRequestHandler
+
+use_quiet_http_request_handler = True
+if len(sys.argv) > 2:
+  use_quiet_http_request_handler = sys.argv[2]
+
+if use_quiet_http_request_handler == True:
```
joshuagl: As this is Python and we expect a boolean value we can just test the variable like:

```suggestion
if use_quiet_http_request_handler:
```

See PEP 8's [Programming Recommendations](https://www.python.org/dev/peps/pep-0008/#id51) and the [Book of Python Anti-Patterns](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html)

tests/simple_server.py
```diff
@@ -70,6 +70,13 @@ def log_request(self, code='-', size='-'):
 # server on those Windows/Py2 to not fill the buffer.
 if six.PY2 and platform.system() == 'Windows':
   handler = QuietHTTPRequestHandler
```
joshuagl: I don't think we still need this logic above if we now default to using `QuietHTTPRequestHandler`? Certainly we don't want two ways to change the HTTP Request Handler for this simple test server.

joshuagl (changes requested):
Thanks for the solid changes to fix this issue @MVrachev.

A few minor suggestions and this is good to go:
* We only need the change to invoke `simple_server.py` with _cwd_ for _tests/test_multiple_repositories_integration.py_ and _tests/test_updater.py_ – could you restore the other tests to their current behaviour of just calling _simple\_server.py_ ?
* I added some comments to _simple\_server.py_ itself to keep that code short and idiomatic, could you please address those?

Thanks

joshuagl:
> @joshuagl will the calls to `os.getcwd()` be affected by the abstract file system work in #1024?

No, these calls are restricted to test code and are to ensure that we have the tests/ directory in order to a) serve the test repo data and b) ensure we are invoking simple_server.py in a way which won't cause errors when there are multiple tests/ directories in the _PYTHONPATH_ (as is the case when we're running with editable securesystemslib and editable tuf installations).

tests/simple_server.py
MVrachev: I didn't want to remove this check so that this strange case for Windows is documented.

If I remove it now and someone change this piece of code again this bug can reappear.

joshuagl: I can see the logic in that, we can either leave the comment and remove the code or consolidate this logic with your change. We should absolutely not set the handler in two places which can conflict with each other. As it stands if I explicitly set `use_quiet_http_request_handler = False` on Windows, I still get `QuietHTTPRequestHandler`.

tests/simple_server.py
MVrachev: Yea, I just wanted to be more explicit, but you are right. 
I will address this.

tests/simple_server.py
MVrachev: Yes, you are right. I will group those checks together so that we don't change the state in two places.

I am wondering which should be with higher priority: `use_quiet_http_request_handler` value or the check if we are on Windows?
I think that if we are on Windows should be with higher priority given that we have bug reported for that specific case.

MVrachev:
> * We only need the change to invoke `simple_server.py` with _cwd_ for _tests/test_multiple_repositories_integration.py_ and _tests/test_updater.py_ – could you restore the other tests to their current behaviour of just calling _simple_server.py_ ?

Yes, you are right. I updated my pr with the above suggestions and the ones left alongside the code.

joshuagl (approved):
Thanks for the changes @MVrachev! Great to have this issue resolved. 🎉

Issue #996, opened 2020-03-11
Make length and hashes optional in snapshot/timestamp metadata


lukpueh:
**Description of issue or feature request**:

As per the specification ([v1.0.1](https://github.com/theupdateframework/specification/releases/tag/v1.0.1)) `length` and `hashes` fields in timestamp and snapshot metadata are optional. The reference implementation should reflect this.

**Current behavior**:
- timestamp metadata always includes length and hashes of snapshot metadata (see `reposoitory_lib.generate_timestamp_metadata`)
- snapshot metadata always excludes length and hashes of (delegated) targets metadata (see `repository_lib.generate_snapshot_metadata`)

**Expected behavior**:
Add option to include/exclude these fields in the relevant functions in `repository_lib`. Since these functions are usually called via `write/writeall` the option should be made available there too. It's also conceivable to configure this via attributes on the Repository object.

MVrachev:
I will work on that issue.

I have one question though: I am planning to add additional parameters to  `generate_timestamp_metadata` and `generate_snapshot_metadata`  which will be used to decide whenever `hashes` and `length` should be used, but my question is what should be their default value?

lukpueh:
> what should be their default value?

I suggest to choose them so that the default behavior of these functions matches their current behavior.

joshuagl:
My initial feeling was that we should default to the current behaviour, but I wonder whether there's an argument to be made for _not_ providing default values. If we don't have default values, we force implementers to explicitly choose whether they want to include lengths and hashes, rather than inheriting a default.

Furthermore I think the specification suggesting the fields are optional implies that they would not be included by default, and thus would expect similar behaviour from a reference implementation.

Conversely, I think the current values probably are sensible defaults, particularly for i.e. the PEP 458 implementation. I'm curious to read what others think.

MVrachev:
I feel like security by default should be implied here and use `length` and `hashes` fields by default.
That way if somebody decides to not use them then he/she should make it explicit by passing the `use_length=False` and `use_hashes=False` when calling the functions.

MVrachev:
@joshuagl  can you close this one? It's fixed by the above pr.

joshuagl:
Fixed in #1031

Issue #955, opened 2019-11-12
Revise intended and actual behavior of  Repository's `status()` method


lukpueh:
**Description of issue or feature request**:

The [`status()`](https://github.com/theupdateframework/tuf/blob/9b93c0bcccffef25b7603318297b2504ede3fd33/tuf/repository_tool.py#L379-L421)-method on the `repository_tool.Repository`-class, wraps [`repository_lib._log_status_of_top_level_roles`](https://github.com/theupdateframework/tuf/blob/9b93c0bcccffef25b7603318297b2504ede3fd33/tuf/repository_lib.py#L1776-L1919), and logs status information about the repository.

**Current behavior:**

This is a non-comprehensive list of concerns identified in these functions:

* `status` is a non-intuitive name for a function that changes and then logs the changed status. E.g. when calling `status` on unsigned metadata, `status` signs the metadata and informs you about the existence of those signatures.

* The control flow is unclear, e.g. `_log_status_of_top_level_roles` may log the following information for any top level role:

  (1) `"ROLENAME role contains K / N public keys"`
  (2) `"ROLENAME role contains K / N signing keys"`
  (3) `"ROLENAME role contains K / N signatures"`

  However the information is shown:
  (1) for roles that have `K < N` public keys,
  (2) for roles that have `K >= N` public keys, and no signatures, and `K < N` signing keys,
  (3) for roles that are processed before `_generate_and_write_metadata` and `_log_status` fails for a role, including the role that failed. The order of processed roles is: root, targets, snapshot, timestamp.

* There is unnecessarily complicated and repetitive code, e.g.:
  ```python
  # This construct is repeated for every role
  root_is_dirty = None
  if 'root' in dirty_rolenames:
    root_is_dirty = True

  else:
    root_is_dirty = False
  ```
  ...should be ...
  ```python
  root_is_dirty = False
  if 'root' in dirty_rolenames:
    root_is_dirty = True
  ```
  ... or just ...
  ```python
  root_is_dirty = 'root' in dirty_rolenames
  ```

  And since this construct is repeated for every top-level role, the best way would be a helper function. 

  ```python
   is_role_dirty(rolename)
  ```

* The separation of `status` and `_log_status_of_top_level_roles` function is not obvious. 
[...]

PR #1031, opened 2020-05-13
Make length and hashes optional for timestamp and snapshot roles

MVrachev:
that is provided, the better.

**Fixes issue #**: https://github.com/theupdateframework/tuf/issues/996

As per the tuf specification (v1.0.1) length and hashes fields are optional for
the timestamp.json and snapshot.json.

There could be a discussion around what should be the default.
My opinion is that we should apply the rule `security by default` here and use hashes and length by default.
That way if somebody decides to not use them then he/she should make it explicit through passing the use_length=False and use_hashes=False arguments when calling the respective functions.

requirements**:


tests/test_repository_lib.py
```diff
@@ -481,6 +481,184 @@ def test_generate_snapshot_metadata(self):
...
+
+          # Check that hashes is not calculated but length is
+          self.assertTrue(
+            metadata_files_info_dict[metadata_filename].get('length') is not None)
+          self.assertTrue(
+            metadata_files_info_dict[metadata_filename].get('hashes') is None)
+
+
+
+  def test_generate_snapshot_metadata_without_hashes_and_length(self):
```
trishankatdatadog: Would it be possible to reuse the logic from the individual tests above?

tests/test_repository_lib.py
```diff
@@ -526,6 +704,120 @@ def test_generate_timestamp_metadata(self):
...
+
+    # Check that hashes is not calculated but length is
+    timestamp_file_info = timestamp_metadata['meta']
+
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('length') is not None)
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('hashes') is None)
+
+
+
+  def test_generate_timestamp_metadata_without_length_and_hashes(self):
```
trishankatdatadog, line 688: Ditto: would it be possible to reuse the logic from the individual tests above?

tuf/repository_lib.py
```diff
@@ -1506,9 +1523,27 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
       # list these roles found in the metadata directory.
       if tuf.roledb.role_exists(rolename, repository_name) and \
           rolename not in ['root', 'snapshot', 'timestamp', 'targets']:
-        fileinfodict[metadata_name] = get_metadata_versioninfo(rolename,
+
+        length = None
```
trishankatdatadog, line 1648: I think we could remove these two assignments and the IF statement altogether.

tuf/repository_lib.py
```diff
@@ -1506,9 +1523,27 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
+        hashes = None
+        if use_length or use_hashes:
+
+          length, hashes = securesystemslib.util.get_file_details(
+            os.path.join(metadata_directory, metadata_filename),
+            tuf.settings.FILE_HASH_ALGORITHMS)
+
+          if use_length == False:
```
trishankatdatadog: Maybe replace this with a conditional assignment like:

`length = (use_length and length) or None`

tuf/repository_lib.py
```diff
@@ -1506,9 +1523,27 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
+            length = None
+
+          if use_hashes == False:
```
trishankatdatadog: Ditto.

tuf/repository_lib.py
```diff
@@ -1577,11 +1622,23 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
...
   securesystemslib.formats.ISO8601_DATETIME_SCHEMA.check_match(expiration_date)
   securesystemslib.formats.NAME_SCHEMA.check_match(repository_name)
+  securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_length)
+  securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_hashes)
 
   # Retrieve the versioninfo of the Snapshot metadata file.
   snapshot_fileinfo = {}
-  length, hashes = securesystemslib.util.get_file_details(snapshot_filename,
-      tuf.settings.FILE_HASH_ALGORITHMS)
+  length = None
```
trishankatdatadog: See my comments above for timestamp, please.

trishankatdatadog (changes requested):
This is great, thanks, Martin!

MVrachev:
Thank you, @trishankatdatadog for your comprehensive review!

I decided to rebase the particular commits, address your comments, and make them ready for merge.
The changes I made are the following:

1.  Used `length = (use_length and length) or None` when calculating the hashes and length in both `generate_snapshot_metadata` and `generate_timestampt_metadata`.

2. Add code in generate_snapshot_metadata function to optionally calculate the hashes and length of the targets file.

3. After step 2 the tests were failing and I had to start using `targets_filename` variable as a full filename with its suffix .json and that's why I created an additional commit.

4. Created `_setup_generate_snapshot_metadata_test` and `_setup_generate_timestamp_metadata_test` functions to reuse setup logic in the `generate_snapshot_metadata` tests and `generate_timestampt_metadata` tests.

tests/test_repository_lib.py
MVrachev: I rebased and addressed that comment.

MVrachev: I rebased and addressed that comment.

tests/test_repository_lib.py
MVrachev: I rebased and addressed that comment.

tuf/repository_lib.py
MVrachev: I rebased and addressed that comment.

tuf/repository_lib.py
MVrachev: I rebased and addressed that comment.

tuf/repository_lib.py
MVrachev: I rebased and addressed that comment.

tuf/formats.py
```diff
@@ -232,8 +232,8 @@
 # and sha512 may be computed for the same file and stored).
 FILEINFO_SCHEMA = SCHEMA.Object(
   object_name = 'FILEINFO_SCHEMA',
-  length = LENGTH_SCHEMA,
-  hashes = HASHDICT_SCHEMA,
+  length = SCHEMA.Optional(LENGTH_SCHEMA),
```
trishankatdatadog, line 243: Not sure this is a good idea. Because in `targets`, they are necessary, but not so in `timestamp` or `snapshot`, so need to distinguish between these two schemas.

tuf/formats.py
```diff
@@ -793,7 +793,13 @@ def make_fileinfo(length, hashes, version=None, custom=None):
     information of a metadata or target file.
   """
 
-  fileinfo = {'length' : length, 'hashes' : hashes}
+  fileinfo = {}
+
+  if length is not None:
```
trishankatdatadog: I think we could just use the "truthey" check here (`if length:`), and in the following IF checks, which excludes all "falsey" values at least

trishankatdatadog (changes requested):
This is great, thanks Martin!

I have one major reservation: the `FILEINFO_SCHEMA` between `timestamp`/`snapshot` and `targets` metadata should be distinguished...

tuf/formats.py
joshuagl: Is it valid to have zero-length files in a repository? For length especially it feels safer to explicitly check for `None`?

tuf/formats.py
joshuagl: Absolutely, as @trishankatdatadog has said. From the specification we can see that:
* in _snapshot.json_, lengths and hashes of the targets metadata files (top-level and delegations) are optional
* in _targets.json_ (and delegated targets metadata), lengths and hashes of the target files are required
* in _timestamp.json_, lengths and hashes of the snapshot.json file are optional

Perhaps we can add a `METADATA_FILEINFO_SCHEMA` to formats.py for use here?

trishankatdatadog: Yes, if so, perhaps rename the original to `TARGETS_FILEINFO_SCHEMA`, just to clarify

tuf/formats.py
trishankatdatadog: I feel like this line can be simplified to `fileinfo['length'] = length or None`, because we really have too many schema checks and such all over the place, which is not necessary for writing metadata (makes code clunky, especially in a dynamically typed language) so much as reading them. It is the readers that should be the most paranoid, and so where the code complexity should be reserved. What do others think?

joshuagl: Agreed, `fileinfo['length'] = length or None` is concise and readable.

MVrachev: Yea, I agree about that `fileinfo['length'] = length or None` is a good approach.
Maybe it would be good to change the checks for the other parameters and make them one-liners as well?

MVrachev: I just realized that by using `fileinfo['length'] = length or None` there is an unexpected side effect.

If I use the proposed one-liner to check all arguments and I call `make_fileinfo` with all arguments with a value of `None` then I will end up with the dictionary:
`fileinfo = {"length": None, "hashes": None, "version": None, "custom": None}`

but if I use the if statements and call the function the same way I will end up with just:
`fileinfo = {}`.

Which of the above effects do we want to achieve?

MVrachev: And I noticed another problem.
If I decide to use this one-liner `fileinfo['length'] = length or None` only for length and run the tests I encounter this problem in 6 of the tests:
image

The problem is caused by the same fact I pointed out in the previous comment. 
At the end of the `make_fileinfo` there is this check `FILEINFO_SCHEMA.check_match(fileinfo)`.
File info dictionary during this check is `fileinfo = {"length": None, "hashes": X, "version": Y, "custom": Z}`.

In `check_match` function `fileinfo["hashes"]` is found and is checked again recursively by the same function:
image

but because the value is none, then the above error is thrown.

tuf/formats.py
MVrachev: I will rename the original `FILE_INFO_SCHEMA` to `TARGETS_FILEINFO_SCHEMA` as suggested, but there is no need to add an additional `METADATA_FILEINFO_SCHEMA`.
There is one schema created by you @joshuagl around 2 months ago: 
```
LOOSE_FILEINFO_SCHEMA = SCHEMA.Object(
  object_name = "LOOSE_FILEINFO_SCHEMA",
  length = SCHEMA.Optional(LENGTH_SCHEMA),
  hashes = SCHEMA.Optional(HASHDICT_SCHEMA),
  version = SCHEMA.Optional(METADATAVERSION_SCHEMA),
  custom = SCHEMA.Optional(SCHEMA.Object())
)
```
https://github.com/theupdateframework/tuf/blob/d7aec6a5f9bbf0eafb7f075e3286d9e8db4c92c9/tuf/formats.py#L307 which could do the work.

joshuagl: Version is not optional for snapshot and timestamp, therefore `LOOSE_FILEINFO_SCHEMA` could allow incorrectly formatted metadata.

Aside: I'm not a fan of `LOOSE_FILEINFO_SCHEMA`, I'm hoping we can remove it as we work on refactoring tuf.

MVrachev: Yes, you are right. Version is not optional for timestamp and snapshot, it seems I missed it because those two look similar.
I made the same mistake with the initial changes too.

Then, I will create a new schema. I was thinking of a name like `FILEINFO_OPTIONAL_FIELDS_SCHEMA` but it's too long...
Other naming suggestions?

joshuagl: I'd propose we:
* rename FILEINFO_SCHEMA -> TARGETS_FILEINFO_SCHEMA
* rename LOOSE_FILEINFO_SCHEMA -> LOOSE_TARGETS_FILEINFO_SCHEMA
* add a new METADATA_FILEINFO_SCHEMA with optional length, optional hashes and version (not optional) fields.

tuf/formats.py
joshuagl: Ah, right. I should have looked at the context for this change/discussion more. Apologies.

I would expect that `make_fileinfo()` becomes `make_targets_fileinfo()` and remains as currently implemented (only changing `FILEINFO_SCHEMA` -> `TARGETS_FILEINFO_SCHEMA`, per the discussion above).

You could add a new helper, `make_metadata_fileinfo(version, length=None, hashes=None)` and then do as Trishank suggested above and check for truthy values, i.e.:
```python
if hashes:
  fileinfo['hashes'] = hashes
```

tuf/formats.py
MVrachev: Yes, I agree with those three points and I will add them.

MVrachev: @joshuagl I addressed your comments in a new commit.

tuf/formats.py
MVrachev: @joshuagl I addressed your comments in a new commit.

MVrachev:
I will fix my merge conflicts when the tests on the develop branch are fixed and pass.

MVrachev:
I just found out that the tests on the development branch are actually passing, but I had problems when I run `tox -e py37` locally.
It seems that tox has cached an older version of securesislib - 0.14.2 and it was using it instead of the latest securesyslib version 0.15.0 when running `tox -e py37`.

Sorry, it was my problem.

MVrachev:
I fixed the merge conflicts.

This commit is big and contains the following changes:


1. Rename FILE_INFO_SCHEMA to TARGETS_FILEINFO_SCHEMA 

2. Add METADATA_FILEINFO_SCHEMA used from snapshot and timestamp 

3. Rename LOOSE_FILEINFO_SCHEMA -> LOOSE_TARGETS_FILEINFO_SCHEMA 

4. Rename make_targets_fileinfo and use it when targets metadata have to be calculated and length and hashes are mandatory 

5. Create make_metadata_fileinfo function which will be used for snapshot and timestamp and fix all places that are calling it. 

6. Rename get_metadata_fileinfo to  get_targets_metadata_fileinfo because it's used only in the _generate_targets_fileinfo where hashes and length are mandatory. 

7. Create unit tests for make_metadata_fileinfo

The code is ready for another review.

tests/test_formats.py
```diff
@@ -797,28 +801,65 @@ def test_make_signable(self):
...
+  def test_make_targets_fileinfo(self):
+    # Test conditions for valid arguments.
+    length = 1024
+    hashes = {'sha256': 'A4582BCF323BCEF', 'sha512': 'A4582BCF323BFEF'}
+    version = 8
+    custom = {'type': 'paintjob'}
+
+    TARGETS_FILEINFO_SCHEMA = tuf.formats.TARGETS_FILEINFO_SCHEMA
+    make_targets_fileinfo = tuf.formats.make_targets_fileinfo
+    self.assertTrue(TARGETS_FILEINFO_SCHEMA.matches(make_targets_fileinfo(length, hashes, version, custom)))
```
joshuagl: See below, but as I understand it we don't keep a _version_ for targets files

tests/test_formats.py
```diff
@@ -797,28 +801,65 @@ def test_make_signable(self):
...
     custom = {'type': 'paintjob'}
 
-    FILEINFO_SCHEMA = tuf.formats.FILEINFO_SCHEMA
-    make_fileinfo = tuf.formats.make_fileinfo
-    self.assertTrue(FILEINFO_SCHEMA.matches(make_fileinfo(length, hashes, version, custom)))
-    self.assertTrue(FILEINFO_SCHEMA.matches(make_fileinfo(length, hashes)))
+    METADATA_FILEINFO_SCHEMA = tuf.formats.METADATA_FILEINFO_SCHEMA
+    make_metadata_fileinfo = tuf.formats.make_metadata_fileinfo
+    self.assertTrue(METADATA_FILEINFO_SCHEMA.matches(make_metadata_fileinfo(
+      version, length, hashes, custom)))
```
joshuagl: We shouldn't have a _custom_ field for metadata files. See more below.

tests/test_repository_lib.py
```diff
@@ -493,7 +501,107 @@ def test_generate_snapshot_metadata(self):
...
+      if metadata_filename[0].isdigit():
+        version_number, metadata_filename = metadata_filename.split('.', 1)
+
+      if metadata_filename.endswith('.json'):
+        if metadata_filename not in \
+            ['root.json', 'targets.json', 'timestamp.json', 'snapshot.json']:
+
+          # Check that length is not calculated but hashes is
+          self.assertTrue(
+            metadata_files_info_dict[metadata_filename].get('length') is None)
```
joshuagl: This would be a bit cleaner if we used the `assertIn` method:
```suggestion
          self.assertIn('length', metadata_files_info_dict[metadata_filename])
```

tests/test_repository_lib.py
```diff
@@ -493,7 +501,107 @@ def test_generate_snapshot_metadata(self):
...
+          self.assertTrue(
+            metadata_files_info_dict[metadata_filename].get('hashes') is not None)
```
joshuagl: Each of the assert*() methods in unittest has an equivalent falsey version. However, I think this would be even cleaner and clearer with [`assertNotIn`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn)

```suggestion
          self.assertNotIn('hashes', metadata_files_info_dict[metadata_filename])
```

tests/test_repository_lib.py
```diff
@@ -493,7 +501,107 @@ def test_generate_snapshot_metadata(self):
...
+      # 1.root.json. The prefix number should be removed.
+      if metadata_filename[0].isdigit():
+        version_number, metadata_filename = metadata_filename.split('.', 1)
+
+      if metadata_filename.endswith('.json'):
+        if metadata_filename not in \
+            ['root.json', 'targets.json', 'timestamp.json', 'snapshot.json']:
+
+          # Check that hashes is not calculated but length is
+          self.assertTrue(
```
joshuagl: as above, let's use `assertIn` here

tests/test_repository_lib.py
```diff
@@ -493,7 +501,107 @@ def test_generate_snapshot_metadata(self):
...
+            metadata_files_info_dict[metadata_filename].get('length') is not None)
+          self.assertTrue(
```
joshuagl: and `assertNotIn` here

tests/test_repository_lib.py
```diff
@@ -493,7 +501,107 @@ def test_generate_snapshot_metadata(self):
...
+      # In the metadata_directory, there are files with format:
+      # 1.root.json. The prefix number should be removed.
+      if metadata_filename[0].isdigit():
+        version_number, metadata_filename = metadata_filename.split('.', 1)
+
+      if metadata_filename.endswith('.json'):
+        if metadata_filename not in \
+            ['root.json', 'targets.json', 'timestamp.json', 'snapshot.json']:
+
+          # Check that both length and hashes are not is not calculated
```
joshuagl:
```suggestion
          # Check that both length and hashes are not calculated
```

tests/test_repository_lib.py
```diff
@@ -493,7 +501,107 @@ def test_generate_snapshot_metadata(self):
...
+          self.assertTrue(
+            metadata_files_info_dict[metadata_filename].get('length') is None)
+          self.assertTrue(
+            metadata_files_info_dict[metadata_filename].get('hashes') is None)
```
joshuagl: we can use `assertNotIn` for both of these tests

tests/test_repository_lib.py
```diff
@@ -538,6 +653,51 @@ def test_generate_timestamp_metadata(self):
...
+
+    timestamp_metadata = repo_lib.generate_timestamp_metadata(snapshot_filename,
+        version, expiration_date, repository_name, use_length=False)
+    self.assertTrue(tuf.formats.TIMESTAMP_SCHEMA.matches(timestamp_metadata))
+
+    # Check that length is not calculated but hashes is
+    timestamp_file_info = timestamp_metadata['meta']
+
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('length') is None)
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('hashes') is not None)
```
joshuagl: Please use `assertIn` and `assertNotIn` here

tests/test_repository_lib.py
```diff
@@ -538,6 +653,51 @@ def test_generate_timestamp_metadata(self):
...
+
+    timestamp_metadata = repo_lib.generate_timestamp_metadata(snapshot_filename,
+        version, expiration_date, repository_name, use_hashes=False)
+    self.assertTrue(tuf.formats.TIMESTAMP_SCHEMA.matches(timestamp_metadata))
+
+    # Check that hashes is not calculated but length is
+    timestamp_file_info = timestamp_metadata['meta']
+
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('length') is not None)
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('hashes') is None)
```
joshuagl: Same again here, `assertIn` / `assertNotIn`

tests/test_repository_lib.py
```diff
@@ -538,6 +653,51 @@ def test_generate_timestamp_metadata(self):
...
+      self._setup_generate_timestamp_metadata_test()
+
+    timestamp_metadata = repo_lib.generate_timestamp_metadata(snapshot_filename,
+        version, expiration_date, repository_name, use_hashes=False, use_length=False)
+    self.assertTrue(tuf.formats.TIMESTAMP_SCHEMA.matches(timestamp_metadata))
+
+    # Check that length and hashes attributes are not added
+    timestamp_file_info = timestamp_metadata['meta']
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('length') is None)
+    self.assertTrue(timestamp_file_info['snapshot.json'].get('hashes') is None)
```
joshuagl: another candidate for `assertIn` / `assertNotIn`

tuf/formats.py
```diff
@@ -230,33 +230,42 @@
...
   custom = SCHEMA.Optional(SCHEMA.Object()))
 
+# Information about snapshot and timestamp files. This schema allows for optional
+# length and hashes, but version is mandatory.
+METADATA_FILEINFO_SCHEMA = SCHEMA.Object(
+  object_name = 'METADATA_FILEINFO_SCHEMA',
+  length = SCHEMA.Optional(LENGTH_SCHEMA),
+  hashes = SCHEMA.Optional(HASHDICT_SCHEMA),
+  version = METADATAVERSION_SCHEMA,
+  custom = SCHEMA.Optional(SCHEMA.Object()))
```
joshuagl: `METADATA_FILEINFO_SCHEMA` should not include a custom field, that property is only valid for targets metadata.

tuf/formats.py
```diff
@@ -230,33 +230,42 @@
 # Information about target files, like file length and file hash(es).  This
 # schema allows the storage of multiple hashes for the same file (e.g., sha256
 # and sha512 may be computed for the same file and stored).
-FILEINFO_SCHEMA = SCHEMA.Object(
-  object_name = 'FILEINFO_SCHEMA',
+TARGETS_FILEINFO_SCHEMA = SCHEMA.Object(
+  object_name = 'TARGETS_FILEINFO_SCHEMA',
   length = LENGTH_SCHEMA,
   hashes = HASHDICT_SCHEMA,
   version = SCHEMA.Optional(METADATAVERSION_SCHEMA),
```
joshuagl: As I understand it, version does not make sense for a targets file. The METAFORMAT for targets in the spec is:
```
 { TARGETPATH : {
       "length" : LENGTH,
       "hashes" : HASHES,
       ("custom" : { ... }) }
   , ...
 }
```

let's drop this field.

tuf/formats.py
```diff
@@ -298,14 +307,14 @@
...
-# developer tools.  The fields match that of FILEINFO_SCHEMA, only all
+# developer tools.  The fields match that of TARGETS_FILEINFO_SCHEMA, only all
 # fields are optional.
 CUSTOM_SCHEMA = SCHEMA.DictOf(
   key_schema = SCHEMA.AnyString(),
   value_schema = SCHEMA.Any()
 )
-LOOSE_FILEINFO_SCHEMA = SCHEMA.Object(
-  object_name = "LOOSE_FILEINFO_SCHEMA",
+LOOSE_TARGETS_FILEINFO_SCHEMA = SCHEMA.Object(
```
joshuagl, line 314: Thank you for renaming this.

tuf/formats.py
```diff
@@ -773,14 +782,11 @@ def make_fileinfo(length, hashes, version=None, custom=None):
       A dict of hashes in 'HASHDICT_SCHEMA' format, which has the form:
        {'sha256': 123df8a9b12, 'sha512': 324324dfc121, ...}
 
-    version:
-      An optional integer representing the version of the file.
-
```
joshuagl, line 778: You've removed the documentation for the _version_ argument, but not the argument itself.

tuf/formats.py
```diff
@@ -789,11 +795,11 @@ def make_fileinfo(length, hashes, version=None, custom=None):
     will raise a 'securesystemslib.exceptions.FormatError' exception.
 
   <Returns>
-    A dictionary conformant to 'FILEINFO_SCHEMA', representing the file
+    A dictionary conformant to 'TARGETS_FILEINFO_SCHEMA', representing the file
     information of a metadata or target file.
   """
 
-  fileinfo = {'length' : length, 'hashes' : hashes}
+  fileinfo = {'length' : length, 'hashes': hashes}
```
joshuagl: Nit: you've changed the whitespace for one of these mappings, but not both. Was that intentional?

tuf/formats.py
```diff
@@ -802,7 +808,59 @@ def make_fileinfo(length, hashes, version=None, custom=None):
...
+
+    length:
+      An optional integer representing the size of the file.
+
+    hashes:
+      An optional dict of hashes in 'HASHDICT_SCHEMA' format, which has the form:
+       {'sha256': 123df8a9b12, 'sha512': 324324dfc121, ...}
+
+    custom:
+      An optional object providing additional information about the file.
```
joshuagl: As mentioned above, please remove the custom field.

tuf/formats.py
```diff
@@ -802,7 +808,59 @@ def make_fileinfo(length, hashes, version=None, custom=None):
...
 
   # Raise 'securesystemslib.exceptions.FormatError' if the check fails.
-  FILEINFO_SCHEMA.check_match(fileinfo)
+  TARGETS_FILEINFO_SCHEMA.check_match(fileinfo)
+
+  return fileinfo
+
+
+
+def make_metadata_fileinfo(version, length=None, hashes=None, custom=None):
```
joshuagl: As discussed above, let's remove custom

tuf/formats.py
```diff
@@ -802,7 +808,59 @@ def make_fileinfo(length, hashes, version=None, custom=None):
...
+  fileinfo = {'version' : version}
+
+  if length:
+    fileinfo['length'] = length
+
+  if hashes:
+    fileinfo['hashes'] = hashes
+
+  if custom:
+    fileinfo['custom'] = custom
```
joshuagl: and remove custom here

tuf/repository_lib.py
```diff
@@ -92,11 +92,15 @@
...
 
 
+
+
 def _generate_and_write_metadata(rolename, metadata_filename,
   targets_directory, metadata_directory, storage_backend,
   consistent_snapshot=False, filenames=None, allow_partially_signed=False,
   increment_version_number=True, repository_name='default',
-  use_existing_fileinfo=False):
+  use_existing_fileinfo=False,use_timestamp_length=True,
```
joshuagl: missing a space after the `,`
```suggestion
  use_existing_fileinfo=False, use_timestamp_length=True,
```

tuf/repository_lib.py
```diff
@@ -1392,6 +1396,16 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
       The name of the repository.  If not supplied, 'rolename' is added to the
       'default' repository.
 
+    use_length:
+      Used to decide if the length attribute should be used. From version 1.0.1
```
```suggestion
      Whether to include the optional length attribute of targets metadata files in the snapshot metadata.
```

tuf/repository_lib.py
```diff
@@ -1392,6 +1396,16 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
+      of the tuf spec, length is an optional attribute in snapshot metadata.
+      Default is true.
```
joshuagl: Nit: should capitalise True here

tuf/repository_lib.py
joshuagl: This implies that this is only true from version 1.0.1 of the spec, which I don't think is true? I'm not convinced the version, even if correct, adds value here either.

tuf/repository_lib.py
```diff
@@ -1392,6 +1396,16 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
+
+    use_hashes:
+      Used to decide if the hashes attribute should be used. From version 1.0.1
```
joshuagl: As above re: spec version.

```suggestion
      Whether to include the optional hashes attribute for targets metadata files in the snapshot metadata.```

File: tuf/repository_lib.py [hunk 32]
```diff
@@ -1392,6 +1396,16 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
+      of the tuf spec, hashes is an optional attribute in snapshot metadata.
+      Default is true.
```
joshuagl: Nit: please capitalise to True here

tuf/repository_lib.py
```diff
@@ -1416,12 +1430,30 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
+  hashes = (use_hashes and hashes) or None
+
+  targets_role = targets_filename[:-len(METADATA_EXTENSION)]
+
+  targets_file_version = get_metadata_versioninfo(targets_role,
+    repository_name)
+
+  # Make file info dictionary with make_metadata_fileinfo because as of
+  # version 1.0.1 of the tuf spec length and hashes are optional for all
+  # METAFILES in snapshot.json including the top-level targets file.
```
joshuagl: As with the docstring, I'm not sure it makes sense to reference a version of the spec here

tuf/repository_lib.py
```diff
@@ -1500,6 +1543,16 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
       An object which implements
       securesystemslib.storage.StorageBackendInterface.
 
+    use_length:
+      Used to decide if the length attribute should be used. From version 1.0.1
```
```suggestion
      Whether to include the optional length attribute of the snapshot metadata file in the timestamp metadata.
```

joshuagl: Same as above, this implies it's only true since this spec version. Does including a version in this docstring help the reader?

tuf/repository_lib.py
```diff
@@ -1500,6 +1543,16 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
...
+      of the tuf spec, length is an optional attribute in timestamp metadata.
+      Default is true.
+
+    use_hashes:
+      Used to decide if the hashes attribute should be used. From version 1.0.1
```
```suggestion
      Used to decide if the hashes attribute should be used. From version 1.0.1
```
```suggestion
      Used to decide if the hashes attribute should be used. From version 1.0.1
```
```suggestion
      Whether to include the optional hashes attribute of the snapshot metadata file in the timestamp metadata.
```

joshuagl: Same question about including the version information.

tuf/repository_lib.py
```diff
@@ -1500,6 +1543,16 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
       An object which implements
       securesystemslib.storage.StorageBackendInterface.
 
+    use_length:
+      Used to decide if the length attribute should be used. From version 1.0.1
+      of the tuf spec, length is an optional attribute in timestamp metadata.
+      Default is true.
```
joshuagl: Nit: please capitalise; True

tuf/repository_lib.py
```diff
@@ -1500,6 +1543,16 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
...
+
+    use_hashes:
+      Used to decide if the hashes attribute should be used. From version 1.0.1
+      of the tuf spec, hashes is an optional attribute in timestamp metadata.
+      Default is true.
```
joshuagl: Nit: please capitalise; True

tuf/repository_lib.py
```diff
@@ -1520,14 +1573,22 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
...
   securesystemslib.formats.NAME_SCHEMA.check_match(repository_name)
+  securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_length)
+  securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_hashes)
 
-  # Retrieve the versioninfo of the Snapshot metadata file.
   snapshot_fileinfo = {}
+
   length, hashes = securesystemslib.util.get_file_details(snapshot_filename,
-      tuf.settings.FILE_HASH_ALGORITHMS)
+    tuf.settings.FILE_HASH_ALGORITHMS)
```
joshuagl: secure-systems-lab coding style is to double-indent continuation lines. Please restore indentation here.

tuf/repository_lib.py
```diff
@@ -1445,9 +1477,20 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
       # list these roles found in the metadata directory.
       if tuf.roledb.role_exists(rolename, repository_name) and \
           rolename not in ['root', 'snapshot', 'timestamp', 'targets']:
-        fileinfodict[metadata_name] = get_metadata_versioninfo(rolename,
+
+        length, hashes = securesystemslib.util.get_file_details(
+          os.path.join(metadata_directory, metadata_filename),
+          tuf.settings.FILE_HASH_ALGORITHMS)
```
joshuagl: Coding style is to double-indent continuation lines, please fix here and throughout.

tuf/repository_lib.py
```diff
@@ -1416,12 +1430,30 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
 
   # Snapshot's 'fileinfodict' shall contain the version number of Root,
   # Targets, and all delegated roles fo the repository.
   fileinfodict = {}
-  fileinfodict[TARGETS_FILENAME] = get_metadata_versioninfo(targets_filename,
-      repository_name)
+
+  length, hashes = securesystemslib.util.get_file_details(
+    os.path.join(metadata_directory, targets_filename),
+    tuf.settings.FILE_HASH_ALGORITHMS)
```
joshuagl: Coding style is to double-indent continuation lines, please fix here and throughout.

tuf/repository_lib.py
```diff
@@ -1416,12 +1430,30 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
+
+  length = (use_length and length) or None
+  hashes = (use_hashes and hashes) or None
+
+  targets_role = targets_filename[:-len(METADATA_EXTENSION)]
+
+  targets_file_version = get_metadata_versioninfo(targets_role,
+    repository_name)
```
joshuagl: Coding style is to double-indent continuation lines, please fix here and throughout.

tuf/repository_lib.py
```diff
@@ -1520,14 +1573,22 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
...
+
+  length = (use_length and length) or None
+  hashes = (use_hashes and hashes) or None
+
+  # Retrieve the versioninfo of the Snapshot metadata file.
   snapshot_version = get_metadata_versioninfo('snapshot', repository_name)
   snapshot_fileinfo[SNAPSHOT_FILENAME] = \
-    tuf.formats.make_fileinfo(length, hashes, version=snapshot_version['version'])
+    tuf.formats.make_metadata_fileinfo(snapshot_version['version'],
+      length, hashes)
```
joshuagl: Coding style is to double-indent continuation lines, please fix here and throughout.

tuf/repository_tool.py
```diff
@@ -179,6 +179,26 @@ class Repository(object):
...
+    use_snapshot_length:
+      Used to decide if the length attribute should be used in the snapshot file.
+      From version 1.0.1 of the tuf spec, length is an optional attribute in
+      snapshot metadata file. Default is true.
+
+    use_snapshot_hashes:
+      Used to decide if the hashes attribute should be used in the snapshot file.
+      From version 1.0.1, of the tuf spec, hashes is an optional attribute in
+      snapshot metadata file. Default is true.
+
```
joshuagl: I suggested phrasing changes to the function docstrings above and that we should remove the spec version information. Please ensure you update here also.

joshuagl (changes requested):
Thank you for your continued diligent work on this Martin. There are two important changes we should make (see below) and some minor stylistic changes that would be good to see (comments in-line).

Because `FILEINFO_SCHEMA` was versatile in its use for both target files and metadata files, it included the optional fields used by both the _METAFILES_ and _TARGETS_ objects defined in the specification.

As we are wisely separating duties here, we can remove the fields from our new objects that don't apply. Specifically:
1. `TARGETS_FILEINFO_SCHEMA` doesn't need a _version_ field
2. `METADATA_FILEINFO_SCHEMA` doesn't need a _custom_ field

tuf/repository_lib.py
MVrachev: Yes, you are right. In tuf spec version `1.0.0` they were optional too.
My mistake.

tuf/repository_lib.py
MVrachev: It's not true. Even in tuf spec `1.0.0` for length in hashes in the snapshot metadata file it's written:
```
LENGTH is the integer length in bytes of the metadata file. It is OPTIONAL for all roles.

HASHES is the dictionary that specifies one or more hashes, including the cryptographic hash function. For example: { "sha256": HASH, ... }. It is OPTIONAL for all roles.
```
I thought that length and hashes were made optional as of version 1.0.1 but I was wrong.
Good catch!

trishankatdatadog:
Thanks for all your hard work, guys! Please let me know when it's ready to review, and I'll take another look

tuf/formats.py
MVrachev: Yea, it seems I forgot to do it :D

tuf/formats.py
MVrachev: Nope, a mistake. 
Thanks for noticing.

MVrachev:
I think the pr should be in good shape now and ready for another set of reviews.
I addressed all your suggestions @joshuagl.

trishankatdatadog:
I'm happy if @joshuagl is happy

tuf/repository_lib.py
MVrachev: Ahh... I just forgot to update that one. I will do that in a minute.

MVrachev:
I have only edited one comment on which Joshua commented, but I had forgotten to change it.

I don't have a clue why the Travis CI build for python 2.7 fails when locally it doesn't...

lukpueh:
> I don't have a clue why the Travis CI build for python 2.7 fails when locally it doesn't...

Could be some timeout. I've seen this every now and then in the TUF tests. Just restarted the job.

tests/test_formats.py
```diff
@@ -119,11 +119,14 @@ def test_schemas(self):
...
                                     'private': 'privkey'}}),
 
-      'FILEINFO_SCHEMA': (tuf.formats.FILEINFO_SCHEMA,
+      'TARGETS_FILEINFO_SCHEMA': (tuf.formats.TARGETS_FILEINFO_SCHEMA,
                           {'length': 1024,
                            'hashes': {'sha256': 'A4582BCF323BCEF'},
                            'custom': {'type': 'paintjob'}}),
 
+      'METADATA_FILEINFO_SCHEMA': (tuf.formats.METADATA_FILEINFO_SCHEMA,
+                          {'version': 1}),
```
joshuagl: Nit: the other tests in `test_schemas` seem to include mandatory and option fields. Was there any reason you didn't include `hashes` and `length` here?

tests/test_repository_lib.py
```diff
@@ -236,31 +236,31 @@ def test_get_metadata_fileinfo(self):
...
                    'sha512': sha512_digest_object.hexdigest()}
     fileinfo = {'length': file_length, 'hashes': file_hashes}
-    self.assertTrue(tuf.formats.FILEINFO_SCHEMA.matches(fileinfo))
+    self.assertTrue(tuf.formats.TARGETS_FILEINFO_SCHEMA.matches(fileinfo))
 
     storage_backend = securesystemslib.storage.FilesystemBackend()
 
-    self.assertEqual(fileinfo, repo_lib.get_metadata_fileinfo(test_filepath,
+    self.assertEqual(fileinfo, repo_lib.get_targets_metadata_fileinfo(test_filepath,
                                                               storage_backend))
```
joshuagl: Nit: this indentation looks odd, to maintain the existing style of the surrounding code (which, admittedly, contradicts the coding style) it should align with `test_filepath` above
```suggestion
                                                                      storage_backend))
```

tests/test_repository_lib.py
```diff
@@ -493,7 +501,101 @@ def test_generate_snapshot_metadata(self):
...
+                                          use_length=False)
+    self.assertTrue(tuf.formats.SNAPSHOT_SCHEMA.matches(snapshot_metadata))
+
+    metadata_files_info_dict = snapshot_metadata['meta']
+    for metadata_filename in sorted(os.listdir(metadata_directory), reverse=True):
+
+      # In the metadata_directory, there are files with format:
+      # 1.root.json. The prefix number should be removed.
+      if metadata_filename[0].isdigit():
+        version_number, metadata_filename = metadata_filename.split('.', 1)
```
joshuagl: Nit: repository_lib has `_strip_version_number`, it might be better to use that here?

tests/test_repository_lib.py
```diff
@@ -493,7 +501,101 @@ def test_generate_snapshot_metadata(self):
...
+                                          consistent_snapshot=False,
+                                          use_hashes=False)
+    self.assertTrue(tuf.formats.SNAPSHOT_SCHEMA.matches(snapshot_metadata))
+
+    metadata_files_info_dict = snapshot_metadata['meta']
+    for metadata_filename in sorted(os.listdir(metadata_directory), reverse=True):
+
+      # In the metadata_directory, there are files with format:
+      # 1.root.json. The prefix number should be removed.
+      if metadata_filename[0].isdigit():
```
joshuagl: As above, it may make sense to use `repository_lib._strip_version_number()` here?

tests/test_repository_lib.py
```diff
@@ -493,7 +501,101 @@ def test_generate_snapshot_metadata(self):
...
+        version_number, metadata_filename = metadata_filename.split('.', 1)
+
+      if metadata_filename.endswith('.json'):
```
joshuagl: Do we expect to find files that aren't metadata files with the json suffix? Is this some defensive coding, or did you run across files in the test repository which caused issues here?

tests/test_repository_lib.py
```diff
@@ -493,7 +501,101 @@ def test_generate_snapshot_metadata(self):
...
```
joshuagl: Same as above, do we expect to find files without a json extension here?

tests/test_repository_lib.py
```diff
@@ -493,7 +501,101 @@ def test_generate_snapshot_metadata(self):
...
+                                          use_hashes=False)
+    self.assertTrue(tuf.formats.SNAPSHOT_SCHEMA.matches(snapshot_metadata))
+
+    metadata_files_info_dict = snapshot_metadata['meta']
+    for metadata_filename in sorted(os.listdir(metadata_directory), reverse=True):
+
+      # In the metadata_directory, there are files with format:
+      # 1.root.json. The prefix number should be removed.
+      if metadata_filename[0].isdigit():
+        version_number, metadata_filename = metadata_filename.split('.', 1)
```
joshuagl: As above, might want to use `repository_lib._strip_version_number()` here? Especially as we have the same logic replicated at three call sites in this file alone.

tests/test_repository_lib.py
```diff
@@ -493,7 +501,101 @@ def test_generate_snapshot_metadata(self):
...
+
+      if metadata_filename.endswith('.json'):
```
joshuagl: As above, are we expecting files that without a json extension here?

tests/test_repository_lib.py
```diff
@@ -493,7 +501,101 @@ def test_generate_snapshot_metadata(self):
...
+        if metadata_filename not in \
+            ['root.json', 'targets.json', 'timestamp.json', 'snapshot.json']:
```
joshuagl: Nit: it might be a bit cleaner to define this list as a constant, i.e. `TOP_LEVEL_METADATA` or similar

tests/test_repository_tool.py
```diff
@@ -113,6 +113,13 @@ def test_init(self):
                       'repository_directory', storage_backend, 3, 'targets_directory')
     self.assertRaises(securesystemslib.exceptions.FormatError, repo_tool.Repository,
                       'repository_directory', 'metadata_directory', 3, storage_backend)
+    self.assertRaises(securesystemslib.exceptions.FormatError, repo_tool.Repository,
+                  'repository_directory/', 'metadata_directory/', 'targets_directory/',
+                  storage_backend, repository_name, use_timestamp_length=3)
```
joshuagl: Indentation here is off:
```suggestion
    self.assertRaises(securesystemslib.exceptions.FormatError, repo_tool.Repository,
                      'repository_directory/', 'metadata_directory/', 'targets_directory/',
                      storage_backend, repository_name, use_timestamp_length=3)
```

tests/test_repository_tool.py
```diff
@@ -113,6 +113,13 @@ def test_init(self):
...
+    self.assertRaises(securesystemslib.exceptions.FormatError, repo_tool.Repository,
+                  'repository_directory/', 'metadata_directory/', 'targets_directory/',
+                  storage_backend, repository_name, use_timestamp_length=False,
+                  use_timestamp_hashes=3)
```
joshuagl: Here too
```suggestion
    self.assertRaises(securesystemslib.exceptions.FormatError, repo_tool.Repository,
                      'repository_directory/', 'metadata_directory/', 'targets_directory/',
                      storage_backend, repository_name, use_timestamp_length=False,
                      use_timestamp_hashes=3)
```

tuf/formats.py
```diff
@@ -230,33 +230,40 @@
...
   key_schema = RELPATH_SCHEMA,
-  value_schema = FILEINFO_SCHEMA)
+  value_schema = TARGETS_FILEINFO_SCHEMA)
 
 # A dict holding a target info.
 TARGETINFO_SCHEMA = SCHEMA.Object(
   object_name = 'TARGETINFO_SCHEMA',
   filepath = RELPATH_SCHEMA,
-  fileinfo = FILEINFO_SCHEMA)
+  fileinfo = TARGETS_FILEINFO_SCHEMA)
```
joshuagl, line 266: Is it just me, or are `FILEDICT_SCHEMA` and `TARGETINFO_SCHEMA` the same? Would be good to remove one of those if so, perhaps as an additional patch in this PR?

`TARGETINFO_SCHEMA` seems to me to be the better name.

tuf/repository_lib.py
```diff
@@ -1350,10 +1355,9 @@ def _generate_targets_fileinfo(target_files, targets_directory,
 
 
 
-
 def generate_snapshot_metadata(metadata_directory, version, expiration_date,
     targets_filename, storage_backend, consistent_snapshot=False,
-    repository_name='default'):
+    repository_name='default', use_length=True, use_hashes=True):
```
joshuagl: Note to maintainers, this is a behaviour change that we should document for the next release. In the current release 
> snapshot metadata always excludes length and hashes of (delegated) targets metadata
from #996

cc: @lukpueh

tuf/repository_lib.py
```diff
@@ -1416,13 +1430,31 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
-  fileinfodict[TARGETS_FILENAME] = get_metadata_versioninfo(targets_filename,
+
+  length, hashes = securesystemslib.util.get_file_details(
+      os.path.join(metadata_directory, targets_filename),
+      tuf.settings.FILE_HASH_ALGORITHMS)
+
+  length = (use_length and length) or None
+  hashes = (use_hashes and hashes) or None
+
+  targets_role = targets_filename[:-len(METADATA_EXTENSION)]
```
joshuagl: We have an odd mix here of using the passed `targets_filename` and using the `TARGETS_FILENAME` constant.

AFAICS there's no need for the targets_filename argument to `generate_snapshot_metadata`, because the only caller assigns to it the value of `TARGETS_FILENAME[:-len(METADATA_EXTENSION)]`. It would be cleaner to drop this argument and instead use `TARGETS_FILENAME` consistently throughout the method.

tuf/repository_lib.py
```diff
@@ -1520,14 +1573,22 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
   tuf.formats.METADATAVERSION_SCHEMA.check_match(version)
   securesystemslib.formats.ISO8601_DATETIME_SCHEMA.check_match(expiration_date)
   securesystemslib.formats.NAME_SCHEMA.check_match(repository_name)
+  securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_length)
+  securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_hashes)
 
-  # Retrieve the versioninfo of the Snapshot metadata file.
   snapshot_fileinfo = {}
+
   length, hashes = securesystemslib.util.get_file_details(snapshot_filename,
```
joshuagl: I realise this is not a change you introduced, but we have similar incoherent code here where we use a passed `snapshot_filename` and the `SNAPSHOT_FILENAME` constant. Although in this instance the passed value isn't just assigned the constant before calling `generate_timestamp_metadata()`. 

Would you be able to  clean this up in a follow-on patch?

joshuagl (commented):
Thanks for your continued diligence on this PR @MVrachev. I have a few minor nits (indentation related and some clarifications) and suggest a couple of clean-ups that would be great additions to this PR.

tests/test_formats.py
MVrachev: No particular, I will add the other fields as well.

tuf/formats.py
MVrachev: I made a good investigation about this and what I saw is that `FILEDICT_SCHEMA` and `TARGETINFO_SCHEMA` are not actually 100 % the same.
FILEDICT_SCHEMA  is a dict in the format `{ "path" : "fileinfo" }` and
TARGETINFO_SCHEMA  is a dict in the format `{ "filepath" : <something> , "fileinfo" : <something>}`.

This could be seen pretty well into the `tests/test_formats.py`:
image

Even though `FILEDICT_SCHEMA` cannot be directly replaced with `TARGETINFO_SCHEMA` I feel that we can drop it out and we don't need it.
I will work to add this in a separate pr.

MVrachev: WIll track the progress here: https://github.com/theupdateframework/tuf/issues/1043

tests/test_repository_lib.py
MVrachev: Yes, there is the `role_file.xml` file created in the `_setup_generate_snapshot_metadata_test` function. 
It's created to make sure it's not loaded by `generate_snapshot_metadata`.
If it's loaded, then this exception will be thrown:
https://github.com/theupdateframework/tuf/blob/95d08cc5b4d786d1ec7a3c5f1f05e54c8c163736/tuf/repository_lib.py#L1452

I will add a comment clarifying that.

tests/test_repository_lib.py
MVrachev: Look at my reply above.

tests/test_repository_lib.py
MVrachev: Look at my reply above.

tuf/repository_lib.py
MVrachev: Yes, I will remove that argument, but it's important to note that this removes the capability to pass a custom `target_filename` arg which could be okay, but it's important to know.

Currently, I don't think it was ever used with a value different than `TARGETS_FILENAME` so it should be okay. I will check that.

tuf/repository_lib.py
MVrachev: It's not the same situation as with `targets_filename` because the `snapshot_filename` value comes from here: https://github.com/theupdateframework/tuf/blob/5d40ffa3c49f6ce8ba5908abacec5630833829e8/tuf/repository_lib.py#L136
and the `fillenames` in one place have a different value than just `snapshot.json`: 
https://github.com/theupdateframework/tuf/blob/5d40ffa3c49f6ce8ba5908abacec5630833829e8/tuf/repository_tool.py#L405.

Still, I noticed an inconsistency that we use both `snapshot_filename` and the constant `SNAPSHOT_FILENAME` at the same function. Will fix that and just use `snapshot_filename`.

MVrachev:
I addressed Joshua's comments and add a two commits `Remove redundant targets_filename argument ` and `Fix snapshot_filename inconsistency usage`.

The pr is ready for yet another set reviews.

tests/test_repository_lib.py
```diff
@@ -463,37 +465,136 @@ def test_generate_snapshot_metadata(self):
...
+    self.assertTrue(tuf.formats.SNAPSHOT_SCHEMA.matches(snapshot_metadata))
+
+    metadata_files_info_dict = snapshot_metadata['meta']
+    for metadata_filename in sorted(os.listdir(metadata_directory), reverse=True):
+
+      # In the metadata_directory, there are files with format:
+      # 1.root.json. The prefix number should be removed.
+      striped_filename, version = \
+        repo_lib._strip_version_number(metadata_filename,
+                                      consistent_snapshot=True)
```
joshuagl: Nit: indentation is off here.

tests/test_repository_lib.py
```diff
@@ -463,37 +465,136 @@ def test_generate_snapshot_metadata(self):
...
+                                          consistent_snapshot=False,
+                                          use_length=False)
+    self.assertTrue(tuf.formats.SNAPSHOT_SCHEMA.matches(snapshot_metadata))
+
+    metadata_files_info_dict = snapshot_metadata['meta']
+    for metadata_filename in sorted(os.listdir(metadata_directory), reverse=True):
+
+      # In the metadata_directory, there are files with format:
+      # 1.root.json. The prefix number should be removed.
+      striped_filename, version = \
```
joshuagl: Nit: you mean stripped, not striped.

tests/test_repository_lib.py
```diff
@@ -463,37 +465,136 @@ def test_generate_snapshot_metadata(self):
...
+        repo_lib._strip_version_number(metadata_filename,
+                                      consistent_snapshot=True)
```
joshuagl: Nit: indentation is off here.

tests/test_repository_lib.py
```diff
@@ -463,37 +465,136 @@ def test_generate_snapshot_metadata(self):
...
```
joshuagl: Nit: indentation is off here

tests/test_repository_lib.py
```diff
@@ -463,37 +465,136 @@ def test_generate_snapshot_metadata(self):
...
+
+      # In the repository, the file "role_file.xml" have been added to make
+      # sure that non-json files aren't loaded. This file should be filtered.
+      if striped_filename.endswith('.json'):
+        if striped_filename not in TOP_LEVEL_METADATA_FILES:
+
```
joshuagl: Nit: you're inconsistent in your usage of a blank line after an if statement. Personally, I prefer no blank line. But here with no blank line following the first `if` and a blank line following the second is jarring.

tests/test_repository_lib.py
```diff
@@ -505,7 +606,7 @@ def test_generate_timestamp_metadata(self):
                                       repo_lib.METADATA_STAGED_DIRECTORY_NAME)
     targets_directory = os.path.join(repository_directory, repo_lib.TARGETS_DIRECTORY_NAME)
 
-    snapshot_filename = os.path.join(metadata_directory,
+    snapshot_file_path = os.path.join(metadata_directory,
                                      repo_lib.SNAPSHOT_FILENAME)
```
joshuagl: Nit: indentation is off here

tuf/formats.py
```diff
@@ -789,20 +793,64 @@ def make_fileinfo(length, hashes, version=None, custom=None):
...
+
+  return fileinfo
+
+
+
+def make_metadata_fileinfo(version, length=None, hashes=None):
+  """
+  <Purpose>
+    Create a dictionary conformant to 'METADATA_FILEINFO_SCHEMA'.
+    This dict describes both metadata and target files.
```
joshuagl: I don't think this is true? `METADATA_FILEINFO_SCHEMA` describes only metadata files

tuf/formats.py
```diff
@@ -759,10 +766,10 @@ def parse_base64(base64_string):
 
 
 
-def make_fileinfo(length, hashes, version=None, custom=None):
+def make_targets_fileinfo(length, hashes, custom=None):
   """
   <Purpose>
-    Create a dictionary conformant to 'FILEINFO_SCHEMA'.
+    Create a dictionary conformant to 'TARGETS_FILEINFO_SCHEMA'.
     This dict describes both metadata and target files.
```
joshuagl: This isn't true, this dict describes only target files.

tuf/formats.py
```diff
@@ -789,20 +793,64 @@ def make_fileinfo(length, hashes, version=None, custom=None):
...
+
+
+  <Exceptions>
+    securesystemslib.exceptions.FormatError, if the 'METADATA_FILEINFO_SCHEMA' to be
+    returned does not have the correct format.
+
+  <Side Effects>
+    If any of the arguments are incorrectly formatted, the dict
+    returned will be checked for formatting errors, and if found,
+    will raise a 'securesystemslib.exceptions.FormatError' exception.
```
joshuagl: I appreciate the diligence here, but I think we don't need this Side Effects section in this case. I'd expect this part of the docstring to describe side-effects that weren't obvious from the rest of the docstring. We already explain the Exception in the section above.

tuf/repository_lib.py
```diff
@@ -1530,18 +1583,27 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
...
+  hashes = (use_hashes and hashes) or None
+
+  _, snapshot_filename = os.path.split(snapshot_file_path)
+  # Retrieve the versioninfo of the Snapshot metadata file.
   snapshot_version = get_metadata_versioninfo('snapshot', repository_name)
-  snapshot_fileinfo[SNAPSHOT_FILENAME] = \
-    tuf.formats.make_fileinfo(length, hashes, version=snapshot_version['version'])
+  snapshot_fileinfo[snapshot_filename] = \
+      tuf.formats.make_metadata_fileinfo(snapshot_version['version'],
+          length, hashes)
```
joshuagl: I don't believe this necessary? The function worked before by using the constant `SNAPSHOT_FILENAME` why would do we need to change away from that? I don't think there's any need for the `snapshot_file_path` argument. AFAICS the one place that wasn't passing in 'snapshot.json' was passing 'path/to/snapshot.json' which your change here will strip down to 'snapshot.json'. Am I misunderstanding?

tuf/repository_lib.py
```diff
@@ -1530,18 +1583,27 @@ def generate_timestamp_metadata(snapshot_filename, version, expiration_date,
...
   snapshot_fileinfo = {}
-  length, hashes = securesystemslib.util.get_file_details(snapshot_filename,
+
+  length, hashes = securesystemslib.util.get_file_details(snapshot_file_path,
       tuf.settings.FILE_HASH_ALGORITHMS, storage_backend)
+
+  length = (use_length and length) or None
+  hashes = (use_hashes and hashes) or None
+
+  _, snapshot_filename = os.path.split(snapshot_file_path)
```
joshuagl: Nit: [`os.path.basename()`](https://docs.python.org/3/library/os.path.html#os.path.basename) would be a bit cleaner here

joshuagl (commented):
Thanks for sticking at this Martin. This is looking really solid. I have only one non-trivial concern, which is the change to `generate_timestamp_metadata` – I really don't think the `snapshot_file_path` argument is necessary. All we really want is the filename, which you get in your changes here by calling `os.path.split()`, and that is defined by the constant `SNAPSHOT_FILENAME` – let's just keep using the constant as the code does today and drop the argument to `generate_timestamp_metadata()`.

tuf/formats.py
MVrachev: Agree. I will remove the side effect section from the `make_targets_fileinfo` as well.

tests/test_repository_lib.py
MVrachev: Yes in this situation it's awkward. I will remove the blank line.
The reason I added it so that the assertions could be easier to read, but I understand your confusion.

tuf/repository_lib.py
MVrachev: `snapshot_file_path` is needed to calculate the hashes and length of the snapshot file by using `get_file_details` from securesyslib: https://github.com/secure-systems-lab/securesystemslib/blob/9b3a78e998412c39418efb29f411591b20cdd236/securesystemslib/util.py#L45
which has the argument `filepath` - the absolute path to the file and in our case to the snapshot file.

Here we can't remove the snapshot_file_path as we did with the target_filename in the `generate_snapshot_metadata` because in `generate_snapshot_metadata` we have the extra `metadata_directory` argument from which we can use to create the absolute path needed for `get_file_details`.

MVrachev:
I had to rebase the second time to remove a commit not made by me which introduced into the changes.

tests/test_formats.py
```diff
@@ -119,11 +119,16 @@ def test_schemas(self):
...
+      'TARGETS_FILEINFO_SCHEMA': (tuf.formats.TARGETS_FILEINFO_SCHEMA,
                           {'length': 1024,
                            'hashes': {'sha256': 'A4582BCF323BCEF'},
                            'custom': {'type': 'paintjob'}}),
 
+      'METADATA_FILEINFO_SCHEMA': (tuf.formats.METADATA_FILEINFO_SCHEMA,
+                          {'length': 1024,
+                           'hashes': {'sha256': 'A4582BCF323BCEF'},
+                           'version': 1}),
+
```
joshuagl: Nit: indentation is not right here

tuf/formats.py
```diff
@@ -759,11 +766,11 @@ def parse_base64(base64_string):
...
 
 
-def make_fileinfo(length, hashes, version=None, custom=None):
+def make_targets_fileinfo(length, hashes, custom=None):
   """
   <Purpose>
-    Create a dictionary conformant to 'FILEINFO_SCHEMA'.
-    This dict describes both metadata and target files.
+    Create a dictionary conformant to 'TARGETS_FILEINFO_SCHEMA'.
+    This dict describes the target files.
```
joshuagl: Nit:
```suggestion
    This dict describes a target file.
```

tuf/repository_lib.py
```diff
@@ -1458,9 +1489,20 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
       # list these roles found in the metadata directory.
       if tuf.roledb.role_exists(rolename, repository_name) and \
           rolename not in ['root', 'snapshot', 'timestamp', 'targets']:
-        fileinfodict[metadata_name] = get_metadata_versioninfo(rolename,
+
+        length, hashes = securesystemslib.util.get_file_details(
```
joshuagl: We're computing length and hashes here, even if we will never use them. Further, because we could be using length, or hashes,  or both, it's suboptimal to compute both in the same function.

joshuagl (approved):
Thanks for your perseverance with these changes @MVrachev.

I realised that in `generate_snapshot_metadata()` we are computing lengths and hashes for all targets metadata files, regardless of whether we are using lengths and hashes or not. This doesn't seem like a major issue until you consider that the PEP 458 implementation will have over 16k delegations (and therefore over 16k targets metadata files to compute the hashes for). 

I anticipate that Warehouse will want to use length and hashes for targets metadata, but we should make sure that if they opt out we aren't doing a lot of unnecessary work. Happy to see that as a patch here or we can file it as an Issue to be implemented in a separate PR before the next release. What do you think?

tests/test_formats.py
MVrachev: I am not sure what you mean by "not right".

joshuagl: The dictionary was originally indented to align with the opening delimiter on the line above (PEP 8 style). With the new longer variable name the indentation no longer matches the existing implicit coding style, nor does it match the explicit coding style of the project.

MVrachev:
> Thanks for your perseverance with these changes @MVrachev.
> 
> I realised that in `generate_snapshot_metadata()` we are computing lengths and hashes for all targets metadata files, regardless of whether we are using lengths and hashes or not. This doesn't seem like a major issue until you consider that the PEP 458 implementation will have over 16k delegations (and therefore over 16k targets metadata files to compute the hashes for).
> 
> I anticipate that Warehouse will want to use length and hashes for targets metadata, but we should make sure that if they opt out we aren't doing a lot of unnecessary work. Happy to see that as a patch here or we can file it as an Issue to be implemented in a separate PR before the next release. What do you think?

A good consideration.
I added one patch fixing that.
Also, fixed one of the nits. The indentation comment I didn't understand.

tests/test_formats.py
MVrachev: Yes, now I get it better.
I was confused that you left the comment in a random location where I placed a new line.

MVrachev:
I fixed the last nit you pointed out @joshuagl.

tuf/repository_lib.py
```diff
@@ -1458,9 +1489,26 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
...
       if tuf.roledb.role_exists(rolename, repository_name) and \
           rolename not in ['root', 'snapshot', 'timestamp', 'targets']:
-        fileinfodict[metadata_name] = get_metadata_versioninfo(rolename,
+
+        length = None
+        hashes = None
+        # We want to make sure we are calculating length and hashes only when
+        # at least one of them is needed. Otherwise, for adoptors of tuf with
+        # lots of delegations, this will cause unnecessary overhead.
+        if use_length or use_hashes:
```
joshuagl, line 1653: Thanks for adding this! Ideally we would only calculate length when `use_length` and hashes when `use_hashes`, but that will require some additions to securesystemslib and as such I think we should open an Issue to track that enhancement once this PR has landed, linked to the corresponding required API additions in securesystemslib https://github.com/secure-systems-lab/securesystemslib/issues/246

joshuagl (approved):
Thank you for this PR @MVrachev. I think this is ready to land. 🎉

I've added a comment in-line about a follow-on enhancement. I think that should be a follow-on to this PR as it requires a change in, and release of, securesystemslib.

@trishankatdatadog please take a look.

tuf/repository_lib.py
```diff
@@ -92,11 +92,15 @@
...
+
+
 def _generate_and_write_metadata(rolename, metadata_filename,
   targets_directory, metadata_directory, storage_backend,
   consistent_snapshot=False, filenames=None, allow_partially_signed=False,
   increment_version_number=True, repository_name='default',
-  use_existing_fileinfo=False):
+  use_existing_fileinfo=False, use_timestamp_length=True,
+  use_timestamp_hashes=True, use_snapshot_length=True,
+  use_snapshot_hashes=True):
```
jku: This is an internal function so it's not a show-stopper for this PR (and I don't have actionable suggestions either)... but 15 (!) arguments is a hint of something being wrong here.

tuf/repository_lib.py
```diff
@@ -1363,10 +1372,9 @@ def _generate_targets_fileinfo(target_files, targets_directory,
 
 
 
-
 def generate_snapshot_metadata(metadata_directory, version, expiration_date,
-    targets_filename, storage_backend, consistent_snapshot=False,
-    repository_name='default'):
+    storage_backend, consistent_snapshot=False, repository_name='default',
+    use_length=True, use_hashes=True):
```
jku: I'm still confused about what is public API and what is not... but is changing the signature like this appropriate here? I guess this is an "internal library" so it's fine?

There's other 'API' changes in the PR of course but this is the one that changes a signature in a potentially subtle way

tuf/repository_lib.py
lukpueh: Well observed, @jku! I agree 100% that this function needs a major re-write... and also the way it is called. E.g. repository_tool's `writeall` function already performs the whole role case handling that's also done in this function, and so does `repository_lib._log_status_of_top_level_roles`, which is a whole different story (see https://github.com/theupdateframework/tuf/issues/955).

MVrachev: I agree that this function is overwhelmed with arguments, but I don't think that this pr is the place to fix that.
We fixed too many bugs in this pr which are not related to`
Make length and hashes optional for timestamp and snapshot roles`.

Also, I don't see an appropriate issue that documents the need to rewrite `__generate_and_write_metadata` and how we call it.
I created an issue about it: https://github.com/theupdateframework/tuf/issues/1070

tuf/repository_lib.py
MVrachev: Well, I don't see another way to achieve the optional length and hashes without passing those additional arguments as 
`_generate_and_write_metadata` is not a member function of any class.

MVrachev:
I rebased upon the latest changes and there are no other merge conflicts left.

Can we merge that pr? Do we expect something more from it?
@trishankatdatadog @lukpueh @jku ?

tuf/repository_lib.py
jku: So my concern was not with the added arguments but the removed one (which in some cases could create a situation where a caller keeps using old signature and it happens to get accepted by accident). But since I think this function is not API (_right?_) so that's fine: you've fixed all of the call sites in tuf so it's all good.

jku (approved):
LGTM (take that with pinch of salt, I'm not familiar with all of the code touched here)

tuf/repository_lib.py
MVrachev: I don't think this function is an API.
The only other place where I see that this function is called or referenced is in the `_generate_and_write_metadata` function which is an internal function.

lukpueh: AFAIK python-TUF does not have a clearly defined API. The closest thing to an API definition is the TUF tutorial, which suggests that a user should interact with the TUF repository mainly through repository_tool.

trishankatdatadog (commented):
Sorry for the late review. A few comments:

1. 💯 on separate metadata and targets file info schemas.
2. Could we default to `use_timestamp_length=True, use_timestamp_hashes=True, use_snapshot_length=False, use_snapshot_hashes=False`, to default to the common py-TUF setup thus far?
3. Can and how do users who use `repository_tool.create_new_repository()` and `repository_tool.load_repository()` control (2)?

Thanks for all your great work, Martin and Joshua!

MVrachev:
> ``Could we default to use_timestamp_length=True, use_timestamp_hashes=True, use_snapshot_length=False, use_snapshot_hashes=False, to default to the common py-TUF setup thus far?

Here, I followed a simple principle that if somebody wants to disable optional security features, then they should mention it clearly and that the defaults should be the secure one (even though I understand that trade-offs between security and performance are often made).

I also understand the motives why you want to keep the older behavior in place and I will change them now.

trishankatdatadog:
> Here, I followed a simple principle that if somebody wants to disable optional security features, then they should mention it clearly and that the defaults should be the secure one (even though I understand that trade-offs between security and performance are often made).
> 
> I also understand the motives why you want to keep the older behavior in place and I will change them now.

Yes, maybe point to [Section 5.6 of the Mercury paper](https://www.usenix.org/conference/atc17/technical-sessions/presentation/kuppusamy) to explain why we are keeping the old behaviour :)

MVrachev:
I updated my pr with the following changes:

1) Updated commits `Make length and hashes optional for snapshot ` and `Add tests for snapshot without hashes or length ` to use `use_snapshot_length=False` and `use_snapshot_hashes=False` as defaults.

2) Add two new commits to allow users to control if they want hashes and length to be computed for snapshot and timestamp roles when using `repository_tool.create_new_repository()` and `repository_tool.load_repository()` functions.

trishankatdatadog (approved):

MVrachev:
Added one final commit to clarify in some of the docstrings which section of the Mercury paper to look for.

I don't think there are other changes that have to be made.
It should be ready to be merged.

tuf/repository_tool.py
```diff
@@ -2919,14 +2919,14 @@ def create_new_repository(repository_directory, repository_name='default',
       Whether to include the optional length attribute for targets
       metadata files in the snapshot metadata.
       Default is False because of bandwidth considerations.
```
trishankatdatadog:
```suggestion
      Default is False to save B/W but w/o losing security from rollback attacks.
```

tuf/repository_tool.py
```diff
@@ -2919,14 +2919,14 @@ def create_new_repository(repository_directory, repository_name='default',
...
-      Read more here:
+      Read more at section 5.6 from the Mercury paper:
       https://www.usenix.org/conference/atc17/technical-sessions/presentation/kuppusamy
 
     use_snapshot_hashes:
       Whether to include the optional hashes attribute for targets
       metadata files in the snapshot metadata.
       Default is False because of bandwidth considerations.
```
trishankatdatadog:
```suggestion
      Default is False to save B/W but w/o losing security from rollback attacks.
```

tuf/repository_tool.py
```diff
@@ -3026,14 +3026,14 @@ def load_repository(repository_directory, repository_name='default',
       Whether to include the optional length attribute for targets
       metadata files in the snapshot metadata.
       Default is False because of bandwidth considerations.
```
trishankatdatadog:
```suggestion
      Default is False to save B/W but w/o losing security from rollback attacks.
```

tuf/repository_tool.py
```diff
@@ -3026,14 +3026,14 @@ def load_repository(repository_directory, repository_name='default',
...
-      Read more here:
+      Read more at section 5.6 from the Mercury paper:
       https://www.usenix.org/conference/atc17/technical-sessions/presentation/kuppusamy
 
     use_snapshot_hashes:
       Whether to include the optional hashes attribute for targets
       metadata files in the snapshot metadata.
       Default is False because of bandwidth considerations.
```
trishankatdatadog:
```suggestion
      Default is False to save B/W but w/o losing security from rollback attacks.
```

MVrachev:
I will add your suggestions, but if you don't mind without the abbreviations of `B/W` and `w/o` because they are little ambiguous to me as a non-native English speaker.

MVrachev:
I updated the last commit with your suggestion on all places where we used flags to control the computation snapshot length and hashes.

trishankatdatadog:
One minor comment: please don't force-push unless absolutely necessary, makes it hard to see new changes...

joshuagl:
Thanks for all of your work on this @MVrachev and the thorough review @trishankatdatadog !

PR #1033, opened 2020-05-20
Clarify generate_targets_metadata()

joshuagl:
**Fixes issue #**: N/A

* Improve documentation for `generate_targets_metadata()`, in the docstring and code comments, to make the `use_existing_fileinfo` and `write_consistent_targets` arguments, and their interactions, clearer.
* Fix `generate_targets_metadata()` so that both `use_existing_fileinfo` and `write_consistent_fileinfo` can be used together, as required by Warehouse for PEP 458

requirements**:

### DIFF
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -1251,9 +1251,12 @@ def generate_targets_metadata(targets_directory, target_files, version,
   securesystemslib.formats.BOOLEAN_SCHEMA.check_match(write_consistent_targets)
   securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_existing_fileinfo)
 
+  # If the caller requested consistent targets *and* supplied their own fileinfo
+  # consistent versions of the targets can not be created, because target
+  # files and their storage are not accessed when using existing fileinfo.
   if write_consistent_targets and use_existing_fileinfo:
-    raise securesystemslib.exceptions.Error('Cannot support writing consistent'
-        ' targets and using existing fileinfo.')
+    logger.info('Consistent targets will not be created because targets '
+        'metadata is being generated from existing fileinfo.')
 
   if delegations is not None:
     tuf.formats.DELEGATIONS_SCHEMA.check_match(delegations)
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -1173,8 +1173,14 @@ def generate_targets_metadata(targets_directory, target_files, version,
 
     target_files:
       The target files tracked by 'targets.json'.  'target_files' is a
-      dictionary of target paths that are relative to the targets directory and
-      a fileinfo dict matching tuf.formats.LOOSE_FILEINFO_SCHEMA
+      dictionary mapping target paths (relative to the targets directory) to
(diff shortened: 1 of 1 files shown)


tuf/repository_lib.py
```diff
@@ -1253,20 +1264,25 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
     for target, fileinfo in six.iteritems(target_files):
 
       # Ensure all fileinfo entries in target_files have a non-empty hashes dict
       if not fileinfo.get('hashes', None):
         raise securesystemslib.exceptions.Error('use_existing_hashes option set'
             ' but no hashes exist in roledb for ' + repr(target))
 
-      if fileinfo.get('length', -1) < 0:
+      # and a non-empty, non-zero, length
+      if fileinfo.get('length', -1) < 1:
```
joshuagl, line 1280: I think it doesn't make sense to have zero length targets files, but maybe this should be restored to the prior check of: `if len < 0` ?

tuf/repository_lib.py
```diff
@@ -1240,9 +1251,12 @@ def generate_targets_metadata(targets_directory, target_files, version,
   securesystemslib.formats.BOOLEAN_SCHEMA.check_match(write_consistent_targets)
   securesystemslib.formats.BOOLEAN_SCHEMA.check_match(use_existing_fileinfo)
 
+  # If the caller requested consistent targets *and* supplied their own fileinfo
```
mnm678, line 1254:
```suggestion
  # If the caller requested consistent targets *and* supplied their own fileinfo,
```

mnm678 (commented):
It looks like setting both `write_consistent_targets` and `use_existing_fileinfo` just ignores the `write_consistent_targets`, which I'm not sure would be the expected behavior. If the caller is responsible for writing the consistent files, it might make more sense for them to set `write_consistent_targets` to false here, then write the consistent files afterwards. But I might be misunderstanding the goal.

joshuagl:
@lukpueh already resolved this issue at the right level by modifying `_generate_and_write_metadata()` to call `generate_targets_metadats()` with an appropriate value for `write_consistent_targets` based on the values of `consistent_snapshot` and `use_existing_fileinfo` in b9891e171bde7626e3218b4e2c3674f447a0b980

Apologies for the noise. I will close this PR and include the first patch in a later PR.

tuf/repository_lib.py
lukpueh: > I think it doesn't make sense to have zero length targets files, but maybe this should be restored to the prior check of: `if len < 0` ?

@joshuagl, the classical `__init__.py` file is a good example for a target file with zero length.

joshuagl: You're right, thanks for pointing out that case @lukpueh. In which case I should restore this to the prior `if len < 0`. Will force push a fix later today.

joshuagl: Oh, I already restored the behaviour for checking `len < 0` in #1034, which contains an updated version of this change. Apologies for the confusion.

lukpueh: No need to apologize. I have no business commenting on closed PRs. :D

Issue #1037, opened 2020-05-26
Targets delegations are not updated when adding a new delegation key


sechkova:
**Description of issue**:

When adding a new key to a delegated role e.g.:

`repository.targets('unclaimed').add_verification_key(public_new_key)`

the delegating role metadata (e.g. `targets.json`) is not updated with the new key.

**Steps to reproduce:** 

1) Delegate to 'unclaimed' role
```
repository.targets.delegate('unclaimed', [public_unclaimed_key], ['*']) 
repository.mark_dirty(['unclaimed', 'targets', 'snapshot', 'timestamp'])
repository.writeall()
```
Delegated role's internal metadata:                                                                                                                                                           

```
tuf.roledb.get_roleinfo('unclaimed')    
{'name': 'unclaimed',
 'keyids': ['9313d41de1204c061f09429197edbe12f71bd6dd354ebae6bd3bdbee4bffb1e1'],
 'signing_keyids': [],
 'threshold': 1,
 'version': 1,
 'expires': '2020-08-25T21:30:11Z',
 'signatures': [],
 'partial_loaded': False,
 'paths': {},
 'delegations': {'keys': {}, 'roles': []}}
```

targets.json:

```
"signed": {
  "_type": "targets",
  "delegations": {
   "keys": {
    "9313d41de1204c061f09429197edbe12f71bd6dd354ebae6bd3bdbee4bffb1e1": {
     "keyid_hash_algorithms": [
      "sha256",
      "sha512"
     ],
     "keytype": "ed25519",
     "keyval": {
      "public": "ede15f800ab11184080552a4edbfa61b0a0920243acebd536cca7d344331d21b"
     },
     "scheme": "ed25519"
    }
   },
   "roles": [
    {
     "keyids": [
      "9313d41de1204c061f09429197edbe12f71bd6dd354ebae6bd3bdbee4bffb1e1"
     ],
     "name": "unclaimed",
     "paths": [
      "*"
     ],
     "terminating": false,
     "threshold": 1
    }
   ]
  },
  "expires": "2020-08-25T21:27:11Z",
  "spec_version": "1.0.0",
  "targets": {},
  "version": 2
 }
}
```

2) Rotate the 'unclaimed' role key

```
repository.targets('unclaimed').remove_verification_key(public_unclaimed_key)
repository.targets('unclaimed').add_verification_key(public_new_key) 
repository.mark_dirty(['unclaimed', 'targets', 'snapshot', 'timestamp'])
[...]

PR #1038, opened 2020-05-27
WIP: Remove the need of manually marking roles as dirty

sechkova:
**Fixes issue #**: #964 #958

**Note:** Although the best solution is undoubtedly to go for a general rework of the internal metadata representation, this pull request aims to be a short-term fix keeping the changes as close as possible to the current functionality.

I am opening it as a WIP to discuss the possible ways to handle some more complicated delegations scenarios.

Remove the need of manually marking dirty all roles whose metadata needs to be (re)written on the filesystem:
- fix functions like `Metadata.load_signing_key()` that mark a role as dirty, although it shouldn't update the corresponding metadata on filesystem.
- add `mark_role_as_dirty` parameter to `roledb.add_role()` and `Targets` class which is needed to mark newly created delegated roles as dirty
- add `roledb. _propagate_dirty_roles()`  which recursively marks as dirty all roles affected by a roledb update according to the role dependencies graph e.g.:
   
```
   any update on targets:  'targets' -> 'snapshot' -> 'timestamp'
   update on targets' key: 'targets' -> 'root'
```

_Note that the dependencies is the implementation are reversed in comparison with the specification: a targets key update is performed on the Targets object which then needs to be propagated to the root metadata._

- add 'delegating_role' to role metadata dictionary. This is needed to be able to propagate an update of the delegated role to the delegating. 

` 'unclaimed' -> 'targets' -> 'snapshot' -> 'timestamp'` 

**Open for discussion:** What would be the purpose of a chain of delegations where two roles both delegate to a third?

```
   targets -> A -> C
   targets -> B -> C
```

Are both A and B delegating the same key and paths/targets to C? Is a scenario where each A and B delegates a different key to a role with the same name (C) meaningful? What should be the chain of interdependencies if C's key is updated?

While working on this I also noticed that delegating role is not updated at all when the delegated role's key is changed (check #1037).

In addition: 
- do some code cleanup and remove repetitive code from `Repository.status()` (#955 )

requirements**:

### DIFF
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -175,7 +175,7 @@ def _generate_and_write_metadata(rolename, metadata_filename,
   else:
     logger.debug('Not incrementing ' + repr(rolename) + '\'s version number.')
 
-  if rolename in ['root', 'targets', 'snapshot', 'timestamp'] and not allow_partially_signed:
+  if rolename in tuf.roledb.TOP_LEVEL_ROLES and not allow_partially_signed:
     # Verify that the top-level 'rolename' is fully signed.  Only a delegated
     # role should not be written to disk without full verification of its
     # signature(s), since it can only be considered fully signed depending on
@@ -394,18 +394,15 @@ def _delete_obsolete_metadata(metadata_directory, snapshot_metadata,
     else:
       logger.debug(repr(metadata_role) + ' found in the snapshot role.')
 
-
-
     # Strip metadata extension from filename.  The role database does not
     # include the metadata extension.
     if metadata_role.endswith(METADATA_EXTENSION):
       metadata_role = metadata_role[:-len(METADATA_EXTENSION)]
-
     else:
       logger.debug(repr(metadata_role) + ' does not match'
           ' supported extension ' + repr(METADATA_EXTENSION))
 
-    if metadata_role in ['root', 'targets', 'snapshot', 'timestamp']:
+    if metadata_role in tuf.roledb.TOP_LEVEL_ROLES:
       logger.debug('Not removing top-level metadata ' + repr(metadata_role))
       return
 
@@ -847,7 +844,7 @@ def get_delegations_filenames(metadata_directory, consistent_snapshot,
 
     # Skip top-level roles, only interested in delegated roles now that the
     # top-level roles have already been loaded.
-    if metadata_name in ['root', 'snapshot', 'targets', 'timestamp']:
+    if metadata_name in tuf.roledb.TOP_LEVEL_ROLES:
       continue
 
     filenames[metadata_name] = metadata_path
@@ -1126,7 +1123,7 @@ def generate_root_metadata(version, expiration_date, consistent_snapshot,
   # Extract the role, threshold, and keyid information of the top-level roles,
   # which Root stores in its metadata.  The necessary role metadata is generated
   # from this information.
-  for rolename in ['root', 'targets', 'snapshot', 'timestamp']:
+  for rolename in tuf.roledb.TOP_LEVEL_ROLES:
 
     # If a top-level role is missing from 'tuf.roledb.py', raise an exception.
     if not tuf.roledb.role_exists(rolename, repository_name):
@@ -1488,7 +1485,7 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
       # snapshot and timestamp roles are not listed in snapshot.json, do not
       # list these roles found in the metadata directory.
       if tuf.roledb.role_exists(rolename, repository_name) and \
-          rolename not in ['root', 'snapshot', 'timestamp', 'targets']:
+          rolename not in tuf.roledb.TOP_LEVEL_ROLES:
         fileinfodict[metadata_name] = get_metadata_versioninfo(rolename,
             repository_name)
 
@@ -1815,9 +1812,9 @@ def _log_status_of_top_level_roles(targets_directory, metadata_directory,
   # metadata is verified in Root -> Targets -> Snapshot -> Timestamp order.
   # Verify the metadata of the Root role.
   dirty_rolenames = tuf.roledb.get_dirty_roles(repository_name)
-  top_level_roles = ['root', 'targets', 'snapshot', 'timestamp']
 
-  for rolename in top_level_roles:
+  for rolename in tuf.roledb.TOP_LEVEL_ROLES:
+
     listed_filenames = None
     if rolename == 'snapshot':
       listed_filenames = {'root': filenames[ROOT_FILENAME],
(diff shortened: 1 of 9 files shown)


tuf/roledb.py
```diff
@@ -395,14 +422,94 @@ def update_roleinfo(rolename, roleinfo, mark_role_as_dirty=True, repository_name
...
+
+    repository_name:
+      The name of the repository to get the dirty roles.  If not supplied, the
+      'default' repository is searched.
+
+  <Returns>
+    None.
+  """
+
+  global _dirty_roles
```
lukpueh, line 486: Since you don't re-bind `_dirty_roles` you don't need `global` here.

tuf/roledb.py
```diff
@@ -480,7 +584,6 @@ def mark_dirty(roles, repository_name='default'):
   securesystemslib.formats.NAMES_SCHEMA.check_match(roles)
   securesystemslib.formats.NAME_SCHEMA.check_match(repository_name)
 
-  global _roledb_dict
```
lukpueh, line 483: Thanks for removing unnecessary `global` statements. Since you're at it, would you mind removing all of them here in `roledb`? :)

As a reminder, the `global` statement is not necessary for reading or **mutating**  global mutables, such as dictionaries (it is necessary for  **re-binding** though).

If I'm not mistaken we really only need it in `clear_roledb`, but please double-check.

tuf/repository_lib.py
```diff
@@ -394,18 +394,15 @@ def _delete_obsolete_metadata(metadata_directory, snapshot_metadata,
     else:
       logger.debug(repr(metadata_role) + ' found in the snapshot role.')
 
-
-
     # Strip metadata extension from filename.  The role database does not
     # include the metadata extension.
     if metadata_role.endswith(METADATA_EXTENSION):
       metadata_role = metadata_role[:-len(METADATA_EXTENSION)]
-
```
lukpueh, line 403: This is a matter of taste, but I personally like blank lines before an `else` clause. I wouldn't impose it as global code style, but I'd definitely leave it here, also for consistency with the the else above and below.

tuf/repository_lib.py
```diff
@@ -812,7 +809,51 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
...
+  except the top-level roles.
+  """
+
+  filenames = {}
+  metadata_files = sorted(storage_backend.list_folder(metadata_directory),
+      reverse=True)
+  for metadata_role in metadata_files:
+    metadata_path = os.path.join(metadata_directory, metadata_role)
+    metadata_name = \
+      metadata_path[len(metadata_directory):].lstrip(os.path.sep)
```
lukpueh, line 826: I might be missing something here, but isn't `metadata_name == metadata_role`?

tuf/repository_tool.py
```diff
@@ -991,7 +1000,7 @@ def add_signature(self, signature, mark_role_as_dirty=True):
 
 
 
-  def remove_signature(self, signature):
+  def remove_signature(self, signature, mark_role_as_dirty=True):
```
lukpueh, line 1003: Marking roles dirty when removing (or adding) signatures seems a bit odd, since the flag is used elsewhere to signal that the actual payload was modified and that it needs to be re-signed and written to disk.

While I can see how we somehow need to indicate that the signature removal (or addition) needs to be persisted on disk, this is different and also does not need to be propagated to any other metadata.

lukpueh (commented):
Thanks a ton for working on this, @sechkova!! ❤️ 

> Note: Although the best solution is undoubtedly to go for a general rework of the internal metadata representation, this pull request aims to be a short-term fix keeping the changes as close as possible to the current functionality.

I don't want to spoil the fun, and I highly value your efforts, but I wonder if it doesn't make more sense to fix this after or in the course of a broader re-write. We do know how to work around the existing issues by marking metadata we want to be written to disk explicitly as dirty. Doing this automatically seems quite cumbersome and error-prone in the current state of the implementation.

> Note that the dependencies is the implementation are reversed in comparison with the specification [...]

Do I understand correctly that you are pointing out the problem which @awwad has described at great length in #660, i.e. "a delegation is not a property of the delegated-to role."? If so, then yes, that is a problem of the current implementation, at least if we want to support more complex delegation graphs as you describe below.

> add 'delegating_role' to role metadata dictionary.

I see how this is needed in the current design, where the signing key is stored on the delegated-to role. But ideally the delegation and its signing key is stored and modified on the delegating role, so that there is no need to propagate a key change upwards in the delegation graph but only downwards (see first bullet point in https://github.com/theupdateframework/tuf/issues/958#issuecomment-553830755).

> What would be the purpose of a chain of delegations where two roles both delegate to a third? [...]

I am not sure if there is a real world use case for multiple roles delegating trust to the same role (@awwad calls this promiscuous delegation, see #660), maybe others do (cc @mnm678, @JustinCappos). At any rate, and at the risk of repeating myself, signing keys should be modified on the delegating role so there is no upwards propagation of the change. Also downwards propagation, in the case of promiscuous delegation, seems to be a bit tricky and probably depends on the actual scenario.

> While working on this I also noticed that delegating role is not updated at all when the delegated role's key is changed (check #1037).

Wow, thanks for discovering this! I will need to take a deeper look at it.

> do some code cleanup and remove repetitive code from Repository.status() 

Awesome! The cleanup should make it easier to re-think the desired behavior (see first two bullet points of #955).

sechkova:
Thanks @lukpueh :))

> I don't want to spoil the fun, and I highly value your efforts, but I wonder if it doesn't make more sense to fix this after or in the course of a broader re-write. We do know how to work around the existing issues by marking metadata we want to be written to disk explicitly as dirty. Doing this automatically seems quite cumbersome and error-prone in the current state of the implementation.

I do share your concerns ... while working on this PR, it seems I have rediscovered many of the already known delegations issues that you are also pointing at. I would agree that the automatically marking metadata as dirty can wait.
However, it will be great to get your feedback on  #1037 which is a consequence of the same design flow of the implementation (updating the delegated role attributes, while they are kept in the delegating one).

In case we agree to _not_ continue with this fix and keep the workaround of manually marking roles as dirty, I think we can still benefit from this PR up to https://github.com/theupdateframework/tuf/pull/1038/commits/4d86eed1c8a3dff7c3815bedd75b1de2ff4151a5 and the code cleanup here and there (and skip https://github.com/theupdateframework/tuf/pull/1038/commits/93968f2f143818ff9400ed7495a22cc2c5b1120e).
This first part of the PR doesn't add major changes in the behaviour, fixes #964  + marks correctly a new delegated role as dirty (no automatic propagation) which will remove the need of manually marking dirty delegated hashed bins (not the most user-friendly feature :grimacing:)  

I appreciate any other comments that would move this work forward in any direction as well as references to old discussions to avoid repetitions :)

sechkova:
Potential solution of the newly described #1045 in https://github.com/theupdateframework/tuf/pull/1038/commits/3fb09c281511c9e663599fdf619d4058b8a6d836.
Most likely with some additional rework.

jku:
This looks like it's not going to proceed any time soon so I'm closing. If it does re-animate, please re-open

Issue #1039, opened 2020-05-28
Silence the logs when running the test suite


MVrachev:
**Description of issue or feature request**:
After my pull request, https://github.com/theupdateframework/tuf/pull/1029 had been merged there are annoying messages when running the unit tests: 
![image](https://user-images.githubusercontent.com/16246778/83158723-cdf12080-a10d-11ea-9a46-2b891c5558da.png).

**Current behavior**:
Those messages are expected when running the unit tests from `tests/test_updater.py` for example and as such we don't want to see them in a successful test run. 

**Expected behavior**:
We want to be able to receive the `stderr` upon errors in the subprocesses, but on a successful `tox` run I don't expect to see the above error messages.

MVrachev:
One possible solution I tried is by using logger files, pipeline the stderr from the subprocess to them, and finally read from them (if they are not empty) and put the information to the `logger`.

The way I did that is by using the `setUpClass` to create a tmp file.
Create subprocesses running the `simple_server.py` like that:
```
subprocess.Popen(command, stderr=log_file)
```
and finally read from the temp file in the `tearDownClass`, if the file is non-empty.

**Cons**: The messages we don't want to see on a successful tox run like:
```
127.0.0.1 - - [29/May/2020 15:07:48] code 404, message File not found
```
comes from the `stderr` of the process too.
Then how do we know **when** to read from the log file and pass to the logger?

Another question is how should we redirect them.
As `log.info `as `log.erro`r?
If we chose `log.info` then they won’t be shown if the tests succeed right?
But if we log them as `log.error` then they will be shown the same way as now

MVrachev:
Another possible solution is by using try could be by using 
```
pr = subprocess.Popen('ls', stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(output, error) = pr.communicate()
print("OUTPUT IS: ", output)
print("ERROR IS: ", error)
```
**Cons:**
From the documentation about `Peopen.communicate`:
```
Interact with process: Send data to stdin. Read data from stdout and stderr, until end-of-file is reached. Wait for the process to terminate.
```
and the result is that when I use this method the test suite freezes if I don't pass a timeout.

MVrachev:
I have confirmed that those messages `127.0.0.1 - - [29/May/2020 15:07:48] code 404, message File not found` come from the stderr.

[...]

PR #1041, opened 2020-06-02
Silence the log messages during a test run

MVrachev:
**Fixes issue #**: https://github.com/theupdateframework/tuf/issues/1039

1. Filter the `GET` request messages from `tests/slow_retrieval.py`:
image

2. Filter the `code 404, message File not found` error messages:
image

3. Remove `We do not have a console handler` warnings.

requirements**:

### DIFF
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -36,10 +36,15 @@ from __future__ import unicode_literals
 import sys
 import random
 import platform
+import os
+import re
 
 import six
 from six.moves.SimpleHTTPServer import SimpleHTTPRequestHandler
 
+# We won't use the HTTPStatus codes from the http module
+# for backward compability with python 2.7
+FILE_NOT_FOUND = 404
 PORT = 0
 
 def _port_gen():
@@ -61,8 +66,27 @@ else:
 class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
   """A SimpleHTTPRequestHandler that does not write incoming requests to
   stderr. """
+
+  # Override the log_request function
   def log_request(self, code='-', size='-'):
     pass
+
+  # Override the log_error function
+  def log_error(self, format, *args):
+
+    file_not_found_name = os.path.basename(self.path)
+
+    # There are errors in the format "code 404, message File not found"
+    # when downloading \d*.root.json (match digits preffix before ".root.json":
+    # 1.root.json, 10.root.json, etc.) files.
+    # We don't want to show them because they are expected.
+    if len(args) > 0 and args[0] == FILE_NOT_FOUND and \
+        re.match("\d*.root.json", file_not_found_name):
+      pass
+    else:
+      SimpleHTTPRequestHandler.log_error(self, format, *args)
+
+
 
 # NOTE: On Windows/Python2 tests that use this simple_server.py in a
 # subprocesses hang after a certain amount of requests (~68), if a PIPE is
--- a/tests/slow_retrieval_server.py
+++ b/tests/slow_retrieval_server.py
@@ -83,6 +83,9 @@ class Handler(six.moves.BaseHTTPServer.BaseHTTPRequestHandler):
     except IOError as e:
       self.send_error(404, 'File Not Found!')
 
+  # Override log_request
+  def log_request(self, code='-', size='-'):
+    pass
 
 
 def get_random_port():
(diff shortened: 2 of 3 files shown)


MVrachev:
I tested those changes to see if other errors appear.

I checkout on commit 3851f38ca19bf3a4cca8e6643630c6947191f300 just before my fix `Fix [Errno 111] Connection refused`.
Then cherry-picked the commits introduced in this pr and when I ran the tests I was able to see all kinds of errors including `FileNotFoundError` which is good and what we expected.

lukpueh:
Thanks for the patch, @MVrachev! Your solution works nicely but it feels a bit hackish to blacklist a certain undesired output in `simple_server.py`. 

A way to fix this in a more scalable fashion would be to configure the tests so that console output is only displayed if tests fail. I remember doing this for in-toto in https://github.com/in-toto/in-toto/pull/240 (it might also be worth taking a look at https://github.com/in-toto/in-toto/pull/183). Let me know if this helps or if you have any questions. 
 And sorry for not having shared that earlier on #1039.

MVrachev:
> A way to fix this in a more scalable fashion would be to configure the tests so that console output is only displayed if tests fail. I remember doing this for in-toto in in-toto/in-toto#240 (it might also be worth taking a look at in-toto/in-toto#183). Let me know if this helps or if you have any questions.
And sorry for not having shared that earlier on #1039.

Yes, I could do that, but even if you see the logs when the tests fail don't you think we should filter the `code 404, message File not found` messages together with logging the request made by the server started in`slow_retrieval.py`?
After all, those are expected messages. They are not errors.

lukpueh:
I think you can, but I don't think it's necessary, assuming that the messages are only displayed on failure and we expect our tests not to fail. :)

MVrachev:
I have tried a couple of things and here are my thoughts.

**First**, I saw we are using `buffer=True` argument for `TextTestRunner` https://github.com/theupdateframework/tuf/blob/8e6ca67f3131cf2639b997a3e0acf3b4e21c637c/tests/aggregate_tests.py#L95 when we call the tests the documentation says

> sys.stdout and sys.stderr will be buffered in between startTest() and stopTest() being called. Collected output will only be echoed onto the real sys.stdout and sys.stderr if the test fails or errors. Any output is also attached to the failure / error message.


then if I redirect `stderr` and `stdout` in the `subprocess.Popen` calls to `sys.stderr` and `sys.stdout` respectively everything will be okay.
I tried it, but it failed with the following error:
image

After researching what other people had discovered for that problem it seems like  this is the right answer:
https://stackoverflow.com/a/31081957/7058471
because when I try to run the same command in an ipython session it works, but not while running the tests with tox...
image.
Maybe the problem is connected that I am not starting the subprocess with shell, but we know that if we do start it with shell this will be insecure. 


**Second**, I thought then I will look again at what you @lukas are suggesting.
Here, there are a few complications. 
We are starting server processes and we are using them through the unit tests and in order to collect logs from them (as you did in in-toto), we have to do it after when they are killed or in the `tearDownClass` functions.
The problem is that logs cannot be gathered from the `tearDownClass` or `setUpClass` functions.
I tried it a couple of times and end up trying with a simple example:
image

and the result was: 

image

Then, I realized we would need to start all server_process from the `setUp` functions (which are called before each unit test) instead of `setUpClass` functions (which are called before tests in an individual class are executed) and get logs into the `tearDown` functions (called after each unit test) instead of the `tearDownClass` functions (called once per class after all unit tests).
With that method, every unit test that needs a server will start its own subprocess.
I tried it and passed the `stdout=subprocess.PIPE` and `stderr=subprocess.PIPE` arguments to `subprocess.Popen` and using `server_process.communicate()` after that to get the logs.
The results weren't good: sometimes the tests just took forever and sometimes they finished after twice as long as usual...

jku:
> The problem is that logs cannot be gathered from the tearDownClass or setUpClass functions.

WRT our chat before : the loggging does work but if you use a testrunner, it gets to decide what to output. In the example it decides to output only the failing test (and the class setup was obviously run much earlier). If you run the same test without a test runner (so no logging is initialized):

```
if __name__ == '__main__':
    unittest.main()
```
you will see the output

MVrachev:
I have another solution which I want to propose @lukpueh.
Please have a look at pr https://github.com/theupdateframework/tuf/pull/1104

joshuagl:
@MVrachev as you have proposed an alternative solution, should we close this PR?

MVrachev:
Yes, I think we can close that one.

joshuagl:
Closing due to being superseded by #1104

Issue #1045, opened 2020-06-05
load_repository() does not load correctly Targets objects of delegated roles


sechkova:
**Description of issue**:
When a repository is loaded from the filesystem by `repository_tool.load_repository()`, delegated roles' objects are missing from the delegating role. 

**Steps to reproduce:**

Create a longer chain of delegations, for example delegate to hashed bins (the same issue applies to any other delegation):

```
repository.targets.delegate('unclaimed', [public_unclaimed_key], [])
repository.targets('unclaimed').delegate_hashed_bins([], [public_bins_key], 32)
repository.mark_dirty(['00-07', '08-0f', '10-17', '18-1f', '20-27', '28-2f', 
   ...: ...   '30-37', '38-3f', '40-47', '48-4f', '50-57', '58-5f', '60-67', '68-6f', 
   ...: ...   '70-77', '78-7f', '80-87', '88-8f', '90-97', '98-9f', 'a0-a7', 'a8-af', 
   ...: ...   'b0-b7', 'b8-bf', 'c0-c7', 'c8-cf', 'd0-d7', 'd8-df', 'e0-e7', 'e8-ef', 
   ...: ...   'f0-f7', 'f8-ff', 'root', 'targets', 'snapshot', 'timestamp', 'unclaimed'])  
repository.writeall()
```
Check the length of the delegations:
```
In [6]: len(repository.targets._delegated_roles)                                                  
Out[6]: 33
In [7]: len(repository.targets('unclaimed')._delegated_roles)                                     
Out[7]: 32
```

Load the created repository from disk:
```
repository = load_repository('repository')
In [5]: len(repository.targets._delegated_roles)                                                  
Out[5]: 33

In [6]: len(repository.targets('unclaimed')._delegated_roles)  
Out[6]: 0

repository.targets('unclaimed').add_target_to_bin('t1', 32)
Error: unclaimed does not have a delegated role 60-67
```

**Current behavior**:
Loaded delegated roles' objects are missing from the delegating object if the role is not the top-level 'targets' role.

**Expected behavior**:
After loading from filesystem, the delegated roles are added to the delegating role's `_delegated_roles` list.

lukpueh:
Excellent work, @sechkova! Out of curiosity I just compared `tuf.roledb._roledb_dict` before and after reloading using below snippet. Looks like the issue you describe above only exists in the repository_tool metadata model and not in the roledb metadata model.
[...]

PR #1052, opened 2020-06-10
Load correctly the delegated Targets objects hierarchy

sechkova:
**Fixes issue #**: #1045 

- Update `load_repository()` function to load the delegations metadata starting from 'targets' and traversing downwards the delegated roles in order to load correctly the delegations hierarchy.
- Improve readability by extracting the loading of `metadata_directory` files in a separate function `get_delegations_filenames()` in a similar way in which top level metadata files are loaded by `get_metadata_filenames()`.
- Update `test_load_repository`.

requirements**:

### DIFF
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -811,6 +811,59 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
   return private_key
 
 
+
+def get_delegations_filenames(metadata_directory, consistent_snapshot,
+    storage_backend=None):
+  """
+  Return a dictionary containing all filenames in 'metadata_directory'
+  except the top-level roles.
+  If multiple versions of a file exist because of a consistent snapshot,
+  only the file with biggest version prefix is included.
+  """
+
+  filenames = {}
+  loaded_metadata = []
+  metadata_files = sorted(storage_backend.list_folder(metadata_directory),
+      reverse=True)
+  for metadata_role in metadata_files:
+    metadata_path = os.path.join(metadata_directory, metadata_role)
+
+    # Strip the version number if 'consistent_snapshot' is True,
+    # or if 'metadata_role' is Root.
+    # Example:  '10.django.json' --> 'django.json'
+    consistent_snapshot = \
+      metadata_role.endswith('root.json') or consistent_snapshot == True
+    metadata_name, junk = _strip_version_number(metadata_role,
+      consistent_snapshot)
+
+    if metadata_name.endswith(METADATA_EXTENSION):
+      extension_length = len(METADATA_EXTENSION)
+      metadata_name = metadata_name[:-extension_length]
+
+    else:
+      logger.debug('Skipping file with unsupported metadata'
+          ' extension: ' + repr(metadata_path))
+      continue
+
+    # Skip top-level roles, only interested in delegated roles now that the
+    # top-level roles have already been loaded.
+    if metadata_name in ['root', 'snapshot', 'targets', 'timestamp']:
+      continue
+
+    # Keep a store of metadata previously loaded metadata to prevent re-loading
+    # duplicate versions.  Duplicate versions may occur with
+    # 'consistent_snapshot', where the same metadata may be available in
+    # multiples files (the different hash is included in each filename).
+    if metadata_name in loaded_metadata:
+      continue
+
+    filenames[metadata_name] = metadata_path
+    loaded_metadata.append(metadata_name)
+
+  return filenames
+
+
+
 def get_metadata_filenames(metadata_directory):
   """
   <Purpose>
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -812,8 +812,8 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
 
 
 
-def get_delegations_filenames(metadata_directory, consistent_snapshot,
-    storage_backend=None):
+def get_delegated_roles_metadata_filenames(metadata_directory,
+    consistent_snapshot, storage_backend=None):
   """
   Return a dictionary containing all filenames in 'metadata_directory'
   except the top-level roles.
@@ -861,7 +861,7 @@ def get_delegations_filenames(metadata_directory, consistent_snapshot,
 
 
 
-def get_metadata_filenames(metadata_directory):
+def get_top_level_metadata_filenames(metadata_directory):
   """
   <Purpose>
     Return a dictionary containing the filenames of the top-level roles.
@@ -1826,7 +1826,7 @@ def _log_status_of_top_level_roles(targets_directory, metadata_directory,
 
   # The expected full filenames of the top-level roles needed to write them to
   # disk.
-  filenames = get_metadata_filenames(metadata_directory)
+  filenames = get_top_level_metadata_filenames(metadata_directory)
   root_filename = filenames[ROOT_FILENAME]
   targets_filename = filenames[TARGETS_FILENAME]
   snapshot_filename = filenames[SNAPSHOT_FILENAME]
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -175,7 +175,7 @@ def _generate_and_write_metadata(rolename, metadata_filename,
   else:
     logger.debug('Not incrementing ' + repr(rolename) + '\'s version number.')
 
-  if rolename in ['root', 'targets', 'snapshot', 'timestamp'] and not allow_partially_signed:
+  if rolename in tuf.roledb.TOP_LEVEL_ROLES and not allow_partially_signed:
     # Verify that the top-level 'rolename' is fully signed.  Only a delegated
     # role should not be written to disk without full verification of its
     # signature(s), since it can only be considered fully signed depending on
@@ -394,18 +394,15 @@ def _delete_obsolete_metadata(metadata_directory, snapshot_metadata,
     else:
       logger.debug(repr(metadata_role) + ' found in the snapshot role.')
 
-
-
     # Strip metadata extension from filename.  The role database does not
     # include the metadata extension.
     if metadata_role.endswith(METADATA_EXTENSION):
       metadata_role = metadata_role[:-len(METADATA_EXTENSION)]
-
     else:
       logger.debug(repr(metadata_role) + ' does not match'
           ' supported extension ' + repr(METADATA_EXTENSION))
 
-    if metadata_role in ['root', 'targets', 'snapshot', 'timestamp']:
+    if metadata_role in tuf.roledb.TOP_LEVEL_ROLES:
       logger.debug('Not removing top-level metadata ' + repr(metadata_role))
       return
 
@@ -850,7 +847,7 @@ def get_delegated_roles_metadata_filenames(metadata_directory,
       continue
 
     # Skip top-level roles, only interested in delegated roles.
-    if metadata_name in ['root', 'snapshot', 'targets', 'timestamp']:
+    if metadata_name in tuf.roledb.TOP_LEVEL_ROLES:
       continue
 
     # Prevent reloading duplicate versions if consistent_snapshot is True
@@ -1131,7 +1128,7 @@ def generate_root_metadata(version, expiration_date, consistent_snapshot,
   # Extract the role, threshold, and keyid information of the top-level roles,
   # which Root stores in its metadata.  The necessary role metadata is generated
   # from this information.
-  for rolename in ['root', 'targets', 'snapshot', 'timestamp']:
+  for rolename in tuf.roledb.TOP_LEVEL_ROLES:
 
     # If a top-level role is missing from 'tuf.roledb.py', raise an exception.
     if not tuf.roledb.role_exists(rolename, repository_name):
@@ -1507,7 +1504,7 @@ def generate_snapshot_metadata(metadata_directory, version, expiration_date,
       # snapshot and timestamp roles are not listed in snapshot.json, do not
       # list these roles found in the metadata directory.
       if tuf.roledb.role_exists(rolename, repository_name) and \
-          rolename not in ['root', 'snapshot', 'timestamp', 'targets']:
+          rolename not in tuf.roledb.TOP_LEVEL_ROLES:
         fileinfodict[metadata_name] = get_metadata_versioninfo(rolename,
             repository_name)
 
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -822,9 +822,13 @@ def get_delegations_filenames(metadata_directory, consistent_snapshot,
   """
 
   filenames = {}
-  loaded_metadata = []
   metadata_files = sorted(storage_backend.list_folder(metadata_directory),
       reverse=True)
+
+  # Iterate over role metadata files, sorted by their version-number prefix, with
+  # more recent versions first, and only add the most recent version of any
+  # (non top-level) metadata to the list of returned filenames. Note that there
+  # should only be one version of each file, if consistent_snapshot is False.
   for metadata_role in metadata_files:
     metadata_path = os.path.join(metadata_directory, metadata_role)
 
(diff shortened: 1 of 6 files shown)


trishankatdatadog:
Thanks, Theodora!

I have just one concern: how does this handle the possibility that Alice and Bob may delegate to Charlie using different sets and thresholds of keys from each other?

lukpueh:
💯, 🎉, 🥳, @sechkova!!

> I have just one concern: how does this handle the possibility that Alice and Bob may delegate to Charlie using different sets and thresholds of keys from each other?

@trishankatdatadog, the current reference implementation already does not support such "promiscuous" delegations (see #589 and for a more general description of the issue #660). 
Supporting that behavior is way out of the scope of #1045.

trishankatdatadog:
> "promiscuous" delegations

Got it. Gods, I hate that term, let's find another name, please...

joshuagl:
> > "promiscuous" delegations
> 
> Got it. Gods, I hate that term, let's find another name, please...

Agreed. Promiscuous has negative connotations.

In theupdateframework/specification#96 Marina used the term multi-role delegations

trishankatdatadog:
> In [theupdateframework/specification#96](https://github.com/theupdateframework/specification/issues/96) Marina used the term multi-role delegations

Thanks, Josh. I don't think the problem is specific to multi-role delegations. I believe the idea is that Alice and Bob, on entirely different parts of the graph, can delegate to Charlie using entirely different thresholds and keys. I don't know what's a good, succinct word for this, except to say something like "delegator-dependent keys for a delegatee" (rolls off the tongue, I know).

lukpueh:
Yes, multi-role delegations as described in [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md) are something else. It's a one-to-many delegation where the delegated-to roles need to agree on a target.

Promiscuous delegations are many-to-many delegations.

joshuagl:
This looks good to me. There's some nice clean-ups of the modified code in here too, thanks @sechkova

tuf/repository_lib.py
```diff
@@ -811,6 +811,59 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
   return private_key
 
 
+
+def get_delegations_filenames(metadata_directory, consistent_snapshot,
```
lukpueh: Minor nit: Would you mind calling the function something like `get_delegated_role_metadata_filenames` just so that it's clear that `delegation != role`?  (see #660) :)

lukpueh: ... or maybe `get_non_top_level_metadata_filenames`? 🤷

tuf/repository_lib.py
```diff
@@ -811,6 +811,59 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
...
+      extension_length = len(METADATA_EXTENSION)
+      metadata_name = metadata_name[:-extension_length]
+
+    else:
+      logger.debug('Skipping file with unsupported metadata'
+          ' extension: ' + repr(metadata_path))
+      continue
+
+    # Skip top-level roles, only interested in delegated roles now that the
+    # top-level roles have already been loaded.
```
lukpueh: I think you can remove the comment, the "now that the  top-level roles have already been loaded" part is not in the scope of this function anymore and given the function name it should be clear what's happening.

tuf/repository_lib.py
```diff
@@ -811,6 +811,59 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
...
+    if metadata_name in ['root', 'snapshot', 'targets', 'timestamp']:
+      continue
+
+    # Keep a store of metadata previously loaded metadata to prevent re-loading
+    # duplicate versions.  Duplicate versions may occur with
+    # 'consistent_snapshot', where the same metadata may be available in
+    # multiples files (the different hash is included in each filename).
```
lukpueh: The comment has a stray "metadata" (first line) and is a bit outdated, as consistent role metadata does not use a hash-prefix anymore but a version number (last line). A comment like this still might be helpful, but maybe rather above, where we define `loaded_metadata` and  call `sorted(..., reverse=True)`.
What do you think about something like this directly above the for loop?

```python
# Iterate over role metadata files, sorted by their version-number prefix, with
# more recent versions first, and only add the most recent version of any
# (non top-level) metadata to the list of returned filenames. Note that there
# should only be one version of each file, if consistent_snapshot is False.
 ```

File: tuf/repository_lib.py [hunk 4]
```diff
@@ -811,6 +811,59 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
...
+    if metadata_name in loaded_metadata:
+      continue
+
+    filenames[metadata_name] = metadata_path
+    loaded_metadata.append(metadata_name)
```
lukpueh: Why not just 
```python
if metadata_name not in filenames:
  filenames[metadata_name] = metadata_path
```
Should be a lot faster (O(1) vs. O(n)), less code and less memory. :)

lukpueh (changes requested):
Thanks again for this great patch, @sechkova! This is just a partial review that touches `repository_lib.py` only. I'll follow up with the rest tomorrow. 

Btw. I am aware that the things I flagged were already there before, but since they are in your diff, I'm asking you to fix them. :P

tuf/repository_tool.py
```diff
@@ -3104,45 +3104,24 @@ def load_repository(repository_directory, repository_name='default',
...
 
-    metadata_path = os.path.join(metadata_directory, metadata_role)
-    metadata_name = \
-      metadata_path[len(metadata_directory):].lstrip(os.path.sep)
+  # Top-level roles are already loaded, fetch targets and get its delegations.
+  # Collect a list of delegated-delegating role pairs, starting from the
+  # top-level targets: [('role1', 'targets'), ('role2', 'targets'), ... ]
+  roleinfo = tuf.roledb.get_roleinfo('targets', repository_name)
+  for role in roleinfo['delegations']['roles']:
+    delegations.append([role['name'], 'targets'])
```
lukpueh: Minor nit: The comment suggests delegations are list of tuples, but the code makes it a list of lists. Please change either one. IMO tuples are fine.

tuf/repository_tool.py
```diff
@@ -3170,18 +3149,26 @@ def load_repository(repository_directory, repository_name='default',
...
+        new_targets_object)
+    targets_objects[delegating_role].add_delegated_role(rolename,
+        new_targets_object)
 
-    targets_object._delegated_roles[(os.path.basename(metadata_name))] = \
-        new_targets_object
+    # Append the next level delegations to the list:
+    # the 'delegated' role becomes the 'delegating'
+    for delegation in metadata_object['delegations']['roles']:
+      delegations.append([delegation['name'], rolename])
```
lukpueh: Wow, this is cool stuff! But I admit it took me a while to understand what's happening. You're adding delegations to the list of delegations as you iterate over them, thus traversing the tree of delegations.  :)

Out of curiosity, is there a name for this pattern? 

At any rate, we should add a big comment that describes it above the outer loop definition (L3123). And we should definitely point out that cycles in the delegation graph, will make this an infinite loop.

lukpueh (commented):
Thanks for the excellent work, @sechkova, you're quite the magician! :) Please consider improving the code comment as suggested inline (and my review comments from yesterday). Otherwise this LGTM!

trishankatdatadog:
Teodora is very good!

tuf/repository_lib.py
sechkova: Done in here: da09a22

tuf/repository_lib.py
sechkova: Also done in da09a22 , thanks for suggesting the text which I directly reused ;) 
I still left one line comment for the lazy ones who won't scroll up to the start of the for loop.

tuf/repository_lib.py
sechkova: Thanks for noticing, here it is da09a22

tuf/repository_lib.py
sechkova: 35f72a3 I renamed both get_*_metadata_filenames ... :shrug:

tuf/repository_tool.py
sechkova: And tuples they are!

tuf/repository_tool.py
sechkova: Let's call it a custom breadth-first search :grin: 
I tried to improve it after you pointed out the potential infinite loop. 
The loaded delegations were still kept in the list of delegations anyway so the options to avoid repetitions seem to be:
- `if x is in list` - O(n)
- replace the list with a dictionary for O(1) lookup but dictionaries are guaranteed to be ordered only in Python 3.x so the traversal won't work
- use a `deque` for traversing the delegations (keep only the next in line) + `set` for storing already loaded ones

If this seems unnecessary overhead since cycles in the delegations are not likely to happen, we can keep only the `deque` and get rid of the `set` + adding a comment.

Hopefully the new comments help to make things look less magical :slightly_smiling_face: 
(and this is the new commit c79ea55)

tuf/repository_tool.py
```diff
@@ -3170,18 +3166,24 @@ def load_repository(repository_directory, repository_name='default',
...
+    targets_objects[rolename] = new_targets_object
 
-    # Generate the Targets objects of the delegated roles of 'metadata_name'
-    # and add it to the top-level 'targets' object.
-    new_targets_object = Targets(targets_directory, metadata_name, roleinfo,
-        repository_name=repository_name)
-    targets_object = targets_objects['targets']
-    targets_objects[metadata_name] = new_targets_object
+    targets_objects['targets'].add_delegated_role(rolename,
+        new_targets_object)
```
lukpueh: **Off-topic:**
I find it extremely odd that the top-level `Targets` object does not make a difference between roles it delegates to directly and indirectly (via other roles).

tuf/repository_tool.py
lukpueh: Haha, I think it just became even more magical. 😆 

Regarding the options:

Using a set as an infinite loop safe-guard seems like a good idea. And definitely better than using a list.
I think we would want to warn the caller if we encounter a loop, what do you think? To me, loops in the delegation graph don't seem desirable. But I may be missing something.

The need for `deque` is not fully obvious to me. Are you doing this to save memory? If so, I suppose that's fine. Maybe you can leave a note about the motivation. :)

tuf/repository_lib.py
```diff
@@ -811,7 +811,57 @@ def import_ed25519_privatekey_from_file(filepath, password=None):
...
+      extension_length = len(METADATA_EXTENSION)
+      metadata_name = metadata_name[:-extension_length]
+
+    else:
+      logger.debug('Skipping file with unsupported metadata'
+          ' extension: ' + repr(metadata_path))
+      continue
+
+    # Skip top-level roles, only interested in delegated roles.
+    if metadata_name in ['root', 'snapshot', 'targets', 'timestamp']:
```
joshuagl: We have this kind of pattern in enough places that I think it would be worthwhile to add a constant for the list of top-level role names.

lukpueh: Agreed. Maybe we can cherry-pick https://github.com/theupdateframework/tuf/pull/1038/commits/7100dc352bea837fc882db35e0d276f7bfbdf1d4

joshuagl: I thought I'd seen that somewhere already, and the same author too. 😄

tuf/repository_tool.py
```diff
@@ -3100,50 +3102,44 @@ def load_repository(repository_directory, repository_name='default',
...
-    else:
-      logger.debug('Skipping file with unsupported metadata'
-          ' extension: ' + repr(metadata_path))
-      continue
-
-    # Skip top-level roles, only interested in delegated roles now that the
-    # top-level roles have already been loaded.
-    if metadata_name in ['root', 'snapshot', 'targets', 'timestamp']:
+  # A deque used to keep the next delegation to be loaded
+  delegations = deque()
```
joshuagl, line 3117: I'm curious about the use of a deque here, can you explain the rationale a bit?
AFAICT you're just using it as a FIFO, at which point a list with `pop(0)` would work just as well and would be a bit more familiar.

sechkova: Answering both of your comments @lukpueh @joshuagl :

Yes, @joshuagl got it right, the idea is to use a FIFO and continuously append the next delegations while removing(`pop(0)/popleft()`) the loaded one (unlike in my first implementation where the delegations were only appended and never removed from the list). 

@joshuagl my reasoning about [list vs deque as a FIFO](https://docs.python.org/3/tutorial/datastructures.html#using-lists-as-queues) 

Using only a FIFO structure would be enough if there were no potential cycles in the graph, but since what we want to achieve is traverse the graph AND avoid loops, we need one ordered structure + one fast lookup structure, hence the deque+set (or one that can do both which would be the dict in Python 3.x).

_And for a completeness of this discussion ..._ 
On the other hand, up to now we don't have exact constraints in `load_repository()` regarding the graph traversal  and we can as well use a [list as a stack (LIFO)](https://docs.python.org/3/tutorial/datastructures.html#using-lists-as-stacks) which will result in a depth-first-search type of traversal.

I'm open to suggestions about less magical code or more pythonic way of implementing this :)

lukpueh: > Yes, @joshuagl got it right, the idea is to use a FIFO and continuously append the next delegations while removing(pop(0)/popleft()) the loaded one (unlike in my first implementation where the delegations were only appended and never removed from the list).

Sure, but your first approach already implemented a FIFO queue, didn't it? The only difference that I saw is that the list grows in your first approach but does not in your second. 

```
# Go from left to right, adding to the right
  ↓
| 0 | 
      ↓
| 0 | 1 |
          ↓
| 0 | 1 | 2 | 
```

```
# Pop from left, add to the right
  ↓
| 0 |
      ↓
    | 1 | 
          ↓
        | 2 |
```

lukpueh: > hence the deque+set (or one that can do both which would be the dict in Python 3.x).

Haven't checked, but would `collections.OrderedDict` have the desired properties cross-version?

lukpueh: > On the other hand, up to now we don't have exact constraints in load_repository() regarding the graph traversal and we can as well use a list as a stack (LIFO) which will result in a depth-first-search type of traversal.

AFAIU, when loading the delegation graph into memory it really shouldn't matter if we do depth or breadth first. This is only important when verifying targets.

sechkova: > > Yes, @joshuagl got it right, the idea is to use a FIFO and continuously append the next delegations while removing(pop(0)/popleft()) the loaded one (unlike in my first implementation where the delegations were only appended and never removed from the list).
> 
> Sure, but your first approach already implemented a FIFO queue, didn't it? The only difference that I saw is that the list grows in your first approach but does not in your second.
> 
> ```
> # Go from left to right, adding to the right
>   ↓
> | 0 | 
>       ↓
> | 0 | 1 |
>           ↓
> | 0 | 1 | 2 | 
> ```
> 
> ```
> # Pop from left, add to the right
>   ↓
> | 0 |
>       ↓
>     | 1 | 
>           ↓
>         | 2 |
> ```

Yes, yes, exactly :) 
Then,  1) Yes, the goal here was to save memory 2) This could've been achieved with a list and pop(0) but this is a slow operation so here comes the deque "which was designed to have fast appends and pops from both ends"

sechkova: > > hence the deque+set (or one that can do both which would be the dict in Python 3.x).
> 
> Haven't checked, but would `collections.OrderedDict` have the desired properties cross-version?

Probably it would do the job, I think I had overlooked it because it sounded outdated from the perspective of the newer Python versions ...

tuf/repository_tool.py
sechkova: > Using a set as an infinite loop safe-guard seems like a good idea. And definitely better than using a list.
> I think we would want to warn the caller if we encounter a loop, what do you think? To me, loops in the delegation graph don't seem desirable. But I may be missing something.

It could be better to warn during the creation of the delegation ... but I think currently there is no mechanism to detect that so ...  better late than never?

tuf/repository_tool.py
lukpueh: IMO really any of the approaches discussed here (FIFO or LIFO with list + set, deque + set or OrderedDict) is absolutely fine, I think I would even have been okay with not guarding against infinite loops, as long as it is documented. :)

So what do you think about keeping it as it is and adding another comment about saving memory by using a deque + leftpop?

lukpueh: + warning the user when a loop is encountered.

sechkova: > * warning the user when a loop is encountered.

Is this related to the sentence above (_"okay with not guarding against infinite loops"_) but appearing as a bullet? 😁 
Otherwise, yep, it works for me :)

lukpueh: haha, this was meant to be a `+` as in:

*... what do you think  about adding another comment ... **plus**  warning the user when a loop is encountered?*

sechkova:
f1a6676 : Amended with an extended comment about the `deque` usage and a warning in case of a cycle in the graph
97eff9e : Discard the dictionary used for temporary storing `new_targets_objects` and reference them by the `parent_targets_object` instead 
6ae3ea6 : cherry-picked 7100dc3

lukpueh (approved):
🥇 Thanks for addressing all the comments! :)

PR #1054, opened 2020-06-17
Update docs on crypto details

jku:
The specifics on crypto dependencies are out of date in a few places: fix setup.py comments and Tutorial.

Obviously this PR reflects my understanding of the situation -- and I'm a newbie here so I may have misunderstood. Review is welcome.

There's a potential additional fix in the Tutorial section that I'm not sure about:

> Which cryptography library to use is determined by the default, or user modified, settings in settings.py

This seems false or misleading -- only hash algorithms can be selected in settings and the signature algorithm selection is done when keys are generated, right?

### DIFF
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -65,15 +65,17 @@ updates.
 The repository tool supports multiple public-key algorithms, such as
 [RSA](https://en.wikipedia.org/wiki/RSA_%28cryptosystem%29) and
 [Ed25519](https://ed25519.cr.yp.to/), and multiple cryptography libraries.
-Which cryptography library to use is determined by the default, or user modified,
-settings in [settings.py](../tuf/settings.py).
 
 Using [RSA-PSS](https://tools.ietf.org/html/rfc8017#section-8.1) or
 [ECDSA](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm)
 signatures requires the [cryptography](https://cryptography.io/) library. If
 generation of Ed25519 signatures is needed
-[PyNaCl](https://github.com/pyca/pynacl) library should be installed.
-Ed25519 and ECDSA keys are stored in JSON format and RSA keys are stored in PEM
+[PyNaCl](https://github.com/pyca/pynacl) library should be installed. This
+tutorial assumes both dependencies are installed: refer to
+[Installation Instructions](INSTALLATION.rst#install-with-more-cryptographic-flexibility)
+for details.
+
+The Ed25519 and ECDSA keys are stored in JSON format and RSA keys are stored in PEM
 format. Private keys are encrypted and passphrase-protected (strengthened with
 PBKDF2-HMAC-SHA256.)  Generating, importing, and loading cryptographic key
 files can be done with functions available in the repository tool.
@@ -81,7 +83,6 @@ files can be done with functions available in the repository tool.
 To start, a public and private RSA key pair is generated with the
 `generate_and_write_rsa_keypair()` function.  The keys generated next are
 needed to sign the repository metadata files created in upcoming sub-sections.
-
 
 Note:  In the instructions below, lines that start with `>>>` denote commands
 that should be entered by the reader, `#` begins the start of a comment, and
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -68,14 +68,13 @@ The repository tool supports multiple public-key algorithms, such as
 Which cryptography library to use is determined by the default, or user modified,
 settings in [settings.py](../tuf/settings.py).
 
-The [PyCrypto](https://www.dlitz.net/software/pycrypto/) library may be
-selected to generate RSA keys and
-[RSA-PSS](https://en.wikipedia.org/wiki/RSA-PSS) signatures.  If generation of
-Ed25519 signatures is needed, the [PyNaCl](https://github.com/pyca/pynacl)
-library setting should be enabled.  PyNaCl is a Python binding to the
-Networking and Cryptography Library.  For key storage, RSA keys may be stored
-in PEM or JSON format, and Ed25519 keys in JSON format.  Private keys, for both
-RSA and Ed25519, are encrypted and passphrase-protected (strengthened with
+Using [RSA-PSS](https://tools.ietf.org/html/rfc8017#section-8.1) or
+[ECDSA](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm)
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -40,12 +40,9 @@
...
   # Installing optional requirements (i.e., after installing tuf).
-  # The 'tools' optional requirement is currently supported, which enables
-  # fast and secure ed25519 key generation, and signature verification
-  # computations with PyNaCl+libsodium.  General-purpose cryptography is also
-  # provided.  'tools' is needed by the TUF repository tools.  Clients that
-  # require verification of RSASSA-PSS signatures must also install tuf[tools].
-  $ pip install tuf[tools]
+  # Support for creation of Ed25519 signatures and support for RSA and ECDSA
+  # signatures in general requires optional dependencies:
+  $ pip install securesystemslib[crypto,pynacl]
```
lukpueh, line 45: IIRC `securesystemslib` with `crypto` and `pynacl` extras must be installed **before** `tuf`. The other way around pip will see securesystemslib as already installed (albeit w/o the extras) and just skip it. The ["Install with More Cryptographic Flexibility" section in the installation guide](https://github.com/theupdateframework/tuf/blob/develop/docs/INSTALLATION.rst#install-with-more-cryptographic-flexibility) has got this right.

Looking at the full docstring here I wonder if it isn't just better to point to the installation guide, for the sake of docs deduplication. Also, the `setup.py` commands don't seem particularly useful.

What do you think, @jku?

docs/TUTORIAL.md
```diff
@@ -68,14 +68,13 @@ The repository tool supports multiple public-key algorithms, such as
 Which cryptography library to use is determined by the default, or user modified,
 settings in [settings.py](../tuf/settings.py).
```
lukpueh: You are absolutely right, this setting was moved from TUF to securesystemslib, when securesystemslib was forked off of TUF, and has been removed from there when PyCrypto as alternative to cryptography was dropped.
Would you mind removing the sentence?

I would say the line above is still true as we do support multiple algorithms (rsa, ecdsa and ed25519) and multiple libraries (OpenSSL/cryptography for rsa and ecdsa, and libsodium/PyNacl  for ed25519, plus a pure python implementation for ed25519 signature verification).

docs/TUTORIAL.md
```diff
@@ -68,14 +68,13 @@ The repository tool supports multiple public-key algorithms, such as
...
-Networking and Cryptography Library.  For key storage, RSA keys may be stored
-in PEM or JSON format, and Ed25519 keys in JSON format.  Private keys, for both
-RSA and Ed25519, are encrypted and passphrase-protected (strengthened with
+Using [RSA-PSS](https://tools.ietf.org/html/rfc8017#section-8.1) or
+[ECDSA](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm)
+signatures requires the [cryptography](https://cryptography.io/) library. If
+generation of Ed25519 signatures is needed
+[PyNaCl](https://github.com/pyca/pynacl) library should be installed.
+Ed25519 and ECDSA keys are stored in JSON format and RSA keys are stored in PEM
+format. Private keys are encrypted and passphrase-protected (strengthened with
```
lukpueh, line 79: Do you think it might be worth informing the reader that the snippet in this tutorial require both RSA and Ed25519 signing capabilities, and thus TUF should be installed with [full crypto support](https://github.com/theupdateframework/tuf/blob/develop/docs/INSTALLATION.rst#install-with-more-cryptographic-flexibility)?

lukpueh (commented):
Many thanks for taking the time to research this and fixing the outdated docs, it is highly appreciated, @jku!

To add some context, @awwad and I compiled a list of docs issues a while ago, which can be found in #808. I fixed many of the tutorial-related issues in #775, but did focus on the code snippets. The accompanying text still requires a larger overhaul (as noted in item one of https://github.com/theupdateframework/tuf/issues/808#issuecomment-487596129). 

So I'm very happy about contributions like this one!  Let me know if you are interested in addressing my comments below, otherwise we can do this in a follow-up PR.

jku:
> Let me know if you are interested in addressing my comments below, otherwise we can do this in a follow-up PR.

I'm leaving computers for a few days to celebrate midsummer in a moment but we can leave it open and I can tweak the PR next week

lukpueh:
Thanks and happy midsummer!

setup.py
jku: > The other way around pip will see securesystemslib as already installed (albeit w/o the extras) and just skip it

I don't think this is true on current pip version at least: 

```
$ pip list | grep securesystemslib
securesystemslib    0.15.0
$ python3 -c "import nacl"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'nacl'
$ pip install securesystemslib[pynacl]
...
Successfully installed pynacl-1.4.0
$ python3 -c "import nacl"
$ 

```

The point about duplicated documentation is quite valid

lukpueh: Oh cool, then I either misremembered or this was changed in a recent release. :)

docs/TUTORIAL.md
jku: yes, done

jku:
I did not remove the install examples from setup.py yet: Replacing them with a git url (like https://github.com/theupdateframework/tuf/blob/develop/docs/INSTALLATION.rst) in  released/installed files feels unclean... but I can do it if needed.

Other comments resolved.

docs/TUTORIAL.md
```diff
@@ -65,25 +65,25 @@ updates.
...
-in PEM or JSON format, and Ed25519 keys in JSON format.  Private keys, for both
-RSA and Ed25519, are encrypted and passphrase-protected (strengthened with
+
+Using [RSA-PSS](https://tools.ietf.org/html/rfc8017#section-8.1) or
+[ECDSA](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm)
+signatures requires the [cryptography](https://cryptography.io/) library. If
+generation of Ed25519 signatures is needed
+[PyNaCl](https://github.com/pyca/pynacl) library should be installed. This
+tutorial assumes both dependencies are installed: refer to
+[installation Instructions](INSTALLATION.rst#install-with-more-cryptographic-flexibility)
```
lukpueh: Minor nit: I'd write "Installation Instructions" or "installation instructions" :)

lukpueh (changes requested):
Thanks for the updates! I have one minor request. Then we can merge.

docs/TUTORIAL.md
jku: oops, fixed

lukpueh (approved):
🎉

Issue #1048, opened 2020-06-08
support updating individual metadata upon addition of target file


lukpueh:
The repository_tool and the underlying repository_lib and roledb modules seem to live on the assumption that all roles (top-level roles and delegated targets roles) are required in memory. 

In large repositories, and this is the case for the implementation of PEP 458, it might not be feasible nor necessary to load and/or have the entire repository metadata in memory at all times.

To be more specific, during ongoing operations, when adding a new target file the following operations are required:

1. find and update relevant targets metadata, e.g. "bin-n" (i.e. target file hashes and size + version number + signature)
1. update snapshot (i.e. targets metadata version number + own version number + signature)
1. update timestamp (i.e. snapshot version number + own version number + signature)

The repository tool should provide API function to support above activities without requiring any other TUF metadata on disk or in memory.

trishankatdatadog:
Hopefully, this is not too hard to do. Happy to help on weeknights and weekends.

lukpueh:
We have to decide if we make this a stand-alone tool, or integrate it with the current repo_tool + repo_lib + roledb + keydb stack, or if we do a combination of both.

That latter seems the most feasible for PEP 485 work. I am thinking of something like a "detached head mode" (apologies for the subpar analogy), where we use some of the existing infrastructure, but don't interfere with functions like `load_repository` or `writeall`.

If we go for a stand-alone tool, we should make use of the thoughts @joshuagl, @sechkova and I have jotted down with the goal to clean up above mentioned stack: https://docs.google.com/document/d/13cY0MD3uzwFYGnj_GkPS_rwvDEzzNpDu6cteyeYP6fY

trishankatdatadog:
@woodruffw @lukpueh @joshuagl @sechkova @MVrachev @mnm678 @SantiagoTorres

Will and I sketched what we hope is a [super-simple API for a TUF repo](https://hackmd.io/8e7WK_EZSBebU66nLMPCQQ)... WDYT?

trishankatdatadog:
I will start hacking on this

joshuagl:
WIP in #1112

joshuagl:
For the purposes of the PEP 458 integration this has been resolved by the merging of #1112.

PR #1058, opened 2020-06-26
Make some helper functions for determining hashed bins part of repository_lib

joshuagl:
**Fixes issue #**: N/A

In order to support users of the WIP low-level API for updating individual metadata on addition of a target (see #1048), specifically Warehouse and PEP 458, make some of the helper functions for determining the appropriate bin for a given target part of the public API in `repository_lib`.

requirements**:

### DIFF
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -980,6 +980,114 @@ def get_metadata_versioninfo(rolename, repository_name):
 
 
 
+def create_bin_name(low, high, prefix_len):
+  """
+  <Purpose>
+    Create a string name of a delegated hash bin, where name will be a range of
+    zero-padded (up to prefix_len) strings i.e. for low=00, high=07,
+    prefix_len=3 the returned name would be '000-007'.
+
+  <Arguments>
+    low:
+      The low end of the prefix range to be binned
+
+    high:
+      The high end of the prefix range to be binned
+
+    prefix_len:
+      The length of the prefix range components
+
+  <Returns>
+    A string bin name, with each end of the range zero-padded up to prefix_len
+  """
+  if low == high:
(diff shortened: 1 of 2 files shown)


joshuagl:
TODO: these functions don't yet have explicit tests (though they are tested implicitly by the `delegate_hashed_bins` tests)

tuf/repository_lib.py
```diff
@@ -980,6 +980,114 @@ def get_metadata_versioninfo(rolename, repository_name):
...
+    raise securesystemslib.exceptions.Error('The "number_of_bins" argument'
+        ' must be a power of 2.')
+
+  return prefix_length, prefix_count, bin_size
+
+
+
+
+
+def find_bin_for_hash(path_hash, number_of_bins):
```
joshuagl: Should perhaps be `find_bin_for_target_hash()` for better consistency with `get_target_hash()`

lukpueh (commented):
Thanks for relocating the code, @joshuagl! I support your `find_bin_for_target_hash` rename suggestion.

Also, I'm for moving on without adding explicit tests. The `delegate_hashed_bins` tests cover the functionality sufficiently for the time being. And we can still add tests when we set the public API in stone.  What do you think?

joshuagl:
> Thanks for relocating the code, @joshuagl! I support your `find_bin_for_target_hash` rename suggestion.

Pushed an update with the rename in ed45fb2

> Also, I'm for moving on without adding explicit tests. The `delegate_hashed_bins` tests cover the functionality sufficiently for the time being. And we can still add tests when we set the public API in stone. What do you think?

I think this makes sense, thanks @lukpueh

lukpueh (approved):
🎉

PR #1060, opened 2020-06-26
A simple, per-file metadata CRUD API

trishankatdatadog:
**Fixes issue #**:

#1048 

A simple, per-file metadata CRUD API that should make it easy to do the most common operations, without resorting to reading and writing the entire repository all at once.

requirements**:

Cc @joshuagl @sechkova @MVrachev @lukpueh @woodruffw

### DIFF
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -55,12 +55,15 @@ class TestTufApi(unittest.TestCase):
     # files.  'temporary_directory' must be deleted in TearDownClass() so that
     # temporary files are always removed, even when exceptions occur.
     cls.temporary_directory = tempfile.mkdtemp(dir=os.getcwd())
-    test_repo_data = os.path.join('repository_data', 'repository')
+
+    test_repo_data = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), 'repository_data')
+
     cls.repo_dir = os.path.join(cls.temporary_directory, 'repository')
-    shutil.copytree(test_repo_data, cls.repo_dir)
-    test_repo_keys = os.path.join('repository_data', 'keystore')
+    shutil.copytree(os.path.join(test_repo_data, 'repository'), cls.repo_dir)
+
     cls.keystore_dir = os.path.join(cls.temporary_directory, 'keystore')
-    shutil.copytree(test_repo_keys, cls.keystore_dir)
+    shutil.copytree(os.path.join(test_repo_data, 'keystore'), cls.keystore_dir)
 
 
 
@@ -100,18 +103,14 @@ class TestTufApi(unittest.TestCase):
     snapshot_path = os.path.join(self.repo_dir, 'metadata', 'snapshot.json')
     md = metadata.Snapshot.read_from_json(snapshot_path)
 
-    self.assertEqual(md.version, 1)
-    md.bump_version()
-    self.assertEqual(md.version, 2)
-
-    self.assertEqual(md.expiration,
-                     iso8601.parse_date("2030-01-01").replace(tzinfo=None))
-    md.bump_expiration()
-    self.assertEqual(md.expiration,
-                     iso8601.parse_date("2030-01-02").replace(tzinfo=None))
-    md.bump_expiration(timedelta(days=365))
-    self.assertEqual(md.expiration,
-                     iso8601.parse_date("2031-01-02").replace(tzinfo=None))
+    self.assertEqual(md.signed.version, 1)
+    md.signed.bump_version()
+    self.assertEqual(md.signed.version, 2)
+    self.assertEqual(md.signed.expires, '2030-01-01T00:00:00Z')
+    md.signed.bump_expiration()
+    self.assertEqual(md.signed.expires, '2030-01-02T00:00:00Z')
+    md.signed.bump_expiration(timedelta(days=365))
+    self.assertEqual(md.signed.expires, '2031-01-02T00:00:00Z')
 
 
   def test_metadata_snapshot(self):
@@ -119,18 +118,17 @@ class TestTufApi(unittest.TestCase):
     snapshot = metadata.Snapshot.read_from_json(snapshot_path)
 
     key_ring = self._load_key_ring()
-    snapshot.keyring = key_ring
-    snapshot.verify()
+    snapshot.verify(key_ring)
 
     # Create a dict representing what we expect the updated data to be
-    fileinfo = snapshot.signed['meta']
+    fileinfo = snapshot.signed.meta
     hashes = {'sha256': 'c2986576f5fdfd43944e2b19e775453b96748ec4fe2638a6d2f32f1310967095'}
     fileinfo['role1.json']['version'] = 2
     fileinfo['role1.json']['hashes'] = hashes
     fileinfo['role1.json']['length'] = 123
 
-    snapshot.update('role1', 2, 123, hashes)
-    self.assertEqual(snapshot.signed['meta'], fileinfo)
+    snapshot.signed.update('role1', 2, 123, hashes)
+    self.assertEqual(snapshot.signed.meta, fileinfo)
 
     # snapshot.signable()
 
@@ -146,39 +144,34 @@ class TestTufApi(unittest.TestCase):
     timestamp = metadata.Timestamp.read_from_json(timestamp_path)
 
     key_ring = self._load_key_ring()
-    timestamp.keyring = key_ring
-    timestamp.verify()
+    timestamp.verify(key_ring)
 
-    self.assertEqual(timestamp.version, 1)
-    timestamp.bump_version()
-    self.assertEqual(timestamp.version, 2)
+    self.assertEqual(timestamp.signed.version, 1)
+    timestamp.signed.bump_version()
+    self.assertEqual(timestamp.signed.version, 2)
 
-    self.assertEqual(timestamp.expiration,
-                     iso8601.parse_date("2030-01-01").replace(tzinfo=None))
-    timestamp.bump_expiration()
-    self.assertEqual(timestamp.expiration,
-                     iso8601.parse_date("2030-01-02").replace(tzinfo=None))
-    timestamp.bump_expiration(timedelta(days=365))
-    self.assertEqual(timestamp.expiration,
-                     iso8601.parse_date("2031-01-02").replace(tzinfo=None))
+    self.assertEqual(timestamp.signed.expires, '2030-01-01T00:00:00Z')
+    timestamp.signed.bump_expiration()
+    self.assertEqual(timestamp.signed.expires, '2030-01-02T00:00:00Z')
+    timestamp.signed.bump_expiration(timedelta(days=365))
+    self.assertEqual(timestamp.signed.expires, '2031-01-02T00:00:00Z')
 
     # Test whether dateutil.relativedelta works, this provides a much easier to
     # use interface for callers
-    saved_expiration = timestamp.expiration
     delta = relativedelta(days=1)
-    timestamp.bump_expiration(delta)
-    self.assertEqual(timestamp.expires, "2031-01-03T00:00:00Z")
+    timestamp.signed.bump_expiration(delta)
+    self.assertEqual(timestamp.signed.expires, '2031-01-03T00:00:00Z')
     delta = relativedelta(years=5)
-    timestamp.bump_expiration(delta)
-    self.assertEqual(timestamp.expires, "2036-01-03T00:00:00Z")
+    timestamp.signed.bump_expiration(delta)
+    self.assertEqual(timestamp.signed.expires, '2036-01-03T00:00:00Z')
 
     hashes = {'sha256': '0ae9664468150a9aa1e7f11feecb32341658eb84292851367fea2da88e8a58dc'}
-    fileinfo = timestamp.signed['meta']['snapshot.json']
+    fileinfo = timestamp.signed.meta['snapshot.json']
     fileinfo['hashes'] = hashes
     fileinfo['version'] = 2
     fileinfo['length'] = 520
-    timestamp.update(2, 520, hashes)
-    self.assertEqual(timestamp.signed['meta']['snapshot.json'], fileinfo)
+    timestamp.signed.update(2, 520, hashes)
+    self.assertEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
 
     # timestamp.sign()
 
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2014 - 2017, New York University and the TUF contributors
+# Copyright 2020, New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
 """
@@ -36,6 +36,7 @@ import sys
 import errno
 import os
 from datetime import timedelta
+from dateutil.relativedelta import relativedelta
 
 from tuf.api import metadata
 from tuf.api import keys
@@ -160,6 +161,16 @@ class TestTufApi(unittest.TestCase):
     timestamp.bump_expiration(timedelta(days=365))
     self.assertEqual(timestamp.expiration,
                      iso8601.parse_date("2031-01-02").replace(tzinfo=None))
+
+    # Test whether dateutil.relativedelta works, this provides a much easier to
+    # use interface for callers
+    saved_expiration = timestamp.expiration
+    delta = relativedelta(days=1)
+    timestamp.bump_expiration(delta)
+    self.assertEqual(timestamp.expires, "2031-01-03T00:00:00Z")
+    delta = relativedelta(years=5)
+    timestamp.bump_expiration(delta)
+    self.assertEqual(timestamp.expires, "2036-01-03T00:00:00Z")
 
     hashes = {'sha256': '0ae9664468150a9aa1e7f11feecb32341658eb84292851367fea2da88e8a58dc'}
     fileinfo = timestamp.signed['meta']['snapshot.json']
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -266,7 +266,7 @@ class TestTufApi(unittest.TestCase):
       NAME = f'test-{key_type}-key'
 
       for signature_algorithm in {s.value for s in VaultKey.SignatureAlgorithms}:
-        for hash_algorithm in {VaultKey.HashAlgorithms.SHA2_224.value, VaultKey.HashAlgorithms.SHA2_384.value, VaultKey.HashAlgorithms.SHA2_512.value,}:
+        for hash_algorithm in {h.value for h in VaultKey.HashAlgorithms}:
           for marshaling_algorithm in {m.value for m in VaultKey.MarshalingAlgorithms}:
             self.assertRaises(ValueError, VaultKey.create_key, VAULT_ADDR, VAULT_TOKEN, NAME, key_type, marshaling_algorithm=marshaling_algorithm,)
 
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -35,11 +35,11 @@ import shutil
 import sys
 import errno
 import os
+from datetime import timedelta
 
 from tuf.api import metadata
 from tuf.api import keys
 
-from dateutil.relativedelta import relativedelta
 import iso8601
 import six
 
@@ -108,7 +108,7 @@ class TestTufApi(unittest.TestCase):
     md.bump_expiration()
     self.assertEqual(md.expiration,
                      iso8601.parse_date("2030-01-02").replace(tzinfo=None))
-    md.bump_expiration(relativedelta(years=1))
+    md.bump_expiration(timedelta(days=365))
     self.assertEqual(md.expiration,
                      iso8601.parse_date("2031-01-02").replace(tzinfo=None))
 
@@ -157,7 +157,7 @@ class TestTufApi(unittest.TestCase):
     timestamp.bump_expiration()
     self.assertEqual(timestamp.expiration,
                      iso8601.parse_date("2030-01-02").replace(tzinfo=None))
-    timestamp.bump_expiration(relativedelta(years=1))
+    timestamp.bump_expiration(timedelta(days=365))
     self.assertEqual(timestamp.expiration,
                      iso8601.parse_date("2031-01-02").replace(tzinfo=None))
 
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+
+# Copyright 2014 - 2017, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""
+<Program Name>
+  test_tuf_api.py
+
+<Author>
+  Joshua Lock <jlock@vmware.com>
+
+<Started>
+  June 30, 2020.
+
+<Copyright>
+  See LICENSE-MIT OR LICENSE for licensing information.
+
+<Purpose>
+  Unit tests for tuf.api
+"""
+
+# Help with Python 3 compatibility, where the print statement is a function, an
+# implicit relative import is invalid, and the '/' operator performs true
+# division.  Example:  print 'hello world' raises a 'SyntaxError' exception.
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+import unittest
+import logging
+import tempfile
+import shutil
+import sys
+import errno
+import os
+
+from tuf.api import metadata
+
+from dateutil.relativedelta import relativedelta
+import iso8601
+import six
+
+logger = logging.getLogger(__name__)
+
+
+class TestTufApi(unittest.TestCase):
+  @classmethod
+  def setUpClass(cls):
+
+    # Create a temporary directory to store the repository, metadata, and target
+    # files.  'temporary_directory' must be deleted in TearDownClass() so that
+    # temporary files are always removed, even when exceptions occur.
+    cls.temporary_directory = tempfile.mkdtemp(dir=os.getcwd())
+    test_repo_data = os.path.join('repository_data', 'repository')
+    cls.repo_dir = os.path.join(cls.temporary_directory, 'repository')
+    shutil.copytree(test_repo_data, cls.repo_dir)
+
+
+
+  @classmethod
+  def tearDownClass(cls):
+
+    # Remove the temporary repository directory, which should contain all the
+    # metadata, targets, and key files generated for the test cases.
+    shutil.rmtree(cls.temporary_directory)
+
+
+
+  def test_metadata_base(self):
+    # Use of Snapshot is arbitrary, we're just testing the base class features
+    # with real data
+    md = metadata.Snapshot()
+    md.read_from_json(os.path.join(self.repo_dir, 'metadata.staged', 'snapshot.json'))
+
+    self.assertEqual(md.version, 1)
+    md.bump_version()
+    self.assertEqual(md.version, 2)
+
+    self.assertEqual(md.expiration, iso8601.parse_date("2030-01-01T00:00:00Z"))
+    md.bump_expiration()
+    self.assertEqual(md.expiration, iso8601.parse_date("2030-01-02T00:00:00Z"))
+    md.bump_expiration(relativedelta(years=1))
+    self.assertEqual(md.expiration, iso8601.parse_date("2031-01-02T00:00:00Z"))
+
+
+  def test_metadata_snapshot(self):
+    snapshot = metadata.Snapshot()
+    snapshot.read_from_json(os.path.join(self.repo_dir, 'metadata.staged', 'snapshot.json'))
+
+    # Create a dict representing what we expect the updated data to be
+    fileinfo = snapshot.signed['meta']
+    hashes = {'sha256': 'c2986576f5fdfd43944e2b19e775453b96748ec4fe2638a6d2f32f1310967095'}
+    fileinfo['role1.json']['version'] = 2
+    fileinfo['role1.json']['hashes'] = hashes
+    fileinfo['role1.json']['length'] = 123
+
+    snapshot.update('role1', 2, 123, hashes)
+    # snapshot.sign()
+    # self.assertEqual(snapshot.signed['meta'], fileinfo)
+
+    # snapshot.update()
+
+    # snapshot.signable()
+
+    # snapshot.sign()
+
+    # snapshot.verify()
+
+    # snapshot.write_to_json(os.path.join(cls.temporary_directory, 'api_snapshot.json'))
+
+
+  def test_metadata_timestamp(self):
+    timestamp = metadata.Timestamp()
+    timestamp.read_from_json(os.path.join(self.repo_dir, 'metadata.staged', 'timestamp.json'))
+
+    self.assertEqual(timestamp.version, 1)
+    timestamp.bump_version()
+    self.assertEqual(timestamp.version, 2)
+
+    self.assertEqual(timestamp.expiration, iso8601.parse_date("2030-01-01T00:00:00Z"))
+    timestamp.bump_expiration()
+    self.assertEqual(timestamp.expiration, iso8601.parse_date("2030-01-02T00:00:00Z"))
+    timestamp.bump_expiration(relativedelta(years=1))
+    self.assertEqual(timestamp.expiration, iso8601.parse_date("2031-01-02T00:00:00Z"))
+
+    hashes = {'sha256': '0ae9664468150a9aa1e7f11feecb32341658eb84292851367fea2da88e8a58dc'}
+    fileinfo = timestamp.signed['meta']['snapshot.json']
+    fileinfo['hashes'] = hashes
+    fileinfo['version'] = 2
+    fileinfo['length'] = 520
+    timestamp.update('snapshot', 2, 520, hashes)
+    # timestamp.sign()
+    # self.assertEqual(timestamp.signed['meta'], fileinfo)
+
+    # timestamp.write_to_json()
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -37,6 +37,7 @@ import errno
 import os
 
 from tuf.api import metadata
+from tuf.api import keys
 
 from dateutil.relativedelta import relativedelta
 import iso8601
@@ -56,6 +57,9 @@ class TestTufApi(unittest.TestCase):
     test_repo_data = os.path.join('repository_data', 'repository')
     cls.repo_dir = os.path.join(cls.temporary_directory, 'repository')
     shutil.copytree(test_repo_data, cls.repo_dir)
+    test_repo_keys = os.path.join('repository_data', 'keystore')
+    cls.keystore_dir = os.path.join(cls.temporary_directory, 'keystore')
+    shutil.copytree(test_repo_keys, cls.keystore_dir)
 
 
 
@@ -68,11 +72,32 @@ class TestTufApi(unittest.TestCase):
 
 
 
+  def _load_key_ring(self):
+    key_list = []
+    root_key = keys.read_key(os.path.join(self.keystore_dir, 'root_key'),
+                             'RSA', 'password')
+    key_list.append(root_key)
+
+    for key_file in os.listdir(self.keystore_dir):
+      if key_file.endswith('.pub'):
+        # ignore public keys
+        continue
+
+      if key_file.startswith('root_key'):
+        # root key is loaded
+        continue
+
+      key = keys.read_key(os.path.join(self.keystore_dir, key_file), 'ED25519',
+                          'password')
+      key_list.append(key)
+    threshold = keys.Threshold(1, 1)
+    return keys.KeyRing(threshold=threshold, keys=key_list)
+
   def test_metadata_base(self):
     # Use of Snapshot is arbitrary, we're just testing the base class features
     # with real data
-    md = metadata.Snapshot()
-    md.read_from_json(os.path.join(self.repo_dir, 'metadata.staged', 'snapshot.json'))
+    snapshot_path = os.path.join(self.repo_dir, 'metadata', 'snapshot.json')
+    md = metadata.Snapshot.read_from_json(snapshot_path)
 
     self.assertEqual(md.version, 1)
     md.bump_version()
@@ -86,8 +111,12 @@ class TestTufApi(unittest.TestCase):
 
 
   def test_metadata_snapshot(self):
-    snapshot = metadata.Snapshot()
-    snapshot.read_from_json(os.path.join(self.repo_dir, 'metadata.staged', 'snapshot.json'))
+    snapshot_path = os.path.join(self.repo_dir, 'metadata', 'snapshot.json')
+    snapshot = metadata.Snapshot.read_from_json(snapshot_path)
+
+    key_ring = self._load_key_ring()
+    snapshot.keyring = key_ring
+    snapshot.verify()
 
     # Create a dict representing what we expect the updated data to be
     fileinfo = snapshot.signed['meta']
@@ -97,10 +126,7 @@ class TestTufApi(unittest.TestCase):
     fileinfo['role1.json']['length'] = 123
 
     snapshot.update('role1', 2, 123, hashes)
-    # snapshot.sign()
-    # self.assertEqual(snapshot.signed['meta'], fileinfo)
-
-    # snapshot.update()
+    self.assertEqual(snapshot.signed['meta'], fileinfo)
 
     # snapshot.signable()
 
@@ -112,8 +138,12 @@ class TestTufApi(unittest.TestCase):
 
 
   def test_metadata_timestamp(self):
-    timestamp = metadata.Timestamp()
-    timestamp.read_from_json(os.path.join(self.repo_dir, 'metadata.staged', 'timestamp.json'))
+    timestamp_path = os.path.join(self.repo_dir, 'metadata', 'timestamp.json')
+    timestamp = metadata.Timestamp.read_from_json(timestamp_path)
+
+    key_ring = self._load_key_ring()
+    timestamp.keyring = key_ring
+    timestamp.verify()
 
     self.assertEqual(timestamp.version, 1)
     timestamp.bump_version()
@@ -130,8 +160,9 @@ class TestTufApi(unittest.TestCase):
     fileinfo['hashes'] = hashes
     fileinfo['version'] = 2
     fileinfo['length'] = 520
-    timestamp.update('snapshot', 2, 520, hashes)
+    timestamp.update(2, 520, hashes)
+    self.assertEqual(timestamp.signed['meta']['snapshot.json'], fileinfo)
+
     # timestamp.sign()
-    # self.assertEqual(timestamp.signed['meta'], fileinfo)
 
     # timestamp.write_to_json()
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -103,11 +103,14 @@ class TestTufApi(unittest.TestCase):
     md.bump_version()
     self.assertEqual(md.version, 2)
 
-    self.assertEqual(md.expiration, iso8601.parse_date("2030-01-01T00:00:00Z"))
+    self.assertEqual(md.expiration,
+                     iso8601.parse_date("2030-01-01").replace(tzinfo=None))
     md.bump_expiration()
-    self.assertEqual(md.expiration, iso8601.parse_date("2030-01-02T00:00:00Z"))
+    self.assertEqual(md.expiration,
+                     iso8601.parse_date("2030-01-02").replace(tzinfo=None))
     md.bump_expiration(relativedelta(years=1))
-    self.assertEqual(md.expiration, iso8601.parse_date("2031-01-02T00:00:00Z"))
+    self.assertEqual(md.expiration,
+                     iso8601.parse_date("2031-01-02").replace(tzinfo=None))
 
 
   def test_metadata_snapshot(self):
@@ -149,11 +152,14 @@ class TestTufApi(unittest.TestCase):
     timestamp.bump_version()
     self.assertEqual(timestamp.version, 2)
 
-    self.assertEqual(timestamp.expiration, iso8601.parse_date("2030-01-01T00:00:00Z"))
+    self.assertEqual(timestamp.expiration,
+                     iso8601.parse_date("2030-01-01").replace(tzinfo=None))
     timestamp.bump_expiration()
-    self.assertEqual(timestamp.expiration, iso8601.parse_date("2030-01-02T00:00:00Z"))
+    self.assertEqual(timestamp.expiration,
+                     iso8601.parse_date("2030-01-02").replace(tzinfo=None))
     timestamp.bump_expiration(relativedelta(years=1))
-    self.assertEqual(timestamp.expiration, iso8601.parse_date("2031-01-02T00:00:00Z"))
+    self.assertEqual(timestamp.expiration,
+                     iso8601.parse_date("2031-01-02").replace(tzinfo=None))
 
     hashes = {'sha256': '0ae9664468150a9aa1e7f11feecb32341658eb84292851367fea2da88e8a58dc'}
     fileinfo = timestamp.signed['meta']['snapshot.json']
--- a/tests/test_tuf_api.py
+++ b/tests/test_tuf_api.py
@@ -40,6 +40,7 @@ from dateutil.relativedelta import relativedelta
 
 from tuf.api import metadata
 from tuf.api import keys
+from tuf.api.keys import VaultKey
 
 import iso8601
 import six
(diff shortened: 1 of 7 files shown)


trishankatdatadog:
There is so much incomplete with this draft PR, but I hope it gives the right idea. Please let me know if you vehemently disagree with the approach here!

Cc @joshuagl @sechkova and @MVrachev who have agreed to help write and polish the code

tuf/api/keys.py
```diff
@@ -0,0 +1,169 @@
...
+# Utility functions.
+
+def get_new_private_keypath(keystore_dir: str, rolename: str, i : int = 1) -> str:
+    return os.path.join(keystore_dir, f'{rolename}_ed25519_key_{i}')
+
+def get_public_keypath(private_keypath: str) -> str:
+    # this is the tuf filename convention at the time of writing.
+    return f'{private_keypath}.pub'
+
+def get_private_keys_from_keyring(keyring: Keyring) -> KeyDict:
```
joshuagl: This might be cleaner as a class method on `Keyring`, i.e. `get_private_keys()`?

tuf/api/keys.py
```diff
@@ -0,0 +1,169 @@
...
+
+    for keypair in keyring.keypairs:
+        privkey = keypair.private.obj
+        keyid = privkey['keyid']
+        assert keyid not in privkeys
+        privkeys[keyid] = privkey
+
+    return privkeys
+
+def get_public_keys_from_keyring(keyring: Keyring) -> KeyDict:
```
joshuagl: Similarly here, would it make more sense as a class method on `Keyring` – `get_public_keys()`?

tuf/api/metadata.py
```diff
@@ -0,0 +1,148 @@
...
+    # And you would use this method to populate it from a file.
+    def read_from_json(self, filename: str) -> None:
+        signable = load_json_file(filename)
+
+        # TODO: use some basic schema checks
+        signatures = signable['signatures']
+        signed = signable['signed']
+
+        self.expiration = datetime.strptime(signed['expiration'], '%b %d %Y %I:%M%p')
+        self.version = signed['version']
```
joshuagl: TODO: should be able to set appropriate values for `self.consistent_snapshot` at load, also.

joshuagl (commented):
Thanks for your initial work on this @trishankatdatadog

Observation: `typing` support requires Python 3.5+ or the [typing package](https://pypi.org/project/typing/) and f-string support requires Python 3.6. 
This needn't block the initial implementation, but assuming we continue to flesh out this low-level API and use it in the refactor it might be exposed to the client, where we want to continue to support Python 2.7 and thus we can't use f-strings and may not want to add an additional dependency on typing.

So far as I can see from a preliminary review the to-do list is:
* keys
  * ~Support non ed25519 keys. Perhaps by making key type a part of: `get_private_keypath()`, `write_keypair()`, `read_keypair()`, etc?~ https://github.com/theupdateframework/tuf/pull/1060/commits/f2861bfd417df30f814cfa25128fdf5e5c2f05f8
* metadata
  * Figure out how to handle consistent snapshots
  * ~Figure out how to handle writing consistent targets~ not relevant at this low-level https://github.com/theupdateframework/tuf/pull/1060#discussion_r448908497
  * ~Add schema checks to `read_from_json()`~ https://github.com/theupdateframework/tuf/pull/1060/commits/1fbff557adec35de1fc80894cca1b2ea982c1960
  * ~Implement `update_signatures()`~ https://github.com/theupdateframework/tuf/pull/1060/commits/5ef60ca1871953d3982458acda736815a4c9bc86
  * Implement `signable()` for each of:
    * ~`Timestamp`~ https://github.com/theupdateframework/tuf/pull/1060/commits/11d76e72ddbd1dafab70c35213db30396dcb7e44
    * ~`Snapshot`~ https://github.com/theupdateframework/tuf/pull/1060/commits/92f97a45e3c8934b2872075fe53e606f459aad22
    * ~`Targets`~ https://github.com/theupdateframework/tuf/pull/1060/commits/916055aa54452283831db0fc0d351e368c8b4ca0
  * Implement `update()` for each of 
    * ~`Timestamp`~ https://github.com/theupdateframework/tuf/pull/1060/commits/11d76e72ddbd1dafab70c35213db30396dcb7e44
    * ~`Snapshot`~ https://github.com/theupdateframework/tuf/pull/1060/commits/92f97a45e3c8934b2872075fe53e606f459aad22
    * ~`Targets`~ https://github.com/theupdateframework/tuf/pull/1060/commits/db0f8a73a70e2dd7103769977a9f939a500f1f94
  * ~Add method to `Metadata` to bump expiration by a delta, `Metadata.bump_expiration(delta='1d')`~
    * ~It would be easy for the implementation if this took a [timedelta](https://docs.python.org/3/library/datetime.html#datetime.timedelta) object, i.e. `Metadata.bump_expiration(delta=timedelta(days=1))` – I'd appreciate some thoughts on how useful/usable such an API is?~ https://github.com/theupdateframework/tuf/pull/1060/commits/721def4eb06acf6e7c2accf2d7201541f0819d71
  * ~Add method to `Metadata` to bump version number; `Metadata.bump_version()`~ https://github.com/theupdateframework/tuf/pull/1060/commits/721def4eb06acf6e7c2accf2d7201541f0819d71
* ~Use `StorageBackendInterface`~ https://github.com/theupdateframework/tuf/pull/1060/commits/0ca471ed2ad151c61647400e3ac07670dc621b9e
* ~We should raise sslib.FormatError or ValueError instead of AssertionErrors (for input validation)~ https://github.com/theupdateframework/tuf/pull/1060/commits/76cb560a4601dc264390d10143a68cc05d16787f
* Add docstrings
* Add unit tests – WIP here https://github.com/theupdateframework/tuf/pull/1060/commits/57c98d45ac570b6947420b9f535a3765c610c0c5 and here https://github.com/theupdateframework/tuf/pull/1060/commits/34a0680947ab5e0afc85572e438b40bcb98cd98a
* ~Streamline `metadata` interaction model so that `signatures` and `signed` are only ever accessed as a property, like `signable`~
* ~make as many thing as possible private variables of the objects~ https://github.com/theupdateframework/tuf/pull/1060/commits/76cb560a4601dc264390d10143a68cc05d16787f

trishankatdatadog:
Hi Joshua,

> Observation: `typing` support requires Python 3.5+ or the [typing package](https://pypi.org/project/typing/) and f-string support requires Python 3.6.
> This needn't block the initial implementation, but assuming we continue to flesh out this low-level API and use it in the refactor it might be exposed to the client, where we want to continue to support Python 2.7 and thus we can't use f-strings and may not want to add an additional dependency on typing.

Agreed. I left the typing in there as to help communicate intent between different developers 🙂 So you can think of it as scaffolding we can remove before hitting production...

> Add bump_role() API somewhere?

Aha, it's not clear, sorry, but it's supposed to be a part of [`Metadata.sign()`](https://github.com/theupdateframework/tuf/pull/1060/files#diff-d2aa622cc58ad47011c52d0adc99323dR73). Let me know whether this is a good idea or not...

How should we divide-and-conquer going forward?

tuf/api/keys.py
```diff
@@ -0,0 +1,169 @@
...
+    get_password,
+    import_ed25519_privatekey_from_file,
+    import_ed25519_publickey_from_file,
+)
+
+# Utility classes.
+
+class Threshold:
+
+    def __init__(self, m: int = 1, n: int = 1):
```
woodruffw: Tiny nit, but can we use more descriptive parameter names + attributes here? I know it's an m-of-n scheme but `minimum` and `keyholders` (or similar) would be a little easier to read 🙂

tuf/api/keys.py
```diff
@@ -0,0 +1,169 @@
...
+        assert m > 0, f'{m} <= 0'
+        assert n > 0, f'{n} <= 0'
+        assert m <= n, f'{m} > {n}'
+        self.m = m
+        self.n = n
+
+class Keypath:
```
lukpueh: Nit: `KeyPath`

tuf/api/keys.py
```diff
@@ -0,0 +1,169 @@
...
+        self.private = private
+        self.public = public
+
+class Key:
+
+    def __init__(self, path: str, obj: Any):
+        self.path = path
+        self.obj = obj
+
+class Keypair:
```
lukpueh: Nit: `KeyPair`

tuf/api/keys.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __init__(self, private: Key, public: Key):
+        self.private = private
+        self.public = public
+
+Keypairs = List[Keypair]
+
+class Keyring:
```
lukpueh: Nit: `KeyRing`

tuf/api/metadata.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def signed(self) -> str:
+        return encode_canonical(self.signable['signed']).encode('utf-8')
+
+    def signatures(self) -> List:
+        return self.signable['signatures']
+
+    # TODO: We need to update the expiration timestamp using self.expiration.
+    # Oh, and bump the version number.
+    # And, oh, take care of consistent snapshot of metadata.
```
lukpueh: I think none of this should be done in the sign function. Let's have a separate API function (or set of functions) for this.

tuf/api/keys.py
```diff
@@ -0,0 +1,169 @@
...
+    private_key = Key(private_keypath, private_key_obj)
+
+    # and its corresponding public key.
+    public_keypath = keypath.public
+    public_key_obj = import_ed25519_publickey_from_file(keypath.public)
+    public_key = Key(public_keypath, public_key_obj)
+
+    return Keypair(private_key, public_key)
+
+def rename_keys_to_match_keyid(keystore_dir: str, keypair: Keypair) -> None:
```
lukpueh: This seems like a crutch that we only need because we don't write the keys to the right location in the first place. Maybe we can rather fix this at the source of that problem (in sslib) instead of working around here?

lukpueh (commented):
Great work, @trishankatdatadog! I especially like the scaffolding in `metadata.py`.

A left a few inline comments and here are two more high-level observations:
- We should raise sslib.FormatError or ValueError instead of AssertionErrors (for input validation)
- When we add docstrings, let's please use a standard format
(see last item of "Issues with the current guideline" in https://github.com/secure-systems-lab/code-style-guidelines/issues/20 and https://github.com/in-toto/in-toto/pull/369)

joshuagl:
> > Observation: `typing` support requires Python 3.5+ or the [typing package](https://pypi.org/project/typing/) and f-string support requires Python 3.6.
> > This needn't block the initial implementation, but assuming we continue to flesh out this low-level API and use it in the refactor it might be exposed to the client, where we want to continue to support Python 2.7 and thus we can't use f-strings and may not want to add an additional dependency on typing.
> 
> Agreed. I left the typing in there as to help communicate intent between different developers 🙂 So you can think of it as scaffolding we can remove before hitting production...

I assumed it was something like that, but thought it worth explicitly noting.

> > Add bump_role() API somewhere?
> 
> Aha, it's not clear, sorry, but it's supposed to be a part of [`Metadata.sign()`](https://github.com/theupdateframework/tuf/pull/1060/files#diff-d2aa622cc58ad47011c52d0adc99323dR73). Let me know whether this is a good idea or not...

Ah, understood! We'll need a way to set/pass a delta by which to increase the expiration, which doesn't feel like it fits naturally into the method signature of `sign()` and is even less clean in the method signature of `write_to_json()`, from which `sign()` is called.

I thought a delta by which to bump the `expiration` could become a property of the `Metadata` object (i.e. PEP 458 [recommends specific expiry deltas per role](https://www.python.org/dev/peps/pep-0458/#metadata-expiry-times) and this would map nicely) however the delta isn't a part of the metadata itself and therefore couldn't be loaded into the object by reading the on-disk representation... therefore – this wouldn't work.

As this is defined as a low-level (or granular?) API, perhaps we should just be explicit and have functions for incrementing version number and increasing expiration. We can build convenience API around these in future as we flesh this out?

tuf/api/metadata.py
joshuagl: @lukpueh beat me to it. I wrote up my though process around this a bit in a comment below.

joshuagl:
> How should we divide-and-conquer going forward?

The to-do list I [sketched above](https://github.com/theupdateframework/tuf/pull/1060#pullrequestreview-439026780) has many of the items broken down by module, perhaps you could tackle `keys` while we work on `metadata`?

Then we can hammer out typing, docstrings, tests, coding style and assertions as a follow-on task?

joshuagl:
I've got some WIP changes https://github.com/joshuagl/tuf/commits/joshuagl/simple-update-api%2BMVrachev-length-hashes-optional and have updated the [to do list](https://github.com/theupdateframework/tuf/pull/1060#pullrequestreview-439026780) to indicate what I have done. 

Note: I'm using a helper from #1031 so my branch includes a merge of that PR

joshuagl:
all of the code I've written so far assumes that the sequence of events is:
```Python
metadata = metadata.Metadata()
metadata.read_from_json('/path/to/metadata.json')
```
and therefore `self.signatures` and `self.signed` exist and are meaningful. That means we can't currently instantiate an object, populate it with code and write it to disk. We can only load and modify. That seems undesirable?

Should we smooth out the interaction model so that `signatures` and `signed` are only ever accessed as a property, like signable? 


Additionally, should we make `read_from_json()` a class method that can be used to create an instance of an object by loading a file?
```Python
snapshot = Snapshot.read_from_json('/path/to/snapshot.json')
```

trishankatdatadog:
> Should we smooth out the interaction model so that `signatures` and `signed` are only ever accessed as a property, like signable?

Yes, we should. Also, `signatures` and `signed` should be objects under `signable`, no?
 
> Additionally, should we make `read_from_json()` a class method that can be used to create an instance of an object by loading a file?

💯

tuf/api/metadata.py
```diff
@@ -0,0 +1,183 @@
...
+class Targets(Metadata):
+    def __init__(self, consistent_snapshot: bool = True, expiration: relativedelta = relativedelta(days=1), keyring: KeyRing = None, version: int = 1):
+        super().__init__(consistent_snapshot, expiration, relativedelta, keyring, version)
+
+    # FIXME
+    def signable(self):
+        return generate_targets_metadata()
+
+    # Add or update metadata about the target.
+    # TODO: how to handle writing consistent targets?
```
sechkova: From the initial API design I assume that it covers only the case of using existing fileinfo. If this is true then writing consistent targets is not applicable here:
https://github.com/theupdateframework/tuf/blob/de8649e12a850fccbc745f0129d43956ef7da33b/tuf/repository_lib.py#L155

If we want to add the option to NOT use existing fileinfo and using tuf for calculating hashes, we need to extend the api (maybe on a later pass) with targets paths and storage_backend.

joshuagl:
> > Should we smooth out the interaction model so that `signatures` and `signed` are only ever accessed as a property, like signable?
> 
> Yes, we should. Also, `signatures` and `signed` should be objects under `signable`, no?

I haven't implemented this suggestion yet as I was worried about the complexity increase of nested classes. I'd be curious to hear/read what everyone thinks of the current implementation.

trishankatdatadog:
> I haven't implemented this suggestion yet as I was worried about the complexity increase of nested classes. I'd be curious to hear/read what everyone thinks of the current implementation.

This is simply looking fabulous, thanks very much, Joshua! 💯 

I have a minor quibble about how we should make as many thing as possible private variables of the objects, but I started that, so it's on me, and it's also easy to fix 🙂

joshuagl:
> > I haven't implemented this suggestion yet as I was worried about the complexity increase of nested classes. I'd be curious to hear/read what everyone thinks of the current implementation.
> 
> This is simply looking fabulous, thanks very much, Joshua! 💯

You're very kind, thanks!

> I have a minor quibble about how we should make as many thing as possible private variables of the objects, but I started that, so it's on me, and it's also easy to fix 🙂

I completely agree. I've added it to the list of things to address in https://github.com/theupdateframework/tuf/pull/1060#pullrequestreview-439026780

tests/test_tuf_api.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env python
+
+# Copyright 2014 - 2017, New York University and the TUF contributors
```
joshuagl: Clearly the wrong date range. 🤦

tuf/api/metadata.py
```diff
@@ -0,0 +1,236 @@
...
+    def signed(self) -> str:
+        raise NotImplementedError
+
+    @property
+    def signatures(self) -> List[JsonDict]:
+        return self._signatures
+
+    @property
+    def expires(self) -> str:
+        return self.expiration.replace(tzinfo=None).isoformat()+'Z'
```
joshuagl: In https://github.com/theupdateframework/tuf/pull/1012#discussion_r403039950 @woodruffw wonders whether `unix_timestamp_to_datetime` should just include the Z suffix by default if we always intend times to be GMT.

trishankatdatadog:
@joshuagl I think we're largely done, no? Should we get @woodruffw to start alpha-testing? 🧪

joshuagl:
> @joshuagl I think we're largely done, no? Should we get @woodruffw to start alpha-testing? 🧪

I'd certainly appreciate some user feedback on this API from @woodruffw !

The to do list before we can remove the draft flag from this PR is:
* Figure out if/how to handle consistent snapshots
  * maybe this would be at the `write_to_json()` level? If `consistent_snapshot` is true we could massage the passed filename into the expected format for a consistent snapshot? Is that too magical? Should consistent snapshot handling be done at a higher level?
* raise sslib.FormatError or ValueError instead of AssertionErrors (for input validation)
* Add docstrings
* Finish unit tests, ensure good (90%+?) coverage

joshuagl:
The interface for metadata currently has two very similar properties:
* `Metadata.expiration` can be used to set/get the `expiration` `datetime` object
* `Metadata.expires` can be used to get the string representation of the `expiration` `datetime` object that we will write to the metadata

The names are a bit too similar for my taste, some thoughts:
* should we make the `expires` property more explicitly named: `expiration_string`, or similar?
* should we make `expires` an internal helper – will users of the public API ever need the string representation?
* if so, should we add a setter for `expires` so that users can set the `expiration` `datetime` from an iso8601 string representation?

trishankatdatadog:
> * maybe this would be at the `write_to_json()` level? If `consistent_snapshot` is true we could massage the passed filename into the expected format for a consistent snapshot? Is that too magical? Should consistent snapshot handling be done at a higher level?

I think maybe we should pass a directory instead of filename to `write_to_json()` and if the `consistent_snapshot` flag is set, we can write an additional file in there. WDYT?

We also need a similar interface for writing consistent targets...

trishankatdatadog:
> * if so, should we add a setter for `expires` so that users can set the `expiration` `datetime` from an iso8601 string representation?

I like this option the best, but we should provide a more friendly interface that takes a `relativedelta` as an input instead of a `datetime`

joshuagl:
> > * maybe this would be at the `write_to_json()` level? If `consistent_snapshot` is true we could massage the passed filename into the expected format for a consistent snapshot? Is that too magical? Should consistent snapshot handling be done at a higher level?
> 
> I think maybe we should pass a directory instead of filename to `write_to_json()` and if the `consistent_snapshot` flag is set, we can write an additional file in there. WDYT?

That seems like a sound proposal to me.

> We also need a similar interface for writing consistent targets...

Should we be handing consistent targets at this low-level API? At least in Warehouse we will be using existing fileinfo and not interacting with the targets directly, should we assume that a higher-level interface does the right thing with consistent targets?

joshuagl:
> > * if so, should we add a setter for `expires` so that users can set the `expiration` `datetime` from an iso8601 string representation?
> 
> I like this option the best, but we should provide a more friendly interface that takes a `relativedelta` as an input instead of a `datetime`

Just to be clear, are you advocating for public properties for setting/getting expiration as a `datetime` and an iso8602 string?

I made some assumptions that this low-level API might eventually end up as part of the client code and so explicitly opted to drop `dateutil` in favour of the standard library (https://github.com/theupdateframework/tuf/pull/1060/commits/54e1f9c03b09bdf05fd0a3cd3e4db35c6480c5a7) to reduce the number of dependencies that i.e. Pip will have to vendor (see #1065).
However, I did add a test and `dateutil.relativedelta` seems to behave as expected, see https://github.com/theupdateframework/tuf/pull/1060/commits/42372872d2a7f83bc4abad34d3f52a1d606fd786.

lukpueh:
> - raise sslib.FormatError or ValueError instead of AssertionErrors (for input validation)

I lean towards consistently using [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) and [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) for bad arguments, since this is what they seem to be meant for.

The only advantage I can see of using a custom sslib's `FormatError` is that our own or user code can easily distinguish our Value/TypeErrors from those of other code. But I don't think that is necessary. Ideally, we validate arguments at the user boundary, so the user should know that we raised the errors, and all errors from further down the stack are our responsibility anyway, even if we didn't raise them. What do others think?

On a very related side-note, if we move away from our custom `FormatError` we also need to re-consider our usage of `check_match`, which anyways would be a good idea for the various reasons outlined in https://github.com/secure-systems-lab/securesystemslib/issues/183. :)

lukpueh:
> - Add docstrings

I volunteer.

lukpueh:
Regarding expiration:

IMO the class should have an `expires` attribute to align with the json metadata representation. The data type of that attribute could then either be a string in the required iso8602 format to fully align with the json metadata representation,  or a datetime object for easier further use. At any rate, we should have additional convenience methods with mnemonic names to set/get the date in the desired format.

FWIW in in-toto I keep the expires attribute as iso8602 string and provide a [`set_relative_expiration`](https://github.com/in-toto/in-toto/blob/6cfa36cfc2c4b864936560294c032d3bfa869064/in_toto/models/layout.py#L150) convenience method.

lukpueh:
> > Should consistent snapshot handling be done at a higher level?
> 
> I think maybe we should pass a directory instead of filename to `write_to_json()` and if the `consistent_snapshot` flag is set, we can write an additional file in there. WDYT?

I like both your ideas. Can we keep `write_to_json` as is, and add a wrapper that does what Trishank says? E.g....
```python
def write(self, metadata_dir, storage_backend=None):
    self.write_to_json(
        os.path.join(metadata_dir, self._name + ".json")

    if self._consistent_snapshot: # ... or isinstance(Root) # not yet relevant
        self.write_to_json(
                os.path.join(metadata_dir, self._version + "." + self._name  + ".json")
```

tuf/api/metadata.py
```diff
@@ -0,0 +1,250 @@
...
+# Types.
+
+JsonDict = Dict[str, Any]
+
+# Classes.
+
+class Metadata(ABC):
+    # By default, a Metadata would be a rather empty one.
+    def __init__(self, consistent_snapshot: bool = True, expiration: datetime = datetime.today(), keyring: Optional[KeyRing] = None, version: int = 1) -> None:
+        self.__consistent_snapshot = consistent_snapshot
```
lukpueh: I am not sure if consistent_snapshot should be an attribute of the base metadata class. In TUF the consistent_snapshot attribute is only available on the root role metadata.

I'd really prefer if we only kept things that belong to the corresponding TUF role on a given metadata class. 

What do others think about an additional management class? Something like `Repository` but without implying that a full repository is required. We can also punt on the decision, where this is info comes from and just require the user of this interface to pass `consistent_snapshot` to the method that needs it (e.g. for writing metadata to disk).

tuf/api/metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+        self.__keyring = keyring
```
lukpueh: I think we should avoid storing verification keys for a role on that role. See #660.

tuf/api/keys.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @abstractmethod
+    def verify(self, signed: str, signature: str) -> bool:
+        raise NotImplementedError
+
+Keys = List[Key]
+
+class KeyRing:
+
+    def __init__(self, threshold: Threshold, keys: Keys):
```
lukpueh, line 94: Which threshold is this?

tuf/api/metadata.py
```diff
@@ -0,0 +1,250 @@
...
+        updated = False
+        keyid_signature = {'keyid':keyid, 'sig':signature}
+        for idx, keyid_sig in enumerate(signatures):
+            if keyid_sig['keyid'] == keyid:
+                signatures[idx] = keyid_signature
+                updated = True
+        if not updated:
+            signatures.append(keyid_signature)
+
+    def sign(self) -> JsonDict:
```
lukpueh: I'd rather pass the signing key(s) to this function instead of having them stored on the object. (see #660)

tuf/api/metadata.py
```diff
@@ -0,0 +1,250 @@
...
+        signatures = self.__signatures
+
+        for key in self.__keyring.keys:
+            signature = key.sign(signed_bytes)
+            self.__update_signature(signatures, key.keyid, signature)
+
+        self.__signatures = signatures
+        return self.signable
+
+    def verify(self) -> bool:
```
lukpueh: I'd rather pass the verification key(s) to this function instead of having them stored on the object. (see #660)

tuf/api/metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+class Snapshot(Metadata):
+    def __init__(self, consistent_snapshot: bool = True, expiration: datetime = datetime.today(), keyring: KeyRing = None, version: int = 1) -> None:
+        super().__init__(consistent_snapshot, expiration, keyring, version)
+        self.targets_fileinfo = {}
+
+    @classmethod
+    def read_from_json(cls, filename: str) -> Metadata:
+        md = Metadata.read_from_json(filename)
+        snapshot = cls(md.consistent_snapshot, md.expiration, md.keyring, md.version)
```
lukpueh: Maybe a stupid question, but is it a common pattern to first create the parent class object and then the child from it?

tuf/api/metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+class Targets(Metadata):
+    def __init__(self, consistent_snapshot: bool = True, expiration: datetime = datetime.today(), keyring: KeyRing = None, version: int = 1) -> None:
+        super().__init__(consistent_snapshot, expiration, keyring, version)
+        self.targets = {}
+        self.delegations = {}
+
+    @classmethod
+    def read_from_json(cls, filename: str) -> Metadata:
+        targets = Metadata.read_from_json(filename)
```
lukpueh: Does this work? I think `targets` is not of type `Targets` here, but of type `Metadata`.

tuf/api/metadata.py
```diff
@@ -0,0 +1,250 @@
...
+                "signed": self.signed}
+
+    @property
+    def signed_bytes(self) -> bytes:
+        return encode_canonical(self.signed).encode('UTF-8')
+
+    @property
+    @abstractmethod
+    def signed(self) -> JsonDict:
+        raise NotImplementedError
```
lukpueh: Do we really need abc/abstractmethod/NotImplementedError, etc... in our Metadata base class? IIUC the list of possible subclasses is: root, snapshot, timestamp and targets. I don't think we need to plan for users implementing arbitrary subclasses.

tuf/api/metadata.py
trishankatdatadog: > We can also punt on the decision, where this is info comes from and just require the user of this interface to pass consistent_snapshot to the method that needs it (e.g. for writing metadata to disk).

Good question. Personally, I agree with this right now, until we figure out something better long-term.

tuf/api/metadata.py
trishankatdatadog: I think I see what you mean, especially in the context of delegations. But how else should we conveniently/naturally sign/verify metadata? It is assumed here that the user can figure out the right keys to use. Is this a bad assumption?

tuf/api/keys.py
trishankatdatadog: Oh, I think I see what you mean. Again, context-dependent, and up to the user to figure out.

tuf/api/metadata.py
trishankatdatadog: Sure, I can we can say `Targets` instead of `Metadata` here w/o breaking `typing`

tuf/api/metadata.py
trishankatdatadog: See my comment above

tuf/api/metadata.py
trishankatdatadog: See my comment above

tuf/api/metadata.py
trishankatdatadog: I guess we could make `__init__()` read a file, if provided, and use default versions when none is provided?

tuf/api/metadata.py
trishankatdatadog: It was just a way to denote that this method must be overridden

tuf/api/metadata.py
trishankatdatadog: Oh, I think I see what you're really saying now. Yeah, sure, we can pass it in per sign/verify call, I guess, less convenient, but maybe philosophically cleaner

lukpueh:
Thanks for getting back to me about my review comments, @trishankatdatadog, and generally thanks a lot for the hard work,  everyone!

trishankatdatadog:
> Thanks for getting back to me about my review comments, @trishankatdatadog, and generally thanks a lot for the hard work, everyone!

Thanks, and where should we go from here? 🙂 Specifically, what can I help with?

tuf/api/metadata.py
```diff
@@ -154,96 +122,166 @@ def verify(self) -> bool:
...
-            targets=self.targets,
-            delegations=self.delegations)
+    def update(
+            self, rolename: str, version: int, length: Optional[int] = None,
+            hashes: Optional[JsonDict] = None) -> None:
+        metadata_fn = f'{rolename}.json'
+
+        self.meta[metadata_fn] = {'version': version}
+        if length is not None:
+            self.meta[metadata_fn]['length'] = length
```
joshuagl, line 263: if it is `None`, do we need to delete an existing value?

tuf/api/metadata.py
```diff
@@ -154,96 +122,166 @@ def verify(self) -> bool:
...
+
+        if hashes is not None:
+            self.meta[metadata_fn]['hashes'] = hashes
```
joshuagl, line 266: Same, what if hashes is `None` (but wasn't before)?

tuf/api/metadata.py
```diff
@@ -154,96 +122,166 @@ def verify(self) -> bool:
...
+
+        # We always intend times to be UTC
+        # NOTE: we could do this with datetime.fromisoformat() but that is not
+        # available in Python 2.7's datetime
+        # NOTE: Store as datetime object for convenient handling, use 'expires'
+        # property to get the TUF metadata format representation
+        self.__expiration = iso8601.parse_date(expires).replace(tzinfo=None)
+
+        if version < 0:
+            raise ValueError(f'version must be < 0, got {version}')
```
joshuagl, line 166: this error message doesn't match the conditional

tuf/api/metadata.py
```diff
@@ -126,25 +92,27 @@ def __update_signature(self, signatures, keyid, signature):
         if not updated:
             signatures.append(keyid_signature)
 
-    def sign(self) -> JsonDict:
+    def sign(self, key_ring: KeyRing) -> JsonDict:
+        # FIXME: Needs documentation of expected behavior
         signed_bytes = self.signed_bytes
```
joshuagl, line 98: unused variable?

tuf/api/metadata.py
```diff
@@ -154,96 +122,166 @@ def verify(self) -> bool:
...
+
                         break
 
-        return len(verified_keyids) >= self.__keyring.threshold.least
+        return len(verified_keyids) >= key_ring.threshold.least
 
-    def write_to_json(self, filename: str, storage_backend: StorageBackendInterface = None) -> None:
+    def write_to_json(
+            self, filename: str,
+            storage_backend: StorageBackendInterface = None) -> None:
```
joshuagl, line 134:
```suggestion
            storage_backend: Optional[StorageBackendInterface] = None) -> None:
```

tuf/api/metadata.py
lukpueh: Good catch. Let's keep the note around and I'll fix it together with other validation updates. What do you think about the related ideas I floated in the doc header todo list?

tuf/api/metadata.py
lukpueh: Good question. Is this something you and @MVrachev discussed on #1031? What do you think should happen? I understood passing `length=None` as, "I don't want length on this".  But it could also mean, "I don't want to re-assign length, use the old one."

tuf/api/metadata.py
lukpueh: Oh, @trishankatdatadog passed `signed_bytes` to `key.sign`. I didn't understand why (same goes for `signatures = self.__signatures`) and started changing it, but then though it might be better to ask him why he did it that way. :) @trishankatdatadog, could you help me understand the `sign` and `verify` functions?

lukpueh:
> > * Add docstrings
> 
> I volunteer.

So I haven't yet had a chance to write the docstrings... 😬 

But following some discussions with @joshuagl I shuffled the metadata model class architecture quite a bit.  My intention was to make it even simpler and above all to make the classes match the metadata format specification more closely. An overview of the changes + rationale can be found in the commit message of 3e249f5bdd9e600475a05eab3587c5388eeedc3a.

tuf/api/metadata.py
joshuagl: We did not. It seemed more apparent here because my imagined workflows for the API are different. This API feels more focused on loading and updating existing metadata, such that we'll often start from a populated class structure and update it. Whereas repository_tool feels like it's more about generating new metadata.

I think passing `length=None` should mean "I don't want a length on this", but we should think about the (unlikely?) case that i.e. a timestamp with a length is loaded, and the next version doesn't want a length.

tuf/api/metadata.py
joshuagl: I'll comment on the ideas you floated in-line 😃

tuf/api/metadata.py
```diff
@@ -1,3 +1,38 @@
...
+ * Validation (some thoughts ...)
+   - Avoid schema, see secure-systems-lab/securesystemslib#183
+   - Provide methods to validate JSON representation (at user boundary)
+   - Fail on bad json metadata in read_from_json method
+   - Be lenient on bad/invalid metadata objects in memory, they might be
+     work in progress. E.g. it might be convenient to create empty metadata
+     and assign attributes later on.
+   - Fail on bad json metadata in write_to_json method, but with option to
+     disable check as there might be a justified reason to write WIP
+     metadata to json.
```
joshuagl, line 23: 💯 This all sounds sensible to me.

tuf/api/metadata.py
```diff
@@ -1,3 +1,38 @@
...
+
+ * It might be nice to have short-cuts on the Metadata class to methods and
+   attributes of the contained Signed object. If we do this, we should only do
+   it on common methods/attributes (e.g. version, bump_version, expires,
+   bump_expiration)
```
joshuagl, line 28: I'm undecided on this. I'm leaning in favour of not mapping attributes of the Signed object into the metadata class, because something like `timestamp.signed.expires` better matches the metadata format in the specification and that's what we set out to achieve with this re-working effort.

tuf/api/metadata.py
trishankatdatadog: You mean `Metadata.(sign|verify)`? To call `Key.(sign|verify)`, really, it should not be concerned with how keys actually do the signing. I have no strong opinions on where `signed` and `signatures` come from, please feel free to adjust according to your new implementation.

tuf/api/keys.py
```diff
@@ -0,0 +1,98 @@
...
+
+# Specific types of keys, such as those in RAM, or on HSMs (TODO).
+
+class RAMKey(Key):
+
+    def __init__(self, obj: Any) -> None: # pylint: disable=super-init-not-called
+        self.__obj = obj
+
+    @classmethod
+    def read_from_file(cls, filename: str,  algorithm: str, passphrase: Optional[str] = None, storage_backend: Optional[StorageBackendInterface] = None) -> Key:
```
trishankatdatadog: We don't need `storage_backend` here (my bad).

tuf/api/metadata.py
lukpueh: Not sure I fully understand. We do override the existing *fileinfo* in L260. So any previously loaded `version`, `length` and `hashes` will be discarded at any rate.

tuf/api/metadata.py
lukpueh: Yeah, I'm a bit torn too. I think as long as it is clear that the `expires`, `version`, etc. are only convenience short-cuts/aliases to those attributes defined on the inner metadata it is okay. 

Still, it might be confusing to have aliases on the common attributes only, but not on those specific to any of timestamp, snapshot, targets... and adding conditional shortcuts to specific fields too might make the short-cutting implementation to complicated.

So if we agree that the typing effort (as in amount of used characters) of e.g. `timestamp.signed.expires` is still reasonable, we can go without aliases for the time being.

Besides, it might make sense to adopt a workflow, where we only work on Metadata objects when needed, and otherwise work directly on the inner metadata, e.g.:
```python

timestamp_md = Timestamp.load_from_json("timestamp.json")
timestamp_md.verify(keys)

# Discard container once existing signatures are verified and
# do all the heavy lifting directly on timestamp
timestamp = timestamp_md.signed
# timestamp.foo()
# timestamp.bar()
# ...

# Re-wrap timestamp in Metadata when signing is needed,
timestamp_md = Metadata(signed=timestamp)
timestamp_md.sign(keys)
timestamp_md.write_to_json()
```

joshuagl: > So if we agree that the typing effort (as in amount of used characters) of e.g. `timestamp.signed.expires` is still reasonable, we can go without aliases for the time being.

I vote we go without aliases to start. In terms of API stability, it's easier to add them later if they are needed than it is to remove them.

> Besides, it might make sense to adopt a workflow, where we only work on Metadata objects when needed, and otherwise work directly on the inner metadata, e.g.:
> 
> ```python
> timestamp_md = Timestamp.load_from_json("timestamp.json")
> timestamp_md.verify(keys)
> 
> # Discard container once existing signatures are verified and
> # do all the heavy lifting directly on timestamp
> timestamp = timestamp_md.signed
> # timestamp.foo()
> # timestamp.bar()
> # ...
> 
> # Re-wrap timestamp in Metadata when signing is needed,
> timestamp_md = Metadata(signed=timestamp)
> timestamp_md.sign(keys)
> timestamp_md.write_to_json()
> ```

Nice idea. We should certainly give ourselves time to experiment with this pattern and see how it feels in use.

tuf/api/metadata.py
joshuagl: 🤦 if we're not updating the existing values what I wrote makes no sense.

tuf/api/metadata.py
lukpueh: > I vote we go without aliases to start. In terms of API stability, it's easier to add them later if they are needed than it is to remove them.

Agreed.

> Nice idea. We should certainly give ourselves time to experiment with this pattern and see how it feels in use.

Forgot to mention, this is what we e.g. do in [layout in-toto's verification routine](https://github.com/in-toto/in-toto/blob/a621c3bf3cbe1dd442b2388c685b3130f032430b/in_toto/verifylib.py#L1462-L1465).

tuf/api/metadata.py
lukpueh: No worries! I thought you meant that someone wanted to update version only and keep the existing hashes and length. Right now this would require to do something like:
```python
obj.update(version=version, length=obj.length, hashes=obj.hashes)
```
But I think that's okay.

tuf/api/metadata.py
trishankatdatadog: I like this idea. Feels a lot better than trying to update signatures in place. Most importantly, we should ask what @woodruffw thinks...

lukpueh:
Here is a summary of my thoughts about signing, signature verification and key handling in regards to this API.

**Keys**
Neither signing keys nor verification keys should be attributes of the role, i.e. `Metadata.signed` instance, they are meant to sign or verify. Instead and in order to align with the role model of the specification, they should be attributes of the *delegating* (this includes root) role. Same goes for the signature threshold. 
Here are two suggestions how to design the signing (or verification) interface.

~~delegating_role.sign(delegated_role) # delegating role signs the delegated role~~ (see https://github.com/theupdateframework/tuf/pull/1060#issuecomment-660108655)
```python
delegated_role.sign(keys) # delegated role signs itself with keys, e.g. from delegating_role, or a detached keyring
```
Caveat: 
- Although signing keys conceptually belong to the delegating role (as in *"delegating role signs for delegated role*"), they must never be part of the corresponding role metadata.
- This API might be used to sign delegated role metadata without having (or requiring) full knowledge of the delegating role (only the delegating role's keys are needed).


**Sign**
This PR currently proposes a sophisticated way of updating signatures by keyids. Although this seems convenient, I am not sure if there is actually a use case. In order to keep the interface minimal, I suggest to implement a `sign` function that just appends the signature, and leave it up to the caller if they want to first remove all or a few signatures from the list.

For comparison, [here is how we do this in in-toto](https://github.com/in-toto/in-toto/blob/a621c3bf3cbe1dd442b2388c685b3130f032430b/in_toto/models/metadata.py#L137-L163).

**Verify** 
I think it would be good two provide two functions, one that only verifies if the signature(s) on the object are valid, and another one that checks if enough validly verified signature are available to meet a threshold.

In-toto has the former as [method on the metadata class](https://github.com/in-toto/in-toto/blob/a621c3bf3cbe1dd442b2388c685b3130f032430b/in_toto/models/metadata.py#L198-L256) and the latter as [part of the full in-toto final product verification routine](https://github.com/in-toto/in-toto/blob/a621c3bf3cbe1dd442b2388c685b3130f032430b/in_toto/verifylib.py#L376-L387).

lukpueh:
> - Although signing keys conceptually belong to the delegating role (as in "delegating role signs for delegated role"), they must never be part of the corresponding role metadata.

I'm actually not so sure about this. A signing key probably does not need to belong to the delegating role, it could come from anywhere. So I suggest we have the caller decide where to get it from, and just pass it to the signing function as argument.

And we should provide the same workflow for the simple signature verification function, i.e. passing an unaffiliated key to just verify a signature. Even if, in the course of the client verification, the key (and threshold) are taken from the delegating role.

trishankatdatadog:
> Here is a summary of my thoughts about signing, signature verification and key handling in regards to this API.

Agree with your thoughts. We could write some convenience functions later to make it easier to find and associate the right Keys with the right Metadata, but we don't need them for now.

joshuagl:
I was just writing a similar comment to @trishankatdatadog's. Really enjoying seeing this API evolve as a collaborative effort, thanks all!

trishankatdatadog:
> I was just writing a similar comment to @trishankatdatadog's. Really enjoying seeing this API evolve as a collaborative effort, thanks all!

Likewise 🙂 Hoping to push soon that new type of Key I talked about yesterday... 🤞

trishankatdatadog:
Good news everyone: with 9285727ec50d1bb746c980525efc03d647f9925f, we finally have a Vault-TUF integration, where metadata can be signed and verified using ECDSA/Ed25519/RSA keys kept entirely in Hashicorp Vault 🚀 

It is a decent start, but there are a few important TODOs left, notably:
- [x] Test that py-TUF can parse public keys produced by this integration
- [x] Test that py-TUF can verify signature produced by this integration
- [ ] Start Vault, get Vault address and root token, and enable Transit Secrets Engine in unit tests
- [ ] Allow passing in a prehashed input instead of the whole input

trishankatdatadog:
Before we finalize, I also think we should consider the notion of [unbounded vs bounded keys](https://github.com/in-toto/in-toto-golang/pull/56#issuecomment-662689093).

joshuagl:
The Vault integration sounds very interesting, great work @trishankatdatadog. 💯 

I wonder whether a key management interface would be better suited to being integrated at the securesystemslib level, rather than TUF?

Even if TUF is the more appropriate home, I think it would be better to keep this PR focused on introducing the CRUD API. With any extra functionality built on top of that introduced in future PR's. What do you think?

On the CRUD API; @woodruffw did you have chance to review the proposed API and see how well it suits your needs for Warehouse integration?

trishankatdatadog:
> Even if TUF is the more appropriate home, I think it would be better to keep this PR focused on introducing the CRUD API. With any extra functionality built on top of that introduced in future PR's. What do you think?

No problemo. I will let @lukpueh, @woodruffw, and yourself decide.

tuf/api/keys.py
```diff
@@ -0,0 +1,427 @@
...
+            salt_length=salt_length
+        )
+
+    def verify(self, signed: BytesOrStr, signature: Dict) -> bool:
+        signed_bytes = self._encode(signed)
+        return verify_signature(self.__obj, signature, signed_bytes)
+
+class VaultKey(Key):
+
+    class AuthenticationError(Exception): pass
```
woodruffw, line 169: Tiny nit, but perhaps don't nest this exception class or the following enums?

Having the former in a standalone `exceptions` module and the latter in a `models` (or similar) module makes it a bit easier to reuse them in other contexts.

tuf/api/metadata.py
woodruffw: Sorry for the late response -- I agree with @joshuagl here.

tuf/api/metadata.py
```diff
@@ -0,0 +1,289 @@
...
+import tuf.formats
+
+# Types.
+
+JsonDict = Dict[str, Any]
+
+# Classes.
+
+
+class Metadata(ABC):
```
sechkova, line 73: I don't think any ABC features are used anymore?

tuf/api/metadata.py
```diff
@@ -0,0 +1,289 @@
...
+            if keyid_sig['keyid'] == keyid:
+                signatures[idx] = keyid_signature
+                updated = True
+        if not updated:
+            signatures.append(keyid_signature)
+
+    def sign(self, key_ring: KeyRing) -> JsonDict:
+        # FIXME: Needs documentation of expected behavior
+        signed_bytes = self.signed_bytes
+        signatures = self.__signatures
```
sechkova, line 99: Is there any intention behind` self.__signatures` and `self.signatures` both existing?

tuf/api/metadata.py
```diff
@@ -0,0 +1,289 @@
...
+
+        for key in key_ring.keys:
+            signature = key.sign(self.signed_bytes)
+            self.__update_signature(signatures, key.keyid, signature)
+
+        self.__signatures = signatures
+        return self.signable
```
sechkova, line 106: `self.signable` seems like a leftover?

tuf/api/metadata.py
```diff
@@ -0,0 +1,289 @@
...
+        # build_dict_conforming_to_schema
+        return {
+            '_type': self._type,
+            'version': self.version,
+            'spec_version': self.spec_version,
+            'expires': self.expires
+        }
+
+    @classmethod
+    def read_from_json(
```
sechkova, line 196: I would expect `read_from_json` and `write_to_json` to be part of the same `Metadata` class. What do you think?

tuf/api/metadata.py
lukpueh: Yes, let's remove.

tuf/api/metadata.py
lukpueh: I don't think so. I haven't touched the sign/verify functions yet, because I wasn't sure about the original intended behavior (that's why I left above comment there).

But I think we agreed that my plans for these functions as outlined in https://github.com/theupdateframework/tuf/pull/1060#issuecomment-660056376 seemed reasonable.

tuf/api/metadata.py
lukpueh: I agree, the only problem is that `read_from_json`, as opposed to `write_to_json`, requires knowledge about the class it should instantiate an object from for the `signed` field. That's why it made sense to me that this method is on the Signed class (i.e. any of Target/Timestamp/Snapshot).

But it would be easy to also have a `Metadata.read_from_json` that instantiates the corresponding child class of `Signed`, based on the "signed.type" JSON field.

That's what we do in [in-toto](https://github.com/in-toto/in-toto/blob/73dcfc4a3e8ae739873377883c3a5a6cb3e15099/in_toto/models/metadata.py#L116-L120).

sechkova:
AFAICT initially this API was created having in mind easily updating a single targets metadata but if we consider it as a base of a tuf rework we need an implementation of a `Root(Signed)` class too.

tuf/api/metadata.py
lukpueh: FYI https://github.com/theupdateframework/tuf/pull/1112/commits/41fe57f58b9351d1bcfbb20f55be0d1be351c6d0

joshuagl:
The core feature of this PR, the metadata API, has landed in #1112. That leaves the Vault functionality of this PR to resolve.

I think securesystemslib would be a good home for it, but either way. Should we close this PR and open an Issue in sslib or here to request the Vault feature with a link to this PR for a reference implementation?

trishankatdatadog:
> The core feature of this PR, the metadata API, has landed in #1112. That leaves the Vault functionality of this PR to resolve.
> 
> I think securesystemslib would be a good home for it, but either way. Should we close this PR and open an Issue in sslib or here to request the Vault feature with a link to this PR for a reference implementation?

Yes, please, we should upstream the Hashicorp Vault functionality against securesystemslib

joshuagl:
Closing this PR as the metadata API landed in #1112. 

Cross-linking to a securesystemslib Issue requesting an enhancement to support Hashicorp Vault: secure-systems-lab/securesystemslib#307

Issue #1059, opened 2020-06-26
Tests fail with recent change to formats.PATH_SCHEMA in securesystemslib master


joshuagl:
**Description of issue or feature request**:

Tests fail with securesystemslib master since the change in secure-systems-lab/securesystemslib#244 to adapt `securesystemslib.formats.PATH_SCHEMA` to use `securesystemslib.schema.AnyNonEmptyString`

See a recent Travis log: https://travis-ci.org/github/theupdateframework/tuf/jobs/702415488

**Current behavior**:

Tests fail with securesystemslib master

**Expected behavior**:

Tests pass with securesystemslib master

joshuagl:
Resolved with #1064

PR #1064, opened 2020-06-30
Fix failing unit tests

MVrachev:
**Fixes issue #**: https://github.com/theupdateframework/tuf/issues/1059

Corresponding to the securesyslib changes made in this pr:
secure-systems-lab/securesystemslib#244
which changed the securesyslib.formats.PATH schema to be of type
AnyNonemptystring.
This made the tuf unit tests to fail because there are to places
where functional arguments should comply with the
securesyslib.formats.PATH schema, but have a default value of
an empty string.

There are already implemented which checks for cases when those
functional arguments are empty strings.
That's why there are failing unit tests.

requirements**:


joshuagl (approved):
Thanks for the fix @MVrachev, this looks appropriate to me as these are optional parameters and therefore we must support empty strings.

Issue #881, opened 2019-05-27
Roadmap for cli tools: repo.py, client.py


lukpueh:
**Description of issue or feature request**:
Recently, several issues and some fixes related to the TUF command line tools `repo.py` and `script.py` have been brought forward: #873, #874, #876, #877, #878, #879, #880 and [*"Some comments on TUF..."* on the mailinglist](https://groups.google.com/forum/#!msg/theupdateframework/sVDSqm_Ixdc/M5CoCkImAwAJ). These include concerns about the script names, unexpected behavior, missing TUF features, etc.

@awwad has [pointed out](https://groups.google.com/d/msg/theupdateframework/sVDSqm_Ixdc/0rAO1ihhAAAJ) that `repo.py` and `client.py` are mostly tutorial tools, useful to get familiar with TUF  (see [`QUICKSTART.md`](https://github.com/theupdateframework/tuf/blob/develop/docs/QUICKSTART.md#next-steps)). Whereas an actual TUF integration will require direct use of the underlying libraries `tuf.repository_tool` and  `tuf.client.updater`.

Given the recent GitHub and mailing list activity around these tool, we need to either clarify that these tools are purely for educational purpose, or develop them more actively so that they can actually used by an integrator.

Thanks, @cminyard and @lixuefeng2 for bringing this to our attention!

**Current behavior**:
The use of and plans for `repo.py` and `client.py` are unclear.

**Expected behavior**:
Decide to what extent `repo.py` and `client.py` should be enhanced, maintained, etc.

trishankatdatadog:
I think we should just remove these old document and code altogether. I think it confuses people more than it does help them. What do others think?

cminyard:
On Tue, May 28, 2019 at 07:32:34PM +0000, Trishank K Kuppusamy wrote:
> I think we should just remove these old document and code altogether. I think it confuses people more than it does help them. What do the others think/
> 

I would disagree strongly.

The client code was useful as a reference.  It's not terribly useful on
it's own, but it could be used for some basic testing.

The repo code is useful in a lot of ways:

* The level of abstraction in TUF currently is fairly low.  This has a
  couple of bad side effects:
  - Everyone will re-create the same code over and over again.
  - People will get things wrong.
  - People won't use it because it's too hard to use.
  If you have a good level of abstraction, it makes things a lot easier.
  For instance, there is no "add this set of files to the repository as
  this role" function.  But that is something that everyone who uses it
[...]

PR #1066, opened 2020-07-03
repo.py: Decode key bytes for SHA key as well

jku:
key bytes were decoded as utf-8 for the JSON keys. Do the same for SHA
key bytes.

---

requirements**:

- [ ] Tests have been added for the bug fix or new feature -- repo.py isn't tested: I don't think I'm adding tests just for this


lukpueh:
Thanks for the patch, @jku! Please see #881 for context about cli tools.

joshuagl (approved):
Thanks for the patch @jku. Good to merge this and have the tools working whilst their futures is decided in #881

PR #1067, opened 2020-07-06
test_repository_tool: Add missing clear_all flag

jku:
Without this the next create_*db() call will fail meaning
   pytest test_repository_tool.py
fails on every test after this one.

### DIFF
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -81,9 +81,6 @@ class TestRepository(unittest.TestCase):
 
 
   def setUp(self):
-    tuf.roledb.clear_roledb(clear_all=True)
-    tuf.keydb.clear_keydb(clear_all=True)
-
     tuf.roledb.create_roledb('test_repository')
     tuf.keydb.create_keydb('test_repository')
 
@@ -1110,8 +1107,8 @@ class TestTimestamp(unittest.TestCase):
 
 
   def tearDown(self):
-    tuf.roledb.clear_roledb()
-    tuf.keydb.clear_keydb()
+    tuf.roledb.clear_roledb(clear_all=True)
+    tuf.keydb.clear_keydb(clear_all=True)
 
 
 


joshuagl:
Thanks for the patch @jku ! I am not familiar with pytest, so I'm not sure why this is required.

The `setUp()` method, which should be called at the start of each test, is already invoking `clear_*db(clear_all=True)`. AIUI that should handle things not being correctly cleaned up in `tearDown()` (though maybe cleaning up correctly is the better way to do things, and we should remove those calls from `setUp()`?

The tests do work in our CI and if I invoke locally from within the tests subdirectory with `python3 -m unittest -q test_repository_tool`.

Should pytest work with standard library unittests? OOI why use pytest and not tox?

jku:
Sorting in order of importance

> The tests do work in our CI and if I invoke locally from within the tests subdirectory with `python3 -m unittest -q test_repository_tool`.
>
> Should pytest work with standard library unittests?

I have no idea because I thought they were the same thing: they don't seem to be. So this is probably  just me not knowing about python testing and not a bug.

> why use pytest and not tox?

To the question "why not use tox": the tox setup seems a little broken to me as there's no way to run a single test because the default environment runs coverage... I've tried to write a new environment config for that but it becomes a little clunky. Calling "python3 -m unittest" seems fine to me: I just didn't know it

> The `setUp()` method, which should be called at the start of each test, is already invoking `clear_*db(clear_all=True)`. AIUI that should handle things not being correctly cleaned up in `tearDown()` (though maybe cleaning up correctly is the better way to do things, and we should remove those calls from `setUp()`?

This is only true for one of the classes I think: the others do not invoke the clear functions in `setUp()` at all -- this one was the only class that does not use clear_all out of all the classes so I figured that was the issue.

jku:
Actually I think it is a real bug: if you rename TestTimestamp so it isn't the last test unittest runs, your command fails as well

lukpueh:
> To the question "why not use tox": the tox setup seems a little broken to me as there's no way to run a single test because the default environment runs coverage... I've tried to write a new environment config for that but it becomes a little clunky. Calling "python3 -m unittest" seems fine to me: I just didn't know it

FYI, you can run `unittest`-powerd modules/classes/tests directly like so:

```bash
python test_repository_tool.py # Runs all test cases (classes) in that module
python test_repository_tool.py TestRepository # Runs all tests (functions) in TestRepository test case
python test_repository_tool.py TestRepository.test_writeall # Runs test_writeall test only
```

Not sure if we can configure tox to allow this fine-grained control, or if that is actually necessary. On my box it would definitely take too long to create a fresh virtual environment and install everything with tox each time I want to run a specific test function. So I usually just run it directly in my dev virtual environment, which uses a recent Python version and has everything installed. I only run tox before I push.

lukpueh:
@jku, do you think it would be helpful to have some of above info in our [contributor testing docs](https://github.com/theupdateframework/tuf/blob/develop/docs/CONTRIBUTORS.rst#testing)?

jku:
> @jku, do you think it would be helpful to have some of above info in our [contributor testing docs](https://github.com/theupdateframework/tuf/blob/develop/docs/CONTRIBUTORS.rst#testing)?

They seem obvious in hindsight ...but yes: I was following those instructions and did not figure out how to tweak them to run a single test case.

jku:
wrote a better commit message (without the pytest confusion), also removed the superfluous cleanup Joshua noticed -- I think that's probably a workaround of a similar issue earlier in history.

joshuagl (approved):
Thank you for the fix, the discussion and the cleaned up commit message.

PR #1074, opened 2020-07-09
Update targets delegations in generate_targets_metadata 

sechkova:
**Fixes issue #**: #1037

- Collect keys and threshold of delegated roles and update delegations in `generate_targets_metadata() `in a similar manner as `generate_root_metadata()` does for top-level roles.
- Move key dictionary generation from `generate_root_metadata()` to a helper function` _add_keys_to_keydict()` 
- Update tests

Fixing this issue was partially blocked by #574: after loading the repository the delegated roles' 'keyids' and 'threshold' are lost which leads to throwing "missing key" exceptions while trying to update the delegations.

I suggest a short-term solution of #574 by using the recently added logic of hierarchically loading the delegations in `load_reposiroty()`. This makes it possible to get the delegated roles `keyids` and `threshold` from its delegating role metadata.
Clearly this won't work if more than one delegation to the same role exists but ... this is not supported correctly anyway in the current implementation so lets cover the simplest use case here?

requirements**:

### DIFF
--- a/tests/test_repository_lib.py
+++ b/tests/test_repository_lib.py
@@ -36,6 +36,7 @@ import tempfile
 import json
 import shutil
 import unittest
+import copy
 
 import tuf
 import tuf.formats
@@ -352,26 +353,47 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     roleinfo = tuf.roledb.get_roleinfo('targets')
     delegations = roleinfo['delegations']
 
-    targets_metadata = \
-      repo_lib.generate_targets_metadata(targets_directory, target_files,
-                                         version, expiration_date, delegations,
-                                         False)
+    targets_metadata = repo_lib.generate_targets_metadata(targets_directory,
+        target_files, version, expiration_date, delegations, False)
     self.assertTrue(tuf.formats.TARGETS_SCHEMA.matches(targets_metadata))
 
     # Valid arguments with 'delegations' set to None.
-    targets_metadata = \
-      repo_lib.generate_targets_metadata(targets_directory, target_files,
-                                         version, expiration_date, None,
-                                         False)
+    targets_metadata = repo_lib.generate_targets_metadata(targets_directory,
+        target_files, version, expiration_date, None, False)
     self.assertTrue(tuf.formats.TARGETS_SCHEMA.matches(targets_metadata))
+
+    # Test update in targets' delegations
+    keystore_path = os.path.join('repository_data', 'keystore')
+    targets_public_keypath = os.path.join(keystore_path, 'targets_key.pub')
+    targets_public_key = securesystemslib.interface.\
+        import_ed25519_publickey_from_file(targets_public_keypath)
+
+    # Add new key and threshold to delegated role
+    repository.targets('role1').add_verification_key(targets_public_key)
+    repository.targets('role1').threshold = 2
+    role1_keyids = tuf.roledb.get_role_keyids('role1')
+    role1_threshold = tuf.roledb.get_role_threshold('role1')
+    roleinfo = tuf.roledb.get_roleinfo('targets')
+    delegations = roleinfo['delegations']
+    old_delegations = copy.deepcopy(delegations)
+
+    targets_metadata  = repo_lib.generate_targets_metadata(targets_directory,
+        target_files, version, expiration_date, delegations, False)
+    self.assertNotEqual(old_delegations, delegations)
+    self.assertEqual(role1_keyids,
+        targets_metadata['delegations']['roles'][0]['keyids'])
+    self.assertEqual(role1_threshold,
+        targets_metadata['delegations']['roles'][0]['threshold'])
+    for keyid in role1_keyids:
+      self.assertIn(keyid, targets_metadata['delegations']['keys'])
+
 
     # Verify that 'digest.filename' file is saved to 'targets_directory' if
     # the 'write_consistent_targets' argument is True.
     list_targets_directory = os.listdir(targets_directory)
-    targets_metadata = \
-      repo_lib.generate_targets_metadata(targets_directory, target_files,
-                                          version, expiration_date, delegations,
-                                          write_consistent_targets=True)
+    targets_metadata = repo_lib.generate_targets_metadata(targets_directory,
+        target_files, version, expiration_date, delegations,
+        write_consistent_targets=True)
     new_list_targets_directory = os.listdir(targets_directory)
 
     # Verify that 'targets_directory' contains only one extra item.
--- a/tests/test_repository_lib.py
+++ b/tests/test_repository_lib.py
@@ -310,11 +310,6 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     # securesystemslib.exceptions.Error exception is raised for duplicate keyids.
     tuf.keydb._keydb_dict['default'][root_keyids[0]]['keytype'] = 'rsa'
 
-    # Add duplicate keyid to root's roleinfo.
-    tuf.roledb._roledb_dict['default']['root']['keyids'].append(root_keyids[0])
-    self.assertRaises(securesystemslib.exceptions.Error, repo_lib.generate_root_metadata, 1,
-                      expires, consistent_snapshot=False)
-
     # Test improperly formatted arguments.
     self.assertRaises(securesystemslib.exceptions.FormatError, repo_lib.generate_root_metadata,
                       '3', expires, False)
@@ -349,27 +344,13 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     file_permissions = oct(os.stat(file1_path).st_mode)[4:]
     target_files = {'file.txt': {'custom': {'file_permission': file_permissions}}}
 
-    delegations = {"keys": {
-      "a394c28384648328b16731f81440d72243c77bb44c07c040be99347f0df7d7bf": {
-       "keytype": "ed25519",
-       "keyval": {
-        "public": "3eb81026ded5af2c61fb3d4b272ac53cd1049a810ee88f4df1fc35cdaf918157"
-       }
-      }
-     },
-     "roles": [
-      {
-       "keyids": [
-        "a394c28384648328b16731f81440d72243c77bb44c07c040be99347f0df7d7bf"
-       ],
-       "name": "targets/warehouse",
-       "paths": [
-        "/file1.txt", "/README.txt", '/warehouse/'
-       ],
-       "threshold": 1
-      }
-     ]
-    }
+    # Delegations data must be loaded into roledb since
+    # generate_targets_metadata tries to update delegations keyids
+    # and threshold
+    repository_path = os.path.join('repository_data', 'repository')
+    repository = repo_tool.load_repository(repository_path)
+    roleinfo = tuf.roledb.get_roleinfo('targets')
+    delegations = roleinfo['delegations']
 
     targets_metadata = \
       repo_lib.generate_targets_metadata(targets_directory, target_files,
@@ -957,6 +938,21 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     filenames = repo_lib.get_top_level_metadata_filenames(metadata_directory)
     repository = repo_tool.create_new_repository(repository_directory, repository_name)
     repo_lib._load_top_level_metadata(repository, filenames, repository_name)
+
+    # Manually add targets delegations to roledb since
+    # repository.write('targets') will try to update its delegations
+    targets_filepath = os.path.join('repository_data', 'repository',
+                                 'metadata', 'targets.json')
+    targets_signable = securesystemslib.util.load_json_file(targets_filepath)
+    delegations = targets_signable['signed']['delegations']
+
+    roleinfo = {}
+    roleinfo['name'] = delegations['roles'][0]['name']
+    roleinfo['keyids'] = delegations['roles'][0]['keyids']
+    roleinfo['threshold'] = delegations['roles'][0]['threshold']
+    roleinfo['version'] = 1
+    tuf.roledb.add_role('role1', roleinfo, repository_name)
+
 
     # Partially write all top-level roles (we increase the threshold of each
     # top-level role so that they are flagged as partially written.
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -166,7 +166,13 @@ def _generate_and_write_metadata(rolename, metadata_filename,
     metadata = generate_targets_metadata(targets_directory,
         roleinfo['paths'], roleinfo['version'], roleinfo['expires'],
         roleinfo['delegations'], consistent_targets, use_existing_fileinfo,
-        storage_backend)
+        storage_backend, repository_name)
+
+    # Update roledb with the latest delegations info collected during
+    # generate_targets_metadata()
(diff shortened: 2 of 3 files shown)


tuf/repository_lib.py
```diff
@@ -155,9 +155,16 @@ def _generate_and_write_metadata(rolename, metadata_filename,
...
-    metadata = generate_targets_metadata(targets_directory, roleinfo['paths'],
-        roleinfo['version'], roleinfo['expires'], roleinfo['delegations'],
-        consistent_targets, use_existing_fileinfo, storage_backend)
+    metadata, delegations_update = generate_targets_metadata(targets_directory,
+        roleinfo['paths'], roleinfo['version'], roleinfo['expires'],
+        roleinfo['delegations'], consistent_targets, use_existing_fileinfo,
+        storage_backend, repository_name)
+
+    # Update roledb with the latest delegations info
+    if delegations_update:
```
joshuagl: I assume the reason we're having `generate_targets_metadata()` return a boolean value and testing it here is because `update_roleinfo()` is non-performant due to its use of `deepcopy()`? What is the impact if we just `update_roleinfo()` without this check?

tuf/repository_lib.py
```diff
@@ -1466,6 +1468,47 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
+  Non-public function that generates keys for the keyids listed and adds
+  them to a key dictionary.
+  """
+
+  for keyid in keyids:
+
+    # Do we have a duplicate?
+    if keyids.count(keyid) > 1:
+      raise securesystemslib.exceptions.Error('Same keyid listed twice:'
+          ' ' + keyid)
```
joshuagl: This feels odd. We do not handle this exception anywhere, is this likely to happen in practice?
Furthermore, this is all data we manage. If duplicates really are an issue, should we just prune them? What are we expecting the 'user' to do when they see this exception?

tuf/repository_lib.py
```diff
@@ -1401,9 +1381,31 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
     tuf.formats.DELEGATIONS_SCHEMA.check_match(delegations)
 
+    # Check if delegated role keys or threshold are changed and update
+    # delegations accordingly. Collect a new delegations key dictionary.
+    keydict = {}
+    delegation_update = False
+    for role in delegations['roles']:
+      rolename = role['name']
+      role_keyids = tuf.roledb.get_role_keyids(rolename, repository_name)
+      role_threshold= tuf.roledb.get_role_threshold(rolename, repository_name)
```
```suggestion
      role_threshold = tuf.roledb.get_role_threshold(rolename, repository_name)
```

tuf/repository_lib.py
```diff
@@ -1401,9 +1381,31 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
+
+      if role_keyids != role['keyids'] or \
+          role_threshold != role['threshold']:
+        role['keyids'] = role_keyids
+        role['threshold'] = role_threshold
+        delegation_update = True
+
+      _add_keys_to_keydict(keydict, role_keyids, repository_name)
```
joshuagl: Should this be part of the `if` conditional?

tuf/repository_lib.py
```diff
@@ -1401,9 +1381,31 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
+
+    if delegation_update:
+      delegations['keys'] = keydict
```
joshuagl: Is there any harm in always assigning keydict to `delegations['keys']`? So far as I can tell `delegations['keys']` is only used when `delegation_update == True`?

joshuagl (commented):
Thanks for the PR @sechkova! Your approach seems sound, but I have a couple of questions in-line about the implementation.

tuf/repository_lib.py
```diff
@@ -1466,6 +1431,47 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
+    # 'securesystemslib.formats.RSAKEY_SCHEMA', and have the form:
+    # {'keytype': 'rsa',
+    #  'keyid': keyid,
+    #  'keyval': {'public': '-----BEGIN RSA PUBLIC KEY----- ...',
+    #             'private': '-----BEGIN RSA PRIVATE KEY----- ...'}}
+    keyid = key['keyid']
+    if keyid not in keydict:
+
+      # This appears to be a new keyid.  Generate the key for it.
+      if key['keytype'] in ['rsa', 'ed25519', 'ecdsa-sha2-nistp256']:
```
jku: I wonder if there's a reason for hard coding the key types here: we could just let format_keyval_to_metadata() throw FormatError (or handle it and throw something better) ?

tuf/repository_lib.py
sechkova: Yes, the idea is to avoid calling `update_roleinfo()` for each delegation on every `write`, especially when their number can grow high if delegated hashed bins are used. 

Anyway: 
- when tested with the scrip from https://github.com/theupdateframework/tuf/issues/1005#issuecomment-601351054 the overhead when updating all delegations and not using the `delegations_update` boolean seems to be not more than ~1-2secs (tested in a regular VM on my laptop)
- assuming that delegated hashed bins all use the same key, key revocation would mean updating all delegations

So if you are hinting towards simplifying the code at the cost of some negligible efficiency, you may have a point :slightly_smiling_face: 

In general, the overall overhead of the code in this PR seems to be in the range of 1-2 seconds (too small to actually calculate an exact delta without a dedicated test setup)

tuf/repository_lib.py
sechkova: No, no obvious harm ...

tuf/repository_lib.py
sechkova: No, it has to be outside the `if`. The key dictionary has to contain delegations keys of all roles, updated or not.

tuf/repository_lib.py
sechkova: There is a test case about it :grin: 
In practice `add_verification_key` has a check to avoid duplicates.
For the rest of your questions, I don't have a strong opinion but I agree this does not seem to be the right place of the check. Unfortunately,  I cannot claim that I can track all possible data entry points for roledb  ... while this is the only "exit" point.  Maybe this is the historical reason for it. 
I would gladly remove the check and its test but I don't have a great insight on what to put instead. Nothing?

tuf/repository_lib.py
sechkova: Thanks for pointing it :) 
After a quick search,  `format_keyval_to_metadata()` is used at a couple of places in `tuf` the way you described it (just letting it throw FormatError) so I agree applying the same here.

It seems also worth trying to merge the functionality of `repository_tool._keys_to_keydict()` and `repository_lib._add_keys_to_keydict()` which seem suspiciously similar :eyes:

tuf/repository_lib.py
joshuagl: Thanks for crunching the numbers. I'd absolutely be in favour of the code simplification.

sechkova:
Some time has passed since the last updates on this pull request so I had to rebase and force push. However I think the reformatted set of changes is addressing all the comments up to now.

joshuagl:
> Some time has passed since the last updates on this pull request so I had to rebase and force push. However I think the reformatted set of changes is addressing all the comments up to now.

Thank you for the updates. They look good on a first pass, I'm aiming to do a more thorough review next week.

tuf/repository_lib.py
```diff
@@ -2256,6 +2228,27 @@ def disable_console_log_messages():
 
 
 
+def keys_to_keydict(keys):
+  """
+  Iterate over a list of keys and return a list of keyids and a dict mapping
+  keyid to key metadata
+  """
```
joshuagl: As we have promoted this to a public method (implicitly, by removing the leading `_`) we should improve the docstring so that it's easier for callers to make use of. Docstrings for public methods should include `<Arguments>` and `<Returns>` sections, at least. The `Arguments` section for this method should help a reader understand the types and any constraints of the objects in the list.

tuf/repository_lib.py
```diff
@@ -1365,6 +1376,9 @@ def generate_targets_metadata(targets_directory, target_files, version,
   <Returns>
     A targets metadata object, conformant to
     'tuf.formats.TARGETS_SCHEMA'.
+
+    delegation_update:
+      Boolean set to True if delegations are updated.
```
joshuagl: I think this value is no longer returned?

tuf/repository_lib.py
```diff
@@ -1382,8 +1396,25 @@ def generate_targets_metadata(targets_directory, target_files, version,
     raise securesystemslib.exceptions.Error('Cannot support writing consistent'
         ' targets and using existing fileinfo.')
 
-  if delegations is not None:
+  if delegations:
```
joshuagl: Can you help me understand this change? 

`if delegations is not None` ensures we only enter this loop if the delegations argument is changed from its default value of `None`.  As I understand it this is the standard pattern for checking whether an argument with a default value of `None` has been changed.

tests/test_repository_lib.py
```diff
@@ -961,6 +942,21 @@ def test__load_top_level_metadata(self):
...
 
+    # Manually add targets delegations to roledb since
+    # repository.write('targets') will try to update its delegations
+    targets_filepath = os.path.join('repository_data', 'repository',
+                                 'metadata', 'targets.json')
+    targets_signable = securesystemslib.util.load_json_file(targets_filepath)
+    delegations = targets_signable['signed']['delegations']
+
+    roleinfo = {}
+    roleinfo['name'] = 'role1'
```
joshuagl: Nit: it's probably better read the role name from the loaded data in `delegations`, as we are for keyids and threshold, in case the generated data changes.

joshuagl (changes requested):
Thank you. I have a few minor comments, mostly on docstring, but otherwise this is looking good.

tuf/repository_lib.py
sechkova: Done in  1acd07c

tuf/repository_lib.py
sechkova: Oops ... 
Removed in a18beb2

tuf/repository_lib.py
sechkova: My bad ... I didn't think it through ...  :facepalm:  
Reverted in a18beb2

tests/test_repository_lib.py
sechkova: Thanks, done in f1a0fb1

sechkova:
Someone with enough permissions to restart travis-ci? Seems like another timeout 👀

jku:
Bad news is  #1096 was merged and this now has conflicts. Good news is the proxy tests _should_ be more reliable once you rebase :)

tuf/repository_lib.py
```diff
@@ -1412,6 +1395,23 @@ def generate_targets_metadata(targets_directory, target_files, version,
...
     tuf.formats.DELEGATIONS_SCHEMA.check_match(delegations)
+    # If targets role has delegations, collect the up-to-date 'keyids' and
+    # 'threshold' for each role. Update the delegations keys dictionary.
+    delegations_keys = []
+    # Update 'keyids' and 'threshold' for each delegated role
+    for role in delegations['roles']:
+      role['keyids'] = \
+          tuf.roledb.get_role_keyids(role['name'], repository_name)
+      role['threshold']= \
+          tuf.roledb.get_role_threshold(role['name'], repository_name)
```
joshuagl: The [coding style](https://github.com/secure-systems-lab/code-style-guidelines#code-lay-out) suggests:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. If necessary, you can add an extra pair of parentheses around an expression, but sometimes using a backslash looks better.

Which suits my personal preference too.
```suggestion
      role['keyids'] = tuf.roledb.get_role_keyids(role['name'],
          repository_name)
      role['threshold']= tuf.roledb.get_role_threshold(role['name'],
          repository_name)
```

tuf/repository_lib.py
```diff
@@ -2256,9 +2256,42 @@ def disable_console_log_messages():
...
+    keyids.append(keyid)
+  return keyids, keydict
+
+
+
+
 if __name__ == '__main__':
   # The interactive sessions of the documentation strings can
-  # be tested by running repository_tool.py as a standalone module:
+  # be tested by running repository_lib.py as a standalone module:
```
joshuagl, line 2291: 👍

tests/test_repository_lib.py
```diff
@@ -387,6 +369,34 @@ def test_generate_targets_metadata(self):
...
+    role1_keyids = tuf.roledb.get_role_keyids('role1')
+    role1_threshold = tuf.roledb.get_role_threshold('role1')
+    roleinfo = tuf.roledb.get_roleinfo('targets')
+    delegations = roleinfo['delegations']
+    old_delegations = copy.deepcopy(delegations)
+
+    targets_metadata  = \
+      repo_lib.generate_targets_metadata(targets_directory, target_files,
+                                         version, expiration_date, delegations,
+                                         False)
```
joshuagl: I realise you've just copied other examples in this file, but it's more inline with the coding style to do something like:
```suggestion
    targets_metadata  = repo_lib.generate_targets_metadata(targets_directory,
        target_files, version, expiration_date, delegations, False)
```

joshuagl (approved):
Thanks for the updates @sechkova, this looks much nicer.

I have made a few minor style suggestions and this PR needs rebasing on the most recent changes in develop, but otherwise this looks good to me.

sechkova:
Thanks @joshuagl, applied your code style suggestions and rebased. Let's see the tests :) 

_Edit:_ Two of the tests fail due to reasons unknown to me ... would need another look ... seems related to https://github.com/secure-systems-lab/securesystemslib/pull/266

jku:
lint failure is #1115

sechkova:
Rebased on develop with the linter failure fix.

tuf/repository_lib.py
```diff
@@ -1238,43 +1239,11 @@ def generate_root_metadata(version, expiration_date, consistent_snapshot,
...
           ' "tuf.roledb".')
 
-    # Keep track of the keys loaded to avoid duplicates.
-    keyids = []
-
-    # Generate keys for the keyids listed by the role being processed.
-    for keyid in tuf.roledb.get_role_keyids(rolename, repository_name):
+    # Collect keys from all roles in a list
+    keyids = tuf.roledb.get_role_keyids(rolename, repository_name)
+    for keyid in keyids:
```
joshuagl, line 1250: Nit: this change to assigning the return value of `roledb.get_role_keyids` seems superfluous, is `keyids` used anywhere other than the iterator here? Why not just keep for `keyid in tuf.roledb....`?

joshuagl (approved):

tuf/repository_lib.py
sechkova: Yes, it is used in https://github.com/theupdateframework/tuf/pull/1074/files#diff-9f585d5478a17bd9e0f6cd5f7484ab60L1284

It replaces the keyids list that was used to avoid duplicates which we already decided not necessary somewhere in the discussion above.

joshuagl: Oh, yes I see. Thanks!

joshuagl:
Good work, thanks!

Issue #1084, opened 2020-07-17
Implement TAP 12


mnm678:
Add [tap12](https://github.com/theupdateframework/taps/blob/master/tap12.md) to the reference implementation. This TAP allows for greater flexibility in how keyids are generated. This will require the following changes to the client:

- [x] Remove all use of keyid_hash_algorithms and instead use the keyid defined by the repository. This step is well documented in #848. 
- [ ] Check that all keys applied to a signature are unique using a standardized representation (ex modulus and exponent for rsa) so that the same key with multiple keyids cannot be applied to a threshold. 
- [ ] Make a local keydb for each delegating targets role so that keys can be locally scoped to their delegating instance (ie where the keys are defined in the delegating role and used in the delegated role) 

The repository code should not need to change as sha256 may still be used to generate keyids.

mnm678:
I have started on step 1 in #1014.

lukpueh:
Two quick questions:
1. Is there a plan to completely remove `keyid_hash_algorithms` in the course of the TAP 12 implementation? If not, we need to create a separate issue to do so.
2. Do others think the removal of the field deserves a *decision record* (#1141)? 
   #848 has become quite a focal point for related discussions (see the many mentions from other GH issues/prs). Is it worth to retain a summary of these discussions?

mnm678:
>     1. Is there a plan to completely remove `keyid_hash_algorithms` in the course of the TAP 12 implementation? If not, we need to create a separate issue to do so.

Removing `keyid_hash_algorithms` would be part of a TAP 12 implementation, but it might be worth adding a separate issue as we decided in #1095 that most of the TAP 12 implementation should wait on the refactor work.

>     2. Do others think the removal of the field deserves a _decision record_ (#1141)?
>        #848 has become quite a focal point for related discussions (see the many mentions from other GH issues/prs). Is it worth to retain a summary of these discussions?

As this is already discussed in a TAP, I don't know if a decision record is needed. However, I'd rather over-document than under-document, so if others disagree we can add one.

joshuagl:
> > ```
> > 1. Is there a plan to completely remove `keyid_hash_algorithms` in the course of the TAP 12 implementation? If not, we need to create a separate issue to do so.
> > ```
> 
> Removing `keyid_hash_algorithms` would be part of a TAP 12 implementation, but it might be worth adding a separate issue as we decided in #1095 that most of the TAP 12 implementation should wait on the refactor work.

[...]

PR #1091, opened 2020-07-28
Each key applies to signature threshold once

mnm678:
This addresses part 2 of #1084.

This pr ensures that each key will only count toward the signature threshold once, even if the keys have different keyids. This will be especially important when repositories are allowed to define their own keyids as part of TAP 12.

The solution is based on @lukpueh's [commit for threshold verification](https://github.com/theupdateframework/tuf/commit/a0397c7c820ec1c30ebc793cc9469b61c8d3f50e)

### DIFF
--- a/tuf/sig.py
+++ b/tuf/sig.py
@@ -245,8 +245,7 @@ def verify(signable, role, repository_name='default', threshold=None,
     NOTE:
     - Signatures with identical authorized keyids only count towards the
       threshold once.
-    - Signatures with different authorized keyids each count towards the
-      threshold, even if the keyids identify the same key.
+    - Signatures with the same key only count toward the threshold once.
 
   <Arguments>
     signable:
@@ -307,7 +306,12 @@ def verify(signable, role, repository_name='default', threshold=None,
   if threshold is None or threshold <= 0: #pragma: no cover
     raise securesystemslib.exceptions.Error("Invalid threshold: " + repr(threshold))
 
-  return len(set(good_sigs)) >= threshold
+  unique_keys = set()
+  for keyid in good_sigs:
+    key = tuf.keydb.get_key(keyid, repository_name)
+    unique_keys.add(key['keyval']['public'])
+
+  return len(unique_keys) >= threshold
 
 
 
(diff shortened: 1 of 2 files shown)


tuf/sig.py
```diff
@@ -307,7 +306,12 @@ def verify(signable, role, repository_name='default', threshold=None,
   if threshold is None or threshold <= 0: #pragma: no cover
     raise securesystemslib.exceptions.Error("Invalid threshold: " + repr(threshold))
 
-  return len(set(good_sigs)) >= threshold
+  unique_keys = set()
+  for keyid in good_sigs:
+    key = tuf.keydb.get_key(keyid, repository_name)
+    unique_keys.add(key['keyval']['public'])
```
trishankatdatadog, line 312: This is great! I believe it is still not entirely robust, as different encodings of the same key, or even the same key with different schemes (e.g., RSA key with different signing/hashing algos) can throw it off.

mnm678: Yes, I think a next step would be to use a consistent representation (modulus and exponent, etc), but I think that would require adding that as a field to securesystemslib. I didn't see anything in the existing key schema could be used.

mnm678: After looking at this some more, I'm not sure if we need to add a consistent representation. The signature will only be verified by a key that has been delegated to, so issues with different encodings, etc will only exist if the delegating party lists the same key in multiple formats. We assume the delegating party is trusted (and if not they could do a much worse attack like reduce the threshold and add their own key). Is there anything I'm missing here?

trishankatdatadog: Yeah, we don't have to be this paranoid, but I think better paranoid than sorry. Suppose, for example, the original delegator got fooled, etc. We can file a separate issue for this.

mnm678: In that case, there is also a possibility that the delegator is tricked into changing the `consistent_representation` (or whatever the name would be) field when delegating to a key

joshuagl (approved):
Excellent change, thank you @mnm678 !

trishankatdatadog (approved):
Definitely more robust than what we had before. We can revisit public key "normalization" later. Thanks!

Issue #1083, opened 2020-07-17
logging in updater tests is weirdly broken


jku:
Something weird goes on in the updater tests logging. In the middle of the test (seemingly around the time the last root.json download fails during Updater.refresh()) logger level seems to somehow change to debug. It's not just that though: even if I change the level of some earlier logger calls (like changing logger.debug() to logger.error()) so that they become visible, they don't look the same as similar  logging later on. The earlier output from e.g. logger.error() isn't prefixed with logger level and component.

**Current behavior**:
```
jku@luna:~/src/tuf/tests$ python3 test_updater_root_rotation_integration.py TestUpdater.test_root_rotation
Adding a verification key that has already been used.
Adding a verification key that has already been used.
127.0.0.1 - - [17/Jul/2020 17:44:27] code 404, message File not found
Could not download URL: 'http://localhost:30379/tmp2kxfm_pt/TestUpdater_nsqxv0n9/repository/metadata/4.root.json'
Traceback (most recent call last):
  File "/home/jku/src/tuf/tuf/download.py", line 264, in _download_file
    response.raise_for_status()
  File "/usr/lib/python3/dist-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: File not found for url: http://localhost:30379/tmp2kxfm_pt/TestUpdater_nsqxv0n9/repository/metadata/4.root.json
Update failed from http://localhost:30379/tmp2kxfm_pt/TestUpdater_nsqxv0n9/repository/metadata/4.root.json.
Traceback (most recent call last):
  File "/home/jku/src/tuf/tuf/client/updater.py", line 1502, in _get_metadata_file
    file_object = tuf.download.unsafe_download(file_mirror,
  File "/home/jku/src/tuf/tuf/download.py", line 150, in unsafe_download
    return _download_file(url, required_length, STRICT_REQUIRED_LENGTH=False)
  File "/home/jku/src/tuf/tuf/download.py", line 264, in _download_file
    response.raise_for_status()
  File "/usr/lib/python3/dist-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: File not found for url: http://localhost:30379/tmp2kxfm_pt/TestUpdater_nsqxv0n9/repository/metadata/4.root.json
[...]

PR #1092, opened 2020-07-29
Improve updater and tests logging

sechkova:
**Fixes issue #**: #1083

- Add NullHandler() to the top-level logger
- Replace logging.exception calls in Updater()

requirements**:

### DIFF
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1151,12 +1151,12 @@ class Updater(object):
           # Otherwise, reraise the error, because it is not a simple HTTP
           # error.
           if neither_403_nor_404(mirror_error):
-            logging.exception('Misc error for root version '+str(next_version))
+            logger.exception('Misc error for root version '+str(next_version))
             raise
           else:
             # Calling this function should give us a detailed stack trace
             # including an HTTP error code, if any.
-            logging.exception('HTTP error for root version '+str(next_version))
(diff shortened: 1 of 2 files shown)


sechkova:
The proposed changes here are related to getting rid of the logging when it is not needed.
On the other hand, probably there's also a similar need of a common way to enable/increase the logging when needed (for debugging etc)

trishankatdatadog (approved):
💯 

Too much noise out-of-the-box right now

We can figure out configurable options later

tuf/client/updater.py
```diff
@@ -1151,12 +1151,12 @@ def neither_403_nor_404(mirror_error):
           # Otherwise, reraise the error, because it is not a simple HTTP
           # error.
           if neither_403_nor_404(mirror_error):
-            logging.exception('Misc error for root version '+str(next_version))
+            logger.exception('Misc error for root version '+str(next_version))
```
joshuagl, line 1154: 💯 excellent find. I've looked at this code numerous times and not noticed the typo here.

joshuagl (approved):
Thank you for finding and fixing this issue @sechkova! I'm not sure I completely agree with the current logging behaviour, but let's figure that out later.

tuf/client/updater.py
trishankatdatadog: My bad, thanks everyone

PR #1095, opened 2020-08-03
[WIP] Use separate keydb for delegated targets

mnm678:
Fixes issue: part 3 of #1084

To implement TAP 12, this pr adds separate keydbs for each delegating instance. This means that the roles delegated to by root will use one keydb, the roles delegated to by targets will use a different one, and so on.

To do so, this pr introduces keydb.create_keydb_from_targets_metadata to make a new keydb for all roles delegated to by a targets metadata file and adds a `parent_role` field to the roledb to keep track of what keydb should be used when verifying a role.

### DIFF
--- a/tests/test_developer_tool.py
+++ b/tests/test_developer_tool.py
@@ -375,7 +375,7 @@ class TestProject(unittest.TestCase):
 
 
     # Load_signing_keys.
-    project('delegation').load_signing_key(delegation_private_key)
+    project('delegation').load_signing_key(delegation_private_key, 'targets')
 
     project.status()
 
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -301,7 +301,6 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
 
   def test_1__load_metadata_from_file(self):
-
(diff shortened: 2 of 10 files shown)


trishankatdatadog:
This is awesome, thanks, Marina!

I'm a bit ambivalent about continuing to use the global key dictionary, even if isolated by repositories. Ideally, this should be done recursively. I understand we need to make progress, but I'm also hesitant about adding code we will eventually need to rewrite anyway.

It's a difficult problem. Thoughts @lukpueh @joshuagl @JustinCappos ?

trishankatdatadog:
@mnm678 would you please kindly mark this as a draft PR until ready? Thanks!

lukpueh:
In a recent TAP meeting we agreed that it might be better to not rework keydb, but instead hold off the implementation of TAP 12 until we have better key handling infrastructure, as e.g. discussed in https://github.com/theupdateframework/tuf/pull/1060#issuecomment-660056376. Should we close here, @mnm678?

mnm678:
Yes, I'll close this and we can revisit the TAP 12 implementation in the future.

Issue #1085, opened 2020-07-18
Remove sleep() from tests


jku:
**Description of issue or feature request**:

Tests use sleep() quite a lot: this makes them slow, and they still don't always protect whatever the sleep is trying to protect. 

**Current behavior**:
* `python3 test_updater.py` takes 25 seconds to run 38 tests
* Running the same command immediately again often fails (the sockets are still in TIME-WAIT state for a minute or so)
* Sometimes subprocess killing fails silently

**Status**:
I have a work-in-progress branch: https://github.com/jku/tuf/tree/remove-sleep-wip : it might not take care of all corner cases yet, the comments are still about the old implementation and it only covers test_updater.py but it does take execution time of `python3 test_updater.py` from 25 seconds to 8 seconds.

 I'm filing this so I don't forget the branch on the holiday -- but :1st_place_medal: for anyone who takes it, polishes and extends it

jku:
Fixed in pr #1096

PR #1096, opened 2020-08-04
Test performance and reliability fixes

jku:
Improve reliability and speed of the tests: altogether a full test run is > 40% faster on my laptop and _should_ be more reliable as well (there's still a possibility that e.g. the `new wait_for_server()` is not handling some exceptions it should but the potential issues should be resolvable over time -- unlike sleep() that is inherently unreliable). 

Also add a missing import in download.py and close some unclosed handles that make it difficult to read test output

Fixes issue #1085

---

There's quite a bit of repetition in e.g. the simple-server using tests... I didn't try to refactor that at the same time but it might make sense to have a SimpleServerTestMixin that does the required initialization and cleanup.

### DIFF
--- a/tests/test_key_revocation_integration.py
+++ b/tests/test_key_revocation_integration.py
@@ -56,6 +56,8 @@ import tuf.repository_tool as repo_tool
 import tuf.unittest_toolbox as unittest_toolbox
 import tuf.client.updater as updater
 
+import utils
+
 import securesystemslib
 import six
 
@@ -89,9 +91,7 @@ class TestKeyRevocation(unittest_toolbox.Modified_TestCase):
     logger.info('\tServing on port: '+str(cls.SERVER_PORT))
     cls.url = 'http://localhost:'+str(cls.SERVER_PORT) + os.path.sep
 
-    # NOTE: Following error is raised if a delay is not applied:
-    # <urlopen error [Errno 111] Connection refused>
-    time.sleep(1)
+    utils.wait_for_server('localhost', cls.SERVER_PORT)
 
 
 
--- a/tests/test_key_revocation_integration.py
+++ b/tests/test_key_revocation_integration.py
@@ -108,6 +108,7 @@ class TestKeyRevocation(unittest_toolbox.Modified_TestCase):
     if cls.server_process.returncode is None:
       logger.info('\tServer process '+str(cls.server_process.pid)+' terminated.')
       cls.server_process.kill()
+      cls.server_process.wait()
 
 
 
--- a/tests/test_key_revocation_integration.py
+++ b/tests/test_key_revocation_integration.py
@@ -38,14 +38,11 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import os
-import time
 import shutil
-import copy
 import tempfile
 import logging
 import random
 import subprocess
-import sys
 import unittest
 
 import tuf
--- a/tests/test_proxy_use.py
+++ b/tests/test_proxy_use.py
@@ -148,6 +148,8 @@ class TestWithProxies(unittest_toolbox.Modified_TestCase):
       if proc.returncode is None:
         logger.info('\tTerminating process ' + str(proc.pid) + ' in cleanup.')
         proc.kill()
+        # Drop return values of communicate()
+        proc.communicate()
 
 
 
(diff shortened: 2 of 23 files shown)


jku:
Interested in seeing the CI results... Even if they're green we might want to have more CI runs to make sure there are no flaky tests: based on git archeology this bit of code has been very prone to those.

EDIT: obviously a bit of work left :|

jku:
MSG_DONTWAIT does not  exist on windows apparently: will have to set the socket to non-blocking mode explicitly

jku:
The other issue is python2.7:
* can't use `with socket.socket(..) as sock:`
* BlockingIOError and ConnectionRefusedError are just OSError: would have to check errno to handle them properly

jku:
Well the windows results imply that Windows is not fine with my hack of calling connect() multiple times... I guess the by-the-book method would be to select() after getting the first EINPROGRESS (I'm assuming that's received first on Windows, just guessing though)

trishankatdatadog:
@jku would you please kindly mark this as a draft PR until ready? Thanks!

jku:
Okay, should be ready for review now.

My original idea of non-blocking connect() did not work on Windows at all: Latest version is using a blocking socket but luckily that's fast on Linux and not slow on Windows ("Connection Refused" can unfortunately take even 2 seconds on Windows because it does retries internally).

I've also bumped the timeout to 10 seconds since a) it does not increase the normal runtime at all and b) Travis actually managed to hit the 5 second mark once. I'm a little worried about Travis not managing to start simple_server.py in 5 seconds though... Doing some more CI runs might not be a bad idea: the full test runtime gives a rough indication of whether there were any really long waits.

joshuagl:
py35 on Travis CI failed the first time, which only indicates that the tests are still a little flakey.

jku:
> py35 on Travis CI failed the first time, which only indicates that the tests are still a little flakey.

That was a fossa upload failure I think (but I don't know how to view that build anymore).

jku:
WRT test speed:
* travis run goes from roughly 16min 30s to 12min 20s
* appveyor (Windows) from roughly 29 mins to 27 mins

tests/test_key_revocation_integration.py
```diff
@@ -516,4 +517,4 @@ def _load_role_keys(keystore_directory):
 
 
 if __name__ == '__main__':
-  unittest.main()
+  unittest.main()
```
joshuagl: NIT: no newline at end of file?

tests/test_proxy_use.py
```diff
@@ -118,15 +120,10 @@ def setUpClass(cls):
...
-    # was long enough. Use 0.5s to be safe, and if issues arise, increase it.
-    # Observed occasional failures at 0.1s, 0.15s, 0.5s, and 2s, primarily on
-    # AppVeyor.  Increasing to 4s.  This setup runs once for the module.
-    time.sleep(4)
-
-
+    utils.wait_for_server('127.0.0.1', cls.http_port)
+    utils.wait_for_server('127.0.0.1', cls.https_port)
+    utils.wait_for_server('127.0.0.1', cls.http_proxy_port)
+    utils.wait_for_server('127.0.0.1', cls.https_proxy_port)
```
joshuagl: NIT: can we just use localhost here, like in other tests?

joshuagl (commented):
Fantastic PR, thanks @jku. Very happy to see `sleep` calls being removed!

I was expecting to see some `import time` lines deleted too, are we still using time in all of the tests that are no longer calling `time.sleep`?

jku:
Thanks @joshuagl all relevant comments. I'll push these changes one by one to get a few more CI runs, will mark the comments resolved after everything is pushed

jku:
Last commit is a fixup for an early commit: I'll squash it once the 2nd review is there (or I can do it now, just let me know).

There are now 7 successive green CI runs for this branch: I feel confident in saying the reliability is on par or better than current develop branch.

joshuagl:
Excellent set of changes, thanks Jussi. The PR looks good to me, please rebase and squash/fixup and we'll be ready to approve.

jku:
This PR is ready to go from my perspective, in case that was unclear.

joshuagl (approved):
Thanks for these great improvements to the test reliability.

Issue #1050, opened 2020-06-09
Delegated roles metadata can be written to disk without signing


sechkova:
**Description of issue or feature request**:
Delegated roles can be written to disk without any signature.

When a role is delegated but no signing keys are loaded to the role, one can still write the metadata on disk without any signature or any error being raised.
```
repository.targets.delegate('unclaimed', [public_unclaimed_key], ['myproject/*.txt'])
repository.mark_dirty(['snapshot', 'targets','timestamp', 'unclaimed'])
repository.writeall()
```

unclaimed.json: 
```
{
 "signatures": [],
 "signed": {
  "_type": "targets",
  "delegations": {
   "keys": {},
   "roles": []
  },
  "expires": "2020-09-08T21:27:27Z",
  "spec_version": "1.0.0",
  "targets": {},
  "version": 1
 }
}
```
**Current behavior**:

Delegated roles are allowed to be written without a signature.

**Expected behavior**:

Raise an error if there is no key loaded for a delegated role.

sechkova:
Delegated roles are allowed to be partially signed and do not go through the same [verification steps](https://github.com/theupdateframework/tuf/blob/5d40ffa3c49f6ce8ba5908abacec5630833829e8/tuf/repository_lib.py#L179) as the top level roles but partially signed should not equal not signed at all.

joshuagl:
`repository_lib.sign_metadata()` takes a list of *keyid*'s for the signing keys and loops over the list, trying to sign the metadata with each associated key.

Only nowhere do we __ensure__ that there are a non-zero number of _keyid_'s passed to `sign_metadata()`. Therefore, if the list is empty (as in the case demonstrated here) the for loop exits and the function returns indicating everything is OK.

Note: `repository_lib.sign_metadata()` attempts to ensure that the generated signable object is legitimate before returning by calling `tuf.formats.check_signable_object_format()`, but that function does not check that the signatures field is non-empty.

We should:
* Ensure that keyids has one or more items (this might just be a check in `sign_metadata()`, or maybe it would make more sense for `securesystemslib.formats.KEYIDS_SCHEMA` to set the *min_count* of `securesystemslib.schema.ListOf`?
* Consider adding verification to `sign_metadata()` to ensure that at least one signature has been added to the returned signable object

sechkova:
[...]

PR #1100, opened 2020-08-06
Raise an error on loading/writing unsigned metadata

sechkova:
**Fixes issue #**: #1050

- Modify  `check_signable_object_format() `to raise an error if signable has an empty 'signatures' list.
- Remove usage of empty signing keys lists in developer_tool 
- Update test cases which use unsigned metadata.

~~**Note** Marked as WIP since `test_root_rotation_missing_keys` correctly fails due to a known verification issue on the Updater (#1101)~~

requirements**:

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -31,6 +31,7 @@ from __future__ import unicode_literals
 import unittest
 import datetime
 import sys
+import os
 
 import tuf
 import tuf.formats
@@ -38,6 +39,7 @@ import tuf.formats
 import utils
 
 import securesystemslib
+import securesystemslib.util
 import six
 
 
@@ -778,20 +780,10 @@ class TestFormats(unittest.TestCase):
 
   def test_make_signable(self):
     # Test conditions for expected make_signable() behavior.
-    root = {'_type': 'root',
-            'spec_version': '1.0.0',
-            'version': 8,
-            'consistent_snapshot': False,
-            'expires': '1985-10-21T13:20:00Z',
-            'keys': {'123abc': {'keytype': 'rsa',
-                                'scheme': 'rsassa-pss-sha256',
-                                'keyval': {'public': 'pubkey',
-                                           'private': 'privkey'}}},
-            'roles': {'root': {'keyids': ['123abc'],
-                               'threshold': 1,
-                               'paths': ['path1/', 'path2']}}}
-
     SIGNABLE_SCHEMA = tuf.formats.SIGNABLE_SCHEMA
+    root_file = os.path.join('repository_data', 'repository', 'metadata',
+        'root.json')
+    root = securesystemslib.util.load_json_file(root_file)
     self.assertTrue(SIGNABLE_SCHEMA.matches(tuf.formats.make_signable(root)))
     signable = tuf.formats.make_signable(root)
     self.assertEqual('root', tuf.formats.check_signable_object_format(signable))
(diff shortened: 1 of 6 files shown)


sechkova:
No longer WIP since the merging of #1101 and ready for review.

tests/test_developer_tool.py
```diff
@@ -348,7 +348,9 @@ def test_write(self):
     name_backup = project.project_name
 
     # Write and reload.
-    self.assertRaises(securesystemslib.exceptions.Error, project.write)
+    self.assertRaises(tuf.exceptions.UnsignedMetadataError, project.write)
+    project.load_signing_key(project_private_key)
+    project('delegation').load_signing_key(delegation_private_key)
```
joshuagl: I think this updated test would be better if we loaded a signing key, did `assertRaises` with `project.write` and then called `project.write(write_partial=True)`.

That way we can test that we continue to support writing partial metadata, even if we _require_ that all metadata be written with at least one signature.

I'd appreciate some input from other maintainers as to whether requiring at least one signature on all metadata is reasonable cc @lukpueh @trishankatdatadog @mnm678

joshuagl (changes requested):
Thanks for the PR @sechkova. Obviously this needs rebasing on recent changes in `develop` and I've requested a change in `test_developer_tool` to ensure we're testing behaviour we expect to support.

I've also added a question to my more experienced colleagues on the use of partially signed metadata.

tests/test_developer_tool.py
mnm678: I think that it's important to at least warn people when metadata is written without a signature, but I worry that requiring this may impact the ability to sign metadata offline. I believe we want to support the case where metadata is generated on one machine and signed on another.

trishankatdatadog: Agree with @mnm678

joshuagl: Thanks for the feedback both. For the purposes of this PR it sounds like we want to change this failure to a warning.

I wonder if in future, i.e. our refactor plans, we should have distinct modes of operating the API such that users can't accidentally write metadata without signatures. Instead they'd have to opt-in to generating unsigned/partially signed metadata.

trishankatdatadog: Yes, let's argue in future 🙂

sechkova:
Thanks all for sharing your opinion!
Taking the discussion into account I decided to keep the error raised in `check_signable_object_format()` if signable has an empty 'signatures' list but to except it and log a warning in `repository_lib` (and `developer_tool`).
This way we both keep the current functionality allowing for signing offline and we require developers to justify their actions when using `check_signable_object_format()` with unsigned metadata. And the error is still raised on the client side when loading unsigned metadata file.

I removed the assert cases expecting an error from the tests but still kept the rest which in my opinion improves their logic in general.

Let me know what do you think

joshuagl:
I like that a client loading unsigned metadata causes an error, yet we still support generating metadata on the repository for offline signing.

What do you think @trishankatdatadog @mnm678 ?

mnm678:
This looks good to me! What should the behavior be is there are less than a threshold of signatures on the metadata? Should this also cause the error?

trishankatdatadog:
_Client_ definitely has to unwaveringly and uncompromisingly strict: no threshold of sigs, no go.

As for _repository/developer tools_, I'm okay with explicit warnings.

sechkova:
> This looks good to me! What should the behavior be is there are less than a threshold of signatures on the metadata? Should this also cause the error?

The threshold of signatures is validated by `sig.verify()` and the changes in this PR don't affect this.

In the client logic `sig.verify()` is always called (often hidden down in the call stack) after `check_signable_object_format()`. 
The only exception is `Updater.__init__` where metadata which is assumed trusted is loaded. In this case `check_signable_object_format()` will now raise an error if the metadata is unsigned.

Anyway, if you have some doubts and you think it is worth the effort I can add a couple of tests for `updater` that will clarify this behaviour.

mnm678:
Sorry, my question wasn't clear. I meant what happens if the repository tries to write metadata that has less than a threshold of signatures? The error will be raised if there are no signatures, but I don't think it will if there are less than a threshold of signatures.

sechkova:
Let me try to explain my vision of it and hopefully all are on the same page.

On the repository side:
1) Current behaviour (before this set of changes):
- `sig.verify` ensures that threshold of signatures is met during writing of metadata
- if `allow_partially_signed` is enabled, there is no check of a threshold of signatures or if metadata is signed at all

A more detailed discussion is in #1050.

2) After the changes included in this PR:
- `sig.verify` ensures that threshold of signatures is met during writing of metadata
- if `allow_partially_signed` is enabled, after taking into account the discussion here, only a warning will be logged when the metadata is unsigned. Nothing will happen if the metadata is signed but a threshold of signatures is not met (I guess this is expected since `allow_partially_signed` is chosen). Maybe another warning in this case?

mnm678:
> Let me try to explain my vision of it and hopefully all are on the same page.

Thanks, I think this helps.

>     Nothing will happen if the metadata is signed but a threshold of signatures is not met (I guess this is expected since `allow_partially_signed` is chosen). Maybe another warning in this case?

This is what I wanted to make sure we clarify. If `allow_partially_signed` is already specified, I'm not sure another warning is needed, but I just want to make sure the behavior makes sense to users.

joshuagl:
Thanks for the explanations @sechkova and for the discussion @mnm678 – seems like we agree this change is good to merge?

joshuagl (approved):
Excellent, thanks!

PR #1101, opened 2020-08-06
Fix updater workflow

joshuagl:
**Fixes issue #**: N/A

The updater was not adhering to the detailed client workflow. Specifically newly downloaded root metadata was not being verified with a threshold of signatures from itself.

Per the detailed client workflow in the specification step 1.2
> Version N+1 of the root metadata file MUST have been signed by:
(1) a threshold of keys specified in the trusted root metadata file (version N), and
(2) a threshold of keys specified in the new root metadata file being validated (version N+1)."

This PR ensures that point two is verified by the updater.

requirements**:

### DIFF
--- a/tests/test_updater_root_rotation_integration.py
+++ b/tests/test_updater_root_rotation_integration.py
@@ -217,6 +217,68 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
 
 
+  def test_verify_root_with_current_keyids_and_threshold(self):
+     """
+     Each root file is signed by the current root threshold of keys as well
+     as the previous root threshold of keys. Test that a root file which is
+     not 'self-signed' with the current root threshold of keys causes the
+     update to fail
+     """
+     # Load repository with root.json == 1.root.json (available on client)
+     # Signing key: "root", Threshold: 1
+     repository = repo_tool.load_repository(self.repository_directory)
+
+     # Rotate keys and update root: 1.root.json --> 2.root.json
+     # Signing key: "root" (previous) and "root2" (current)
+     # Threshold (for both): 1
+     repository.root.load_signing_key(self.role_keys['root']['private'])
+     repository.root.add_verification_key(self.role_keys['root2']['public'])
+     repository.root.load_signing_key(self.role_keys['root2']['private'])
+     # Remove the previous "root" key from the list of current
+     # verification keys
+     repository.root.remove_verification_key(self.role_keys['root']['public'])
+     repository.writeall()
+
+     # Move staged metadata to "live" metadata
+     shutil.rmtree(os.path.join(self.repository_directory, 'metadata'))
+     shutil.copytree(os.path.join(self.repository_directory, 'metadata.staged'),
+         os.path.join(self.repository_directory, 'metadata'))
+
+     # Intercept 2.root.json and tamper with "root2" (current) key signature
+     root2_path_live = os.path.join(
+         self.repository_directory, 'metadata', '2.root.json')
+     root2 = securesystemslib.util.load_json_file(root2_path_live)
+
+     for idx, sig in enumerate(root2['signatures']):
+       if sig['keyid'] == self.role_keys['root2']['public']['keyid']:
+         sig_len = len(root2['signatures'][idx]['sig'])
+         root2['signatures'][idx]['sig'] = "deadbeef".ljust(sig_len, '0')
+
+     roo2_fobj = tempfile.TemporaryFile()
+     roo2_fobj.write(tuf.repository_lib._get_written_metadata(root2))
+     securesystemslib.util.persist_temp_file(roo2_fobj, root2_path_live)
+
+     # Update 1.root.json -> 2.root.json
+     # Signature verification with current keys should fail because we replaced
+     with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
+       self.repository_updater.refresh()
+
+     for mirror_url, mirror_error in six.iteritems(cm.exception.mirror_errors):
+       self.assertTrue(mirror_url.endswith('/2.root.json'))
+       self.assertTrue(isinstance(mirror_error,
+           securesystemslib.exceptions.BadSignatureError))
+
+     # Assert that the current 'root.json' on the client side is the verified one
+     self.assertTrue(filecmp.cmp(
+       os.path.join(self.repository_directory, 'metadata', '1.root.json'),
+       os.path.join(self.client_metadata_current, 'root.json')))
+
+
+
+
+
+
+
   def test_root_rotation_full(self):
     """Test that a client whose root is outdated by multiple versions and who
     has none of the latest nor next-to-latest root keys can still update and
@@ -340,22 +402,26 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     shutil.copytree(os.path.join(self.repository_directory, 'metadata.staged'),
                     os.path.join(self.repository_directory, 'metadata'))
 
-    try:
+    with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
       self.repository_updater.refresh()
 
-    except tuf.exceptions.NoWorkingMirrorError as exception:
-      for mirror_url, mirror_error in six.iteritems(exception.mirror_errors):
-        url_prefix = self.repository_mirrors['mirror1']['url_prefix']
-        url_file = os.path.join(url_prefix, 'metadata', '2.root.json')
-
-        # Verify that '2.root.json' is the culprit.
-        self.assertEqual(url_file.replace('\\', '/'), mirror_url)
-        self.assertTrue(isinstance(mirror_error,
+    for mirror_url, mirror_error in six.iteritems(cm.exception.mirror_errors):
+      self.assertTrue(mirror_url.endswith('/2.root.json'))
+      self.assertTrue(isinstance(mirror_error,
           securesystemslib.exceptions.BadSignatureError))
 
+    # Assert that the current 'root.json' on the client side is the verified one
+    self.assertTrue(filecmp.cmp(
+        os.path.join(self.repository_directory, 'metadata', '1.root.json'),
+        os.path.join(self.client_metadata_current, 'root.json')))
 
 
-  def test_root_rotation_unmet_threshold(self):
+
+
+  def test_root_rotation_unmet_last_version_threshold(self):
+    """Test that client detects a root.json version that is not signed
+     by a previous threshold of signatures """
+
     repository = repo_tool.load_repository(self.repository_directory)
 
     # Add verification keys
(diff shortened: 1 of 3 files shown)


tuf/client/updater.py
```diff
@@ -1379,12 +1379,50 @@ def verify_target_file(target_file_object):
...
+      # threshold of keys listed in the current root role, therefore skip
+      # keys with a keyid that is not listed in the current root role.
+      if keyid not in keyids:
+        continue
+
+      key = keys[keyid]
+      # The ANYKEY_SCHEMA check in verify_signature expects the keydict to
+      # include a keyid
+      key['keyid'] = keyid
+      valid_sig = securesystemslib.keys.verify_signature(key, signature, signed)
```
trishankatdatadog, line 1399: Should we bail on a single exception here? Feels like we need a try-catch...

tuf/client/updater.py
```diff
@@ -1379,12 +1379,50 @@ def verify_target_file(target_file_object):
 
 
 
-  def _verify_uncompressed_metadata_file(self, metadata_file_object,
+  def _verify_root_sigs(self, signable):
```
trishankatdatadog: I think we need a better name, like `_self_verify_root_sigs()`, because it wasn't clear where the keys are coming from (the new root metadata file itself, which seems dangerous until you consider that this should be called only _after_ verifying that new follows from old, then you check that new follows from new).

tuf/client/updater.py
```diff
@@ -1448,6 +1486,14 @@ def _verify_uncompressed_metadata_file(self, metadata_file_object,
     if not valid:
       raise securesystemslib.exceptions.BadSignatureError(metadata_role)
 
+    # For root metadata, verify the downloaded root metadata object with the
+    # new threshold of new signatures contained within the downloaded root
+    # metadata object
+    if valid and metadata_role == 'root':
+      valid = self._verify_root_sigs(metadata_signable)
```
trishankatdatadog: I also wouldn't call this function inside _verify_metadata_file(), because there is too much magic. Best to call it explicitly in the _update_root_metadata() code IMHO.

tuf/client/updater.py
```diff
@@ -1811,9 +1839,6 @@ def _update_metadata(self, metadata_role, upperbound_filelength, version=None):
     updated_metadata_object = metadata_signable['signed']
     current_metadata_object = self.metadata['current'].get(metadata_role)
 
-    self._verify_root_chain_link(metadata_role, current_metadata_object,
```
trishankatdatadog, line 1805: Why did we remove this code again?

tests/test_updater_root_rotation_integration.py
```diff
@@ -219,6 +219,80 @@ def test_root_rotation(self):
 
 
 
+  def test_BUG_ROOT_NOT_VERIFIED_WITH_CURRENT_KEYIDS_AND_THRESHOLD(self):
```
trishankatdatadog: Hold on: according to the documentation, this test should fail. Why doesn't it?

tests/test_updater_root_rotation_integration.py
```diff
@@ -413,8 +495,112 @@ def test_root_rotation_unmet_threshold(self):
...
+        securesystemslib.exceptions.BadSignatureError))
+
+    # Assert that the current 'root.json' on the client side is the verified one
+    self.assertTrue(filecmp.cmp(
+        os.path.join(self.repository_directory, 'metadata', '2.root.json'),
+        os.path.join(self.client_metadata_current, 'root.json')))
+
+
+
+  def test_root_rotation_discard_untrusted_version(self):
```
trishankatdatadog, line 541: Lovely test.

tests/test_updater_root_rotation_integration.py
```diff
@@ -413,8 +495,112 @@ def test_root_rotation_unmet_threshold(self):
...
+        securesystemslib.exceptions.BadSignatureError))
+
+    # Assert that the current 'root.json' on the client side is the verified one
+    self.assertTrue(filecmp.cmp(
+        os.path.join(self.repository_directory, 'metadata', '2.root.json'),
+        os.path.join(self.client_metadata_current, 'root.json')))
+
+
+
+  def test_root_rotation_unmet_new_threshold(self):
```
trishankatdatadog, line 495: Lovely test, but there's some overlap between this test and the next. Maybe clarify that there is at least one successful intermediate rotation involved or something.

tests/test_updater_root_rotation_integration.py
```diff
@@ -219,6 +219,80 @@ def test_root_rotation(self):
...
+       os.path.join(self.repository_directory, 'metadata', '1.root.json'),
+       os.path.join(self.client_metadata_current, 'root.json')))
+
+
+
+
+
+
+
   def test_root_rotation_full(self):
```
trishankatdatadog, line 282: Hold on: how does this test full rotation? We're still at version 1 of root at the end.

trishankatdatadog (changes requested):

tuf/client/updater.py
joshuagl: Agreed, this does make `_verify_metadata_file()` much more magical. I opted to call it inside `_verify_metadata_file()` because this enable us to perform the check before the tempfile is persisted, and makes it easier to perform the check per download mirror.

If we do the check in `_update_root_metadata()` we need to clean up a file from the clients metadata directory and, perhaps less importantly, can only perform the check on a single file returned by `_update_root_metadata()`

tuf/client/updater.py
joshuagl: It's just duplicating steps that have already been performed in `_verify_metadata_file` and not performing those steps until after the tempfile has been persisted.

tests/test_updater_root_rotation_integration.py
joshuagl: Good catch, the doc string needs updating now that the fix is implemented.

tuf/client/updater.py
jku: `_verify_root_self_signed()`?

tuf/client/updater.py
jku: +1. I think the point of never writing unverified files is _critical_ and should probably be encoded in comments or docs... If we do not adhere to that it's almost impossible to ensure that our state of metadata stays valid.

tests/test_updater_root_rotation_integration.py
jku: Are you mixing up tests? `test_root_rotation_full()` seems fine to me and was not touched in this commit.

tests/test_updater_root_rotation_integration.py
```diff
@@ -342,22 +415,30 @@ def test_root_rotation_missing_keys(self):
...
-        url_file = os.path.join(url_prefix, 'metadata', '2.root.json')
+    for mirror_url, mirror_error in six.iteritems(cm.exception.mirror_errors):
+      url_prefix = self.repository_mirrors['mirror1']['url_prefix']
+      url_file = os.path.join(url_prefix, 'metadata', '2.root.json')
 
-        # Verify that '2.root.json' is the culprit.
-        self.assertEqual(url_file.replace('\\', '/'), mirror_url)
-        self.assertTrue(isinstance(mirror_error,
+      # Verify that '2.root.json' is the culprit.
+      self.assertEqual(url_file.replace('\\', '/'), mirror_url)
```
jku: I know this is how most tests do this but replacing the 3 lines with this should do the job and makes the test more readable:
`self.assertTrue(mirror_url.endswith("/2.root.json"))`

(same comment applies to other modified/added tests)

tests/test_updater_root_rotation_integration.py
```diff
@@ -413,8 +494,112 @@ def test_root_rotation_unmet_threshold(self):
...
-        self.repository_updater.refresh)
+    with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
+      self.repository_updater.refresh()
+
+    for mirror_url, mirror_error in six.iteritems(cm.exception.mirror_errors):
+      url_prefix = self.repository_mirrors['mirror1']['url_prefix']
+      url_file = os.path.join(url_prefix, 'metadata', '3.root.json')
+
+      # Verify that '2.root.json' is the culprit.
+      self.assertEqual(url_file.replace('\\', '/'), mirror_url)
```
jku, line 506: same comment as above

tuf/client/updater.py
joshuagl: good suggestion, I like it

tuf/client/updater.py
joshuagl: `_get_metadata_file` has some logic to collect all exceptions and pack them into `NoWorkingMirrorError`, so I *think* this just continues the existing pattern?

tests/test_updater_root_rotation_integration.py
joshuagl: Thanks for this suggestion, I've pushed a patch implementing it: adbb5fe

tests/test_updater_root_rotation_integration.py
```diff
@@ -413,8 +481,103 @@ def test_root_rotation_unmet_threshold(self):
...
     # The following refresh should fail because root must be signed by the
     # previous self.role_keys['root'] key, which wasn't loaded.
-    self.assertRaises(tuf.exceptions.NoWorkingMirrorError,
-        self.repository_updater.refresh)
+    with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
+      self.repository_updater.refresh()
+
+    for mirror_url, mirror_error in six.iteritems(cm.exception.mirror_errors):
+      # Verify that '2.root.json' is the culprit.
+      self.assertTrue(mirror_url.endswith('/3.root.json'))
```
jku: Now that the code is simpler it's easy to see an issue with the comment :)
Could even drop the comment: code seems fairly obvious to me...

tests/test_updater_root_rotation_integration.py
```diff
@@ -413,8 +481,103 @@ def test_root_rotation_unmet_threshold(self):
...
+
+
+   # The following refresh should fail because root must be signed by the
+   # current self.role_keys['root3'] key, which wasn't loaded.
+    with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
+      self.repository_updater.refresh()
+
+    for mirror_url, mirror_error in six.iteritems(cm.exception.mirror_errors):
+      # Verify that '2.root.json' is the culprit.
+      self.assertTrue(mirror_url.endswith('/3.root.json'))
```
jku: comment clashes with code here too

tests/test_updater_root_rotation_integration.py
joshuagl: 🤦 thank you for the diligent review. Fixed in 6eaf194

tuf/client/updater.py
trishankatdatadog: Also, we want to make sure that other functions that depend on this function do the same thing. Let's leave the magic alone for now, but document clearly why it is there, and not somewhere else.

jku (approved):

joshuagl:
I believe I've addressed all review comments. Once the PR is approved I'd like to squash some of the changes addressing review comments into the patches that introduced the changes.

trishankatdatadog (approved):
LGTM

Issue #1080, opened 2020-07-16
Updater: Way to disable hash prefix for consistent_snapshot


jku:
Currently if the repository is `consistent_snapshot`, Updater will prefix the target filename with the hash when constructing the download URL. In the Warehouse case this is not wanted (warehouse target file paths are "consistent", even if the filenames are not).

This is an example of a working URL:

https://files.pythonhosted.org/packages/ca/ab/5e004afa025a6fb640c6e983d4983e6507421ff01be224da79ab7de7a21f/Django-3.0.8-py3-none-any.whl

But currently Updater will try to download

https://files.pythonhosted.org/packages/ca/ab/5e004afa025a6fb640c6e983d4983e6507421ff01be224da79ab7de7a21f/caab5e004afa025a6fb640c6e983d4983e6507421ff01be224da79ab7de7a21f.Django-3.0.8-py3-none-any.whl

If there is no way to figure this out automatically, then there should probably be a way to tell Updater that _I Know What I'm Doing, don't prefix filenames_.

MVrachev:
I am not sure I understand that:

> `warehouse target file paths are "consistent", even if the filenames are not).`

if the filenames are not consistent, then the file path won't be too?

Also, in the spec in section 5.2 it's written:
```
If consistent snapshots are not used (see Section 7), 
then the filename used to download the target file is of the fixed form FILENAME.EXT (e.g., foobar.tar.gz).
Otherwise, the filename is of the form HASH.FILENAME.EXT 
(e.g., c14aeb4ac9f4a8fc0d83d12482b9197452f6adf3eb710e3b1e2b79e8d14cb681.foobar.tar.gz), 
where HASH is one of the hashes of the targets file listed in the targets metadata file found earlier in step 4. 
In either case, the client MUST write the file to non-volatile storage as FILENAME.EXT.
```

so, it seems that they didn't think of a case when you will use `consistent_snapshot` and the naming `FILENAME.EXT `

joshuagl:
> I am not sure I understand that:
> 
> > `warehouse target file paths are "consistent", even if the filenames are not).`
> 
> if the filenames are not consistent, then the file path won't be too?

[...]

PR #1102, opened 2020-08-06
Add a way to disable hash prefix when using consistent_snapshot

MVrachev:
**Fixes issue #**: https://github.com/theupdateframework/tuf/issues/1080

Currently, if the repository is consistent_snapshot,
Updater will prefix the target filename with the hash
when constructing the download URL.
For some adopters of TUF (like Warehouse) this is not wanted
(warehouse target file paths are "consistent",
even if the filenames are not).

For example, Warehouse doesn't follow what tuf
(the reference implementation and specification) advice for naming
consistent filenames, which is to prefix the filename with the hash
of the files contents.
However, the target filenames it does use are consistent,
only the hash is part of the target's file path
not the target's file name.

requirements**:

### DIFF
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1291,6 +1291,39 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     self.repository_updater.download_target(targetinfo2,
                                             destination_directory)
 
+    # Checks if the file has been successfully downloaded
+    download_filepath = os.path.join(destination_directory, target_filepath2)
+    self.assertTrue(os.path.exists(download_filepath))
+
+    # Removes the file so that it can be downloaded again in the next test
+    os.remove(download_filepath)
+
+    # Test downloading with consistent snapshot enabled, but without adding
+    # the hash of the file as a prefix to its name.
+
+    file1_path = targetinfo2['filepath']
+    file1_hashes = securesystemslib.util.get_file_hashes(
+        os.path.join(self.repository_directory, 'targets', file1_path),
+        hash_algorithms=['sha256', 'sha512'])
+
+    # Currently in the repository directory, those three files exists:
+    # "file1.txt", "<sha256_hash>.file1.txt" and "<sha512_hash>.file1.txt"
+    # where both sha256 and sha512 hashes are for file file1.txt.
+    # Remove the files with the hash digest prefix to ensure that
+    # the served target file is not prefixed.
+    os.remove(os.path.join(self.repository_directory, 'targets',
+        file1_hashes['sha256'] + '.' + file1_path))
+    os.remove(os.path.join(self.repository_directory, 'targets',
+        file1_hashes['sha512'] + '.' + file1_path))
+
+
+    self.repository_updater.download_target(targetinfo2,
+                                            destination_directory,
+                                            prefix_filename_with_hash=False)
+
+    # Checks if the file has been successfully downloaded
+    self.assertTrue(os.path.exists(download_filepath))
+
     # Test for a destination that cannot be written to (apart from a target
     # file that already exists at the destination) and which raises an
     # exception.
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1323,7 +1323,8 @@ class Updater(object):
 
 
 
-  def _get_target_file(self, target_filepath, file_length, file_hashes):
+  def _get_target_file(self, target_filepath, file_length, file_hashes,
+      prefix_filename_with_hash):
     """
     <Purpose>
       Non-public method that safely (i.e., the file length and hash are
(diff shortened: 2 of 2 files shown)


MVrachev:
I decided to add `use_existing_hash_prefixes` argument to the `updater.download_target()` function instead as an argument to the `__init__` function for the Updater because `updater.download_target()` is really the only place where this argument is needed.
Plus, this made it easier for testing.

tests/test_updater.py
```diff
@@ -1303,6 +1303,22 @@ def test_6_download_target(self):
...
+    os.remove(download_filepath)
+
+    # Test downloading with consistent snapshot enabled, but without adding
+    # the hash of the file as a prefix to its name.
+    self.repository_updater.download_target(targetinfo2,
+                                            destination_directory,
+                                            use_existing_hash_prefixes=True)
+
+    # Checks if the file has been successfully downloaded
+    self.assertTrue(os.path.exists(download_filepath))
```
jku: How does the test work without any changes in the test data? Does the non-prefixed version exist on fileserver already?

If both versions exist on the server, maybe we should make sure only one does (for at least one download with each setting) -- otherwise we don't really know what we're downloading... or am I being overly cautious here?

tuf/client/updater.py
```diff
@@ -3192,7 +3197,8 @@ def updated_targets(self, targets, destination_directory):
 
 
 
-  def download_target(self, target, destination_directory):
+  def download_target(self, target, destination_directory,
+      use_existing_hash_prefixes=False):
```
jku: I think putting the argument in this function (as opposed to Updater initializer) is fine. The name feels a bit off however: 
* `use_existing_hash_prefixes` might not be even technically correct (as Updater does not really know _why_ hash prefixed names are not used)
* the name does not help the Updater user understand what the argument does

The second point is the important one. Would `prefix_filename_with_hash=True` be too long?

tuf/client/updater.py
```diff
@@ -3209,6 +3215,11 @@ def download_target(self, target, destination_directory):
       destination_directory:
         The directory to save the downloaded target file.
 
+      use_existing_hash_prefixes:
+        Whether to prefix the targets file names with their hash when using
+        consistent snapshot.
+        Default is False.
```
jku: Can you make it clear that the prefix is added to the _download query_ made to the server, not to the filename that's saved to destination_directory. 

Mentioning "consistent snapshot" is good but I fear most readers will not understand what it means... How about adding another sentence like this: 

> This should be set to True when the served target filenames are not prefixed with hashes (in this case the server uses other means to ensure snapshot consistency).

jku (changes requested):
Thanks! I have comments mostly on phrasing plus a question on the test implementation

tests/test_updater.py
MVrachev: No, it's a good question and I should leave a comment somewhere about this.

In the begging when a repository is loaded and copied (lines 1280 - 1296) there are three files:
`file1.txt`, `file2.txt`, and `file3.txt`. 

So, when we download without adding the hash of the file as a prefix we are downloading `file1.txt` which is available from the begging.

tuf/client/updater.py
MVrachev: No, it's not too long, on the contrary - it's one character shorter than the current one. :D

I agree that this argument name doesn't give good information for the user. I have been thinking a lot about how to call it.
`prefix_filename_with_hash` name is okay.

tuf/client/updater.py
MVrachev: A good point.
When you pointed it out I remembered I was a little confused about what file is saved on the backend with the hash prefix or without it.

tests/test_updater.py
jku: :+1: so then the question is, should we be stricter in the test? I'm thinking of something like us mixing up True and False in handling this: this test would not notice that...

MVrachev:
I updated my pr adding the necessary explanations in the test and changing the variable name.

tests/test_updater.py
```diff
@@ -1291,6 +1291,39 @@ def test_6_download_target(self):
...
+    file1_path = targetinfo2['filepath']
+    file1_hashes = securesystemslib.util.get_file_hashes(
+        os.path.join(self.repository_directory, 'targets', file1_path),
+        hash_algorithms=['sha256', 'sha512'])
+
+    # Currently in the repository directory, those three files exists:
+    # "file1.txt", "<sha256_hash>.file1.txt" and "<sha512_hash>.file1.txt"
+    # where both sha256 and sha512 hashes are for file file1.txt.
+    # I will remove the files with the hash digest prefix to ensure that
+    # the served target file is not prefixed.
```
joshuagl: Avoid talking in the first person in code comments.
```suggestion
    # Remove the files with the hash digest prefix to ensure that
    # the served target file is not prefixed.
```

joshuagl (approved):
This looks like a nice clean change, thanks @MVrachev.

@jku could you take another look to ensure it suits your needs and your prior comments have been addressed?

jku (approved):
LGTM, thanks Martin.

(Btw, the Travis timeout happened in a place where we get server output: It is using a port that is used already! Good info but not related to this bug)

joshuagl:
Thanks for the patches @MVrachev and for the review @jku

joshuagl:
Thanks for the patches @MVrachev and for the review @jku

PR #1103, opened 2020-08-06
PoC(tap13): User Selection of the Top-Level Target Files Through Mapping Metadata

mnm678:
This pr adds support for [TAP 13](https://github.com/theupdateframework/taps/pull/118/files). It adds support for targets mapping metadata that is defined by the client. This mapping metadata allows the client to specify which targets metadata file on the repository they would like to use as top-level targets. This mechanism allows repository managers to split a repository into multiple namespaces and for clients to specify which of these namespaces to trust.

### DIFF
--- a/tests/repository_data/client/targets_map.json
+++ b/tests/repository_data/client/targets_map.json
@@ -1,0 +1,16 @@
+{
+ "targets_filename": "role1",
+ "keys":{
+  "c8022fa1e9b9cb239a6b362bbdffa9649e61ad2cb699d2e4bc4fdf7930a0e64a": {
+   "keyid_hash_algorithms": [
+    "sha256",
+    "sha512"
+   ],
+   "keytype": "ed25519",
+   "keyval": {
+    "public": "fcf224e55fa226056adf113ef1eb3d55e308b75b321c8c8316999d8c4fd9e0d9"
+   },
+   "scheme": "ed25519"
+  }
+ }
+}
--- a/tests/repository_data/client/targets_map.json
+++ b/tests/repository_data/client/targets_map.json
@@ -1,5 +1,5 @@
 {
- "targets_filename": "role1",
+ "targets_rolename": "role1",
  "keys":{
   "c8022fa1e9b9cb239a6b362bbdffa9649e61ad2cb699d2e4bc4fdf7930a0e64a": {
    "keyid_hash_algorithms": [
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -186,6 +186,10 @@ class MultiRepoUpdater(object):
       The path of the map file.  The map file is needed to determine which
       repositories to query given a target file.
 
+    targets_map_filename:
+      The path of the targets map file. This targets map file
+      will be used by all repositories.
+
   <Exceptions>
     securesystemslib.exceptions.FormatError, if the map file is improperly
     formatted.
@@ -595,6 +599,9 @@ class Updater(object):
     self.repository_name:
       The name of the updater instance.
 
+    self.targets_map_file:
+      The contents of the targets map file.
+
   <Updater Methods>
     refresh():
       This method downloads, verifies, and loads metadata for the top-level
@@ -673,6 +680,10 @@ class Updater(object):
                                           'metadata_path': 'metadata',
                                           'targets_path': 'targets',
                                           'confined_target_dirs': ['']}}
+      targets_map_file:
+        The name of the targets map file. If one is not provided,
+        self.targets_map_file will be set to None and the top-level
+        targets metadata from the repository will be used.
 
     <Exceptions>
       securesystemslib.exceptions.FormatError:
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -232,14 +232,13 @@ class MultiRepoUpdater(object):
     #  }
     self.repository_names_to_mirrors = self.map_file['repositories']
 
-    self.targets_map_filename = None
-
+    # If there is a targets_map_file, set self.targets_map_filename.
+    # This map file will be used with all repositories in place of the
+    # top level targets file.
     if targets_map_filename is not None:
-      # Is 'targets_map_file' a path?  If not, raise
-      # 'securesystemslib.exceptions.FormatError'.  The actual content of the map
-      # file is validated later on in this method.
       securesystemslib.formats.PATH_SCHEMA.check_match(targets_map_filename)
-      self.targets_map_filename = targets_map_filename
+
+    self.targets_map_filename = targets_map_filename
 
 
   def get_valid_targetinfo(self, target_filename, match_custom_field=True):
@@ -641,7 +640,8 @@ class Updater(object):
     http://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables
   """
 
-  def __init__(self, repository_name, repository_mirrors, targets_map_file = None):
+  def __init__(self, repository_name, repository_mirrors,
+      targets_map_filename=None):
     """
     <Purpose>
       Constructor.  Instantiating an updater object causes all the metadata
@@ -680,7 +680,7 @@ class Updater(object):
                                           'metadata_path': 'metadata',
                                           'targets_path': 'targets',
                                           'confined_target_dirs': ['']}}
-      targets_map_file:
+      targets_map_filename:
         The name of the targets map file. If one is not provided,
         self.targets_map_file will be set to None and the top-level
         targets metadata from the repository will be used.
@@ -715,14 +715,13 @@ class Updater(object):
     self.mirrors = repository_mirrors
     self.targets_map_file = None
 
-    if targets_map_file is not None:
-      # Is 'targets_map_file' a path?  If not, raise
-      # 'securesystemslib.exceptions.FormatError'.  The actual content of the map
-      # file is validated later on in this method.
-      securesystemslib.formats.PATH_SCHEMA.check_match(targets_map_file)
+    # If targets_map_filename is provided, load the targets_map_file.
+    if targets_map_filename is not None:
+      securesystemslib.formats.PATH_SCHEMA.check_match(targets_map_filename)
 
       try:
-        self.targets_map_file = securesystemslib.util.load_json_file(targets_map_file)
+        self.targets_map_file = securesystemslib.util.load_json_file(
+            targets_map_filename)
 
       except (securesystemslib.exceptions.Error) as e:
         raise tuf.exceptions.Error('Cannot load the targets map file: ' + str(e))
@@ -859,7 +858,9 @@ class Updater(object):
     # Save and construct the full metadata path.
     metadata_directory = self.metadata_directory[metadata_set]
     # For top-level targets, the targets map file may overwrite the
-    # targets metadata on the repository
+    # targets metadata on the repository. If this is the case, ignore
+    # the 'targets.json' file on the repository and instead load
+    # the targets metadata indicated by the targets map file.
     if metadata_role == 'targets' and self.targets_map_file is not None:
       metadata_filename = self.targets_map_file['targets_filename'] + '.json'
     else:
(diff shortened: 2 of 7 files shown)


tests/repository_data/client/targets_map.json
```diff
@@ -0,0 +1,16 @@
...
+    "sha512"
+   ],
+   "keytype": "ed25519",
+   "keyval": {
+    "public": "fcf224e55fa226056adf113ef1eb3d55e308b75b321c8c8316999d8c4fd9e0d9"
+   },
+   "scheme": "ed25519"
+  }
+ }
+}
```
lukpueh, line 16: Is there an agreement about the targets map metadata format? AFAIR TAP 12 still misses the definition.

tuf/keydb.py
```diff
@@ -111,6 +116,44 @@ def create_keydb_from_root_metadata(root_metadata, repository_name='default'):
...
+        # Although keyid duplicates should *not* occur (unique dict keys), log a
+        # warning and continue.  However, 'key_dict' may have already been
+        # adding to the keydb elsewhere.
+        except tuf.exceptions.KeyAlreadyExistsError as e: # pragma: no cover
+          logger.warning(e)
+          continue
+
+      else:
+        logger.warning('Root Metadata file contains a key with an invalid keytype.')
+
```
lukpueh: Please try not to copy-paste huge blocks like these (especially if they include dubious `keyid_hash_algorithm`-code that is prone to change .. see https://github.com/secure-systems-lab/securesystemslib/pull/227). :)

Why not just merge the two dictionaries `targets_map_file['keys']` and `root_metadata['keys']` into a new dict and then iterate over that one? We need to be careful though and define (and implement) the desired behavior for when  there are matching keyids in both dicts.

tuf/roledb.py
```diff
@@ -76,7 +76,7 @@
 TOP_LEVEL_ROLES = ['root', 'targets', 'snapshot', 'timestamp']
 
 
-def create_roledb_from_root_metadata(root_metadata, repository_name='default'):
+def create_roledb_from_root_metadata(root_metadata, repository_name='default', targets_map_file = None):
```
lukpueh: Style nit: Please remove space around operator.

tuf/keydb.py
```diff
@@ -60,7 +60,7 @@
 _keydb_dict['default'] = {}
 
 
-def create_keydb_from_root_metadata(root_metadata, repository_name='default'):
+def create_keydb_from_root_metadata(root_metadata, repository_name='default', targets_map_file = None):
```
lukpueh: Style nit: Please remove space around operator.

tuf/roledb.py
```diff
@@ -140,6 +144,11 @@ def create_roledb_from_root_metadata(root_metadata, repository_name='default'):
       roleinfo['previous_keyids'] = roleinfo['keyids']
       roleinfo['previous_threshold'] = roleinfo['threshold']
 
+    if rolename == 'targets' and targets_map_file is not None:
```
lukpueh: Minor nit: Could be an `elif`, right?

tuf/client/updater.py
```diff
@@ -228,6 +232,14 @@ def __init__(self, map_file):
     #  }
     self.repository_names_to_mirrors = self.map_file['repositories']
 
+    self.targets_map_filename = None
+
+    if targets_map_filename is not None:
+      # Is 'targets_map_file' a path?  If not, raise
+      # 'securesystemslib.exceptions.FormatError'.  The actual content of the map
+      # file is validated later on in this method.
```
lukpueh: Please don't copy those comments. They are already everywhere (not your fault) and IMO not particularly helpful.

tuf/client/updater.py
```diff
@@ -228,6 +232,14 @@ def __init__(self, map_file):
...
+      securesystemslib.formats.PATH_SCHEMA.check_match(targets_map_filename)
+      self.targets_map_filename = targets_map_filename
```
lukpueh: Minor style suggestion:
```suggestion
    if targets_map_filename is not None:
      securesystemslib.formats.PATH_SCHEMA.check_match(targets_map_filename)
    self.targets_map_filename = targets_map_filename
```

tuf/client/updater.py
```diff
@@ -626,7 +641,7 @@ class Updater(object):
     http://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables
   """
 
-  def __init__(self, repository_name, repository_mirrors):
+  def __init__(self, repository_name, repository_mirrors, targets_map_file = None):
```
lukpueh: Style nit: Please remove space around operator.

tuf/client/updater.py
```diff
@@ -665,6 +680,10 @@ def __init__(self, repository_name, repository_mirrors):
                                           'metadata_path': 'metadata',
                                           'targets_path': 'targets',
                                           'confined_target_dirs': ['']}}
+      targets_map_file:
+        The name of the targets map file. If one is not provided,
+        self.targets_map_file will be set to None and the top-level
+        targets metadata from the repository will be used.
```
lukpueh, line 686: I find it confusing that the argument `targets_map_file` has the filename and the attribute `targets_map_file` has its contents. Would you mind renaming one other the other? Above in `MultiRepoUpdater` you use `targets_map_filename` for the parameter.

tuf/client/updater.py
```diff
@@ -694,6 +713,23 @@ def __init__(self, repository_name, repository_mirrors):
     # Save the validated arguments.
     self.repository_name = repository_name
     self.mirrors = repository_mirrors
+    self.targets_map_file = None
+
+    if targets_map_file is not None:
+      # Is 'targets_map_file' a path?  If not, raise
+      # 'securesystemslib.exceptions.FormatError'.  The actual content of the map
+      # file is validated later on in this method.
```
lukpueh: See comment about this comment block above.

lukpueh (commented):
Thanks for taking a stab at TAP13, @mnm678! Could you please take a look why the tests fail>

Also, my review is mostly about programming- and code-style. I admit that I had a hard time to know if  the patch indeed implements TAP 13. Would you mind adding a few more code comments to make the intentions clearer? More details in the commit message would also help.  :)

One more style-nit: Please try to stay within 80 characters per line.

tests/repository_data/client/targets_map.json
mnm678: There isn't. I based this format on a suggestion from the comments of the TAP 13 pr. It would probably be good to get some more consensus about this or a different format.

tuf/keydb.py
mnm678: I removed the repeated block in the latest commit. When merging the dictionaries, I chose to prioritize keys from root metadata if there is a conflict. With the current implementation's use of sha2-256 for all keyids, this conflict should not occur, but it is something that might require some more thought.

tuf/client/updater.py
```diff
@@ -694,6 +713,22 @@ def __init__(self, repository_name, repository_mirrors):
...
     self.repository_name = repository_name
     self.mirrors = repository_mirrors
+    self.targets_map_file = None
+
+    # If targets_map_filename is provided, load the targets_map_file.
+    if targets_map_filename is not None:
+      securesystemslib.formats.PATH_SCHEMA.check_match(targets_map_filename)
+
+      try:
+        self.targets_map_file = securesystemslib.util.load_json_file(\
```
lukpueh: No need to escape the newline, there is implied line continuation inside parentheses.

tests/repository_data/client/targets_map.json
lukpueh: cc @joshuagl, @trishankatdatadog, @JustinCappos

tuf/keydb.py
```diff
@@ -111,10 +117,21 @@ def create_keydb_from_root_metadata(root_metadata, repository_name='default'):
...
+
+  keys_to_add = root_metadata['keys']
+
+  if targets_map_file is not None:
+    tuf.formats.TARGETS_MAPFILE_SCHEMA.check_match(targets_map_file)
+
+    # If any keyids from root metadata match those from targets_map_metadata,
+    # only the root metadata keys will be added. Add all keys to root_metadata
+    keys_to_add = dict(list(targets_map_file['keys'].items()) +
+        list(root_metadata['keys'].items()))
```
lukpueh: Minor nit: This technique is discouraged (see e.g. https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-in-python-taking-union-o). I suggest to use the copy + update technique.

lukpueh (commented):
Thanks for quickly addressing my comments, @mnm678. The patch looks quite good to me (modulo some minor nits in my inline comments). I will need to take another look to say for sure that this fully implements TAP13. It would be nice if other TAP authors could take a look too (especially in regards to the metadata format). 

Also, please git rid of the merge commit in the PRs commit history.

mnm678:
Thanks @lukpueh . I force pushed to clean up the merge commit. It looks like the linter is failing due to #1115.

jku:
It would be great if commits (and the PR cover letter as well) answered the question "why?", even if the answer is just a single sentence.

I understand the TAP exists but 

1. Once this is committed and no longer part of a PR the context isn't as clear and 
1. Going from no explanation at all to the somewhat academic TAP is big jump

tests/repository_data/client/targets_map.json
jku: Some of the other metadata files refer to other metadata files as well: they use the full name "targets.json" so `targets_filename` should probably do that too

tuf/client/updater.py
```diff
@@ -694,6 +713,22 @@ def __init__(self, repository_name, repository_mirrors):
...
+    # If targets_map_filename is provided, load the targets_map_file.
+    if targets_map_filename is not None:
+      securesystemslib.formats.PATH_SCHEMA.check_match(targets_map_filename)
+
+      try:
+        self.targets_map_file = securesystemslib.util.load_json_file(
+            targets_map_filename)
+
+      except (securesystemslib.exceptions.Error) as e:
+        raise tuf.exceptions.Error('Cannot load the targets map file: ' + str(e))
```
jku, line 732: Please avoid generic exceptions in public API if possible.

Also document the exception (alternatively remove the 'Exceptions' section from the docs if it's known to be outdated).

mnm678: Do you think a FormatError would make more sense here (to indicate that the targets map file specified an invalid filename)? Or something else?

jku: Good question: securesystemslib seems to have a wide variety of errors here (some of which clearly aren't formaterrors) so maybe you can't avoid a generic one. In that case at least document the exceptions that may be raised.

By the way, the check_match() call just before this seems to not be needed as securesystemslib does that (unless the point was to trigger the FormatError specifically before the try-except)

mnm678:
> It would be great if commits (and the PR cover letter as well) answered the question "why?", even if the answer is just a single sentence.

I just force-pushed to update the commit messages.

joshuagl:
As this is a PoC for the TAP process I've converted it to a draft.

tests/repository_data/client/targets_map.json
joshuagl: in the intervening 1 year we've discovered multiple reasons _not_ to use the full filename.

joshuagl (commented):
I'd suggest `"targets_filename"` -> `"targets_rolename"` as we want to move away from being file-centric in the spec

tests/repository_data/client/targets_map.json
```diff
@@ -0,0 +1,16 @@
+{
+ "targets_filename": "role1",
```
mnm678:
```suggestion
 "targets_rolename": "role1",
```

lukpueh:
Now with the [legacy code gone](https://github.com/theupdateframework/python-tuf/blob/develop/docs/1.0.0-ANNOUNCEMENT.md) this PR will need quite a re-write. 

Since most of above review comments and discussion don't seem to be relevant anymore, I suggest to close here and submit a fresh PR for TAP13 ... it should be a lot easier to build on top of the new APIs.

PR #1104, opened 2020-08-11
Log subproceses stdout and stderr in temp files

MVrachev:
**Fixes issues #**: https://github.com/theupdateframework/tuf/issues/1039, https://github.com/theupdateframework/tuf/issues/1119

**This pr replaces pr https://github.com/theupdateframework/tuf/pull/1041.**

Logging the stdout and stderr from the test subprocesses into
temporary files clean the console from unnecessary messages from
the server-side such as "code 404, message File not found" or
"GET" queries.

The unit tests are executed in sequential order and that's why
we can reuse one temporary file object for multiple tests.

I have used this pr to remove some redundant functions, comments, and variables.

For this pr I received great help from @jku!
Thanks!

requirements**:

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -36,8 +36,6 @@ from __future__ import unicode_literals
 import hashlib
 import logging
 import os
-import random
-import subprocess
 import sys
 import unittest
 
@@ -73,15 +71,11 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     self.target_data_length = len(self.target_data)
 
     # Launch a SimpleHTTPServer (serves files in the current dir).
-    self.PORT = random.randint(30000, 45000)
-    self.server_proc = popen_python(['simple_server.py', str(self.PORT)])
-    logger.info('\n\tServer process started.')
-    logger.info('\tServer process id: ' + str(self.server_proc.pid))
-    logger.info('\tServing on port: ' + str(self.PORT))
-    junk, rel_target_filepath = os.path.split(target_filepath)
-    self.url = 'http://localhost:'+str(self.PORT)+'/'+rel_target_filepath
+    self.server_process_handler = utils.TestServerProcess(log=logger)
 
-    utils.wait_for_server('localhost', self.PORT)
+    rel_target_filepath = os.path.basename(target_filepath)
+    self.url = 'http://localhost:' \
+        + str(self.server_process_handler.port) + '/' + rel_target_filepath
 
     # Computing hash of target file data.
     m = hashlib.md5()
@@ -93,11 +87,11 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
   # Stop server process and perform clean up.
   def tearDown(self):
     unittest_toolbox.Modified_TestCase.tearDown(self)
-    if self.server_proc.returncode is None:
-      logger.info('\tServer process '+str(self.server_proc.pid)+' terminated.')
-      self.server_proc.kill()
-      # Drop return values of communicate()
-      self.server_proc.communicate()
+
+    # Logs stdout and stderr from the server subprocess and then it
+    # kills it and closes the temp file used for logging.
+    self.server_process_handler.clean()
+
     self.target_fileobj.close()
 
 
@@ -176,14 +170,17 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
                       download_file,
                       self.random_string(), self.target_data_length)
 
+    url = 'http://localhost:' \
+        + str(self.server_process_handler.port) + '/' + self.random_string()
     self.assertRaises(requests.exceptions.HTTPError,
                       download_file,
-                      'http://localhost:' + str(self.PORT) + '/' + self.random_string(),
+                      url,
                       self.target_data_length)
-
+    url1 = 'http://localhost:' \
+      + str(self.server_process_handler.port + 1) + '/' + self.random_string()
     self.assertRaises(requests.exceptions.ConnectionError,
                       download_file,
-                      'http://localhost:' + str(self.PORT+1) + '/' + self.random_string(),
+                      url1,
                       self.target_data_length)
 
     # Specify an unsupported URI scheme.
@@ -258,27 +255,30 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     # 3: run with an HTTPS certificate with an unexpected hostname
     # 4: run with an HTTPS certificate that is expired
     # Be sure to offset from the port used in setUp to avoid collision.
-    port1 = str(self.PORT + 1)
-    port2 = str(self.PORT + 2)
-    port3 = str(self.PORT + 3)
-    port4 = str(self.PORT + 4)
-    good_https_server_proc = popen_python(
-        ['simple_https_server.py', port1, good_cert_fname])
-    good2_https_server_proc = popen_python(
-        ['simple_https_server.py', port2, good2_cert_fname])
-    bad_https_server_proc = popen_python(
-        ['simple_https_server.py', port3, bad_cert_fname])
-    expd_https_server_proc = popen_python(
-        ['simple_https_server.py', port4, expired_cert_fname])
 
-    for port in range(self.PORT + 1, self.PORT + 5):
-      utils.wait_for_server('localhost', port)
+    port1 = self.server_process_handler.port + 1
+    port2 = self.server_process_handler.port + 2
+    port3 = self.server_process_handler.port + 3
+    port4 = self.server_process_handler.port + 4
 
-    relative_target_fpath = os.path.basename(target_filepath)
-    good_https_url = 'https://localhost:' + port1 + '/' + relative_target_fpath
-    good2_https_url = good_https_url.replace(':' + port1, ':' + port2)
-    bad_https_url = good_https_url.replace(':' + port1, ':' + port3)
-    expired_https_url = good_https_url.replace(':' + port1, ':' + port4)
+    good_https_server_handler = utils.TestServerProcess(log=logger,
+        server='simple_https_server.py', port=port1,
+        extra_cmd_args=[good_cert_fname])
+    good2_https_server_handler = utils.TestServerProcess(log=logger,
+        server='simple_https_server.py', port=port2,
+        extra_cmd_args=[good2_cert_fname])
+    bad_https_server_handler = utils.TestServerProcess(log=logger,
+        server='simple_https_server.py', port=port3,
+        extra_cmd_args=[bad_cert_fname])
+    expd_https_server_handler = utils.TestServerProcess(log=logger,
+        server='simple_https_server.py', port=port4,
+        extra_cmd_args=[expired_cert_fname])
+
+    suffix = '/' +  os.path.basename(target_filepath)
+    good_https_url = 'https://localhost:' + str(port1) + suffix
+    good2_https_url = 'https://localhost:' + str(port2) + suffix
+    bad_https_url = 'https://localhost:' + str(port3) + suffix
+    expired_https_url = 'https://localhost:' + str(port4) + suffix
 
     # Download the target file using an HTTPS connection.
 
@@ -354,36 +354,15 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
       download.unsafe_download(good2_https_url, target_data_length).close()
 
     finally:
-      for proc in [
-          good_https_server_proc,
-          good2_https_server_proc,
-          bad_https_server_proc,
-          expd_https_server_proc]:
-        if proc.returncode is None:
-          logger.info('Terminating server process ' + str(proc.pid))
-          proc.kill()
-          # drop return values
-          proc.communicate()
+      for proc_handler in [
+          good_https_server_handler,
+          good2_https_server_handler,
+          bad_https_server_handler,
+          expd_https_server_handler]:
 
-
-
-
-
-# TODO: Move this to a common test module (tests/common.py?)
-#       and strip it test_proxy_use.py and test_download.py.
-def popen_python(command_arg_list):
-  """
-  Run subprocess.Popen() to produce a process running a Python interpreter.
-  Uses the same Python interpreter that the current process is using, via
-  sys.executable.
-  """
-  assert sys.executable, 'Test cannot function: unable to determine ' \
-      'current Python interpreter via sys.executable.'
-
-  return subprocess.Popen(
-      [sys.executable] + command_arg_list, stderr=subprocess.PIPE)
-
-
+        # Logs stdout and stderr from the server subprocess and then it
+        # kills it and closes the temp file used for logging.
+        proc_handler.clean()
 
 
 
--- a/tests/test_proxy_use.py
+++ b/tests/test_proxy_use.py
@@ -37,10 +37,8 @@ from __future__ import unicode_literals
 
 import logging
 import os
-import random
-import subprocess
-import sys
 import unittest
+import sys
 
 import tuf
 import tuf.download as download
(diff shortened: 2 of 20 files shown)


MVrachev:
On a successful test run the console logs look like this: 
image

I wanted to simulate an error in the subprocess and I changed the name of the server script from `simple_server.py` to `tsimple_server.py` and here how the error was logged: 
image

jku:
Two comments from a quick test:
* the child process output does get buffered for me (so `python3 test_updater.py TestUpdater.test_4_refresh` does not notice the 404s): we should start the server using  "python -u simple_server.py" to prevent buffering
* the server output should not be warning because warnings are printed by default. Also we don't really know what the messages are and mostly do not care. info or debug level maybe?

MVrachev:
> * the child process output does get buffered for me (so `python3 test_updater.py TestUpdater.test_4_refresh` does not notice the 404s): we should start the server using  "python -u simple_server.py" to prevent buffering

Yes, I checked that and you are right.
The child process buffers.

The solution that I propose is to use **_two temporary files_**: one for stdout and one for stderr.
I imagine that there could be an option which will give you control if you want to see the stdout or not.
I propose that by the default, if the stderr temp file has any information in it, then we would print it and the stdout file will be ignored.


> * the server output should not be a warning because warnings are printed by default. Also, we don't really know what the messages are and mostly do not care. info or debug level maybe?

If we separate stderr from stdout as I proposed above, then we can use `logger.warning()` or `logger.error()` for stderr and `logger.debug()` or `logger.info()` for stdout.
My opinion is that `logger.error()` is appropriate for `stderr` and `logger.debug()` for stdout.

My bigger concern with that approach is how we are going to propagate the cmd option when using tox or even calling the tests with `python aggregate_tests.py`...

jku:
> The solution that I propose is to use two temporary files: one for stdout and one for stderr.

You can do that but I don't think it's necessarily worth the effort: 
* what simple_server.py considers stderr/stdout might not map to our categories anyway (as an example: we might expect the server to print errors in some situations: then it's not an error for us)
* all of simple_server.py output is mostly debugging aid anyway: We'll want to look at it only when something isn't going the way we expected
* if you separate the streams, you no longer know if the stderr line came before the stdout line...

I think dumping all the server output to debug or info is fine

jku:
You know what: we should think about combining (or re-ordering) this PR and issue #1111 .

If we had this in utils.py (edited after chat with Martin):

```
class TestServerProcess():
 
  # starts server, sets up the TempFile, waits for server response before returning
  def __init__(self, server='simple_server.py', port, extra_args=None, timeout=10)
  
  # logs contents of TempFile with given logger, truncates buffer
  def flush_log(self, logger)
  
  # handles kill() and communicate() on the subprocess
  def kill(self)
```
 I think this PR could be vastly smaller and easier to modify later, AND we would have the infra needed by #1111

MVrachev:
I am working to create an abstraction that will handle the subprocess and the temporary files used for logging.
While I am doing this I will release a couple of commits and I will mark this pr as a draft before it's ready for full review.

MVrachev:
I have pushed the initial set of commits in which I created the class which handles the server subprocess and the temporary files used for logging.

I have decided to use `logging.info` for logging in the `flus_log` function because stdout and stderr are logged together and thus it seems like a reasonable default value.

On a successful test right now no logs like `code 404, message File not found` or "GET" queries are reported.

I am planning to push at least 1 commit more in which I will remove or reword some of the test comments.

tests/test_download.py
```diff
@@ -260,27 +259,30 @@ def test_https_connection(self):
...
+        extra_cmd_args=[bad_cert_fname])
+    expd_https_server_handler = utils.TestServerProcess(log=logger,
+        server='simple_https_server.py', port=port4,
+        extra_cmd_args=[expired_cert_fname])
+
+    suffix = '/' +  os.path.basename(target_filepath)
+    good_https_url = 'https://localhost:' + str(port1) + suffix
+    good2_https_url = 'https://localhost:' + str(port2) + suffix
+    bad_https_url = 'https://localhost:' + str(port3) + suffix
+    expired_https_url = 'https://localhost:' + str(port3) + suffix
```
jku: port number typo -- if this did not come up in tests, maybe check why

tests/utils.py
```diff
@@ -74,3 +77,118 @@ def wait_for_server(host, port, timeout=10):
...
+
+      self.__logger.info("Logs from the server subprocess!")
+      if test_name:
+        self.__logger.info("Logs for test " + test_name)
+
+      self.__logger.info("INFO: " + log_message)
+      self.__logger.info("*******************")
+      # Make sure the file is empty before the next test logs new information.
+      self.__temp_log_file.truncate(0)
+
```
jku: We already discussed but I'll leave my opinion here as well for transparency: I think this does things it does not need to... 
* We don't print test names in any other logging either: if it's not clear which test is running, why not improve that in general instead of just for server output
* if the log message from the server does not look good, I think we should fix the test server output instead of replacing things in the output here
* the output missing the server name however: e.g. "simple_server.py": this would be important sometimes

these are opinions though, feel free to disregard

tests/test_download.py
```diff
@@ -260,10 +260,13 @@ def test_https_connection(self):
...
-    port2 = self.server_process_handler.port + 2
-    port3 = self.server_process_handler.port + 3
-    port4 = self.server_process_handler.port + 4
+    # Use different ranges for each one of the ports to make sure
+    # that there are no collisions between the values which are
+    # generated for them.
+    port1 = random.randint(30000, 35000)
+    port2 = random.randint(35001, 40000)
+    port3 = random.randint(40001, 45000)
+    port4 = random.randint(45001, 50000)
```
jku: Hmmm, this now makes comment on line 261 untrue I think

jku:
Looks good, thanks! I'll take this for a test run later to see what the output looks like

tests/test_download.py
MVrachev: I am planning to add a function `_generate_port()` in the TestServerProcess class where I will add a function that generates a port and checks if the port is busy.
So, this code will be removed.

tests/test_download.py
MVrachev: It is not shown in the tests because we are checking that there is an SSL error and not what exact type is:
```
with self.assertRaises(requests.exceptions.SSLError):
        download.unsafe_download(expired_https_url, target_data_length)
```

tests/utils.py
MVrachev: > We don't print test names in any other logging either: if it's not clear which test is running, why not improve that in general instead of just for server output

I think the `test_name` argument can be removed because as we discussed on a test failure the logs are showed once before all tests and a second time just after the test name:
image

I don't think we can add the test names to all logs because many of the logs are coming from public API functions and there is no need to add `test_name` argument just for that.

> if the log message from the server does not look good, I think we should fix the test server output instead of replacing things in the output here

Well, yes, maybe you are right. I will remove that part.

> the output missing the server name however: e.g. "simple_server.py": this would be important sometimes

Will add that.

MVrachev:
I don't know why but when I run `python test_arbitrary_package_attack.py` I get a resource warning even though I have closed the `temporary_log_file` in the `clean` function: 
`/home/mvrachev/src/tuf/.editable_both/lib/python3.6/posixpath.py:86: ResourceWarning: unclosed file <_io.FileIO name=5 mode='rb+' closefd=True>
  for b in map(os.fspath, p):`

Same for every file where I am using this new abstraction...

jku:
> I don't know why but when I run `python test_arbitrary_package_attack.py` I get a resource warning even though I have closed the `temporary_log_file` in the `clean` function:
> `/home/mvrachev/src/tuf/.editable_both/lib/python3.6/posixpath.py:86: ResourceWarning: unclosed file <_io.FileIO name=5 mode='rb+' closefd=True> for b in map(os.fspath, p):`

I'm guessing this is not _your_ unclosed file -- updater currently leaves unclosed files on pretty much every "error" situation I think. Maybe check if same warning appears on develop branch? If it does then you don't need to handle it here.

You can also get more details on the warning with` PYTHONTRACEMALLOC=50` if you want

MVrachev:
> > I don't know why but when I run `python test_arbitrary_package_attack.py` I get a resource warning even though I have closed the `temporary_log_file` in the `clean` function:
> > `/home/mvrachev/src/tuf/.editable_both/lib/python3.6/posixpath.py:86: ResourceWarning: unclosed file <_io.FileIO name=5 mode='rb+' closefd=True> for b in map(os.fspath, p):`
> 
> I'm guessing this is not _your_ unclosed file -- updater currently leaves unclosed files on pretty much every "error" situation I think. Maybe check if same warning appears on develop branch? If it does then you don't need to handle it here.
> 
> You can also get more details on the warning with` PYTHONTRACEMALLOC=50` if you want


Yes, those warnings are visible in the develop branch too.

tests/utils.py
```diff
@@ -74,3 +77,117 @@ def wait_for_server(host, port, timeout=10):
...
+      self.__logger.info("Logs from the subprocess running server: " \
+          + self.server)
+
+      messages = log_message.split('\n')
+      for line in messages:
+        if line != "":
+          self.__logger.info("INFO: " + line)
+
+      self.__logger.info("*******************")
```
jku: I still have issues with this bit:
* very verbose: most of the time the log itself is a single line: this prints 4 times as much
* should not print "INFO: ": this is what the log level is for. If the user wants it printed, they will use a log formatter that prints the log level
* I think the earlier version of a single log entry for the server log was better (it can still be multiline) as that can be made to clearly show the beginning and end of server log without need of extra lines of "************"  (I'll paste the example I tested with below) -- do you see a problem with this approach?

```
    # Build a single log entry with multiline content
    title = "Test server (" + self.server + ") output:"
    message = [title] + log_message.splitlines()
    self.__logger.info('\n| '.join(message))
```

This way the normal oneliner now looks like this (assuming basicConfig default formatter)

```
INFO:logger_name:Test server (simple_server.py) output:
| code 404, message File not found
```

tests/test_proxy_use.py
```diff
@@ -78,25 +75,21 @@ def setUpClass(cls):
...
-        ['simple_server.py', str(cls.http_port)])
+    cls.http_server_handler = utils.TestServerProcess(log=logger)
 
     # Launch an HTTPS server (serves files in the current dir).
-    cls.https_port = cls.http_port + 1
-    cls.https_server_proc = popen_python(
-        ['simple_https_server.py', str(cls.https_port)])
-
+    cls.https_server_handler = utils.TestServerProcess(log=logger,
+        server='simple_https_server.py')
```
jku: This has the same issue as `test_https_connection()` -- there is now a (very small) chance of collision with cls.http_server_handler.port.

I think the real solution is to make simple_server actually report back that it did succeed in binding the port (so TestServerProcess can retry if it did not). It should probably (?) not be part of this PR though: you'll have to choose an interim solution here: I think I'd use `port=cls.http_server_handler.port + 1` because that is closest to current functionality

jku: this applies to all servers below as well

tests/utils.py
```diff
@@ -74,3 +77,117 @@ def wait_for_server(host, port, timeout=10):
...
+    self.__logger.info('Server process started.')
+    self.__logger.info('Server process id: ' + str(self.__server_process.pid))
+    self.__logger.info('Serving on port: ' + str(self.port))
+
+    if timeout > 0:
+      try:
+        wait_for_server('localhost', self.port, timeout)
+      except Exception as e:
+        self.__logger.error(
+            "Error while waiting for the server to start: " + repr(e))
```
jku: Not sure logging an extra error is useful? if the TimeoutError lacks detail (which I think it does, sorry about that) we should add an informative message to the error where its raised maybe.

jku:
I left a few comments, otherwise I think it looks good

trishankatdatadog:
Just wanted to say: thanks for doing all this tedious but necessary work! 🥇

tests/test_proxy_use.py
MVrachev: Yes, I won't fix this in this pr, but I will return it to the previous state in both `test_proxy_use.py` and `test_download.py`.

tests/utils.py
MVrachev: I fixed it in an upcoming commit and removed the commit where I introduce the `INFO` prefix before each line.

tests/utils.py
MVrachev: I created a new commit where I remove that line and make `TimeoutError` message more comprehensive.

MVrachev:
Together with the great help of @jku, we have been elaborating over those changes in the last couple of weeks and now I think this pr is ready for a review!

I have rebased and cleared the commit history to make more sense for a newcomer.

This pr hides the logs from the server because it's using `logger.info()` to log information from the stdout and stderr of the server subprocesses. That's why it's better to merge this pr after https://github.com/theupdateframework/tuf/pull/1145 which will provide a way to receive logs with `info` logging level. This shouldn't be a problem to receive a review.

tests/test_download.py
```diff
@@ -73,15 +71,11 @@ def setUp(self):
...
-    logger.info('\tServer process id: ' + str(self.server_proc.pid))
-    logger.info('\tServing on port: ' + str(self.PORT))
-    junk, rel_target_filepath = os.path.split(target_filepath)
-    self.url = 'http://localhost:'+str(self.PORT)+'/'+rel_target_filepath
+    self.server_process_handler = utils.TestServerProcess(log=logger)
 
-    utils.wait_for_server('localhost', self.PORT)
+    rel_target_filepath = os.path.basename(target_filepath)
+    self.url = 'http://localhost:' \
+        + str(self.server_process_handler.port) + '/' + rel_target_filepath
```
joshuagl, line 78: Nice. 

I'd guess we don't even need to store the url during setup in many tests now that we can get the port number from the `server_process_handler`? 👍

tests/test_proxy_use.py
```diff
@@ -77,26 +75,24 @@ def setUpClass(cls):
       raise NotImplementedError("TestWithProxies only works with Python 2"
                                 " (proxy_server.py is Python2 only)")
 
-    # Launch a simple HTTP server (serves files in the current dir).
     cls.http_port = random.randint(30000, 45000)
```
joshuagl: Why not just let `TestServerProcess` generate this port number?

tests/test_updater.py
```diff
@@ -2065,7 +2044,6 @@ def test_get_valid_targetinfo(self):
     shutil.rmtree(os.path.join(self.repository_directory2, 'metadata'))
     shutil.copytree(os.path.join(self.repository_directory2, 'metadata.staged'),
         os.path.join(self.repository_directory2, 'metadata'))
-
```
joshuagl: Nit: I'm curious why this empty line was removed?

tests/utils.py
```diff
@@ -74,3 +77,111 @@ def wait_for_server(host, port, timeout=10):
...
+        # Make sure that errors from the server side will be logged.
+        self.flush_log()
+        raise e
+
+
+
+  # Logs contents from TempFile, truncates buffer
+  def flush_log(self):
```
joshuagl: Nit: here, and throughout `TestServerProcess`, I think this kind of comment might be more idiomatic as a docstring.
```suggestion
  def flush_log(self):
    """Logs contents from TempFile, truncates buffer"""
```

joshuagl (approved):
Thanks for this great set of cleanups @MVrachev. I added a few minor comments and questions, but this is pretty much ready to merge. Excellent PR.

tests/test_download.py
MVrachev: Well, 3 out of 4 tests in `test_download.py` are using the `self.url.` 
That's why it make sense to keep it as a variable.

tests/test_proxy_use.py
MVrachev: It could, I just tried to make the lines shorter and just doesn't change it.
We are planning to leave all instances of port generation outside `tests/utils.py` and it's in the best interest of us if we do that as well.
I will change it.

tests/test_updater.py
MVrachev: Honestly, it seems it's a mistake I did while I was resolving prs.
I will fix it right away.

MVrachev:
I rebased my changes upon the latest changes in develop and changed the second commit `Log subproceses stdout and stderr in temp files` with the comments from Joshua.

MVrachev:
I have decided that because I am working in the same code area I will use this pr to fix the issue https://github.com/theupdateframework/tuf/issues/1119 too and that's what the last commit is all about.

jku (approved):
LGTM. 

Maybe you are aware but I'll mention for future commits: instead of 

> For reference read:
> #1119

you can do `Fixes #1119` (or use a few other keywords) to auto-close on merge -- obviously it's useful only if you know the issue should be closed when commit is merged.

Also, the format `Fixes issue #N` does not seem to work -- this is why this PR cover letter is probably not going to trigger auto-closing either.

MVrachev:
> LGTM.
> 
> Maybe you are aware but I'll mention for future commits: instead of
> 
> > For reference read:
> > #1119
> 
> you can do `Fixes #1119` (or use a few other keywords) to auto-close on merge -- obviously it's useful only if you know the issue should be closed when commit is merged.
> 
> Also, the format `Fixes issue #N` does not seem to work -- this is why this PR cover letter is probably not going to trigger auto-closing either.

Thanks! 
That's what motivated me to create pr https://github.com/theupdateframework/tuf/pull/1152.

tests/utils.py
```diff
@@ -97,3 +100,112 @@ def configure_test_logging(argv):
...
+  <Purpose>
+    Creates a child process with the subprocess.Popen object and
+    TempFile object used for logging.
+
+   <Arguments>
+      log:
+        Logger which will be used for logging.
+
+      server:
+        Path to the server which will be runned by the subprocess.
```
```suggestion
        Path to the server to run in the subprocess.
```
Runned is not a word, perhaps rephrase like this?

joshuagl (approved):
This is an excellent set of changes, thank you. I'd like to see a minor fixup of a docstring. After that, I think this is good to merge.

joshuagl:
Looks good, thanks Martin. Could you rebase and squash/fixup b9ef099 into e7a8a78?

MVrachev:
Updated.

joshuagl:
Awesome work, thanks @MVrachev

PR #1106, opened 2020-08-11
PoC(tap): Succinct hashed bin delegations

mnm678:
This pr adds support for succinct hashed bin delegations as described in theupdateframework/taps#120.

### DIFF
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1640,14 +1640,9 @@ class TestTargets(unittest.TestCase):
 
     delegated_rolename = self.targets_object.rolename + '.hbd'
 
-    bin_names = [delegated_rolename + '-0', delegated_rolename + '-1',
-                delegated_rolename + '-2', delegated_rolename + '-3',
-                delegated_rolename + '-4', delegated_rolename + '-5',
-                delegated_rolename + '-6', delegated_rolename + '-7',
-                delegated_rolename + '-8', delegated_rolename + '-9',
-                delegated_rolename + '-a', delegated_rolename + '-b',
-                delegated_rolename + '-c', delegated_rolename + '-d',
-                delegated_rolename + '-e', delegated_rolename + '-f']
+    # Create a list of hex-suffixed names from "<name>-0" to "<name>-f"
+    bin_names = ["{}-{:x}".format(delegated_rolename,
+        i) for i in range(0, 16)]
 
     delegated_rolenames.append(delegated_rolename)
 
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1484,6 +1484,7 @@ class TestTargets(unittest.TestCase):
     threshold = 1
     paths = ['*']
     path_hash_prefixes = ['e3a3', '8fae', 'd543']
+    succinct_hash_delegations = {'prefix_bit_length': 2}
 
     self.targets_object.delegate(rolename, public_keys, paths,
         threshold, terminating=False, list_of_targets=list_of_targets,
@@ -1491,6 +1492,15 @@ class TestTargets(unittest.TestCase):
 
     self.assertEqual(self.targets_object.get_delegated_rolenames(),
                      ['tuf'])
+
+    # Tests succinct hashed bin delegations
+    rolename = 'succinct'
+    self.targets_object.delegate(rolename, public_keys, paths,
+        threshold, terminating=False, list_of_targets=list_of_targets,
+        succinct_hash_delegations=succinct_hash_delegations)
+
+    self.assertEqual(self.targets_object.get_delegated_rolenames(),
+                    ['tuf', 'succinct'])
 
     # Test for delegated paths that do not exist.
     # An exception should not be raised for non-existent delegated paths, since
@@ -1506,6 +1516,7 @@ class TestTargets(unittest.TestCase):
     self.targets_object.delegate(rolename, public_keys, [], threshold,
         terminating=False, list_of_targets=['non-existent.txt'],
         path_hash_prefixes=path_hash_prefixes)
+
 
     # Test improperly formatted arguments.
     self.assertRaises(securesystemslib.exceptions.FormatError,
@@ -1531,6 +1542,10 @@ class TestTargets(unittest.TestCase):
     self.assertRaises(securesystemslib.exceptions.FormatError,
         self.targets_object.delegate, rolename, public_keys, paths, threshold,
         list_of_targets, 3)
+
+    self.assertRaises(securesystemslib.exceptions.FormatError,
+        self.targets_object.delegate, rolename, public_keys, paths, threshold,
+        list_of_targets, None, succinct_hash_delegations=5)
 
     # Test invalid arguments (e.g., already delegated 'rolename', non-existent
     # files, etc.).
@@ -1577,7 +1592,7 @@ class TestTargets(unittest.TestCase):
       have_prefixes = False
 
       for delegated_role in roleinfo['delegations']['roles']:
-        if len(delegated_role['path_hash_prefixes']) > 0:
+        if '.hbd' not in delegated_role['name'] and len(delegated_role['path_hash_prefixes']) > 0:
           rolename = delegated_role['name']
           prefixes = delegated_role['path_hash_prefixes']
           have_prefixes = True
@@ -1605,6 +1620,29 @@ class TestTargets(unittest.TestCase):
     self.assertEqual(sorted(self.targets_object.get_delegated_rolenames()),
                      sorted(delegated_rolenames))
     check_prefixes_match_range()
+
+
+    # Test succinct hashed bin delegations
+    self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
+                                            number_of_bins=16, succinct=True)
+
+    delegated_rolename = self.targets_object.rolename + '.hbd'
+
+    bin_names = [delegated_rolename + '-0', delegated_rolename + '-1',
+                delegated_rolename + '-2', delegated_rolename + '-3',
+                delegated_rolename + '-4', delegated_rolename + '-5',
+                delegated_rolename + '-6', delegated_rolename + '-7',
+                delegated_rolename + '-8', delegated_rolename + '-9',
+                delegated_rolename + '-a', delegated_rolename + '-b',
+                delegated_rolename + '-c', delegated_rolename + '-d',
+                delegated_rolename + '-e', delegated_rolename + '-f']
+
+    delegated_rolenames.append(delegated_rolename)
+
+    self.assertEqual(sorted(self.targets_object.get_delegated_rolenames()),
+                    sorted(delegated_rolenames))
+    self.assertEqual(sorted(self.targets_object._succinct_delegations),
+                    sorted(bin_names))
 
     # For testing / coverage purposes, try to create delegated bins that
     # hold a range of hash prefixes (e.g., bin name: 000-003).
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -1604,7 +1604,7 @@ class TestTargets(unittest.TestCase):
       have_prefixes = False
 
       for delegated_role in roleinfo['delegations']['roles']:
-        if '.hbd' not in delegated_role['name'] and len(delegated_role['path_hash_prefixes']) > 0:
+        if 'delegation_hash_prefix_len' not in delegated_role and len(delegated_role['path_hash_prefixes']) > 0:
           rolename = delegated_role['name']
           prefixes = delegated_role['path_hash_prefixes']
           have_prefixes = True
@@ -1638,10 +1638,10 @@ class TestTargets(unittest.TestCase):
     self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
                                             number_of_bins=16, succinct=True)
 
-    delegated_rolename = self.targets_object.rolename + '.hbd'
+    delegated_rolename = self.targets_object.rolename + '.hbd-'
 
     # Create a list of hex-suffixed names from "<name>-0" to "<name>-f"
-    bin_names = ["{}-{:x}".format(delegated_rolename,
+    bin_names = ["{}{:x}".format(delegated_rolename,
         i) for i in range(0, 16)]
 
     delegated_rolenames.append(delegated_rolename)
@@ -1650,6 +1650,21 @@ class TestTargets(unittest.TestCase):
                     sorted(delegated_rolenames))
     self.assertEqual(sorted(self.targets_object._succinct_delegations),
                     sorted(bin_names))
+
+    # Test succinct hashed bin delegation with a custom prefix
+    self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
+                            number_of_bins=2, succinct=True, prefix='prefix')
+
+    delegated_rolename = "prefix-"
+
+    bin_names.append("prefix-0")
+    bin_names.append("prefix-1")
+    delegated_rolenames.append(delegated_rolename)
+
+    self.assertEqual(sorted(self.targets_object.get_delegated_rolenames()),
+                  sorted(delegated_rolenames))
+    self.assertEqual(sorted(self.targets_object._succinct_delegations),
+                  sorted(bin_names))
 
     # For testing / coverage purposes, try to create delegated bins that
     # hold a range of hash prefixes (e.g., bin name: 000-003).
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -295,7 +295,7 @@ class TestRepository(unittest.TestCase):
     # Test if succinct delegations are written on writeall
     list_of_targets = [target1, target2]
     repository.targets.delegate_hashed_bins(list_of_targets, [role1_pubkey],
-                                            4, True)
+                                            4, True, prefix='targets.hbd-')
     repository.targets.load_signing_key_succinct_delegations(role1_privkey)
     repository.writeall()
 
@@ -1623,7 +1623,7 @@ class TestTargets(unittest.TestCase):
     # Test delegate_hashed_bins() and verify that 16 hashed bins have
     # been delegated in the parent's roleinfo.
     self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
-                                             number_of_bins=16, prefix='bin')
+                                             number_of_bins=16, prefix='bin-')
 
     # The expected child rolenames, since 'number_of_bins' = 16
     delegated_rolenames = ["{}{:x}".format('bin-',
@@ -1635,7 +1635,7 @@ class TestTargets(unittest.TestCase):
 
     # Test delegate_hashed_bins with a custom prefix
     self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
-                                  number_of_bins=4, prefix="pre")
+                                  number_of_bins=4, prefix="pre-")
 
     delegated_rolenames.append("pre-0")
     delegated_rolenames.append("pre-1")
@@ -1646,10 +1646,10 @@ class TestTargets(unittest.TestCase):
                      sorted(delegated_rolenames))
 
     # Test succinct hashed bin delegations
-    self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
-                                            number_of_bins=16, succinct=True)
-
     delegated_rolename = self.targets_object.rolename + '.hbd-'
+
+    self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
+        number_of_bins=16, succinct=True, prefix=delegated_rolename)
 
     # Create a list of hex-suffixed names from "<name>-0" to "<name>-f"
     bin_names = ["{}{:x}".format(delegated_rolename,
@@ -1664,7 +1664,7 @@ class TestTargets(unittest.TestCase):
 
     # Test succinct hashed bin delegation with a custom prefix
     self.targets_object.delegate_hashed_bins(list_of_targets, public_keys,
-                            number_of_bins=2, succinct=True, prefix='prefix')
+                            number_of_bins=2, succinct=True, prefix='prefix-')
 
     delegated_rolename = "prefix-"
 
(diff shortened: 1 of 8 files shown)


tests/test_repository_tool.py
```diff
@@ -1609,6 +1636,29 @@ def check_prefixes_match_range():
...
+    delegated_rolename = self.targets_object.rolename + '.hbd'
+
+    bin_names = [delegated_rolename + '-0', delegated_rolename + '-1',
+                delegated_rolename + '-2', delegated_rolename + '-3',
+                delegated_rolename + '-4', delegated_rolename + '-5',
+                delegated_rolename + '-6', delegated_rolename + '-7',
+                delegated_rolename + '-8', delegated_rolename + '-9',
+                delegated_rolename + '-a', delegated_rolename + '-b',
+                delegated_rolename + '-c', delegated_rolename + '-d',
+                delegated_rolename + '-e', delegated_rolename + '-f']
```
lukpueh: Looks like a job for a list comprehension.
```python
# Create a list of hex-suffixed names from "<name>-0" to "<name>-f"
bin_names = ["{}-{:x}".format(delegated_rolename, i) for i in  xrange(0, 16)]
```

Just saying... :)

tuf/formats.py
```diff
@@ -126,6 +126,9 @@
 # A list of path hash prefixes.
 PATH_HASH_PREFIXES_SCHEMA = SCHEMA.ListOf(PATH_HASH_PREFIX_SCHEMA)
 
+SUCCINCT_HASH_DELEGATIONS_SCHEMA = SCHEMA.Object(
+  prefix_bit_length = SCHEMA.Integer(lo=0, hi=32))
```
lukpueh: I know the tap defines it this way, so this comment is probably a better fit for https://github.com/theupdateframework/taps/pull/120, but is it really necessary to use an object/dict with only one item? Can't we just assign the prefix bit length value directly to the `succinct_hash_delegations` field (and maybe rethink its name)?

tuf/repository_tool.py
```diff
@@ -2545,6 +2627,12 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
         from [000]... - [003]..., where the series of digits in brackets is
         considered the hash prefix).
 
+      succinct:
+        Whether the hashed bins should use succinct hashed bin delegations.
+
+      succinct:
+        Whether the hashed bins should use succinct hashed bin delegations.
+
```
lukpueh:
```suggestion
      succinct:
        Whether the hashed bins should use succinct hashed bin delegations.

```

tuf/repository_tool.py
```diff
@@ -3111,7 +3246,14 @@ def load_repository(repository_directory, repository_name='default',
   # [('role1', 'targets'), ('role2', 'targets'), ... ]
   roleinfo = tuf.roledb.get_roleinfo('targets', repository_name)
   for role in roleinfo['delegations']['roles']:
-    delegations.append((role['name'], 'targets'))
+    if role['name'].endswith('.hbd'):
+      prefix_len = role['succinct_hash_delegations']['prefix_bit_length']
+      num_bins = 2 ** prefix_len
+      for i in range(num_bins):
+        bin_name = role['name'] + '-' + "{num:0{len}x}".format(num=i, len=prefix_len)
```
lukpueh: Programming-style nit: I suggest to only use one of format strings or string concatenation, e.g.:
```python
"{name}-{num:0{len}x}".format(name=role['name'], num=i, len=prefix_len)
```

lukpueh (changes requested):
Thanks for taking a stab at the hashed bin delegation POC. Before I continue reviewing I have a high-level question. Why do we have different naming conventions for normal hashed bin delegated roles and succinct hashed bin delegated roles?

It would save us a lot of implementation trouble, if we used the same convention. And I don't see a good reason why we shouldn't.

IIUC the delegated role names serve the same purpose in both cases, that is to indicate what range of hashed bins the role is responsible for.

The main advantage of the new naming scheme for succinct hbd, i.e. `"<DELEGATING_ROLE>.hbd-(0..2^BIT_LENGTH-1)"`, compared to the naming scheme of normal hbd, i.e. `"<LOW>-<HIGH>"`, is that it adds the ability of different roles delegating to different hashed bins in one repository. Is there a use case for this? It seems more like a side-effect of the TAP than an original motivation. Also, if we didn't need that flexibility for normal hbd, why add it now for succinct hbd?

Please let me know if I'm missing something.

cc @joshuagl, @trishankatdatadog, @JustinCappos

mnm678:
> The main advantage of the new naming scheme for succinct hbd, i.e. `"<DELEGATING_ROLE>.hbd-(0..2^BIT_LENGTH-1)"`, compared to the naming scheme of normal hbd, i.e. `"<LOW>-<HIGH>"`, is that it adds the ability of different roles delegating to different hashed bins in one repository. Is there a use case for this? It seems more like a side-effect of the TAP than an original motivation. Also, if we didn't need that flexibility for normal hbd, why add it now for succinct hbd?

Yes, I believe the intention of the new naming scheme is to allow for different roles to create independent hashed bin delegations. I don't know that this ability has been requested by any existing implementer, but I can imagine a use case for registries as different uploaders to the same registry may each want to use succinct delegations. Does this seem like a likely use case? If not, we should probably stick with the existing naming scheme.

tuf/formats.py
mnm678: I agree, the object with one item feels a bit redundant. I'll post in the TAP to propose a change.

tuf/repository_tool.py
mnm678: lol, thanks for catching this :)

JustinCappos:
> > The main advantage of the new naming scheme for succinct hbd, i.e. `"<DELEGATING_ROLE>.hbd-(0..2^BIT_LENGTH-1)"`, compared to the naming scheme of normal hbd, i.e. `"<LOW>-<HIGH>"`, is that it adds the ability of different roles delegating to different hashed bins in one repository. Is there a use case for this? It seems more like a side-effect of the TAP than an original motivation. Also, if we didn't need that flexibility for normal hbd, why add it now for succinct hbd?
> 
> Yes, I believe the intention of the new naming scheme is to allow for different roles to create independent hashed bin delegations. I don't know that this ability has been requested by any existing implementer, but I can imagine a use case for registries as different uploaders to the same registry may each want to use succinct delegations. Does this seem like a likely use case? If not, we should probably stick with the existing naming scheme.

Yes, different parties need to be able to use this mechanism.  I agree we must not omit the <DELEGATING_ROLE> from the name.

lukpueh:
Thanks for addressing my comments and taking the naming discussion to the TAP page. Could you please also make sure to follow (our) [commit guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/commits.md).

lukpueh:
I still think that we should harmonize the naming conventions, in order to reduce the delegated bin role name case handling in the code. Since the traditional hashed bin delegation does not rely on a name convention (at least not on the client), we could:
- allow a custom role name prefix such as `"<DELEGATING ROLE>.hbd-"` there as well,
- and we could also use the bin number as suffix instead of the hash prefix range.

tuf/repository_tool.py
```diff
@@ -1793,6 +1852,11 @@ def add_delegated_role(self, rolename, targets_object):
       targets_object:
         A Targets() object.
 
+      succinct:
+        Whether the delegated role is used by succinct hashed bin delegations.
+        If succinct delegations are used, the role is added to '_succinct_delegations'
+        instead of '_delegated_roles'.
```
lukpueh: Why?

joshuagl:
> I still think that we should harmonize the naming conventions, in order to reduce the delegated bin role name case handling in the code. Since the traditional hashed bin delegation does not rely on a name convention (at least not on the client), we could:
> 
> * allow a custom role name prefix such as `"<DELEGATING ROLE>.hbd-"` there as well,
> * and we could also use the bin number as suffix instead of the hash prefix range.

I am also in favour of using consistent hashed bin naming conventions

lukpueh:
The idea behind my [suggestion from above](https://github.com/theupdateframework/tuf/pull/1106#issuecomment-679043312) was that we could use the same (slightly updated) hash bin helper functions for both types of hashed bin delegation. So that we don't have to concatenate strings / construct names anywhere else.

```python
def create_bin_name(bin_info, prefix=""):
  """Returns prefix + suffix. 
  Suffix is computed based on the passed bin identifying info and follows a convention ,
  e.g. lo-hi of hex hash target prefix, or binary target prefix, or bin index, etc. """
  ...

def find_bin_for_target_hash(target_hash, bins_info):
  """Returns either the bin name or abstract bin info that can be used to create the name.
  bins_info is used to find out how many bins there are and what target prefixes each bin
  serves. It could be number of bins, or hash prefix bit length, etc."""
  ...

```

mnm678:
@lukpueh I updated this pr to try to simplify the naming convention, as well as rebasing onto develop. One thing to note is that this updates the default way that non-succinct hashed bins are named. I don't think this breaks backwards compatibility as the previous naming convention (or any arbitrary one) is still supported by the client.

lukpueh:
Thanks for the updates, @mnm678! I agree with you that the update shouldn't be a problem for clients. It is still a backwards incompatible change to the repository tool though. But I don't think that's a big problem. Although, we could just keep using the hash-prefix-range as role name suffix for both classic and succinct hbd, then we wouldn't have a problem at all.

Also, I noticed that you still repeat the bin role name creation in quite a few places (see various occurrences of `{num:0{len}x}` or `name_prefix = prefix + '-'`), don't you think it's better to do this in a single place as I suggested above in https://github.com/theupdateframework/tuf/pull/1106#issuecomment-681866823?

lukpueh:
@mnm678, please ping me when you have addressed my comments and the PR is ready for another review, or if you need any help. :) 

Also, could you please try to stick closer to our git [commit](https://github.com/secure-systems-lab/lab-guidelines/blob/master/commits.md) and [history](https://github.com/secure-systems-lab/lab-guidelines/blob/master/git-history.md) guidelines (separation of commits, commit message wording and line length, no merge commits in feature branch, etc...). Thanks!

mnm678:
@lukpueh I think this is ready for another look. I did some git history cleanup, but can make some more updates to the commit messages if needed.

tuf/client/updater.py
```diff
@@ -2971,8 +2971,24 @@ def _visit_child_role(self, child_role, target_filepath):
...
+      # Find the bin associated with this hash and return the bin name
+      prefix_bits = child_role_delegation_hash_prefix_len
+      number_of_bins = 2 ** prefix_bits
+      prefix_length = len("{:x}".format(number_of_bins - 1))
+      bin_size = 16 ** prefix_length // number_of_bins
+      prefix = int(target_filepath_hash[:prefix_length], 16)
+
+      bin_num = (prefix - (prefix % bin_size)) / bin_size
+
+      bin_name = child_role_name + "{num:0{len}x}".format(num=int(bin_num), len=prefix_length)
```
```suggestion
      bin_name = "{role_name}{num:0{len}x}".format(
          role_name=child_role_name, num=int(bin_num), len=prefix_length)
```
We should try and avoid mixing string concatenation and string formatting

tuf/repository_tool.py
```diff
@@ -2341,6 +2414,9 @@ def delegate(self, rolename, public_keys, paths, threshold=1,
         hashed bin delegations.  Targets may be located and stored in hashed
         bins by calculating the target path's hash prefix.
 
+      succinct:
+        If the delegation is to bins using succinct hashed bin delegations.
```
joshuagl: I think this comment is out of date? Aren't we expecting an integer argument here?

tuf/repository_tool.py
```diff
@@ -2341,6 +2414,9 @@ def delegate(self, rolename, public_keys, paths, threshold=1,
         hashed bin delegations.  Targets may be located and stored in hashed
         bins by calculating the target path's hash prefix.
 
+      succinct:
```
joshuagl:
```suggestion
      succinct_hash_delegations:
```

tuf/repository_lib.py
```diff
@@ -1035,31 +1035,28 @@ def get_metadata_versioninfo(rolename, repository_name):
...
 
-def create_bin_name(low, high, prefix_len):
+def create_bin_name(bin_num, bin_num_length, prefix=''):
   """
   <Purpose>
-    Create a string name of a delegated hash bin, where name will be a range of
-    zero-padded (up to prefix_len) strings i.e. for low=00, high=07,
-    prefix_len=3 the returned name would be '000-007'.
+    Create a string name of a delegated hash bin, where name will be
+    prefix-NUM where NUM will be a
```
joshuagl: this comment implies we insert the '-', but we don't appear to be doing that.

tuf/repository_lib.py
```diff
@@ -1035,31 +1035,28 @@ def get_metadata_versioninfo(rolename, repository_name):
...
+    zero-padded (up to prefix_len) string i.e. for bin_num=0, bin_num_length=2
+    prefix="pre", the name will be pre-00
```
joshuagl: Same here, implies we're inserting the '-'.

tuf/client/updater.py
joshuagl: It's a shame we have to duplicate this here and don't have a way to share functions with the client *without* including all of repository_lib – there's clearly a need for it.

tuf/repository_lib.py
```diff
@@ -1035,31 +1035,28 @@ def get_metadata_versioninfo(rolename, repository_name):
...
   <Returns>
-    A string bin name, with each end of the range zero-padded up to prefix_len
+    A string bin name
   """
-  if low == high:
-    return "{low:0{len}x}".format(low=low, len=prefix_len)
-
-  return "{low:0{len}x}-{high:0{len}x}".format(low=low, high=high,
-      len=prefix_len)
+  return '{pre}{num:0{len}x}'.format(pre=prefix, num=bin_num, len=bin_num_length)
```
joshuagl, line 1059: While I am 💯 in support of having unified hbd naming for succinct and non-succinct hbd, this change introduces a backward incompatible change to the repository tools in how we create name hashed bins in the non-succinct case.

Is there anything in the TAP which prevents us from maintaining backwards compatibility by having the bin names be: `"{pre}{low:0{len}x}-{high:0{len}x}".format(pre=prefix, low=low, high=high, len=prefix_len)`?

tuf/repository_tool.py
```diff
@@ -1777,7 +1778,64 @@ def __call__(self, rolename):
...
+    """
+    # Does 'key' have the correct format?
+    # Ensure the arguments have the appropriate number of objects and object
+    # types, and that all dict keys are properly named.  Raise
+    # 'securesystemslib.exceptions.FormatError' if any are improperly formatted.
+    securesystemslib.formats.ANYKEY_SCHEMA.check_match(key)
+
+    # Ensure the private portion of the key is available, otherwise signatures
+    # cannot be generated when the metadata file is written to disk.
+    if 'private' not in key['keyval'] or not len(key['keyval']['private']):
```
joshuagl: I like dict's `get` here, it's slightly shorter (and therefore more readable) and avoids querying the dict twice.

```suggestion
    if not key['keyval'].get('private'):
```

tuf/repository_tool.py
```diff
@@ -1777,7 +1778,64 @@ def __call__(self, rolename):
...
+      raise securesystemslib.exceptions.Error('This is not a private key.')
+
+    # Has the key, with the private portion included, been added to the keydb?
+    # The public version of the key may have been previously added.
+    try:
+      tuf.keydb.add_key(key, repository_name=self._repository_name)
+
+    except tuf.exceptions.KeyAlreadyExistsError:
+      tuf.keydb.remove_key(key['keyid'], self._repository_name)
+      tuf.keydb.add_key(key, repository_name=self._repository_name)
```
joshuagl, line 1826: should we be handling a possible exception on the second `add_key` call here?

tuf/repository_tool.py
```diff
@@ -1814,12 +1880,18 @@ def add_delegated_role(self, rolename, targets_object):
       raise securesystemslib.exceptions.FormatError(repr(targets_object) + ' is'
           ' not a Targets object.')
 
-
-    if rolename in self._delegated_roles:
-      logger.debug(repr(rolename) + ' already exists.')
+    if succinct:
+      if rolename in self._succinct_delegations:
+        logger.debug(repr(rolename) + ' already exists.')
```
joshuagl: Do we really want to ignore this? What happens if we load a repository with succinct hashed bin delegations and want to change i.e. the number of succinct hashed bins?

Is there any harm in adding the role to `self._succinct_delegations` even when it's already present?

tuf/repository_tool.py
```diff
@@ -1814,12 +1880,18 @@ def add_delegated_role(self, rolename, targets_object):
...
+      else:
+        self._succinct_delegations[rolename] = targets_object
 
     else:
-      self._delegated_roles[rolename] = targets_object
```
joshuagl, line 1822: I'd just leave this as-is and always assign. The debug info isn't useful, I don't think.

tuf/repository_tool.py
```diff
@@ -2579,24 +2671,33 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
...
+      name = repo_lib.create_bin_name(idx, prefix_length, prefix)
+      if succinct:
+      # For succinct hased bin delegations, these metadata files will still
+      # be created, but they will not all be listed in the delegation
+        role = {"name":name,
+            "keyids": keyids,
+            "threshold": 1,
+            "terminating": False,
+            "target_paths": []}
+        ordered_roles.append(role)
```
joshuagl: The role is already appended at the end of the loop, outside of the conditional. I think we're adding the role to ordered_roles twice?
```suggestion
```

tuf/repository_tool.py
```diff
@@ -2579,24 +2671,33 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
...
-        target_hash_prefixes = [name]
+    for idx in range(0, number_of_bins):
+      name = repo_lib.create_bin_name(idx, prefix_length, prefix)
+      if succinct:
+      # For succinct hased bin delegations, these metadata files will still
+      # be created, but they will not all be listed in the delegation
+        role = {"name":name,
+            "keyids": keyids,
+            "threshold": 1,
+            "terminating": False,
```
joshuagl: I think this is not required? We're setting these defaults later when iterating over ordered_roles

tuf/repository_tool.py
```diff
@@ -2639,34 +2738,60 @@ def delegate_hashed_bins(self, list_of_targets, keys_of_hashed_bins,
...
+                    'path_hash_prefixes': bin_role['target_hash_prefixes']}
+        delegated_roleinfos.append(roleinfo)
+      else:
+        roleinfo = {'name': bin_role['name'],
+                    'keyids': keyids,
+                    'signing_keyids': [],
+                    'version':1,
+                    'threshold': 1,
+                    'terminating': False,
+                    'paths': []}
```
joshuagl: Are an empty `paths` and the `version` required? If so we should add them for non-succinct hbd.

joshuagl: In fact, we don't appear to be doing anything with this roleinfo dict? Let's just delete this else clause?

joshuagl (changes requested):
Good work here @mnm678, thank you.

I have a few questions and comments in line. Most significantly I wonder if the TAP requires that we change the way we name hbd. Could we support succinct hashed bin delegations with the hash-prefix range as calculated by the current code as the rolename suffix? That way we could harmonise our hashed bin code and not introduce a backwards incompatible change to repository_tool?

Hopefully this could help us implement succinct hashed bin delegations with fewer branches too.

mnm678:
> I have a few questions and comments in line. Most significantly I wonder if the TAP requires that we change the way we name hbd. Could we support succinct hashed bin delegations with the hash-prefix range as calculated by the current code as the rolename suffix? That way we could harmonise our hashed bin code and not introduce a backwards incompatible change to repository_tool?

Thanks for the review @joshuagl! The main downside I see to using the current naming scheme is that it could make very long filenames if the prefix is sufficiently long. For example `rolename-00000000-00000001`, etc. Either way I can see the advantage to having at least an initial implementation that is backwards compatible.

tuf/client/updater.py
mnm678: I agree. Maybe we can create a separate library of helper functions as part of the refactor (this would also be useful for the snapshot merkle implementation, and probably others).

tuf/repository_lib.py
mnm678: The TAP specifies the new naming scheme, but we can discuss this further to see if the existing naming makes more sense.

tuf/repository_tool.py
mnm678: I don't think so. If the KeyAlreadyExistsError is raised on the first call, the formatting and existence checks should have already passed.

tuf/repository_tool.py
mnm678: You're right, I think replacing any existing value makes more sense.

tuf/roledb.py
```diff
@@ -963,7 +971,10 @@ def get_delegated_rolenames(rolename, repository_name='default'):
   delegated_roles = []
 
   for delegated_role in roleinfo['delegations']['roles']:
-    delegated_roles.append(delegated_role['name'])
+    if 'succinct_hash_delegations' in delegated_role:
+      delegated_roles.append(delegated_role['succinct_hash_delegations']['bin_name_prefix'])
```
joshuagl, line 975: Is the prefix enough here, or do we need the full role name?

mnm678: I'm not sure what the expected behavior here would be. A single succinct delegation in the metadata corresponds to many actual delegations, so should this function return names of the delegated roles in the metadata file (what it currently does), or the names of all roles delegated to by the current role (including the full names of all succinct delegations)?

Either way, I'll be sure to document it in the docstring.

joshuagl:
As this is a proof of concept for a TAP I've converted it to a draft PR.

lukpueh:
Now with the [legacy code gone](https://github.com/theupdateframework/python-tuf/blob/develop/docs/1.0.0-ANNOUNCEMENT.md) this PR will need quite a re-write.

Most of above comments were either relevant to the legacy code, or to the TAP itself, and have been addressed in the taps repo. So I suggest to close here and re-open a fresh PR.

For the repository side, since we currently don't have a repository tool, it's probably best to prepare something like the [hashed bin delegation example](https://github.com/theupdateframework/python-tuf/blob/develop/examples/repo_example/hashed_bin_delegation.py), and the client-side code should work with `ngclient`.

PR #1112, opened 2020-08-18
Add simple TUF role metadata model

lukpueh:
This PR copies over the new `api/metadata.py` module (+ tests) added in #1060, which has shifted focus to implementing a Hashicorp Vault interface.

Co-authorship by @trishankatdatadog, @joshuagl, @sechkova and me plus note-worthy commit message details are preserved in a single squashed commit.

Missing TODOs items for this PR include at least.

- [x] add code documentation (docstrings)
- [x] finalize signature creation and verification interface
- [ ] add input validatation

See doc header of `api/metadata.py` for more details

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -89,7 +89,7 @@ class TestMetadata(unittest.TestCase):
                 ('targets', Targets)]:
 
             path = os.path.join(self.repo_dir, 'metadata', metadata + '.json')
-            metadata_obj = Metadata.read_from_json(path)
+            metadata_obj = Metadata.from_json_file(path)
 
             # Assert that generic method instantiates the right inner class for
             # each metadata type
@@ -103,31 +103,31 @@ class TestMetadata(unittest.TestCase):
             f.write(json.dumps(bad_metadata).encode('utf-8'))
 
         with self.assertRaises(ValueError):
-            Metadata.read_from_json(bad_metadata_path)
+            Metadata.from_json_file(bad_metadata_path)
 
         os.remove(bad_metadata_path)
 
 
     def test_compact_json(self):
         path = os.path.join(self.repo_dir, 'metadata', 'targets.json')
-        metadata_obj = Metadata.read_from_json(path)
+        metadata_obj = Metadata.from_json_file(path)
         self.assertTrue(
-                len(metadata_obj.as_json(compact=True)) <
-                len(metadata_obj.as_json()))
+                len(metadata_obj.to_json(compact=True)) <
+                len(metadata_obj.to_json()))
 
 
     def test_read_write_read_compare(self):
         for metadata in ['snapshot', 'timestamp', 'targets']:
             path = os.path.join(self.repo_dir, 'metadata', metadata + '.json')
-            metadata_obj = Metadata.read_from_json(path)
+            metadata_obj = Metadata.from_json_file(path)
 
             path_2 = path + '.tmp'
-            metadata_obj.write_to_json(path_2)
-            metadata_obj_2 = Metadata.read_from_json(path_2)
+            metadata_obj.to_json_file(path_2)
+            metadata_obj_2 = Metadata.from_json_file(path_2)
 
             self.assertDictEqual(
-                    metadata_obj.as_dict(),
-                    metadata_obj_2.as_dict())
+                    metadata_obj.to_dict(),
+                    metadata_obj_2.to_dict())
 
             os.remove(path_2)
 
@@ -135,7 +135,7 @@ class TestMetadata(unittest.TestCase):
     def test_sign_verify(self):
         # Load sample metadata (targets) and assert ...
         path = os.path.join(self.repo_dir, 'metadata', 'targets.json')
-        metadata_obj = Metadata.read_from_json(path)
+        metadata_obj = Metadata.from_json_file(path)
 
         # ... it has a single existing signature,
         self.assertTrue(len(metadata_obj.signatures) == 1)
@@ -180,7 +180,7 @@ class TestMetadata(unittest.TestCase):
         # with real data
         snapshot_path = os.path.join(
                 self.repo_dir, 'metadata', 'snapshot.json')
-        md = Metadata.read_from_json(snapshot_path)
+        md = Metadata.from_json_file(snapshot_path)
 
         self.assertEqual(md.signed.version, 1)
         md.signed.bump_version()
@@ -195,7 +195,7 @@ class TestMetadata(unittest.TestCase):
     def test_metadata_snapshot(self):
         snapshot_path = os.path.join(
                 self.repo_dir, 'metadata', 'snapshot.json')
-        snapshot = Metadata.read_from_json(snapshot_path)
+        snapshot = Metadata.from_json_file(snapshot_path)
 
         # Create a dict representing what we expect the updated data to be
         fileinfo = snapshot.signed.meta
@@ -211,7 +211,7 @@ class TestMetadata(unittest.TestCase):
     def test_metadata_timestamp(self):
         timestamp_path = os.path.join(
                 self.repo_dir, 'metadata', 'timestamp.json')
-        timestamp = Metadata.read_from_json(timestamp_path)
+        timestamp = Metadata.from_json_file(timestamp_path)
 
         self.assertEqual(timestamp.signed.version, 1)
         timestamp.signed.bump_version()
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -3,7 +3,10 @@
 # Copyright 2020, New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 """ Unit tests for api/metdata.py
+
 """
+
+import sys
 import logging
 import os
 import shutil
@@ -13,10 +16,21 @@ import unittest
 from datetime import timedelta
 from dateutil.relativedelta import relativedelta
 
-from tuf.api.metadata import (
-    Snapshot,
-    Timestamp,
-)
+# TODO: Remove case handling when fully dropping support for versions >= 3.6
+IS_PY_VERSION_SUPPORTED = sys.version_info >= (3, 6)
+
+# Use setUpModule to tell unittest runner to skip this test module gracefully.
+def setUpModule():
+    if not IS_PY_VERSION_SUPPORTED:
+        raise unittest.SkipTest("requires Python 3.6 or higher")
+
+# Since setUpModule is called after imports we need to import conditionally.
+if IS_PY_VERSION_SUPPORTED:
+    from tuf.api.metadata import (
+        Snapshot,
+        Timestamp,
+    )
+
 
 logger = logging.getLogger(__name__)
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -23,7 +23,7 @@ IS_PY_VERSION_SUPPORTED = sys.version_info >= (3, 6)
 # Use setUpModule to tell unittest runner to skip this test module gracefully.
 def setUpModule():
     if not IS_PY_VERSION_SUPPORTED:
-        raise unittest.SkipTest("requires Python 3.6 or higher")
+        raise unittest.SkipTest('requires Python 3.6 or higher')
 
 # Since setUpModule is called after imports we need to import conditionally.
 if IS_PY_VERSION_SUPPORTED:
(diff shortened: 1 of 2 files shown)


tests/test_api.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python
+
+# Copyright 2020, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+""" Unit tests for api/metdata.py
```
trishankatdatadog:
```suggestion
""" Unit tests for api/metadata.py
```

sechkova:
_I will be leaving some thoughts here while looking at the way the client side (Updater) may use the api/metadata._

- There are several places in the code where Metadata needs to be loaded by passing a file object instead of a filename, in particular when Updater downloads temporary metadata files and verifies them. Should we plan for adding another method like `read_from_json_object` or we may use the Metadata() constructor directly by passing the file object already serialised as a dictionary and not complicating this api?

lukpueh:
> * There are several places in the code where Metadata needs to be loaded by passing a file object instead of a filename, in particular when Updater downloads temporary metadata files and verifies them. Should we plan for adding another method like `read_from_json_object` or we may use the Metadata() constructor directly by passing the file object already serialised as a dictionary and not complicating this api?

Great idea! What about the following interface?

```python
# Classmethod on all classes
def from_dict(json_dict):
    """
    Initialize object from Python dictionary, and call from_dict for any
    inner complex attribute that is represented by a class.

    In the case of Metadata.from_dict(), it must dispatch to the corresponding
    from_dict of the base class of Signed, identified by
    json_dict['signed']['_type'].

    NOTE: This is basically a mix of the constructor and load_from_json and
    the inverse of `as_dict` in the current version of metadata.py. """

# Classmethods on Metadata (and optionally on Signed)
def from_json(json_object):
    """Deserialize a json string and pass to from_dict(). """

def from_json_file(filename):
    """Read file from storage and pass to from_json(). """
```

tuf/api/metadata.py
```diff
@@ -94,51 +90,68 @@ def as_json(self, compact: bool = False) -> None:
...
-    #                     break
-
-    #     return len(verified_keyids) >= key_ring.threshold.least
+    def sign(self, key: JsonDict, append: bool = False) -> JsonDict:
+        """Creates signature over 'signed' and assigns it to 'signatures'.
+
+        Arguments:
+            key: A securesystemslib-style private key object used for signing.
+            append: A boolean indicating if the signature should be appended
+                to the list of signatures or replace them.
```
```suggestion
                to the list of signatures or replace any existing signatures.
                The default behaviour is to replace signatures.
```

tuf/api/metadata.py
```diff
@@ -94,51 +90,68 @@ def as_json(self, compact: bool = False) -> None:
...
+                Signing errors.
+
+        Returns:
+            A boolean indicating if all identified signatures are valid. False
+            if no signature was found for the keyid or any of the found
+            signatures is invalid.
+
+            FIXME: Is this behavior expected? An alternative approach would be
+            to raise an exception if no signature is found for the keyid,
+            and/or if more than one sigantures are found for the keyid.
```
```suggestion
            and/or if more than one signatures are found for the keyid.
```

lukpueh:
Hey @sechkova, I added an `Metadata.from_json(metadat_json_string)` in a7a33ce. Let me know if this is what you had in mind.  1cdae24 generally cleans up the deserialization interface as I proposed in https://github.com/theupdateframework/tuf/pull/1112#issuecomment-683701445 (see detailed commit message).

@woodruffw, @joshuagl, @trishankatdatadog, I'd say this is ready for review aka. throwing rocks at it. It still misses some features  but what's there is tested, documented and should serve our purposes for PEP 458. I think we can work off the todo list from `metadata.py`'s doc header independently in follow-up PRs.

tuf/api/metadata.py
```diff
@@ -0,0 +1,539 @@
...
+    Objects with base class Signed are usually included in a Metablock object
+    on the signed attribute. This class provides attributes and methods that
+    are common for all TUF metadata types (roles).
+
+    Attributes:
+        _type: The metadata type string.
+        version: The metadata version number.
+        spec_version: The TUF specification version number (semver) the
+            metadata format adheres to.
+        expires: The metadata expiration date in 'YYYY-MM-DDTHH:MM:SSZ' format.
```
lukpueh: Now a `datetime`object.

tuf/api/metadata.py
```diff
@@ -0,0 +1,539 @@
...
+
+        # TODO: Should we separate data validation from constructor?
+        if version < 0:
+            raise ValueError(f'version must be < 0, got {version}')
+        self.version = version
+
+
+    # Deserialization (factories).
+    @classmethod
+    def from_dict(cls, signed_dict) -> 'Signed':
```
```suggestion
    def from_dict(cls, signed_dict: JsonDict) -> 'Signed':
```

tuf/api/metadata.py
```diff
@@ -0,0 +1,539 @@
...
+ * Validation (some thoughts ...)
+   - Avoid schema, see secure-systems-lab/securesystemslib#183
+   - Provide methods to validate JSON representation (at user boundary)
+   - Fail on bad json metadata in read_from_json method
+   - Be lenient on bad/invalid metadata objects in memory, they might be
+     work in progress. E.g. it might be convenient to create empty metadata
+     and assign attributes later on.
+   - Fail on bad json metadata in write_to_json method, but with option to
+     disable check as there might be a justified reason to write WIP
+     metadata to json.
```
lukpueh: Btw. what do others think about the [ValidationMixin](https://github.com/in-toto/in-toto/blob/74da7aabce663a5539ad3e1e17b2c871eeaa227a/in_toto/models/common.py#L27-L40) methods we use in in-toto? (also see [`Layout._validate_*`](https://github.com/in-toto/in-toto/blob/74da7aabce663a5539ad3e1e17b2c871eeaa227a/in_toto/models/layout.py#L420-L438) methods for example usage) 

I think it's a commendable way of doing this (credits to @SantiagoTorres)

trishankatdatadog (approved):
Great job, @lukpueh! 🎉 

I don't have strong opinions about this PR: it looks largely like the Metadata API I originally proposed, except for changes like removing KeyRing and Threshold so on (we can argue about that later). As long as @woodruffw and @joshuagl  also approve, I'm 👍🏽

tests/test_api.py
```diff
@@ -0,0 +1,165 @@
...
+    @classmethod
+    def setUpClass(cls):
+
+        # Create a temporary directory to store the repository, metadata, and target
+        # files.  'temporary_directory' must be deleted in TearDownClass() so that
+        # temporary files are always removed, even when exceptions occur.
+        cls.temporary_directory = tempfile.mkdtemp(dir=os.getcwd())
+
+        test_repo_data = os.path.join(
+                os.path.dirname(os.path.realpath(__file__)), 'repository_data')
```
joshuagl, line 57: 8 space indentation on a continuation?

tuf/api/metadata.py
joshuagl: That `ValidationMixin` looks nice! I'd be happy to see it used here in tuf

tuf/api/metadata.py
```diff
@@ -0,0 +1,539 @@
...
+
+    # Deserialization (factories).
+    @classmethod
+    def from_dict(cls, signed_dict) -> 'Signed':
+        """Creates Signed object from its JSON/dict representation. """
+
+        # Convert 'expires' TUF metadata string to a datetime object, which is
+        # what the constructor expects and what we store. The inverse operation
+        # is implemented in 'to_dict'.
+        signed_dict['expires'] = iso8601.parse_date(
```
joshuagl: Do we still need an iso8601 dependency if we're only targeting Python 3.5+, can we just use `datetime.fromisoformat()` here?

tests/test_api.py
```diff
@@ -3,7 +3,12 @@
 # Copyright 2020, New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 """ Unit tests for api/metdata.py
+
+Skipped on Python < 3.6.
```
joshuagl: I presume the goal would be to revert this patch once all Python 2.7 supporting code has been removed?

tuf/api/metadata.py
```diff
@@ -186,6 +206,21 @@ def write_to_json(
 
 
 class Signed:
+    """A base class for the signed part of TUF metadata.
+
+    Objects with base class Signed are usually included in a Metablock object
```
```suggestion
    Objects with base class Signed are usually included in a Metadata object
```

tuf/api/metadata.py
joshuagl: I prefer the alternative approach of raising an exception if no signature is found, or more than one signature is found, for the keyid.

joshuagl (commented):
Very nice streamlining of the metadata API PR. Thanks @lukpueh! I have a few minor questions and comments inline. I'm also curious to understand why `version` and `spec_version` have different types? `version: int` and `spec_version: str`, is there any reason `spec_version` can't be an int?

tuf/api/metadata.py
lukpueh: I think I do too. Any advice on the exception type? At some point we should discuss [exception taxonomy guidelines](https://github.com/theupdateframework/tuf/projects/1#card-44476165). But we shouldn't block on that.

joshuagl: We don't seem to have an exception in `tuf.exceptions` or `securesystemslib.exceptions` that would fit, we could either throw a generic exception _or_ define a new one. Either way, it will probably need some cleanup once guidelines are defined.

tests/test_api.py
lukpueh: Hm. Good question. I always thought [PEP 8 required double-indentation for hanging indent](https://www.python.org/dev/peps/pep-0008/#indentation), but it is only a suggestion, and also only to distinguish it from a subsequent nested block. 

IMO it's a good idea to always differentiate line continuation indentation from functional indentation, even if there is no subsequent nested block. But I'm fine either way.

FYI, the [google style guide does not say anything about double indent](https://google.github.io/styleguide/pyguide.html#indentation) but has some [examples that could be interpreted that way](https://google.github.io/styleguide/pyguide.html#s3.2-line-length).

joshuagl: > IMO it's a good idea to always differentiate line continuation indentation from functional indentation, even if there is no subsequent nested block. But I'm fine either way.
> 
That seems like a reasonable coding style for us to adopt. We should document it and teach the linter about it 😃

lukpueh:
> version: int and spec_version: str, is there any reason spec_version can't be an int?

Yes, `spec_version` is in semver format.

lukpueh:
Thanks for the reviews, @joshuagl and @trishankatdatadog!. I addressed the minor comments in 0a9fc93, which I'm happy to squash with previous commits before merging. The more serious change of `Metadata.verify` behaviour is in a separate commit, to make review easier. :)

tests/test_api.py
```diff
@@ -19,6 +17,7 @@
 from datetime import datetime, timedelta
 from dateutil.relativedelta import relativedelta
 
+# TODO: Remove case handling when fully dropping support for versions >= 3.6
```
joshuagl, line 20:
```suggestion
# TODO: Remove case handling when fully dropping support for versions < 3.6
```

tuf/api/metadata.py
```diff
@@ -39,11 +45,7 @@
 )
 from securesystemslib.storage import StorageBackendInterface
 from securesystemslib.keys import create_signature, verify_signature
-from tuf.repository_lib import (
-    _strip_version_number
-)
 
-import iso8601
```
joshuagl: 🎉

tuf/api/metadata.py
```diff
@@ -308,14 +311,15 @@ def __init__(
...
         """Creates Signed object from its JSON/dict representation. """
 
         # Convert 'expires' TUF metadata string to a datetime object, which is
         # what the constructor expects and what we store. The inverse operation
         # is implemented in 'to_dict'.
-        signed_dict['expires'] = iso8601.parse_date(
-                signed_dict['expires']).replace(tzinfo=None)
+        signed_dict['expires'] = datetime.strptime(
+                signed_dict['expires'],
+                "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=None)
```
joshuagl, line 299: Awesome, thanks for adding this!

joshuagl (approved):
Thanks @lukpueh, this looks solid. I'm very happy to see this long-discussed model implemented.

One minor typo in a comment, but otherwise this looks good to rebase and merge.

tuf/api/metadata.py
joshuagl: We have three uses of `iso8601.parse_date` in the current codebase, so this might be a good candidate for a helper function in future.

tests/test_api.py
lukpueh: 🤦 thanks for catching this!

lukpueh:
Thanks again for the reviews! I just cleaned up the git history, fixed a minor inconsistency and removed the todo list after created GitHub issues for each item.
The diff of the new force-pushed tip to your last review can be seen here: https://github.com/lukpueh/tuf/compare/simple-tuf-api-before-rebase..simple-tuf-api

Issue #1120, opened 2020-09-08
logging in test runner / CI : higher log level but only print failing tests


jku:
We should do what in-toto does: have the test runner output much more logging but only for failing tests. This should be especially useful for CI but also for developers.

Assuming we hard-code a good log level (see #1093 for potential alternatives), I think we only need roughly this in aggregate_tests.py:

```
class TestHandler(logging.StreamHandler):
  def __init__(self):
    super().__init__(self)

  @property
  def stream(self):
    return sys.stderr

  @stream.setter
  def stream(self, value):
    pass

if __name__ == '__main__':
  logging.basicConfig(level=logging.INFO, handlers=[TestHandler()])
  # current testrunner code here...
```

This is because the testrunner buffering will not work with long lived loggers otherwise: the default handler caches `sys.stderr` value and testrunner wants to modify it between test runs (and buffering is required to only print the failing test output).

jku:
I have a commit for this (see link above) but it might require some discussion and work:
* My original idea was to increase the default tox log level significantly (to INFO or even DEBUG) and this is what the branch does: The logic is that on CI successful test output should be hidden but failing tests should output as much as possible to give the needed context. However, I'm not sure if other developers use tox in their development workflows and if it should _not_ print that much by default (and instead CI should have a separate tox config that uses higher log level)?
* The unittest buffering does not seem to work properly on python 2? This means logs from successful tests are not hidden so there's a huge amount of output, even at INFO level
* Failing test output is currently printed twice: first all of the output from all failing tests is printed, then individual test output is printed nicely separated by test names and results. We would like to prevent former and only have the latter but this seems to be a unittest issue: in-toto has the same problem

I should maybe make a draft PR of this and make some tests fail to showcase this...

MVrachev:
> My original idea was to increase the default tox log level significantly (to INFO or even DEBUG) and this is what the branch does: The logic is that on CI successful test output should be hidden but failing tests should output as much as possible to give the needed context. However, I'm not sure if other developers use tox in their development workflows and if it should not print that much by default (and instead CI should have a separate tox config that uses higher log level)?

We cannot assess how all developers are using tox, but probably the best we can do is to ask that on a tuf meeting.
[...]

PR #1145, opened 2020-09-15
Silence errors to be raised

### LINKED ISSUE #1093: Option to show log output when running individual tests
In #1083 @sechkova dug into the current logging behaviour and documented that 

1) Executing the tests by `aggregate_tests.py` buffers the output on a successful run. 
2) Executing a test "by hand", the logging output remains hidden until an error occurs.

IMNSHO we should only buffer individual tests run by hand when the -b command-line option is passed.

### LINKED ISSUE #1108: logging level consistency
There are still some logging annoyances that make it hard to A) see actual errors in e.g. unittest logging but also B) difficult to decide what to do WRT tuf logging in an application:
* tuf logs things as errors even when things are working fine: e.g. tuf.download and tuf.client.updater do this when a N.root.json is not found
* tuf logs things as errors when it has correctly found a security issue: e.g. BadSignatureError from tuf.client.updater. This one is more debatable but I'd argue the correct response is an exception for the application to handle and no ERROR output: tuf is working exactly as expected in this case

There have been some improvements in this #1083, #1039, #1093, #983 but the core issue is still there: In my opinion ERROR and CRITICAL should only be used when tuf is in an error state, not when it has correctly identified a security issue.

For testing this I suggest modifying e.g. test_update.py:

```
if __name__ == '__main__':
  verbose = '-v' in sys.argv or '--verbose' in sys.argv
  loglevel = logging.DEBUG if verbose else logging.ERROR
  logging.basicConfig()
  tuf.log.set_log_level(loglevel)

  unittest.main()

```
and running `python3 test_updater.py TestUpdater` (this will only print ERROR and higher levels)

jku:
**Fixes  #1108, fixes #1093**

This does not fix every logging inconsistency but it does fix issues that become visible in normal usage when running the updater at log level ERROR.

I decided to not try issue #1120 in the same PR (as it might require more discussion), but did fix #1093 (_Option to show log output when running individual tests_) because that is needed to easily verify these fixes. Now you can run e.g.
 
    python3 test_updater.py -vvvv TestUpdater.test_4_refresh

to run a specific test with DEBUG level output. The default level is ERROR.

* Avoid logging inside the library when all info is included in the exception that will be raised
* When logging, avoid using log levels ERROR, CRITICAL (and the exception logger) when tuf is not actually in error or critical state itself.
* remove test_init.py and test_exceptions.py: they do not seem to be tests
* Configure logging for all individual test files

I still have these complaints/questions (but personally can live with the PR as is):
* test_log.py is strange: I'm not 100% sure how it works or should work, but I dislike it printing errors and warnings on normal operation
* repository_tool and repository_lib modify log handlers and their log levels. The fact that just importing repository_tool affects logging is fishy... but I did not want to modify this here as I'm not sure if there are any visible issues
* Unittest already uses "-v": this is fine except that the effects I added are now a little hard to discover -- I did not want to implement "--help" since that would hide the help output from unittest itself...

requirements**:


tuf/client/updater.py
```diff
@@ -1142,12 +1142,10 @@ def neither_403_nor_404(mirror_error):
           # Otherwise, reraise the error, because it is not a simple HTTP
           # error.
           if neither_403_nor_404(mirror_error):
-            logger.exception('Misc error for root version '+str(next_version))
+            logger.info('Misc error for root version '+str(next_version) + str(exception))
```
MVrachev: Nit: missing space between those two `'Misc error for root version '+str(next_version)`

tuf/client/updater.py
```diff
@@ -1142,12 +1142,10 @@ def neither_403_nor_404(mirror_error):
...
             raise
           else:
-            # Calling this function should give us a detailed stack trace
-            # including an HTTP error code, if any.
-            logger.exception('HTTP error for root version '+str(next_version))
+            logger.debug('HTTP error for root version '+str(next_version))
```
MVrachev: Nit: missing space between those two `'HTTP error for root version '+str(next_version)`

MVrachev:
Looks good to me.

jku:
Something that may not be obvious: Running some tests like test_updater.py (at default ERROR level) is currently still very chatty because we get a lot of warnings.  Warnings are not integrated into the logging system by default, but we should be able to make them into e.g. WARNING level log entries quite easily: then the default output would be quite nice.

However, ideally we should fix the warnings instead of hiding them so I've not done the integration here (_EDIT: this was annoying enough that I finished my ResourceWarning cleanup: #1147_).

tests/utils.py
```diff
@@ -74,3 +77,23 @@ def wait_for_server(host, port, timeout=10):
...
 
 
+def configure_test_logging(argv):
+  # parse arguments but only handle '-v': argv may contain 
+  # other things meant for unittest argument parser
+  parser = argparse.ArgumentParser(add_help=False)
+  parser.add_argument('-v', '--verbose', action='count', default=0)
+  args, _ = parser.parse_known_args(argv)
+  
+  if args.verbose <= 1:
```
MVrachev, line 87: In your commit message, you have mentioned that the number of `v` means the following:
```
default: ERROR
"-v":    ERROR, but unittest prints test names
"-vv":   WARNING
"-vvv":  INFO
"-vvvv": DEBUG
```

I think this should be documented somewhere.
Can we have like `--help` argument and functionality?

MVrachev:
Looks good to me besides my comment.

I am glad you have removed `tests/test_exceptions.py` and `tests/test_init.py`.
I don't quite understand the need for those tests...

tests/utils.py
jku: I wish we could... but I don't see a way to do that without breaking unittest --help. The lack of discoverability is annoying but I just don't know how to make it better (without essentially hiding unittest functionality, which I also don't want to do)

joshuagl:
> * remove test_init.py and test_exceptions.py: they do not seem to be tests

I half-expected this to change the coverage, but it's still at 97% 🤷‍♂️ 

> * repository_tool and repository_lib modify log handlers and their log levels. The fact that just importing repository_tool affects logging is fishy... but I did not want to modify this here as I'm not sure if there are any visible issues

I think this one is worth an issue, would you mind creating one?

jku:
#1148 is about `import repository_tool` side effects

tuf/download.py
```diff
@@ -278,7 +278,7 @@ def _download_file(url, required_length, STRICT_REQUIRED_LENGTH=True):
   except Exception:
     # Close 'temp_file'.  Any written data is lost.
     temp_file.close()
-    logger.exception('Could not download URL: ' + repr(url))
+    logger.debug('Could not download URL: ' + repr(url))
```
joshuagl, line 281: Can you elaborate on why you chose debug level here?

joshuagl (approved):
Great set of cleanups for logging, thanks @jku. I have a question in-line about how you're choosing logging levels, I think I figured it out in most cases apart from that one.

tuf/download.py
jku: I don't have a strong opinion here, but I think the reason I went with debug() was to avoid info() for things that 
* do happen in normal usage
* are then raised as exceptions and handled in calling code: in this case Updater will usually log 'Update failed from ...' also at debug level

jku: just so I'm not misleading: 'Update failed from ...' log level was also chosen by me in this PR.

The point that I tried to make was that the 'lower level code' (download) probably should not log at INFO if the 'higher level code' (Updater) logs the same thing at DEBUG. A case could be made for the Updater logs for failed updates to be INFO though...

joshuagl: And that logging in updater is only a DEBUG because an exception is raised if all mirrors fail. This makes sense, thanks for providing that extra context.

I agree, a case could be made for info. But I'm not making that case here.

Issue #1077, opened 2020-07-13
os.path.join() used with URLs in Updater


jku:
mirrors.py does this:

    base = os.path.join(mirror_info['url_prefix'], mirror_info['metadata_path'])

This seems like an incorrect use of `os.path.join()`. I'm not sure what it does on Windows but there's probably room for bugs here. `urllib.parse.join()` is probably the safe choice (although it has some annoying corner cases that makes it different from `os.path.join()`) ?

There is a related issue here: I'm trying to figure out how to handle the case where metadata and targets are hosted on different hosts and the use of `os.path.join()` affects it. The discussion is in https://python.zulipchat.com/#narrow/stream/223926-pep458-implementation/topic/separate.20hosting.20for.20metadata.20and.20targets . Fixing this probably makes sense after we know the resolution of that discussion.

jku:
I think we actually have multiple os separator bugs in updater that end up somewhat compensating for each other...
* get_list_of_mirrors() uses `os.path.join()` to build a URL which is wrong on Windows, but tries to fix that by running `url.replace('\\', '/')` in the end
* Updater._get_target_file() and Updater._update_metadata() use `os.path.join()` to create the final target path (which should be a relative URL/posix path). This probably creates a broken path on Windows with Windows separators -- but this path is later fed to get_list_of_mirrors() which 'fixes' this by doing the mentioned replace
* updated_targets() and download_targets() use the target file path (a relative URL/posix path) to build an OS path to save the file in... I wonder what does on windows? _Edit: this actually probably works: python seems to swap the separators correctly_ 
* download_targets() also strips os.sep from the target file path (which again should be a relative posix path)

I'm pretty sure it will be impossible to fix all of these in a way that wouldn't break some client mirror configuration where Windows separators were used in an innovative way... but probably it's still worth doing: If config like that exists it is just waiting to be broken

I no longer agree with "good first issue" label  :) (well, some items on the list may be easy, some are tricky)

sechkova:
I think the reference implementation should adopt the same strategy when dealing with URLs/paths on the client side as well as on the repository side. And while the specification at least recommends the format of paths in the metadata:

> To avoid surprising behavior when matching targets with PATHPATTERN, it is RECOMMENDED that PATHPATTERN uses the forward slash (/) as directory separator and does not start with a directory separator, akin to TARGETSPATH.

mirrors.py is a fully implementation product ... 

[...]

PR #1153, opened 2020-09-25
mirrors: Make targets_path and metadata_path optional

### LINKED ISSUE #1079: Updater: mirrors configuration tweaks
For the Warehouse/pip use case we are not so much interested multiple actual mirrors but separating the target and metadata hosting to two hosts.

Current Updater mirror configuration is close but not quite up to the task:

    mirrors = {
        'mirror1': {
            'url_prefix': 'https://hostname/basedir/',
            'metadata_path': 'metadata',
            'targets_path': 'targets',
            'confined_target_dirs': ['']
        },
    }

I would like to have a way of signaling that a particular mirror should not be used for metadata or should not be used for targets (`'confined_target_dirs': []` does offer that, but only for targets). There's multiple ways to solve this: I'm open to suggestions.

The mirrors role (which is unrelated to this but deals with the same problem) solves the issue by having PATH_PATTERNS for both metadata and targets, but for this specific use case it would be enough if metadata_path and targets_path could be set to None. Then I could do this in pip:

    mirrors = {
        'metadata': {
            'url_prefix': 'https://pypi.org/',
            'metadata_path': 'tuf',
            'targets_path': None,
            'confined_target_dirs': []
        },
        'targets': {
            'url_prefix': 'https://files.pythonhosted.org/',
            'metadata_path': None,
            'targets_path': 'packages',
            'confined_target_dirs': ['']
        },
    }

This would probably be a trivial change in mirrors.py and should not break any existing users

jku:
Now clients can leave out targets_path or metadata_path if the
client knows the mirror does not have that type of targets.

This is backwards compatible: old mirror configs continue to work.

Fixes #1079.

--- 

**PR notes:**

This is a minor improvement in the warehouse case where one mirror has metadata and targets, and the other mirror has only targets.

Note that this leaves the issue of using os.path.join() on URLs as is: Fixing #1077 in a provably backwards compatible way seems tricky.

I believe this is a good if minor change... but I'll make the case against merging as well:
* the usefulness in pip/warehouse is minimal: the directory confinement is not expressive enough to be useful so pip has to use two separate mirror configurations anyway -- this should mean that in normal operation no requests will be made to the 'wrong' mirror: I believe they would only happen if the metadata mirror does not respond to a request for e.g. a specific targets bin file. In that case the wrong mirror would be asked as well -- this would be an annoyance, not a security issue
* it can be argued that the correct solution is to make directory confinement more powerful instead:
  * it should work on both metadata and targets
  * it should be expressive enough to say "allow everything under this directory"

Unfortunately improving directory confinement is likely to be a breaking change.

requirements**:

### DIFF
--- a/tests/test_mirrors.py
+++ b/tests/test_mirrors.py
@@ -58,21 +58,29 @@ class TestMirrors(unittest_toolbox.Modified_TestCase):
                  'confined_target_dirs' : ['targets/release/',
                                             'targets/release/']},
      'mirror3': {'url_prefix' : 'http://mirror3.com',
-                 'metadata_path' : 'metadata',
                  'targets_path' : 'targets',
-                 'confined_target_dirs' : ['targets/release/',
-                                            'targets/release/']}}
+                 'confined_target_dirs' : ['targets/release/v2/']},
+     'mirror4': {'url_prefix' : 'http://mirror4.com',
+                 'metadata_path' : 'metadata',
+                 'confined_target_dirs' : []},
+    }
 
 
 
   def test_get_list_of_mirrors(self):
     # Test: Normal case.
+
+    # 3 matches: Metadata found on 3 mirrors
     mirror_list = mirrors.get_list_of_mirrors('meta', 'release.txt', self.mirrors)
     self.assertEqual(len(mirror_list), 3)
-    for mirror, mirror_info in six.iteritems(self.mirrors):
-      url = mirror_info['url_prefix'] + '/metadata/release.txt'
-      self.assertTrue(url in mirror_list)
+    self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/metadata/release.txt' in \
+                    mirror_list)
+    self.assertTrue(self.mirrors['mirror2']['url_prefix']+'/metadata/release.txt' in \
+                    mirror_list)
+    self.assertTrue(self.mirrors['mirror4']['url_prefix']+'/metadata/release.txt' in \
+                    mirror_list)
 
+    # 1 match: a mirror without target directory confinement
     mirror_list = mirrors.get_list_of_mirrors('target', 'a.txt', self.mirrors)
     self.assertEqual(len(mirror_list), 1)
     self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a.txt' in \
@@ -83,11 +91,20 @@ class TestMirrors(unittest_toolbox.Modified_TestCase):
     self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a/b' in \
                     mirror_list)
 
+    # No matches
     mirror1 = self.mirrors['mirror1']
     del self.mirrors['mirror1']
     mirror_list = mirrors.get_list_of_mirrors('target', 'a/b', self.mirrors)
     self.assertFalse(mirror_list)
     self.mirrors['mirror1'] = mirror1
+
+    # 2 matches: One with non-confined targets and one with matching confinement
+    mirror_list = mirrors.get_list_of_mirrors('target', 'release/v2/c', self.mirrors)
+    self.assertEqual(len(mirror_list), 2)
+    self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/release/v2/c' in \
+                    mirror_list)
+    self.assertTrue(self.mirrors['mirror3']['url_prefix']+'/targets/release/v2/c' in \
+                    mirror_list)
 
     # Test: Invalid 'file_type'.
     self.assertRaises(securesystemslib.exceptions.Error, mirrors.get_list_of_mirrors,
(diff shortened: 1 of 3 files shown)


jku:
In case it's relevant, these are the configs I keep currently switching in pip depending on what I'm downloading:

config 1 (for downloading index file targets from pypi.org):
```
        {
            'https://pypi.org': {
                'url_prefix': 'https://pypi.org',
                'metadata_path': 'tuf/',
                'targets_path': 'simple/',
                'confined_target_dirs': ['']
            }
        }
```
config 2 (for downloading distribution target files from pythonhosted.org):
```
        {
            'https://pypi.org': {
                'url_prefix': 'https://pypi.org',
                'metadata_path': 'tuf/',
                'targets_path': 'not_real_path/', # TODO should not exist
                'confined_target_dirs': []
            },
            'https://files.pythonhosted.org/packages': {
                'url_prefix': 'https://files.pythonhosted.org/packages',
                'metadata_path': 'not_real_path', # TODO should not exist
                'targets_path': '',
                'confined_target_dirs': ['']
            }
        }
```

jku:
> directory confinement is not expressive enough to be useful so pip has to use two separate mirror configurations anyway

Actually, it's possible that even a very expressive directory confinement might not be able to handle the warehouse setup in a single config... I'd have to really think the solution through (but maybe it's not needed for this PR)

tests/test_mirrors.py
```diff
@@ -58,21 +58,29 @@ def setUp(self):
                  'confined_target_dirs' : ['targets/release/',
                                             'targets/release/']},
      'mirror3': {'url_prefix' : 'http://mirror3.com',
-                 'metadata_path' : 'metadata',
                  'targets_path' : 'targets',
-                 'confined_target_dirs' : ['targets/release/',
-                                            'targets/release/']}}
+                 'confined_target_dirs' : ['targets/release/v2/']},
```
MVrachev, line 62: Why did you have to change that?

MVrachev (approved):
Аside from my comment LGTM.

tests/test_mirrors.py
jku: ah, that could maybe have been a separate change: there was nothing testing if confinement matches subdirectories (it does not). 
The test I added (one with 2 matches) tests that while it also makes sure things work without a metadata_path.

MVrachev: > The test I added (one with 2 matches) tests that while it also makes sure things work without a metadata_path.

It tests that you have added makes sure things work without a metadata_path, yes, but you haven't tested if confinement matches subdirectories.

I don't think we need such a test, because to check if a directory is confinement in a list is a function of securesystemslib and it's tested there:
https://github.com/secure-systems-lab/securesystemslib/blob/94d83b943f2bcc020619547d83cc037bd78dfb61/tests/test_util.py#L165

jku: > It tests that you have added makes sure things work without a metadata_path, yes, but you haven't tested if confinement matches subdirectories.

If subdirectories did match, mirror2 would be in the results -> test would fail.

> I don't think we need such a test, because to check if a directory is confinement in a list is a function of securesystemslib

That's an implementation detail of `get_list_of_mirrors()`... if we change the implementation (or if securesystemslib changes their implementation) we'd want to know if results change.

In any case: I'm not adding a confinement test here: I'm adding a test for missing metadata_path that also happens to check that confinement works they way it was designed to.

MVrachev: > If subdirectories did match, mirror2 would be in the results -> test would fail.

Yes, you are right.
Maybe it will be even better if you add a comment that subdirectories are not expected to match?
It's possible that this is well documented on multiple places.
That's a NIT and it's your decision. 

> That's an implementation detail of get_list_of_mirrors()... if we change the implementation (or if securesystemslib changes their implementation) we'd want to know if results change.
> In any case: I'm not adding a confinement test here: I'm adding a test for missing metadata_path that also happens to check that confinement works they way it was designed to.

Yes, you are right.

tuf/formats.py
```diff
@@ -403,8 +403,8 @@
 MIRROR_SCHEMA = SCHEMA.Object(
   object_name = 'MIRROR_SCHEMA',
   url_prefix = securesystemslib.formats.URL_SCHEMA,
-  metadata_path = RELPATH_SCHEMA,
-  targets_path = RELPATH_SCHEMA,
+  metadata_path = SCHEMA.Optional(RELPATH_SCHEMA),
+  targets_path = SCHEMA.Optional(RELPATH_SCHEMA),
```
joshuagl, line 407: Prior to this change a `check_match()` call will ensure that `both metadata_path` and `targets_path` are set. But following this change it's possible to pass an object conformant to `MIRROR_SCHEMA` with neither value set.

I'm not sure whether we should worry too much about this, it should be fairly obvious to a client if they misconfigure by not setting either value?

jku: This is true, does not seem like a major issue to me either.

joshuagl (approved):
LGTM, thanks!

PR #1157, opened 2020-09-30
tuf/api: Expose tuf.api as a package

woodruffw:
This just adds `tuf/api/__init__.py`, turning it into a full package for distribution purposes.

- [N/A] Tests have been added for the bug fix or new feature
- [N/A] Docs have been added for the bug fix or new feature


trishankatdatadog (approved):

woodruffw:
Amusingly this now breaks the CI, since `tuf.api` is now correctly discovered and linted 😉

joshuagl:
According to the [pylintrc docs](https://docs.pylint.org/en/1.6.0/run.html#command-line-options)

> If the current working directory is in a Python module, Pylint searches up the hierarchy of Python modules until it finds a pylintrc file. This allows you to specify coding standards on a module-by-module basis. Of course, a directory is judged to be a Python module if it contains an \_\_init\_\_.py file.

I think we could add a `pylintrc` in tuf/api to support the style there?

lukpueh:
> I think we could add a pylintrc in tuf/api to support the style there?

That's actually a really good idea. The linter configuration would need a redo anyway even without style guide change (see https://github.com/in-toto/in-toto/pull/296#issue-297658607, which also applies here). But I'm reluctant to change the default `pylintrc` as it would require many fixes in code that won't survive the upcoming [refactor](https://github.com/theupdateframework/tuf/projects/2).

If we create a new config in `tuf/api` we can use it for new code only and make it the default later. :)

joshuagl:
I've taken this change to a new PR which also adds a linter for tuf/api https://github.com/theupdateframework/tuf/pull/1177

PR #1162, opened 2020-10-05
MANIFEST.in: Include LICENSE files

### LINKED ISSUE #1160: tar.gz release does not contain LICENSE
Release tarballs (like tuf-0.14.0.tar.gz) do not contain a LICENSE file

This prevents vendoring tuf using the `vendoring` tool: I could patch the license in but unfortunately the license file check is done on unpatched sources

`python3 setup.py sdist` produces a tarball with the LICENSE file (and many other files that are not in our release tarball) -- what's this about?

jku:
This seems to be a common way to handle license files. Also, vendoring
tool fails to handle tuf without a LICENSE file.

Fixes #1160

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
+include LICENSE*
 include README.md
 include tox.ini
 include tests/repository_data/keystore/delegation_key


joshuagl (approved):
Thanks for this

PR #1163, opened 2020-10-05
Update release process and mark unit tests as executable

joshuagl:
**Fixes issue #**: N/A

**Description of the changes being introduced by the pull request**: Following some discussion around #1160 this PR updates the release docs to recommend explicitly using Python3 for distribution generation and cleaning the tuf tree before making a release.

I've also included a patch that adds the executable bit to some of our unit tests that didn't already have it.

requirements**:


docs/RELEASE.md
```diff
@@ -1,12 +1,16 @@
 # Release process
 
+* Ensure you have a backup of all working files and then remove files not tracked by git
+  `git clean -xdf`. **NOTE**: this will delete all files in the tuf tree that aren't
```
MVrachev, line 4: Does this mean that if I am to release something, this will have to delete all my venvs in the current directory (which of course are not tracked by git) if I don't back them up?

joshuagl: Yes, it does. The theory is that it's better to ensure the directory is clean and matches the git tree, so that we don't accidentally end up packaging things we don't intend to ship.

MVrachev: Yes, it makes sense.
Just wanted to make sure I understand it right.

docs/RELEASE.md
```diff
@@ -1,12 +1,16 @@
...
+  tracked by git
 * Ensure `docs/CHANGELOG.md` contains a one-line summary of each [notable
   change](https://keepachangelog.com/) since the prior release
 * Update `setup.py` and `tuf/__init__.py` to the new version number vA.B.C
 * Test packaging, uploading to Test PyPI and installing from a virtual environment
+  (ensure commands invoking `python` below are using Python 3)
```
mnm678, line 10: I'm not sure we need this note (here and below) if the commands also say `python3`

mnm678 (approved):
One nit, but lgtm

docs/RELEASE.md
lukpueh: @MVrachev, you could also keep your venvs in an [`ENV_DIR`](https://docs.python.org/3/library/venv.html) that's not under cwd. :)

docs/RELEASE.md
lukpueh: I think it's okay to be very clear about it. But I'm fine either way.

lukpueh (approved):

docs/RELEASE.md
joshuagl: I opted to be very clear, just in case readers are not diligently reading each step

docs/RELEASE.md
jku: @MVrachev the alternative I've used is that I never do releases from my development directory: I make a new git clone for every release. This ensures two things
1. all files required for the release actually are in git 
1. no files from outside of git end up in the release

trishankatdatadog:
Haven't read everything, but is it possible to codify the release process in a Python script? Would make it a lot more automatic and less error-prone...

lukpueh:
> Haven't read everything, but is it possible to codify the release process in a Python script?

That's a great idea, @trishankatdatadog! Mind creating a ticket? I think it's out of the scope of this PR, because it needs some tinkering in regards to non-automatable (e.g. "update the changelog", ...), or slightly harder to automate ("check test.pypi.org", "create release on GitHub", ...) steps.

Speaking of steps, as follow-up work we might want to consider creating and publishing in-toto link metadata for all of these and a corresponding layout too. (see https://github.com/in-toto/in-toto/issues/278 for some thoughts about doing this for the very similar release process of in-toto itself).

joshuagl:
> Speaking of steps, as follow-up work we might want to consider creating and publishing in-toto link metadata for all of these and a corresponding layout too. (see [in-toto/in-toto#278](https://github.com/in-toto/in-toto/issues/278) for some thoughts about doing this for the very similar release process of in-toto itself).

Absolutely. That is one of several tasks I've been wanting to get around to for some time.

PR #1166, opened 2020-10-06
Make confined_target_dirs optional field

### LINKED ISSUE #1155: Make confined_target_dirs optional field in  the MIRROR_SCHEMA
I was thinking about why `confined_target_dirs` is not an optional field in the MIRROR_SCHEMA.
https://github.com/theupdateframework/tuf/blob/fdccb8dc0bf34e2b016ce20f1570e582dcd459e7/tuf/formats.py#L408

Given that `['']` matches everything and `[]` matches nothing this API can be confusing and lead to a mistake.
For me, it sounds more logical to not provide `confined_target_dirs` list at all if I want to accept queries for all targets for that mirror and provide a list of `confined_target_dirs` if it's otherwise.

**Current behavior**:
`confined_target_dirs` is a mandatory field in the MIRROR_SCHEMA.

**Expected behavior**:

`confined_target_dirs`  to be an optional field in the MIRROR_SCHEMA.

MVrachev:
Fixes: https://github.com/theupdateframework/tuf/issues/1155

The field confined_target_dirs from the MIRROR_SCHEMA  is
a list of strings. Those strings define the accessible target
paths for that mirror. For one target to be available for that mirror,
its path should have as a prefix at least one of the strings defined
in confined_target_dirs.

That's why when confined_target_dirs is a list with one element empty
string (e.g. `['']`) this means all targets files on that mirror are
available and if confined_target_dirs is empty list (e.g. `[]`) this
would be interpreted as none of the target files is available.

This is a confusing API that could easily lead to mistakes.
That's why it's better we promote to not set confined_target_dirs
at all if a user wants targets to be available.

requirements**:

### DIFF
--- a/tests/test_mirrors.py
+++ b/tests/test_mirrors.py
@@ -50,8 +50,7 @@ class TestMirrors(unittest_toolbox.Modified_TestCase):
     self.mirrors = \
     {'mirror1': {'url_prefix' : 'http://mirror1.com',
                  'metadata_path' : 'metadata',
-                 'targets_path' : 'targets',
-                 'confined_target_dirs' : ['']},
+                 'targets_path' : 'targets'},
      'mirror2': {'url_prefix' : 'http://mirror2.com',
                  'metadata_path' : 'metadata',
                  'targets_path' : 'targets',
--- a/tests/test_mirrors.py
+++ b/tests/test_mirrors.py
@@ -59,15 +59,47 @@ class TestMirrors(unittest_toolbox.Modified_TestCase):
      'mirror3': {'url_prefix' : 'http://mirror3.com',
                  'targets_path' : 'targets',
                  'confined_target_dirs' : ['targets/release/v2/']},
+     # confined_target_dirs = [] means that none of the targets on
+     # that mirror is available.
      'mirror4': {'url_prefix' : 'http://mirror4.com',
                  'metadata_path' : 'metadata',
                  'confined_target_dirs' : []},
+     # Make sure we are testing when confined_target_dirs is [''] which means
+     # that all targets are available on that mirror.
+     'mirror5': {'url_prefix' : 'http://mirror5.com',
+                 'targets_path' : 'targets',
+                 'confined_target_dirs' : ['']}
     }
 
 
 
   def test_get_list_of_mirrors(self):
     # Test: Normal case.
+
+    # 1 match: a mirror without target directory confinement
+    mirror_list = mirrors.get_list_of_mirrors('target', 'a.txt', self.mirrors)
+    self.assertEqual(len(mirror_list), 2)
+    self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a.txt' in \
+                    mirror_list)
+    self.assertTrue(self.mirrors['mirror5']['url_prefix']+'/targets/a.txt' in \
+                    mirror_list)
+
+    mirror_list = mirrors.get_list_of_mirrors('target', 'a/b', self.mirrors)
+    self.assertEqual(len(mirror_list), 2)
+    self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a/b' in \
+                    mirror_list)
+    self.assertTrue(self.mirrors['mirror5']['url_prefix']+'/targets/a/b' in \
+                    mirror_list)
+
+    # 2 matches: One with non-confined targets and one with matching confinement
+    mirror_list = mirrors.get_list_of_mirrors('target', 'release/v2/c', self.mirrors)
+    self.assertEqual(len(mirror_list), 3)
+    self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/release/v2/c' in \
+                    mirror_list)
+    self.assertTrue(self.mirrors['mirror3']['url_prefix']+'/targets/release/v2/c' in \
+                    mirror_list)
+    self.assertTrue(self.mirrors['mirror5']['url_prefix']+'/targets/release/v2/c' in \
+                    mirror_list)
 
     # 3 matches: Metadata found on 3 mirrors
     mirror_list = mirrors.get_list_of_mirrors('meta', 'release.txt', self.mirrors)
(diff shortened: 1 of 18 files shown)


joshuagl (commented):
Looks good, thanks Martin. It would be good to ensure we're testing that the existing behaviour when `confined_target_dirs: ['']` continues to work. I don't think that's covered by test_mirrors.py any more with this PR?

MVrachev:
I added a commit testing that.

tests/test_mirrors.py
```diff
@@ -60,30 +60,36 @@ def setUp(self):
...
+    self.assertEqual(len(mirror_list), 2)
     self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a/b' in \
                     mirror_list)
 
     mirror1 = self.mirrors['mirror1']
     del self.mirrors['mirror1']
+    del self.mirrors['mirror4']
     mirror_list = mirrors.get_list_of_mirrors('target', 'a/b', self.mirrors)
     self.assertFalse(mirror_list)
     self.mirrors['mirror1'] = mirror1
```
joshuagl: Do we need to re-add mirror4 here also?

tests/test_mirrors.py
```diff
@@ -60,30 +60,36 @@ def setUp(self):
                  'metadata_path' : 'metadata',
                  'targets_path' : 'targets',
                  'confined_target_dirs' : ['targets/release/',
-                                            'targets/release/']}}
+                                            'targets/release/']},
+     # Make sure we are testing if confined_target_dirs is ['']
```
joshuagl:
```suggestion
     # Make sure we are testing when confined_target_dirs is ['']
```

I wonder if this is overloading the current test too much, perhaps it would be cleaner to add a new test case just for the `confined_target_dirs = ['']` scenario?

tests/test_mirrors.py
MVrachev: We could in the name of consistency, but we can say we don't need to re-add mirror1 too because the following tests are regarding exception handling and don't actually check the content of `self.mirrors`.

Maybe I can even delete this re-adding because I don't see any usage of it.

tests/test_mirrors.py
MVrachev: This test doesn't contain complicated logic it just contains multiple test cases.
If we are going to add a separate test for this case with the new mirror then it makes sense to split the whole current test case into 4 or 5 tests.

tests/test_mirrors.py
```diff
@@ -60,30 +60,36 @@ def setUp(self):
...
+    self.assertEqual(len(mirror_list), 4)
     for mirror, mirror_info in six.iteritems(self.mirrors):
       url = mirror_info['url_prefix'] + '/metadata/release.txt'
       self.assertTrue(url in mirror_list)
 
     mirror_list = mirrors.get_list_of_mirrors('target', 'a.txt', self.mirrors)
-    self.assertEqual(len(mirror_list), 1)
+    self.assertEqual(len(mirror_list), 2)
     self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a.txt' in \
                     mirror_list)
```
joshuagl, line 83: I think we want/need a similar assertion for mirror4?

tests/test_mirrors.py
```diff
@@ -60,30 +60,36 @@ def setUp(self):
...
 
     mirror_list = mirrors.get_list_of_mirrors('target', 'a/b', self.mirrors)
-    self.assertEqual(len(mirror_list), 1)
+    self.assertEqual(len(mirror_list), 2)
     self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a/b' in \
                     mirror_list)
```
joshuagl, line 90: I think we want a similar assertion for mirror4?

tests/test_mirrors.py
joshuagl: Agreed that the logic isn't too complicated. My concern was that by changing the number of expected entries in the mirror, are we making it harder to ensure the results are as we expect? I see now in the unchanged context lines we're asserting not only the number of matched mirrors but also that they are the mirrors we expect. We should extend that check for mirror4.

joshuagl (changes requested):
Apologies for not catching this on the first review, but I think we need to check that mirror4 is in the matched mirror list for the [''] cases.

jku (commented):
with the changes so far discussed, LGTM

tests/test_mirrors.py
```diff
@@ -80,31 +111,13 @@ def test_get_list_of_mirrors(self):
...
-    self.assertTrue(self.mirrors['mirror1']['url_prefix']+'/targets/a/b' in \
-                    mirror_list)
-
     # No matches
     mirror1 = self.mirrors['mirror1']
     del self.mirrors['mirror1']
+    del self.mirrors['mirror5']
     mirror_list = mirrors.get_list_of_mirrors('target', 'a/b', self.mirrors)
     self.assertFalse(mirror_list)
-    self.mirrors['mirror1'] = mirror1
```
MVrachev, line 99: Deleted this line because the tests below don't actually care about the content of `self.mirrors`.

tests/test_mirrors.py
```diff
@@ -60,16 +59,48 @@ def setUp(self):
      'mirror3': {'url_prefix' : 'http://mirror3.com',
                  'targets_path' : 'targets',
                  'confined_target_dirs' : ['targets/release/v2/']},
+     # confined_target_dirs = [] means that none of the targets on
+     # that mirror is available.
```
MVrachev, line 63: Decided to leave a comment about this, because I made a mistake during testing connected with how we interpreted `confined_target_dirs = []`.

MVrachev:
I had to rebase upon the latest develop branch.
I decided to reorder the test cases in `test_mirrors.py` and that's why the lines changed increased.
Also, I added additional test cases for `mirror5` where we were testing about target confinement.

tests/test_mirrors.py
joshuagl: Thanks for adding the comment. I wonder if we should document this better throughout the code, most notably in updater.py. AFAICT if `confined_target_dirs` is a `None` value the mirror will never be considered/used. Does that match your understanding?

joshuagl (approved):
Thank you for adding the extra tests!

tests/test_mirrors.py
MVrachev: Yes, in these code changes we are checking if `confined_target_dirs` is a falsely value like that:

```
confined_target_dirs = mirror_info.get('confined_target_dirs')
# confined_target_dirs is an optional field
if confined_target_dirs and not in_confined_directory(full_filepath,
          confined_target_dirs):
     ...
```

I don't want us to leave a lot of comments (especially publicly) about what `confined_target_dirs = []` means, because I prefer if we advertise to not use the field if you don't want confinement for your mirrors. 

This unit test is the only place where `confined_target_dirs = []` could be seen and that's why it's worth leaving that comment.

Issue #1165, opened 2020-10-06
import-style prevents vendoring


jku:
I just tried to vendor TUF into pip (in other words embed the tuf sources inside pip sources) and it fails (even with #1160 fixed) because of the import style used in TUF source code itself:

`import tuf.download`

Using this (instead of relative imports) means tuf needs to be installed to python path for the imports to work.

I've looked at the dozen other projects pip vendors and the system seems to cope with two styles:

1. Most projects use relative imports like 
```
from . import treebuilders
from .treebuilders.base import Marker
```
1. The other method is 
```
from toml.decoder import InlineTableDict
```

The second style works because vendoring (the tool) modifies `import NAME` to `from vendored.path import NAME` and `from NAME.module import something` to `from vendored.path.NAME.module import something`. I assume this is mostly done so vendored dependencies of vendored  projects work (so e.g. vendored tuf can "import securesystemslib"). In any case the TUF style cannot be automatically modified as far as I can tell.

I'll try to add reproduction instructions that don't require my pip branch...

jku:
Example you can run or [have a look at](https://gist.github.com/jku/22d5f08083c8b7b2f3e552f7ca6c301e):
`git clone https://gist.github.com/22d5f08083c8b7b2f3e552f7ca6c301e.git`

jku:
The issue is any "submodule import" really, not just internal submodule imports: `import securesystemslib.util` will be an issue as well

joshuagl:
This reads like in order to support vendoring TUF (and securesystemslib) we need to change the way we do imports to either
1. explicit relative imports (`from . import downloads`)
2. importing functions (and other attributes) of a module into the current modules namespace (`from tuf.download import safe_download`)

All things being equal, I prefer absolute imports – I believe they make the code easier to read and follow. Indeed, I recently rejected a securesystemslib PR which added explicit relative imports (secure-systems-lab/securesystemslib#254), but now I'm wondering if that's the direction we should take?

I certainly prefer explicit relative imports to importing functions from a dependent module into our own namespace.

jku:
If the important bit is the complete code namespacing, I suppose that would still be possible if package initialization imported the submodules? As in you could do

```
import tuf
tuf.download.safe_download(...)
```
if `__init__.py` actually imported submodules (or a reasonable subset of them) . I don't know much about python library design so maybe there's a reason that isn't being done?

joshuagl:
[...]

PR #1167, opened 2020-10-07
Potential fix for imports

jku:
Maybe fixes issue #1165 ?

This style of import:
`import tuf.download`
is incompatible with vendoring tool (typically vendoring modifies `from X import Y` into `from vendored.path.X import Y` but this can't be succesfully done with our style).

So I'd like to talk about changing the import style. Unfortunately just switching to the style `from tuf import download` means massive code churn as the local name changes from `tuf.download` to `download`: at least 700 lines of change and significant chance of namespace collision.

This PR is a exploration of another way: using package imports to keep the code mostly as is.

* Removes all cases of `import tuf.something`
* Introduces package level default imports tuf and tuf.client, so a plain "import tuf" or "from tuf import client" is enough to get the basic imports in place with minimal changes to source code
* Leaves repository_tool, repository_lib, developer_tool and unittest_toolbox as they are: not part of a default import set
* leaves log out of the default import set because importing it has side-effects, and fixes the local naming issues resulting from that

This should be 100% backwards compatible: all import styles still work.
The important changes in functionality are:
* importing any tuf module typically ends up importing one or both default sets
* tuf.log is no longer imported uselessly in a lot of places (this should be a good thing)

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -29,7 +29,7 @@ def setUpModule():
 
 # Since setUpModule is called after imports we need to import conditionally.
 if IS_PY_VERSION_SUPPORTED:
-    import tuf.exceptions
+    import tuf
     from tuf.api.metadata import (
         Metadata,
         Snapshot,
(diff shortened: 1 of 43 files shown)


jku:
closing this: it was useful as a test for the workaround I'm going to use in pip , but will never be a real solution we'd want to merge

PR #1156, opened 2020-09-30
WIP: Remove slow retrieval attacks

joshuagl:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #**:

**Description of the changes being introduced by the pull request**:

Slow retrievals have been removed from the specification, update the reference implementation accordingly.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #1169, opened 2020-10-08
Add a retry mechanism on server startup failure and use TestServerProcess for port generation

### LINKED ISSUE #1111: tests (simple_server.py) still unreliable

### LINKED ISSUE #1124: Add a method to create a server subprocesses only on unused local ports

**Current behavior**:

**Expected behavior**:

MVrachev:
Fixes #1124, Fixes #1111 

Adds a retry mechanism on server startup failure. This mechanism is useful for the tests
which are using server subprocesses and if any of the ports they are trying to connect on is already taken.
The other changes in this pr make sure we are using only TestServerProcess for port generation and finally
I removed the `port` argument as a whole from the TestServerProcess constructor.

requirements**:

### DIFF
--- a/tests/proxy_server.py
+++ b/tests/proxy_server.py
@@ -489,11 +489,15 @@ def test(HandlerClass=ProxyRequestHandler, ServerClass=ThreadingHTTPServer, prot
 
 
     HandlerClass.protocol_version = protocol
-    httpd = ServerClass(server_address, HandlerClass)
+    try:
+      httpd = ServerClass(server_address, HandlerClass)
+      sa = httpd.socket.getsockname()
+      print("bind succeeded")
+      print("Serving HTTP Proxy on", sa[0], "port", sa[1], "...")
+      httpd.serve_forever()
+    except:
+      print("bind failed")
 
-    sa = httpd.socket.getsockname()
-    print "Serving HTTP Proxy on", sa[0], "port", sa[1], "..."
-    httpd.serve_forever()
 
 
 if __name__ == '__main__':
--- a/tests/proxy_server.py
+++ b/tests/proxy_server.py
@@ -39,6 +39,8 @@
   Note that this is not thread-safe, in part due to its use of globals.
 """
 
+from utils import ImproperNumberOfArguments
+
 import sys
 import os
 import socket
@@ -462,10 +464,13 @@ def test(HandlerClass=ProxyRequestHandler, ServerClass=ThreadingHTTPServer, prot
     global INTERCEPT
     global TARGET_SERVER_CA_FILEPATH
 
-    if sys.argv[1:]:
-        port = int(sys.argv[1])
-    else:
-        port = 8080
+    # sys.argv[0] contains the name of the executed script
+    # and that's why we require length > 1
+    if len(sys.argv) <= 1:
+      raise ImproperNumberOfArguments("Failure on " + __file__ + " startup!" \
+          + "Port number should be provided and should be the first argument!")
+
+    port = int(sys.argv[1])
     server_address = ('localhost', port)
 
     # MODIFIED: Argument added, conditional below added to control INTERCEPT
--- a/tests/simple_https_server.py
+++ b/tests/simple_https_server.py
@@ -62,11 +62,14 @@ if len(sys.argv) > 2:
     print('simple_https_server: cert file not found: ' + sys.argv[2] +
         '; using default: ' + certfile)
 
-httpd = six.moves.BaseHTTPServer.HTTPServer(('localhost', PORT),
-                            six.moves.SimpleHTTPServer.SimpleHTTPRequestHandler)
+try:
+  httpd = six.moves.BaseHTTPServer.HTTPServer(('localhost', PORT),
+                              six.moves.SimpleHTTPServer.SimpleHTTPRequestHandler)
 
-httpd.socket = ssl.wrap_socket(
-    httpd.socket, keyfile=keyfile, certfile=certfile, server_side=True)
+  httpd.socket = ssl.wrap_socket(
+      httpd.socket, keyfile=keyfile, certfile=certfile, server_side=True)
 
-#print('Starting https server on port: ' + str(PORT))
-httpd.serve_forever()
+  print("bind succeeded")
+  httpd.serve_forever()
+except:
+  print("bind failed")
--- a/tests/simple_https_server.py
+++ b/tests/simple_https_server.py
@@ -37,22 +37,24 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import unicode_literals
 
+from utils import ImproperNumberOfArguments
+
 import sys
 import random
 import ssl
 import os
 import six
 
-PORT = 0
-
 keyfile = os.path.join('ssl_certs', 'ssl_cert.key')
 certfile = os.path.join('ssl_certs', 'ssl_cert.crt')
 
-if len(sys.argv) > 1:
-  PORT = int(sys.argv[1])
+# sys.argv[0] contains the name of the executed script
+# and that's why we require length > 1
+if len(sys.argv) <= 1:
+  raise ImproperNumberOfArguments("Failure on " + __file__ + " startup!" \
+      + "Port number should be provided and should be the first argument!")
 
-else:
-  PORT = random.randint(30000, 45000)
+PORT = int(sys.argv[1])
 
 if len(sys.argv) > 2:
 
--- a/tests/simple_https_server.py
+++ b/tests/simple_https_server.py
@@ -40,7 +40,6 @@ from __future__ import unicode_literals
 from utils import ImproperNumberOfArguments
 
 import sys
-import random
 import ssl
 import os
 import six
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -73,6 +73,10 @@ else:
 # Allow re-use so you can re-run tests as often as you want even if the
 # tests re-use ports. Otherwise TCP TIME-WAIT prevents reuse for ~1 minute
 six.moves.socketserver.TCPServer.allow_reuse_address = True
-httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
 
-httpd.serve_forever()
+try:
+  httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
+  print("bind succeeded")
+  httpd.serve_forever()
+except:
+  print("bind failed")
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -34,18 +34,19 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import sys
-import random
+
+from utils import ImproperNumberOfArguments
 
 import six
 from six.moves.SimpleHTTPServer import SimpleHTTPRequestHandler
 
-PORT = 0
+# sys.argv[0] contains the name of the executed script
+# and that's why we require length > 1
+if len(sys.argv) <= 1:
+  raise ImproperNumberOfArguments("Failure on " + __file__ + " startup!" \
+      + "Port number should be provided and should be the first argument!")
 
-if len(sys.argv) > 1:
-  PORT = int(sys.argv[1])
-
-else:
-  PORT = random.randint(30000, 45000)
+PORT = int(sys.argv[1])
 
 
 class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -48,6 +48,13 @@ if len(sys.argv) <= 1:
 
 PORT = int(sys.argv[1])
 
+# Used for tests in tests/test_utils.py
+if len(sys.argv) > 2:
+  if sys.argv[2] == "stop":
+    sys.exit()
+  elif sys.argv[2] == "endless":
+    while True:
+      pass
 
 class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
   """A SimpleHTTPRequestHandler that does not write incoming requests to
--- a/tests/slow_retrieval_server.py
+++ b/tests/slow_retrieval_server.py
@@ -67,5 +67,10 @@ class Handler(six.moves.BaseHTTPServer.BaseHTTPRequestHandler):
 if __name__ == '__main__':
   port = int(sys.argv[1])
   server_address = ('localhost', port)
-  httpd = six.moves.BaseHTTPServer.HTTPServer(server_address, Handler)
-  httpd.handle_request()
+
+  try:
+    httpd = six.moves.BaseHTTPServer.HTTPServer(server_address, Handler)
+    print("bind succeeded")
+    httpd.serve_forever()
+  except:
+    print("bind failed")
--- a/tests/slow_retrieval_server.py
+++ b/tests/slow_retrieval_server.py
@@ -37,14 +37,6 @@ import random
 import six
 
 
-# Modify the HTTPServer class to pass the 'test_mode' argument to
-# do_GET() function.
-class HTTPServer_Test(six.moves.BaseHTTPServer.HTTPServer):
-  def __init__(self, server_address, Handler, test_mode):
-    six.moves.BaseHTTPServer.HTTPServer.__init__(self, server_address, Handler)
-    self.test_mode = test_mode
-
-
 
 # HTTP request handler.
 class Handler(six.moves.BaseHTTPServer.BaseHTTPRequestHandler):
@@ -62,38 +54,18 @@ class Handler(six.moves.BaseHTTPServer.BaseHTTPRequestHandler):
       self.send_header('Content-length', str(len(data)))
       self.end_headers()
 
-      if self.server.test_mode == 'mode_1':
-        # Before sending any data, the server does nothing for a long time.
-        DELAY = 40
-        time.sleep(DELAY)
-        self.wfile.write(data)
-
-        return
-
-      # 'mode_2'
-      else:
-        DELAY = 1
-        # Throttle the file by sending a character every DELAY seconds.
-        for i in range(len(data)):
-          self.wfile.write(data[i].encode('utf-8'))
-          time.sleep(DELAY)
-
-        return
+      # Before sending any data, the server does nothing for a long time.
+      DELAY = 40
+      time.sleep(DELAY)
+      self.wfile.write((data.encode('utf-8')))
 
     except IOError as e:
       self.send_error(404, 'File Not Found!')
 
 
 
-def run(port, test_mode):
-  server_address = ('localhost', port)
-  httpd = HTTPServer_Test(server_address, Handler, test_mode)
-  httpd.handle_request()
-
-
-
 if __name__ == '__main__':
   port = int(sys.argv[1])
-  test_mode = sys.argv[2]
-  assert test_mode in ('mode_1', 'mode_2')
-  run(port, test_mode)
+  server_address = ('localhost', port)
+  httpd = six.moves.BaseHTTPServer.HTTPServer(server_address, Handler)
+  httpd.handle_request()
--- a/tests/slow_retrieval_server.py
+++ b/tests/slow_retrieval_server.py
@@ -29,6 +29,8 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import unicode_literals
 
+from utils import ImproperNumberOfArguments
+
 import os
 import sys
 import time
@@ -65,6 +67,12 @@ class Handler(six.moves.BaseHTTPServer.BaseHTTPRequestHandler):
 
 
 if __name__ == '__main__':
+  # sys.argv[0] contains the name of the executed script
+  # and that's why we require length > 1
+  if len(sys.argv) <= 1:
+    raise ImproperNumberOfArguments("Failure on " + __file__ + " startup!" \
+        + "Port number should be provided and should be the first argument!")
+
   port = int(sys.argv[1])
   server_address = ('localhost', port)
 
--- a/tests/slow_retrieval_server.py
+++ b/tests/slow_retrieval_server.py
@@ -34,7 +34,6 @@ from utils import ImproperNumberOfArguments
 import os
 import sys
 import time
-import random
 
 import six
 
--- a/tests/test_slow_retrieval_attack.py
+++ b/tests/test_slow_retrieval_attack.py
@@ -169,7 +169,7 @@ class TestSlowRetrieval(unittest_toolbox.Modified_TestCase):
     repository_basepath = self.repository_directory[len(os.getcwd()):]
 
     self.server_process_handler = utils.TestServerProcess(log=logger,
-        server='slow_retrieval_server.py', timeout=0)
+        server='slow_retrieval_server.py')
 
     logger.info('Slow Retrieval Server process started.')
 
--- a/tests/test_slow_retrieval_attack.py
+++ b/tests/test_slow_retrieval_attack.py
@@ -46,7 +46,6 @@ from __future__ import unicode_literals
 
 import os
 import tempfile
-import random
 import time
 import shutil
 import logging
@@ -222,7 +221,6 @@ class TestSlowRetrieval(unittest_toolbox.Modified_TestCase):
 
     # Verify that the TUF client detects replayed metadata and refuses to
     # continue the update process.
-    client_filepath = os.path.join(self.client_directory, 'file1.txt')
     try:
       file1_target = self.repository_updater.get_one_valid_targetinfo('file1.txt')
       self.repository_updater.download_target(file1_target, self.client_directory)
--- a/tests/test_slow_retrieval_attack.py
+++ b/tests/test_slow_retrieval_attack.py
@@ -172,14 +172,6 @@ class TestSlowRetrieval(unittest_toolbox.Modified_TestCase):
 
     logger.info('Slow Retrieval Server process started.')
 
-    # NOTE: Following error is raised if a delay is not long enough:
-    # <urlopen error [Errno 111] Connection refused>
-    # or, on Windows:
-    # Failed to establish a new connection: [Errno 111] Connection refused'
-    # 1s led to occasional failures in automated builds on AppVeyor, so
-    # increasing this to 3s, sadly.
-    time.sleep(3)
-
     url_prefix = 'http://localhost:' \
       + str(self.server_process_handler.port) + repository_basepath
 
--- a/tests/test_slow_retrieval_attack.py
+++ b/tests/test_slow_retrieval_attack.py
@@ -68,61 +68,19 @@ logger = logging.getLogger(__name__)
 repo_tool.disable_console_log_messages()
 
 
-class TestSlowRetrievalAttack(unittest_toolbox.Modified_TestCase):
 
-  @classmethod
-  def setUpClass(cls):
-    # Create a temporary directory to store the repository, metadata, and target
-    # files.  'temporary_directory' must be deleted in TearDownModule() so that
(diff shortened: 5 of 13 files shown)


MVrachev:
I was wondering should we remove `wait_for_server()` function calls and the function itself if we are to merge this pr?
We have a mechanism to check if a server has successfully connected, so what advantages do we get from calling `wait_for_server()` after those checks?
What do you think @jku @joshuagl?

PS: Can somebody restart Travis CI? It failed with a strange error...

tests/proxy_server.py
```diff
@@ -489,11 +489,15 @@ def test(HandlerClass=ProxyRequestHandler, ServerClass=ThreadingHTTPServer, prot
 
 
     HandlerClass.protocol_version = protocol
-    httpd = ServerClass(server_address, HandlerClass)
+    try:
+      httpd = ServerClass(server_address, HandlerClass)
+      sa = httpd.socket.getsockname()
+      print "bind succeded"
```
joshuagl: Note typo in the spelling of succeeded, here and throughout this patch.

```suggestion
      print("bind succeeded")
```

Please call [print as a function](https://www.python.org/dev/peps/pep-3105/), not as a statement. I was surprised to see we have uses of print as a statement.

tests/simple_https_server.py
```diff
@@ -62,11 +62,14 @@
...
+                              six.moves.SimpleHTTPServer.SimpleHTTPRequestHandler)
 
-httpd.socket = ssl.wrap_socket(
-    httpd.socket, keyfile=keyfile, certfile=certfile, server_side=True)
+  httpd.socket = ssl.wrap_socket(
+      httpd.socket, keyfile=keyfile, certfile=certfile, server_side=True)
 
-#print('Starting https server on port: ' + str(PORT))
-httpd.serve_forever()
+  print("bind succeded")
```
joshuagl:
```suggestion
  print("bind succeeded")
```

tests/simple_server.py
```diff
@@ -73,6 +73,10 @@ def log_request(self, code='-', size='-'):
 # Allow re-use so you can re-run tests as often as you want even if the
 # tests re-use ports. Otherwise TCP TIME-WAIT prevents reuse for ~1 minute
 six.moves.socketserver.TCPServer.allow_reuse_address = True
-httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
 
-httpd.serve_forever()
+try:
+  httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
+  print("bind succeded")
```
joshuagl:
```suggestion
  print("bind succeeded")
```

tests/utils.py
```diff
@@ -154,13 +154,33 @@ def __init__(self, log, server='simple_server.py',
...
+      # Some tests (like those in test_slow_retrieval_attack.py) require no
+      # checks if the server has started.
+      if timeout == 0:
+        break
+
+      started = self.has_server_started(timeout)
+      if not started:
+        # If the server has not started for whatever reason
+        self.__logger.info("Could not connect to the " + server + " on port " \
+           + str(self.port) + "! Generating a new port and retry.")
```
joshuagl: Minor nit:
```suggestion
           + str(self.port) + "! Generating a new port and retrying.")
```

tests/utils.py
```diff
@@ -154,13 +154,33 @@ def __init__(self, log, server='simple_server.py',
...
+        self.__server_process.kill()
+        self.__server_process.wait()
+        self.port = random.randint(30000, 45000)
+
+    # Make sure the file is empty so you won't print log messages related
```
joshuagl: Nit:
```suggestion
    # Make sure the file is empty so we don't print log messages related
```

tests/utils.py
```diff
@@ -176,6 +196,33 @@ def __init__(self, log, server='simple_server.py',
...
+    while remaining_timeout > 0:
+      # Seek is needed to move the pointer to the beginning of the file, because
+      # the subprocess could have read and/or write and thus moved the pointer.
+      self.__temp_log_file.seek(0)
+      log_message = self.__temp_log_file.read()
+
+      if len(log_message) > 0:
+        lines = log_message.splitlines()
+
+        if "bind succeded" in lines:
```
```suggestion
        if "bind succeeded" in lines:
```

tests/utils.py
```diff
@@ -154,13 +154,33 @@ def __init__(self, log, server='simple_server.py',
     self.port = port or random.randint(30000, 45000)
```
joshuagl: By the end of this series, are we still using the `port` argument to the `TestServerProcess` constructor anywhere? Can we remove it?

If we are still using it and it can not be removed, should it be an error when the caller requests a specific port number and binding to that specific port number fails?

tests/test_updater.py
```diff
@@ -1856,27 +1857,29 @@ def setUp(self):
...
 
-    url_prefix = 'http://localhost:' + str(self.SERVER_PORT)
-    url_prefix2 = 'http://localhost:' + str(self.SERVER_PORT2)
+    url_prefix = 'http://localhost:' + str(self.server_process_handler.port)
+    url_prefix2 = 'http://localhost:' + str(self.server_process_handler2.port)
+
+    self.map_file_path = os.path.join(self.client_directory, 'map.json')
+    data = securesystemslib.util.load_json_file(self.map_file_path)
+
+    # Add the url_prefixes to the map file.
```
joshuagl: Nice change to remove these hard-coded port numbers! I wonder if it's worth adding a bit more detail to this comment, so that a future reader without the wider context of this PR can easily understand that we are writing generated numbers to a copy of the test file.

tests/test_slow_retrieval_attack.py
```diff
@@ -201,29 +153,53 @@ def setUp(self):
...
+
+    logger.info('Slow Retrieval Server process started.')
+
+    # NOTE: Following error is raised if a delay is not long enough:
+    # <urlopen error [Errno 111] Connection refused>
+    # or, on Windows:
+    # Failed to establish a new connection: [Errno 111] Connection refused'
+    # 1s led to occasional failures in automated builds on AppVeyor, so
+    # increasing this to 3s, sadly.
+    time.sleep(3)
```
joshuagl: Can we not pass this as a timeout value to `TestServerProcess`, rather than having a `sleep()` here? I thought `wait_for_server` was designed for cases such as this?

cc @jku

tests/test_slow_retrieval_attack.py
```diff
@@ -68,68 +68,20 @@
 repo_tool.disable_console_log_messages()
 
 
-class TestSlowRetrievalAttack(unittest_toolbox.Modified_TestCase):
 
-  @classmethod
-  def setUpClass(cls):
+def setUpTuf(obj, repo_name, mode):
```
joshuagl: Two nits:
1. can we use snake_case here?
2. could we have a clearer name for this function? It seems to be setting up a copy of the test repo metadata, starting a server and creating an updater – `set_up_slow_retrieval_tests` is rather wordy but might suffice?

joshuagl (changes requested):
Very nice work here, thanks. I have some comments and suggestions inline.

tests/proxy_server.py
MVrachev: I thought I should use print like that because in the `test_proxy_use.py` file is said: 
```
  THIS module requires Python2.7 (not 2.8.x, not 3.x, just 2.7.x) as the test
  proxy it uses only supports Python2.7.
```
and I thought I should use python 2 syntax.
I will fix it.

tests/utils.py
MVrachev: I remove the port arg in my last commit from this series:
`Remove port arg from __init__ in TestServerProcess`.

tests/test_slow_retrieval_attack.py
MVrachev: `set_up_slow_retrieval_tests` will work as a title.

tests/test_slow_retrieval_attack.py
MVrachev: I am not sure why, but if we use `wait_for_server` with whatever timeout I come up with then the test always fails with:
 
image

MVrachev:
Updated the commits with all of your suggestions @joshuagl except on one of your comments.
There I hope Jussi can give his opinion as you have suggested.

tests/proxy_server.py
joshuagl: Thanks! Fortunately print as a statement was introduced in Python 2.7, so we don't have to carry code that [Guido has long regretted](https://legacy.python.org/doc/essays/ppt/regrets/PythonRegrets.pdf) 😁

jku:
> I was wondering should we remove `wait_for_server()` function calls and the function itself if we are to merge this pr?
> We have a mechanism to check if a server has successfully connected, so what advantages do we get from calling `wait_for_server()` after those checks?

I'm not 100% sure -- _maybe_ successful binding is enough to guarantee that there is a response when the port is tried?  It seems reasonable but I would not bet retirement money on that being true on every platform.

Maybe don't remove it now, but add a calendar notification a few weeks after merging. Then we can check CI logs to see whether the combination helped and can remove `wait_for_server()` as a test to see if your work here is enough alone -- that would be ideal.

tests/simple_server.py
```diff
@@ -73,6 +73,10 @@ def log_request(self, code='-', size='-'):
...
 six.moves.socketserver.TCPServer.allow_reuse_address = True
-httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
 
-httpd.serve_forever()
+try:
+  httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
+  print("bind succeeded")
+  httpd.serve_forever()
+except:
+  print("bind failed")
```
jku, line 90: We don't really know it's the bind failing (and not something else)... but I think this is fine

tests/utils.py
```diff
@@ -141,26 +136,45 @@ class TestServerProcess():
...
+        break
+
+      started = self.has_server_started(timeout)
+      if not started:
+        # If the server has not started for whatever reason
+        self.__logger.info("Could not connect to the " + server + " on port " \
+           + str(self.port) + "! Generating a new port and retrying.")
+        self.__server_process.kill()
+        self.__server_process.wait()
+        self.port = random.randint(30000, 45000)
```
jku: nit: this self.port assignment could be moved to beginning of loop: then you wouldn't need the original initialization outside the loop

tests/utils.py
```diff
@@ -176,6 +190,33 @@ def __init__(self, log, server='simple_server.py',
...
+      if len(log_message) > 0:
+        lines = log_message.splitlines()
+
+        if "bind succeeded" in lines:
+          return True
+        elif "bind failed" in lines:
+          return False
+        # If the server process has exited we consider this as a
+        # failed attempt to connect to the server.
+        elif self.__server_process.poll():
```
jku: this last check should be done outside the  `if len(log_message) > 0` block ?

tests/test_slow_retrieval_attack.py
```diff
@@ -68,68 +68,20 @@
 repo_tool.disable_console_log_messages()
 
 
-class TestSlowRetrievalAttack(unittest_toolbox.Modified_TestCase):
 
-  @classmethod
-  def setUpClass(cls):
+def set_up_slow_retrieval_tests(obj, repo_name, mode):
```
jku: this looks like a new SlowRetrievalBase class Setup() masquerading as normal function: `obj` really must be an instance of Modified_TestCase here, because you use `make_temp_directory()`?

Looks a bit fishy as it is.

tests/test_slow_retrieval_attack.py
```diff
@@ -68,68 +68,20 @@
...
-    # kills it and closes the temp file used for logging.
-    self.server_process_handler.clean()
-
-
-  def setUp(self):
-    # We are inheriting from custom class.
-    unittest_toolbox.Modified_TestCase.setUp(self)
-
-    self.repository_name = 'test_repository1'
+    obj.temporary_directory = tempfile.mkdtemp(dir=os.getcwd())
```
jku: make_temp_directory() is used later to create another temporary directory? I imagine one would be enough

tests/simple_server.py
MVrachev: Yes, we don't care what exactly is the error within the try block.
The result is that the server doesn't work as expected and it's better to retry.

tests/test_slow_retrieval_attack.py
jku: I did not originally remove this sleep since one of the tests actually tests for too slow initial response -- thinking that wait_for_server() would break the test...  but now that I actually read the server code again, I think it may be fine for us to remove the sleep, and give a timeout to TestServerProcess.

If no-one else is sure how it works either, then I guess we'll have to test it to see...

tests/test_slow_retrieval_attack.py
```diff
@@ -234,6 +210,13 @@ def tearDown(self):
     tuf.roledb.clear_roledb(clear_all=True)
     tuf.keydb.clear_keydb(clear_all=True)
```
jku, line 211: these tuf calls should not be needed if this function just wipes everything anyway

jku (changes requested):
otherwise looks good, left comments about slow retrieval mostly...

tests/utils.py
MVrachev: Yes, it seems I did a mistake.

jku:
Commenting to write down results from chat: I think there is probably a `slow_retrieval_server.py` refactoring that would mean the class separation and the new function are not required at all in `test_slow_retrieval_attack.py` -- but I'm definitely not saying it has to be done: spending too much time on these tests is probably not a good investment.

tests/utils.py
```diff
@@ -141,26 +136,45 @@ class TestServerProcess():
...
 
-    # The "-u" option forces stdin, stdout and stderr to be unbuffered.
-    command = ['python', '-u', server, str(self.port)] + extra_cmd_args
-
-    # We are reusing one server subprocess in multiple unit tests, but we are
-    # collecting the logs per test.
-    self.__server_process = subprocess.Popen(command,
-        stdout=self.__temp_log_file, stderr=subprocess.STDOUT, cwd=popen_cwd)
+    started = False
+    while not started:
```
jku: this loop needs to handle timeout somehow: if `has_server_started()` keeps saying no, this loop never exits.

it might make more sense to keep both loops in this function for that reason (the timeout being shared), and instead have the separate function just be the check for logfile contents and poll()

joshuagl:
> Commenting to write down results from chat: I think there is probably a `slow_retrieval_server.py` refactoring that would mean the class separation and the new function are not required at all in `test_slow_retrieval_attack.py` -- but I'm definitely not saying it has to be done: spending too much time on these tests is probably not a good investment.

Agreed. Particularly as slow retrieval is currently not part of the spec and could be removed from the reference implementation (see #1156).

tests/utils.py
MVrachev: Great discovery!

tests/test_slow_retrieval_attack.py
MVrachev: That was a great observation.
I removed this temporary directory creation and found out that this pattern is used in a lot of tests.
Removed all instances of this pattern and will show it in an upcoming commit.

tests/test_slow_retrieval_attack.py
MVrachev: If I remove the sleep I get the result I have screenshotted above.
I am not sure why, but I don't think we can remove the sleep here...

MVrachev:
I addressed all of @jku feedbacks.

I rebased and changed the following thinks:

1. In `tests/utils.py`  I set up a hardcoded amount of possible port generation attempts.

2. Changed the place of the `If the server process has exited` check in `has_server_started` in `tests/utils.py`.

3. Removed the `expected to fail` test from `tests/test_slow_retrieval_attack.py`. As I said in my modified commit message: 
```
Slow retrievals have been removed from the specification and soon 
it will be removed from the tuf reference implementation as a whole.
This means that the chances of making this test useful are close to 0 if not none.
```
4. Because of this comment https://github.com/theupdateframework/tuf/pull/1169#discussion_r505488484 made by @jku I made a research and found that we creating one additional temporary folder per test class in many of our test files.
Fixed that in the last commit.

tests/utils.py
```diff
@@ -163,16 +163,37 @@ def __init__(self, log, server='simple_server.py',
...
-    # The "-u" option forces stdin, stdout and stderr to be unbuffered.
-    command = ['python', '-u', server, str(self.port)] + extra_cmd_args
-
-    # We are reusing one server subprocess in multiple unit tests, but we are
-    # collecting the logs per test.
-    self.__server_process = subprocess.Popen(command,
-        stdout=self.__temp_log_file, stderr=subprocess.STDOUT, cwd=popen_cwd)
+    started = False
+    port_generation_attempts = 5
+    while not started and port_generation_attempts > 0:
```
jku: That probably works but I don't understand why we need another variable: Does it become too complex somehow if you use timeout here as well?

When you have two different "escape" variables you might end up with a situation where e.g. _has_server_started() times out (10seconds) but you keep retrying for 5 times: so TestServerProcess() takes 50 seconds to fail even though it was given 10 seconds as a timeout.

tests/utils.py
```diff
@@ -188,6 +203,34 @@ def __init__(self, log, server='simple_server.py',
 
 
 
+  def has_server_started(self, remaining_timeout=10):
+    """ Checks if the server has started and returns boolean value"""
```
jku:
```suggestion
    """Waits until server has successfully started or 'remaining_timeout' seconds have elapsed. 
       Returns True if server succesfully started"""
```

tests/utils.py
```diff
@@ -153,26 +148,46 @@ class TestServerProcess():
...
+
+      # We are reusing one server subprocess in multiple unit tests, but we are
+      # collecting the logs per test.
+      self.__server_process = subprocess.Popen(command,
+          stdout=self.__temp_log_file, stderr=subprocess.STDOUT, cwd=popen_cwd)
+
+      # Some tests (like those in test_slow_retrieval_attack.py) require no
+      # checks if the server has started.
+      if timeout == 0:
+        break
```
jku: I think we established that this should not be true. Does this not work with slow server (if  it also prints the bind messages)?

tests/utils.py
```diff
@@ -153,26 +148,46 @@ class TestServerProcess():
...
+
+      started = self.has_server_started(timeout)
+      if not started:
+        # If the server has not started for whatever reason
+        self.__logger.info("Could not connect to the " + server + " on port " \
```
jku:
```suggestion
        self.__logger.info("Failed to start " + server + " on port " \
```
We've not tried connecting

tests/utils.py
```diff
@@ -188,6 +203,34 @@ def __init__(self, log, server='simple_server.py',
...
+        lines = log_message.splitlines()
+
+        if "bind succeeded" in lines:
+          return True
+        elif "bind failed" in lines:
+          return False
+
+      # If the server process has exited we consider this as a
+      # failed attempt to connect to the server.
+      if self.__server_process.poll():
```
jku:
```suggestion
      if self.__server_process.poll() is not None:
```
Please review this yourself as well but I think it's more pythonic to avoid even potential issues with Truthy/Falsey return values from poll()

tests/utils.py
```diff
@@ -163,16 +163,37 @@ def __init__(self, log, server='simple_server.py',
...
+        break
+
+      started = self.has_server_started(timeout)
+      if not started:
+        # If the server has not started for whatever reason
+        self.__logger.info("Could not connect to the " + server + " on port " \
+           + str(self.port) + "! Generating a new port and retrying.")
+        self.__server_process.kill()
+        self.__server_process.wait()
+        port_generation_attempts -= 1
```
joshuagl: Haven't seen a decrement assignment operator in Python in a long time!

tests/slow_retrieval_server.py
```diff
@@ -62,38 +54,18 @@ def do_GET(self):
...
-        DELAY = 40
-        time.sleep(DELAY)
-        self.wfile.write(data)
-
-        return
-
-      # 'mode_2'
-      else:
-        DELAY = 1
-        # Throttle the file by sending a character every DELAY seconds.
```
joshuagl, line 76: It would be useful to update the commit message to make it clear that only the throttle/mode_2 slow-retrieval test has been removed and that the delayed/mode_1 slow retrieval test remains

tests/utils.py
```diff
@@ -168,7 +162,7 @@ def __init__(self, log, server='simple_server.py',
     started = False
     port_generation_attempts = 5
     while not started and port_generation_attempts > 0:
-      self.port = port or random.randint(30000, 45000)
```
joshuagl: Nit: the commit message does not tell us _why_ the argument was removed. Presumably because none of the tests are using it?

joshuagl (approved):
Very nice set of changes, thanks Martin!

tests/utils.py
```diff
@@ -153,26 +148,46 @@ class TestServerProcess():
...
+      # checks if the server has started.
+      if timeout == 0:
+        break
+
+      started = self.has_server_started(timeout)
+      if not started:
+        # If the server has not started for whatever reason
+        self.__logger.info("Could not connect to the " + server + " on port " \
+           + str(self.port) + "! Generating a new port and retrying.")
+        self.__server_process.kill()
```
jku: I think this will fail if the process has exited

tests/utils.py
```diff
@@ -188,6 +203,34 @@ def __init__(self, log, server='simple_server.py',
...
+          return True
+        elif "bind failed" in lines:
+          return False
+
+      # If the server process has exited we consider this as a
+      # failed attempt to connect to the server.
+      if self.__server_process.poll():
+          return False
+
+      time.sleep(0.2)
```
jku: have you experimented if this value affects test suite runtime -- I'd expect a shorter sleep would work and would mean faster runtime overall?

jku (commented):
I did not look at the new commit yet -- it sounds good but also looks like a lot of unrelated changes that are not the easiest to review... If you want to keep it in this PR we can do that but it could be a separate one as well

MVrachev:
> I did not look at the new commit yet -- it sounds good but also looks like a lot of unrelated changes that are not the easiest to review... If you want to keep it in this PR we can do that but it could be a separate one as well

Yea, maybe you are right.
This is a big cleanup and it deserves its own pr and it's maybe hard to review.
Additionally, there could be a discussion it.

tests/utils.py
MVrachev: I haven't, but I will decrease it to 0.1 for now.

MVrachev:
I didn't want to rebase this time, but because I had to change the commit messages I had to.

I found a lot of different small or bigger problems while addressing your comments @joshuagl and @jku.
Thanks for the nice reviews!

There are a lot of changes (for which I am sorry that I didn't submit them in the begging) which I will try to summarize:

1. **Removed** the last commit which removed the extra temporary folder. Will submit a new pr for it.

2. Made `utils.py` to use **one variable as a condition** for the two loops.

3. **Moved** the startup of the server in `utils.py` in a separate function.

4. **Added tests** for `utils.py` to make sure that we catch the usual use-cases for our abstraction,  but also the corner cases (e.g. when the server hangs on forever or exits unexpectedly).

5. **Added a helper function** in `utils.py` called `is_process_running` which I then used through `utils.py` and its tests.

6. **Removed the need to use timeout = 0** and make `slow_retrieval.py` to use the standard workflow in `utils.py`.

7. Made sure that we should use the port generation workflow from `utils.py` when we run the four server files (e.g. `simple_server.py`, `slow_retrieval.py` ...)

8. **Added a custom exception** when the port is not provided.

9. **Changed the commit messages** for `Remove test_slow_retrieval expected failure test` and `Remove port arg from __init__ in TestServerProcess`

10. **Removed sleep** from `test_slow_retrieval_attack.py` . We no longer get `TimeoutError` without it.

11. **Updated some of the comments** am introducing them now and not in the begging), but I will summarize them:

tests/test_slow_retrieval_attack.py
MVrachev: This one is fixed in my last update.

tests/utils.py
```diff
@@ -49,6 +49,11 @@ def __str__(self):
       return repr(self.value)
 
 
+class ImproperNumberOfArguments(Exception):
+    """Raised when the number of argumnets is wrong."""
+    pass
```
jku, line 54: Let's not add exceptions if we don't intend to ever handle them (also, server scripts importing the utils.py that starts the server scripts... this does not look right)

tests/utils.py
```diff
@@ -153,44 +152,110 @@ class TestServerProcess():
...
+
+        if self.is_process_running():
+          self.__server_process.kill()
+          self.__server_process.wait()
 
-    # We are reusing one server subprocess in multiple unit tests, but we are
-    # collecting the logs per test.
-    self.__server_process = subprocess.Popen(command,
-        stdout=self.__temp_log_file, stderr=subprocess.STDOUT, cwd=popen_cwd)
+        timeout = timeout - (time.time() - start)
```
jku, line 207: this does not do what it should :)

tests/simple_server.py
```diff
@@ -34,19 +34,27 @@
...
+  raise ImproperNumberOfArguments("Failure on " + __file__ + " startup!" \
+      + "Port number should be provided and should be the first argument!")
 
-else:
-  PORT = random.randint(30000, 45000)
+PORT = int(sys.argv[1])
 
+# Used for tests in tests/test_utils.py
+if len(sys.argv) > 2:
+  if sys.argv[2] == "stop":
```
jku, line 53: first, this script already uses argv[2] for something else. Second: I'm not 100% convinced these new features should be here: If you need this kind of "breaking server", add a new script instead

tests/utils.py
```diff
@@ -215,8 +280,15 @@ def clean(self):
...
           ' terminated.')
       self.__server_process.kill()
       self.__server_process.wait()
+
+
+
+  def is_process_running(self):
+    """Returns a boolean value if the server process is currently running."""
+
+    return True if self.__server_process.returncode is None else False
```
jku, line 294: after testing and checking docs I found out that `returncode` does not actually work if poll() or wait() has not been called: so we should just use poll() here instead.

I may have been the one who said "use returncode", sorry about that

jku (changes requested):
Left a few specific comments, but I have some general ones as well:
* I'm not 100% sold on the tests: the idea is good (TestServerProcess is complex enough that it should ideally be tested) but testing this properly is _not_ trivial and the implementation should not mean actual code (e.g. simple_server.py) gets more complex because of it
* The server startup functionality still smells complex: I think this is the code that will make or break the CI reliability, and it should be both correct and _simple enough for everyone to see that it's correct_. Not going to details here: let's talk it through later

MVrachev:
While addressing the comments made by @jku I read somewhere that we can use a simpler way to generate unused ports as showed here: https://docs.python.org/3/library/socketserver.html#asynchronous-mixins
```
    # Port 0 means to select an arbitrary unused port
    HOST, PORT = "localhost", 0
    server = ThreadedTCPServer((HOST, PORT), ThreadedTCPRequestHandler)
```
The same method applies for `BaseHTTPServer.HTTPServer`, `TCPServer ` and  `ThreadingHTTPServer`  which we are using in our server scripts used for testing.

This means that instead of us generating a port, verifying that it's not busy, and using a retry mechanism if the port is busy we can simply start the server like that 
```
six.moves.socketserver.TCPServer(('localhost', 0), handler)
```
and arbitrary unused port will be chosen for us.

I have implemented that new mechnaism in a different branch.

**I will close this pr** because there is no longer sense for us to generate a new port and validate it as I did and we discussed in this pr **and I will create a new pr where I will push this new branch.**

Issue #1142, opened 2020-09-14
client/updater design: let user download targets?


jku:
This is a future wishlist item for client API (#1135), based on working on pip integration.

Current issues with TUF from pip point of view:
* No way to implement progress indication (a current feature) in pip 
* Difficult to implement parallel downloads in pip (a potential future feature)
* No way to control low level HTTP details (authentication, timeouts, concurrency)

Some of these could be fixed by adding more complexity to TUF... but potentially the correct fix is to separate downloading from the other TUF functionality and let the clients do that themselves.

High level functionality list of the client currently looks like this:
1. generic metadata update (`client.updater.Updater.refresh()`)
1. target metadata update (`client.updater.Updater.get_one_valid_targetinfo()`) 
1. resolving potential download URLs for target (`mirrors.get_list_of_mirrors()`, happens during all metadata and target downloads)
1. local cache check (`client.updater.Updater.updated_targets()`)
1. downloading target (`client.updater.Updater.download_target()`)
1. verifying target (happens inside `client.updater.Updater.download_target()`)

Maybe we can optionally let user handle "_downloading target_" and expose "_verifying target_" step in a more usable way to the user. This means that "_resolving download URLs_" must be better exposed to user. Also "_local cache check_" need re-design: if user downloads files, they also control caching so we can't expect tuf cache directory structure to exist.

jku:
See also #1135 for more details on mirror configuration (which relates to _resolving download URLs_ part)

sechkova:
I had a look on the current implementation and I would vote for implementing minimal changes in the current Updater code trying to achieve "let user download targets" sooner and then transfer this to the client refactor.

I have some questions and comments for @jku but I'd appreciate general comments related to the client API:

> let user handle "downloading target" and expose "verifying target"

I don't see a reason why we cannot expose a `verify_target` function and keep the current `download_target` (which I would rename to a more explicit `download_verified_target` but this will change the current API so maybe not). 
Anyway, the community may have its security reasons about not letting the client handle downloads on its own and I would like to hear them before implementing it. (@joshuagl @lukpueh @trishankatdatadog )

> "resolving download URLs" must be better exposed to user

I can speculate what do you mean @jku but can you be more specific :) Do you mean adding `get_list_of_mirrors`  or similar to the client module or you dislike something from the function internals?

> "local cache check" need re-design

[...]

PR #1171, opened 2020-10-08
PoC: Let user download targets

sechkova:
**Fixes issue #**: #1142 

Give client-side users the option to provide their own custom download function
performing the actual target files download.

The user has to ensure that the function has the following signature and functionality:
```

def download_handler_func(url, required_length)
""" 
  Given the 'url' and 'required_length' of the desired file, open a
  connection to 'url', download it, and return the contents of the file.
  Also ensure the length of the downloaded file matches 'required_length'
  exactly
"""
```
By default `tuf.download.safe_download `is still used.

requirements**:

### DIFF
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1228,7 +1228,7 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     download_filepath = os.path.join(destination_directory, target_filepath)
 
     # Test passing a handler with correct signature but incorrect implementation
-    def test_1(param_1, param2):
+    def test_1(param_1):
       pass
 
     self.assertRaises(tuf.exceptions.NoWorkingMirrorError,
@@ -1239,7 +1239,7 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     self.assertFalse(os.path.exists(download_filepath))
 
     # Test passing a handler with incorrect number of parameters
-    def test_2(param_1, param_2, param_3):
+    def test_2(param_1, param_2):
       pass
 
     with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
@@ -1253,7 +1253,7 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     self.assertFalse(os.path.exists(download_filepath))
 
     # Test passing a handler throwing an unexpected error
-    def test_3(param_1, param2):
+    def test_3(param_1):
       raise IOError
 
     with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
@@ -1272,13 +1272,8 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     # Checks if the file has been successfully downloaded
     self.assertTrue(os.path.exists(download_filepath))
 
-    # Define a simple custom download handler mimicking
-    # the tuf.download call stack:
-    # safe_download()
-    #   _download_file()
-    #     _download_fixed_amount_of_data()
-    #       _check_downloaded_length()
-    def download_handler(url, required_length):
+    # Define a simple custom download handler
+    def download_handler(url):
       url = six.moves.urllib.parse.unquote(url).replace('\\', '/')
       temp_file = tempfile.TemporaryFile()
 
@@ -1287,32 +1282,12 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
         with session.get(url, stream=True,
             timeout=tuf.settings.SOCKET_TIMEOUT) as response:
 
-          # Check response status.
+          # Check response status
           response.raise_for_status()
 
-          # Download fixed amount of data
-          average_download_speed = 0
-          number_of_bytes_received = 0
-
-          start_time = timeit.default_timer()
-
+          # Read the raw socket response
           for chunk in response.iter_content(chunk_size=tuf.settings.CHUNK_SIZE):
-            number_of_bytes_received += len(chunk)
             temp_file.write(chunk)
-            if number_of_bytes_received >= required_length:
-              break
-
-            stop_time = timeit.default_timer()
-            average_download_speed = number_of_bytes_received / (stop_time - start_time)
-            if average_download_speed < tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED:
-              break
-
-        #Check downloaded length
-        if number_of_bytes_received != required_length:
-          if average_download_speed < tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED:
-            raise tuf.exceptions.SlowRetrievalError(average_download_speed)
-
-          raise tuf.exceptions.DownloadLengthMismatchError(required_length, number_of_bytes_received)
 
       except Exception:
         temp_file.close()
@@ -1321,7 +1296,7 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
       else:
         return temp_file
 
-    #Test passing a custom download function
+    # Test passing a custom download function
     self.repository_updater.download_target(targetinfo, destination_directory,
         custom_download_handler=download_handler)
 
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1782,10 +1782,7 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
       self.repository_updater._check_hashes(targets_path, file_hashes)
 
     self.repository_updater._get_file('targets.json', verify_target_file,
-        file_type, file_size, download_safely=True).close()
-
-    self.repository_updater._get_file('targets.json', verify_target_file,
-        file_type, file_size, download_safely=False).close()
+        file_type, file_size).close()
 
   def test_13__targets_of_role(self):
     # Test case where a list of targets is given.  By default, the 'targets'
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1213,6 +1213,73 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
 
 
+  def test_download_target_custom_download_handler(self):
+    # Create temporary directory (destination directory of downloaded targets)
+    # that will be passed as an argument to 'download_target()'.
+    destination_directory = self.make_temp_directory()
+    target_filepaths = \
+      list(self.repository_updater.metadata['current']['targets']['targets'].keys())
+
+    # Get the target info, which is an argument to 'download_target()'.
+    target_filepath = target_filepaths.pop()
+    targetinfo = self.repository_updater.get_one_valid_targetinfo(target_filepath)
+    download_filepath = os.path.join(destination_directory, target_filepath)
+
+    # Test passing a handler with correct signature but incorrect implementation
+    def test_1(param_1, param2):
+      pass
+
+    self.assertRaises(tuf.exceptions.NoWorkingMirrorError,
+        self.repository_updater.download_target, targetinfo, destination_directory,
+        custom_download_handler=test_1)
+
+    # Checks if the file is permanently stored
+    self.assertFalse(os.path.exists(download_filepath))
+
+    # Test passing a handler with incorrect number of parameters
+    def test_2(param_1, param_2, param_3):
+      pass
+
+    with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
+      self.repository_updater.download_target(targetinfo, destination_directory,
+          custom_download_handler=test_2)
+
+    for _, mirror_error in six.iteritems(cm.exception.mirror_errors):
+      self.assertTrue(isinstance(mirror_error, TypeError))
+
+    # Checks if the file is permanently stored
+    self.assertFalse(os.path.exists(download_filepath))
+
+    # Test passing a handler throwing an unexpected error
+    def test_3(param_1, param2):
+      raise IOError
+
+    with self.assertRaises(tuf.exceptions.NoWorkingMirrorError) as cm:
+      self.repository_updater.download_target(targetinfo, destination_directory,
+          custom_download_handler=test_3)
+
+    for _, mirror_error in six.iteritems(cm.exception.mirror_errors):
+      self.assertTrue(isinstance(mirror_error, IOError))
+
+    # Checks if the file has been successfully downloaded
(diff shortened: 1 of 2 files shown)


MVrachev (approved):
Looks good to me!
Good job noticing that `_get_file` doesn't need the `download_safely` argument.

tuf/client/updater.py
```diff
@@ -3237,6 +3253,22 @@ def download_target(self, target, destination_directory,
         to ensure snapshot consistency).
         Default is True.
 
+     custom_download_handler:
+        A user provided function performing the actual target file download.
+        In order to comply with the TUF specification, the function signature
+        should have the form:
+
+        'def download_handler_func(url, required_length)'
```
joshuagl: Can we encode the expected API of the handler somehow and verify that a passed `custom_download_handler` adheres to it? 

What must be included in that API? Certainly the number, order and types of the arguments the handler expects. What about exceptions? The `download_target()` call stack can potentially throw quite a few, for example `_download_file` throws a `tuf.exceptions.DownloadLengthMismatchError`

I'd very much welcome some input here from users of Updater – @trishankatdatadog @jku any thoughts?

joshuagl (changes requested):
Thanks for this PR @sechkova, especially the removal of unused code!

I would definitely like to see some tests for this custom_download_handler. Both testing a custom handler works and also ensuring we error sensibly if an incorrect handler is passed.

I've commented inline regarding my desire to encode the expected API of the custom handler and do some sanity checks that the API is adhered to, let me know what you think about those ideas.

tuf/client/updater.py
sechkova: We can use [inspect.signature](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object) for that purpose. Unfortunately the doc says _New in version 3.3._ but there should be an equivalent way for older versions.

trishankatdatadog: As long as it does not affect current users (can upgrade transparently w/o changing anything or seeing any changes), I have no strong opinion 🙂

sechkova:
I pushed two new commits that only add tests for the new `custom_download_handler` and extend the docs with expected exceptions (_the rebase was rather unintentional ... there is no change in the old commits_) 

About encoding an API, given the dynamic nature of Python, I see two possible approaches:
1) Using the `inspect` module to dynamically check the handler signature against a predefined one or against some expected parameters.
- In practice this will allow us to raise an exception with a custom description when calling `download_target` if the handler has incorrect parameters.
- This option does not require big changes of the code.

2) Defining an interface by and Abstract Base Class.
- In this case the user must implement a custom download method by inheriting from a base class. This will enforce the function signature or an exception will be raised during the class instantiation. At least I don't know a way to enforce rather than recommend by design the types of exceptions raised.
- Requires some changes of the current client code: reformatting the `download` module as a class and defining an ABC as interface. 

I actually prefer none of the above :shrug: 

I don't see a big added value in 1) in comparison with the current behavior. Currently a `TypeError` will be raised when calling the handler with incorrect parameters and will be re-raised as `tuf.exceptions.NoWorkingMirrorError:` with the original `TypeError` added to `mirror_errors`.

I agree that option 2) would be the neater one. But given the current code and the plans to abandon it, is it worth formally defining an interface which the client can brake in practice and we can actually only "recommend it by design"?

If I am missing a clever python way to achieve this, please, I will be glad to learn about it!
In any case, let me know your opinion.

tests/test_updater.py
```diff
@@ -1213,6 +1213,73 @@ def test_6_get_one_valid_targetinfo(self):
...
+
+    # Test: default case.
+    self.repository_updater.download_target(targetinfo, destination_directory)
+
+    # Checks if the file has been successfully downloaded
+    self.assertTrue(os.path.exists(download_filepath))
+
+    # Test: normal case.
+    self.repository_updater.download_target(targetinfo, destination_directory,
+        custom_download_handler=tuf.download.safe_download)
```
jku: is it a lot of work to actually make a working custom handler?

tuf/client/updater.py
```diff
@@ -3241,6 +3247,30 @@ def download_target(self, target, destination_directory,
...
+        Also ensure the length of the downloaded file matches 'required_length'
+        exactly'
+
+        Raise 'securesystemslib.exceptions.DownloadLengthMismatchError', if
+        STRICT_REQUIRED_LENGTH is True and total_downloaded is not equal
+        required_length.
+
+        Raise 'tuf.exceptions.SlowRetrievalError', if the total downloaded was
+        done in less than the acceptable download speed (as set in
+        tuf.settings.py).
```
jku: Last paragraph does not make sense IMO: if the whole file has already been downloaded, what's the point of raising?

Anyway, I think slow retrieval and even the above length mismatch error should be optional: the implementation _may_ raise these things in these conditions. Requiring things that even TUF does not successfully do is a bit much I think :)

on the length check: TUF should check the length after the handler in any case (because it can) whether the handlers raising is documented as optional or not

tuf/client/updater.py
```diff
@@ -3241,6 +3247,30 @@ def download_target(self, target, destination_directory,
...
+        A user provided function performing the actual target file download.
+        In order to comply with the TUF specification, the function signature
+        should have the form:
+
+        'def download_handler_func(url, required_length)'
+
+        The function implementation should provide the following functionality:
+
+        'Given the 'url' and 'required_length' of the desired file, open a
+        connection to 'url', download it, and return the contents of the file.
```
jku: I think "contents of the file" means a file object, that should be explicit. It might even be useful to mention that it really should be a TemporaryFile() -- nothing else really makes sense.

Maybe makes sense to document the handler in the same way TUF functions are documented (even if the format is a bit weird)?

jku (changes requested):
I think this looks very reasonable. I would like to see an actual custom handler in the test -- it might show any undocumented requirements (like the fact that the handler probably really should return a TemporaryFile() -- other things probably don't make sense)

tuf/client/updater.py
jku: I'm pretty sure we do check the length after the handler is finished (in `verify_target_file()`) but please double check :)

tests/test_updater.py
sechkova: I tried! Here it is: ca6f280 
What I did is a very rudimentary implementation of the logic in `tuf.download` (skipping all extra checks, logging, etc).
I am hoping that this will help demonstrating what `tuf` does without the need to follow all the details.
Not sure if this was your expectation?

tuf/client/updater.py
sechkova: I updated the docstring as you suggest (aa3a635 ). The text is copied from the current `tuf.download` module documentation and I don't insist on this content. It is what currently TUF claims doing.

tuf/client/updater.py
sechkova: As said below, this is copied from the current documentation and I am open for discussion on the exact formulation.

Related to the length verification, I checked ... :shrug: 
```

  def _hard_check_file_length(self, file_object, trusted_file_length):
    """
    <Purpose>
      Non-public method that ensures the length of 'file_object' is strictly
      equal to 'trusted_file_length'.  This is a deliberately redundant
      implementation designed to complement
      tuf.download._check_downloaded_length().
```

tests/test_updater.py
jku: Yeah this is good. I don't think it needs to demonstrate what tuf does: it needs to demonstrate how a custom handler could be written -- and it does that. I have opinions about including SlowRetrievalError but since it relates to the larger issue of what we would require from the handler, I'll write that in a separate comment.

tuf/client/updater.py
```diff
@@ -3241,6 +3247,41 @@ def download_target(self, target, destination_directory,
...
+          <Exceptions>
+            DownloadLengthMismatchError, if there was a
+            mismatch of observed vs expected lengths while downloading the file.
+
+            SlowRetrievalError, if the total downloaded was
+            done in less than the acceptable download speed (as set in
+             tuf.settings.py).
+
+          <Returns>
+            A temporay file object that points to the contents of 'url'.
```
trishankatdatadog:
```suggestion
            A temporary file object that points to the contents of 'url'.
```

jku:
I'l let a maintainer make a call but I'll document my opinion on SlowRetrieval with regards to this PR:
* I believe the slow retrieval attack prevention in TUF does not really work (*)
* We should not _require_ a custom handler to implement something we don't know how to do properly ourselves
* We should not provide a bad example of the slow retrieval prevention in the test  -- it's complexity without benefit

--- 

*) I've not tried to implement this but here are the numbers that lead me to believe the code is not useful:
* chunk size is a massive 400000 bytes: this means any checks are only done every 400000 bytes -- even if the expected file size is 10 bytes
* sending 1 byte every 3 seconds would avoid triggering requests socket timeout
* sending 1 byte every 3 seconds 400000 times takes almost 14 days

The client notices that the server is doing a slow retrieval attack *two weeks* after the attack started.

tuf/client/updater.py
```diff
@@ -1706,17 +1712,17 @@ def _get_file(self, filepath, verify_file_function, file_type, file_length,
...
-        # TODO: Instead of the more fragile 'download_safely' switch, unroll
-        # the function into two separate ones: one for "safe" download, and the
-        # other one for "unsafe" download? This should induce safer and more
-        # readable code.
-        if download_safely:
-          file_object = tuf.download.safe_download(file_mirror, file_length)
-
-        else:
-          file_object = tuf.download.unsafe_download(file_mirror, file_length)
+        # Eensure the length of the downloaded file matches 'file_length'
```
```suggestion
        # Ensure the length of the downloaded file matches 'file_length'
```

tests/test_updater.py
```diff
@@ -1213,6 +1215,122 @@ def test_6_get_one_valid_targetinfo(self):
...
+          raise tuf.exceptions.DownloadLengthMismatchError(required_length, number_of_bytes_received)
+
+      except Exception:
+        temp_file.close()
+        raise
+
+      else:
+        return temp_file
+
+    #Test passing a custom download function
```
joshuagl:
```suggestion
    # Test passing a custom download function
```

tuf/client/updater.py
```diff
@@ -3241,6 +3247,41 @@ def download_target(self, target, destination_directory,
...
+            required_length:
+              An integer value representing the length of the file.  This is an exact
+              limit.
+
+          <Side Effects>
+            A temprorary file object is created to store the contents of 'url'.
+
+          <Exceptions>
+            DownloadLengthMismatchError, if there was a
+            mismatch of observed vs expected lengths while downloading the file.
```
joshuagl: `DownloadLengthMismatchError` is thrown after the download handler, or `safe_download()`, call in `_get_file()` by `_hard_check_file_length()` (we also have an unused, except by tests, `_soft_check_file_length()` 🤦 ).

I don't think we need handlers to throw this. Better that we check lengths in the framework and leave the check as is.

joshuagl (commented):
I agree with @jku's assertion that we should not require a custom handler to implement slow retrieval protection. It's not currently part of the spec and we most likely won't re-add it until we can specify how we expect implementations to, well, implement it.

Let's go ahead and remove the expectation for the custom handler to throw `DownloadLengthMismatchError` (as it's handled elsewhere) and `SlowRetrievalError` (as above) and simplify the tests by removing those too.

tuf/client/updater.py
sechkova: Yes, this is the "_deliberately redundant implementation designed to complement tuf.download.\_check\_downloaded\_length()_"
https://github.com/theupdateframework/tuf/pull/1171#discussion_r510943049

Seems like it can nicely complement `custom_download_handler` too without the redundancy part!

sechkova:
Pushed two new commits simplifying the expectation for the custom handler and the tests.
Now an external download handler should perform only the download and the file length is of the returned temporary file is verified by TUF in `_hard_check_file_length()`.

I did not amend the old commits to make the discussion easier to follow but I will probably want to squash/reorder some commits before a final merge.

_Edit: force-pushed the last two commits because of a typo ..._

tuf/client/updater.py
```diff
@@ -3241,6 +3248,29 @@ def download_target(self, target, destination_directory,
...
+
+          <Arguments>
+            url:
+              A URL string that represents the location of the file.
+
+          <Side Effects>
+            A temprorary file object is created to store the contents of 'url'.
+
+          <Returns>
+            A temporary file object that points to the contents of 'url'.
```
joshuagl, line 3270: This description of the return value feels a bit awkward, but I see you've copied it from `tuf.download.safe_download()`

joshuagl (commented):
This looks like a reasonable addition, thanks @sechkova. I'm a bit worried about how complex/jumpy the `download_target()` call stack is already and this addition doesn't help. 

But I think we can clean that up a bit by collapsing some of the functions as a follow-on.

I propose we not merge this until it's clear whether we need it to support pip (see pypa/pip#9041) – does that seem reasonable?

sechkova:
> This looks like a reasonable addition, thanks @sechkova. I'm a bit worried about how complex/jumpy the `download_target()` call stack is already and this addition doesn't help.
> 
> But I think we can clean that up a bit by collapsing some of the functions as a follow-on.
> 
> I propose we not merge this until it's clear whether we need it to support pip (see [pypa/pip#9041](https://github.com/pypa/pip/pull/9041)) – does that seem reasonable?

I'm ok to merge it when there is a clear need of such complication.

jku:
> I propose we not merge this until it's clear whether we need it to support pip (see [pypa/pip#9041](https://github.com/pypa/pip/pull/9041)) – does that seem reasonable?

No news on this yet but I'll note that I'm collecting issues related to this in my pip fork with label 'download': https://github.com/jku/pip/labels/download: it's maybe worth noting that some of these issues will not be completely fixed by the PR here -- e.g. currently pip supports `--proxy` to set a proxy server and since current PR still leaves tuf in control of the metadata downloads, the proxy setting would not be used for the metadata downloads.

Some issues will need individual fixes (possibly just changes in pip or things like issue #1188) in addition to something like this PR...

joshuagl:
I've converted this PR to a draft as it's effectively a proof of concept.

sechkova:
Closing after the discussions in #1213 and opening of #1250.

PR #1173, opened 2020-10-09
Add MissingLocalRepositoryError

### LINKED ISSUE #1063: Updater: Add ability to check for existence of (local) repository
For the purpose of supporting both tuf-enabled servers and legacy servers without additional configuration I'd like to be able to ask tuf.Updater "do you have local metadata for `<repository>`?". 

As a workaround I can check for existence of `<repositories_directory>/<repository>/metadata/current/root.json` but that feels like poking into tuf implementation details.

**Current behavior**:
```
>> updater = tuf.client.updater.Updater('non-existent', repository_mirrors)
  File "<stdin>", line 1, in <module>
  File "/home/jku/src/tuf/tuf/client/updater.py", line 740, in __init__
    raise tuf.exceptions.RepositoryError('Missing'
tuf.exceptions.RepositoryError: Missing './non-existent/metadata/current'.  This path must exist and, at a minimum, contain the Root metadata file.
```

This is very close to what I'd like -- the only issue is that RepositoryError is very generic, it's used for a lot of other things than Missing as well.

**Expected behavior**:
Maybe something like this

```
try:
    updater = tuf.client.updater.Updater('non-existent', repository_mirrors)
except MissingLocalRepositoryError:
    # we don't have root.json on device: use non-tuf-enabled code path


```

jku:
This allows clients to separate
a) missing local repository and
b) error while loading local repository

This is fully backwards-compliant: MissingLocalRepositoryError derives
from RepositoryError and every situation that now results in
MissingLocalRepositoryError used to result in a RepositoryError.

Fixes #1063

---

### DIFF
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -733,6 +733,12 @@ class Updater(object):
     # Set the path for the current set of metadata files.
     repositories_directory = tuf.settings.repositories_directory
     repository_directory = os.path.join(repositories_directory, self.repository_name)
+
+    # raise MissingLocalRepository if the repo does not exist at all.
+    if not os.path.exists(repository_directory):
+      raise tuf.exceptions.MissingLocalRepositoryError('Local repository ' +
+        repr(repository_directory) + ' does not exist.')
+
     current_path = os.path.join(repository_directory, 'metadata', 'current')
 
     # Ensure the current path is valid/exists before saving it.
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -224,6 +224,9 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     # directory.
     tuf.settings.repositories_directory = self.client_directory
 
+    # Test: repository does not exist
+    self.assertRaises(tuf.exceptions.MissingLocalRepositoryError, updater.Updater,
+                      'test_non_existing_repository', self.repository_mirrors)
 
     # Test: empty client repository (i.e., no metadata directory).
     metadata_backup = self.client_metadata + '.backup'
--- a/tuf/exceptions.py
+++ b/tuf/exceptions.py
@@ -115,6 +115,10 @@ class RepositoryError(Error):
   """Indicate an error with a repository's state, such as a missing file."""
 
 
+class MissingLocalRepositoryError(RepositoryError):
+  """Raised when a local repository could not be found."""
+
+
 class InsufficientKeysError(Error):
   """Indicate that metadata role lacks a threshold of pubic or private keys."""
 


tuf/client/updater.py
```diff
@@ -733,6 +733,12 @@ def __init__(self, repository_name, repository_mirrors):
     # Set the path for the current set of metadata files.
     repositories_directory = tuf.settings.repositories_directory
     repository_directory = os.path.join(repositories_directory, self.repository_name)
+
+    # raise MissingLocalRepository if the repo does not exist at all.
+    if not os.path.exists(repository_directory):
+      raise tuf.exceptions.MissingLocalRepositoryError('Local repository ' +
+        repr(repository_directory) + 'does not exist.')
```
MVrachev: NIT: Maybe add space before `does not exists` to make it more readable, like that:
`repr(repository_directory) + ' does not exist.')` 
Because if you printed like that it will be concatenated without spaces between `repository_directory` and `does not exist`.

MVrachev (approved):
LGTM!

jku:
force-pushed the whitespace fix martin suggested, thanks.

trishankatdatadog (approved):
Looks reasonable to me, thank you. The only comment I would add that, while maintaining the current version is important where necessary, we should focus our efforts on the new version as much as possible going forward.

jku:
> The only comment I would add that, while maintaining the current version is important where necessary, we should focus our efforts on the new version as much as possible going forward.

While I recognise and accept the point I'll mention that my change requests have a goal: to enable TUF support in pip. Not doing (most of) the changes more or less means not implementing TUF support in pip and instead waiting for the possible rewrite to first appear at some point in the future. Some of the changes obviously could be worked around on the pip side but that's not appealing either...

Pushing the pip integration into the future is a completely valid choice for the project (and might happen anyway if there are large enough features missing), but so far I've felt the process has been a net positive: some churn sure, but also a lot of feedback, design notes and feature requests. Essentially the point is this: _If we don't polish the code we have, how do we actually know which features we really need in the new code?_

This might be a good subject for the community meeting?

trishankatdatadog:
> This might be a good subject for the community meeting?

💯 and thanks for all your team's hard work, we would be worse off without it!

joshuagl (approved):
LGTM as a minimal, non-breaking change.

joshuagl:
> This might be a good subject for the community meeting?

I fully support discussing this in a community meeting.

For what it's worth, my current position is that moving pip integration forward is valuable. Jussi's work on a proof of concept pip integration has helped identify issues in the warehouse integration, identified areas for improvement in `tuf.client.updater`, and hopefully will allow us to collect feedback from the pip community on changes they want from tuf before they'd be willing to accept the tuf integration into pip (i.e. how many of the identified gaps are actually blockers for pip integration and how many are nice to have?).

I also want to avoid a situation where PyPI is producing tuf metadata but pip isn't using it yet because we're waiting on the refactor, having an integration with the current `tuf.client.updater` implementation seems like a safe way to avoid that.

PR #1176, opened 2020-10-13
Remove iso8601 dependency

### LINKED ISSUE #1065: Remove iso8601 dependency
I've been going through the tuf dependency chain with an eye on integrating tuf with pip: The issue with pip is that it's a package manager so needs to vendor everything it needs -- so being conservative with dependencies is a good idea. The good news is that tuf does not have many direct or indirect dependencies that would be a problem (good work!).

The one that possibly sticks out is "iso8601". The module is currently used in two places:
* repository_tool: Metadata.expiration property getter uses it to return a datetime object
* updater.py: Updater::_ensure_not_expired()

I'm mostly interested in that last one. it's used to compare the expiration stamp to current time and to format the error message:
```
    expires_datetime = iso8601.parse_date(expires)
    expires_timestamp = tuf.formats.datetime_to_unix_timestamp(expires_datetime)

    if expires_timestamp < current_time:
      message = 'Metadata '+repr(metadata_rolename)+' expired on ' + \
        expires_datetime.ctime() + ' (UTC).'
      logger.error(message)

      raise tuf.exceptions.ExpiredMetadataError(message)
```
I'm not familiar with date handling in python so my question is: Is this dependency valid or could this code be replaced with something that did not depend on iso8601?

jku:
Our 'expires' strings are constrained by the ISO8601_DATETIME_SCHEMA
(meaning regex '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z'). This can be
parsed with just a datetime.strptime(): iso8601 module is not needed.

* Add formats.expiry_string_to_datetime() helper function
* Modify the 3 locations that used iso8601 and the api/metadata.py usage
  of datetime.strptime()
* Remove related unnecessary logger setup
* Add the missing exception documentation to relevant functions (in many
  cases the exception is rather unlikely as the schema has been verified
  many times before this though...)

Fixes #1065

---

The reason for doing this is again vendoring into pip: I'd like to not have to vendor iso8601 as well.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -294,9 +294,8 @@ class Signed:
         # Convert 'expires' TUF metadata string to a datetime object, which is
         # what the constructor expects and what we store. The inverse operation
         # is implemented in 'to_dict'.
-        signed_dict['expires'] = datetime.strptime(
-                signed_dict['expires'],
-                "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=None)
+        signed_dict['expires'] = tuf.formats.expiry_string_to_datetime(
+                signed_dict['expires'])
         # NOTE: We write the converted 'expires' back into 'signed_dict' above
         # so that we can pass it to the constructor as  '**signed_dict' below,
         # along with other fields that belong to Signed subclasses.
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -6,7 +6,6 @@ cryptography==3.1.1         # via securesystemslib
 enum34==1.1.6             ; python_version < '3' # via cryptography
 idna==2.10                 # via requests
 ipaddress==1.0.23         ; python_version < '3' # via cryptography
-iso8601==0.1.13
 pycparser==2.20           # via cffi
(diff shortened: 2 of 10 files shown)


tuf/api/metadata.py
```diff
@@ -294,9 +294,8 @@ def from_dict(cls, signed_dict: JsonDict) -> 'Signed':
         # Convert 'expires' TUF metadata string to a datetime object, which is
         # what the constructor expects and what we store. The inverse operation
         # is implemented in 'to_dict'.
-        signed_dict['expires'] = datetime.strptime(
-                signed_dict['expires'],
-                "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=None)
```
jku, line 299: note that I'm not doing the replace() in my version: I could not think of a case where it would be useful.

lukpueh: Makes perfect sense. I guess I thought that `strptime` might adopt the system timezone. Thanks for removing unnecessary code!

lukpueh (approved):
Impeccable work! Thanks, @jku!

PR #1177, opened 2020-10-15
tuf/api: Expose tuf.api as a package (take 2)

joshuagl:
Taking over from #1157 

* add tuf/api/__init__.py, turning tuf/api into a full package for distribution purposes
* add a pylintrc for tuf/api and update tox to use it (and ignore tuf/api with the default pylintrc)
* minor stylistic changes to tuf/api/metadata.py to resolve some pylint warnings

Note: ac8d273 disables a single pylintrc check in a code comment. I was loath to do this, but I think it's better to explicitly disable that site than disable the message for all code before we've decided on coding style.

requirements**:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,8 @@ skipsdist = true
 changedir = tests
 
 commands =
-    pylint {toxinidir}/tuf
+    pylint {toxinidir}/tuf --ignore={toxinidir}/tuf/api
+    pylint {toxinidir}/tuf/api --rcfile={toxinidir}/tuf/api/pylintrc
     bandit -r {toxinidir}/tuf
     coverage run aggregate_tests.py
     coverage report -m --fail-under 97
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27, py35, py36, py37, py38
+envlist = lint,py{27,35,36,37,38}
 skipsdist = true
 
 [testenv]
@@ -14,9 +14,6 @@ skipsdist = true
 changedir = tests
 
 commands =
-    pylint {toxinidir}/tuf --ignore={toxinidir}/tuf/api
-    pylint {toxinidir}/tuf/api --rcfile={toxinidir}/tuf/api/pylintrc
-    bandit -r {toxinidir}/tuf
     coverage run aggregate_tests.py
     coverage report -m --fail-under 97
 
@@ -40,3 +37,9 @@ deps =
 commands =
     coverage run aggregate_tests.py
     coverage report -m
+
+[testenv:lint]
+commands =
+    pylint {toxinidir}/tuf --ignore={toxinidir}/tuf/api
+    pylint {toxinidir}/tuf/api --rcfile={toxinidir}/tuf/api/pylintrc
+    bandit -r {toxinidir}/tuf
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -14,7 +14,8 @@ skipsdist = true
 changedir = tests
 
 commands =
-    pylint {toxinidir}/tuf
+    pylint {toxinidir}/tuf --ignore={toxinidir}/tuf/api
+    pylint {toxinidir}/tuf/api --rcfile={toxinidir}/tuf/api/pylintrc
```
lukpueh: What about https://github.com/theupdateframework/tuf/pull/1157#issuecomment-702025224? Or do you purposefully want to be explicit here?

lukpueh: Also [looks](https://travis-ci.org/github/theupdateframework/tuf/builds/736003163) like we can only run pylint on supported versions for `tuf.api`.

tuf/api/pylintrc
```diff
@@ -0,0 +1,9 @@
+[BASIC]
+good-names=e, fn, fp
```
lukpueh: Minor nit: Right now we don't use those names. What do you think about removing the line and only add when needed, or when fleshing out the config in order to align with the style guide TBD (https://github.com/secure-systems-lab/code-style-guidelines/issues/20)?

tox.ini
lukpueh: I wonder if the linter really needs to run for every supported version. In [in-toto we have a separate lint tox env](https://github.com/in-toto/in-toto/blob/c775e3b260a50ec5fb673a9b725e2d2eacedfb6b/tox.ini#L22-L25) that uses the latest supported Python version.

lukpueh (changes requested):
Thanks a lot, @joshuagl! Looks mostly good to me.

tuf/api/pylintrc
joshuagl: Completely agree, I'll remove this line.

tox.ini
joshuagl: > What about [#1157 (comment)](https://github.com/theupdateframework/tuf/pull/1157#issuecomment-702025224)? Or do you purposefully want to be explicit here?

It didn't seem to work when I tried it. May be because cwd isn't a Python module when we invoke pylint? Being explicit seemed easier than relying on pylint magic I couldn't fully decipher from the documentation.

> I wonder if the linter really needs to run for every supported version. In in-toto we have a separate lint tox env that uses the latest supported Python version.

This feels like a sensible option, I'll add a separate lint env.

tox.ini
```diff
@@ -39,3 +38,8 @@ deps =
 commands =
     coverage run aggregate_tests.py
     coverage report -m
+
+[testenv:lint]
```
lukpueh, line 41: We also need to [adopt this change in .travis.yml](https://github.com/in-toto/in-toto/blob/c775e3b260a50ec5fb673a9b725e2d2eacedfb6b/.travis.yml#L20-L21). And should we maybe move bandit to the lint env too?

joshuagl: Ugh, I always forget to touch `.travis.yml` – thanks!

joshuagl:
I've force pushed an update to squash fixups, which prevents the DCO from complaining. 

You can see the pre rebase history here https://github.com/theupdateframework/tuf/compare/develop...joshuagl:ww/tuf-api-package-hist where I've added 81e0ef6 to add the lint env to travis, and move bandit into the lint env. I also changed 427759b while I was in the .travis.yml to use newer Python for with-sslib-master

woodruffw:
Thanks @joshuagl!

lukpueh (approved):
🎉

PR #1180, opened 2020-10-20
Make colorama dependency optional

joshuagl:
Fixes N/A

* Switch repo script from using colorama directly to using the `TERM_` values defined in `securesystemslib.interface`. These values are set automatically depending on whether colorama is importable, and therefore make the colorama dependency an optional/soft dependency
* Remove colorama from the requirements

requirements**:

### DIFF
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -1,7 +1,6 @@
 certifi==2020.6.20       # via requests
 cffi==1.14.3              # via cryptography, pynacl
 chardet==3.0.4            # via requests
-colorama==0.4.3           # via securesystemslib
 cryptography==3.1.1         # via securesystemslib
 enum34==1.1.6             ; python_version < '3' # via cryptography
 idna==2.10                 # via requests
@@ -10,7 +9,7 @@ pycparser==2.20           # via cffi
 pynacl==1.4.0             # via securesystemslib
 python-dateutil==2.8.1    # via securesystemslib
 requests==2.24.0
-securesystemslib[colors,crypto,pynacl]==0.16.0
+securesystemslib[crypto,pynacl]==0.16.0
 six==1.15.0
 subprocess32==3.5.4       ; python_version < '3' # via securesystemslib
 urllib3==1.25.10           # via requests
--- a/requirements.txt
+++ b/requirements.txt
@@ -41,6 +41,6 @@
 # 3. Use this command to remove per-version files
 #    `rm requirements-?.?.txt`
 #
-securesystemslib[colors, crypto, pynacl]
+securesystemslib[crypto, pynacl]
 requests
 six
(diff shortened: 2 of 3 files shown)


trishankatdatadog (approved):

jku (approved):

lukpueh:
Sorry for the late comment. But I don't think we actually want to remove `coloroma` from the requirements files. As (maybe not clearly enough) [stated in `requirements.txt`s doc header](https://github.com/theupdateframework/tuf/blob/c67fbb66d2d15cafba0170185e8bd1eccb597509/requirements.txt#L1-L7), the idea is to use `requirements.txt` to generate a `requirements-pinned.txt` with all required and optional, immediate and transitive dependencies, to be auto-bumped and auto-tested via `Dependabot`. 

Now, if we remove `colorama` from these files Dependabot won't notify us about breaking updates.

We use `setup.py` to encode the minimal set of requirements, which already didn't list `coloroma` prior to this PR. Also note that the `crypto` and `pynacl` extra markers, which indicate optional dependencies, are not in `setup.py` (but still in the requirements files).

What do you think... @joshuagl, @jku? I'm happy to once more revisit the requirements handling (see #978 and #982 for prior work).

jku:
so I think you're making a lot of sense (I had missed the idea of generating requirements-pinned.txt completely). 

The only thing I'm missing is _why_ do we want colorama as a dependency at all -- as far as I can tell where now using securesystemslib features only, there's essentially no colorama anywhere?

EDIT: or is the idea to document a version for every possible (even transient optional) dependency?

lukpueh:
> so I think you're making a lot of sense (I had missed the idea of generating requirements-pinned.txt completely).

I admit that the documentation is not ideal.

lukpueh:
> EDIT: or is the idea to document a version for every possible (even transient optional) dependency?

I think `pip-tools` goes one level deep into transitive dependencies. IMO we should at least test securesystemslib, which is basically a part of TUF in a separate package, with all its optional dependencies, because they were added for TUF.

lukpueh:
If let's say `six` has an optional dependency that we are not aware of and don't care for, we don't need to test it.

lukpueh:
That said, I think the easiest thing to do would be to just nuke `colorama` from securesystemslib. IIRC @trishankatdatadog indicated some regret that it was added in the first place.

jku:
> IMO we should at least test securesystemslib, which is basically a part of TUF in a separate package, with all its optional dependencies, because they were added for TUF.

Fair point, I guess this does make sense. Is this something you'd like fixed before upcoming release?

On the point of nuking colorama completely -- that probably makes sense? there should be a fairly high bar to adding dependencies to projects like this and coloramas usefulness does not seem to reach that bar...

lukpueh:
> Is this something you'd like fixed before upcoming release?

I think there is no fixing needed. Let's just leave this PR merged as is (I did add a note about it to the changelog) and remove colorama in securesystemslib rather soon, hopefully before they push an update that breaks tuf without us noticing, which I think is fairly unlikely.

PR #1187, opened 2020-10-22
Simplify aggregate_tests.py

MVrachev:

There is a simpler way to skip modules or particular tests
built-in into the unittest module.
That's why it doesn't make sense for us to manually filter
modules based on the python version we are running.

### DIFF
--- a/tests/aggregate_tests.py
+++ b/tests/aggregate_tests.py
@@ -36,61 +36,9 @@ from __future__ import unicode_literals
 
 import sys
 import unittest
-import glob
-import random
-
-# Generate a list of pathnames that match a pattern (i.e., that begin with
-# 'test_' and end with '.py'.  A shell-style wildcard is used with glob() to
-# match desired filenames.  All the tests matching the pattern will be loaded
-# and run in a test suite.
-available_tests = glob.glob('test_*.py')
-
-# A dictionary of test modules that should only run in certain python versions.
-# Carefully consider the impact of only testing these in a given version.
-# test_proxy_use.py: uses a proxy that only runs in Python2.7. TUF's
-# compatibility with proxies is not likely to vary based on the Python version
-# in use, so this is OK for now. See comments in that module.
-# The semantics here are: only add to this list the particular tests that are
-# to be run in a single major version or a single minor version. An entry must
-# include major version, and may include minor version.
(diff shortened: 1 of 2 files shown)


lukpueh:
Cool stuff! I'd even go so far and and remove the `glob`+ for loop. `unittest` has built-in test loader discovery function for that.

lukpueh:
Btw. here's said test loader in action: https://github.com/in-toto/in-toto/blob/develop/tests/runtests.py

@MVrachev, do you have capacities to finalize this PR?

MVrachev:
Yes, sorry. I was busy with my other pr `Delegate port generation to the OS`.
I used a discover `unittest.TestLoader().discover()` successfully, I just wanted to find a way how to randomize the unit tests execution order.

I didn't want to lose functionality with this cleanup.

lukpueh:
TBH I'm not convinced about randomizing unit test modules invocation order. I have the feeling that, if anything, it might only randomly reveal errors in the test code and not in the tested code, and its ability to do so is very limited, because test cases and functions in a module are still executed in a non-random order. And all that at the cost of determinism. 🤷 

What do others think, @joshuagl, @jku, @sechkova?

jku:
mixed feelings: finding issues in tests earlier rather than later is usually better (and randomization can help with that) but ... flaky tests are also the worst. :shrug: indeed

MVrachev:
Okay, for now, I will just upload the version with `discover` until we wait for the opinion of Joshua and Tedi.
I don't have strong feelings about the order of the tests.

sechkova:
I don't have a strong opinion as long as we don't end up randomly executing test cases from different modules which is discouraged by the [docs](https://docs.python.org/3.8/library/unittest.html#class-and-module-fixtures).

joshuagl:
> TBH I'm not convinced about randomizing unit test modules invocation order. I have the feeling that, if anything, it might only randomly reveal errors in the test code and not in the tested code, and its ability to do so is very limited, because test cases and functions in a module are still executed in a non-random order. And all that at the cost of determinism. 🤷

Agreed, I don't think the way we randomise test invocation order adds any value.

lukpueh:
Let's drop randomization then.

joshuagl (approved):
Nice simplification here. I'm happy with _not_ having our test modules be run in a random order.

PR #919, opened 2019-09-18
Streamline TUFs crypto interface 

lukpueh:
**Fixes issue #**:
Closes #656
Supersedes #804
Paves the way for #840 

**Description of the changes being introduced by the pull request**:

- Remove `securesystemslib`-wrappers  in `repository_lib` that don't add any new functionality. Also remove corresponding tests that already exist in `securesystemslib`.
This means the functions are no longer available via `repository_lib`, which should be fine, because `repository_lib` does not seem to be a public interface.

- Replace `securesystemslib`-  and `repository_lib`-wrappers, in `repository_tool` and `developer_tool` with direct imports.
This means, the functions are still available via the `repository_tool` (or `developer_tool`), which they should be (see e.g. TUTORIAL.md), but internally the modules get slimmer. 

See commit messages for more details.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #1191, opened 2020-10-28
Adopt sslib keygen interface encryption changes

lukpueh:
**Blocks on https://github.com/secure-systems-lab/securesystemslib/pull/288 + release**

Fixes #<ISSUE NUMBER> -

secure-systems-lab/securesystemslib#288 changes the key generation interface functions to no longer auto-prompt for an encryption password if no password is passed, in order to not suprise the caller with a blocking prompt.

~~The downside of this change is that the keys are stored in plain text per default, which may be mitigated by recommending encryption in the docs.~~

This PR updates related TUF documentation.

NOTE: The securesystemslib private key import functions do not auto-prompt for decryption passwords either, however TUF only exposes custom wrappers (see repository_lib) that do auto-prompt.

See commit message for details.

requirements**:

### DIFF
--- a/tuf/repository_tool.py
+++ b/tuf/repository_tool.py
@@ -74,8 +74,14 @@ from tuf.repository_lib import (
 
 from securesystemslib.interface import (
     generate_and_write_rsa_keypair,
+    generate_and_write_rsa_keypair_with_prompt,
+    generate_and_write_unencrypted_rsa_keypair,
     generate_and_write_ecdsa_keypair,
+    generate_and_write_ecdsa_keypair_with_prompt,
+    generate_and_write_unencrypted_ecdsa_keypair,
     generate_and_write_ed25519_keypair,
+    generate_and_write_ed25519_keypair_with_prompt,
+    generate_and_write_unencrypted_ed25519_keypair,
     import_rsa_publickey_from_file,
     import_ecdsa_publickey_from_file,
     import_ed25519_publickey_from_file,
--- a/docs/TUTORIAL.md
+++ b/docs/TUTORIAL.md
@@ -96,14 +96,15 @@ text without prepended symbols is the output of a command.
 # following function creates an RSA key pair, where the private key is saved to
 # "root_key" and the public key to "root_key.pub" (both saved to the current
 # working directory).
->>> generate_and_write_rsa_keypair("root_key", bits=2048, password="password")
+>>> generate_and_write_rsa_keypair(password="password", filepath="root_key", bits=2048)
 
 # If the key length is unspecified, it defaults to 3072 bits. A length of less
-# than 2048 bits raises an exception. A password may be supplied as an
-# argument, otherwise a user prompt is presented.  If an empty password
-# is entered, the private key is saved unencrypted.
->>> generate_and_write_rsa_keypair("root_key2")
-Enter a password for the RSA key (/path/to/root_key2):
+# than 2048 bits raises an exception. A similar function is available to supply
+# a password on the prompt. If an empty password is entered, the private key
+# is saved unencrypted.
+>>> generate_and_write_rsa_keypair_with_prompt(filepath="root_key2")
+enter password to encrypt private key file '/path/to/root_key2'
+(leave empty if key should not be encrypted):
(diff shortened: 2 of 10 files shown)


JustinCappos:
Would it be better to have an option to decide how to do this?  If you want the caller to not encrypt the file on disk, we should make them pass in some scary flag, I think.

trishankatdatadog:
> Would it be better to have an option to decide how to do this? If you want the caller to not encrypt the file on disk, we should make them pass in some scary flag, I think.

I agree. Storing keys unencrypted by default is a sort of a backwards-incompatible change, and not a great one.

lukpueh:
@JustinCappos, @trishankatdatadog, I agree with you but there also seems to be a consensus that a default prompt in an interface function is a bad idea (see https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r511420534 for the related discussion).

trishankatdatadog:
> @JustinCappos, @trishankatdatadog, I agree with you but there also seems to be a consensus that a default prompt in an interface function is a bad idea (see [secure-systems-lab/securesystemslib#288 (comment)](https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r511420534) for the related discussion).

Makes sense. If that is the case, then perhaps we need to make a backwards-incompatible change, and split the function into two: one for generating unencrypted keys, and one for encrypted keys (no prompt, because the passphrase must be explicitly specified by the caller). This should follow the principle of least surprise to users. What do you think?

JustinCappos:
> > @JustinCappos, @trishankatdatadog, I agree with you but there also seems to be a consensus that a default prompt in an interface function is a bad idea (see [secure-systems-lab/securesystemslib#288 (comment)](https://github.com/secure-systems-lab/securesystemslib/pull/288#discussion_r511420534) for the related discussion).
> 
> Makes sense. If that is the case, then perhaps we need to make a backwards-incompatible change, and split the function into two: one for generating unencrypted keys, and one for encrypted keys (no prompt, because the passphrase must be explicitly specified by the caller). This should follow the principle of least surprise to users. What do you think?

I think Trishank's suggestion is a good way to do what I was trying to suggest. 

I more meant that we should force the behavior to specify some option (which sort of means there is no default).  I wasn't really meaning to keep the current behavior by default.

lukpueh:
@trishankatdatadog and @JustinCappos, I posted a recap of the discussion at https://github.com/secure-systems-lab/securesystemslib/pull/288#issuecomment-718831595. Let's discuss there what makes sense most.

lukpueh:
The issue of non-default encryption has been resolved in #288 and adopted here. May I kindly ask for a re-review?

lukpueh:
Note that tests fail, because the changes in sslib have not yet been released. The `with-sslib-master` should pass though...

trishankatdatadog (approved):
Looks like a straightforward application to me

tuf/repository_tool.py
```diff
@@ -74,8 +74,14 @@
 
 from securesystemslib.interface import (
     generate_and_write_rsa_keypair,
+    generate_and_write_rsa_keypair_with_prompt,
```
trishankatdatadog, line 77: Do we actually use these new functions here, or are they just being imported w/o being used?

lukpueh: The latter is correct. They are just added to the TUF namespace (see comments above), because a lot of docs suggest to do `from repository_tool import *` to access them.

I started this in https://github.com/theupdateframework/tuf/pull/919 (see 7306446118b89e3e395340e748a2717b03e0666e for details), in order to remove many oneliner function wrappers that served the same purpose but added cruft to the code base and duplicated the maintenance burden above all in regards to docstrings.

A better but also more invasive fix would be to remove these functions from the  tuf namespace altogether, and update the docs to directly use securesystemslib akin to what we now do in in-toto (see https://github.com/in-toto/in-toto/pull/402, https://github.com/in-toto/in-toto/pull/408 and https://in-toto-lukpueh.readthedocs.io/en/latest/api.html#key-utilities).

lukpueh:
Closes #1204

PR #1192, opened 2020-10-28
Delegate port generation for the tests to the OS

### LINKED ISSUE #1111: tests (simple_server.py) still unreliable

### LINKED ISSUE #1124: Add a method to create a server subprocesses only on unused local ports

MVrachev:
Fixes https://github.com/theupdateframework/tuf/issues/1124, Fixes https://github.com/theupdateframework/tuf/issues/1111

This pr aims to make the creation of new server subprocesses more stable.
It fixes issues related to the usage of ports already in use like 
`[Errno 98] Address already in use` or `ConnectionRefusedError: [Errno 111]`

By using `0` as a port argument when we start the servers we ask the OS to give us
an arbitrary unused port. 
This method is much better than if we generate a port, because
instead of us generating a port, verifying that it's not busy, and use a retry mechanism (if the port is busy)
we can simply rely on the OS to do that job for us.

This means that the server scripts receive their ports when they are created and then send it back to the 
father process which had created them.

Other changes/cleanups in this pr include 
- remove test_slow_retrieval expected failure test
-  make sure there is no place where we are generating server ports outside the server scripts.
- remove port argument from TestServerProcess in `tests/utils.py`
- remove unused imports
- remove try/except at server test files 

This pr is related to the closed https://github.com/theupdateframework/tuf/pull/1169 pr but because 
I decided to use a whole different approach I thought it made more sense to close that one and create this one.

PS: Great thanks to @jku with whom we had discussions about this pr on regular occasions!

requirements**:

### DIFF
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -39,14 +39,6 @@ import random
 import six
 from six.moves.SimpleHTTPServer import SimpleHTTPRequestHandler
 
-PORT = 0
-
-if len(sys.argv) > 1:
-  PORT = int(sys.argv[1])
-
-else:
-  PORT = random.randint(30000, 45000)
-
 
 class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
   """A SimpleHTTPRequestHandler that does not write incoming requests to
@@ -73,6 +65,12 @@ else:
 # Allow re-use so you can re-run tests as often as you want even if the
 # tests re-use ports. Otherwise TCP TIME-WAIT prevents reuse for ~1 minute
 six.moves.socketserver.TCPServer.allow_reuse_address = True
-httpd = six.moves.socketserver.TCPServer(('', PORT), handler)
 
-httpd.serve_forever()
+try:
+  httpd = six.moves.socketserver.TCPServer(('localhost', 0), handler)
+  port_message = 'bind succeeded, server port is: ' \
+      + str(httpd.server_address[1])
+  print(port_message)
+  httpd.serve_forever()
+except:
+  print("bind failed")
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -66,11 +66,8 @@ else:
 # tests re-use ports. Otherwise TCP TIME-WAIT prevents reuse for ~1 minute
 six.moves.socketserver.TCPServer.allow_reuse_address = True
 
-try:
-  httpd = six.moves.socketserver.TCPServer(('localhost', 0), handler)
-  port_message = 'bind succeeded, server port is: ' \
-      + str(httpd.server_address[1])
-  print(port_message)
-  httpd.serve_forever()
-except:
-  print("bind failed")
+httpd = six.moves.socketserver.TCPServer(('localhost', 0), handler)
+port_message = 'bind succeeded, server port is: ' \
+    + str(httpd.server_address[1])
+print(port_message)
+httpd.serve_forever()
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -113,12 +113,11 @@ class TestServerProcess(unittest_toolbox.Modified_TestCase):
 
 
   def test_server_exit_before_timeout(self):
-    # Test starting a non existing server file."
-    self.assertRaises(ChildProcessError, utils.TestServerProcess, logger,
+    self.assertRaises(utils.ChildProcessError, utils.TestServerProcess, logger,
         server='non_existing_server.py')
 
     # Test starting a server which immediately exits."
-    self.assertRaises(ChildProcessError, utils.TestServerProcess, logger,
+    self.assertRaises(utils.ChildProcessError, utils.TestServerProcess, logger,
         server='fast_server_exit.py')
 
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -112,6 +112,16 @@ class TestServerProcess(unittest_toolbox.Modified_TestCase):
     self.assertFalse(self.server_process_handler.is_process_running())
 
 
+  def test_server_exit_before_timeout(self):
+    # Test starting a non existing server file."
+    self.assertRaises(ChildProcessError, utils.TestServerProcess, logger,
+        server='non_existing_server.py')
+
+    # Test starting a server which immediately exits."
+    self.assertRaises(ChildProcessError, utils.TestServerProcess, logger,
+        server='fast_server_exit.py')
+
+
 if __name__ == '__main__':
   utils.configure_test_logging(sys.argv)
   unittest.main()
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python
+
+# Copyright 2020, TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""
+<Program Name>
+  test_utils.py
+
+<Author>
+  Martin Vrachev.
+
+<Started>
+  October 21, 2020.
+
+<Copyright>
+  See LICENSE-MIT OR LICENSE for licensing information.
+
+<Purpose>
+  Provide tests for some of the functions in utils.py module.
+"""
+
+import os
+import logging
+import unittest
+import socket
+import sys
+
+import tuf.unittest_toolbox as unittest_toolbox
+
+import utils
+
+logger = logging.getLogger(__name__)
+
+class TestServerProcess(unittest_toolbox.Modified_TestCase):
+
+  def tearDown(self):
+    # Make sure we are calling clean on existing attribute.
+    if hasattr(self, 'server_process_handler'):
+      self.server_process_handler.clean()
+
+
+  def can_connect(self):
+    succeed = False
+    try:
+      sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+      sock.connect(('localhost', self.server_process_handler.port))
+      succeed = True
+    except:
+      pass
+    finally:
+      if sock:
+        sock.close()
+    return succeed
+
+
+  def test_simple_server_startup(self):
+    # Test normal case
+    self.server_process_handler = utils.TestServerProcess(log=logger)
+
+    # Make sure we can connect to the server
+    self.assertTrue(self.can_connect())
+
+
+  def test_simple_https_server_startup(self):
+    # Test normal case
+    good_cert_path = os.path.join('ssl_certs', 'ssl_cert.crt')
+    self.server_process_handler = utils.TestServerProcess(log=logger,
+        server='simple_https_server.py', extra_cmd_args=[good_cert_path])
+
+    # Make sure we can connect to the server
+    self.assertTrue(self.can_connect())
+
+
+  @unittest.skipIf(sys.version_info.major != 2, "Test for Python 2.X")
+  def test_proxy_server_startup(self):
+    # Test normal case
+    self.server_process_handler = utils.TestServerProcess(log=logger,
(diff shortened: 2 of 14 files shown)


MVrachev:
I had some discussions with @jku  should we add tests for the `broken cases` when we create a server process.
For example, when we don't receive the expected message in the temp log file from the server or the server process has just died, etc.
The problem is, that this would mean to wait for the whole `timeout` to finish and wait for `Timeout` exception to be raised.
We won't want to use the default timeout of 10 seconds for each of those tests given that the server process is started for like `0.10 ~ 0.20` seconds on a successful run on my VM.
On another hand, we don't want to make the timeout too low, because Travis CI or AppVeyor could be too slow to start the process and not start it at all in which case when we receive `Timeout` exception we haven't tested anything...


One solution I can think of is using the default timeout and moving these tests to a different dir which will be executed only by
Travis CI and AppVeyor. Those tests are not exactly unit tests but more of integration tests, so for me, it sounds logical.

tests/test_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    if hasattr(self, 'server_process_handler'):
+      self.server_process_handler.clean()
+
+
+  def can_connect(self):
+    succeed = False
+    try:
+      sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+      sock.connect(('localhost', self.server_process_handler.port))
+      succeed = True
```
jku: nit: you can just `return True` here (and False in the exception handler) without a variable

tests/utils.py
```diff
@@ -214,9 +278,11 @@ def clean(self):
...
 
-    if self.__server_process.returncode is None:
-      self.__logger.info('Server process ' + str(self.__server_process.pid) +
-          ' terminated.')
-      self.__server_process.kill()
-      self.__server_process.wait()
+
+
+  def is_process_running(self):
+    """Returns a boolean value if the server process is currently running."""
```
jku: nit: this comment literally does not tell me anything... It probably does not need to exit

tests/utils.py
```diff
@@ -154,43 +147,114 @@ class TestServerProcess():
...
-    if timeout > 0:
-      try:
-        wait_for_server('localhost', self.server, self.port, timeout)
-      except Exception as e:
-        # Make sure that errors from the server side will be logged.
-        self.flush_log()
-        raise e
+
+  def _is_port_found_in_log(self):
+    """Checks if the port number is sent from the server subprocess."""
```
jku: nit: this also sets self.port -- so I guess the name could imply it's not just a check. I don't have great suggestions though

tests/utils.py
```diff
@@ -154,43 +147,114 @@ class TestServerProcess():
...
+
+  def _start_server(self, timeout, extra_cmd_args, popen_cwd):
+    """Start the server subprocess. Uses a retry mechanism
+    if the bind fails."""
+
+    success = False
+    retries = 0
+    elapsed = 0
+    start = time.time()
+    while not success and elapsed < timeout:
```
jku: I understood this outer loop should not needed anymore  -- why are we retrying if we expect the bind() to choose a free port for us in the subprocess?

The docstring above should be then changed as well: no retries

jku (changes requested):
I like this! The only hack here is parsing server logs to get the port but I believe that's a decent price to pay (and arguably less of a hack than retrying on bind() failure). Tests should be more reliable and faster because of this -- and not just because one very slow test was removed.

Only real question is `why _start_server()` still has two loops:  I thought the point was that the outer loop was replaced by letting bind() pick a port...

MVrachev:
> I like this! The only hack here is parsing server logs to get the port but I believe that's a decent price to pay (and arguably less of a hack than retrying on bind() failure). Tests should be more reliable and faster because of this -- and not just because one very slow test was removed.
> 
> Only real question is `why _start_server()` still has two loops: I thought the point was that the outer loop was replaced by letting bind() pick a port...

Yes, I was thinking that if we have a problem starting the server we can try again if we haven't finished the timeout?
Or probably it's best just to fail because it won't be a problem related to already taken port...

tests/test_utils.py
MVrachev: Yes, but that way I can use finally and close the socket there.
Otherwise, I will have to add the `close the socket` part in both execpt and try blocks.

jku:
> Or probably it's best just to fail because it won't be a problem related to already taken port...

Yes: we have no idea why it might have failed so no reason to believe it would be fixed by retrying.

tests/test_utils.py
jku: finally will work even if you return

tests/utils.py
MVrachev: I agree.
I just felt obligated to add it when all of the other functions have docstrings.

jku:
Oh one more thing about the new tests: Apart from the clean-test I think everything else is already tested by the current tests. The new tests are obviously more "unit-testy" since they only test a single thing ... but still: is there value in adding tests that are basically subsets of existing tests?

tests/test_utils.py
MVrachev: Didn't know that and tested and you are right.
That's cool!

MVrachev:
> Oh one more thing about the new tests: Apart from the clean-test I think everything else is already tested by the current tests. The new tests are obviously more "unit-testy" since they only test a single thing ... but still: is there value in adding tests that are basically subsets of existing tests?

They provide us a way to clearly see when we have a bug while starting any of the subprocesses running a server script, which is not always easy to be seen given that most of those server subprocesses are initialized in `setUpClass` where we can't receive stdout or stderr.

The tests on my virtual machine are run for around `0.5 - 1.0` seconds and with python2 (or none of the tests is skipped) is the same.

trishankatdatadog:
Just wanted to say: keep up the great work, y'all! cc @sechkova @joshuagl

tests/utils.py
MVrachev: Yes, I will remove this loop.

tests/utils.py
```diff
@@ -167,40 +167,33 @@ def __init__(self, log, server='simple_server.py',
...
+    while elapsed < timeout:
+      if not self.is_process_running():
+        break
+      elif self._set_port_if_in_logs():
+        # If the port is in the log, then the bind was successful.
+        success = True
+        break
 
         time.sleep(0.01)
         elapsed = time.time() - start
```
jku: wrong indent now -- should not be inside the conditional?

jku (commented):
Getting close... LGTM except for the loop removal that seems incorrect (indent change needed for the elapsed time calculation)

tests/utils.py
```diff
@@ -167,40 +167,33 @@ def __init__(self, log, server='simple_server.py',
...
-          # If the port is in the log, then the bind was successful.
-          success = True
-          break
+
+    self._start_process(extra_cmd_args, popen_cwd)
+
+    # loop until bind succeeds, server exits or we timeout
+    while elapsed < timeout:
+      if not self.is_process_running():
+        break
```
jku: actually one more thing: this now leads to TimeoutError -- that's not entirely correct anymore (as we only know that the server exited unexpectedly)

jku: `raise ChildProcessError` here maybe? (it's not entirely correct but maybe close enough if you don't want to make a new exception)

MVrachev: Yes, that's a good point.
Just forget about it.
Will use `continue` here.

tests/utils.py
MVrachev: Yes, a good point.
Forgot to fix it.

tests/utils.py
MVrachev: But wait if the process hasn't started then `_set_port_if_in_logs` will return `False` and the process will sleep for a little and start again.
I think we don't need that check at all.
I will remove it.

MVrachev: I added a comprehensive `ChildProcessError` exception.

MVrachev:
I realized that now when we had removed the outer loop I can add tests that check the case when a server has exited before the timeout has expired.

tests/simple_server.py
```diff
@@ -73,6 +65,12 @@ def log_request(self, code='-', size='-'):
...
 
-httpd.serve_forever()
+try:
+  httpd = six.moves.socketserver.TCPServer(('localhost', 0), handler)
+  port_message = 'bind succeeded, server port is: ' \
+      + str(httpd.server_address[1])
+  print(port_message)
+  httpd.serve_forever()
+except:
+  print("bind failed")
```
jku: we discussed this and came to the conclusion that the exception handling is not useful in this version -- the log message  is not needed and actively hides possible other error (exception) messages

jku: this applies to all servers

jku (approved):
Man from Del Monte says yes :+1:

tests/utils.py
```diff
@@ -37,6 +37,7 @@
 try:
   # is defined in Python 3
   TimeoutError
+  ChildProcessError
```
jku, line 40: This will not work:
* if TimeoutError fails, you define class ChildProcessError?
* the code that uses this is in another file and uses utils.ChildProcessError -- that won't be defined in python 3

jku: I think it makes sense to just always define an error of your own since ChildProcessError is not in python2 (it's not 100% correct even in python3): `TestServerError` or something.

jku:
Since the failures are now hidden in the UI: Earlier test runs (e.g. https://ci.appveyor.com/project/theupdateframework/tuf/builds/36033643/job/0pm05aeiihcn9nad for commit 9e53c4a3b) had error  like this:
 
```
ERROR: test_get_valid_targetinfo (test_updater.TestMultiRepoUpdater)
  File "C:\projects\tuf\tests\test_updater.py", line 1875, in setUp
    server=self.SIMPLE_SERVER_PATH, popen_cwd=self.repository_directory2)
  File "C:\projects\tuf\tests\utils.py", line 165, in __init__
    raise e
TimeoutError: u'Failure during C:\\projects\\tuf\\tests\\simple_server.py startup!'
```

That's worrying and needs to be investigated.

tests/utils.py
MVrachev: Both `TimeoutError` and `ChildProcessError` are available on python 3.X and not available on python 2.X, so if `TimeoutError` is not found then it makes sense to define both of them.

> the code that uses this is in another file and uses utils.ChildProcessError -- that won't be defined in python 3

for that you are right. I was mistaken in my assumption.

I think that ChildProcessError is a suitable name for that error.
I just think that in `test_utils.py` I will fix that error with:
```
    child_process_error = None
    if sys.version_info.major == 2:
      child_process_error = utils.ChildProcessError
    else:
      child_process_error = ChildProcessError
```

jku: > Both `TimeoutError` and `ChildProcessError` are available on python 3.X and not available on python 2.X, so if `TimeoutError` is not found then it makes sense to define both of them.

no. You can either decide things based on python version or you can decide things based on feature testing. Using feature testing as if it is python version testing is wrong even if it happens to work right now.

MVrachev:
After the findings made by @jku and described here https://github.com/theupdateframework/tuf/issues/1196
I made changes to the `Delegate port generation for the tests to the OS` commit following his idea about using a thread-safe Queue for process communication and also had to squash a few other commits.

That's why **it makes sense to close this pr and open another pr which uses this new method.**
Of course, I could reuse this pr, but given that I have rebased all of the commits and I will have to change the pr message together with some of the commit messages I think this would be more confusing.

Issue #1139, opened 2020-09-10
Add classes for complex metadata fields


lukpueh:
Coordinate with #1133 and #1137 

**Description of issue or feature request**:
For consistent use of OOP we should provide classes for all complex metadata fields
and not only for the outer metadata signature wrapper and the inner `signed` metadata.

This would make the used metadata format more explicit and might have a positive effect on metadata schema validation (see discussion in #1130).

**Current behavior**:
Mixed use of classes and dictionaries for TUF metadata

**Expected behavior**:
Provide classes for all complex metadata fields, i.e. `'signatures'` (in
`Metadata`) `'meta'`/`'targets'` (in `Timestamp`, `Snapshot`, `Targets`), `'delegations'`
(in `Targets`), `'keys'`/`'roles'` (in not yet existent `Delegation`), ...

Issue #1134, opened 2020-09-10
How much metadata behaviour should be implemented on their classes?


lukpueh:
The TUF refactor aims at a more idiomatic use of OOP (see #1112). As such it seems reasonable to implement metadata entity behaviour as instance methods on the corresponding classes.

However, not all users of the metadata model need the same behaviour. For instance, a TUF client is likely to only need `Metadata.verify` and read access on all metadata object attributes, but none of the methods to modify attributes, which are needed by a  repository library/tool,  (e.g. `sign`, `bump_version`, `bump_expiration`, `delegate`, `add_keys`, `add_targets`, etc.).

This question about how to draw the line is especially important if unneeded functionality requires 3rd party dependencies, which we would have to vendor on a client.

Some possible approaches (brain storming):
- Use classes exclusively for attributes (except maybe for methods on sslib's `Metadata` and `Signed` classes, such as `sign`, `verify`, `canonicalise`) and implement all behaviour on metadata user specific controller classes, e.g. `Repository`, `Client`  (or something like `Project`/`Developer`/`Workspace` for PEP480.
- Use Subclasses, e.g. `RepositoryTargets`, `ClientTargets`, etc...
- Expose all methods to all users of the model, but handle missing optional dependencies. E.g. raise something like an `UnsupportedLibraryError` if a client calls `Metadata.sign` but does not have the underlying optional  3rd-party dependencies (see https://github.com/secure-systems-lab/securesystemslib/issues/179)

Premise: Find a balance between OOP purism and pragmatism

lukpueh:
Related discussion: Is there a need for library functions that should be implemented neither on a metadata class nor on a controller class?

sechkova:
_dropping some thoughts here ..._ 

On one hand the general preference should be towards composition over inheritance, in this case the controller classes option.
On the other hand, `Client` needs only a subset of all metadata methods, meaning there is a common base functionality between `Client` and `Repository` which repository extends which hints toward inheritance. However, `Signed` class already has one level of inheritance hierarchy which I find meaningful and I don't want to suggest ruining.

Having said that maybe something like option one: 

> Use classes exclusively for attributes (except maybe for methods on sslib's Metadata and Signed classes, such as sign, verify, canonicalise)

plus a base Controller class that can be extended by a RepositoryController, if needed, seems like the best compromise.

I like option three too, 
[...]

PR #1193, opened 2020-10-28
Add root metadata class to new TUF metadata model

### LINKED ISSUE #1137: Add root metadata class to new TUF metadata model

Add root  metadata class to `tuf.api.metadata` module and implement (de)serialisation methods (`to_dict` and `from_dict`) and adequate metadata update methods (see `tuf.api.Signed`-subclasses for inspiration).

**Current behavior**:
No Root class in `tuf.api.metadata`

**Expected behavior**:
```python
# In tuf/api/metdata.py
class Root(Signed):
    ...

```

sechkova:
Fixes #1137

- Added root metadata class to `tuf.api.metadata` module and implements serialization and update methods.
- Added root and targets metadata tests

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -88,8 +88,7 @@ class Metadata():
         elif _type == 'timestamp':
             inner_cls = Timestamp
         elif _type == 'root':
-            # TODO: implement Root class
-            raise NotImplementedError('Root not yet implemented')
+            inner_cls = Root
         else:
             raise ValueError(f'unrecognized metadata type "{_type}"')
 
@@ -333,6 +332,89 @@ class Signed:
     def bump_version(self) -> None:
         """Increments the metadata version number by 1."""
         self.version += 1
+
+
+class Root(Signed):
+    """A container for the signed part of root metadata.
+
+    Attributes:
+        consistent_snapshot: A boolean indicating whether the repository
+            supports consistent snapshots.
+        keys: A dictionary that contains a public key store used to verify
+            top level roles metadata signatures::
+            {
+                '<KEYID>': {
+                    'keytype': '<KEY TYPE>',
+                    'scheme': '<KEY SCHEME>',
+                    'keyid_hash_algorithms': [
+                        '<HASH ALGO 1>',
+                        '<HASH ALGO 2>'
+                        ...
+                    ],
+                    'keyval': {
+                        'public': '<PUBLIC KEY HEX REPRESENTATION>'
+                    }
+                },
+                ...
+            },
+        roles: A dictionary that contains a list of signing keyids and
+            a signature threshold for each top level role::
+            {
+                '<ROLE>': {
+                    'keyids': ['<SIGNING KEY KEYID>', ...],
+                    'threshold': <SIGNATURE THRESHOLD>,
+                },
+                ...
+            }
+
+    """
+    # TODO: determine an appropriate value for max-args and fix places where
+    # we violate that. This __init__ function takes 7 arguments, whereas the
+    # default max-args value for pylint is 5
+    # pylint: disable=too-many-arguments
+    def __init__(
+            self, _type: str, version: int, spec_version: str,
+            expires: datetime, consistent_snapshot: bool,
+            keys: JsonDict, roles: JsonDict) -> None:
+        super().__init__(_type, version, spec_version, expires)
+        # TODO: Add classes for keys and roles
+        self.consistent_snapshot = consistent_snapshot
+        self.keys = keys
+        self.roles = roles
+
+
+    # Serialization.
+    def to_dict(self) -> JsonDict:
+        """Returns the JSON-serializable dictionary representation of self. """
+        json_dict = super().to_dict()
+        json_dict.update({
+            'consistent_snapshot': self.consistent_snapshot,
+            'keys': self.keys,
+            'roles': self.roles
+        })
+        return json_dict
+
+
+    # Update key for a role.
+    def add_key(self, role: str, keyid: str, key_metadata: JsonDict) -> None:
+        """Adds new key for 'role' and updates the key store. """
+        if keyid not in self.roles[role]['keyids']:
+            self.roles[role]['keyids'].append(keyid)
+            self.keys[keyid] = key_metadata
+
+
+    # Remove key for a role.
+    def remove_key(self, role: str, keyid: str) -> None:
+        """Removes key for 'role' and updates the key store. """
+        if keyid in self.roles[role]['keyids']:
+            self.roles[role]['keyids'].remove(keyid)
+            for keyinfo in self.roles.values():
+                if keyid in keyinfo['keyids']:
+                    return
+
+            del self.keys[keyid]
+
+
 
 
 class Timestamp(Signed):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -43,6 +43,10 @@ if IS_PY_VERSION_SUPPORTED:
         import_ed25519_privatekey_from_file
     )
 
+    from securesystemslib.keys import (
+        format_keyval_to_metadata
+    )
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -335,6 +334,76 @@ def bump_version(self) -> None:
         self.version += 1
 
 
+class Root(Signed):
+    """A container for the signed part of root metadata.
+
+    Attributes:
+        consistent_snapshot: A boolean indicating whether the repository
+            supports consistent snapshots,
```
```suggestion
            supports consistent snapshots.
```

tuf/api/metadata.py
```diff
@@ -335,6 +334,76 @@ def bump_version(self) -> None:
...
+        keys: A dictionary that contains a public key store used to verify
+            top level roles metadata signatures::
```
joshuagl, line 344:
```suggestion
            top level roles metadata signatures:
```

tuf/api/metadata.py
```diff
@@ -335,6 +334,76 @@ def bump_version(self) -> None:
...
+                        ...
+                    ],
+                    'keyval': {
+                        'public': '<PUBLIC KEY HEX REPRESENTATION>'
+                    }
+                },
+                ...
+            },
+        roles: A dictionary that contains a list of signing keyids and
+            a signature threshold for each top level role::
```
joshuagl, line 361:
```suggestion
            a signature threshold for each top level role:
```

tuf/api/metadata.py
```diff
@@ -335,6 +334,76 @@ def bump_version(self) -> None:
...
+            'consistent_snapshot': self.consistent_snapshot,
+            'keys': self.keys,
+            'roles': self.roles
+        })
+        return json_dict
+
+
+    # Modification.
+    def update(self, keys: JsonDict, roles: JsonDict,
+        consistent_snapshot: bool) -> None:
```
joshuagl: I think we might have an API granularity mismatch here. 

`Snapshot.update()` takes information for a single targets metadata file, to update/add to the listed files in the snapshot metadata file.

`Targets.update()` takes information for a single target file, to update/add the fileinfo for a single target file. 

Whereas `Root.update()` is overwriting all of the keys and roles defined in the root metadata.
Perhaps `keys` and `roles` parameters should be optional and singular (`key` and `role`), taking a single keydict to add/update and a single roleinfo to add/update? That probably gets easier with classes for complex metadata fields (#1139).

I'd appreciate hearing from some library users (@trishankatdatadog, @woodruffw, @jku) and co-maintainers (@lukpueh) on this topic of API granularity.

joshuagl (commented):
Great catch on the use of dictionaries in the test!

I think your proposed interface for `Root.update()` here might be inconsistent with the API in the other metadata classes. More details on that inline, what do you think?

tuf/api/metadata.py
sechkova: You didn't mind the double colon before 😉 
I mean I kept them because they seemed intentionally used throughout the docs of the Signed class 🙂

lukpueh: It's sphinx markup for a subsequent verbatim code block. Probably something we should mention in our shadow/diff code style guideline (https://github.com/secure-systems-lab/code-style-guidelines/issues/20).

tuf/api/metadata.py
sechkova: Now that you've said it, it seems you have a point. I think I like an api of the form:
`Root.update('timestamp', key)` 
but what if you want to revoke a key and remove it, not only add/update one?
Should we then extend `Root` with `Root.remove/delete(role, key)` or implement this logic ... somewhere else?

How do you feel about consistent snapshot? Maybe we need a setter/getter pattern for toggling it independently? I am not sure ... a user of the class can access it anyway directly

tuf/api/metadata.py
joshuagl: 🤦

lukpueh: Easy to mistake for a typo.

tuf/api/metadata.py
lukpueh: Good questions! I think the `update` methods on Targets/Snapshot/Timestamp are supposed to facilitate the  chain of (currently manual) updates that are necessary when the target files in the repository are updated.

Root is not part of that chain, so it may not need to be consistent with the other `update` functions. Regardless, something like `Root.update/delete(role_name, key)` IMHO seems to add more value than mere (multi-)setter logic. 

Generally, I think we should do in these methods whatever is convenient for updating an individual role either manually or from within methods like  `Repository.update_targets(...)`,  `Repository.create_snapshot(...)`, `Repository.rotate_keys(...)` etc., i.e. the methods that actually *do TUF*.

This discussion seems related to #1134.

joshuagl: > Generally, I think we should do in these methods whatever is convenient for updating an individual role either manually or from within methods like `Repository.update_targets(...)`, `Repository.create_snapshot(...)`, `Repository.rotate_keys(...)` etc., i.e. the methods that actually _do TUF_.

I think that sounds like a sensible approach.


I like the `Root.update()` and `Root.remove()` interfaces you've proposed here @sechkova. Do they fulfil the needs which led you to create this PR?

I think the best path forward for this PR would be to implement as much of `Root` functionality as you need to proceed. We can modify/extend the metadata API as we work on making use of it from higher-level pieces of the refactor.

sechkova: I worked on this PR in the context of the client meaning that I only need read access to  'signed' dictionaries.
Anyway I think we all agree on implementing a set of `Root.update()` and `Root.remove()` interfaces which won't cost me much and will be valuable for moving the metadata API forward.

sechkova:
Added two root methods that should be enough for a key revocation:

`root.add_key(role, keyid, key_metadata)` - root.update seemed ambiguous since a new keyid is appended to the role's list of keyids
`root.remove_key(role, keyid)`

I think this is what we can do for now, until we decide how to represent the internal metadata fields. For the same reason I haven't added any errors/warnings on missing/already existing keyids but I will if you don't like the idea of silently ignoring these cases.

joshuagl:
> `root.add_key(role, keyid, key_metadata)` - root.update seemed ambiguous since a new keyid is appended to the role's list of keyids

`update()` would be consistent with `Targets.update()`, which either replaces an existing entry in the targets dict, or adds a new one when the filename (key) is new.

sechkova:
> > `root.add_key(role, keyid, key_metadata)` - root.update seemed ambiguous since a new keyid is appended to the role's list of keyids
> 
> `update()` would be consistent with `Targets.update()`, which either replaces an existing entry in the targets dict, or adds a new one when the filename (key) is new.

I see your point about consistency but what would `root.update('targets', keyid ...)` mean with this metadata format:

```
 "targets": {
    "keyids": [
     "135c2f50e57ff11e744d234a62cebad8c38daf399604a7655661cc9199c69164"
    ],
    "threshold": 1
   },
```

We don't expect different keys with the same keyid, do we?

joshuagl:
> > > `root.add_key(role, keyid, key_metadata)` - root.update seemed ambiguous since a new keyid is appended to the role's list of keyids
> > 
> > 
> > `update()` would be consistent with `Targets.update()`, which either replaces an existing entry in the targets dict, or adds a new one when the filename (key) is new.
> 
> I see your point about consistency but what would `root.update('targets', keyid ...)` mean with this metadata format:
> 
> ```
>  "targets": {
>     "keyids": [
>      "135c2f50e57ff11e744d234a62cebad8c38daf399604a7655661cc9199c69164"
>     ],
>     "threshold": 1
>    },
> ```

You're right, `update()` doesn't make much sense with a list. 👍 

> We don't expect different keys with the same keyid, do we?

Not as I understand it. From [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md)

> keyids defined within a metadata file must be unique. For example, a root file that delegates trust to root, snapshot, timestamp, and top-level targets should provide unique keyids for each key trusted to sign metadata for these roles.

joshuagl (approved):
Thanks for working through the details with me

joshuagl:
Trying a close/re-open to trigger CI.

PR #1194, opened 2020-10-30
Speedup indefinite freeze tests

jku:
Remove `sleep()` calls worth ~15 seconds from indefinite freeze tests. Instead mock `time.time()` so Updater thinks it's running in the future.

Also add mentions of ExpiredMetadataError in Updater docs

### DIFF
--- a/tests/test_indefinite_freeze_attack.py
+++ b/tests/test_indefinite_freeze_attack.py
@@ -45,7 +45,6 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import datetime
-import unittest.mock as mock
 import os
 import time
 import tempfile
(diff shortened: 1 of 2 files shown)


tests/test_indefinite_freeze_attack.py
```diff
@@ -44,6 +44,8 @@
 from __future__ import division
 from __future__ import unicode_literals
 
+import datetime
+import unittest.mock as mock
```
lukpueh: We need a conditional mock import à la...
```
if sys.version_info >= (3, 3):
  import unittest.mock as mock
else:
  import mock
```

MVrachev (approved):
Really liked this pr!
A perfect place to use mocking and the time is decreased by a significant margin!
LGTM!

joshuagl (approved):
Nice set of changes here, thanks @jku !

lukpueh (approved):
Thanks for the patch, @jku! This is great work! 🎉

PR #1198, opened 2020-11-04
Tests: Use Queue for process communication which replaces tmp files and use OS for port creation

### LINKED ISSUE #1111: tests (simple_server.py) still unreliable

### LINKED ISSUE #1124: Add a method to create a server subprocesses only on unused local ports

### LINKED ISSUE #1196: tests: server logging  via tempfile may have threading issues
Filing this after debugging rare errors in https://github.com/theupdateframework/tuf/pull/1192

The tempfile approach to capturing server logging works fine in current develop branch but it triggers issues in PR 1192 where we start reading from the file at a rapid pace -- based on the results this seems to be a threading issue in file object as the log contents become corrupted (rarely but still reproducibly)

I just tried another approach (worker thread + Queue instead of a TemporaryFile) and it seems to work (this is completely unintegrated into the tests, just an example):

```python
try:
  import queue
except ImportError:
  import Queue as queue # python2

import subprocess
import sys
import threading
import time


# Worker function to run in separate thread
# Reads from 'stream' (stdout), puts lines in Queue 'line_queue' (Queue is thread-safe)
def log_queue_worker(stream, line_queue):
  while True:
    log_line = stream.readline()
    line_queue.put(log_line)
    if len(log_line) == 0:
      # EOS
      break

# Start child process
proc = subprocess.Popen([sys.executable, "-u", "proxy_server.py"],
    stdout=subprocess.PIPE, stderr=subprocess.STDOUT)

# Run log_queue_worker() in a thread
# The thread will exit when child process dies
log_queue = queue.Queue()
log_thread = threading.Thread(target=log_queue_worker, args=(proc.stdout, log_queue))
log_thread.daemon = True
log_thread.start()

# setup timeout variables
start = time.time()
elapsed = 0
timeout = 10

# Get lines from log_queue until port is found, child process exits, or we timeout
while True:
  try:
    line = log_queue.get(timeout=timeout - elapsed)
    if len(line) == 0:
      print ("process exit")
      break
    elif line.startswith(b"bind succeeded, server port is: "):
      print (line.rstrip('\n'))
      break
  except queue.Empty:
    print("timeout")
    break
  finally:
    elapsed = time.time() - start

```

This has the added benefit of not needing a single sleep (because the worker thread can block and main thread uses `Queue.get()` timeout)...

MVrachev:
Fixes #1124, Fixes #1111, Fixes #1196 

These changes can be summarized with the following bullets:
- Delegate generation of ports used for the tests to the OS
- Use thread-safe Queue for processes communication
instead of temporary files
- Remove all instances of port generation or hardcoded ports
- Make test_slow_retrieval.py fully conform with TestServerProcess
- Remove test_slow_retrieval expected failure test  (the test with mode_2)
- Add tests for `tests/utils.py`
- Remove all instances of `temp file` mentions in the comments regarding the instances of `TestServerProcess`. 

Delegate generation of ports used for the tests to the OS is much
better than if we manually generate them, because there is always
a chance that the port we have randomly pick turns out to be taken.
By giving 0 to the port argument we ask the OS to give us
an arbitrary unused port.

Use thread-safe Queue for processes communication instead of temporary
files became a necessity because of findings made by Jussi Kukkonen.
With the latest changes made in pr 1192 we were rapidly reading
from the temporary files and Jussi found that it happened rarely
the successful message "bind succeded..." to be corrupted.
It seems, this is a thread issue related to the thread redirecting
the subprocess stdout to the temp file and our thread rapidly
reading from the file.
By using a thread-safe Queue we eliminate this possibility.
For reference read:
#1196

Lastly, test_slow_retrieval.py and slow_retrieval.py were refactored.
Until now, slow_retrieval.py couldn't use the TestServerProcess class
from utils.py for a port generation because of a bug related to
httpd.handle_request().
Now, when we use httpd.serve_forever() we can refactor both of those
files and fully conform with TestServerProcess.

----------------------------------------------------------

This is the 3-rd pr which aims to fix the issues mentioned above.
The other two were closed:

1. [Add a retry mechanism on server startup failure and use TestServerProcess for port generation](https://github.com/theupdateframework/tuf/pull/1169)
2. [Delegate port generation for the tests to the OS ](https://github.com/theupdateframework/tuf/pull/1192)

The first one was closed because I realized that delegating the port generation to the OS was much better
then if we generated a port and then validate it.
The reasons are described here: https://github.com/theupdateframework/tuf/pull/1169#issuecomment-717908695

The second pr was closed because @jku found a rarely appearing thread bug.
You can read my comment about it here: https://github.com/theupdateframework/tuf/pull/1192#issuecomment-721723063

**PS:** Again, many thanks to @jku for reviewing and discussing those issues with me!
Those changes are a good example of what a community can achieve together!

requirements**:

### DIFF
--- a/tests/test_multiple_repositories_integration.py
+++ b/tests/test_multiple_repositories_integration.py
@@ -119,13 +119,6 @@ class TestMultipleRepositoriesIntegration(unittest_toolbox.Modified_TestCase):
     # the pre-generated metadata files have a specific structure, such
     # as a delegated role 'targets/role1', three target files, five key files,
     # etc.
-    self.SERVER_PORT = random.SystemRandom().randint(30000, 45000)
-    self.SERVER_PORT2 = random.SystemRandom().randint(30000, 45000)
-
-    # Avoid duplicate port numbers, to prevent multiple localhosts from
-    # listening on the same port.
-    while self.SERVER_PORT == self.SERVER_PORT2:
-      self.SERVER_PORT2 = random.SystemRandom().randint(30000, 45000)
 
     # Needed because in some tests simple_server.py cannot be found.
     # The reason is that the current working directory
@@ -134,20 +127,18 @@ class TestMultipleRepositoriesIntegration(unittest_toolbox.Modified_TestCase):
 
     # Creates a subprocess running server and uses temp file for logging.
     self.server_process_handler = utils.TestServerProcess(log=logger,
-        port=self.SERVER_PORT, server=SIMPLE_SERVER_PATH,
-        popen_cwd=self.repository_directory)
+        server=SIMPLE_SERVER_PATH, popen_cwd=self.repository_directory)
 
     logger.debug('Server process started.')
 
     # Creates a subprocess running server and uses temp file for logging.
     self.server_process_handler2 = utils.TestServerProcess(log=logger,
-        port=self.SERVER_PORT2, server=SIMPLE_SERVER_PATH,
-        popen_cwd=self.repository_directory2)
+        server=SIMPLE_SERVER_PATH, popen_cwd=self.repository_directory2)
 
     logger.debug('Server process 2 started.')
 
-    url_prefix = 'http://localhost:' + str(self.SERVER_PORT)
-    url_prefix2 = 'http://localhost:' + str(self.SERVER_PORT2)
+    url_prefix = 'http://localhost:' + str(self.server_process_handler.port)
+    url_prefix2 = 'http://localhost:' + str(self.server_process_handler2.port)
 
     self.repository_mirrors = {'mirror1': {'url_prefix': url_prefix,
                                            'metadata_path': 'metadata',
@@ -265,8 +256,10 @@ class TestMultipleRepositoriesIntegration(unittest_toolbox.Modified_TestCase):
 
     # Test the behavior of the multi-repository updater.
     map_file = securesystemslib.util.load_json_file(self.map_file)
-    map_file['repositories'][self.repository_name] = ['http://localhost:' + str(self.SERVER_PORT)]
-    map_file['repositories'][self.repository_name2] = ['http://localhost:' + str(self.SERVER_PORT2)]
+    map_file['repositories'][self.repository_name] = ['http://localhost:' \
+        + str(self.server_process_handler.port)]
+    map_file['repositories'][self.repository_name2] = ['http://localhost:' \
+        + str(self.server_process_handler2.port)]
     with open(self.map_file, 'w') as file_object:
       file_object.write(json.dumps(map_file))
 
--- a/tests/test_multiple_repositories_integration.py
+++ b/tests/test_multiple_repositories_integration.py
@@ -124,13 +124,13 @@ class TestMultipleRepositoriesIntegration(unittest_toolbox.Modified_TestCase):
     # has been changed when executing a subprocess.
     SIMPLE_SERVER_PATH = os.path.join(os.getcwd(), 'simple_server.py')
 
-    # Creates a subprocess running server and uses temp file for logging.
+    # Creates a subprocess running a server.
     self.server_process_handler = utils.TestServerProcess(log=logger,
         server=SIMPLE_SERVER_PATH, popen_cwd=self.repository_directory)
 
     logger.debug('Server process started.')
 
-    # Creates a subprocess running server and uses temp file for logging.
+    # Creates a subprocess running a server.
     self.server_process_handler2 = utils.TestServerProcess(log=logger,
         server=SIMPLE_SERVER_PATH, popen_cwd=self.repository_directory2)
 
@@ -160,8 +160,7 @@ class TestMultipleRepositoriesIntegration(unittest_toolbox.Modified_TestCase):
     # directories that may have been created during each test case.
     unittest_toolbox.Modified_TestCase.tearDown(self)
 
-    # Logs stdout and stderr from the server subprocesses and then it
-    # kills them and closes the temp files used for logging.
+    # Cleans the resources and flush the logged lines (if any).
     self.server_process_handler.clean()
     self.server_process_handler2.clean()
 
--- a/tests/test_multiple_repositories_integration.py
+++ b/tests/test_multiple_repositories_integration.py
@@ -31,7 +31,6 @@ from __future__ import unicode_literals
 
 import os
 import tempfile
-import random
 import logging
 import shutil
 import unittest
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -58,6 +58,7 @@ import logging
 import errno
 import sys
 import unittest
+import json
 
 import tuf
 import tuf.exceptions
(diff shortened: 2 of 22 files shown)


tests/utils.py
```diff
@@ -154,69 +161,169 @@ class TestServerProcess():
...
+    log_thread = threading.Thread(target=self._log_queue_worker,
+        args=(self.__server_process.stdout, self._log_queue))
+
+    # "daemon = True" means the thread won't interfere with the process exit.
+    log_thread.daemon = True
+    log_thread.start()
+
+
+
+  def _log_queue_worker(self, stream, line_queue):
```
jku: If the worker function is an object method (which should be fine), I think it would be good to mark it `@staticmethod `to clearly show that this function should never touch `self` state in any way, except via line_queue.

tests/utils.py
```diff
@@ -154,69 +161,169 @@ class TestServerProcess():
...
+
+
+
+  def _set_port_if_in_logs(self, timeout):
+    """
+    Checks if the port is logged from the server subprocess.
+    If it's found, self.port is set and returns True.
+    Otherwise, returns False.
+    TestServerProcessError is throwned if the process has exited
+    or if it has failed to bind.
```
jku:
```suggestion
    Raises TestServerProcessError if the process has exited.
```

tests/utils.py
```diff
@@ -154,69 +161,169 @@ class TestServerProcess():
...
+      else:
+        # An exception or some other message is printed from the server.
+        self.__logged_messages.append(line)
+        # Check if more lines are logged.
         self.flush_log()
-        raise e
+        raise TestServerProcessError(self.server + ' failed to bind!')
+    except queue.Empty:
+      self.__logger.info("No log lines are found in the Queue!")
+      return False
```
jku: Since it's not called in a loop anymore, this function does not need a return value, I think. If you raise TimeoutError right here, change the method name to something like `wait_for_port()` and update the docstring, the boolean is no longer needed

tests/utils.py
```diff
@@ -154,69 +161,169 @@ class TestServerProcess():
...
+
+    while True:
+      # Get lines from log_queue
+      try:
+        line = self._log_queue.get(block=False)
+        if len(line) == 0:
+          # This means no more lines will be sent in the queue.
+          break
+        else:
+          self.__logged_messages.append(line)
```
jku:
```suggestion
        if len(line) > 0:
          self.__logged_messages.append(line)
```
Just a suggestion to make it easier to read: the break should not be required, next get() raises anyway

tests/utils.py
```diff
@@ -154,69 +161,170 @@ class TestServerProcess():
...
+      except queue.Empty:
+        # No more lines are logged in the queue.
+        break
+
+    if len(self.__logged_messages) > 0:
       title = "Test server (" + self.server + ") output:"
-      message = [title] + log_message.splitlines()
+      message = [title] + self.__logged_messages
       self.__logger.info('\n| '.join(message))
```
jku: I've not noticed `rstrip()` anywhere (and readline() results include '\n') so I expect this leads to double-newlines... Can you confirm that multiline logs still look good ?

jku (commented):
It's been a long process (and that's not your fault) but I think we're about there now: this looks good. I have a couple of suggestions on the newest code (new compared to PR 1192): None of them show stoppers but please take a look.

Would be nice to get some more green CI runs in any case to confirm that this improves reliability like we think it will.

tests/utils.py
```diff
@@ -154,69 +161,170 @@ class TestServerProcess():
...
+      elif line.startswith(expected_msg):
+        self.port = int(line[len(expected_msg):])
+        return True
+      else:
+        # An exception or some other message is printed from the server.
+        self.__logged_messages.append(line)
+        # Check if more lines are logged.
         self.flush_log()
-        raise e
+        raise TestServerProcessError(self.server + ' failed to bind!')
```
jku:
```suggestion
        raise TestServerProcessError(self.server + ' did not print port message as first stdout line as expected')
```
this might be fairer for the future testserver writer whose server binds just fine but does not print what we expect.

tests/test_multiple_repositories_integration.py
```diff
@@ -134,20 +126,18 @@ def setUp(self):
 
     # Creates a subprocess running server and uses temp file for logging.
```
MVrachev: I have to remove all occurrences of `temp file` in the comments when creating the server subprocess...

tests/utils.py
MVrachev: Yes, you were right:
image

I will remove the `\n` from the ` self.__logger.info` call.

A good catch!

MVrachev:
I have updated the comments to not mention the usage of temp file
for logging regarding the instances of the TestServerProcess class.

That's why the number of changed files increased to 22, but most of the new changes are one-liners.

jku (approved):
LGTM, thanks

MVrachev:
I decided to rebase on top of the latest changes in `develop` branch.
I used that opportunity to squash some of my commits.

PS: Running the tests locally with `python aggregate_tests.py` on my Ubuntu 18.04 VM takes only 25~30 seconds after those changes!

tests/test_updater.py
```diff
@@ -1863,27 +1864,35 @@ def setUp(self):
...
-    url_prefix = 'http://localhost:' + str(self.SERVER_PORT)
-    url_prefix2 = 'http://localhost:' + str(self.SERVER_PORT2)
+    url_prefix = 'http://localhost:' + str(self.server_process_handler.port)
+    url_prefix2 = 'http://localhost:' + str(self.server_process_handler2.port)
+
+    # We have all of the necessary information for two repository mirrors
+    # in map.json, except for url prefixes.
+    # For the url prefixes, we create subprocesses that run a server script.
+    # In server scripts we get a free port from the OS which is sent
+    # back to the father process.
```
joshuagl: Nit:
```suggestion
    # back to the parent process.
```

tests/utils.py
```diff
@@ -154,69 +161,161 @@ class TestServerProcess():
...
-    # Create temporary log file used for logging stdout and stderr
-    # of the subprocess. In the mode "r+"" stands for reading and writing
-    # and "t" stands for text mode.
-    self.__temp_log_file = tempfile.TemporaryFile(mode='r+t')
+      timeout=10, popen_cwd=".", extra_cmd_args=[]):
 
     self.server = server
-    self.port = port or random.randint(30000, 45000)
     self.__logger = log
+    # Stores poped messages from the queue.
```
joshuagl: Nit:
```suggestion
    # Stores popped messages from the queue.
```

tests/utils.py
```diff
@@ -154,69 +161,161 @@ class TestServerProcess():
...
+    Worker function to run in a seprate thread.
+    Reads from 'stream', puts lines in a Queue (Queue is thread-safe).
+    """
 
-      # Make sure the file is empty before the next test logs new information.
-      self.__temp_log_file.truncate(0)
+    while True:
+      # readline() is a blocking operation.
+      # decode to push a string in the queue instead of 8-bit bytes.
+      log_line = stream.readline().decode()
```
```suggestion
      log_line = stream.readline().decode('utf-8')
```
Wasn't the default value of 'utf-8' for encoding only added in Python 3? Throughout the rest of the code we consistently call `decode()` with an explicit encoding of 'utf-8' and that seems to be a fairly consistent pattern in my experience. Is that redundant?

joshuagl (commented):
This looks like a really solid set of improvements. Nice work @MVrachev! I've added a few minor comments while we wait for Travis CI.

tests/utils.py
MVrachev: It seems that for python2 the default is `ascii`. 
It's better to specify the encoding while we have python2 code.
By the way: a good catch.

joshuagl (approved):
Looks great, thanks @MVrachev! Could you squash 9bf3faa into cc95abb?

joshuagl:
Close/re-open to prompt CI

PR #1202, opened 2020-11-09
Simplify updater logic for downloading and verifying target files

joshuagl:

The logic for downloading/verifying target files is fairly complex to follow and debug. This PR flattens several internal methods (including a callback) to make the logic simpler to follow.

requirements**:

### DIFF
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1616,13 +1616,13 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
 
 
-  def test_10__hard_check_file_length(self):
+  def test_10__check_file_length(self):
     # Test for exception if file object is not equal to trusted file length.
     with tempfile.TemporaryFile() as temp_file_object:
       temp_file_object.write(b'X')
       temp_file_object.seek(0)
       self.assertRaises(tuf.exceptions.DownloadLengthMismatchError,
-                      self.repository_updater._hard_check_file_length,
+                      self.repository_updater._check_file_length,
                       temp_file_object, 10)
 
 
@@ -1743,26 +1743,6 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
           self.repository_updater._verify_metadata_file,
           metadata_file_object, 'root')
 
-
-  def test_12__get_file(self):
-    # Test for an "unsafe" download, where the file is downloaded up to
-    # a required length (and no more).  The "safe" download approach
-    # downloads an exact required length.
-    targets_path = os.path.join(self.repository_directory, 'metadata', 'targets.json')
-
-    file_size, file_hashes = securesystemslib.util.get_file_details(targets_path)
-    file_type = 'meta'
-
-    def verify_target_file(targets_path):
-      # Every target file must have its length and hashes inspected.
-      self.repository_updater._hard_check_file_length(targets_path, file_size)
-      self.repository_updater._check_hashes(targets_path, file_hashes)
-
-    self.repository_updater._get_file('targets.json', verify_target_file,
-        file_type, file_size, download_safely=True).close()
-
-    self.repository_updater._get_file('targets.json', verify_target_file,
-        file_type, file_size, download_safely=False).close()
 
   def test_13__targets_of_role(self):
     # Test case where a list of targets is given.  By default, the 'targets'
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1629,25 +1629,6 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
 
 
-  def test_10__soft_check_file_length(self):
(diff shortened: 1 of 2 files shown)


tuf/client/updater.py
```diff
@@ -1351,24 +1304,44 @@ def _get_target_file(self, target_filepath, file_length, file_hashes,
...
+
+    # file_mirror (URL): error (Exception)
+    file_mirror_errors = {}
+    file_object = None
+
+    for file_mirror in file_mirrors:
+      try:
+        file_object = tuf.download.safe_download(file_mirror, file_length)
+
+        # Verify 'file_object' against the expected length and hashes.
```
sechkova, line 1327: Do you think it is possible to explain why there is a duplicated check of file length here after the one in `safe_download`?

jku: strictly speaking the checks that are done in functions that safe_download() calls are for the downloaded byte count, whereas _check_file_length() actually looks at the file object. I don't know how that could be of practical difference but...

tuf/client/updater.py
```diff
@@ -1217,7 +1217,7 @@ def _check_hashes(self, file_object, trusted_hashes):
 
 
 
-  def _hard_check_file_length(self, file_object, trusted_file_length):
+  def _check_file_length(self, file_object, trusted_file_length):
```
jku, line 1220: Now that you made me look at this function, I'll comment on it (this may be a bit unrelated to your PR): 

It does 
```
file_object.seek(0)
observed_length = len(file_object.read())
```
which seems wildly unoptimal. I imagine this would be better:
```
file_object.seek(0,2)
observer_length = file_object.tell()
```

JustinCappos: Looks good to me too!  Please add a comment to explain "2" if you add this as it's a magic number here...

jku (approved):

tuf/client/updater.py
joshuagl: Thanks for the suggestion @jku, I pushed an additional patch to implement this optimisation with you listed as co-author.

joshuagl:
Trying close/re-open to handle the CI move

tuf/client/updater.py
trishankatdatadog: Good question, but I think better safe than sorry. I think redundant checks are fine as long as no real impact on performance.

tuf/client/updater.py
```diff
@@ -1351,24 +1306,44 @@ def _get_target_file(self, target_filepath, file_length, file_hashes,
...
+        self._check_hashes(file_object, file_hashes)
+
+      except Exception as exception:
+        # Remember the error from this mirror, and "reset" the target file.
+        logger.debug('Update failed from ' + file_mirror + '.')
+        file_mirror_errors[file_mirror] = exception
+        file_object = None
+
+      else:
+        break
```
trishankatdatadog: Instead of in the `else` here, considering moving the `break` to right after all the checks in `try`.

tests/test_updater.py
```diff
@@ -1763,26 +1744,6 @@ def test_11__verify_metadata_file(self):
           metadata_file_object, 'root')
 
 
-  def test_12__get_file(self):
```
trishankatdatadog, line 1766: Should we replace with a similar test for `_get_target_file()`? Is it tested somewhere else?

tuf/client/updater.py
joshuagl: Nice. Even simpler if we just return `file_object` here and raise the `NoWorkingMirrorError` exception if we exit the loop.

tests/test_updater.py
joshuagl: `_get_target_file()` is exercised by the existing tests for `updater.download_target()` – not sure it warrants an extra test? `download_target()` only marshals the parameters for `_get_target_file()` and ensures that the directory we fetch into exists.

joshuagl: A test for `_get_target_file()` would only be ~5LOC, so no great effort to add.

trishankatdatadog: As long as the function is covered somewhere, I have no strong objection.

trishankatdatadog:
Why is it called `_get_target_file()`? How are metadata downloaded and verified?

joshuagl:
> Why is it called `_get_target_file()`? How are metadata downloaded and verified?

with [`_get_metadata_file()`](https://github.com/theupdateframework/tuf/blob/60dcb95656e8dca156e075dba68ed66715e94c30/tuf/client/updater.py#L1492). It does things slightly differently: doesn't prefix the file with a hash, has different/additional checks for metadata fields (`version` and `spec_version`), checks for expiry, verifies signature, etc.

lukpueh (approved):
Impeccable work, @joshuagl! 🎉  Thanks for deobfuscating. :)

PR #1203, opened 2020-11-09
ADR0002: document deprecation strategy for current release series post 1.0

### LINKED ISSUE #1127: Decide on deprecation strategy
The current plan is to develop the new TUF alongside the old TUF (see #1126). 
How long *after* the refactor is complete will we continue to support the old code base? What does support look like? Bug fixes? Security fixes? We need a roadmap there. The main reason to keep the old code base around is to have a client that supports Python 2.7 (see  #1125). Not sure if this is a good reason for very much longer.

Independently of the refactor, we might generally want to think of a deprecation strategy for the future. E.g. do we want to display deprecation warnings for *n* releases before we drop interface functions? This probably only makes sense if we follow a stricter release schedule than we do now.

joshuagl:

Per the discussion in #1127 opt to support the old release on a best-effort
basis.

requirements**:

### DIFF
--- a/docs/adr/index.md
+++ b/docs/adr/index.md
@@ -6,6 +6,7 @@ This log lists the architectural decisions for tuf.
 
 - [ADR-0000](0000-use-markdown-architectural-decision-records.md) - Use Markdown Architectural Decision Records
 - [ADR-0001](0001-python-version-3-6-plus.md) - Default to Python 3.6 or newer for new development
+- [ADR-0002](0002-pre-1-0-deprecation-strategy.md) - Deprecation strategy
 
 <!-- adrlogstop -->
 
--- a/docs/adr/0002-pre-1-0-deprecation-strategy.md
+++ b/docs/adr/0002-pre-1-0-deprecation-strategy.md
@@ -1,0 +1,52 @@
+# Deprecation strategy for current release series (0.x)
+
+* Date: 2020-11-05
+
+Technical Story: https://github.com/theupdateframework/tuf/issues/1127
+
+## Context and Problem Statement
+
+We plan to refactor the reference implementation significantly and, as part of
+that effort, drop support for no-longer maintained versions of Python
+(see ADR 0001).
+
+However, a major user of (and contributor to) the project has users of the
+client stuck on older Python versions.
+
+We would like to define a reasonable support policy for the current, Python 2.7
+supporting, codebase.
+
+## Decision Drivers
+
+* We have finite resources.
+* A major adopter/user of the project has a need to maintain support for
+  Python 2.7 clients.
+
+## Considered Options
+
+* Maintain the code in parallel for a fixed period of time after releasing the
+  refactored code.
+* Abandon the old code once the refactored code is released.
+* Support the old code on a best-effort basis once the refactored code is
+  released.
+
+## Decision Outcome
+
+Chosen option: "Support the old code on a best-effort basis once the refactored
+code is released", because we only have finite resources and want to focus them
+on moving the project forward, including supporting PyPI/pip integration and
+providing a solid implementation for developing specification enhancements in.
+
+We should document this outcome clearly in a governance document describing
+the release process with words along the lines of:
+
+"Support for older releases:
+Bugs reported with tuf versions prior to 1.0.0 will likely not be addressed
+directly by tuf’s maintainers. Pull Requests to fix bugs in the last release
+prior to 1.0.0 will be considered, and merged (subject to normal review
+processes). Note that there may be delays due to the lack of developer resources
+for reviewing such pull requests."
+
+## Links
+
+* [ADR 0001](0001-python-version-3-6-plus.md) Python version


trishankatdatadog (commented):
Thanks for documenting this, Joshua!

Do we have a policy for backporting security fixes?

joshuagl:
Do you think security fixes should be handled differently to what is proposed here? If so, what additional steps do you think the maintainers should commit to?

I'm absolutely open to input here, I'm just not sure what to propose we commit to.

trishankatdatadog:
> Do you think security fixes should be handled differently to what is proposed here? If so, what additional steps do you think the maintainers should commit to?
> 
> I'm absolutely open to input here, I'm just not sure what to propose we commit to.

I think we should at least commit in writing to prioritizing and guaranteeing security fixes for Python 2, but nothing else. WDYT?

joshuagl:
> > Do you think security fixes should be handled differently to what is proposed here? If so, what additional steps do you think the maintainers should commit to?
> > I'm absolutely open to input here, I'm just not sure what to propose we commit to.
> 
> I think we should at least commit in writing to prioritizing and guaranteeing security fixes for Python 2, but nothing else. WDYT?

Prioritising and guaranteeing review/merge/release of security fixes seems reasonable. Is that what you're suggesting? Or did you mean to also suggest implementation of security fixes?

trishankatdatadog:
> Prioritising and guaranteeing review/merge/release of security fixes seems reasonable. Is that what you're suggesting? Or did you mean to also suggest implementation of security fixes?

Both

mnm678:
I think that we can reasonable commit to review and merge security fixes, but actually implementing the fixes may require more resources. To mitigate this, we can adopt a reporting strategy so that any users of the Python 2 implementation will be notified before any security issues become public, and will have a change to submit a patch.

trishankatdatadog:
> I think that we can reasonable commit to review and merge security fixes, but actually implementing the fixes may require more resources. To mitigate this, we can adopt a reporting strategy so that any users of the Python 2 implementation will be notified before any security issues become public, and will have a change to submit a patch.

If someone else privately reports a security issue to us, then we could ask them to submit a patch, too. But what if we find the issue ourselves?

lukpueh:
Why the change of heart, @trishankatdatadog?

@joshuagl already described the strategy proposed here in https://github.com/theupdateframework/tuf/issues/1127#issuecomment-713663707, and you seemed to deem it reasonable. (https://github.com/theupdateframework/tuf/issues/1127#issuecomment-713670216).

I admit that in the original proposal the best-effort basis sounded a bit more optimistic ...
> * that we will review any functional bug and security fixes proposed to the branch (PRs)
> * that severe security issues and functional bugs will be addressed on a best effort basis
> * that we will release updates with any generated fixes in a timely fashion

... than it does now ...

>  Bugs reported with tuf versions prior to 1.0.0 will likely not be addressed directly by tuf’s maintainers.

But I think the essence is the same.

lukpueh (commented):
Many thanks for the great write-up, @joshuagl! In #1127 you talked about:
> a deprecation strategy for the current release series [...] and
> a deprecation/API stability strategy for releases after the refactor.

IIUC this document only addresses the former. Would you mind clarifying the distinction in the title (and maybe document name)?

trishankatdatadog:
> Why the change of heart, @trishankatdatadog?

Sorry, what change of heart?

lukpueh:
> Sorry, what change of heart?

Oh, apologies, I missed an important pronoun in my comment above.

I was just curious why you now wanted to commit to more (https://github.com/theupdateframework/tuf/pull/1203#issuecomment-726801429) than what @joshuagl had already described in https://github.com/theupdateframework/tuf/issues/1127#issuecomment-713663707, which you seemed to deem reasonable (https://github.com/theupdateframework/tuf/issues/1127#issuecomment-713670216).

trishankatdatadog:
> I was just curious why you now wanted to commit to more ([#1203 (comment)](https://github.com/theupdateframework/tuf/pull/1203#issuecomment-726801429)) than what @joshuagl had already described in [#1127 (comment)](https://github.com/theupdateframework/tuf/issues/1127#issuecomment-713663707), which you seemed to deem reasonable ([#1127 (comment)](https://github.com/theupdateframework/tuf/issues/1127#issuecomment-713670216)).

I see. Yes, I still agree. I just think it's important to clearly state we should make an exception for security fixes, just like [Python 2](https://www.python.org/doc/sunset-python-2/)... until I saw that, nope, they don't guarantee security fixes either. Huh, wow. If that's the case, it may not be a terrible precedent. May encourage people to seriously upgrade. Not a bad idea. I know we're a more security-based project, but that doesn't mean we have to support everything forever. Ok, let's stick to the original plan.

trishankatdatadog (approved):

docs/adr/0002-deprecation-strategy.md
```diff
@@ -0,0 +1,52 @@
...
+
+* Date: 2020-11-05
+
+Technical Story: https://github.com/theupdateframework/tuf/issues/1127
+
+## Context and Problem Statement
+
+We plan to refactor the reference implementation significantly and, as part of
+that effort, drop support for no-longer maintained versions of Python
+(see ADR 0002).
```
sechkova: I think this should be a reference to ADR 0001

joshuagl: Absolutely right, good catch!

joshuagl:
Thank you all for continuing the discussion in my absence.

> Many thanks for the great write-up, @joshuagl! In #1127 you talked about:
> 
> > a deprecation strategy for the current release series [...] and
> > a deprecation/API stability strategy for releases after the refactor.
> 
> IIUC this document only addresses the former. Would you mind clarifying the distinction in the title (and maybe document name)?

Well spotted, thank you for the thorough review. I've pushed a fixup commit 8685898 which renames the document and changes the document title.

docs/adr/index.md
```diff
@@ -6,6 +6,7 @@ This log lists the architectural decisions for tuf.
 
 - [ADR-0000](0000-use-markdown-architectural-decision-records.md) - Use Markdown Architectural Decision Records
 - [ADR-0001](0001-python-version-3-6-plus.md) - Default to Python 3.6 or newer for new development
+- [ADR-0002](0002-deprecation-strategy.md) - Deprecation strategy
```
lukpueh:
```suggestion
- [ADR-0002](0002-pre-1-0-deprecation-strategy.md) - Deprecation strategy
```

lukpueh (changes requested):
Please update link, other than that LGTM! :)

lukpueh (approved):

lukpueh:
Thanks for the updates, @joshuagl. No need to wait for appveyor here. Merging...

PR #1205, opened 2020-11-10
Refactor _get_metadata_file() from updater.py

MVrachev:

Currently, the `_get_metadata_file()` from `tuf/client/updater.py` is really large 
and does too many things and some of them partially.
It validates the specification version and the metadata version in it,
but doesn't validate if the metadata has expired or if the signature could be trusted.
So, it's partially validating the metadata file and at the same time calling 
`_verify_metadata_file()` from updater.py for the rest of the work which doesn't make
sense.

It's logically better if we move the validation of the specification
and the metadata version in separate functions which would
be then called in ` _verify_metadata_file()` and that way move all of the validation
will be done in ` _verify_metadata_file()`.

Also, this pr is needed if we want to validate locally available metadata files.
Currently, the problem is that part of the validation cannot be done for local files, because
the spec and metadata versions validation is done in `get_metadata_file()` which is 
used for accessing a metada file somewhere on the internet.

requirements**:

### DIFF
--- a/tests/test_endless_data_attack.py
+++ b/tests/test_endless_data_attack.py
@@ -269,8 +269,10 @@ class TestEndlessDataAttack(unittest_toolbox.Modified_TestCase):
       self.repository_updater.refresh()
 
     except tuf.exceptions.NoWorkingMirrorError as exception:
-      for mirror_url, mirror_error in six.iteritems(exception.mirror_errors):
-        self.assertTrue(isinstance(mirror_error, securesystemslib.exceptions.Error))
+      for _, mirror_error in six.iteritems(exception.mirror_errors):
+        # Throw tuf.exceptions.InvalidMetadataJSONError
+        # because the metadata is not a valud JSON file.
+        self.assertTrue(isinstance(mirror_error, tuf.exceptions.InvalidMetadataJSONError))
 
     else:
       self.fail('TUF did not prevent an endless data attack.')
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1422,15 +1422,13 @@ class Updater(object):
 
 
 
-  def _validate_spec_version(self, metadata_spec_version):
+  def _validate_spec_version(self, metadata_spec_version_str):
     """
     <Purpose>
       Non-public method verifying a metadata specification version.
-      It is assumed that "spec_version" is in (major.minor.fix) format,
-      (for example: "1.4.3") and that releases with the same major version
-      number maintain backward compatibility.
-      Consequently, if the major version number of new metadata equals our
-      expected major version number, the new metadata is safe to parse.
+      It is assumed that "spec_version" is in (major.minor.fix) format
+      and that releases with the same major version number maintain
+      backward compatibility.
 
     <Arguments>
       metadata_spec_version:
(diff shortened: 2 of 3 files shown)


tests/test_endless_data_attack.py
```diff
@@ -269,8 +269,8 @@ def test_with_tuf(self):
       self.repository_updater.refresh()
 
     except tuf.exceptions.NoWorkingMirrorError as exception:
-      for mirror_url, mirror_error in six.iteritems(exception.mirror_errors):
-        self.assertTrue(isinstance(mirror_error, securesystemslib.exceptions.Error))
+      for _, mirror_error in six.iteritems(exception.mirror_errors):
+        self.assertTrue(isinstance(mirror_error, tuf.exceptions.InvalidMetadataJSONError))
```
MVrachev, line 275: Now, when all of the validation is done in `_verify_metadata_file` another type of exception will
be thrown for `timestamp.json` and thus I had to change it here.

MVrachev:
While I was working on that, I thought that maybe it's not so logical to verify the metadata file by calling `_verify_metadata_file()` in `_get_metadata_file`.

`_get_metadata_file` is called in `_update_metadata` and I think it makes sense to call separately `_verify_metadata_file()` after calling `_get_metadata_file` inside the `_update_metadata` function.

I didn't change it here, because I wanted to hear an opinion on that first.

tuf/client/updater.py
```diff
@@ -1521,8 +1602,8 @@ def _get_metadata_file(self, metadata_role, remote_filename,
         downloaded.
 
       expected_version:
-        The expected and required version number of the 'metadata_role' file
-        downloaded.  'expected_version' is an integer.
+        An integer representing the expected and required version number
+        of the 'metadata_role' file downloaded.
```
joshuagl, line 1597: Or `None`?

tuf/client/updater.py
```diff
@@ -1412,9 +1412,81 @@ def _verify_root_self_signed(self, signable):
...
+    """
+
+    if expected_version is not None:
+      if version_downloaded != expected_version:
+        raise tuf.exceptions.BadVersionNumberError('Downloaded'
+          ' version number: ' + repr(version_downloaded) + '. Version'
+          ' number MUST be: ' + repr(expected_version))
+
+    # The caller does not know which version to download.
+    # Verify that the downloaded version is at least greater
```
joshuagl: Below we're checking that the current version is not less (<) than the current version, which contradicts this comment (which implies <=).

tuf/client/updater.py
```diff
@@ -1429,6 +1501,10 @@ def _verify_metadata_file(self, metadata_file_object,
         The role name of the metadata (e.g., 'root', 'targets',
         'unclaimed').
 
+      expected_version:
+        An integer representing the expected and required version number
+        of the 'metadata_role' file downloaded.
+
     <Exceptions>
```
jku, line 1489: Documented exceptions are unlikely to still be complete after these changes, please check

tests/test_endless_data_attack.py
jku: this is a sign of an API change -- maybe a good one but it should still be a conscious and documented decision. I assume it happens because the initial `load_json_string()` result is now caught when previously it was not?

Which one is buggy (old or new way) and why?

tuf/client/updater.py
```diff
@@ -1412,9 +1412,81 @@ def _verify_root_self_signed(self, signable):
...
+    # The caller does not know which version to download.
+    # Verify that the downloaded version is at least greater
+    # than the one locally available.
+    else:
+      try:
+        current_version = self.metadata['current'][metadata_role]['version']
+
+        if version_downloaded < current_version:
+          raise tuf.exceptions.ReplayedMetadataError(metadata_role,
+              version_downloaded, current_version)
```
jku, line 1418: This check (`version_downloaded < current_version`) is actually not consistent with the docstring or the comment above... what should really happen when `version_downloaded == current_version`?

I realize this inconsistency exists in 'develop' right now, but I'd like us to either fix the comment & docstring or file an issue while we do this.

jku (changes requested):
There's probably a debate here: do we want to do fixes like this now, does this make e.g. Teodoras work more difficult?

Personally I think there's value in this and the idea is correct: it makes the core update loop easier to read and makes us find issues in it. That said, the update loop feels brittle: let's make sure we do not break it... I left comments in code.

I also think we should try to fix the comments and logging when we touch them (make sure they're correct, useful and not too verbose) -- did not leave comments on those yet

MVrachev:
> There's probably a debate here: do we want to do fixes like this now, does this make e.g. Teodoras work more difficult?

It will be best if @sechkova could look into those changes and give an opinion as well.
I need those changes for https://github.com/theupdateframework/tuf/pull/1214.
I need to be able to abstract the spec and version verifications so that they could be reused when verifying older versions
of root.json.

> I also think we should try to fix the comments and logging when we touch them (make sure they're correct, useful and not too verbose) -- did not leave comments on those yet

Yes, a good reminder.
Will have a look at the comments again.
Please leave more comments if you want me to change a specific comment or line.

sechkova:
I am by no means against improving the code given that this is the currently used version and the rework is only just a dream ... yet.

jku:
> Please leave more comments if you want me to change a specific comment or line.

From memory: expected_version argument would have the typing hint "Optional[int]" but the possibility of None is not documented in several function docstrings. Both `_validate_*()` functions are quite long considering how little they do: comments and logging are so verbose.

tuf/client/updater.py
joshuagl: Let's fix the comment and docstring.

tests/test_endless_data_attack.py
MVrachev: I think the old one is buggy and not clear.
I catch it in `_verify_metadata_file` and instead throw a more concrete exception.
I will leave comments on both places: where I catch it and in the test.

tuf/client/updater.py
MVrachev: In the spec, it says:
```
The version number of the trusted timestamp metadata file, if any, MUST be less than or equal 
to the version number of the new timestamp metadata file. 
If the new timestamp metadata file is older than the trusted timestamp metadata file, discard it, 
abort the update cycle, and report the potential rollback attack.
```
The same applies to `Root` and `Snapshot`.

tuf/client/updater.py
MVrachev: In the spec, it says:
```
The version number of the trusted timestamp metadata file, if any, MUST be less than or equal 
to the version number of the new timestamp metadata file. 
If the new timestamp metadata file is older than the trusted timestamp metadata file, discard it, 
abort the update cycle, and report the potential rollback attack.
```
The same applies to `Root` and `Snapshot`.

So, I will change the above comment and docstring to  
```
Verify that the downloaded version is equal or 
greater than the locally available.

MVrachev:
Updated this pr according to @joshuagl and @jku comments.
Had to rebase because there were changes in `tuf/client/updater.py`.

MVrachev:
After a discussion with @joshuagl, I added the last commit to simplify some of the comments and the way we check for spec conformance.

PR #1214, opened 2020-11-17
Add optional feature to verify the chain of trust from bootstrap trusted root metadata to trusted root metadata

### LINKED ISSUE #1168: Updater feature request: verify chain of trust from bootstrapped root metadata

Context:
* In order to comply with the detailed client workflow a TUF client must ship a bootstrap trusted root metadata file out-of band (i.e. with the client installer).

Observations:
* On general purpose operating systems (Linux, Windows, macOS) this bootstrap root metadata will often be installed into operating system owned storage (i.e. the Python that comes with macOS, a distro installed Python on Linux, or Python from the Microsoft Store on Windows). 
* In the case of pip, and possibly other integrations, the trusted root metadata downloaded and stored during the detailed client workflow will often be persisted in user owned storage (i.e somewhere in a UNIX users home directory).
* In many cases the bootstrap root metadata will be afforded a much higher level or protection from tampering (SELinux/MAC, integrity verification with IMA/EVM, etc on some Linux distros. [System integrity protection](https://support.apple.com/en-gb/HT204899) for the default Python install on recent macOS.)

With these considerations in mind, it feels like providing a way to verify the chain of trust from the boostrap root metadata to the trusted root metadata will provide some additional integrity protection for the client, to better detect tampering with trusted root metadata that isn't stored in OS protected locations.

**Current behavior**:

No verification of chain of trust from bootstrap trusted root metadata to current trusted root metadata.

**Expected behavior**:

Expose an optional method on tuf.client.updater.Updater, to be called before `refresh()`, which will walk the chain of trust from the bootstrap root metadata to the current trusted root metadata. This method will perform a similar sequence of steps as defined in 5.1 of the detailed client workflow, only it will _prefer_ local copies of the intermediate trusted root metadata before reaching out to the repository to download intermediate root metadata file version.

MVrachev:
**NOTE:** This pr includes commits from https://github.com/theupdateframework/tuf/pull/1205

Fixes https://github.com/theupdateframework/tuf/issues/1168

Add an optional feature to establish a chain of trust between
a bootstrap root metadata file and the trusted current root
metadata.

If bootstrap_root_path is provided, the folder of the bootstrap root
will be expected to contain all intermediate root files between
the bootstrap root version and the current root version.

Also, if this feature is enabled, when updating a root.json file
with _update_metadata call (from updater.py) the root metadata files
marked as "previous" won't be deleted, but instead, they will be moved
to the provided bootstrap root folder.
That way after updating to a new version of the root metadata,
the chain of trust will be preserved.

New root.json metadata files had to be generated in order to
tests properly the establishment of a chain of trust between
a bootstrap root file and the current trustworthy root file.
The sequence `1.root.json`, `2.root.json` is generated by marking
the older one `X.root.json` as "dirty" and using `writeall()`
function from `repository_tool.py` and this will generate `X+1.root.json`.

**PS:** The diff is big because of the 5 new root.json files...

requirements**:

### DIFF
--- a/tests/repository_data/repository/metadata/root_sequence/1.root.json
+++ b/tests/repository_data/repository/metadata/root_sequence/1.root.json
@@ -1,0 +1,87 @@
+{
+ "signatures": [
+  {
+   "keyid": "4e777de0d275f9d28588dd9a1606cc748e548f9e22b6795b7cb3f63f98035fcb",
+   "sig": "a337d6375fedd2eabfcd6c2ef6c8a9c3bb85dc5a857715f6a6bd41123e7670c4972d8548bcd7248154f3d864bf25f1823af59d74c459f41ea09a02db057ca1245612ebbdb97e782c501dc3e094f7fa8aa1402b03c6ed0635f565e2a26f9f543a89237e15a2faf0c267e2b34c3c38f2a43a28ddcdaf8308a12ead8c6dc47d1b762de313e9ddda8cc5bc25aea1b69d0e5b9199ca02f5dda48c3bff615fd12a7136d00634b9abc6e75c3256106c4d6f12e6c43f6195071355b2857bbe377ce028619b58837696b805040ce144b393d50a472531f430fadfb68d3081b6a8b5e49337e328c9a0a3f11e80b0bc8eb2dc6e78d1451dd857e6e6e6363c3fd14c590aa95e083c9bfc77724d78af86eb7a7ef635eeddaa353030c79f66b3ba9ea11fab456cfe896a826fdfb50a43cd444f762821aada9bcd7b022c0ee85b8768f960343d5a1d3d76374cc0ac9e12a500de0bf5d48569e5398cadadadab045931c398e3bcb6cec88af2437ba91959f956079cbed159fed3938016e6c3b5e446131f81cc5981"
+  }
+ ],
+ "signed": {
+  "_type": "root",
+  "consistent_snapshot": false,
+  "expires": "2030-01-01T00:00:00Z",
+  "keys": {
+   "4e777de0d275f9d28588dd9a1606cc748e548f9e22b6795b7cb3f63f98035fcb": {
+    "keyid_hash_algorithms": [
+     "sha256",
+     "sha512"
+    ],
+    "keytype": "rsa",
+    "keyval": {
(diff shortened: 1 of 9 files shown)


MVrachev:
I haven't addressed edge cases like the possible situation where a user uses two versions of pip with two different bootstrap files. Read more in this comment https://github.com/theupdateframework/tuf/issues/1168#issuecomment-713546471.

Also, I decided that initially, I will focus on the root chain of trust, instead of making a more general function that could be used with snapshot too. I wanted to make sure we catch all of the edge cases here, before making something more general.
Read more here: https://github.com/theupdateframework/tuf/issues/1168#issuecomment-713546471 and here https://github.com/theupdateframework/tuf/issues/1168#issuecomment-714301736.

jku:
I think this requires a bit more design work or at least discussion -- either here or maybe rather in the issue. As an end result I'd like to see the core algorithm written as text, with explanation and other options discussed (e.g. why are we loading current runtime metadata first, and only then looking at the bootstrap if we think the bootstrap is more trustworthy?) and the variables at play as well  (what should happen if root chain is not complete locally, what if bootstrap root is newer than current runtime root, what if bootstrap and runtime root are just different).

Also might be useful to consider other (non-root) bootstrap metadata already (even if the intention is not to implement that) -- even if just to clearly state that this approach is / is not viable for the other metadata types. The related TAP might be useful to refer to here.

My comments might be a bit all over the place, hope you can find the threads...

> If bootstrap_root_path is provided, the folder of the bootstrap root
> will be expected to contain all intermediate root files between
> the bootstrap root version and the current root version.

Is this a reasonable requirement? What happens if it does not?

> Also, if this feature is enabled, when updating a root.json file
> with _update_metadata call (from updater.py) the root metadata files
> marked as "previous" won't be deleted, but instead, they will be moved
> to the provided bootstrap root folder.

This is not possible:  bootstrap location may be on a read-only filesystem (or not writable by current user) -- the core idea is that bootstrap may be more trustworthy because user cannot modify it.

I don't see a reason not to store the "old" root metadata files somewhere in the normal runtime metadata directory: please explain if you've noticed one. 

The question of whether bootstrap checking should modify runtime metadata (e.g. should it store the bootstrap metadata itself in runtime metadata if it's not there already, should it download new files needed to chain up to current runtime metadata root) needs to be decided... I would think downloading and storing the missing files is appropriate but this probably needs some thinking and playing out the different scenarios -- what if I run pip X that has bootstrap A, and then run pip Y that has bootstrap B, with various values of A and B?

I think I'd like to see a variant of the process that starts the local metadata verification process from the bootstrap root, and continues using the files in the runtime metadata (as a design to compare with this one) -- maybe it even makes sense to put this all in the existing root metadata update loop?

> I haven't addressed edge cases like the possible situation where a user uses two versions of pip with two different bootstrap files

The comments above probably make this obvious but: I don't think that is an edge case. Even just upgrading the same pip could mean new, unexpected bootstrap root

MVrachev:
Good points @jku!
This could be considered as a prototype.
I had a brainstorming session with @joshuagl where I wrote down the assumptions I am making in the current proposal
and what and why can be approved.
The steps I am going to take are:
1. Familiarize with the proposal made in this TAP: https://github.com/theupdateframework/taps/pull/128
2. Address @jku comments and change this pr.

MVrachev:
On hold, until this [TAP](https://github.com/theupdateframework/taps/pull/128) is clarified to a point where I would be able to make this pr a prototype of the TAP.

jku:
> On hold, until this [TAP](https://github.com/theupdateframework/taps/pull/128) is clarified to a point where I would be able to make this pr a prototype of the TAP.

I'm closing this: no point in keeping PRs cluttering the list if they're not going anywhere. I don't think we'll want to touch the legacy updater anymore anyway.

Please re-open if needed.

PR #1216, opened 2020-11-19
Fix imports and default value for function arg

MVrachev:
This quote from the Google Python style guide made me realize
why empty list as a default value for an argument could be
dangerous:

`Default arguments are evaluated once at module load time.
This may cause problems if the argument is a mutable object
such as a list or a dictionary. If the function modifies the object
(e.g., by appending an item to a list), the default value is modified.`

Read more here:
https://google.github.io/styleguide/pyguide.html#2123-cons

Also, currently, we are importing the "utils" module in tests/utils
with "import utils".
This could become a problem when there is another module with
the same general name "utils" and could lead to import mistakes.

requirements**:

### DIFF
--- a/tests/test_arbitrary_package_attack.py
+++ b/tests/test_arbitrary_package_attack.py
@@ -52,7 +52,7 @@ import tuf.log
 import tuf.client.updater as updater
 import tuf.unittest_toolbox as unittest_toolbox
 
-import utils
+from tests import utils
 
 import securesystemslib
 import six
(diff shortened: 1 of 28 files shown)


tests/test_arbitrary_package_attack.py
```diff
@@ -52,7 +52,7 @@
 import tuf.client.updater as updater
 import tuf.unittest_toolbox as unittest_toolbox
 
-import utils
+import tests.utils as utils
```
jku:
```suggestion
from tests import utils
```
should be enough

jku (approved):
the default value is only dangerous if the function actually modified the value -- but as defensive and consistent programming  these both look like solid changes to me.

joshuagl (approved):
These look like good improvements, particularly the empty list as default argument fix. Thanks @MVrachev!

PR #1217, opened 2020-11-19
Make metadata signatures ordered by keyid

### LINKED ISSUE #1154: repository_lib.py lists signatures in an non-deterministic order

When some metadata is signed with multiple keys, `tuf/repository_lib.py` doesn't write the signatures in a deterministic order. This occurs because of this [line](https://github.com/theupdateframework/tuf/blob/develop/tuf/repository_lib.py#L121). If this line was changed to `signing_keyids = sorted(set(roleinfo['signing_keyids']))`. I hacked this in and it worked, but I wasn't sure how to write a test for it, and ran out of time.

 doesn't sign the metadata in a consistent order, and so it's not guaranteed to generate the same file if metadata is resigned with the same keys. This came up in the [go-tuf interoperability test suite](https://github.com/theupdateframework/go-tuf/pull/122), where each time I regenerated the metadata I might get different file hashes.

**Current behavior**:

It's not guaranteed what order signatures are written in.

**Expected behavior**:

Signatures to be sorted in some order. My proposal is we just do alphabetical order.

### LINKED ISSUE #1211: Non-determinism in JSON export

I've been working to create a deterministic test fixture generator for PHP-TUF. I've rooted out the apparent sources of most meaningful non-determinism by fixing the clock and using a fixed well of keypairs. However, some of the JSON export appears to have different behavior on different systems.

Shown below is the diff I see when comparing generated data on GitHub Actions (on Python 3.9 with `ubuntu-latest`) versus on my laptop (also Python 3.9 but with Fedora 33). We've pinned all known dependencies using `pipenv`, so I don't think it's that.

This causes a cascading set of differences because other files use hashes of `snapshot.json`.

Could TUF canonicalize even the JSON data that isn't directly signed?

**Current behavior**:
Screenshot from 2020-11-12 12-30-35

**Expected behavior**:

Deterministic (ideally canonical) output of JSON that contains the same functional data.

lukpueh:
Fixes #1154 and parts of #1211.

In `repository_lib._generate_and_write_metadata` sort the set of signing key keyids alphabetically before passing them on to signing functions, to make the order in which signatures are added deterministic. This is above all beneficial for testing.

This commit also adds an exemplary test for signatures on root metadata using the `repository_tool` interface to setup all the state that is required to test _generate_and_write_metadata.

Kudos to @erickt for proposing this fix, and to him and @davidstrauss for bringing the issue to our attention.

requirements**:

### DIFF
--- a/tests/test_repository_tool.py
+++ b/tests/test_repository_tool.py
@@ -716,6 +716,34 @@ class TestRepository(unittest.TestCase):
       tuf.formats.check_signable_object_format(role_signable)
 
 
+  def test_signature_order(self):
+    """Test signatures are added to metadata in alphabetical order. """
+    # Create empty repo dir and init default repo in memory
+    repo_dir = tempfile.mkdtemp(dir=self.temporary_directory)
+    repo = repo_tool.create_new_repository(repo_dir)
+
+    # Dedicate any two existing test keys as root signing keys
+    for key_name in ["targets_key", "snapshot_key"]:
+      repo.root.load_signing_key(
+          repo_tool.import_ed25519_privatekey_from_file(
+              os.path.join("repository_data", "keystore", key_name),
+              "password"))
+
+    # Write root metadata with two signatures
+    repo.write("root")
+
+    # Load signed and written json metadata back into memory
+    root_metadata_path = os.path.join(
+        repo_dir, repo_tool.METADATA_STAGED_DIRECTORY_NAME, "root.json")
+    root_metadata = securesystemslib.util.load_json_file(root_metadata_path)
+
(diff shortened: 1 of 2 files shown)


mnm678 (approved):
This looks good, but IIUC the metadata may still be non-deterministic as the body is not canonical json in transit.

jku (approved):

lukpueh:
Thanks for the reviews, @jku and @mnm678!  And yes, @mnm678, metadata may still be non-deterministic due to JSON not being deterministic. This PR is only about the order of signatures. I'll only close #1154 and leave #1211 open.

PR #1218, opened 2020-11-23
Fix root self verification to only count a keyid once towards the threshold

joshuagl:
Fixes #N/A

@trishankatdatadog perceptively observed that `_verify_root_self_signed()`, introduced in #1101, will incorrectly count multiple signatures with the same keyid towards the threshold for new signatures listed inside an updated root metadata file.

This PR introduces:
1. a test to catch this issue, by having a root metadata file with a threshold of two list the same keyid:sig dict twice in the metadata file's signatures.
2. a fix to `_verify_root_self_signed()` to uniquify a list of keyids for which signatures have been verified, and compare only the number of unique keyids against the threshold.

requirements**:

### DIFF
--- a/tests/test_updater_root_rotation_integration.py
+++ b/tests/test_updater_root_rotation_integration.py
@@ -262,6 +262,63 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
 
 
+  def test_verify_root_with_duplicate_current_keyids(self):
+     """
+     Each root file is signed by the current root threshold of keys as well
+     as the previous root threshold of keys. In each case, a keyid must only
+     count once towards the threshold. Test that the new root signatures
+     specific signature verification implemented in _verify_root_self_signed()
+     only counts one signature per keyid towards the threshold.
+     """
+     # Load repository with root.json == 1.root.json (available on client)
+     # Signing key: "root", Threshold: 1
+     repository = repo_tool.load_repository(self.repository_directory)
+
+     # Add an additional signing key and bump the threshold to 2
+     repository.root.load_signing_key(self.role_keys['root']['private'])
+     repository.root.add_verification_key(self.role_keys['root2']['public'])
+     repository.root.load_signing_key(self.role_keys['root2']['private'])
+     repository.root.threshold = 2
+     repository.writeall()
+
+     # Move staged metadata to "live" metadata
+     shutil.rmtree(os.path.join(self.repository_directory, 'metadata'))
+     shutil.copytree(os.path.join(self.repository_directory, 'metadata.staged'),
+         os.path.join(self.repository_directory, 'metadata'))
+
+     # Modify 2.root.json and list two signatures with the same keyid
+     root2_path_live = os.path.join(
+         self.repository_directory, 'metadata', '2.root.json')
+     root2 = securesystemslib.util.load_json_file(root2_path_live)
+
+     signatures = []
+     signatures.append(root2['signatures'][0])
+     signatures.append(root2['signatures'][0])
+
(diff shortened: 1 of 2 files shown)


tuf/client/updater.py
```diff
@@ -1385,7 +1385,7 @@ def _verify_root_self_signed(self, signable):
     signatures = signable['signatures']
     signed = securesystemslib.formats.encode_canonical(
         signable['signed']).encode('utf-8')
-    validated = 0
+    verified_sig_keyids = []
```
trishankatdatadog: Might as well just use a set() from the outset, no?

trishankatdatadog (changes requested):

tuf/client/updater.py
jku: That's still a list. Proper empty set requires `verified_sig_keyids = set()` (and then you need to use use `verified_sig_keyids.add(keyid)`, but don't need to transform it before getting the length).

Either way works for me, maybe it's cleaner to be an actual set to begin with -- this is also what `sig.verify()` does.

tuf/client/updater.py
```diff
@@ -1403,9 +1403,12 @@ def _verify_root_self_signed(self, signable):
       valid_sig = securesystemslib.keys.verify_signature(key, signature, signed)
 
       if valid_sig:
-        validated = validated + 1
+        verified_sig_keyids.append(keyid)
 
-    if validated >= threshold:
+    # A signature with a given keyid should only count towards the threshold
+    # once. Convert the list of keyids into a set, in order to uniquify them
+    # and therefore count each keyid only once towards the threshold.
```
jku:
```suggestion
    # Each unique signature/keyid counts only once towards the threshold
```
would this be enough?

tuf/client/updater.py
joshuagl: Quite right, thanks both. Pushed a fixup to change this to a set in 44d48af

tuf/client/updater.py
joshuagl: That's much more concise while remaining informative, thanks. I ended up removing the comment entirely when switching to using a set as suggested above.

tuf/client/updater.py
jku: oh and if my comment sounded weird that was because I misread the trishanks comment to be a github 'suggestion'. Didn't mean to lecture :)

jku (approved):

tests/test_updater_root_rotation_integration.py
```diff
@@ -262,6 +262,63 @@ def test_verify_root_with_current_keyids_and_threshold(self):
...
+
+     signatures = []
+     signatures.append(root2['signatures'][0])
+     signatures.append(root2['signatures'][0])
+
+     root2['signatures'] = signatures
+
+     roo2_fobj = tempfile.TemporaryFile()
+     roo2_fobj.write(tuf.repository_lib._get_written_metadata(root2))
+     securesystemslib.util.persist_temp_file(roo2_fobj, root2_path_live)
```
lukpueh: Given that you have to rebase one more time anyway (see DCO) I suggest to address this minor nit while doing so:
```suggestion
     root2_fobj = tempfile.TemporaryFile()
     root2_fobj.write(tuf.repository_lib._get_written_metadata(root2))
     securesystemslib.util.persist_temp_file(root2_fobj, root2_path_live)
```

lukpueh (approved):
Thanks for the quick fix and the test, @joshuagl! Please add DCO and we are good to go.


Also, kudos to @trishankatdatadog, for bringing this to our attention.

PR #1219, opened 2020-11-23
Avoid reading target in memory

### LINKED ISSUE #1215: Updater: target hash calculation should not read the whole file in memory

This is not a bug (although it could be on memory limited client device), but a performance improvement:

After PR #1202 there is only one place where the updater loads the whole target file in memory. We should avoid doing that as targets could be very large and memory could be limited.

**Current behavior**:

`_check_hashes()` does this:
```python
digest_object = securesystemslib.hash.digest(algorithm)
digest_object.update(file_object.read())
computed_hash = digest_object.hexdigest()
```

**Expected behavior**:
something like *handwaves*
```python
digest_object = securesystemslib.hash.digest(algorithm)
while True:
    chunk = file_object.read(CHUNK_SIZE)
    if not chunk:
        break
    digest_object.update(chunk)
computed_hash = digest_object.hexdigest()
```
or even more simply just let SSLib handle this with its default chunk size:

```python
digest_object = securesystemslib.hash.digest_fileobject(file_object, algorithm)
computed_hash = digest_object.hexdigest()
```

jku:
Fixes #1215

Updater: Avoid reading whole target file in memory

We don't want to read the whole file in memory as it can be huge. Use `securesystemslib.hash.digest_fileobject()` instead: This way Securesystemslib will read the file in chunks.

Also make the comments and logging in `Updater._check_hashes()` more to the point.

### DIFF
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1196,12 +1196,8 @@ class Updater(object):
     # Verify each trusted hash of 'trusted_hashes'.  If all are valid, simply
     # return.
     for algorithm, trusted_hash in six.iteritems(trusted_hashes):
-      digest_object = securesystemslib.hash.digest(algorithm)
-      # Ensure we read from the beginning of the file object
-      # TODO: should we store file position (before the loop) and reset after we
-      # seek about?
-      file_object.seek(0)
-      digest_object.update(file_object.read())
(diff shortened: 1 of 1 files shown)


mnm678 (approved):

joshuagl (approved):
Lovely set of cleanups, thanks @jku !

PR #1220, opened 2020-11-24
ADR0003: where to develop TUF 1.0.0

### LINKED ISSUE #1126: Decide where to develop TUF 1.0.0
The current plan is to implement TUF 1.0.0 alongside the current code base, in order to not disrupt existing  usage and keep providing a Python 2.7 client (see #1125). We need to decide where to best develop the new TUF.  A few ideas:

- In its own repo
  Con: Given limited engineering power, the old repo is prone to bit rot while we work on the other.

- In a separate develop branch of the current tuf implementation
  Con: Similar to above + git mental overhead

- In a subdirectory of the current tuf implementation (which we already do see https://github.com/theupdateframework/tuf/pull/1112)
 Con: Messiness in plain sight.

sechkova:
Fixes #1126 

Document the outcome of #1126: develop TUF 1.0.0 in a subdirectory of the current TUF implementation.

requirements**:

### DIFF
--- a/docs/adr/0003-where-to-develop-TUF-1-0-0.md
+++ b/docs/adr/0003-where-to-develop-TUF-1-0-0.md
@@ -1,0 +1,46 @@
+# Develop TUF 1.0.0 in a subdirectory of the current TUF implementation
+
+* Status: accepted
+* Date: 2020-11-23
+
+Technical Story: https://github.com/theupdateframework/tuf/issues/1126
+
+## Context and Problem Statement
+
+The plan is to implement a refactored TUF (1.0.0) alongside the current
+code base, in order to not disrupt existing usage and keep providing
+a Python 2.7 client.
+
+We need to decide on the best place to do this development.
+
+## Decision Drivers
+
+* Developing the new code piecemeal
+* Continuing to make releases in the interim
+* Avoiding maintenance overhead
+
+## Considered Options
+
+Develop TUF 1.0.0:
+
+* In its own repository
+* In a separate development branch of the current TUF implementation
+* In the default branch, archiving the current implementation
+* In a subdirectory of the current TUF implementation
+
+## Decision Outcome
+
+Chosen option: "Develop TUF 1.0.0 in a subdirectory of the current TUF
+implementation", because we want to add the new TUF code gradually
+while keep maintaining the current implementation given limited
+maintenance resources.
+
+### Negative Consequences
+
+* In progress development in the default branch causes messiness
+  in plain sight.
+
+## Links
+
+* [Discussion of Python version support in TUF 1.0.0](https://github.com/theupdateframework/tuf/issues/1125)
+* [Discussion of deprecation policy for the pre-1.0, Python 2.7 supporting, code](https://github.com/theupdateframework/tuf/issues/1127)
--- a/docs/adr/0003-where-to-develop-TUF-1-0-0.md
+++ b/docs/adr/0003-where-to-develop-TUF-1-0-0.md
@@ -35,6 +35,18 @@ implementation", because we want to add the new TUF code gradually
 while keep maintaining the current implementation given limited
 maintenance resources.
 
+Once development of the new version is complete, we will transition
+from TUF 1.0.0 in a subdirectory to stand-alone TUF 1.0.0 by the following
+procedure:
+
+* flesh out tuf/api/*
+* implement tuf/client/new-updater.py
+* implement tuf/repository/*
+* \<iterate\>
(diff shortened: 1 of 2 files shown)


docs/adr/0003-where-to-develop-TUF-1-0-0.md
```diff
@@ -0,0 +1,46 @@
...
+* In a separate development branch of the current TUF implementation
+* In the default branch, archiving the current implementation
+* In a subdirectory of the current TUF implementation
+
+## Decision Outcome
+
+Chosen option: "Develop TUF 1.0.0 in a subdirectory of the current TUF
+implementation", because we want to add the new TUF code gradually
+while keep maintaining the current implementation given limited
+maintenance resources.
```
lukpueh, line 36: Minor nit: Should we somehow underline that this option seems to have the least maintenance overhead compared to option 1 and 2, while allowing us to continue making releases with the old code unlike option 3?

lukpueh (approved):
Thanks for the PR, @sechkova. I have one minor nit (see inline) but other than that it looks solid.

lukpueh:
Not sure if this should be part of AD0003, or should have been part of [ADR0002](https://github.com/theupdateframework/tuf/blob/develop/docs/adr/0002-pre-1-0-deprecation-strategy.md), or neither, but we also need to decide how to transition from TUF 1.0.0 in a subdirectory to stand-alone TUF 1.0.0. Im fine if we cross that bridge when we get there (e.g. with the help of `git filter-branch`-sorcery). cc @joshuagl

trishankatdatadog:
I'm ok with developing in a separate dir in the current default branch, as long as we're not committed to legacy code. Should we clarify this in the ADR, if we all agree?

joshuagl:
> Not sure if this should be part of AD0003, or should have been part of [ADR0002](https://github.com/theupdateframework/tuf/blob/develop/docs/adr/0002-pre-1-0-deprecation-strategy.md), or neither, but we also need to decide how to transition from TUF 1.0.0 in a subdirectory to stand-alone TUF 1.0.0. Im fine if we cross that bridge when we get there (e.g. with the help of `git filter-branch`-sorcery). cc @joshuagl

The way I've been imaginging it, filter-branch wizardry would not be required. In my mind we'd have something like:
* flesh out tuf/api/*
* implement tuf/client/new-updater.py
* implement tuf/repository/*
* \<iterate\> 💻 
* git mv tuf/client/new-updater.py tuf/client/updater.py
* git rm tuf/*.py
* tag 1.0.0

joshuagl (approved):
Thanks for writing this up @sechkova !

joshuagl:
> I'm ok with developing in a separate dir in the current default branch, as long as we're not committed to legacy code. Should we clarify this in the ADR, if we all agree?

I believe we all agree. I think it's implicit in ADR 0002 that we are not committed to legacy code?

joshuagl:
Per https://github.com/theupdateframework/tuf/pull/1203#pullrequestreview-535266482 and the discussion in #1127 we still need to document:

> a deprecation/API stability strategy for releases after the refactor

Perhaps we can ensure we are not committed to maintaining legacy code with that document?

lukpueh:
> The way I've been imaginging it, filter-branch wizardry would not be required. In my mind we'd have something like:
> 
> * flesh out tuf/api/*
> * implement tuf/client/new-updater.py
> * implement tuf/repository/*
> * <iterate> 💻
> * git mv tuf/client/new-updater.py tuf/client/updater.py
> * git rm tuf/*.py
> * tag 1.0.0

Sounds like a good plan and a good fit for an addendum to ADR0003?

joshuagl:
> > The way I've been imaginging it, filter-branch wizardry would not be required. In my mind we'd have something like:
> > 
> > * flesh out tuf/api/*
> > * implement tuf/client/new-updater.py
> > * implement tuf/repository/*
> > * 💻
> > * git mv tuf/client/new-updater.py tuf/client/updater.py
> > * git rm tuf/*.py
> > * tag 1.0.0
> 
> Sounds like a good plan and a good fit for an addendum to ADR0003?

Agreed. Would you mind outlining the above in ADR0003 @sechkova ?

sechkova:
Updated with your suggestions.

lukpueh:
> Updated with your suggestions.

Thanks! Could you please also resolve the conflict in `docs/adr/index.md`?

sechkova:
> > Updated with your suggestions.
> 
> Thanks! Could you please also resolve the conflict in `docs/adr/index.md`?

oops .. I forgot .. here it is

lukpueh:
picobello!

Issue #1161, opened 2020-10-05
Modernise and rationalise packaging


joshuagl:
**Description of issue or feature request**:

Python packaging has changed significantly since the tuf packaging was created. There's been a move to declarative files, the formalisation of build system independent source tree layout in [PEP 517](https://www.python.org/dev/peps/pep-0517/), and the addition of new metadata to specify the build system dependencies in [PEP 518](https://www.python.org/dev/peps/pep-0518/).

Furthermore our current packaging is somewhat ad-hoc, with some expected files like LICENSE missing from the sdist (see #1160) and some unexpected files like tuf/ATTACKS.md included.

For reference see:
https://packaging.python.org/
https://snarky.ca/what-the-heck-is-pyproject-toml/

**Current behavior**:

Packaging isn't up-to-date with PyPA recommendations and our sdist includes files which don't match conventions.

**Expected behavior**:

Packaging is in line with current PyPA best practices and generated distributions match ecosystem conventions.

joshuagl:
We should add some CI steps to ensure our packaging metadata is sane, for example our current `MANIFEST.in` fails a [check-manifest](https://github.com/mgedmin/check-manifest) test with:

> lists of files in version control and sdist do not match!

jku:
A gotcha that we may care about with regards to setup.py/setup.cfg/pyproject.toml: I just tried to do `pip install -e <directory>` to a project that only has a pyproject.toml and:
```
pip install -e ../vendoring
ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /home/jku/src/vendoring
(A "pyproject.toml" file was found, but editable mode currently requires a setup.py based build.)
```
Which would mean current requirements-dev.txt in tuf would fail to install in this case.

sechkova:
> A gotcha that we may care about with regards to setup.py/setup.cfg/pyproject.toml: I just tried to do `pip install -e <directory>` to a project that only has a pyproject.toml and:
> 
> ```
> pip install -e ../vendoring
> ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /home/jku/src/vendoring
> (A "pyproject.toml" file was found, but editable mode currently requires a setup.py based build.)
> ```
> 
> Which would mean current requirements-dev.txt in tuf would fail to install in this case.

Yep, it is a known limitation, it can be circumvented by:

```
import setuptools

if __name__ == "__main__":
    setuptools.setup()
```

[...]

PR #1221, opened 2020-11-24
[WIP] tuf-client as a standalone installation

### LINKED ISSUE #1181: Identify components of client and determine how to split out client from current code base

For both the pip integration, and in order to provide some semblance of support for Python 2.7 users when we proceed with the refactor (#1127), it would be very useful to have the client be installable by itself __without__ any of the modules which are only used for a repository.

We should identify what pieces of the current code are required by the client. Can we create a branch with only the client and its dependencies? Can we adapt the current client tests to work without `tuf.repository_lib` and `tuf.repository_tool`?

The tuf updater requires at least the following modules:
* tuf.client.updater
* tuf.download
* tuf.mirrors
* tuf.sig
* tuf.formats
* tuf.settings
* tuf.keydb
* tuf.roledb
* tuf.exceptions

**Current behavior**:

Client and repository code is installed as one big package

**Expected behavior**:

Possible to install only the client, without any modules used only by a repository

sechkova:
Fixes #1181

This is a WIP  trying to identify whether splitting the client and repository code and maintaining a client as
a separate installation (in a client branch for now) will be useful for client side users of tuf (e.g. pip) as well as for providing some kind of support for older versions.

Up to now I have only removed the pure repository code and generated a tuf-client distribution.
This only reduces the lines of code while the dependencies remain the same, since client and repository share all dependencies (see [comment](https://github.com/theupdateframework/tuf/pull/1221#issuecomment-733033354)). Client even upgrades them with requests.

I added one dummy test simply calling updater.refresh without using the repository code.

Please give me your opinion on:
- Does this even make sense?
- Did I get the idea right?
- Is this bringing a significant value given the effort to rewrite the test suite?
- If all of the above is yes, should we continue with fixing all tests and the configuration?

requirements**:


sechkova:
In case I have successfully regenerated `requirements-pinned.txt ` following the guidelines in r`equirements.txt`, the resulting diff is only in version numbers:

```
diff --git a/requirements-pinned.txt b/requirements-pinned.txt
index 848e99a8..48e73a38 100644
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -1,14 +1,14 @@
-certifi==2020.6.20       # via requests
+certifi==2020.11.8        # via requests
 cffi==1.14.3              # via cryptography, pynacl
 chardet==3.0.4            # via requests
-cryptography==3.2.1         # via securesystemslib
-enum34==1.1.6             ; python_version < '3' # via cryptography
-idna==2.10                 # via requests
-ipaddress==1.0.23         ; python_version < '3' # via cryptography
+cryptography==3.2.1       # via securesystemslib
+enum34==1.1.10            # via cryptography
+idna==2.10                # via requests
+ipaddress==1.0.23         # via cryptography
 pycparser==2.20           # via cffi
 pynacl==1.4.0             # via securesystemslib
-requests==2.25.0
-securesystemslib[crypto,pynacl]==0.18.0
-six==1.15.0
-subprocess32==3.5.4       ; python_version < '3' # via securesystemslib
+requests==2.25.0          # via -r requirements.txt
+securesystemslib[crypto,pynacl]==0.18.0  # via -r requirements.txt
+six==1.15.0               # via -r requirements.txt, cryptography, pynacl, securesystemslib
+subprocess32==3.5.4       # via securesystemslib
 urllib3==1.26.2           # via requests

```

setup.py
```diff
@@ -124,7 +124,6 @@
   ],
   packages = find_packages(exclude=['tests']),
   scripts = [
-    'tuf/scripts/repo.py',
     'tuf/scripts/client.py'
```
joshuagl, line 127: Might as well scrap this, too. It's not useful for clients using `updater.py`

tests/test_updater_no_repo.py
```diff
@@ -0,0 +1,254 @@
...
+
+        # Creating a repository instance.  The test cases will use this client
+        # updater to refresh metadata, fetch target files, etc.
+        self.repository_updater = updater.Updater(self.repository_name,
+                                                self.repository_mirrors)    
+
+        # Metadata role keys are needed by the test cases to make changes to the
+        # repository (e.g., adding a new target file to 'targets.json' and then
+        # requesting a refresh()).
+        self.role_keys = _load_role_keys(self.keystore_directory)
```
joshuagl, line 178: I don't think we want/need this if we're not using repository_tool?

joshuagl (commented):
Thank you for taking a stab at this @sechkova! 🎉 

My motivation for filing #1181 was the idea that an effort such as this might be useful:
a) for vendoring tuf into pip, to reduce the number of files and LOC added to that project (see pypa/pip#9041)
b) for theoretically providing an as small as possible base for interested parties to pick up if they want/need to continue to use tuf/client/updater.py on a version of the interpreter that will no longer be supported following the [refactor towards 1.0](https://github.com/theupdateframework/tuf/projects/2) (i.e. 2.7)

I'll leave others to comment on whether they feel this work is worthwhile cc @lukpueh @trishankatdatadog @jku

sechkova:
Change of heart!
@jku gave us convincing arguments about grouping repository/client/common code in a proper directory structure and packaging two separate repository/client distributions all in the current development branch.

I think this suggestion achieves the goals listed by Joshua:
> My motivation for filing #1181 was the idea that an effort such as this might be useful:
> a) for vendoring tuf into pip, to reduce the number of files and LOC added to that project (see [pypa/pip#9041](https://github.com/pypa/pip/pull/9041))
> b) for theoretically providing an as small as possible base for interested parties to pick up if they want/need to continue to use tuf/client/updater.py on a version of the interpreter that will no longer be supported following the [refactor towards 1.0](https://github.com/theupdateframework/tuf/projects/2) (i.e. 2.7)
> 

while in addition:
- no need to rewrite a big part of the test cases
- no need to maintain common code in two places

As a drawback, one repository will host configuration for two packages. We can use the occasion to experiment with "modern packaging" #1161, however.

sechkova:
In case there are no objections on the proposal in the last comment, I will close this PR and reopen another one with the suggested approach.

jku:
Closing this based on last comment :)

Issue #1140, opened 2020-09-10
Add input validation to simple metadata api


lukpueh:
Coordinate with validation guidelines https://github.com/theupdateframework/tuf/issues/1130

**Description of issue or feature request**:
Some suggestions:
   - Avoid schema, see secure-systems-lab/securesystemslib#183 (just don't use it)
   - Make use of type hints for simple type validation
   - Perform additional non-metadata parameter validation at user boundary
   - Provide methods to validate JSON representation at user boundary, i.e. fail on bad json metadata in `from_json_file`/`to_json_file` method, but with option to
     disable check as there might be a justified reason to read or write WIP
     metadata to json.
   - Be lenient on bad/invalid metadata objects in memory, they might be
     work in progress. E.g. it might be convenient to create empty metadata
     and assign attributes later on.
   - Consider using in-toto style `ValidationMixin` (see the [`mixin`](https://github.com/in-toto/in-toto/blob/74da7a/in_toto/models/common.py#L27-L40) and it's [usage](https://github.com/in-toto/in-toto/blob/74da7a/in_toto/models/layout.py#L420-L438) for details).

**Current behavior**:
No input validation

**Expected behavior**:
Add input validation

joshuagl:
> * Consider using in-toto style `ValidationMixin` (see the [`mixin`](https://github.com/in-toto/in-toto/blob/74da7a/in_toto/models/common.py#L27-L40) and it's [usage](https://github.com/in-toto/in-toto/blob/74da7a/in_toto/models/layout.py#L420-L438) for details).

This looks like a nice pattern. Each class defines appropriate validation methods, which are called by the Mixin's `validate()` method at the end of the object's construction. Any thoughts on whether a decorator might be a more idiomatic approach than a Mixin? We can

This article suggests how to to type validation with type hints https://aboutsimon.com/blog/2018/04/04/Python3-Type-Checking-And-Data-Validation-With-Type-Hints.html

joshuagl:
> * Consider using in-toto style `ValidationMixin` (see the [`mixin`](https://github.com/in-toto/in-toto/blob/74da7a/in_toto/models/common.py#L27-L40) and it's [usage](https://github.com/in-toto/in-toto/blob/74da7a/in_toto/models/layout.py#L420-L438) for details).

This looks like a nice pattern. Each class defines appropriate validation methods, which are called by the Mixin's `validate()` method at the end of the object's construction. 

[...]

PR #1223, opened 2020-11-25
New metadata API: add MetadataInfo and TargetFile classes

MVrachev:
This pr addresses https://github.com/theupdateframework/tuf/issues/1139, but doesn't fix it yet.

I will add a separate pr which will handle `Roles` and `Keys` classes.

It includes the following changes:
- Add a missing root use case in a couple of our tests
- Make `length` and `hashes` optional in `Timestamp` as it is in our specification
- Add `MetaFile` class and use it in the `Timestamp` and `Snapshot` classes
- Add `TargetFile` class and use it in the `Targets` class
- Rename a few instances of the `update` method in our classes and use a more verbose name.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -570,6 +570,58 @@ class Snapshot(Signed):
         self.meta[metadata_fn] = MetadataInfo(version, length, hashes)
 
 
+class TargetInfo:
+    """A container with information about a particular target file.
+    Instances of TargetInfo are used as values in a dictionary
+    called "targets" in Targets.
+
+    Attributes:
+        length: An integer indicating the length of the target file.
+        hashes: A dictionary containing hash algorithms and the
+            hashes resulted from applying them over the target file::
+
+                'hashes': {
+                    '<HASH ALGO 1>': '<TARGET FILE HASH 1>',
+                    '<HASH ALGO 2>': '<TARGET FILE HASH 2>',
+                    ...
+                }
+
+        custom: An optional dictionary which may include version numbers,
+            dependencies, or any other data that the application wants
+            to include to describe the target file::
+
+                'custom': {
+                    'type': 'metadata',
+                    'file_permissions': '0644',
+                    ...
+                } // optional
+
+    """
+
+    def __init__(self, length: int, hashes: JsonDict,
+            custom: Optional[JsonDict] = None) -> None:
+        self.length = length
+        self.hashes = hashes
+        self.custom = custom
+
+
+    def __eq__(self, other: 'TargetInfo') -> bool:
+        """Compare objects by their values instead of by their addresses."""
+        return (self.length == other.length and
+                self.hashes == other.hashes and
+                self.custom == other.custom)
+
+
+    def to_dict(self) -> JsonDict:
+        """Returns the JSON-serializable dictionary representation of self. """
+        json_dict = {'length': self.length, 'hashes': self.hashes}
+
+        if self.custom is not None:
+            json_dict['custom'] = self.custom
+
+        return json_dict
+
+
 class Targets(Signed):
     """A container for the signed part of targets metadata.
 
@@ -577,15 +629,7 @@ class Targets(Signed):
         targets: A dictionary that contains information about target files::
 
             {
-                '<TARGET FILE NAME>': {
-                    'length': <TARGET FILE SIZE>,
-                    'hashes': {
-                        '<HASH ALGO 1>': '<TARGET FILE HASH 1>',
-                        '<HASH ALGO 2>': '<TARGETS FILE HASH 2>',
-                        ...
-                    },
-                    'custom': <CUSTOM OPAQUE DICT> // optional
-                },
+                '<TARGET FILE NAME>': <TargetInfo INSTANCE>,
                 ...
             }
 
@@ -629,20 +673,35 @@ class Targets(Signed):
     # pylint: disable=too-many-arguments
     def __init__(
             self, _type: str, version: int, spec_version: str,
-            expires: datetime, targets: JsonDict, delegations: JsonDict
-            ) -> None:
+            expires: datetime, targets: Dict[str, TargetInfo],
+            delegations: JsonDict) -> None:
         super().__init__(_type, version, spec_version, expires)
-        # TODO: Add class for meta
         self.targets = targets
+
+        # TODO: Add Key and Role classes
         self.delegations = delegations
+
+
+    @classmethod
+    def from_dict(cls, signed_dict: JsonDict) -> 'Targets':
+        """Creates Targets object from its JSON/dict representation. """
+        for target_path in signed_dict['targets'].keys():
+            signed_dict['targets'][target_path] = TargetInfo(
+                    **signed_dict['targets'][target_path])
+
+        return super().from_dict(signed_dict)
 
 
     # Serialization.
     def to_dict(self) -> JsonDict:
         """Returns the JSON-serializable dictionary representation of self. """
         json_dict = super().to_dict()
+        target_dict = {}
+        for target_path, target_file_obj in self.targets.items():
+            target_dict[target_path] = target_file_obj.to_dict()
+
         json_dict.update({
-            'targets': self.targets,
+            'targets': target_dict,
             'delegations': self.delegations,
         })
         return json_dict
@@ -650,4 +709,5 @@ class Targets(Signed):
     # Modification.
     def update(self, filename: str, fileinfo: JsonDict) -> None:
         """Assigns passed target file info to meta dict. """
-        self.targets[filename] = fileinfo
+        self.targets[filename] = TargetInfo(fileinfo['length'],
+            fileinfo['hashes'], fileinfo.get('custom'))
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -381,6 +381,10 @@ class Root(Signed):
             self, _type: str=None, version: int=None, spec_version: str=None,
             expires: datetime=None, consistent_snapshot: bool=None,
             keys: JsonDict=None, roles: JsonDict=None) -> None:
+
+        # TODO: Add sensible defaults when we have input validation.
+        # See issue https://github.com/theupdateframework/tuf/issues/1140
+        # We need default values to create empty objects in Signed.from_dict()
         super().__init__(_type, version, spec_version, expires)
         # TODO: Add classes for keys and roles
         self.consistent_snapshot = consistent_snapshot
@@ -493,6 +497,10 @@ class Timestamp(Signed):
             self, _type: str=None, version: int=None, spec_version: str=None,
             expires: datetime=None, meta: Dict[str, MetadataInfo]=None
             ) -> None:
+
+        # TODO: Add sensible defaults when we have input validation.
+        # See issue https://github.com/theupdateframework/tuf/issues/1140
+        # We need default values to create empty objects in Signed.from_dict()
         super().__init__(_type, version, spec_version, expires)
         self.meta = meta
 
@@ -528,6 +536,8 @@ class Timestamp(Signed):
             hashes: Optional[JsonDict] = None) -> None:
         """Assigns passed info about snapshot metadata to meta dict. """
 
+        # TODO: Consider renaming this function:
+        # see: https://github.com/theupdateframework/tuf/issues/1230
         self.meta['snapshot.json'] = MetadataInfo(version, length, hashes)
 
 
@@ -549,6 +559,10 @@ class Snapshot(Signed):
             self, _type: str=None, version: int=None, spec_version: str=None,
             expires: datetime=None, meta: Dict[str, MetadataInfo]=None
             ) -> None:
+
+        # TODO: Add sensible defaults when we have input validation.
+        # See issue https://github.com/theupdateframework/tuf/issues/1140
+        # We need default values to create empty objects in Signed.from_dict()
         super().__init__(_type, version, spec_version, expires)
         self.meta = meta
 
@@ -587,6 +601,9 @@ class Snapshot(Signed):
             self, rolename: str, version: int, length: Optional[int] = None,
             hashes: Optional[JsonDict] = None) -> None:
         """Assigns passed (delegated) targets role info to meta dict. """
+
+        # TODO: Consider renaming this function:
+        # see: https://github.com/theupdateframework/tuf/issues/1230
         metadata_fn = f'{rolename}.json'
 
         self.meta[metadata_fn] = MetadataInfo(version, length, hashes)
@@ -697,6 +714,10 @@ class Targets(Signed):
             self, _type: str=None, version: int=None, spec_version: str=None,
             expires: datetime=None, targets: Dict[str, TargetInfo]=None,
             delegations: JsonDict=None) -> None:
+
+        # TODO: Add sensible defaults when we have input validation.
+        # See issue https://github.com/theupdateframework/tuf/issues/1140
+        # We need default values to create empty objects in Signed.from_dict()
         super().__init__(_type, version, spec_version, expires)
         self.targets = targets
 
@@ -737,5 +758,8 @@ class Targets(Signed):
     # Modification.
     def update(self, filename: str, fileinfo: JsonDict) -> None:
         """Assigns passed target file info to meta dict. """
+
+        # TODO: Consider renaming this function:
+        # see: https://github.com/theupdateframework/tuf/issues/1230
         self.targets[filename] = TargetInfo(fileinfo['length'],
             fileinfo['hashes'], fileinfo.get('custom'))
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -281,8 +281,6 @@ class Signed:
         self.expires = expires
 
         # TODO: Should we separate data validation from constructor?
-        if version < 0:
-            raise ValueError(f'version must be < 0, got {version}')
         self.version = version
 
 
@@ -291,22 +289,24 @@ class Signed:
     def from_dict(cls, signed_dict: JsonDict) -> 'Signed':
         """Creates Signed object from its JSON/dict representation. """
 
+        # Create empty object with default or parametrized constructor with
+        # default arguments.
+        obj = cls()
+        obj._type = signed_dict['_type']
+        obj.version = signed_dict['version']
+        obj.spec_version = signed_dict['spec_version']
         # Convert 'expires' TUF metadata string to a datetime object, which is
         # what the constructor expects and what we store. The inverse operation
         # is implemented in 'to_dict'.
-        signed_dict['expires'] = tuf.formats.expiry_string_to_datetime(
+        obj.expires = tuf.formats.expiry_string_to_datetime(
                 signed_dict['expires'])
-        # NOTE: We write the converted 'expires' back into 'signed_dict' above
-        # so that we can pass it to the constructor as  '**signed_dict' below,
-        # along with other fields that belong to Signed subclasses.
-        # Any 'from_dict'(-like) conversions of fields that correspond to a
-        # subclass should be performed in the 'from_dict' method of that
-        # subclass and also be written back into 'signed_dict' before calling
-        # super().from_dict.
+        # NOTE: Any 'from_dict'(-like) conversions of fields that correspond
+        # to a subclass should be performed in the 'from_dict' method of that
+        # subclass and assigned to the below returned object.
 
         # NOTE: cls might be a subclass of Signed, if 'from_dict' was called on
         # that subclass (see e.g. Metadata.from_dict).
-        return cls(**signed_dict)
+        return obj
 
 
     # Serialization.
@@ -378,14 +378,26 @@ class Root(Signed):
     # default max-args value for pylint is 5
     # pylint: disable=too-many-arguments
     def __init__(
-            self, _type: str, version: int, spec_version: str,
-            expires: datetime, consistent_snapshot: bool,
-            keys: JsonDict, roles: JsonDict) -> None:
+            self, _type: str=None, version: int=None, spec_version: str=None,
+            expires: datetime=None, consistent_snapshot: bool=None,
+            keys: JsonDict=None, roles: JsonDict=None) -> None:
         super().__init__(_type, version, spec_version, expires)
         # TODO: Add classes for keys and roles
         self.consistent_snapshot = consistent_snapshot
         self.keys = keys
         self.roles = roles
+
+
+    @classmethod
+    def from_dict(cls, signed_dict: JsonDict) -> 'Root':
+        """Creates Root object from its JSON/dict representation. """
+
+        # Get a parent object with its attributes already assigned.
+        obj = super().from_dict(signed_dict)
+        obj.consistent_snapshot = signed_dict['consistent_snapshot']
+        obj.keys = signed_dict['keys']
+        obj.roles = signed_dict['roles']
+        return obj
 
 
     # Serialization.
@@ -478,8 +490,9 @@ class Timestamp(Signed):
 
     """
     def __init__(
-            self, _type: str, version: int, spec_version: str,
-            expires: datetime, meta: Dict[str, MetadataInfo]) -> None:
+            self, _type: str=None, version: int=None, spec_version: str=None,
+            expires: datetime=None, meta: Dict[str, MetadataInfo]=None
+            ) -> None:
         super().__init__(_type, version, spec_version, expires)
         self.meta = meta
 
@@ -500,10 +513,14 @@ class Timestamp(Signed):
     def from_dict(cls, signed_dict: JsonDict) -> 'Timestamp':
         """Creates Timestamp object from its JSON/dict representation. """
 
-        signed_dict['meta']['snapshot.json'] = MetadataInfo(
-                **signed_dict['meta']['snapshot.json'])
+        # Get a parent object with its attributes already assigned.
+        obj = super().from_dict(signed_dict)
+        obj.meta = {}
+        meta = signed_dict['meta']['snapshot.json']
+        obj.meta['snapshot.json'] = MetadataInfo(meta['version'],
+            meta.get('length'), meta.get('hashes'))
 
-        return super().from_dict(signed_dict)
+        return obj
 
 
     # Modification.
@@ -529,8 +546,9 @@ class Snapshot(Signed):
 
     """
     def __init__(
-            self, _type: str, version: int, spec_version: str,
-            expires: datetime, meta: Dict[str, MetadataInfo]) -> None:
+            self, _type: str=None, version: int=None, spec_version: str=None,
+            expires: datetime=None, meta: Dict[str, MetadataInfo]=None
+            ) -> None:
         super().__init__(_type, version, spec_version, expires)
         self.meta = meta
 
@@ -539,11 +557,15 @@ class Snapshot(Signed):
     def from_dict(cls, signed_dict: JsonDict) -> 'Snapshot':
         """Creates Snapshot object from its JSON/dict representation. """
 
+        # Get a parent object with its attributes already assigned.
+        obj = super().from_dict(signed_dict)
+        obj.meta = {}
         for meta_path in signed_dict['meta'].keys():
-            signed_dict['meta'][meta_path] = MetadataInfo(
-                    **signed_dict['meta'][meta_path])
+            meta = signed_dict['meta'][meta_path]
+            obj.meta[meta_path] = MetadataInfo(meta['version'],
+                meta.get('length'), meta.get('hashes'))
 
-        return super().from_dict(signed_dict)
+        return obj
 
 
     # Serialization.
@@ -672,9 +694,9 @@ class Targets(Signed):
     # default max-args value for pylint is 5
     # pylint: disable=too-many-arguments
     def __init__(
-            self, _type: str, version: int, spec_version: str,
-            expires: datetime, targets: Dict[str, TargetInfo],
-            delegations: JsonDict) -> None:
+            self, _type: str=None, version: int=None, spec_version: str=None,
+            expires: datetime=None, targets: Dict[str, TargetInfo]=None,
+            delegations: JsonDict=None) -> None:
         super().__init__(_type, version, spec_version, expires)
         self.targets = targets
 
@@ -685,11 +707,17 @@ class Targets(Signed):
     @classmethod
     def from_dict(cls, signed_dict: JsonDict) -> 'Targets':
         """Creates Targets object from its JSON/dict representation. """
-        for target_path in signed_dict['targets'].keys():
-            signed_dict['targets'][target_path] = TargetInfo(
-                    **signed_dict['targets'][target_path])
 
-        return super().from_dict(signed_dict)
+        # Get a parent object with its attributes already assigned.
+        obj = super().from_dict(signed_dict)
+        obj.targets = {}
+        for target_path in signed_dict['targets'].keys():
+            info = signed_dict['targets'][target_path]
+            obj.targets[target_path] = TargetInfo(info['length'],
+                info['hashes'], info.get('custom'))
+
+        obj.delegations = signed_dict['delegations']
+        return obj
 
 
     # Serialization.
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -431,7 +431,7 @@ class Timestamp(Signed):
                         '<HASH ALGO 1>': '<SNAPSHOT METADATA FILE HASH 1>',
                         '<HASH ALGO 2>': '<SNAPSHOT METADATA FILE HASH 2>',
                         ...
-                    }
+                    } // optional
                 }
             }
 
@@ -455,13 +455,16 @@ class Timestamp(Signed):
 
 
     # Modification.
-    def update(self, version: int, length: int, hashes: JsonDict) -> None:
+    def update(self, version: int, length: Optional[int] = None,
+            hashes: Optional[JsonDict] = None) -> None:
         """Assigns passed info about snapshot metadata to meta dict. """
-        self.meta['snapshot.json'] = {
-            'version': version,
-            'length': length,
-            'hashes': hashes
-        }
+
+        self.meta['snapshot.json'] = {'version': version}
+        if length is not None:
+            self.meta['snapshot.json']['length'] = length
+
+        if hashes is not None:
+            self.meta['snapshot.json']['hashes'] = hashes
 
 
 class Snapshot(Signed):
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -420,6 +420,50 @@ class Root(Signed):
             del self.keys[keyid]
 
 
+class MetadataInfo:
+    """A container with information about a particular metadata file.
+    Instances of MetadataInfo are used as values in a dictionary called
+    "meta" in Timestamp and Snapshot.
+
+    Attributes:
+        version: An integer indicating the version of the metadata file.
+        length: An optional integer indicating the length of the metadata file.
+        hashes: A optional dictionary containing hash algorithms and the
+            hashes resulting from applying them over the metadata file.::
+
+                'hashes': {
+                    '<HASH ALGO 1>': '<METADATA FILE HASH 1>',
+                    '<HASH ALGO 2>': '<METADATA FILE HASH 2>',
+                    ...
+                }
+
+    """
+
+    def __init__(self, version: int, length: Optional[int] = None,
+            hashes: Optional[JsonDict] = None) -> None:
+        self.version = version
+        self.length = length
+        self.hashes = hashes
+
+
+    def __eq__(self, other: 'MetadataInfo') -> bool:
+        """Compare objects by their values instead of by their addresses."""
+        return (self.version == other.version and
+                self.length == other.length and
+                self.hashes == other.hashes)
+
+
+    def to_dict(self) -> JsonDict:
+        """Returns the JSON-serializable dictionary representation of self. """
+        json_dict = {'version': self.version}
+
+        if self.length is not None:
+            json_dict['length'] = self.length
+
+        if self.hashes is not None:
+            json_dict['hashes'] = self.hashes
+
+        return json_dict
 
 
 class Timestamp(Signed):
@@ -429,23 +473,14 @@ class Timestamp(Signed):
         meta: A dictionary that contains information about snapshot metadata::
 
             {
-                'snapshot.json': {
-                    'version': <SNAPSHOT METADATA VERSION NUMBER>,
-                    'length': <SNAPSHOT METADATA FILE SIZE>, // optional
-                    'hashes': {
-                        '<HASH ALGO 1>': '<SNAPSHOT METADATA FILE HASH 1>',
-                        '<HASH ALGO 2>': '<SNAPSHOT METADATA FILE HASH 2>',
-                        ...
-                    } // optional
-                }
+                'snapshot.json': <MetadataInfo INSTANCE>
             }
 
     """
     def __init__(
             self, _type: str, version: int, spec_version: str,
-            expires: datetime, meta: JsonDict) -> None:
+            expires: datetime, meta: Dict[str, MetadataInfo]) -> None:
         super().__init__(_type, version, spec_version, expires)
-        # TODO: Add class for meta
         self.meta = meta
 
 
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
         metadata_fn = f'{rolename}.json'
 
-        self.meta[metadata_fn] = {'version': version}
-        if length is not None:
-            self.meta[metadata_fn]['length'] = length
+        self.meta[metadata_fn] = MetaFile(version, length, hashes)
+
+
+class TargetFile:
+    """A container with information about a particilur target file.
```
```suggestion
    """A container with information about a particular target file.
```

tuf/api/metadata.py
```diff
@@ -415,53 +415,102 @@ def remove_key(self, role: str, keyid: str) -> None:
             del self.keys[keyid]
 
 
+class MetaFile:
+    """A container with information about a particilur file.
```
```suggestion
    """A container with information about a particular file.
```

tuf/api/metadata.py
```diff
@@ -415,53 +415,102 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+    Instances of Metafile are used as values in a dictionary called "meta" in
+    Timestamp and Snapshot:
+
+        meta: A dictionary that contains information about metadata files::
```
lukpueh: From the [reST/sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks):
>  The literal block must be [...] separated from the surrounding [...] by blank lines

Please fix here and elsewhere (also in the recently added `Root` class docstring if you don't mind).

tuf/api/metadata.py
lukpueh: And maybe say a "particular *metadata* file"?

tuf/api/metadata.py
lukpueh: Although I think I would remove the description of `meta` (+ code snippet) here, because `meta` should be and is documented in Snapshot and Timestamp.

tuf/api/metadata.py
```diff
@@ -415,53 +415,102 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+        length: An optional integer indicating the length of the metafile.
+        hashes: An optional dictionary containing hash algorithms and the
+            hashes resulted from applying them over the metafile::
+                'hashes': {
+                    '<HASH ALGO 1>': '<META FILE HASH 1>',
+                    '<HASH ALGO 2>': '<META FILE HASH 2>',
+                    ...
+                }
+
+    """
```
lukpueh, line 461: Would you mind replacing "metafile" and "META FILE" with "metadata file" in the docs? I know the spec says `"meta" : METAFILES`, but `METAFILES`  in the spec is actually a variable for the dictionary `{<METADATA FILE PATH>: <METADATA FILE INFO>, ...}`.  I think it would be easier if we sticked to "metadata file", which seems to be more prevalent in the project.

And I might also call the class `MetadataFileInfo` to express its relationship to a metadata file and also to distinguish it from the `Metadata` class. What do you think?

This might seem like nit picking, but we are likely to stick with these names for a long time. So I'd like them to be as clear. :) 

Also see https://github.com/theupdateframework/specification/issues/27 for a complaint from a former TUF maintainer about the name `meta`.

tuf/api/metadata.py
```diff
@@ -415,53 +415,102 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+        self.length = length
+        self.hashes = hashes
+
+
+    def __eq__(self, other: Any) -> bool:
+        """Used to compare objects by their values instead of by their
+        addresses."""
+        if not isinstance(other, MetaFile):
+            # don't attempt to compare against unrelated types
+            return NotImplemented
```
lukpueh: Why not use type hints and (in a follow-up PR) validate that instead of doing an `isinstance` on the passed arg?

```suggestion
    def __eq__(self, other: 'MetaFile') -> bool:
        """Used to compare objects by their values instead of by their
        addresses."""
```

tuf/api/metadata.py
```diff
@@ -415,53 +415,102 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+
+        return self.version == other.version and \
+            self.length == other.length and \
+            self.hashes == other.hashes
```
lukpueh: I think our new style guide would suggest:
```suggestion
        return (self.version == other.version and
                    self.length == other.length and
                    self.hashes == other.hashes)
```

tuf/api/metadata.py
```diff
@@ -415,53 +415,102 @@ def remove_key(self, role: str, keyid: str) -> None:
...
             expires: datetime, meta: JsonDict) -> None:
         super().__init__(_type, version, spec_version, expires)
-        # TODO: Add class for meta
-        self.meta = meta
+
+        self.meta = {}
+        self.meta['snapshot.json'] = MetaFile(
+            meta['snapshot.json']['version'],
+            meta['snapshot.json'].get('length'),
+            meta['snapshot.json'].get('hashes'))
```
lukpueh: So the way I have designed this in the other classes the constructor does not do any dict-to-object deserialization, but just receives objects and assign them to its attributes. This means the constructor here would stay as it was but with a different type hint.
```python
def __init__(
            self, _type: str, version: int, spec_version: str,
            expires: datetime, meta: Dict[str, Metafile]):
    self.meta = meta
```

And deserialization would happen  in a `from_dict` factory method that looks something like this).
```python
@classmethod
def from_dict(cls, metadata: JsonDict) -> 'Timestamp':
    metadata['meta'] = MetaFile(**metadata['meta']['snapshot.json'])
    return super().from_dict(metadata)
      
```

Please take a look at the constructors and `from_dict` methods of `Metadata` and `Signed` classes for reference (especially the latter has some details in the comments).

tuf/api/metadata.py
```diff
@@ -471,15 +520,7 @@ class Snapshot(Signed):
...
-                'targets.json': {
-                    'version': <TARGETS METADATA VERSION NUMBER>,
-                    'length': <TARGETS METADATA FILE SIZE>, // optional
-                    'hashes': {
-                        '<HASH ALGO 1>': '<TARGETS METADATA FILE HASH 1>',
-                        '<HASH ALGO 2>': '<TARGETS METADATA FILE HASH 2>',
-                        ...
-                    } // optional
-                },
+                'targets.json': <MetaFile INSTANCE>
```
lukpueh: Place also use `<MetaFile INSTANCE>` as values for `'<DELEGATED TARGETS ROLE 1>'` and `'<DELEGATED TARGETS ROLE 2>'`

tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
             expires: datetime, meta: JsonDict) -> None:
         super().__init__(_type, version, spec_version, expires)
-        # TODO: Add class for meta
-        self.meta = meta
+
+        self.meta = {}
+        for meta_path, meta_info in meta.items():
+            self.meta[meta_path] = MetaFile(meta_info['version'],
+                meta_info.get('length'), meta_info.get('hashes'))
+
```
lukpueh, line 588: See comment about `__init__` vs. `from_dict` in `Timestamp` class above.

tuf/api/metadata.py
lukpueh: Let me know if this is unclear. Also, I can see how this is not  immediately obvious when extending the model. I'm happy for suggestions for better development documentation.

tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
+        targets: A dictionary that contains information about target files::
+            {
+                'target_file_path_1': <TargetFile INSTANCE_1>,
+                'target_file_path_2': <TargetFile INSTANCE_2>,
+                ...
+            }
+    Attributes:
+        length: An integer indicating the length of the target file.
+        hashes: An dictionary containing hash algorithms and the
```
```suggestion
        hashes: A dictionary containing hash algorithms and the
```

tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
+    """A container with information about a particilur target file.
+    Instances of TargetFile are used as values in a dictionary
+    called "targets" in Targets.
+
+        targets: A dictionary that contains information about target files::
+            {
+                'target_file_path_1': <TargetFile INSTANCE_1>,
+                'target_file_path_2': <TargetFile INSTANCE_2>,
+                ...
+            }
```
lukpueh: Same comment as the one about `meta` in `MetaInfo` docstring applies here.

tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
         """Assigns passed (delegated) targets role info to meta dict. """
         metadata_fn = f'{rolename}.json'
 
-        self.meta[metadata_fn] = {'version': version}
-        if length is not None:
-            self.meta[metadata_fn]['length'] = length
+        self.meta[metadata_fn] = MetaFile(version, length, hashes)
+
+
+class TargetFile:
```
lukpueh: What to you think about `TargetFileInfo` akin to my suggestion `MetadataFileInfo` above?

tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
+    def __eq__(self, other: Any) -> bool:
+        """Used to compare objects by their values instead of by their
+        addresses."""
+        if not isinstance(other, TargetFile):
+            # don't attempt to compare against unrelated types
+            return NotImplemented
+
+        return self.length == other.length and \
+            self.hashes == other.hashes and \
+            self.custom == other.custom
```
lukpueh: See suggestion in `MetaFile` above.

tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
+        self.hashes = hashes
+        self.custom = custom
+
+
+    def __eq__(self, other: Any) -> bool:
+        """Used to compare objects by their values instead of by their
+        addresses."""
+        if not isinstance(other, TargetFile):
+            # don't attempt to compare against unrelated types
+            return NotImplemented
```
lukpueh: See suggestion in `MetaFile` above.

tuf/api/metadata.py
```diff
@@ -584,22 +680,32 @@ def __init__(
...
         super().__init__(_type, version, spec_version, expires)
-        # TODO: Add class for meta
-        self.targets = targets
+
+        self.targets = {}
+        for target_path, target_dict in targets.items():
+            self.targets[target_path] = TargetFile(target_dict['length'],
+                target_dict['hashes'], target_dict.get('custom'))
+
+        # TO DO: Add Key and Role classes
```
```suggestion
        # TODO: Add Key and Role classes
```

tuf/api/metadata.py
```diff
@@ -584,22 +680,32 @@ def __init__(
             expires: datetime, targets: JsonDict, delegations: JsonDict
             ) -> None:
         super().__init__(_type, version, spec_version, expires)
-        # TODO: Add class for meta
-        self.targets = targets
+
+        self.targets = {}
+        for target_path, target_dict in targets.items():
+            self.targets[target_path] = TargetFile(target_dict['length'],
+                target_dict['hashes'], target_dict.get('custom'))
```
lukpueh: See comment about `__init__` vs. `from_dict` in `Timestamp` class above.

lukpueh (changes requested):
Excellent work, @MVrachev and many thanks for taking a stab at this!  Don't let the many comments discourage you, they are mostly minor nit picks that should be easy to fix. Let me know when this is ready for another round for review.

tuf/api/metadata.py
MVrachev: Yep, it makes sense to rename the class to `MetadataFileInfo` and `"meta" : METAFILES` to `<METADATA FILE PATH>: <METADATA FILE INFO>, ...}`
Will do it.

tuf/api/metadata.py
MVrachev: I was not sure how to do it, because I couldn't use `MetaFile` directly, because it was not yet defined.
I will use your approach. :)

tuf/api/metadata.py
MVrachev: Yes, it looks much better that way.
It seems I have missed that part from the Google style guideline.

tuf/api/metadata.py
lukpueh: See https://www.python.org/dev/peps/pep-0484/#forward-references for more info about that scenario.

tuf/api/metadata.py
```diff
@@ -494,32 +535,95 @@ def __init__(
...
         json_dict.update({
-            'meta': self.meta
+            'meta': meta_dict
         })
         return json_dict
 
 
     # Modification.
-    def update(
+    def update_meta_file(
```
sechkova: I have mixed feelings about this API change. I can understand why you wanted to make the methods' names less ambiguous but it starts reminding me of the current implementation with long function names and even longer parameters list.

Maybe when the update is on a single `MetaFile` the `update()` method should be implemented by the `MetaFile` class. This way the interface will very roughly look like `Snapshot.{rolename}.update(version, length, hashes`).
Also you are not updating the `MetaFile` object but creating a new one which may or may not be a problem.

Same train of thoughts applies for `TargetFile`.

Since the classes' design and their interaction is a WIP we can revise this later, not in this PR.
Any thoughts on this @lukpueh?

MVrachev: I understand your concerns and you have a point.
I realized it's an ambiguous name when I and @jku were discussing something about the new API and he misunderstood what the update function does.

Also on that point, @lukpueh said `MetaFile` is a little ambiguous name too https://github.com/theupdateframework/tuf/pull/1223#discussion_r531060868 and I agreed, but then when I renamed the class to `MetadataFileInfo` I feel it became too long...
What do you think @sechkova @lukpueh?

sechkova: That's a different case of ambiguity 😁 
My issue is not that much with the name itself, more with the uniformity of the interface and the classes design.

lukpueh: This reminds me of the conversation about the `update` methods we had in https://github.com/theupdateframework/tuf/pull/1193#discussion_r514572393 pp. Let me repurpose my comment from that discussion:

I would say the name "update" in `{Targets, Snapshot, Timestamp}.update(...)` does not stand for  *"__update__ attributes of object"*  but rather for *"update attributes of object in the course of a repository __update__"*. I am talking about the Update in The Update Framework.

Right now a TUF integrator needs to call exactly these `update` methods explicitly whenever target files are added, thus above meaning of *update* made sense to me.

A bit further down the road, I want us to have a repository controller class that provides an `update` method that handles the whole chain of updates on related targets, snapshot, timestamp objects, whenever target files are added (see #1134).

Once we implement that logic in a controller, and integrators no longer need to handle updates of individual roles explicitly, above meaning of the method names becomes a bit less important. I suggest that we then (or already now) use a bit more verbose and mnemonic names than `update`.

Regarding the concerns about the length of `MetadataFileInfo` or `update_meta_file`, I can see where they come from, but I think there are still worlds between this and something like ...

```python
def _generate_and_write_metadata(rolename, metadata_filename, targets_directory, metadata_directory, storage_backend, consistent_snapshot=False, filenames=None, allow_partially_signed=False, increment_version_number=True, repository_name='default', use_existing_fileinfo=False, use_timestamp_length=True, use_timestamp_hashes=True, use_snapshot_length=False, use_snapshot_hashes=False):
```

MVrachev: > A bit further down the road, I want us to have a repository controller class that provides an update method that handles the whole chain of updates on related targets, snapshot, timestamp objects, whenever target files are added (see #1134).
Once we implement that logic in a controller, and integrators no longer need to handle updates of individual roles explicitly, above meaning of the method names becomes a bit less important. I suggest that we then (or already now) use a bit more verbose and mnemonic names than update.

I agree with this.
When we have a controller, then eventually our users will stop using those `update` methods directly and we would be able to rename them with something less ambiguous.
I think the best solution here will be to drop that commit and file an issue describing when it's suitable to make that change and discuss even if we want that change in the first place.
Agreed @lukpueh, @sechkova?

MVrachev: After a discussion with Lukas, we both agreed that `MetadataInfo` sounds like the best balance between clear, but yet a not verbose name for the class used to describing `meta` for Snapshot and Timestamp.

MVrachev:
It took me longer than expected, but I updated my pr.
I had to rebase because I dropped the commit in which I renamed `update` methods.
Additionally, I 
1. added a new commit `New API: Make sure we are not changing dict arg `
2. added a new commit `Fix documentation indentation` as proposed by Lukas.
3. Addressed all comments made by Lukas.

lukpueh:
@MVrachev and @jku have rightfully pointed out that in my hierarchic `from_dict` design, which recursively instantiates and assigns nested TUF metadata objects from their json/dict representation, the input data (i.e. the json/dict) is mutated while being parsed, which might be surprising to the caller.  

Note that this mutation only occurs in classes that use inheritance, i.e. in `Signed` and its subclasses `Targets/Snapshot/Timestamp`. Here is a simplified/abstract demonstration of the issue:
```python
class Parent:
    def __init__(self, some_parent_attr: SomeParentAttr):
        self.some_parent_attr = some_parent_attr

    @classmethod
    def from_dict(cls, data: Dict):
        # Write back into passed 'data' dict so that we can pass it as '**data'
        # to constructor, with parsed value for 'data["some_parent_attr"]',
        # alongside other 'data' values that were parsed in subclasses.
        # FIXME: Mutation of 'data' (passed by reference) might surprise caller
        data["some_parent_attr"] = SomeParentAttr.from_dict(data["some_parent_attr"])
        return cls(**data)

class Child(Parent):
    def __init__(self,
                 some_parent_attr: SomeParentAttr,
                 some_child_attr: SomeChildAttr):
        super().__init__(some_parent_attr)
        self.some_child_attr = some_child_attr

    @classmethod
    def from_dict(cls, data: Dict):
        # Parse and write back into passed 'data' dict so that we can hand it
        # on to the parent class for any additional parsing of 'data' values.
        # FIXME: Mutation of 'data' (passed by reference) might surprise caller
        data["some_child_attr"] = SomeChildAttr.from_dict([data["some_child_attr"]])
        return super().from_dict(data)
```


@MVrachev proposes to `copy` (probably rather `deepcopy`) the input data in the relevant `from_dict` methods. While this seems like a fair fix, it introduces unnecessary overhead. Here is an alternative solution:

```python
class Parent:
    @classmethod
    def from_dict(cls, data: Dict):
        # Create empty object with default or parametrized constructor with
        # default arguments.
        obj = cls()

        # Parse 'data["some_parent_attr"]' and assign it to the empty object.
        # Any child attributes are parsed and assigned later in the child
        # class.
        obj.some_parent_attr = SomeParentAttr.from_dict(data["some_parent_attr"])
        return obj

class Child(Parent):
    @classmethod
    def from_dict(cls, data: Dict):
        # Call parent parse to get object with parent attributes already
        # assigned and ...
        obj = super().from_dict(data)

        # ... parse and assign any child attributes here.
        obj.some_child_attr = SomeChildAttr.from_dict([data["some_child_attr"]])

        return obj
```

trishankatdatadog:
Just my 0.02 BTC as an observer:

One thing I regretted adding in my original Metadata API was native JSON decoding/encoding support as class methods. While it works well, it does not not decouple our internal Metadata objects from the data transport format (which need not be JSON).

In my new design, I use an abstract [parser](https://github.com/trishankatdatadog/tuf-on-a-plane/tree/0f51b6e10046e92726c893a5d79201e91ec897ec/src/tuf_on_a_plane/parsers) which can be overriden to implement different parsers for different data transport formats that nevertheless return the same type of internal Metadata objects we expect.

My new design does not yet account for how to write those internal Metadata objects back into JSON, but I don't believe it would be too difficult while keeping separation of concerns.

Anyway, please carry on, don't let me derail you, I was just making an observation based on something I caused.

lukpueh:
Good point, @trishankatdatadog. This would be a third, more procedural option, for our considerations in https://github.com/secure-systems-lab/securesystemslib/issues/272#issuecomment-691056186. And it indeed completely decouples internal Metadata objects from external transport format. 

What I liked about my proposal is that serialization/deserialization logic is organized in small chunks that are encapsulated in the related class.

But I'm really fine either way. Curious to hear what others think.

MVrachev:
> @MVrachev and @jku have rightfully pointed out that in my hierarchic `from_dict` design, which recursively instantiates and assigns nested TUF metadata objects from their json/dict representation, the input data (i.e. the json/dict) is mutated while being parsed, which might be surprising to the caller.
> 
> Note that this mutation only occurs in classes that use inheritance, i.e. in `Signed` and its subclasses `Targets/Snapshot/Timestamp`. Here is a simplified/abstract demonstration of the issue:
> 
> ```python
> class Parent:
>     def __init__(self, some_parent_attr: SomeParentAttr):
>         self.some_parent_attr = some_parent_attr
> 
>     @classmethod
>     def from_dict(cls, data: Dict):
>         # Write back into passed 'data' dict so that we can pass it as '**data'
>         # to constructor, with parsed value for 'data["some_parent_attr"]',
>         # alongside other 'data' values that were parsed in subclasses.
>         # FIXME: Mutation of 'data' (passed by reference) might surprise caller
>         data["some_parent_attr"] = SomeParentAttr.from_dict(data["some_parent_attr"])
>         return cls(**data)
> 
> class Child(Parent):
>     def __init__(self,
>                  some_parent_attr: SomeParentAttr,
>                  some_child_attr: SomeChildAttr):
>         super().__init__(some_parent_attr)
>         self.some_child_attr = some_child_attr
> 
>     @classmethod
>     def from_dict(cls, data: Dict):
>         # Parse and write back into passed 'data' dict so that we can hand it
>         # on to the parent class for any additional parsing of 'data' values.
>         # FIXME: Mutation of 'data' (passed by reference) might surprise caller
>         data["some_child_attr"] = SomeChildAttr.from_dict([data["some_child_attr"]])
>         return super().from_dict(data)
> ```
> 
> @MVrachev proposes to `copy` (probably rather `deepcopy`) the input data in the relevant `from_dict` methods. While this seems like a fair fix, it introduces unnecessary overhead. Here is an alternative solution:
> 
> ```python
> class Parent:
>     @classmethod
>     def from_dict(cls, data: Dict):
>         # Create empty object with default or parametrized constructor with
>         # default arguments.
>         obj = cls()
> 
>         # Parse 'data["some_parent_attr"]' and assign it to the empty object.
>         # Any child attributes are parsed and assigned later in the child
>         # class.
>         obj.some_parent_attr = SomeParentAttr.from_dict(data["some_parent_attr"])
>         return obj
> 
> class Child(Parent):
>     @classmethod
>     def from_dict(cls, data: Dict):
>         # Call parent parse to get object with parent attributes already
>         # assigned and ...
>         obj = super().from_dict(data)
> 
>         # ... parse and assign any child attributes here.
>         obj.some_child_attr = SomeChildAttr.from_dict([data["some_child_attr"]])
> 
>         return obj
> ```

I agree with your proposal @lukpueh and will change it.
This is a more optimized way to fix that problem.

MVrachev:
I implemented the suggestions from @lukpueh.
I had to give default function arguments for all class arguments assigned in `__init__`
for Root, Snapshot, Targets, and Timestamp classes.
I chose `None` as the easiest solution as a default argument, but
we definitely want to add proper validation which will ensure
we are not creating empty or partially populated objects.
I didn't want to create a discussion for sensible defaults
and argument validation here.
There is already an existing issue for that:
#1140 

It will be awesome if we can hear the opinions of others.
@joshuagl, @sechkova?

tuf/api/metadata.py
```diff
@@ -289,22 +287,25 @@ def __init__(
...
                 signed_dict['expires'])
-        # NOTE: We write the converted 'expires' back into 'signed_dict' above
-        # so that we can pass it to the constructor as  '**signed_dict' below,
-        # along with other fields that belong to Signed subclasses.
-        # Any 'from_dict'(-like) conversions of fields that correspond to a
-        # subclass should be performed in the 'from_dict' method of that
+        # NOTE: Any 'from_dict'(-like) conversions of fields that correspond
+        # to a subclass should be performed in the 'from_dict' method of that
         # subclass and also be written back into 'signed_dict' before calling
         # super().from_dict.
```
```suggestion
        # NOTE: Any 'from_dict'(-like) conversions of fields that correspond
        # to a subclass should be performed in the 'from_dict' method of that
        # subclass and assigned to the below returned object.
```

tuf/api/metadata.py
```diff
@@ -449,19 +501,32 @@ def to_dict(self) -> JsonDict:
...
 
+    @classmethod
+    def from_dict(cls, signed_dict: JsonDict) -> 'Timestamp':
+        """Creates Timestamp object from its JSON/dict representation. """
+
+        # Get a parent object with its attributes already assigned.
+        obj = super().from_dict(signed_dict)
+        obj.meta = {}
+        obj.meta['snapshot.json'] = MetadataInfo(
+                **signed_dict['meta']['snapshot.json'])
```
lukpueh: For the sake of consistency we might want to add and use a `MetadataInfo.from_dict` here and below instead of calling the constructor directly, but maybe that's overkill.

tuf/api/metadata.py
```diff
@@ -471,38 +536,45 @@ class Snapshot(Signed):
...
-                        ...
-                    } // optional
-                },
-                '<DELEGATED TARGETS ROLE 1>.json': {
-                    ...
-                },
-                '<DELEGATED TARGETS ROLE 2>.json': {
-                    ...
-                },
+                'targets.json': <MetadataInfo INSTANCE>
```
lukpueh: Minor nit: Can you add a trailing coma?

tuf/api/metadata.py
```diff
@@ -342,47 +343,62 @@ class Root(Signed):
...
     # we violate that. This __init__ function takes 7 arguments, whereas the
     # default max-args value for pylint is 5
     # pylint: disable=too-many-arguments
     def __init__(
-            self, _type: str, version: int, spec_version: str,
-            expires: datetime, consistent_snapshot: bool,
-            keys: JsonDict, roles: JsonDict) -> None:
+            self, _type: str=None, version: int=None, spec_version: str=None,
+            expires: datetime=None, consistent_snapshot: bool=None,
+            keys: JsonDict=None, roles: JsonDict=None) -> None:
```
lukpueh, line 388: Would you mind adding a TODO comment that briefly mentions why we need default values and that we are likely to assign only valid not-None defaults once we implement metadata object validation as discussed in https://github.com/theupdateframework/tuf/issues/1140?

tuf/api/metadata.py
lukpueh: Same goes for `TargetInfo` below.

lukpueh (commented):
Thanks for the updates, @MVrachev! Speaking of updates, should we add TODO comments to mark the `update` methods for revision as discussed in https://github.com/theupdateframework/tuf/pull/1223#discussion_r532537746 pp? Otherwise this looks quite good to me. :)

tuf/api/metadata.py
lukpueh: On a second thought, I lean towards it being overkill. Martin also just made a good point that `Metadata.from_dict` is not very likely to be used outside of `{Timestamp, Snapsho}.from_dict`.

tuf/api/metadata.py
```diff
@@ -455,13 +455,16 @@ def to_dict(self) -> JsonDict:
 
 
     # Modification.
-    def update(self, version: int, length: int, hashes: JsonDict) -> None:
+    def update(self, version: int, length: Optional[int] = None,
+            hashes: Optional[JsonDict] = None) -> None:
```
joshuagl, line 541: I'm wondering if `JsonDict` is the right type here, and elsewhere in the API? I wonder whether in the interest of not encoding our wireline format in the API we should have a more specialised type for what the specification calls Objects?

Maybe the solution is as simple as renaming `JsonDict` to something else?

I'm happy for this to be addressed in a future change, but wanted to hear what others think.

tuf/api/metadata.py
```diff
@@ -415,6 +434,50 @@ def remove_key(self, role: str, keyid: str) -> None:
             del self.keys[keyid]
 
 
+class MetadataInfo:
+    """A container with information about a particular metadata file.
+    Instances of MetadataInfo are used as values in a dictionary called
+    "meta" in Timestamp and Snapshot.
+
+    Attributes:
+        version: An integer indicating the version of the metafile.
```
```suggestion
        version: An integer indicating the version of the metadata file.
```

tuf/api/metadata.py
```diff
@@ -415,6 +434,50 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+        length: An optional integer indicating the length of the metafile.
```
```suggestion
        length: An optional integer indicating the length of the metadata file.
```

tuf/api/metadata.py
```diff
@@ -415,6 +434,50 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+        hashes: A optional dictionary containing hash algorithms and the
+            hashes resulted from applying them over the metafile::
```
```suggestion
        hashes: An optional dictionary containing hash algorithms and the
            hashes resulting from applying them over the metafile::
```

tuf/api/metadata.py
```diff
@@ -454,22 +489,31 @@ def to_dict(self) -> JsonDict:
         """Returns the JSON-serializable dictionary representation of self. """
         json_dict = super().to_dict()
         json_dict.update({
-            'meta': self.meta
+            'meta': {
+                'snapshot.json': self.meta['snapshot.json'].to_dict()
```
joshuagl, line 519: Nit: we might want to introduce constants for things like the role names and the file extensions, 'meta', etc.

joshuagl (commented):
This is looking really neat, great work @MVrachev and thanks for the superb review @lukpueh.

tuf/api/metadata.py
lukpueh: Agreed. Maybe `StrDict` is less biased. Or we just use the aliased value, i.e. `Dict[str, Any]`, which is not that much longer. Maybe this is something we could revisit with #1140?

joshuagl: I don't have a good name suggestion, so would lean towards the explicit aliased value of `Dict[str, Any]` to start with. I'm more than happy to revisit in a future PR.

MVrachev: I agree that something more specific is needed.
`StrDict` sounds good. I am not sure `Dict[str, Any]` is what we need, especially that values could be `Any` when they
should be strings. Maybe like `Dict[str, str]` is better?

tuf/api/metadata.py
MVrachev: I agree we need a module containing different constants. 
I will add an issue for that so we can discuss it as a change in a future pr.

MVrachev:
I addressed all comments made by @joshuagl and rebased to squash some of the commits and prepare this pr for merging.

joshuagl:
Closing to try and re-trigger GHA.

joshuagl:
There are linter failures on the new changes, could you please take a look @MVrachev? Note: it's likely that we need to update pylint configuration for the new coding style.

MVrachev:
I added one commit where I disable inline some of the irrelevant warnings and for warning `E1101` I decided to just disable it because it was creating a lot of false-positives.

tuf/api/metadata.py
```diff
@@ -289,8 +289,11 @@ def from_dict(cls, signed_dict: JsonDict) -> 'Signed':
 
         # Create empty object with default or parametrized constructor with
         # default arguments.
-        obj = cls()
-        obj._type = signed_dict['_type']
+        # Warnings about rules E1120 and W0212 are not relevant here because
+        # cls is most likely a child class of "Signed" and we need to setup
+        # the appropriate fields.
+        obj = cls() # pylint: disable=E1120
```
joshuagl, line 295: 🤔 this error is `no-value-for-parameter`, which is that we're creating an instance of the class without passing values `__init__`'s arguments. Seems OK given the pattern we decided to adopt here.

tuf/api/metadata.py
```diff
@@ -289,8 +289,11 @@ def from_dict(cls, signed_dict: JsonDict) -> 'Signed':
...
+        obj._type = signed_dict['_type'] # pylint: disable=W0212
```
joshuagl, line 300: W0212 is `protected-access`, we're accessing a protected member outside of the class or a descendent. This makes sense given the pattern of creating an empty object, updating the fields, then returning the populated objet.

tuf/api/metadata.py
joshuagl, line 299: E1120 is `no-value-for-parameter`, which is that we're creating an instance of the class without passing values for `__init__`'s arguments.

tuf/api/pylintrc
```diff
@@ -1,5 +1,5 @@
 [MESSAGE_CONTROL]
-disable=fixme
+disable=fixme,E1101
```
joshuagl, line 2: E1101 is `no-member`, a variable has been accessed which the type checker doesn't think is a valid member of that type. That makes sense, we're upcasting to the base 


Rather than disabling this globally, it would probably make more sense to disable it for only the `tuf.api.metadata` module (by including the `#pylint: disable=` comment at the top of the module's file. It may be cleaner to do this for the 4 specific fields by using the typechecker's `generated-members` configuration, although these aren't really generated members so perhaps not.

joshuagl: E1101 is `no-member`, a variable has been accessed which the type checker doesn't think is a valid member of that type. That makes sense, we're upcasting to the base `Signed` type, then accessing members of the subtype.

Rather than disabling this globally, it would probably make more sense to disable it for only the `tuf.api.metadata` module (by including the `#pylint: disable=` comment at the top of the module's file).

joshuagl (commented):
Orthogonal to this PR, given the number of linter warnings we're having to disable related to our `from_dict` pattern, I wonder whether we should document this (and consider other options?) with an ADR. cc @lukpueh

tuf/api/metadata.py
lukpueh: AFAIU `_type` is not actually meant to be "protected" in the Python OOP sense (the naming dates back before our transition to more OOP), but rather is a workaround to not shadow the builtin `type` name.

lukpueh:
> Orthogonal to this PR, given the number of linter warnings we're having to disable related to our from_dict pattern, I wonder whether we should document this (and consider other options?) with an ADR. cc @lukpueh

Agreed. I'll put something together.

tuf/api/metadata.py
joshuagl: Of course, let's add a comment to that effect?

MVrachev:
Updated the pr.
Added a more detailed comment about `E1120 and W0212` and disabled `E1101` only for `tuf/api/metadata.py`.
Also, added a commit with a comment for `_type`.

Had to rebase in order to return `pylintrc` back to its original state.

MVrachev:
I updated this pr and had to rebase in order to replace the `**` key-value syntax I was using when creating class instances with
directly passing the arguments needed for the `__init__` functions.

tuf/api/metadata.py
joshuagl: `Dict[str, str]` would work for this specific case, but would not be a suitable generic type for all the places we are currently using `JsonDict`.

I have added [a comment](https://github.com/theupdateframework/tuf/issues/1140#issuecomment-766781447) to #1140, let's move the discussion there.

tuf/api/metadata.py
```diff
@@ -415,6 +441,50 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+class MetadataInfo:
+    """A container with information about a particular metadata file.
+    Instances of MetadataInfo are used as values in a dictionary called
+    "meta" in Timestamp and Snapshot.
+
+    Attributes:
+        version: An integer indicating the version of the metadata file.
+        length: An optional integer indicating the length of the metadata file.
+        hashes: A optional dictionary containing hash algorithms and the
+            hashes resulting from applying them over the metadata file.::
```
joshuagl, line 453: I wonder if something like the following might be clearer? Or am I nitpicking?

```suggestion
        hashes: An optional dictionary mapping hash algorithms to the
            hashes resulting from applying them over the metadata file
            contents.::
```

tuf/api/metadata.py
```diff
@@ -415,6 +441,50 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+
+    def __init__(self, version: int, length: Optional[int] = None,
+            hashes: Optional[JsonDict] = None) -> None:
+        self.version = version
+        self.length = length
+        self.hashes = hashes
+
+
+    def __eq__(self, other: 'MetadataInfo') -> bool:
+        """Compare objects by their values instead of by their addresses."""
```
joshuagl, line 471: Nit: by default Python compares `object`s by their identity, which is [_like_ the address](https://docs.python.org/3/reference/datamodel.html?highlight=identity) but may not be. Can we use the Pythonic term identity here?

tuf/api/metadata.py
```diff
@@ -512,14 +606,64 @@ def update(
...
+
+class TargetInfo:
+    """A container with information about a particular target file.
+    Instances of TargetInfo are used as values in a dictionary
+    called "targets" in Targets.
+
+    Attributes:
+        length: An integer indicating the length of the target file.
+        hashes: A dictionary containing hash algorithms and the
+            hashes resulted from applying them over the target file::
```
joshuagl, line 625: Similar to the above suggestion, I'd welcome feedback on whether this is clearer or not?
```suggestion
        hashes: A dictionary mapping hash algorithms to the
            hashes resulting from applying them over the metadata file
            contents::
```

tuf/api/metadata.py
```diff
@@ -570,22 +570,66 @@ def update(
...
+
+    def __init__(self, length: int, hashes: JsonDict,
+            custom: Optional[JsonDict] = None) -> None:
+        self.length = length
+        self.hashes = hashes
+        self.custom = custom
+
+
+    def __eq__(self, other: 'TargetInfo') -> bool:
+        """Compare objects by their values instead of by their addresses."""
```
joshuagl, line 653: Same comment about identity here.

tuf/api/metadata.py
```diff
@@ -289,22 +291,27 @@ def __init__(
     def from_dict(cls, signed_dict: JsonDict) -> 'Signed':
         """Creates Signed object from its JSON/dict representation. """
 
+        # Create empty object with default or parametrized constructor with
```
joshuagl, line 294:
```suggestion
        # Create empty object with default or parameterized constructor with
```

tuf/api/metadata.py
```diff
@@ -291,22 +289,24 @@ def __init__(
...
+        obj.spec_version = signed_dict['spec_version']
         # Convert 'expires' TUF metadata string to a datetime object, which is
         # what the constructor expects and what we store. The inverse operation
         # is implemented in 'to_dict'.
-        signed_dict['expires'] = tuf.formats.expiry_string_to_datetime(
+        obj.expires = tuf.formats.expiry_string_to_datetime(
                 signed_dict['expires'])
-        # NOTE: We write the converted 'expires' back into 'signed_dict' above
-        # so that we can pass it to the constructor as  '**signed_dict' below,
-        # along with other fields that belong to Signed subclasses.
```
joshuagl, line 299: Should the removal of this comment indicate we should no longer be doing the conversion of the expires string to a datetime object here?

joshuagl (commented):
Thank you for your continued attention on this @MVrachev. I left a few comments, only one of which feels like a blocker – you removed a comment about converting `expires` so that it can be passed to the constructor as `**signed_dict`, but you've changed the code away from using that pattern. Should the conversion of expires to a datetime object be removed in favour of it being handled elsewhere?

Nice catch on the sphinx doc strings, can we add some kind of CI/linter check for these in future to prevent them being merged in an incorrect state? This feels issue worthy, rather than something we need to resolve here.

@lukpueh I'd appreciate your review on this, I don't feel quite up to speed on the new metadata API yet.

MVrachev:
On hold, until we merge #1270 and #1279.

joshuagl:
I've converted this PR to draft until it is ready for review. Mostly to save myself from checking the discussion each time to see what the status is.

tuf/api/metadata.py
MVrachev: Yep, didn't know about that.

MVrachev:
I feel like this pr has a lot of discussions and changes that are outdated compared to our current code.

I made a new branch migrating the changes from this pr and found out that: c37bdd2608dbbcce7b86c161f30243e4af8d9ed7 and the discussion we had with @lukpueh starting from [here](https://github.com/theupdateframework/tuf/pull/1223#issuecomment-737188686) are no longer relevant to our code, commits 
c37bdd2608dbbcce7b86c161f30243e4af8d9ed7 and b9c70aae06d5c16d266ff739d151ce6de0a8280d are different and
I dropped commit 383e26044ee96e8616861838d7e620c07ae89b5f.

I feel it makes more sense to close this pr and open a new one.

MVrachev:
Replaced by https://github.com/theupdateframework/tuf/pull/1329.

PR #1224, opened 2020-11-25
Remove redundant temp folder and unused imports

MVrachev:

In many of the tests classes, we are creating two
temporary directories: one with the name "temp_<random_string>"
and inside it, we generate a new directory for each test with the name
`Test<Class_Name>_<random_string>`.

I think we don't need the "temp_<random_string>" directory.
The only benefit I can think of is that it could contain multiple
temp folders from failing tests from a particular test run.
But even then, this is not a big bonus because the name
`temp_<random_string>` is not really descriptive from which test
run this directory was created.

PS: Thanks to Jussi Kukkonen who noticed we are using two temp
folders per class in our tests.

requirements**:

### DIFF
--- a/tests/test_arbitrary_package_attack.py
+++ b/tests/test_arbitrary_package_attack.py
@@ -37,7 +37,6 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import os
-import tempfile
 import shutil
 import json
 import logging
@@ -50,7 +49,7 @@ import tuf.roledb
 import tuf.keydb
 import tuf.log
 import tuf.client.updater as updater
-import tuf.unittest_toolbox as unittest_toolbox
+from tuf import unittest_toolbox
 
 from tests import utils
 
@@ -64,11 +63,6 @@ class TestArbitraryPackageAttack(unittest_toolbox.Modified_TestCase):
 
   @classmethod
   def setUpClass(cls):
-    # Create a temporary directory to store the repository, metadata, and target
-    # files.  'temporary_directory' must be deleted in TearDownModule() so that
-    # temporary files are always removed, even when exceptions occur.
-    cls.temporary_directory = tempfile.mkdtemp(dir=os.getcwd())
-
     # Launch a SimpleHTTPServer (serves files in the current directory).
     # Test cases will request metadata and target files that have been
     # pre-generated in 'tuf/tests/repository_data', which will be served by the
@@ -85,16 +79,11 @@ class TestArbitraryPackageAttack(unittest_toolbox.Modified_TestCase):
     # Cleans the resources and flush the logged lines (if any).
     cls.server_process_handler.clean()
 
-    # Remove the temporary repository directory, which should contain all the
-    # metadata, targets, and key files generated of all the test cases.
-    shutil.rmtree(cls.temporary_directory)
-
 
 
 
   def setUp(self):
-    # We are inheriting from custom class.
-    unittest_toolbox.Modified_TestCase.setUp(self)
+    super().setUp()
 
     self.repository_name = 'test_repository1'
 
@@ -103,7 +92,7 @@ class TestArbitraryPackageAttack(unittest_toolbox.Modified_TestCase):
     # The 'repository_data' directory is expected to exist in 'tuf/tests/'.
     original_repository_files = os.path.join(os.getcwd(), 'repository_data')
     temporary_repository_root = \
-      self.make_temp_directory(directory=self.temporary_directory)
+        self.make_temp_directory(directory=os.getcwd())
 
     # The original repository, keystore, and client directories will be copied
     # for each test case.
@@ -142,9 +131,7 @@ class TestArbitraryPackageAttack(unittest_toolbox.Modified_TestCase):
 
 
   def tearDown(self):
-    # Modified_TestCase.tearDown() automatically deletes temporary files and
-    # directories that may have been created during each test case.
-    unittest_toolbox.Modified_TestCase.tearDown(self)
+    super().tearDown()
     # updater.Updater() populates the roledb with the name "test_repository1"
     tuf.roledb.clear_roledb(clear_all=True)
     tuf.keydb.clear_keydb(clear_all=True)
--- a/tests/test_repository_lib.py
+++ b/tests/test_repository_lib.py
@@ -32,7 +32,6 @@ import os
 import time
 import datetime
 import logging
-import tempfile
 import json
 import shutil
 import unittest
@@ -42,10 +41,11 @@ import sys
 import tuf
 import tuf.formats
 import tuf.log
-import tuf.formats
 import tuf.roledb
 import tuf.keydb
 import tuf.settings
+
+from tuf import unittest_toolbox
 
 import tuf.repository_lib as repo_lib
 import tuf.repository_tool as repo_tool
@@ -67,16 +67,11 @@ TOP_LEVEL_METADATA_FILES = ['root.json', 'targets.json', 'timestamp.json',
                             'snapshot.json']
 
 
-class TestRepositoryToolFunctions(unittest.TestCase):
+class TestRepositoryToolFunctions(unittest_toolbox.Modified_TestCase):
   @classmethod
   def setUpClass(cls):
-    # Create a temporary directory to store the repository, metadata, and target
-    # files.  'temporary_directory' must be deleted in TearDownClass() so that
-    # temporary files are always removed, even when exceptions occur.
     tuf.roledb.clear_roledb(clear_all=True)
     tuf.keydb.clear_keydb(clear_all=True)
-    cls.temporary_directory = tempfile.mkdtemp(dir=os.getcwd())
-
 
 
   @classmethod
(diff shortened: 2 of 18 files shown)


tests/test_repository_lib.py
```diff
@@ -86,24 +81,25 @@ def tearDownClass(cls):
     tuf.roledb.clear_roledb(clear_all=True)
     tuf.keydb.clear_keydb(clear_all=True)
 
-    shutil.rmtree(cls.temporary_directory)
-
 
   def setUp(self):
+    # We are inheriting from custom class.
+    unittest_toolbox.Modified_TestCase.setUp(self)
```
joshuagl: This pattern looks odd, isn't the normal way to do this with `super().setUp(self)` ?

tests/test_arbitrary_package_attack.py
```diff
@@ -102,8 +92,10 @@ def setUp(self):
     # any modifications made to repository files are restricted to the copies.
     # The 'repository_data' directory is expected to exist in 'tuf/tests/'.
     original_repository_files = os.path.join(os.getcwd(), 'repository_data')
+    # Create a temporary directory to store the repository, metadata,
+    # and target files.
```
joshuagl: I don't think we need to include this comment here or throughout the tests you've edited.

tests/test_repository_lib.py
```diff
@@ -86,24 +81,25 @@ def tearDownClass(cls):
...
     tuf.keydb.create_keydb('test_repository')
+    self.temporary_directory = self.make_temp_directory(directory=os.getcwd())
 
 
   def tearDown(self):
     tuf.roledb.clear_roledb(clear_all=True)
     tuf.keydb.clear_keydb(clear_all=True)
-
+    # Modified_TestCase.tearDown() automatically deletes temporary files and
+    # directories that may have been created during each test case.
```
joshuagl: Feels like an unnecessary comment throughout the PR

tests/test_repository_tool.py
```diff
@@ -42,6 +42,7 @@
 import tuf.formats
 import tuf.roledb
 import tuf.keydb
+import tuf.unittest_toolbox as unittest_toolbox
```
joshuagl: `from tuf import unittest_toolbox`

tests/test_repository_lib.py
```diff
@@ -42,11 +41,12 @@
 import tuf
 import tuf.formats
 import tuf.log
-import tuf.formats
 import tuf.roledb
 import tuf.keydb
 import tuf.settings
 
+import tuf.unittest_toolbox as unittest_toolbox
```
joshuagl: `from tuf import unittest_toolbox` should suffice

tests/test_root_versioning_integration.py
```diff
@@ -38,6 +38,7 @@
 import tuf.exceptions
 import tuf.roledb
 import tuf.keydb
+import tuf.unittest_toolbox as unittest_toolbox
```
joshuagl: `from tuf import unittest_toolbox`?

tests/test_repository_lib.py
```diff
@@ -86,24 +81,25 @@ def tearDownClass(cls):
     tuf.roledb.clear_roledb(clear_all=True)
     tuf.keydb.clear_keydb(clear_all=True)
 
-    shutil.rmtree(cls.temporary_directory)
-
 
   def setUp(self):
+    # We are inheriting from custom class.
```
joshuagl: I don't think we need the comment to justify chaining up to the parent implementation of the method here, or throughout the tests you have edited.

joshuagl (commented):
Thank you for taking the time to clean this up. I have a few minor comments, mostly on not adding semi-redundant comments to the code.

MVrachev:
Addressed all of @joshuagl comments and:
1. Replaced `import tuf.unittest_toolbox as unittest_toolbox` with `from tuf import unittest_toolbox` in all of our test files.
2. Removed the comments I have added such as `We are inheriting from custom class.` from all of the tests.
3. Replaced all `unittest_toolbox.Modified_TestCase.function(self)` calls with `super().function()` in all of our tests.

MVrachev:
It seems like because we are supporting `Python 2` we have a problem with 
`super().function()` which I used because in Python2  it is `super(self).function()`
Probably I should replace it with `unittest_toolbox.Modified_TestCase.function(self)` the way it was?

joshuagl:
Ah, yes. The syntax changed for `super()` between Python 2 and Python 3. Let's leave it as is for now and we can fix this later when we've purged Python 2 from our codebase.

joshuagl:
Would be good to get this cleanup landed. @MVrachev could you revert back to `unittest_toolbox.Modified_TestCase.function(self)` when you have some cycles? Thank you.

jku:
I'm a bit worried about
* using os.getcwd() for the location where temp dirs are created
* the fact that sometimes temp dirs are not cleaned up (is this only when test run is cancelled? not sure)
* after this PR probably creating more temp dirs in current dir (one per instance instead of one per class I guess?)

So I wonder if this should only be done if we also start creating the temp dirs somewhere more reasonable?

Does anyone know when the temp dirs are not being removed?

sechkova:
> Does anyone know when the temp dirs are not being removed?

Besides the expected case when `tearDown` is not called due to some error, I've noticed temp dirs not being removed on Windows with "being used by another process" error. Unfortunately, I can't say which is the process and I haven't observed this consistently in other environments.

sechkova:
> So I wonder if this should only be done if we also start creating the temp dirs somewhere more reasonable?

I would vote for that if possible. Polluting the tests source code dir on a failure is not particularly harmful but ... unpleasant

MVrachev:
> I'm a bit worried about
> 
> * using os.getcwd() for the location where temp dirs are created
>
> * the fact that sometimes temp dirs are not cleaned up (is this only when test run is cancelled? not sure)


> * after this PR probably creating more temp dirs in current dir (one per instance instead of one per class I guess?)

Yes, the side effect is that it will create one temp dir per class instance.

> So I wonder if this should only be done if we also start creating the temp dirs somewhere more reasonable?

As I understand it, the idea is to have those temp files available for you on test failure for analysis.
Probably it will be better to choose a deterministic path where they can be found, but where should that be?
Maybe `~/tmp?`


> Does anyone know when the temp dirs are not being removed?

The [doc](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown) says:
`tearDown`:
```
Method called immediately after the test method has been called and the result recorded. 
This is called even if the test method raised an exception, so the implementation 
in subclasses may need to be particularly careful about checking the internal state.
...
This method will only be called if the setUp() succeeds, regardless of the outcome 
of the test method. 
```

Again from the [doc](https://docs.python.org/3/library/unittest.html#setupclass-and-teardownclass) for tearDownClass:
```
If an exception is raised during a setUpClass then the tests in the class are not run and 
the tearDownClass is not run. 
```

Also again in the [doc](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures)
```
Skipped tests will not have setUp() or tearDown() run around them. 
Skipped classes will not have setUpClass() or tearDownClass() run.
```

joshuagl:
> As I understand it, the idea is to have those temp files available for you on test failure for analysis.
> Probably it will be better to choose a deterministic path where they can be found, but where should that be?
> Maybe `~/tmp?`

Python has a function to get an OS appropriate, administrator configurable, temporary directory [`tempfile.gettempdir()`](https://docs.python.org/3/library/tempfile.html?tempfile.mkdtemp#tempfile.gettempdir). As well as providing functions to create directories within that host appropriate location: [`tempfile.mkdtemp()`](https://docs.python.org/3/library/tempfile.html?tempfile.mkdtemp#tempfile.mkdtemp), and [`tempfile.TemporaryFile()`](https://docs.python.org/3/library/tempfile.html?tempfile.mkdtemp#tempfile.TemporaryFile).

sechkova:
> > Does anyone know when the temp dirs are not being removed?
> 
> Besides the expected case when `tearDown` is not called due to some error, I've noticed temp dirs not being removed on Windows with "being used by another process" error. Unfortunately, I can't say which is the process and I haven't observed this consistently in other environments.

Sorry for giving you a bit random hints and not direct answers but in the Windows case temporary files created by `unittest_toolbox.Modified_TestCase.make_temp_data_file` cannot be cleaned due to 
`PermissionError(13, 'This process cannot access the file because it is being used by another process')`. 

But this is a Windows VM so it can be something specific to this case ...

MVrachev:
Blocked, until we remove Python2 from our tests as Joshua has mentioned.
https://github.com/theupdateframework/tuf/pull/1224#issuecomment-738333840

MVrachev:
After the removal of python2 test cases in https://github.com/theupdateframework/tuf/commit/13b085712f2b4b83fc2e7014c34016d55061ab7f I rebased and updated this pr.

MVrachev:
Honestly, I am wondering about this change.
On one hand, on a successful test run, we had generated multiple additional folders for each test class,
on another hand, on an error probably we will end up with many folders for each test class that failed, right?

jku:
I might have broken this PR quite badly with a recent bug fix (moved chained up teardown calls to the correct place in #1346 to fix issues found in another PR). Sorry about that -- I did not remember this one.

I think that PR also makes this one a bit less needed: at least in the tests that I touched I made sure that we only try to remove the "topmost" directory (because that will remove every other directory inside it). In some cases there might still be unneeded temp directories.

Let's discuss when you are back.

MVrachev:
As I said before, I don't feel so confident about this change anymore.
I will just close it for now and if somebody complains about it in the future, I will reopen or push another one.

PR #1229, opened 2020-11-30
ADR0004: Justify extent of OOP in metadata model

### LINKED ISSUE #1133: Decide on extent of OOP in metadata model
Should we create classes for dictionary attributes of inner metadata classes (Targets/Snapshot/Timestamp/Root)?

Con: Seems a bit over-engineered, removes the light-footedness that comes with Python dictionaries

Pro: Makes our use of OOP more consistent.
Why would we use classes for Metadata/Signed/Targets/Snapshot, etc. but not for the inner dicts? (see #1112)

Pro: Makes it easier to validate formats, if complex data structures are defined by classes and simple data types use type annotations. (see #1130)

lukpueh:
Fixes #1133 

Add MADR that justifies why we want to add custom classes for complex tuf metadata attributes.

requirements**:

### DIFF
--- a/docs/adr/0004-extent-of-OOP-in-metadata-model.md
+++ b/docs/adr/0004-extent-of-OOP-in-metadata-model.md
@@ -1,0 +1,48 @@
+# Add classes for complex metadata attributes
+
+* Status: accepted
+* Date: 2020-11-30
+
+Technical Story: https://github.com/theupdateframework/tuf/issues/1133
+
+## Context and Problem Statement
+Custom classes for the TUF signed metadata wrapper (Metadata) and metadata
+payload containers (Root, Timestamp, Snapshot, Targets) were added recently.
+Complex attributes on these classes are still represented as dictionaries.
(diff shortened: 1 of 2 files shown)


MVrachev (approved):
LGTM

mnm678 (approved):

trishankatdatadog:
💯 💯 

In my own [experience](https://github.com/trishankatdatadog/tuf-on-a-plane), it has been extremely useful to decouple the internal, higher-level representation of metadata from the data format themselves. Not only does this let you use different data formats with the same high-level representation, but it also lets you write [better](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/0f51b6e10046e92726c893a5d79201e91ec897ec/src/tuf_on_a_plane/models/metadata.py#L133-L144) code that is independent of the data format (as should be the case).

lukpueh:
Thanks for making a strong case for this proposal, @trishankatdatadog!

PR #1235, opened 2020-12-08
client: update expiration check to match spec

### LINKED ISSUE #1231: Expiration check does not follow the spec

The metadata expiry check [code](https://github.com/theupdateframework/tuf/blob/fc4b77f92ce3ffc492ea2f1e630a4217c2646f3e/tuf/client/updater.py#L2269) does not follow the [specification](https://github.com/theupdateframework/specification/blob/56ef9545ea244d9c5397bce72278eb01483888d2/tuf-spec.md):

`Check for a freeze attack. The expiration timestamp in the trusted $ROLE metadata file MUST be higher than the fixed update expiration time.`

**Current behavior**:

`expires < now`

**Expected behavior**:

`expires <= now`

joshuagl:
Fixes #1231

The specification, as of 1.0.16, describes an update expiration check as:

> The expiration timestamp in the trusted $ROLE metadata file MUST be
  higher than the fixed update expiration time.

Having done some research into how other security providers are comparing
expiration equivalents (i.e. OpenSSL x509 certificate checking code, and
GnuPG expiration checks), and how other TUF implementations are performing
the same check (rust-tuf, go-tuf), we came to a consensus that the correct
way to implement expiration comparisons is:

    expiration <= now

Where:
  expiration: is the metadata's expiration datetime
  now: is the current system time, or the fixed notion of time in the
       detailed client workflow (introduced in 1.0.16 of the spec)

requirements**:

### DIFF
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -60,6 +60,11 @@ import sys
 import unittest
 import json
 
+if sys.version_info >= (3, 3):
+  import unittest.mock as mock
+else:
+  import mock
+
 import tuf
 import tuf.exceptions
 import tuf.log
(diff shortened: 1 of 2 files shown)


mnm678 (approved):

trishankatdatadog (commented):
LGTM, thanks, Joshua!

Could we quickly mock some tests, please, if that's not too much to ask?

Might also help to test we're handling UTC correctly, it was not obvious to me from reading the code...

joshuagl:
> Could we quickly mock some tests, please, if that's not too much to ask?

Done in follow-on commit fccd078

> Might also help to test we're handling UTC correctly, it was not obvious to me from reading the code...

I'm not sure the test I added is doing this, could you elaborate a bit more on what your concern is? Perhaps we can add a test to check.

trishankatdatadog:
> I'm not sure the test I added is doing this, could you elaborate a bit more on what your concern is? Perhaps we can add a test to check.

I think your tests cover `<`, `==`, and `>`, so we should be OK.

MVrachev (approved):
LGTM!

jku (approved):

lukpueh (approved):

PR #1238, opened 2020-12-10
Drop support for Python 3.5

lukpueh:
that is provided, the better.

Fixes #<ISSUE NUMBER>

Python 3.5 has now reached its end-of-life and has been retired:
 https://www.python.org/dev/peps/pep-0478/

The optional (but highly recommended) 'cryptography' dependency  has also just dropped support for 3.5. Continuing support for 3.5  in TUF does not seem worth the effort.

This PR also updates the test configuration accordingly and recomputes the pinned dependencies. 

requirements**:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ matrix:
   include:
     - python: "2.7"
       env: TOXENV=py27
-    - python: "3.5"
-      env: TOXENV=py35
     - python: "3.6"
       env: TOXENV=py36
     - python: "3.7"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,5 @@
 environment:
   matrix:
-    - PYTHON: "C:\\Python39"
-      PYTHON_VERSION: 3.9
-      PYTHON_ARCH: 32
-
     - PYTHON: "C:\\Python38"
       PYTHON_VERSION: 3.8
       PYTHON_ARCH: 32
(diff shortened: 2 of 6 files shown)


mnm678 (approved):

joshuagl (approved):
Odd that Dependabot didn't detect Enum, I wonder if it's not picking up dependencies which are only used on Python2?

lukpueh:
> ... I wonder if it's not picking up dependencies which are only used on Python2?

Could be. I'll keep an eye on it.

Btw. I'm also thinking of a better way to auto-update `requirements-pinned.txt` (see [FIXME comment](https://github.com/in-toto/in-toto/blob/6337ce766a77166f58d769e3c7096819c900f3b2/requirements.txt#L20-L30) in in-toto requirements docs, which are the same as for tuf/sslib). I will at least follow up with an issue shortly.

joshuagl:
> Btw. I'm also thinking of a better way to auto-update `requirements-pinned.txt` (see [FIXME comment](https://github.com/in-toto/in-toto/blob/6337ce766a77166f58d769e3c7096819c900f3b2/requirements.txt#L20-L30) in in-toto requirements docs, which are the same as for tuf/sslib). I will at least follow up with an issue shortly.

Could be a good use of a scheduled GitHub action, much like securesystemlib's ed25519 check?

lukpueh:
> Could be a good use of a scheduled GitHub action, much like securesystemlib's ed25519 check?

Exactly what I had in mind. :)

Issue #1213, opened 2020-11-13
updater: abstract out the network IO


jku:
This might be relevant to Updater redesign (#1135) and if accepted would deprecate #1142 and the PR #1171  

We (me, Joshua, Martin, Teodora) have been talking about abstracting some of the client functionality out of the Updater itself. The biggest issue from my perspective is network IO. Teodora already made a PR to let the application download targets but it seems like there are still issues with TUF handling metadata downloads.

## Why is this needed?

* In the real world applications are already using a network stack and will be using it after integrating TUF as well: we should not force another one on them
* Even if the network stacks of the application and TUF are same, the fact that they use different sessions and configurations is not great
* Complex applications have legitimate needs to configure a lot of things we don't want to provide API for: user agent, proxies, basic authentication, custom request headers. This applies to both metadata and targets
* Complex applications have legitimate needs to control the download process (e.g. progress information, canceling)
* Complex applications have (legitimate?) needs to poke at low level details like timeouts

## Potential solutions

We identified two main solutions to this:
1. Make a new event-based non-blocking client API. This would be most flexible but also more complex for TUF maintainers to maintain and application developers to customize 
1. Keep the current API but add a new Fetcher interface that applications can optionally implement. This is likely fairly easy and non-invasive to implement but remains a blocking API

I'm proposing option 2 but for reference please see the [draft](https://gist.github.com/jku/e100bc2a676867e56c8e830a92bed751) of option 1 as well.

### Proposal

Add a Fetcher interface that applications can implement. Provide a default implementation of Fetcher. Add a new method to Updater that Fetcher can use to provide the data it fetches. 

The Updater processes (`refresh()`, `get_one_valid_targetinfo() `and `download_target()`) will now look like this:
* Whenever a remote file (metadata or target) is needed:
  * setup a temporary file to write results to
  * call `Fetcher.fetch()`
    * fetcher calls `Updater.provide_fetched_data()` zero or more times to provide chunks of data. Updater writes these chunks into the file
  * when fetcher returns without exceptions, the download is finished and written to the file

[...]

Issue #1188, opened 2020-10-26
Provide a way to customize requests.Session


gdetrez:
**Description of issue or feature request**:
Would it be possible to have a safe way for the user of the update client to customize the Session object used to download files?  E.g. providing (limited?) custom parameters for the `Session()`?

I have two use cases in mind:
- provide Authorization credentials (e.g. basic auth or bearer token) that may be needed to access the files
- customizing the user-agent header to prepend the application's own `<name>/<version>`

trishankatdatadog:
@gdetrez you might be interested in following #1171, which should let you do what you want...

joshuagl:
See also #1213, which includes a proposal for abstracting out the networking pieces for greater customisation.

jku:
I am closing this: both ngclient and the legacy client include a download abstraction that allows any shenanigans a client app might want to do. The included RequestsFetcher implementation does not explicitly expose the requests session but for this purpose a client could easily create their own customized fetcher that modifies the session (this is what I did in my pip POC).

PR #1250, opened 2020-12-22
Abstract out the network IO

sechkova:

Based on the discussion in #1213 and mainly the idea of @jku, implement an abstract `FetcherInterface` calss and
a concrete `RequestsFetcher` class .

The idea is to extract the network IO related code out of `tuf.download` and put it in `RequestsFetcher` class which is also the default `FetcherInterface` implementation which TUF uses if not given an external one.

`RequestsFetcher` implements only one public method '`fetch`' which returns a bytes iterator. All file handling and verification
logic remains in `tuf.download` which is now wrapper in a `FileDownload` class for convenience.

requirements**:

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -43,6 +43,7 @@ import warnings
 
 import tuf
 import tuf.download as download
+import tuf.fetcher
 import tuf.log
 import tuf.unittest_toolbox as unittest_toolbox
 import tuf.exceptions
@@ -85,6 +86,10 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     digest = m.hexdigest()
     self.target_hash = {'md5':digest}
 
+    # Initialize the default fetcher for the download
+    self.fetcher = tuf.fetcher.RequestsFetcher()
+
+
 
   # Stop server process and perform clean up.
   def tearDown(self):
@@ -100,7 +105,7 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
   def test_download_url_to_tempfileobj(self):
 
     download_file = download.safe_download
-    with download_file(self.url, self.target_data_length) as temp_fileobj:
+    with download_file(self.url, self.target_data_length, self.fetcher) as temp_fileobj:
       temp_fileobj.seek(0)
       temp_file_data = temp_fileobj.read().decode('utf-8')
       self.assertEqual(self.target_data, temp_file_data)
@@ -118,18 +123,18 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     # the server-reported length of the file does not match the
     # required_length.  'updater.py' *does* verify the hashes of downloaded
     # content.
-    download.safe_download(self.url, self.target_data_length - 4).close()
-    download.unsafe_download(self.url, self.target_data_length - 4).close()
+    download.safe_download(self.url, self.target_data_length - 4, self.fetcher).close()
+    download.unsafe_download(self.url, self.target_data_length - 4, self.fetcher).close()
 
     # We catch 'tuf.exceptions.DownloadLengthMismatchError' for safe_download()
     # because it will not download more bytes than requested (in this case, a
     # length greater than the size of the target file).
     self.assertRaises(tuf.exceptions.DownloadLengthMismatchError,
-        download.safe_download, self.url, self.target_data_length + 1)
+        download.safe_download, self.url, self.target_data_length + 1, self.fetcher)
 
     # Calling unsafe_download() with a mismatched length should not raise an
     # exception.
-    download.unsafe_download(self.url, self.target_data_length + 1).close()
+    download.unsafe_download(self.url, self.target_data_length + 1, self.fetcher).close()
 
 
 
@@ -165,31 +170,29 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     unsafe_download_file = download.unsafe_download
 
     self.assertRaises(securesystemslib.exceptions.FormatError,
-                      download_file, None, self.target_data_length)
-
-    self.assertRaises(tuf.exceptions.URLParsingError,
-                      download_file,
-                      self.random_string(), self.target_data_length)
+                      download_file, None, self.target_data_length, self.fetcher)
 
     url = 'http://localhost:' \
         + str(self.server_process_handler.port) + '/' + self.random_string()
     self.assertRaises(requests.exceptions.HTTPError,
                       download_file,
                       url,
-                      self.target_data_length)
+                      self.target_data_length,
+                      self.fetcher)
     url1 = 'http://localhost:' \
       + str(self.server_process_handler.port + 1) + '/' + self.random_string()
     self.assertRaises(requests.exceptions.ConnectionError,
                       download_file,
                       url1,
-                      self.target_data_length)
+                      self.target_data_length,
+                      self.fetcher)
 
     # Specify an unsupported URI scheme.
     url_with_unsupported_uri = self.url.replace('http', 'file')
     self.assertRaises(requests.exceptions.InvalidSchema, download_file, url_with_unsupported_uri,
-                      self.target_data_length)
+                      self.target_data_length, self.fetcher)
     self.assertRaises(requests.exceptions.InvalidSchema, unsafe_download_file,
-                      url_with_unsupported_uri, self.target_data_length)
+                      url_with_unsupported_uri, self.target_data_length, self.fetcher)
 
 
 
@@ -290,7 +293,7 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
       os.environ['REQUESTS_CA_BUNDLE'] = bad_cert_fname
       # Clear sessions to ensure that the certificate we just specified is used.
       # TODO: Confirm necessity of this session clearing and lay out mechanics.
-      tuf.download._sessions = {}
+      self.fetcher._sessions = {}
 
       # Try connecting to the server process with the bad cert while trusting
       # the bad cert. Expect failure because even though we trust it, the
@@ -302,9 +305,9 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
             category=urllib3.exceptions.SubjectAltNameWarning)
 
         with self.assertRaises(requests.exceptions.SSLError):
-          download.safe_download(bad_https_url, target_data_length)
+          download.safe_download(bad_https_url, target_data_length, self.fetcher)
         with self.assertRaises(requests.exceptions.SSLError):
-          download.unsafe_download(bad_https_url, target_data_length)
+          download.unsafe_download(bad_https_url, target_data_length, self.fetcher)
 
         # Try connecting to the server processes with the good certs while not
         # trusting the good certs (trusting the bad cert instead). Expect failure
@@ -312,31 +315,31 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
         # trust it.
         logger.info('Trying HTTPS download of target file: ' + good_https_url)
         with self.assertRaises(requests.exceptions.SSLError):
-          download.safe_download(good_https_url, target_data_length)
+          download.safe_download(good_https_url, target_data_length, self.fetcher)
         with self.assertRaises(requests.exceptions.SSLError):
-          download.unsafe_download(good_https_url, target_data_length)
+          download.unsafe_download(good_https_url, target_data_length, self.fetcher)
 
         logger.info('Trying HTTPS download of target file: ' + good2_https_url)
         with self.assertRaises(requests.exceptions.SSLError):
-          download.safe_download(good2_https_url, target_data_length)
+          download.safe_download(good2_https_url, target_data_length, self.fetcher)
         with self.assertRaises(requests.exceptions.SSLError):
-          download.unsafe_download(good2_https_url, target_data_length)
+          download.unsafe_download(good2_https_url, target_data_length, self.fetcher)
 
 
         # Configure environment to now trust the certfile that is expired.
         os.environ['REQUESTS_CA_BUNDLE'] = expired_cert_fname
         # Clear sessions to ensure that the certificate we just specified is used.
         # TODO: Confirm necessity of this session clearing and lay out mechanics.
-        tuf.download._sessions = {}
+        self.fetcher._sessions = {}
 
         # Try connecting to the server process with the expired cert while
         # trusting the expired cert. Expect failure because even though we trust
         # it, it is expired.
         logger.info('Trying HTTPS download of target file: ' + expired_https_url)
         with self.assertRaises(requests.exceptions.SSLError):
-          download.safe_download(expired_https_url, target_data_length)
+          download.safe_download(expired_https_url, target_data_length, self.fetcher)
         with self.assertRaises(requests.exceptions.SSLError):
-          download.unsafe_download(expired_https_url, target_data_length)
+          download.unsafe_download(expired_https_url, target_data_length, self.fetcher)
 
 
         # Try connecting to the server processes with the good certs while
@@ -346,18 +349,18 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
         os.environ['REQUESTS_CA_BUNDLE'] = good_cert_fname
         # Clear sessions to ensure that the certificate we just specified is used.
         # TODO: Confirm necessity of this session clearing and lay out mechanics.
-        tuf.download._sessions = {}
+        self.fetcher._sessions = {}
         logger.info('Trying HTTPS download of target file: ' + good_https_url)
-        download.safe_download(good_https_url, target_data_length).close()
-        download.unsafe_download(good_https_url, target_data_length).close()
+        download.safe_download(good_https_url, target_data_length, self.fetcher).close()
+        download.unsafe_download(good_https_url, target_data_length,self.fetcher).close()
 
         os.environ['REQUESTS_CA_BUNDLE'] = good2_cert_fname
         # Clear sessions to ensure that the certificate we just specified is used.
         # TODO: Confirm necessity of this session clearing and lay out mechanics.
-        tuf.download._sessions = {}
+        self.fetcher._sessions = {}
         logger.info('Trying HTTPS download of target file: ' + good2_https_url)
-        download.safe_download(good2_https_url, target_data_length).close()
-        download.unsafe_download(good2_https_url, target_data_length).close()
+        download.safe_download(good2_https_url, target_data_length, self.fetcher).close()
+        download.unsafe_download(good2_https_url, target_data_length, self.fetcher).close()
 
     finally:
       for proc_handler in [
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -43,7 +43,7 @@ import warnings
 
 import tuf
 import tuf.download as download
-import tuf.fetcher
+import tuf.requests_fetcher
 import tuf.log
 import tuf.unittest_toolbox as unittest_toolbox
 import tuf.exceptions
@@ -87,7 +87,7 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     self.target_hash = {'md5':digest}
 
     # Initialize the default fetcher for the download
-    self.fetcher = tuf.fetcher.RequestsFetcher()
+    self.fetcher = tuf.requests_fetcher.RequestsFetcher()
 
 
 
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -112,6 +112,29 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
       self.assertEqual(self.target_data_length, len(temp_file_data))
 
 
+  # Test: Download url in more than one chunk.
+  def test_download_url_in_chunks(self):
+
+    # Set smaller chunk size to ensure that the file will be downloaded
+    # in more than one chunk
+    default_chunk_size = tuf.settings.CHUNK_SIZE
+    tuf.settings.CHUNK_SIZE = 4
+    # We don't have access to chunks from  download_file()
+    # so we just confirm that the expectation of more than one chunk is
+    # correct and verify that no errors are raised during download
+    chunks_count = self.target_data_length/tuf.settings.CHUNK_SIZE
+    self.assertGreater(chunks_count, 1)
+
+    download_file = download.safe_download
+    with download_file(self.url, self.target_data_length, self.fetcher) as temp_fileobj:
+      temp_fileobj.seek(0)
+      temp_file_data = temp_fileobj.read().decode('utf-8')
+      self.assertEqual(self.target_data, temp_file_data)
+      self.assertEqual(self.target_data_length, len(temp_file_data))
+
+    # Restore default settings
+    tuf.settings.CHUNK_SIZE = default_chunk_size
+
 
   # Test: Incorrect lengths.
   def test_download_url_to_tempfileobj_and_lengths(self):
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -77,7 +77,7 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     self.server_process_handler = utils.TestServerProcess(log=logger)
 
     rel_target_filepath = os.path.basename(target_filepath)
-    self.url = 'http://localhost:' \
+    self.url = 'http://127.0.0.1:' \
         + str(self.server_process_handler.port) + '/' + rel_target_filepath
 
     # Computing hash of target file data.
@@ -172,14 +172,14 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     self.assertRaises(securesystemslib.exceptions.FormatError,
                       download_file, None, self.target_data_length, self.fetcher)
 
-    url = 'http://localhost:' \
+    url = 'http://127.0.0.1:' \
         + str(self.server_process_handler.port) + '/' + self.random_string()
     self.assertRaises(requests.exceptions.HTTPError,
                       download_file,
                       url,
                       self.target_data_length,
                       self.fetcher)
-    url1 = 'http://localhost:' \
+    url1 = 'http://127.0.0.1:' \
       + str(self.server_process_handler.port + 1) + '/' + self.random_string()
     self.assertRaises(requests.exceptions.ConnectionError,
                       download_file,
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -192,23 +192,19 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     download_file = download.safe_download
     unsafe_download_file = download.unsafe_download
 
-    self.assertRaises(securesystemslib.exceptions.FormatError,
-                      download_file, None, self.target_data_length, self.fetcher)
+    with self.assertRaises(securesystemslib.exceptions.FormatError):
+      download_file(None, self.target_data_length, self.fetcher)
 
     url = 'http://127.0.0.1:' \
         + str(self.server_process_handler.port) + '/' + self.random_string()
-    self.assertRaises(requests.exceptions.HTTPError,
-                      download_file,
-                      url,
-                      self.target_data_length,
-                      self.fetcher)
+    with self.assertRaises(tuf.exceptions.FetcherHTTPError) as cm:
+      download_file(url, self.target_data_length, self.fetcher)
+    self.assertEqual(cm.exception.status_code, 404)
+
     url1 = 'http://127.0.0.1:' \
       + str(self.server_process_handler.port + 1) + '/' + self.random_string()
-    self.assertRaises(requests.exceptions.ConnectionError,
-                      download_file,
-                      url1,
-                      self.target_data_length,
-                      self.fetcher)
+    with self.assertRaises(requests.exceptions.ConnectionError):
+      download_file(url1, self.target_data_length, self.fetcher)
 
     # Specify an unsupported URI scheme.
     url_with_unsupported_uri = self.url.replace('http', 'file')
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -77,7 +77,7 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     self.server_process_handler = utils.TestServerProcess(log=logger)
 
     rel_target_filepath = os.path.basename(target_filepath)
-    self.url = 'http://127.0.0.1:' \
+    self.url = 'http://' + utils.TEST_HOST_ADDRESS + ':' \
         + str(self.server_process_handler.port) + '/' + rel_target_filepath
 
     # Computing hash of target file data.
@@ -195,13 +195,13 @@ class TestDownload(unittest_toolbox.Modified_TestCase):
     with self.assertRaises(securesystemslib.exceptions.FormatError):
       download_file(None, self.target_data_length, self.fetcher)
 
-    url = 'http://127.0.0.1:' \
+    url = 'http://' + utils.TEST_HOST_ADDRESS + ':' \
         + str(self.server_process_handler.port) + '/' + self.random_string()
     with self.assertRaises(tuf.exceptions.FetcherHTTPError) as cm:
       download_file(url, self.target_data_length, self.fetcher)
     self.assertEqual(cm.exception.status_code, 404)
 
-    url1 = 'http://127.0.0.1:' \
+    url1 = 'http://' + utils.TEST_HOST_ADDRESS + ':' \
       + str(self.server_process_handler.port + 1) + '/' + self.random_string()
     with self.assertRaises(requests.exceptions.ConnectionError):
       download_file(url1, self.target_data_length, self.fetcher)
--- a/tests/test_fetcher.py
+++ b/tests/test_fetcher.py
@@ -55,7 +55,7 @@ class TestFetcher(unittest_toolbox.Modified_TestCase):
     self.server_process_handler = utils.TestServerProcess(log=logger)
 
     rel_target_filepath = os.path.basename(target_filepath)
-    self.url = 'http://localhost:' \
+    self.url = 'http://127.0.0.1:' \
         + str(self.server_process_handler.port) + '/' + rel_target_filepath
 
     # Create a temporary file where the target file chunks are written
--- a/tests/test_fetcher.py
+++ b/tests/test_fetcher.py
@@ -58,7 +58,7 @@ class TestFetcher(unittest_toolbox.Modified_TestCase):
     self.server_process_handler = utils.TestServerProcess(log=logger)
 
     rel_target_filepath = os.path.basename(target_filepath)
-    self.url = 'http://127.0.0.1:' \
+    self.url = 'http://' + utils.TEST_HOST_ADDRESS + ':' \
         + str(self.server_process_handler.port) + '/' + rel_target_filepath
 
     # Create a temporary file where the target file chunks are written
--- a/tests/test_fetcher.py
+++ b/tests/test_fetcher.py
@@ -30,7 +30,7 @@ import math
 
 import tuf
 import tuf.exceptions
-import tuf.fetcher
+import tuf.requests_fetcher
 import tuf.unittest_toolbox as unittest_toolbox
 
 from tests import utils
@@ -64,7 +64,7 @@ class TestFetcher(unittest_toolbox.Modified_TestCase):
     # Create a temporary file where the target file chunks are written
     # during fetching
     self.temp_file = tempfile.TemporaryFile()
-    self.fetcher = tuf.fetcher.RequestsFetcher()
+    self.fetcher = tuf.requests_fetcher.RequestsFetcher()
 
 
   # Stop server process and perform clean up.
--- a/tests/test_fetcher.py
+++ b/tests/test_fetcher.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+
+"""
+<Program>
+  test_download.py
+
+<Author>
+  Teodora Sechkova tsechkova@vmware.com
+
+<Started>
+  December 18, 2020.
+
+<Copyright>
+  See LICENSE-MIT OR LICENSE for licensing information.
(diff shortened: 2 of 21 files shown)


sechkova:
The new GitHub actions have been harsh on me :)) Will review the errors after the holiday break.

sechkova:
> The new GitHub actions have been harsh on me :)) Will review the errors after the holiday break.

Fixed the linter issues, however according to Windows "Download was too slow." :roll_eyes:

sechkova:
Fixed a small issue and force-pushed but (un)fortunately I see no errors while testing on a local windows VM.

sechkova:
First thing, apologize for the spam and using the CI as a test server but I could not reproduce the Windows tests failure locally.

The problem can be summarized in two parts:

1. Slow 'localhost' lookup on Windows #1257
2. A change in this PR which adds the time to connect to the server to the calculation of `average_download_speed` for the first downloaded chunk which raises `SlowRetreivalError` in the case of 1.

While 1. can quickly be fixed as suggested by @jku in #1257, 2. is related to the network abstraction api which may need some discussion. The main point is, given all the uncertainty around protection against slow retrieval attack and its recent removal from the specification, where should we implement the slow retrieval check:

1. Keep the fetcher unaware of it and implement the download speed calculations in its caller function.
   - has the obvious pros that fetcher implementation is unaware of it and the check is performed by TUF itself
   - cons: incorrectly adds connection time to the actual chunk download time
2. Keep the fetcher unaware of it but make fetcher API more complicated by splitting session establishment/connection and downloading data
    - pros: solves the timing issue
    - cons: undesirably complicated api
3. Move the download speed calculation and slow retrieval checks as part of TUF's fetcher implementation
    - pros: replicates current behaviour
    - cons: different fetcher implementations can(and most likely will) totally ignore the slow retrieval check

My choice is in favour of point 3 but I would like to hear more opinions before going ahead.

lukpueh:
Thanks for the detailed comment, @sechkova (and of course for the patch, which I haven't looked at yet). Here are my thoughts:
- I'd rule out (2). This attack, which is no longer part of the spec (https://github.com/theupdateframework/specification/pull/111), does not seem to justify a more complicated API.
- I think (3) is perfectly fine, maybe we can add a big disclaimer to the API docs that says that only the default fetcher provides this protection, and any custom implementation must do it themselves.
- Regardless, wouldn't it be possible to achieve (1) by granting some grace period in the beginning, where the fetcher may drop below the rate? AFAIU a successful slow-retrieval attack requires quite some time, so if we start checking the average rate only after enough time has passed to likely smooth out the spike caused by connection/session establishment, we are probably able to detect the attack early enough.

trishankatdatadog:
I agree with @lukpueh: (1) makes more sense to me, depending on the philosophical motivation for the fetcher. Are we simply trying to fetch bytes or do something more?

jku:
I lean towards option 3 (mostly because I do not believe our slow retrieval protection works without false positives -- Teodoras test failures are a strange example of this IMO -- and I would like to not have these false positives in pip)

> wouldn't it be possible to achieve (1) by granting some grace period in the beginning, where the fetcher may drop below the rate? AFAIU a successful slow-retrieval attack requires quite some time, so if we start checking the average rate only after enough time has passed to likely smooth out the spike caused by connection/session establishment, we are probably able to detect the attack early enough.

I originally suggested the same thing, as this might also help with the massive bias that the current implementation has. Example of the bias: Imagine a download of 10 chunks where 9 chunks transfer at "normal internet speeds" but one is slow:  
* if the slow chunk is the first one the download will be canceled if speed is < 50B/s
* if the slow chunk is the last one the download will happily succeed even if the chunk speed is 1 B/s (a very long time later but it will succeed)

However, solving the above issue issue properly is no doubt complex -- probably the best we could do is special case the first chunk handling, just giving it a few extra seconds... I don't see why this is better than letting the fetcher handle it.

jku:
Oh additional reason for choice 3 (that Teodora kind of mentioned): the patch will probably be cleaner as it will be easier to just move existing code from download.py to Fetcher instead of developing similar but not quite same code.

jku:
> I agree with @lukpueh: (1) makes more sense to me, depending on the philosophical motivation for the fetcher. Are we simply trying to fetch bytes or do something more?

The issue is that the first call to the fetch() generator also creates the connection because there is no separate initialization function... now that I said it aloud, I think it has to be possible to initialize the generator separately. I'll look this up. Thanks.

joshuagl:
I am in favour of option three, both because we don't have a specification for slow retrieval attacks, and because we are not confident about the effectiveness of the current implementation of slow retrieval attacks.

> First thing, apologize for the spam and using the CI as a test server but I could not reproduce the Windows tests failure locally.

I didn't receive any spam about CI failures, so I think no need to apologise here 😄

sechkova:
Thanks all for giving your opinion and especially @jku for the ideas, most of which I think are included in the latest changes. 
Feel free to review :wink:

tuf/fetcher.py
```diff
@@ -113,12 +113,11 @@ def fetch(self, url, required_length):
     # Always set the timeout. This timeout value is interpreted by requests as:
     #  - connect timeout (max delay before first byte is received)
     #  - read (gap) timeout (max delay between bytes received)
-    with session.get(url, stream=True,
-        timeout=self.SOCKET_TIMEOUT) as response:
-
-      # Check response status.
-      response.raise_for_status()
+    response = session.get(url, stream=True, timeout=self.SOCKET_TIMEOUT)
```
joshuagl: As there's no longer a context manager in use for `response`, do we need to call some cleanup method?

jku: also maybe double check what happens when connect times out. I expect that's a ReadTimeoutError which should be handled

tuf/fetcher.py
```diff
@@ -0,0 +1,201 @@
...
+      data from a server
+
+    <Returns>
+      A bytes iterator
+    """
+    raise NotImplementedError # pragma: no cover
+
+
+
+class FetcherSettings():
```
jku: Can you explain the construct? I can see it should work but what is the advantage of a separate FetcherSettings and RequestsFetcher?

tuf/settings.py
```diff
@@ -74,12 +74,6 @@
 # download Targets metadata.
 DEFAULT_TARGETS_REQUIRED_LENGTH = 5000000 #bytes
 
-# Set a timeout value in seconds (float) for non-blocking socket operations.
-SOCKET_TIMEOUT = 4 #seconds
-
-# The maximum chunk of data, in bytes, we would download in every round.
-CHUNK_SIZE = 400000 #bytes
-
```
jku: These are settings that unfortunately maybe used by someone so I don't think we should just silently delete them without deprecation.

I wonder if the new implementation should still use these as the default values for its properties... and possibly mark these as deprecated, explaining that these values are now fetcher specific and that the default RequestsFetcher exposes the same properties (but that those properties might not be considered API(?) and those changing them should consider making their own fetcher)

tuf/client/updater.py
```diff
@@ -659,6 +659,10 @@ def __init__(self, repository_name, repository_mirrors):
                                           'targets_path': 'targets',
                                           'confined_target_dirs': ['']}}
 
+      fetcher:
+        A conctrete 'FetcherInterface' implementation. If not provided,
+        tuf.fetcher.RequestsFetcher is used.
```
jku: typo on concrete. Also it's lacking a sentence on what fetcher does.

jku (commented):
I like it except for the fact that it rewrites download.py completely :) I understand it's mostly an indent change but I was really hoping that wasn't needed.

I have not yet reviewed download.py changes -- I will do that but I'll leave an early review for this question: _What compromises would we need to make to keep download.py changes to a minimum?_

tuf/client/updater.py
sechkova: Fixed in b0e8acb

tuf/settings.py
sechkova: I totally forgot about `settings` being part of the API. I dropped the commit and just used `tuf.settings` in `RequestsFetcher`. 

I suggest we revise `tuf.settings` in general and then apply the decisions here.

tuf/fetcher.py
sechkova: The idea was to separate `FetcherSettings` so that other `FetcherInterface` implementations can reuse them (by inheriting) or replace them with their own. But given that we cannot remove them from `tuf.settings` just like that, I dropped this change too.

tuf/fetcher.py
sechkova: Did some cleanup in 77900af.

Connect time out will raise an exception that will be captured in `NoWorkingMirrorError`. This is the same as the current behaviour.

sechkova:
Reworked some changes after discussion with @jku. Download.py should be more appealing for a review now!

tests/test_fetcher.py
```diff
@@ -0,0 +1,119 @@
...
+  # Test: Normal case.
+  def test_fetch(self):
+
+    for chunk in self.fetcher.fetch(self.url, self.file_length):
+        self.temp_file.write(chunk)
+
+    self.temp_file.seek(0)
+    temp_file_data = self.temp_file.read().decode('utf-8')
+    self.assertEqual(self.file_contents, temp_file_data)
+    self.assertEqual(self.file_length, len(temp_file_data))
```
MVrachev: Do you need to check that when you know that the `self.file_contents` is equal to `temp_file_data` from the previous `assertEqual`?

tests/test_fetcher.py
```diff
@@ -0,0 +1,119 @@
...
+    for chunk in self.fetcher.fetch(self.url, self.file_length+4):
+        self.temp_file.write(chunk)
+
+    self.temp_file.seek(0, io.SEEK_END)
+    self.assertEqual(self.temp_file.tell(), self.file_length)
+
+
+  # Test incorrect URL parsing
+  def test_url_parsing(self):
+   with self.assertRaises(tuf.exceptions.URLParsingError) as cm:
```
MVrachev: What exactly will happen if this test fails? What about when it succeeds?  
I just haven't seen the usage of a context manager with an exception...

jku:
```
with self.assertRaises(ExpectedError):
  function_under_test(argument1, argument2)
```
is functionally the same as
```
self.assertRaises(ExpectedError, function_under_test, argument1, argument2)
```
but more readable especially with longer names and more arguments, and also more extensible (as you can have more than one line in the block as test_url_parsing() does).

The `as cm` seems unneeded (but harmless) in this case as it's not used.

sechkova: > but more readable especially with longer names and more arguments, and also more extensible (as you can have more than one line in the block as test_url_parsing() does).

This is what I've meant when writing it but I guess I had plans to use `cm` that didn't happen ...

tests/test_fetcher.py
sechkova: Nope. Thanks for noting, seems like I'll have to take another look on the test cases.

tuf/download.py
```diff
@@ -210,174 +191,46 @@ def _download_file(url, required_length, STRICT_REQUIRED_LENGTH=True):
...
-      number_of_bytes_received = number_of_bytes_received + len(data)
-
-      # Data successfully read from the response.  Store it.
-      temp_file.write(data)
-
-      if number_of_bytes_received == required_length:
-        break
+    chunks = fetcher.fetch(url, required_length)
+    start_time = timeit.default_timer()
+    for chunk in chunks:
```
jku, line 201: This is not totally Fetcher related but maybe should be fixed now that we're making changes: I think we never test a download with more than 1 chunk. This is not ideal especially now that a download loop exists in both fetcher and download.py.

Should one or both cases ensure that multiple chunks are tested, maybe by lowering the chunk size to something small or creating a large target file dynamically? Small chunk size is probably a better option (although artificially small chunk size might lead to SlowRetrieval checks being even more unreliable...)

tuf/fetcher.py
```diff
@@ -0,0 +1,188 @@
...
+    session = self._get_session(url)
+
+    # Get the requests.Response object for this URL.
+    #
+    # Defer downloading the response body with stream=True.
+    # Always set the timeout. This timeout value is interpreted by requests as:
+    #  - connect timeout (max delay before first byte is received)
+    #  - read (gap) timeout (max delay between bytes received)
+    response = session.get(url, stream=True,
+        timeout=tuf.settings.SOCKET_TIMEOUT)
```
jku: At some point I think we want to do something about the settings here (like deprecate the ones in settings module and make them just properties in the fetcher implementation) -- but I think this is the correct approach for this PR

tuf/fetcher.py
```diff
@@ -0,0 +1,188 @@
...
+          if not data:
+            logger.debug('Downloaded ' + repr(bytes_received) + '/' +
+              repr(required_length) + ' bytes.')
+
+            # Finally, we signal that the download is complete.
+            break
+
+          yield data
+
+          if bytes_received == required_length:
```
jku: this is copied from old implementation but...
the defensive version of this check would be `if bytes_received >= required_length:` -- there should be no harm in using it. download.py already ensures that the length matches so this would just make sure the response is closed etc.

tuf/fetcher.py
```diff
@@ -0,0 +1,188 @@
...
+    # Defer downloading the response body with stream=True.
+    # Always set the timeout. This timeout value is interpreted by requests as:
+    #  - connect timeout (max delay before first byte is received)
+    #  - read (gap) timeout (max delay between bytes received)
+    response = session.get(url, stream=True,
+        timeout=tuf.settings.SOCKET_TIMEOUT)
+    # Check response status.
+    response.raise_for_status()
+
+    def chunks():
```
jku: could add a comment describing the trick here: chunks is the actual iterator that gets returned from fetch() after response object has been initialized (and most importantly after connect() has happened)

tuf/fetcher.py
```diff
@@ -0,0 +1,188 @@
...
+      # Attach some default headers to every Session.
+      requests_user_agent = session.headers['User-Agent']
+      # Follows the RFC: https://tools.ietf.org/html/rfc7231#section-5.5.3
+      tuf_user_agent = 'tuf/' + tuf.__version__ + ' ' + requests_user_agent
+      session.headers.update({
+          # Tell the server not to compress or modify anything.
+          # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#Directives
+          'Accept-Encoding': 'identity',
+          # The TUF user agent.
+          'User-Agent': tuf_user_agent})
```
jku: user-agent could be a property of the fetcher -- but this makes sense to do in a later PR

jku (commented):
This is very much what I imagined it would be, thanks! I'm documenting my understanding here, do fix if there are mistakes:
* the issue with connect-time being counted in the first chunk download time was solved by separating the iterator initialization and first chunk download: now it operates just like the current code in this regard
* practical results should be identical to current functionality (and no tests were changed except for interface changes mentioned below and the unrelated localhost bugfix)
* The effects on TUF API are:
  - updater.py API is backwards-compatible (but has a new optional fetcher argument)
  - fetcher.py was added: FetcherInterface is now part of client API. A default implementation is provided and other implementations can be used
  - download.py was considered an implementation detail of updater.py and so download.py interface was modified: some functions now have a new required argument. The new argument could be made optional but that would lead to download.py (in the default case) to always instantiate a new fetcher, and thus new session which is not optimal. I guess it could also keep a "global default fetcher"  much like it now has a "global _sessions" but that seems like a worse idea
  - TUF no longer strictly depends on Requests (but the default fetcher does)

I left some comments in code. I've not had time to integrate this into my pip branch yet but from my POV this looks good to go (with the comments addressed).

tuf/fetcher.py
sechkova: I think for a later PR it is worth considering going even further and providing a way to [customise requests.Session](https://github.com/theupdateframework/tuf/issues/1188).

tuf/fetcher.py
sechkova: I was sure that `tuf.settings` has been discussed in terms of TUF 1.0.0 plans but I couldn't find anything documented. Will file an issue about it if nothing comes up.

sechkova:
> This is very much what I imagined it would be, thanks! I'm documenting my understanding here, do fix if there are mistakes:
> 

Thanks for the summary, @jku, I couldn't have described it better!
Comments are addressed in two new commits.

jku:
Thank you.

With regards to the new tests: I'm not against running 2MB of data through the code but writing so much to disk per test makes me a bit uncomfortable if there are alternatives... The other observation I have is that even the fetcher test never actually confirms that multiple chunks were used: we just assume requests is working like we expect. Maybe this is me being a bit paranoid. Thoughts on these?

jku:
Other discussion items (Teodora already has good ideas for these but I'm putting the questions out there for others to chime in): 

1. This PR currently suffers from the common issue of class & module name "conflict":
```
from tuf import fetcher
fetcher = fetcher.Fetcher() # oops name clash with variable and module
```
Basically the "logical" name for an instance of the class happens to be the name of the module.  This is not horrible and Updater is similar but if we can avoid it, it would be nice

2. The "API boundary" is as unclear as ever: would be nice if client developers had a clear boundary (e.g. "only files inside client/ are considered public client API")

sechkova:
> With regards to the new tests: I'm not against running 2MB of data through the code but writing so much to disk per test makes me a bit uncomfortable if there are alternatives... The other observation I have is that even the fetcher test never actually confirms that multiple chunks were used: we just assume requests is working like we expect. Maybe this is me being a bit paranoid. Thoughts on these?

I amended the last commit which adds tests with more than one chunk download (https://github.com/theupdateframework/tuf/pull/1250/commits/1e4b944310845974709f61ac84d6722159e56d9c). Instead of writing big files on disk, default chunk size is decreased. Let me know what do you think. 
I have doubts about the new test case in `test_download` since there is no visibility on the actual chunks from the `download_file` output so the test only confirms that file is downloaded without errors. It is harmless but is it also useless?

jku:
tests look good to me.

I found one real issue: Updater still uses HTTPError from requests (because it wants to know if error is 404 or 403 to decide if root download should error our or not). I think we need to define a HTTPError of our own and require the Fetcher to wrap any http errors in that. A bit annoying but easy to implement in the fetcher: let's experiment with this a bit before committing to a solution.

jku:
> I found one real issue: Updater still uses HTTPError from requests (because it wants to know if error is 404 or 403 to decide if root download should error our or not). I think we need to define a HTTPError of our own and require the Fetcher to wrap any http errors in that. A bit annoying but easy to implement in the fetcher: let's experiment with this a bit before committing to a solution.

I've tested this and the above is correct. The interface has to require fetch() to throw a specific exception on HTTP errors, otherwise we can't check for 403 & 404 in updater. The actual implementation work is minimal. I'll update this PR (as Teodora is out-of-office this week) hopefully tomorrow.

trishankatdatadog:
@jku Ah, right, I totally forgot about that, good catch

jku:
I'm using this same PR to add two changes. We've discussed these with Teodora but she has not seen the actual commits:
 * Add a custom HTTP error for Fetcher implementations to raise
 * Move the components around a bit (to make it clear what we consider client API)
 
The last point maybe deserves a definition: Through these changes we've had this idea:
* tuf.client contains the "client API"
* any other tuf.* modules are _from the client POV_ more or less implementation details

This is why there is a (minimal) module interface change in tuf.download in the PR and the last commit makes Fetcher components respect this idea as well.

There is no absolute "API boundary" defined in TUF at the moment but I think the above is a good goal to have for the client. To make this even clearer download.py, mirrors.py and requests_fetcher.py could be moved to e.g. tuf/client/_internal/ but I did not go that far in this PR.

jku:
Since it seems I did not report on the pip tests before this: 

I've tested this branch with pip: the custom fetcher works and is pleasant to implement, 15 LOC was enough to get pip progress bars working. From my POV this is ready (but I'm not adding an official review as I'm now an author as well).

tuf/fetcher.py
```diff
@@ -16,6 +16,16 @@
...
 
 import abc
+import requests
+import six
+import logging
+import time
+
+import urllib3.exceptions
+import tuf.exceptions
+import tuf.settings
```
joshuagl: Is this import style compatible with vendoring @jku? Or should we be doing `from foo import bar`?

tuf/fetcher.py
joshuagl: Let's be sure to file an issue to track this.

tuf/fetcher.py
joshuagl: We have an [issue in securesystemslib](https://github.com/secure-systems-lab/securesystemslib/issues/219), but not a corresponding issue in tuf. We should file one as part of the refactor project and milestones.

tuf/exceptions.py
```diff
@@ -335,3 +335,12 @@ class URLParsingError(Error):
 class InvalidConfigurationError(Error):
   """If a configuration object does not match the expected format."""
 
+class FetcherHTTPError(Exception):
+  """
+  Returned by FetcherInterface implementations for HTTP errors.
+  status_code property is the (int) HTTP status code
+  """
```
joshuagl, line 345: Can we document `message` and `status_code` here?

tests/test_download.py
```diff
@@ -118,18 +146,18 @@ def test_download_url_to_tempfileobj_and_lengths(self):
...
-    download.unsafe_download(self.url, self.target_data_length - 4).close()
+    download.safe_download(self.url, self.target_data_length - 4, self.fetcher).close()
+    download.unsafe_download(self.url, self.target_data_length - 4, self.fetcher).close()
 
     # We catch 'tuf.exceptions.DownloadLengthMismatchError' for safe_download()
     # because it will not download more bytes than requested (in this case, a
     # length greater than the size of the target file).
     self.assertRaises(tuf.exceptions.DownloadLengthMismatchError,
-        download.safe_download, self.url, self.target_data_length + 1)
+        download.safe_download, self.url, self.target_data_length + 1, self.fetcher)
```
joshuagl, line 156: should there be a `.close()` call on the downloaded file here, like the calls to `safe_download()` above and below?

tests/test_download.py
```diff
@@ -164,32 +192,26 @@ def test_download_url_to_tempfileobj_and_urls(self):
...
-                      download_file, None, self.target_data_length)
-
-    self.assertRaises(tuf.exceptions.URLParsingError,
-                      download_file,
-                      self.random_string(), self.target_data_length)
+    with self.assertRaises(securesystemslib.exceptions.FormatError):
+      download_file(None, self.target_data_length, self.fetcher)
 
-    url = 'http://localhost:' \
+    url = 'http://127.0.0.1:' \
```
joshuagl: Nit: a change like this in future might be easier if we define the local URL as a test level constant

tests/test_download.py
joshuagl: Ah, now I see that we close the file when we catch an exception. Hmm, I wonder if that's expected (anticipated) behaviour?

tests/test_fetcher.py
```diff
@@ -0,0 +1,146 @@
...
+  Teodora Sechkova tsechkova@vmware.com
+
+<Started>
+  December 18, 2020.
+
+<Copyright>
+  See LICENSE-MIT OR LICENSE for licensing information.
+
+<Purpose>
+  Unit test for 'fetcher.py'.
```
joshuagl: Specifically unit tests for requests_fetcher?

tests/test_fetcher.py
```diff
@@ -0,0 +1,146 @@
...
+import logging
+import os
+import io
+import sys
+import unittest
+import tempfile
+import math
+
+import tuf
+import tuf.unittest_toolbox as unittest_toolbox
```
joshuagl: Nit: should this be sorted alphabetically? `unittest_toolbox` below `requests_fetcher`

tests/test_proxy_use.py
```diff
@@ -202,8 +205,8 @@ def test_baseline_no_proxy(self):
     """
 
     logger.info('Trying HTTP download with no proxy: ' + self.url)
-    download.safe_download(self.url, self.target_data_length)
-    download.unsafe_download(self.url, self.target_data_length)
+    download.safe_download(self.url, self.target_data_length, self.fetcher)
```
joshuagl: do the `safe_download()` and `unsafe_download()` calls in test_proxy_use.py need to close the downloaded file?

tuf/download.py
```diff
@@ -128,6 +111,9 @@ def unsafe_download(url, required_length):
       An integer value representing the length of the file.  This is an upper
       limit.
 
+    fetcher:
+      A FetcherInterface object that performs the network IO operations.
```
joshuagl:
```suggestion
      An object implementing FetcherInterface that performs the network IO operations.
```

tuf/download.py
```diff
@@ -210,174 +190,46 @@ def _download_file(url, required_length, STRICT_REQUIRED_LENGTH=True):
...
+    # Does the total number of downloaded bytes match the required length?
+    _check_downloaded_length(number_of_bytes_received, required_length,
+                             STRICT_REQUIRED_LENGTH=STRICT_REQUIRED_LENGTH,
+                             average_download_speed=average_download_speed)
 
-        # Finally, we signal that the download is complete.
-        break
+  except Exception:
+    # Close 'temp_file'.  Any written data is lost.
+    temp_file.close()
```
joshuagl, line 228: Not strictly related to this PR, however – Is it odd that we close the file on an exception, but otherwise leave it up to the caller?

joshuagl (approved):
This looks really nice, great work @sechkova and @jku! I've made a few minor comments inline.

>I've tested this branch with pip: the custom fetcher works and is pleasant to implement, 15 LOC was enough to get pip progress bars working. From my POV this is ready (but I'm not adding an official review as I'm now an author as well).

15 LOC, that's quite satisfying. Really nice work all!

joshuagl:
Let's be sure to follow the new coding style in the new files, especially for the docstrings.

tuf/fetcher.py
jku: It is not compatible but I figured this will land before the imports (and since merge fixes are going to be needed anyway I'd change these in the imports PR where I change everything else as well)

tests/test_download.py
jku: In the exception case there is no (and can't be a) file to close at this point, I think this is correct (it just looks weird in the tests since we don't ever do anything with the file -- normally you'd use a `with download.safe_download() as f:` construct)

tests/test_proxy_use.py
jku: Yes I think they should. For some reason there are no warnings about unclosed files on these though...

tuf/download.py
jku: this is handled by Teodoras "Test downloading data in more than one chunk" commit

tuf/download.py
jku: Not in the new code because this code is now in _download_file() which opens the temp_file and returns it: the caller would have no way of closing the file in the exception case

tuf/fetcher.py
sechkova: Here it is  https://github.com/theupdateframework/tuf/issues/1273

tuf/fetcher.py
sechkova: Sorry, my link to the existing issue above was not working.
Do we need another one more specific to this PR's changes? I don't want to open an issue about something which has not yet landed in the develop branch.

joshuagl: We should probably update #1188 with recommendations for how to achieve the request within the new fetcher world, once this PR has landed.

tuf/exceptions.py
sechkova: Added some explanation in 7cd6177

tests/test_download.py
sechkova: A suggestion in 5d3133d.

tests/test_fetcher.py
sechkova: Fixed in be399d2.

tests/test_fetcher.py
sechkova: Fixed in be399d2.

tests/test_proxy_use.py
sechkova: Fixed in 8df22a9.

tuf/download.py
sechkova: Updated in 8b58f36.

sechkova:
Addressed all comments and suggestions (links to commits added since there was a rebase). Let me know if I am missing something or something else comes up.

tuf/download.py
```diff
@@ -192,33 +192,33 @@ def _download_file(url, required_length, fetcher, STRICT_REQUIRED_LENGTH=True):
...
   try:
-
-    for chunk in fetcher.fetch(url, required_length):
+    chunks = fetcher.fetch(url, required_length)
+    start_time = timeit.default_timer()
+    for chunk in chunks:
 
       stop_time = timeit.default_timer()
-      seconds_spent_receiving = stop_time - start_time
+      temp_file.write(chunk)
```
MVrachev, line 204: It seems I  am missing context, but why did you decide to move the write chunk to temp file before exiting the fetcher loop on error?

tuf/download.py
```diff
@@ -192,33 +192,33 @@ def _download_file(url, required_length, fetcher, STRICT_REQUIRED_LENGTH=True):
...
   number_of_bytes_received = 0
 
   try:
-
-    for chunk in fetcher.fetch(url, required_length):
+    chunks = fetcher.fetch(url, required_length)
+    start_time = timeit.default_timer()
+    for chunk in chunks:
 
       stop_time = timeit.default_timer()
```
MVrachev, line 203: I am wondering what exactly will happen between the `start_time` and `stop_time` given that they are just one after another.
Do `chunks` remain blocked until it's fully downloaded and then start the for loop (but that doesn't make sense, this would mean you have to have the whole file in memory which is what we want to escape)? 

Sorry for those probably obvious questions in this stage of the review process, but I wanted to make sure I understand this.

tuf/download.py
jku: I believe it's because the old code can lead to an exception even if all bytes have already been downloaded (which seems a bit useless). Teodoras implementation still warns in that case but does not throw.

tuf/download.py
jku: good question. the blocking points should be:
* `fetcher.fetch()` -- connect(), DNS lookup, TLS negotiation, etc
* each iteration of `chunks`: the iterator blocks until it has downloaded data for the chunk

this is the intention and "obvious implementation" at least: of course the Fetcher implementation could do things differently.

MVrachev: > fetcher.fetch() -- connect(), DNS lookup, TLS negotiation, etc
This means that the `start_time` won't catch the delay when starting the fetch process which makes sense

> each iteration of chunks: the iterator blocks until it has downloaded data for the chunk

This means it will count the average download on each chunk which seems right.
Thanks, I think I understand it now.

tuf/download.py
sechkova: In both cases writing the chunk to the temp_file is in the for loop. I moved it to be before the average download speed calculation and mostly before the `break` statement on line 216 since it was resulting in misleading errors on low average download speed as Jussi has explained.

sechkova: @MVrachev ah sorry, misread your question, so yeah ... what Jussi said :grin:

joshuagl:
Would it be possible to ensure the new files (tuf/client/fetcher.py and tuf/requests_fetcher.py) adhere to the new coding style? Of particular note is the docstring format.

sechkova:
> Would it be possible to ensure the new files (tuf/client/fetcher.py and tuf/requests_fetcher.py) adhere to the new coding style? Of particular note is the docstring format.

The docstrings are updated. I haven't applied the rules from the new coding style which cause linter errors. Since the changes in the PR are meant to work with the existing code they follow the old style. Let's review the coding style during the transition to TUF 1.0.0

tests/test_fetcher.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python
+
+"""
+<Program>
+  test_download.py
```
joshuagl: Nit:
```suggestion
  test_fetcher.py
```

Maybe we should be explicit and call this test_requests_fetcher?

tests/test_fetcher.py
```diff
@@ -0,0 +1,146 @@
...
+    # Cleans the resources and flush the logged lines (if any).
+    self.server_process_handler.clean()
+
+    self.target_fileobj.close()
+    self.temp_file.close()
+
+
+  # Test: Normal case.
+  def test_fetch(self):
+
```
joshuagl: Nit:
```suggestion
```

tests/test_fetcher.py
```diff
@@ -0,0 +1,146 @@
...
+    self.temp_file.seek(0)
+    temp_file_data = self.temp_file.read().decode('utf-8')
+    self.assertEqual(self.file_contents, temp_file_data)
+
+  # Test if fetcher downloads file up to a required length
+  def test_fetch_restricted_length(self):
+    for chunk in self.fetcher.fetch(self.url, self.file_length-4):
+        self.temp_file.write(chunk)
+
+    self.temp_file.seek(0, io.SEEK_END)
```
joshuagl, line 84: Nice, I didn't realise these constants were available in Python2.7.

tests/test_fetcher.py
```diff
@@ -0,0 +1,146 @@
...
+    self.assertEqual(self.temp_file.tell(), self.file_length)
+
+
+  # Test incorrect URL parsing
+  def test_url_parsing(self):
+    with self.assertRaises(tuf.exceptions.URLParsingError):
+      for chunk in self.fetcher.fetch(self.random_string(), self.file_length):
+          self.temp_file.write(chunk)
```
joshuagl: I assume the write never happens here? Do we need the for loop, or could we just call `self.fetcher.fetch()` to trigger the expected exception?

tests/test_fetcher.py
```diff
@@ -0,0 +1,146 @@
...
+    chunks_count = 0
+    for chunk in self.fetcher.fetch(self.url, self.file_length):
+      self.temp_file.write(chunk)
+      chunks_count+=1
+
+    self.temp_file.seek(0)
+    temp_file_data = self.temp_file.read().decode('utf-8')
+    self.assertEqual(self.file_contents, temp_file_data)
+    # Double check that we download more than one chunk
+    self.assertGreater(chunks_count, 1)
```
joshuagl: I think this check is unnecessary? We are checking for the expected number of chunks below, what additional value does this provide?

tuf/download.py
```diff
@@ -192,12 +180,6 @@ def _download_file(url, required_length, STRICT_REQUIRED_LENGTH=True):
   <Returns>
     A file object that points to the contents of 'url'.
   """
-
-  # Do all of the arguments have the appropriate format?
-  # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
-  securesystemslib.formats.URL_SCHEMA.check_match(url)
-  tuf.formats.LENGTH_SCHEMA.check_match(required_length)
```
joshuagl, line 199: Was removing the format checks for url and required_length intentional? I was half expecting them to have moved to the fetcher implementation.

joshuagl (approved):
A few minor comments, which you may choose to address or not, and a couple of questions. I will be happy to merge this in its current state once we've released v0.17.0 (#1284).

Thank you for the solid work on this!

tuf/download.py
sechkova: It is fully intentional. The call stack is `(un)safe_download` -> `_download_file` and the same checks already exist in `(un)safe_download`

joshuagl: Got it, thanks for clarifying!

tests/test_fetcher.py
sechkova: I reworked it a bit.

Does this line express my intention better:
https://github.com/theupdateframework/tuf/blob/c794e0b917ba79327277f2846eda0c9a005b6730/tests/test_fetcher.py#L122-L123

The goal is to make sure that the calculation of `expected_chunks` is correct and results in more than one chunk downloaded.

sechkova:
Addressed the latest comments and rebased on the latest changes from the development branch.

tests/test_fetcher.py
sechkova: Fixed the docstring but test_requests_fetcher sounds too verbose :)

tests/test_fetcher.py
joshuagl: Oh, I had confused variables. Apologies. 🤦 
The locality of the check makes the intention clearer, I don't think we should be using `assert()` over `self.assert*` here though.

sechkova: I did change the variable to make the test clearer 🙄 
Here it is, using  the unittest's "assert"  50b3b19

jku:
Had another look, I agree with my previous self: LGTM.

Notes:
* The whole thing is a mostly faithful reproduction of the earlier code, just in a RequestsFetcher.fetch() generator function
* requests_fetcher.py, download.py and mirrors.py are implementation details of updater.py: this is why download.py "API change" was seen as reasonable in the PR
* download.py does not provide much value anymore but I think keeping it is the correct call to keep the PR to a reasonable size
* This PR fixes the issues that I can see with pip integration (ability to tweak network stack tweakables, ability to do progress notifications, ability to compress metadata on the wire)
* The test server address fix could be more integrated (like `TestServerProcess.get_base_url()` that returns something like "http://127.0.0.1:1234/" but I think this is good enough for this PR)

There's an approval already so I'll merge this soon if no complaints appear.

lukpueh:
Thanks for all the efforts you put into this PR, @sechkova and @jku! And kudos to the reviewers, @joshuagl and @MVrachev. ❤️ 

I won't add any meaningful comments here, but based on a quick glance through the diff and all the other comments and discussions above, I'm sure the approval is well justified. :)

Issue #1246, opened 2020-12-17
GitHub Actions: Consider hardening setup against CVE-2020-15228


lukpueh:
**Description of issue or feature request**:
#1242 added a GitHub Actions based CI workflow . GitHub Actions has a moderate injection vulnerability (CVE-2020-15228) against which we should consider hardening our setup.

**Current behavior**:
No particular hardening against CVE-2020-15228

**Expected behavior**:
Assess necessity and harden against CVE-2020-15228.

@jku proposes a strategy in https://github.com/theupdateframework/tuf/issues/1195#issuecomment-744600836 also described in https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

joshuagl:
Additional hardening we might consider is in  the actions we use.

AFAICT an Action release on the marketplace corresponds to a tag in the repository that publishes the action, which leaves users open to malicious maintainer attacks - tags are not immutable, so a repo owner (or attacker) can replace the tag and point to a malicious version of the action.

GitHub allows us to restrict actions to either: only use actions defined in our org, only allow a predefined list of actions, or allow all actions.

lukpueh:
Good observation, @joshuagl! Here are some ideas to reduce the attack surface:

- [use the commit sha](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/finding-and-customizing-actions#using-shas) instead of the tag (with the downside of pinning full versions, whereas with tags you can also pin major or minor only)
- allow [only actions created/hosted by GitHub](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository#allowing-specific-actions-to-run) (and maybe popular verified 3rd-party actions)
- try to use custom shell scripts on the [`run:` keyword](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun) instead of non-verified 3rd-party actions where feasible

trishankatdatadog:
You guys are really thinking this through. Very impressive! 👏🏽

joshuagl:
Another reason to distrust 3rd-party Actions – they can push commits (and tags?) to the repository[[1](https://hikari.noyu.me/blog/2020-05-05-github-private-repos-access-control-violation.html)].

[...]

Issue #1247, opened 2020-12-17
GitHub Actions: Consider further caching in CI workflow


lukpueh:
**Description of issue or feature request**:
#1242 added a GitHub Actions based CI workflow, which configures basic pip caching.

**Current behavior**:
Basic pip caching in CI workflow

**Expected behavior**:
Explore and implement further caching to further speed up CI workflow (@jku suggests to try caching the tox directory).

lukpueh:
Also cache pip install in non-"test" jobs, e.g. "coveralls-fin" introduced in #1252. It would be nice to find a cross-job caching solution.

jku:
Leaving a note for later: Defining an action of our own in this same repo should allow using the same code across different jobs: Maybe it's not hard to define a "pip-install" action that sets up cache, updates pip and installs/updates the given list of packages.

jku:
> cache pip install in non-"test" jobs, e.g. "coveralls-fin" introduced in #1252

this one might be the only relevant thing left here as
* caching tox is not a good idea: tox is not able to handle some changes in dependencies, so we'd need to build something ourselves... not worth it
* setup-python now handles cache (#1718) so that's not longer ugly

h4l0gen:
Hi @jku @lukpueh, If this issue is available to take on, I would like to work on it.

jku:
this issue is old... There may be caching opportunities in the workflows but there is no check-list to go through, sorry.

lukpueh:
> > cache pip install in non-"test" jobs, e.g. "coveralls-fin" introduced in [#1252](https://github.com/theupdateframework/python-tuf/pull/1252)
> 
> this one might be the only relevant thing left here as

"coveralls-fin" Python installs are cached since https://github.com/theupdateframework/python-tuf/pull/2173/files

IMO GHA builds are swift enough to close here.

PR #1252, opened 2020-12-23
Reinstate publishing coverage on coveralls.io

### LINKED ISSUE #1243: GitHub Actions: Reinstate coveralls.io integration
A recent switch from Travis (and Appveyor) to GitHub Actions removed coveralls.io integration from our CI setup. (see #1242)

**Current behavior**:
No publication of coverage data on coveralls.io
No coveralls.io badge 

**Expected behavior**:
Re-instate publication of coverage data on coveralls.io (see WIP in lukpueh@df73242)
Re-add coveralls.io badge (revert 0ab9ee7)

lukpueh:
Fixes #1243

Re-instates publishing coverage on coveralls.io using the `coveralls` command line tool, which works better than the official, and a popular unofficial coveralls GitHub action and seems preferable for the security reasons outlined in #1246. 

This PR also configures the Python 2.7 runner to setup an additional 3.x service Python version to run tox and coveralls tools (only inside tox Python 2.7 is used).

See commit messages for more details.

requirements**:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,39 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install tox
+          pip install --upgrade tox coveralls
 
       - name: Run tox
         run: tox -e ${{ matrix.toxenv }}
+
+      - name: Publish on coveralls.io
+        # TODO: Maybe make 'lint' a separate job instead of case handling here
+        if: ${{ matrix.toxenv != 'lint' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COVERALLS_FLAG_NAME: ${{ runner.os }} / Python ${{ matrix.python-version }} / ${{ matrix.toxenv }}
+          COVERALLS_PARALLEL: true
+        # Use cp workaround to publish coverage reports with relative paths
+        # FIXME: Consider refactoring the tests to not require the test
+        # aggregation script being invoked from the `tests` directory, so
+        # that `.coverage` is written to and .coveragrc can also reside in
+        # the project root directory as is the convention.
+        run: |
+          cp tests/.coverage .
+          coveralls --service=github --rcfile=tests/.coveragerc
+
+  coveralls-fin:
+    # Always run when all 'build' jobs have finished even if they failed
+    # TODO: Replace always() with a 'at least one job succeeded' expression
+    if: always()
+    needs: build
+    runs-on: ubuntu-latest
+    container: python:3-slim
+    steps:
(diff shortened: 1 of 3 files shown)


jku:
This looks quite reasonable, but  now that I see the trouble you have because of python version being part of the GH matrix... I wonder if we should just use single python version and use explicit toxenv values as the "matrix axis"?

This should mean our "infra code" would always run on the know-good python (probably 3.8 currently) and the explicit toxenv would take care of running the tests on the correct python version... Would that simplify things?

lukpueh:
Hm. That is actually not such a bad idea and it might make the workflow file a little cleaner/clearer (no ugly ternary env var assignment). But it would also require setting up two Python versions on each run. Not sure how much of an overhead that is.

Do you think it's worth the changing now? Once we drop Python 2.7, we can get rid of the wonky case handling anyway, and in the Python 3 versions we should be fine with "infra python" == "test python"...

But let me quickly try what it looks like. :)

lukpueh:
Now that I'm trying, I'm less convinced. In the setup-python step I now need to infer the correct version "X.Y" from the toxenv string "pyXY", but not on "with-sslib-master" and "lint" runners. I think I don't want to spare the cycles to make this nicer right now, but I'm happy to review a patch if you want to submit one. ;)

jku:
Do we really need to know the version "X.Y" somewhere?  I can have a quick look later today -- maybe it's obvious once I try it

lukpueh:
> Do we really need to know the version "X.Y" somewhere?

From the [`setup-python` usage docs](https://github.com/actions/setup-python#usage):

```yaml
  with:
    python-version: '3.x' # Version range or exact version of a Python version to use, using SemVer's version range syntax
```
>  can have a quick look later today -- maybe it's obvious once I try it

Thanks!

jku:
Yes, I get it now: tox needs the correct system python to be installed. I'm sure there is a clever way to convert tox version to py version (or other way round) but we'd still need to have 2 setup-python steps so...

My only complaint now is that the coveralls-fin "pip install" probably does not use pip cache. We could fix that later (would be nice to figure out how to define the cache steps just once and not separately for coveralls-fin)...

jku (approved):

lukpueh:
> .. clever way to convert tox version to py version (or other way round)

Yeah, something like a step that runs `sed` and patches the toxenv variable or so ... 🤷 

> My only complaint now is that the coveralls-fin "pip install" probably does not use pip cache. We could fix that later (would be nice to figure out how to define the cache steps just once and not separately for coveralls-fin)...

Well spotted! I had made a (mental) note to look into this later with https://github.com/theupdateframework/tuf/issues/1247 exactly for that reason, i.e. code reuse. I'll make an actual note and merge here.

Thanks for the review!

PR #1259, opened 2021-01-11
Configure dependabot to ignore 'idna'

lukpueh:
Fixes -
Related to #1256 

New releases of the transitive (via 'requests') dependency 'idna' break Python 2.7 builds. To fix this we configure dependabot to not bump 'idna' in requirements-pinned.txt, which lists and auto-updates all immediate and transitive dependencies for CI/CD testing.

An alternative would be to add and restrict 'idna' in 'requirements.txt' but this is less preferable because 'requirements.txt' should only have direct dependencies.

requirements**:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,8 @@ updates:
     interval: daily
     time: "10:00"
   open-pull-requests-limit: 10
+  ignore:
+    # New 'idna' (see 'requests') releases break Python 2.7 builds. Ignore here
+    # to avoid listing/pinning transitive dependencies in requirements.txt.
+    # FIXME: Un-ignore when dropping Python 2.7 or resolving #1249
+    - dependency-name: "idna"


.github/dependabot.yml
```diff
@@ -6,3 +6,8 @@ updates:
     interval: daily
     time: "10:00"
   open-pull-requests-limit: 10
+  ignore:
```
joshuagl, line 9: The [docs](https://dependabot.com/docs/config-file/#ignored_updates) you linked to earlier suggest this should be `ignored_updates`? Did this change when Dependabot was integrated into GitHub?

.github/dependabot.yml
```diff
@@ -6,3 +6,8 @@ updates:
...
+    # New 'idna' (see 'requests') releases break Python 2.7 builds. Ignore here
+    # to avoid listing/pinning transitive dependencies in requirements.txt.
+    # FIXME: Un-ignore when dropping Python 2.7 or resolving #1249
+    - dependency-name: "idna"
```
joshuagl: Nit:
```suggestion
    - dependency-name: "idna"

```

.github/dependabot.yml
lukpueh: > Did this change when Dependabot was integrated into GitHub?

Exactly. I used the "Dependabot Preview" (legacy non-native version) web interface to PR the initial dependabot.yml (#1258), which also did the migration to GitHub-native, which seems to have a slightly different config syntax: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-and-disabling-version-updates#disabling-dependabot-version-updates

lukpueh:
Thanks for the quick review, @joshuagl! Just force-pushed an updated commit that adds the newline and the relevant docs in the commit message.

lukpueh:
> Just force-pushed an updated commit that adds the newline and the relevant docs in the commit message.

Sorry, forgot to stage the change. 🤦  Now the newline is indeed included.

joshuagl (approved):
LGTM, thanks @lukpueh !

Issue #1131, opened 2020-09-10
Add exception handling guidelines


lukpueh:
Which exceptions should be handled, which exceptions should be propagated to the user? When should we use custom exceptions, when should we use built-in exceptions? etc...

The Google Python style-guide has some reasonable [suggestions about exceptions](https://google.github.io/styleguide/pyguide.html#24-exceptions).

lukpueh:
See related #967

MVrachev:
Don't you think this issue needs a label `decision record `?
When you create guidelines for something used in multiple places like exceptions maybe it's a good idea to record it.

Same question for issues https://github.com/theupdateframework/tuf/issues/1132, https://github.com/theupdateframework/tuf/issues/1130, https://github.com/theupdateframework/tuf/issues/1129 and https://github.com/theupdateframework/tuf/issues/1125.

joshuagl:
I'm not sure if it's deserving of a separate label, but we should make it clear that the result of every discussion should be documented. Whether a decision record, contributor guideline or some other form.

lukpueh:
Good point @MVrachev. But how would the decision record about exception guidelines differ from the exception guideline document itself? I think the reason why I didn't use the label here is because I envisioned the guideline document itself a good central place that justifies the way exceptions are handled in multiple places, and probably this issue page as easily identifiable place to look up any additional discussion, that are not part of the guideline document.

That said, I do see how it's not always clear if something classifies for a decision record or not.  A general rule of thumb to classify something could be:

*"The decision is worth being tracked and there is no easily identifiable place (e.g. PR or issue page, commit message, code comment, other documentation or guideline, etc...) for it to be tracked at."*

What do you think? At any rate, we should make this clear as part of #1141.

lukpueh:
As development has slowed down, due to the maturity of the code base, this request has become a lot less relevant. I suggest to close until this becomes an issue again.

PR #1261, opened 2021-01-12
Vendoring-compatible imports

jku:
The goal here is to make TUF compatible with the vendoring tool so that tuf sources can be vendored into pip. As a reminder vendoring will modify all import-lines in a vendored project like this patch:

```
- import package 
- from package import module 
- from package.module import func
+ from _vendoring import package 
+ from _vendoring.package import module 
+ from _vendoring.package.module import func

```

This does not work in the case of `import package.module` that is so common in tuf.

Notes
* tuf modules are now imported with this style `from tuf import download`. This works with just a few localised issues: I'm generally happy with how code calling internal modules looks
* securesystemslib modules can't be handled this way:
  * module names clash with tuf modules names
  * module names are quite generic and likely to lead to conflicts later on
* as result securesystemslib imports are now like `from securesystemslib import formats as sslib_formats` in most cases. I don't particularly like it but it is readable and I can't think of a better solution
* Very minor changes to how other dependencies are used
* linter now understands our imports so some fixes and workarounds have been added. 
* The import lists that we're changed have been sorted so tuf imports are last
* The patch is surprisingly large, ~900 lines. Roughly 350 are needed for the client vendoring but I modified the whole code base to at least be consistent. The tests have not been touched.

### DIFF
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -143,7 +143,7 @@ from tuf import download
 from tuf import exceptions
 from tuf import formats
 from tuf import keydb
-from tuf import log
+from tuf import log # pylint: disable=unused-import
 from tuf import mirrors
 from tuf import roledb
 from tuf import settings
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -133,6 +133,7 @@ import io
 
 from securesystemslib import exceptions as sslib_exceptions
 from securesystemslib import formats as sslib_formats
+from securesystemslib import hash as sslib_hash
 from securesystemslib import keys as sslib_keys
 from securesystemslib import util as sslib_util
 
@@ -148,7 +149,6 @@ from tuf import sig
 import tuf.requests_fetcher
 import tuf.keydb
 
-import securesystemslib.hash
 import six
 
 # The Timestamp role does not have signed metadata about it; otherwise we
@@ -1207,7 +1207,7 @@ class Updater(object):
 
     # Verify each hash, raise an exception if any hash fails to verify
     for algorithm, trusted_hash in six.iteritems(trusted_hashes):
-      digest_object = securesystemslib.hash.digest_fileobject(file_object,
+      digest_object = sslib_hash.digest_fileobject(file_object,
           algorithm)
       computed_hash = digest_object.hexdigest()
 
@@ -2933,7 +2933,7 @@ class Updater(object):
     # Calculate the hash of the filepath to determine which bin to find the
     # target.  The client currently assumes the repository (i.e., repository
     # tool) uses 'hash_function' to generate hashes and UTF-8.
-    digest_object = securesystemslib.hash.digest(hash_function)
+    digest_object = sslib_hash.digest(hash_function)
     encoded_target_filepath = target_filepath.encode('utf-8')
     digest_object.update(encoded_target_filepath)
     target_filepath_hash = digest_object.hexdigest()
@@ -3088,7 +3088,7 @@ class Updater(object):
       for algorithm, digest in six.iteritems(target['fileinfo']['hashes']):
         digest_object = None
         try:
-          digest_object = securesystemslib.hash.digest_filename(target_filepath,
+          digest_object = sslib_hash.digest_filename(target_filepath,
             algorithm=algorithm)
 
         # This exception would occur if the target does not exist locally.
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -135,10 +135,10 @@ from tuf import download
 from tuf import exceptions
 from tuf import formats
 from tuf import log
+from tuf import mirrors
 import tuf.requests_fetcher
 import tuf.settings
 import tuf.keydb
-import tuf.mirrors
 import tuf.roledb
 import tuf.sig
 
@@ -1313,7 +1313,7 @@ class Updater(object):
       dirname, basename = os.path.split(target_filepath)
       target_filepath = os.path.join(dirname, target_digest + '.' + basename)
 
-    file_mirrors = tuf.mirrors.get_list_of_mirrors('target', target_filepath,
+    file_mirrors = mirrors.get_list_of_mirrors('target', target_filepath,
         self.mirrors)
 
     # file_mirror (URL): error (Exception)
@@ -1511,7 +1511,7 @@ class Updater(object):
       A file object containing the metadata.
     """
 
-    file_mirrors = tuf.mirrors.get_list_of_mirrors('meta', remote_filename,
+    file_mirrors = mirrors.get_list_of_mirrors('meta', remote_filename,
         self.mirrors)
 
     # file_mirror (URL): error (Exception)
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -136,10 +136,10 @@ from tuf import exceptions
 from tuf import formats
 from tuf import log
 from tuf import mirrors
+from tuf import roledb
 import tuf.requests_fetcher
 import tuf.settings
 import tuf.keydb
-import tuf.roledb
 import tuf.sig
 
 import securesystemslib.exceptions
@@ -767,7 +767,7 @@ class Updater(object):
 
     # Load current and previous metadata.
     for metadata_set in ['current', 'previous']:
-      for metadata_role in tuf.roledb.TOP_LEVEL_ROLES:
+      for metadata_role in roledb.TOP_LEVEL_ROLES:
         self._load_metadata_from_file(metadata_set, metadata_role)
 
     # Raise an exception if the repository is missing the required 'root'
@@ -914,7 +914,7 @@ class Updater(object):
     tuf.keydb.create_keydb_from_root_metadata(self.metadata['current']['root'],
         self.repository_name)
 
-    tuf.roledb.create_roledb_from_root_metadata(self.metadata['current']['root'],
+    roledb.create_roledb_from_root_metadata(self.metadata['current']['root'],
         self.repository_name)
 
 
@@ -983,11 +983,11 @@ class Updater(object):
     # Add the roles to the role database.
     for roleinfo in roles_info:
       try:
-        # NOTE: tuf.roledb.add_role will take care of the case where rolename
+        # NOTE: roledb.add_role will take care of the case where rolename
         # is None.
         rolename = roleinfo.get('name')
         logger.debug('Adding delegated role: ' + str(rolename) + '.')
-        tuf.roledb.add_role(rolename, roleinfo, self.repository_name)
+        roledb.add_role(rolename, roleinfo, self.repository_name)
 
       except exceptions.RoleAlreadyExistsError:
         logger.warning('Role already exists: ' + rolename)
@@ -2237,7 +2237,7 @@ class Updater(object):
     # Remove knowledge of the role.
     if metadata_role in self.metadata['current']:
       del self.metadata['current'][metadata_role]
-    tuf.roledb.remove_role(metadata_role, self.repository_name)
+    roledb.remove_role(metadata_role, self.repository_name)
 
 
 
@@ -2341,8 +2341,8 @@ class Updater(object):
     # Fetch the targets of the delegated roles.  get_rolenames returns
     # all roles available on the repository.
     delegated_targets = []
-    for role in tuf.roledb.get_rolenames(self.repository_name):
-      if role in tuf.roledb.TOP_LEVEL_ROLES:
+    for role in roledb.get_rolenames(self.repository_name):
+      if role in roledb.TOP_LEVEL_ROLES:
         continue
 
       else:
@@ -2477,7 +2477,7 @@ class Updater(object):
     targets_of_role = list(targets)
     logger.debug('Getting targets of role: ' + repr(rolename) + '.')
 
-    if not tuf.roledb.role_exists(rolename, self.repository_name):
+    if not roledb.role_exists(rolename, self.repository_name):
       raise exceptions.UnknownRoleError(rolename)
 
     # We do not need to worry about the target paths being trusted because
@@ -2579,7 +2579,7 @@ class Updater(object):
       self._refresh_targets_metadata(refresh_all_delegated_roles=True)
 
 
-    if not tuf.roledb.role_exists(rolename, self.repository_name):
+    if not roledb.role_exists(rolename, self.repository_name):
       raise exceptions.UnknownRoleError(rolename)
 
     return self._targets_of_role(rolename, skip_refresh=True)
@@ -2973,7 +2973,7 @@ class Updater(object):
 
     # Iterate the rolenames and verify whether the 'previous' directory
     # contains a target no longer found in 'current'.
-    for role in tuf.roledb.get_rolenames(self.repository_name):
+    for role in roledb.get_rolenames(self.repository_name):
       if role.startswith('targets'):
         if role in self.metadata['previous'] and self.metadata['previous'][role] != None:
           for target in self.metadata['previous'][role]['targets']:
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -132,6 +132,7 @@ import io
 
 import tuf
 from tuf import download
+from tuf import exceptions
 import tuf.requests_fetcher
 import tuf.formats
 import tuf.settings
@@ -140,7 +141,6 @@ import tuf.log
 import tuf.mirrors
 import tuf.roledb
 import tuf.sig
-import tuf.exceptions
 
 import securesystemslib.exceptions
 import securesystemslib.hash
@@ -208,7 +208,7 @@ class MultiRepoUpdater(object):
       self.map_file = securesystemslib.util.load_json_file(map_file)
 
     except (securesystemslib.exceptions.Error) as e:
-      raise tuf.exceptions.Error('Cannot load the map file: ' + str(e))
+      raise exceptions.Error('Cannot load the map file: ' + str(e))
 
     # Raise securesystemslib.exceptions.FormatError if the map file is
     # improperly formatted.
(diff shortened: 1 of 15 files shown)


tuf/keydb.py
```diff
@@ -344,7 +344,7 @@ def get_key(keyid, repository_name='default'):
     return copy.deepcopy(_keydb_dict[repository_name][keyid])
 
   except KeyError as error:
-    six.raise_from(tuf.exceptions.UnknownKeyError('Key: ' + keyid), error)
+    six.raise_from(exceptions.UnknownKeyError('Key: ' + keyid), error)
```
joshuagl, line 351: Looks like this was missed in the prior commit? 😃

tuf/developer_tool.py
```diff
@@ -251,12 +251,12 @@ def write(self, write_partial=False):
     # Raise 'securesystemslib.exceptions.FormatError' if any are improperly formatted.
     securesystemslib.formats.BOOLEAN_SCHEMA.check_match(write_partial)
 
-    # At this point the tuf.keydb and tuf.roledb stores must be fully
+    # At this point the tuf.keydb and roledb stores must be fully
```
joshuagl: should a prior commit have changed tuf.keydb -> keydb ?

tuf/developer_tool.py
```diff
@@ -310,7 +310,7 @@ def add_verification_key(self, key, expires=None):
         securesystemslib.exceptions.Error, if the project already contains a key.
 
       <Side Effects>
-        The role's entries in 'tuf.keydb.py' and 'tuf.roledb.py' are updated.
+        The role's entries in 'tuf.keydb.py' and 'roledb' are updated.
```
joshuagl: should a prior commit have changed tuf.keydb.py -> keydb ?

tuf/repository_lib.py
```diff
@@ -1195,7 +1195,7 @@ def generate_root_metadata(version, expiration_date, consistent_snapshot,
   repository_name='default'):
   """
   <Purpose>
-    Create the root metadata.  'tuf.roledb.py' and 'tuf.keydb.py'
+    Create the root metadata.  'roledb' and 'tuf.keydb.py'
```
joshuagl: tuf.keydb.py -> keydb, in a prior commit?

tuf/repository_lib.py
```diff
@@ -1224,10 +1224,10 @@ def generate_root_metadata(version, expiration_date, consistent_snapshot,
 
     securesystemslib.exceptions.Error, if an error is encountered while
     generating the root metadata object (e.g., a required top-level role not
-    found in 'tuf.roledb'.)
+    found in 'roledb'.)
 
   <Side Effects>
-    The contents of 'tuf.keydb.py' and 'tuf.roledb.py' are read.
+    The contents of 'tuf.keydb.py' and 'roledb' are read.
```
joshuagl: tuf.keydb.py -> keydb, in a prior commit?

tuf/repository_tool.py
```diff
@@ -314,7 +314,7 @@ def writeall(self, consistent_snapshot=False, use_existing_fileinfo=False):
     # formatted.
     securesystemslib.formats.BOOLEAN_SCHEMA.check_match(consistent_snapshot)
 
-    # At this point, tuf.keydb and tuf.roledb must be fully populated,
+    # At this point, tuf.keydb and roledb must be fully populated,
```
joshuagl: tuf.keydb -> keydb, in a prior commit?

tuf/repository_tool.py
```diff
@@ -721,8 +721,7 @@ def add_verification_key(self, key, expires=None):
       expired.
 
     <Side Effects>
-      The role's entries in 'tuf.keydb.py' and 'tuf.roledb.py' are
-      updated.
+      The role's entries in 'tuf.keydb.py' and 'roledb' are updated.
```
joshuagl: tuf.keydb.py -> keydb, in a prior commit?

tuf/repository_tool.py
```diff
@@ -883,7 +882,7 @@ def load_signing_key(self, key):
       securesystemslib.exceptions.Error, if the private key is not found in 'key'.
 
     <Side Effects>
-      Updates the role's 'tuf.keydb.py' and 'tuf.roledb.py' entries.
+      Updates the role's 'tuf.keydb.py' and 'roledb' entries.
```
joshuagl: tuf.keydb.py -> keydb, in a prior commit?

tuf/repository_tool.py
```diff
@@ -2353,7 +2352,7 @@ def delegate(self, rolename, public_keys, paths, threshold=1,
     <Side Effects>
       A new Target object is created for 'rolename' that is accessible to the
       caller (i.e., targets.<rolename>).  The 'tuf.keydb.py' and
```
joshuagl: tuf.keydb.py -> keydb, in a prior commit?

tuf/sig.py
```diff
@@ -71,14 +71,14 @@ def get_signature_status(signable, role=None, repository_name='default',
 
     * bad -- Invalid signature
     * good -- Valid signature from key that is available in 'tuf.keydb', and is
```
joshuagl, line 75: tuf.keydb -> keydb, in a prior commit?

tuf/sig.py
```diff
@@ -71,14 +71,14 @@ def get_signature_status(signable, role=None, repository_name='default',
...
-      authorized for the passed role as per 'tuf.roledb' (authorization may be
+      authorized for the passed role as per 'roledb' (authorization may be
       overwritten by passed 'keyids').
     * unknown -- Signature from key that is not available in 'tuf.keydb', or if
```
joshuagl, line 78: tuf.keydb -> keydb, in a prior commit?

tuf/sig.py
```diff
@@ -71,14 +71,14 @@ def get_signature_status(signable, role=None, repository_name='default',
...
       'role' is None.
     * unknown signing schemes -- Signature from key with unknown signing
       scheme.
     * untrusted -- Valid signature from key that is available in 'tuf.keydb',
```
joshuagl, line 82: tuf.keydb -> keydb, in a prior commit?

tuf/download.py
```diff
@@ -357,14 +358,14 @@ def _download_fixed_amount_of_data(response, temp_file, required_length):
       # Measure the average download speed.
       average_download_speed = number_of_bytes_received / seconds_spent_receiving
 
-      if average_download_speed < tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED:
+      if average_download_speed < settings.MIN_AVERAGE_DOWNLOAD_SPEED:
         logger.debug('The average download speed dropped below the minimum'
-          ' average download speed set in tuf.settings.py.')
+          ' average download speed set in tuf.settings.')
```
joshuagl: We're somewhat inconsistent across this PR on whether we refer to modules – in comments and log messages – as tuf.module, or just module. Let's choose one and stick to it.

tuf/client/updater.py
```diff
@@ -129,6 +129,8 @@
 import copy
 import warnings
 
+from securesystemslib import exceptions as sslib_exceptions
```
joshuagl, line 135: Given that the exceptions module is one of the few places in sslib where we have classes, wouldn't it be cleaner to import only the exceptions we want/need? Or do some of the conflicting exception names between sslib and tuf prevent this?

tuf/log.py
```diff
@@ -269,7 +271,7 @@ def set_console_log_level(log_level=_DEFAULT_CONSOLE_LOG_LEVEL):
       'log_level' examples: logging.INFO; logging.CRITICAL.
 
   <Exceptions>
-    securesystems.exceptions.Error, if the 'log.py' console handler has not
+    securesystemslib.exceptions.Error, if the 'log.py' console handler has not
```
joshuagl, line 273: log.py -> log?

tuf/scripts/repo.py
```diff
@@ -147,17 +147,19 @@
 import time
 import fnmatch
 
+# 'pip install securesystemslib[crypto,pynacl]' is required for the CLI,
+# which installs the cryptography and pynacl.
```
joshuagl: "which installs the cryptography and pynacl dependedencies"?

tuf/client/updater.py
```diff
@@ -130,6 +130,7 @@
 import warnings
 
 from securesystemslib import exceptions as sslib_exceptions
+from securesystemslib import formats as sslib_formats
```
joshuagl, line 136: Can we just import the formats we care about?
`from securesystemslib import PATH_SCHEMA, NAME_SCHEMA, ...` ?

tuf/developer_tool.py
```diff
@@ -75,7 +74,8 @@
     import_rsa_privatekey_from_file)
 
 from securesystemslib.keys import (
-    format_keyval_to_metadata)
+    format_keyval_to_metadata,
+    format_metadata_to_key)
```
joshuagl, line 77: Curious why you chose this pattern here, importing specific functions, but not in updater?

joshuagl (commented):
Thank you for tackling this @jku! 

FYI the series has multiple "imports: Fix securesystemslib.exceptions imports" patches. They aren't all modifying exceptions. Similarly with "imports: Fix securesystemslib.interface imports".

> This is a draft since I'll have to do something similar in securesystemslib before I can test that it all actually works with vendoring... I am also going to go through this PR again to make sure there are no typos and to improve commit messages.

Should have paid more attention to this before reviewing commit by commit, I hope the numerous comments I've left are useful.
 

> * as result securesystemslib inports are now like `from securesystemslib import formats as sslib_formats` in most cases. I don't particularly like it and am willing to consider alternatives

I'm not opposed to this, though I should point out that we can avoid many of the sslib_* imports by importing only the names we care about, you've done this in a couple of places – importing only the functions from a securesystemslib module that the tuf module is using – I assume expediency is the main reason you didn't do this throughout?

> * The patch is surprisingly large, ~900 lines. Roughly 350 are needed for the client vendoring but I modified the whole code base to at least be consistent. The tests have not been touched.

Completely agree with this rationale.

tuf/developer_tool.py
jku: oh just because developer_tool was already using the "from x import y" pattern for keys (just not consistently). It's likely that other places would benefit from doing that but in those cases I would have had more new local name conflicts to worry about (when I import the module, I can just check the module name is not used)

tuf/client/updater.py
jku: This would probably be a good idea

tuf/developer_tool.py
jku: it looks like I have two commits with keydb changes, the latter one seems to handle these. Will combine them.

tuf/client/updater.py
jku: probably reasonable, will test this for sslib_formats and sslib_exceptions

lukpueh:
Thanks a ton for taking a stab at this thankless task, and especially for the thorough separation of commits! As requested, I only skimmed the patches here and in secure-systems-lab/securesystemslib#316. Let me know when you want a full review.

Here are some thoughts regarding your notes from above.

> * tuf modules are now imported with this style `from tuf import download`. 

I'd like to add that this does not only fix the vendoring issues but also aligns with our [style recommendation](https://github.com/secure-systems-lab/code-style-guidelines/blame/28bccecd4ff0783e9e0395c8e6d7cbc9e532bc34/python.md#L55-L56), even though the [Google guide accepts both](https://github.com/secure-systems-lab/code-style-guidelines/pull/21#discussion_r533310544).


> * as result securesystemslib inports are now like `from securesystemslib import formats as sslib_formats` in most cases. I don't particularly like it and am willing to consider alternatives

I'm fine with this approach. IMO the secondary goal (besides making vendoring work) should be to help the reader attribute function or other names to modules or packages. And `sslib_interface.import_publickeys_from_file` seems to serve that purpose better than `interface.import_publickeys_from_file`.

> * The import lists that we're changed have been sorted so tuf imports are last

Excellent, this too is a style guide recommendation.

* The patch is surprisingly large, ~900 lines. Roughly 350 are needed for the client vendoring but I modified the whole code base to at least be consistent. The tests have not been touched.

👍 


Do you think it's worth to add an ADR for the rationale in this PR? Or to amend our style guide? Specifically in regards to:

`from package import module` vs. `from package import module as xyz_module` vs. `from package.module import name`

jku:
> To you think it's worth to add an ADR for the rationale in this PR? Or to amend our style guide? Specifically in regards to:
> 
> `from package import module` vs. `from package import module as xyz_module` vs. `from package.module import name`

Eh, I think it maybe depends... So far I've mostly used the first (and second where confusion was more likely) but the third one makes sense for some modules. I've avoided the third option for now mostly to keep the changes to a minimum and so I only had to worry about one new local name (the module name) and not multiple as might be the case if we import individual things from modules.

Related thought (this might not be a good idea, just something I'm wondering), for some of our most common imports I've been thinking: If TUF uses e.g. schemas from `tuf.formats` and `securesystemslibs.formats` wouldn't it make sense for `tuf.formats` to expose the relevant items from `securesystemslibs.formats`? This way there would be one place (tuf.formats) that takes care of possible conflicts and other TUF code could just `import tuf.formats` or individual items from there (module name not really needed because now there's no confusion over where they come from). Same goes to a degree for exceptions -- where it's a bit weird that code that does not call securesystemslib code might have to import `securesystemslib.exceptions` to handle errors.

lukpueh:
> Related thought (this might not be a good idea, just something I'm wondering), for some of our most common imports I've been thinking: If TUF uses e.g. schemas from tuf.formats and securesystemslibs.formats wouldn't it make sense for tuf.formats to expose the relevant items from securesystemslibs.formats? 

I generally think that's a good idea, but given that we want to get away from schemas (https://github.com/secure-systems-lab/securesystemslib/issues/183) maybe not worth the effort?

> Same goes to a degree for exceptions -- where it's a bit weird that code that does not call securesystemslib code might have to import securesystemslib.exceptions to handle errors.

I agree that it's weird. But I think the solution should be a revision of the exception taxonomy (https://github.com/theupdateframework/tuf/issues/1131, secure-systems-lab/securesystemslib#191). AFAICS there is no big win in defining any tuf exceptions in securesystemslib.

tuf/client/updater.py
jku: I've not done this for now... there are cases where clashes are definitely possible and a couple where they actually exist right now: the main issue is the generic "Error" exception that both projects use.

It's an idea worth exploring but I think not as part of this PR.

tuf/log.py
jku: the whole file is full of this naming, decided not to touch it

tuf/client/updater.py
jku: shoot, I had missed this suggestion in this pass. This is not done yet...

jku: Decided not to do this after all: in some cases this approach would work but there are a lot of uses in repository_tool, repository_lib, keydb and roledb . Some of them would end up with clashes (at least exceptions.Error)... The schemas probably would work but I'm not sure the amount of work is reasonable if we want to change all uses.

I think I would prefer leaving this outside of the PR.

jku:
I've marked this ready for review. 

I think this is the one question that hasn't been answered (in this issue):
> I'm not opposed to this, though I should point out that we can avoid many of the sslib_* imports by importing only the names we care about, you've done this in a couple of places – importing only the functions from a securesystemslib module that the tuf module is using – I assume expediency is the main reason you didn't do this throughout?

I think the places where I did it were already using this format for some other formats from this module (so I was aiming for least disruption to existing code). Otherwise I tried to follow the guideline of importing modules, not module contents. It's true that in some instances the latter would work better: schemas are an example but modifying them is a lot of work that might not be 100% trivial and that I'd rather leave for a follow up if needed

tuf/client/updater.py
joshuagl: Sounds reasonable, thanks for looking into it.

tuf/client/updater.py
joshuagl: Sounds reasonable to me.

tuf/requests_fetcher.py
```diff
@@ -14,7 +14,7 @@
 import urllib3.exceptions
 
 from tuf import exceptions
-import tuf.settings
+from tuf import settings
```
lukpueh, line 17: Nit pick: I think the patch in this file should have gone into 7841008a. (Note that I mostly point this out so that you see that I actually read the diff 😄 . Feel free to ignore.)

lukpueh (approved):
Thanks again for doing this dull deed, @jku! Here is how I reviewed your PR:

I have **not** verified if this patch
- indeed makes tuf vendoring-compatible,
- is complete,
- doesn't shadow any names.

Also, I have **not** commented on inconsistent adoption of new module/sybmol paths in docstrings, most notably in `<Exceptions>`, or `Raises:` sections. I noticed some and briefly considered pointing them out, but I'm unsure if it is worth the effort.


What **I did do** is I
- went through each commit,
- read each diff line by line, and
- checked that each diff corresponds to the commit message.
- I also made sure you addressed @joshuagl's comments.

Based on that, your PR LGTM.

If that's enough, I'd say we merge, after you have rebased and fixed the conflict in `tuf/api/metadata.py`. 

ps: I'd also like to land this before https://github.com/theupdateframework/tuf/pull/1314 because that way it will be easier to resolve conflicts (nothing big, I checked, but still).

jku:
Changes since last push:
* rebased to develop
* fixed rebase issues in "imports: Make 'formats' imports vendoring-compatible" 
* fixed rebase issues in "imports: Make 'exceptions' imports vendoring-compatible" 
* Added a commit for a missed requests_fetcher import
* Added a commit to bundle imports like isort likes them

lukpueh:
Still LGTM! :)

PR #1270, opened 2021-01-29
ADR0006: Where to implement model serialization

lukpueh:
Fixes #-
Related to #1139 (Implements Option 5 of this ADR)

Add decision record about the design of de/serialization between TUF metadata class model and tuf wire line metadata formats.

requirements**:

### DIFF
--- a/docs/adr/0006-where-to-implemenent-model-serialization.md
+++ b/docs/adr/0006-where-to-implemenent-model-serialization.md
@@ -1,0 +1,116 @@
+# Separate metadata serialization from metadata class model but keep helpers
+
+Technical Story: https://github.com/theupdateframework/tuf/pull/1279
+
+## Context and Problem Statement
+In the course of implementing a class-based role metadata model we have also
+reviewed options on how to design serialization infrastructure between wire
+formats and the class model. In an initial attempt we have implemented
+serialization on the metadata class (see option 1), but issues with inheritance
+and calls for more flexibility have caused us to rethink this approach.
+
+## Decision Drivers
+* A class-based role metadata model (see ADR4) requires serialization routines
+  from and to wire formats
+* TUF integrators may require custom serialization implementations for custom
+  wire formats
+* Readability and simplicity of implementation for users and maintainers
+* Recognizability of specification
+
+## Considered Options
+1. Serialization in metadata classes
+2. Serialization in metadata subclasses
+3. Serialization separated from metadata classes
+4. Compromise 1: Default serialization methods in metadata classes /
+   non-default serialization separated
+5. Compromise 2: Serialization separated / dict conversion helper methods for
+   default serialization in metadata classes
+
+## Decision Outcome
+Chosen option: "Compromise 2", because implementing dict conversion as methods
+on a corresponding class is idiomatic and allows for well-structured code.
+Together with a separated serialization interface, it provides both ease of use
+and maintenance, and full flexibility with regards to custom serialization
+implementations and wire formats.
+
+## Pros and Cons of the Options
+
+### Option 1: Serialization in metadata classes
+
+Serialization is implemented on metadata classes, e.g.
+`Metadata.serialize_as_json()`, etc.
+
+* Good, because serialization for any object is encapsulated within the
+  corresponding class and thus structured in small code chunks, using the
+  already existing hierarchical class model structure.
+* Good, because the TUF specification is heavily based on json, even if only
+  for illustrative purposes, thus this option facilitates recognizability.
+* Bad, because it might suggest that TUF is limited to json alone.
+* Bad, because it does not facilitate custom serialization implementations.
+* Bad, because it can get complicated with inheritance in the class model.
+  *NOTE: a workaround exists in #1279.*
+
+### Option 2: Serialization in metadata subclasses
+Serialization is implemented on metadata subclasses, e.g.
+`JsonMetadata.serialize()`, etc.
+
+* Good, because the wire format is decoupled from the base classes, not giving
+  the impression that TUF is limited to json, and facilitating custom
+  implementations.
+* Bad, because a user needs to decide on serialization ahead of time, when
+  instantiating the metadata objects.
+* Bad, because the metadata model has many classes, which would all need to be
+  subclassed accordingly.
+
+### Option 3: Serialization separated from metadata classes
+Serialization is implemented independently of the metadata class, e.g. by
+defining an abstract `Serializer` interface, which must be implemented in
+subclasses, e.g. `JsonSerializer`, etc.
+
+* Good, because the wire format is completely decoupled from the class model,
+  not giving the impression that TUF is limited to json, and facilitating
+  custom implementations.
+* Good, because it can serve as exact blueprint for custom implementations.
+* Bad, because a decoupled serialization implementation needs to "re-implement"
+  the entire class hierarchy, likely in a procedural manner.
+
+### Option 4: Compromise 1
+Default json serialization is implemented on the metadata class as described in
+(1), but can be overridden using an independent `Serializer` interface as
+described in (3).
+
+* Good, for the reasons outlined in options (1) and (3), i.e. encapsulation
+  within classes but decoupled class model and wire format.
+* Bad, because it creates two different code paths for default and non-default
+  wire formats making the code more complex and prone to deteriorate,
+  especially on the non-default path.
+* Bad, because the on-the-class default implementation can not be used as
+  blueprint for custom implementations.
+
+### Option 5: Compromise 2
+Serialization is implemented independently of the metadata class as described
+in (3). However, the *meat* of the default `JsonSerializer`, i.e. conversion
+between metadata objects and dicts, is implemented on the metadata class, e.g.
+as `Metadata.to_dict()`, etc.
+
+* Good, for the reasons outlined in options (1) and (3), i.e. encapsulation
+  within classes but decoupled class model and wire format, without the
+  disadvantage in (4) of having two completely different code paths.
+* Good, because it makes the separate default serializer a minimal wrapper
+  around the dict conversion methods.
+* Good, because other serialization implementations might also make use of dict
+  conversion methods.
+* Good, because conversion between class objects and dicts is akin to type
+  casting, which is idiomatic to implement on the class.
+* Bad, because the on-the-class default implementation can not be used as
+  blueprint for custom implementations.
+
+## Links
+* [ADR4: Add classes for complex metadata attributes (decision driver)](/Users/lukp/tuf/tuf/docs/adr/0004-extent-of-OOP-in-metadata-model.md)
+* [PR: Add simple TUF role metadata model (implements option 1)](https://github.com/theupdateframework/tuf/pull/1112)
+  - [details about separation of serialization and instantiation](https://github.com/theupdateframework/tuf/commit/f63dce6dddb9cfbf8986141340c6fac00a36d46e)
+  - [code comment about issues with inheritance](https://github.com/theupdateframework/tuf/blob/9401059101b08a18abc5e3be4d60e18670693f62/tuf/api/metadata.py#L297-L306)
+* [PR: New metadata API: add MetadataInfo and TargetFile classes (recent ADR discussion impetus)](https://github.com/theupdateframework/tuf/pull/1223)
+  - [more discussion about issues with inheritance](https://github.com/theupdateframework/tuf/pull/1223#issuecomment-737188686)
+* [SSLIB/Issue: Add metadata container classes (comparison of options 1 and 2)](https://github.com/secure-systems-lab/securesystemslib/issues/272)
+* [tuf-on-a-plane parser (implements option 3)](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/master/src/tuf_on_a_plane/parsers/)
--- a/docs/adr/index.md
+++ b/docs/adr/index.md
@@ -10,6 +10,7 @@ This log lists the architectural decisions for tuf.
 - [ADR-0003](0003-where-to-develop-TUF-1-0-0.md) - Develop TUF 1.0.0 in a subdirectory of the current TUF implementation
 - [ADR-0004](0004-extent-of-OOP-in-metadata-model.md) - Add classes for complex metadata attributes
 - [ADR-0005](0005-use-google-python-style-guide.md) - Use Google Python style guide with minimal refinements
+- [ADR-0006](0006-where-to-implemenent-model-serialization.md) - Separate metadata serialization from metadata class model but keep helpers
 
 <!-- adrlogstop -->
 


MVrachev:
I will leave some thoughts here for the different options:

1. `De/serialization on metadata classes`: 

About the `first` bad point: do we plan to add support for additional metadata types additional than JSON? 
Is there a user that asked for that and do we consider switching? 
For the `second` bad point I agree it's a bit complicated. When I wrote my first pr for the new api I had a couple of questions about the behavior of the de/serialization methods. Still, I am not sure how to simplify it...
With the current API we are just supporting many stuff (which all are meaningful?) : 
- the possibility to add  support for additional file formats (by providing for example `from_yaml` which calls `from_dict()`)
- custom filenames for the files representing the roles, that's why we have to check the `_type` field in `from_dict()`
- additional helpful methods `to_json`, `to_json_file`, `from_json`, `from_json_file`. 

I agree with the `good` points. From an OOP perspective, it feels for me this is the best solution.
 `Sidenote`: it seems `from_json` and `from_json_file` are the public methods to use for deserialization, then it's good to mark `from_dict` as a private (with `_` for example), internal method. The same applies to other internal methods as well. If we chose that approach for the adr, I can create an issue/pr for this.

2. `De/serialization on metadata subclasses`

`About the bad point`: this seems to add not trivial complecity to the model. If I understand the idea correctly, we would need to add the following classes `JsonMetadata`, `JsonRoot`, `JsonTargets`, etc.? This seems an overkill to support `possible` custom parsers...

3. `De/serialization separated`

`About the bad point`: At first, this seemed like a reasonable option, but then I realized there are two cons:
- the smaller one: as you mentioned "re-implement" the entire class hierarchy
- `what I am most concerned about`:  the procedural implementation we have to provide. I don't have any bad feelings about Trishank or `tuf-on-plane`, but at least for me, it feels harder to read, follow and understand what's going on. From an OOP perspective, this option seems not right to me.


`Summary`:  Overall I feel like the first option we have now is the best. It's the most logical, fits in my mental model and the deserialization functions don't worry about types and finding out what kind of JSON they are deserializing. It's unclear to me how many of our users would want custom file formats or do we considering supporting other file formats and that's why it's not the top priority from my perspective.
What I would like is to discuss in a different issue is how we can simplify this model more? Is there a way? Do we want to be so extendable and support all of the above points I mentioned?

trishankatdatadog:
Clearly, I'm for the third option.

In my experience building [tuf-on-a-plane](https://github.com/trishankatdatadog/tuf-on-a-plane), it was very helpful to separate the two concerns, as I could: (1) cleanly separate the concerns and reason about them (a bug in one did not have to do with the other), and (2) easily extend custom parsers/decoders/deserializers without changing anything about the intermediate Metadata representation.

The problem with the first option is that both the (de)serialization and Metadata code are all in one place, and it does not help to read or reason about them at the same time. I argue that they are both conceptually and practically separate.

I should also note that whatever we decide (e.g., global key databases instead of functionally local key lookup tables) does end up influencing other implementations. The easier we make it for other developers to reason about how TUF works, the better. So if that means a better separation of the concerns, then we should do that.

Anyway, that's just my 0.02 BTC (even though it's something I feel relatively strong about). What do others think?

joshuagl:
> I should also note that whatever we decide (e.g., global key databases instead of functionally local key lookup tables) does end up influencing other implementations. The easier we make it for other developers to reason about how TUF works, the better. So if that means a better separation of the concerns, then we should do that.

This is something we're seeing over and over as more implementations of TUF are being developed. Let's make "The easier we make it for other developers to reason about how TUF works, the better." our maintainer mantra.

trishankatdatadog:
> This is something we're seeing over and over as more implementations of TUF are being developed. Let's make "The easier we make it for other developers to reason about how TUF works, the better." our maintainer mantra.

Thanks, Joshua. Another concern I'd like to us address is removing the optics that TUF is officially tied exclusively to canonical JSON as data transport, so if separating the two concerns (parsing vs models) helps us do that, then all the better.

joshuagl:
I've been thinking about this a for a few days while reviewing the current `tuf.api.metadata` implementation. As @trishankatdatadog indicated, a primary goal for any metadata model in the reference implementation should be recognisability – that the model represents what is described by the specification, and shared on the wire.

I also feel like that the reference implementation has some other goals: 
* be easy to understand (and copy) – the reference implementation is a model for other implementations
* concise, yet extensible – the reference implementation should be easily extensible, with a simple core that can easily be adapted to a variety of update systems

Given those two things, I feel like proposal 4 in the ADR is not the way to go. As I understand it, it would necessitate a separate serialisation implementation for formats that aren't part of the default implementation. This would mean that 1) it is harder for people to implement a new serialisation method – many will want to start by copying, then modifying, the default mechanism and that won't work if JSON serialisation is baked in, and 2) that there's a separate codepath for the non-default serialisation mechanism – which may lead to subtle issues around serialisation to non-default formats.

Given the goals proposed above, I am in favour of a generic and abstract serialisation/deserialisation interface which is separate from the metadata model, extensible without directly modifying the code of python-tuf, and shared by the upstream and any downstream serialisation/deserialisation implementations.

lukpueh:
Thanks for your comments, everyone! Based on the discussion here I have implemented option 3 in #1279. 


I do, however, have another option, or rather a variant of option 3, to offer. See [option 5](https://github.com/theupdateframework/tuf/blame/7009595f9194c9ed91df3976a695435fd2db2bc1/docs/adr/0006-where-to-implemenent-model-serialization.md#L97-L109) in the just updated ADR, which I have generally fleshed out a bit.

Note that #1279 is already crafted in such a way that we can easily switch to option 5 by chopping off the last commit (b786bc0).

MVrachev:
> Thanks for your comments, everyone! Based on the discussion here I have implemented option 3 in #1279.
> 
> I do, however, have another option, or rather a variant of option 3, to offer. See [option 5](https://github.com/theupdateframework/tuf/blame/7009595f9194c9ed91df3976a695435fd2db2bc1/docs/adr/0006-where-to-implemenent-model-serialization.md#L97-L109) in the just updated ADR, which I have generally fleshed out a bit.
> 
> Note that #1279 is already crafted in such a way that we can easily switch to option 5 by chopping off the last commit ([b786bc0](https://github.com/theupdateframework/tuf/commit/b786bc0282efb1866f930f559e476d324f3dfcf7)).

I had a look into option 5 and read the code without commit b786bc0.
I see the benefits of this approach and totally agree with the second one. 
It makes sense to instantiate new objects inside a member
function instead of a separate function.

jku:
At first option 5 felt like a bit of smoke-and-mirrors to hide the fact that metadata still implements the json de/serialization... but on second look it doesn't bother me much. The potential issues might be:
* maybe in the future the de/serialization is not actually possible using the public metadata API and we fail to notice because we do it inside the metadata classes
* it's not really a 1:1 example for the hypothetical developer who is implementing another non-json de/serializer

but I think the advantages you listed are more important than those issues (let's not optimize too much for the hypothetical case)... So LGTM.

---

I did try to come up with a clever-er way of separating to_dict()/from_dict() from the actual metadata just a little bit. E.g. some "JsonMixin" classes that lived with the actual Metadata classes but were still clearly separate. I could not come up with anything cleaner than this version.

docs/adr/0006-where-to-implemenent-model-serialization.md
```diff
@@ -0,0 +1,118 @@
...
+* Readability and simplicity of implementation
+* Recognizability of specification
+
+## Considered Options
+1. De/serialization on metadata classes
+2. De/serialization on metadata subclasses
+3. De/serialization separated
+4. Compromise 1: Default json de/serialization on metadata classes,
+   non-default de/serialization separated
+5. Compromise 2: Dictitionary conversion for default json de/serialization on
```
trishankatdatadog:
```suggestion
5. Compromise 2: Dictionary conversion for default json de/serialization on
```

lukpueh:
> - maybe in the future the de/serialization is not actually possible using the public metadata API and we fail to notice because we do it inside the metadata classes

I think that's only an [issue in option 4](https://github.com/theupdateframework/tuf/pull/1270/files#diff-e29ede7b722105e47d49d3f03232233f11b2b5d601f47ecc28848bfcbb4f60b5R91-R93). In option 5 we still serialize via the external serializer and thus [don't have two different code paths](https://github.com/theupdateframework/tuf/pull/1270/files#diff-e29ede7b722105e47d49d3f03232233f11b2b5d601f47ecc28848bfcbb4f60b5R104). Or do I miss something?

> - it's not really a 1:1 example for the hypothetical developer who is implementing another non-json de/serializer

You are right, I even briefly had that listed as con for option 5, but removed it because it felt so weak. The inherit-from-abstract-serializer scaffolding can be copy-pasted from the default json de/serializer, and adopting the to/from_dict method stack seemed like a fair ask.

trishankatdatadog:
I'm not sure I'm convinced about Option 5. Can someone explain why it's conceptually easier to go from a high-level Metadata object to an intermediate-level dictionary object and then to a low-level serialization format such as JSON or XML? Why is the intermediate step necessary?

lukpueh:
What about the arguments in the ADR, @trishankatdatadog?

jku:
> > * maybe in the future the de/serialization is not actually possible using the public metadata API and we fail to notice because we do it inside the metadata classes
> 
> I think that's only an [issue in option 4](https://github.com/theupdateframework/tuf/pull/1270/files#diff-e29ede7b722105e47d49d3f03232233f11b2b5d601f47ecc28848bfcbb4f60b5R91-R93). In option 5 we still serialize via the external serializer and thus [don't have two different code paths](https://github.com/theupdateframework/tuf/pull/1270/files#diff-e29ede7b722105e47d49d3f03232233f11b2b5d601f47ecc28848bfcbb4f60b5R104). Or do I miss something?

I just mean if we later add an internal function in Metadata and use it in to_dict()... this internal function is now really part of serialization code but not available to someone doing it outside of Metadata. I don't think this is something we should worry about here. 

Maybe related point: The way metadata.py refers to JsonDicts in a lot of places does make it a little confusing: are they _really_ json or just very specific dictionaries? I'm not sure if I can tell how separate metadata.py and json are at this point (maybe worth documenting things like `Metadata.signature` & `sign()` return value and `Root.__init__()` arguments at some point)

trishankatdatadog:
> What about the arguments in the ADR, @trishankatdatadog?

Sorry, I'm not sure I fully follow. Why is it easier for third-party developers to write decoders from JSON/XML to Python dictionaries instead of more natural, higher-level Metadata objects? What are we saving here? I must be missing something.

jku:
> Sorry, I'm not sure I fully follow. Why is it easier for third-party developers to write decoders from JSON/XML to Python dictionaries instead of more natural, higher-level Metadata objects? What are we saving here? I must be missing something.

I don't think the idea was for 3rd parties to to use to_dict/from_dict: they are just helpers for the default de/serializer -- it's code that could be in the de/serializer but becames a bit simpler when you can just include it in metadata. The status and purpose of these functions could maybe be better documented.

lukpueh:
> Sorry, I'm not sure I fully follow. Why is it easier for third-party developers to write decoders from JSON/XML to Python dictionaries instead of more natural, higher-level Metadata objects? What are we saving here? I must be missing something.

JSON to dictionary is really easy:

```python
import json
json.loads(data)
```

For formats that don't have libraries with builtin dict de/serialization support it doesn't make sense to use a dictionary intermediary, but you also don't have to.

lukpueh:
> I just mean if we later add an internal function in Metadata and use it in to_dict()... this internal function is now really part of serialization code but not available to someone doing it outside of Metadata.

Hm. I if it were all implemented in `JsonSerializer.serialize` it wouldn't be available to a different serializer to start with. 

> The way metadata.py refers to JsonDicts in a lot of places does make it a little confusing: are they really json or just very specific dictionaries? I'm not sure if I can tell how separate metadata.py and json are at this point (maybe worth documenting things like Metadata.signature & sign() return value and Root.__init__() arguments at some point)

Agreed, we should remove references to JSON in metadata.py. Removing the usage of `JsonDict` as type hint is documented in https://github.com/theupdateframework/tuf/issues/1140#issuecomment-766781447.

trishankatdatadog:
I think I'm a bit lost here. There are too many options, and I've got too many things going on to properly switch context. Would appreciate a quick 1/2 hour meeting to quickly discuss options. Especially w/o quick code to glance at, I can't immediately tell what's going on, sorry.

lukpueh:
@trishankatdatadog, did you take a look at #1279, which implements your preferred option 3, as [suggested above](https://github.com/theupdateframework/tuf/pull/1270#issuecomment-780660222)? Its PR description explains the alternative option 5 including pros and cons, and it even points to quick code to glance at that shows the diff between option 3 and 5 (see b786bc0282efb1866f930f559e476d324f3dfcf7).

I think reading the PR description and skimming the diff shouldn't take longer than half an hour and give you a good idea what's going on here. Let me know if you still need a meeting afterwards.

trishankatdatadog:
> I think reading the PR description and skimming the diff shouldn't take longer than half an hour and give you a good idea what's going on here. Let me know if you still need a meeting afterwards.

It's good. I read the code, and I like how Option 3 is implemented, thanks. What still I don't understand is the difference in Option 5. Is there any good reason to expose those utility from/to dictionary methods in the base Metadata class? Would other (de)serializers benefit from it?

lukpueh:
Thanks for the feedback, @trishankatdatadog!  Regarding your question whether other de/serializers benefit from `from/to_dict`, yes they do. As a matter of fact, #1279 implements different De/Serializers for [JSON](https://github.com/theupdateframework/tuf/blob/b786bc0282efb1866f930f559e476d324f3dfcf7/tuf/api/serialization/json.py#L26-L61) and for [canonical JSON](https://github.com/theupdateframework/tuf/blob/b786bc0282efb1866f930f559e476d324f3dfcf7/tuf/api/serialization/json.py#L64-L74), which are each just thin wrappers around `from/to_dict`. And I can very well imagine that for other formats there also exist libraries that make it easy to de/serialize from/to generic dictionaries.

Although it doesn't make a big difference whether the `from/to_dict` functions are available as [methods on the metadata classes or in a separate public module](https://github.com/theupdateframework/tuf/commit/b786bc0282efb1866f930f559e476d324f3dfcf7), I do think that they are more easily found on the corresponding class, e.g. when using a Python interpreter with auto-completion, and also because users might expect classes to have such methods (see 3rd-party libraries such as `attrs`, `pydantic`, `marshmallow`, or built-in `vars()`/`__dict__` or `dict()`/`__iter__` properties and methods that provide default and customizable dict representation for custom objects.

So one argument is that it seems **idiomatic** to implement these functions as methods on the class, and the other argument is that the code seems a bit **better structured** when each method is scoped within the corresponding class, which in turn should make maintaining/modifying/extending the class model easier.

trishankatdatadog:
Do what you think is best, Lukas and friends. I trust you will make the right decision. From my point of view, what is critical is that developers can easily write new (de)serializers targeting the same abstract, high-level metadata interface, and not worry about dictionaries and JSON and so on.

MVrachev:
I think the decision is already made - we have implemented model (de)serialization by merging https://github.com/theupdateframework/tuf/pull/1279. 
My opinion is that for now, it works fine, but if we want to change our approach we can update this ADR or create a new ADR overwriting this one. In order to merge this pr,  we have to update it with our decision.

What do you think @jku?

joshuagl (approved):
thanks to @MVrachev for following up on this long-lived PR both here and offline. We have been moving ahead with option 5 (aka compromise 2) since the merge of https://github.com/theupdateframework/tuf/pull/1279

Therefore it makes sense to merge this ADR documenting the approach and the other options we considered.

PR #1271, opened 2021-02-02
Docs: change the editable venv installation order

MVrachev:
Fixes NA

If you follow the instructions we provide for our contributors in
docs/CONTRIBUTORS.rst your sys.path (used to search for imports)
will put securesystemlib project directory first and tuf directory
second.
This creates a problem with imports from tuf modules because we can
import the wrong file or on relative imports (as currently we
use in the tests when we import utils), the imports cannot be resolved.

If we change the installation order, then tuf directory will be the
first in the import resolution path and those problems will be fixed.

PS: I want to express my gratitude towards Jussi who helped me find
this problem.

requirements**:

### DIFF
--- a/docs/CONTRIBUTORS.rst
+++ b/docs/CONTRIBUTORS.rst
@@ -155,13 +155,13 @@ via PyPI).
 
 To develop and test ``tuf`` with above commands alongside its in-house dependency
 `securesystemslib <https://github.com/secure-systems-lab/securesystemslib>`_,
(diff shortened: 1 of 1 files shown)


jku (approved):
LGTM.

For context the issue is this combination
* when editable install is used both projects make "tests" module available as top level import
* Code in tests cannot do `from . import ...` because then you cannot execute the module as file

So this seems like a reasonable way to minimize damage

PR #1279, opened 2021-02-10
Move metadata class model de/serialization to sub-package

lukpueh:
Fixes #-
Related to #1270 

This PR moves de/serialization logic of TUF metadata from the metadata class model (`api.metadata`) to a newly added serialization sub-package (`api.serialization`) and adopts the metadata API accordingly, with the goal of making it easier for users to swap de/serialization implementations. In particular this PR adds:
- abstract base classes to serialize and deserialize to and from wireline metadata bytes (for transport), plus default json implementations, and
- an abstract base class to serialize signed bytes (for signature creation and verification) and a default canonical json implementation.

*More infos can be found in ...*
- #1270, which drafts a corresponding ADR and discusses the pros and cons of this and alternative approaches,
-  79ce0346, whose commit message has very specific design considerations, and
- the other commit messages, which are informative too. ;)

**Discussion:**

The meat of the default json and canonical json de/serializers is the `from/to_dict` logic, which is currently implemented in a `serialization.util` module. I wonder if we shouldn't just keep it on the metadata class model instead (we'd only need to revert b786bc0).

*Pro `from/to_dict` as metadata class methods:*
   - better structured because encapsulated on each class
   - keeps the serialization package minimal
   - `from/to_dict` is hardly serialization, which implies bytes, but rather some sort of type casting / conversion, which seems appropriate to be implemented on the class
   - it would spare us the stupid cyclic imports
   
*Pro `*_from/to_dict` in serialization sub-package:*
   - might make it easier to copy-paste-customize when creating a custom de/serializer implementation.

requirements**:


tuf/api/metadata.py
```diff
@@ -57,144 +55,94 @@ def __init__(self, signed: 'Signed', signatures: list) -> None:
...
+                The metadata object cannot be serialized.
             securesystemslib.exceptions.StorageError:
                 The file cannot be written.
 
         """
+        if serializer is None:
+            # Function-scope import to avoid circular dependency. Yucky!!!
+            # TODO: At least move to a _get_default_metadata_serializer helper.
+            from tuf.api.serialization.json import JSONSerializer # pylint: disable=import-outside-toplevel
+            serializer = JSONSerializer(True) # Pass True to compact JSON
```
```suggestion
            serializer = JSONSerializer(compact=True)
```
Using the named argument makes it clearer what's going on and obviates the need for the comment.

tuf/api/serialization/util.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+from typing import Any, Dict, Mapping
+
+from tuf import formats
+from tuf.api.metadata import (Signed, Metadata, Root, Timestamp, Snapshot,
+                              Targets)
+
+
+def _get_signed_common_args_from_dict(_dict: Mapping[str, Any]) -> list:
```
MVrachev: How did you decide to accept `typing.Mapping` instead of `typing.Dict` here and elsewhere in this file?

lukpueh: Good question! The [`typing.Dict` docs](https://docs.python.org/3/library/typing.html#typing.Dict) have the answer:
> To annotate arguments it is preferred to use an abstract collection type such as Mapping.

I didn't really question that advice, but it makes sense when you consider the API design principle of returning specific types, while accepting generic types.

tuf/api/pylintrc
```diff
@@ -4,3 +4,6 @@ disable=fixme
 [FORMAT]
 indent-string="    "
 max-line-length=79
+
+[DESIGN]
+min-public-methods=0
```
joshuagl, line 12: I'm not sure what change in this first commit (79ce03467cf480f5922b2fa3de85a83b766a5e29) necessitated this?  Do we want to make this decision for the entirety of tuf/api? Or should we disable this on a case-by-case basis (at the class or module level)?

tuf/api/serialization/__init__.py
```diff
@@ -0,0 +1,43 @@
+"""TUF role metadata de/serialization.
```
joshuagl, line 4: Should we add SPDX style license headers to new files?
```
# Copyright New York University and the TUF contributors
```

tuf/api/serialization/json.py
```diff
@@ -0,0 +1,55 @@
+"""TUF role metadata JSON serialization and deserialization.
```
joshuagl, line 4: Same comment about SPDX license headers:
```
# Copyright New York University and the TUF contributors
```

tuf/api/serialization/json.py
```diff
@@ -0,0 +1,55 @@
...
+
+"""
+import json
+
+from securesystemslib.formats import encode_canonical
+
+from tuf.api.metadata import Metadata, Signed
+from tuf.api.serialization import (MetadataSerializer,
+                                   MetadataDeserializer,
+                                   SignedSerializer)
```
joshuagl: Note: the recently adopted style guide recommends _against_ importing individual classes or functions.

tuf/api/serialization/json.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class JSONDeserializer(MetadataDeserializer):
+    """Provides JSON-to-Metadata deserialize method. """
+
+    def deserialize(self, raw_data: bytes) -> Metadata:
+        """Deserialize utf-8 encoded JSON bytes into Metadata object. """
+        _dict = json.loads(raw_data.decode("utf-8"))
```
joshuagl: Nit: could we rename this variable, rather than using the `_` trick? Maybe json_dict, json_object (as the return type of `json.loads` is a Python object), or loaded_json.

tuf/api/serialization/json.py
```diff
@@ -0,0 +1,55 @@
...
+        compact: A boolean indicating if the JSON bytes generated in
+                'serialize' should be compact by excluding whitespace.
+
+    """
+    def __init__(self, compact: bool = False) -> None:
+        self.compact = compact
+
+    def serialize(self, metadata_obj: Metadata) -> bytes:
+        """Serialize Metadata object into utf-8 encoded JSON bytes. """
+        indent = (None if self.compact else 1)
```
joshuagl: Out of curiosity, why did you opt for 1 space indent for the non-compact version?

tuf/api/metadata.py
```diff
@@ -12,19 +12,19 @@
...
 from securesystemslib.keys import create_signature, verify_signature
 
+from tuf.api.serialization import (MetadataSerializer, MetadataDeserializer,
+                                   SignedSerializer)
+
 import tuf.formats
 import tuf.exceptions
 
 
+
```
joshuagl: What happened here? :-)

tuf/api/metadata.py
```diff
@@ -101,32 +101,17 @@ class also that has a 'from_dict' factory method. (Currently this is
...
-
-
-    @classmethod
-    def from_json_file(
-            cls, filename: str,
-            storage_backend: Optional[StorageBackendInterface] = None
-            ) -> 'Metadata':
-        """Loads JSON-formatted TUF metadata from file storage.
+    def from_file(
+        cls, filename: str, deserializer: MetadataDeserializer = None,
```
joshuagl: Should the typing annotation mark deserializer as optional, like storage_backend ?

tuf/api/metadata.py
```diff
@@ -161,37 +158,48 @@ def to_json(self, compact: bool = False) -> None:
                 sort_keys=True)
 
 
-    def to_json_file(
-            self, filename: str, compact: bool = False,
-            storage_backend: StorageBackendInterface = None) -> None:
-        """Writes the JSON representation of self to file storage.
+    def to_file(self, filename: str, serializer: MetadataSerializer = None,
+                storage_backend: StorageBackendInterface = None) -> None:
```
joshuagl: As above, I think the type annotations for serializer and storage_backend should mark them as optional

tuf/api/metadata.py
```diff
@@ -161,37 +158,48 @@ def to_json(self, compact: bool = False) -> None:
...
         with tempfile.TemporaryFile() as temp_file:
-            temp_file.write(self.to_json(compact).encode('utf-8'))
+            temp_file.write(serializer.serialize(self))
             persist_temp_file(temp_file, filename, storage_backend)
 
 
     # Signatures.
-    def sign(self, key: JsonDict, append: bool = False) -> JsonDict:
+    def sign(self, key: JsonDict, append: bool = False,
+             serializer: SignedSerializer = None) -> JsonDict:
```
joshuagl: Also here, serializer is optional.

Nit: I wonder if we should rename the parameter signed_serializer. Using the same name for named parameters that take different types could be confusing.

tuf/api/metadata.py
```diff
@@ -213,11 +227,15 @@ def sign(self, key: JsonDict, append: bool = False) -> JsonDict:
         return signature
 
 
-    def verify(self, key: JsonDict) -> bool:
+    def verify(self, key: JsonDict,
+               serializer: SignedSerializer = None) -> bool:
```
joshuagl: Same comments as above, serializer is optional and might be better named signed_serializer

tuf/api/serialization/json.py
```diff
@@ -16,16 +17,22 @@
...
 
 class JSONDeserializer(MetadataDeserializer):
     """Provides JSON-to-Metadata deserialize method. """
 
     def deserialize(self, raw_data: bytes) -> Metadata:
         """Deserialize utf-8 encoded JSON bytes into Metadata object. """
-        _dict = json.loads(raw_data.decode("utf-8"))
-        return Metadata.from_dict(_dict)
+        try:
+            _dict = json.loads(raw_data.decode("utf-8"))
```
joshuagl: Same comment here, with the code moved, can we rename _dict? Possible alternatives include  json_object and json_dict.

tuf/api/metadata.py
```diff
@@ -357,6 +261,7 @@ def bump_version(self) -> None:
         self.version += 1
 
 
+
```
joshuagl: Are we aiming for three blank lines between top-level definitions? The style guide suggests two: https://google.github.io/styleguide/pyguide.html#35-blank-lines

tuf/api/metadata.py
```diff
@@ -487,6 +369,7 @@ def update(self, version: int, length: int, hashes: JsonDict) -> None:
         }
 
 
+
```
joshuagl: Same comment on blank lines between top-level definitions.

tuf/api/metadata.py
```diff
@@ -545,6 +419,7 @@ def update(
             self.meta[metadata_fn]['hashes'] = hashes
 
 
+
```
joshuagl: Same blank lines comment here, too.

joshuagl (commented):
Lovely PR, apologies for the delay in getting to review this @lukpueh.

I have a few minor questions and suggestions in-line; mostly around typing annotations, and variable & parameter names. 

I also want to point out that the new code we are writing since introducing a new coding style seems to consistently violate the style guide's recommendations around imports, [specifically](https://google.github.io/styleguide/pyguide.html#22-imports):

> Use `import` statements for packages and modules only, not for individual classes or functions.

New code in tuf.api is fairly consistently importing individual classes and functions from modules. Should we consider whether we want to make a style guide modification at this point?

> * [79ce034](https://github.com/theupdateframework/tuf/commit/79ce03467cf480f5922b2fa3de85a83b766a5e29), whose commit message has very specific design considerations

Thank you for the detailed commit messages. I wonder whether we might want to capture some of this detail in the ADR?

> **Discussion:**
[snip]
> _Pro `from/to_dict` as metadata class methods:_
> 
> * better structured because encapsulated on each class
> * keeps the serialization package minimal
> * `from/to_dict` is hardly serialization, which implies bytes, but rather some sort of type casting / conversion, which seems appropriate to be implemented on the class
> * it would spare us the stupid cyclic imports

Having reviewed this PR, I am in favour of keeping `[from|to]_dict` as metadata class methods. Curious to read what others think.

tuf/api/pylintrc
lukpueh: > I'm not sure what change in this first commit (79ce034) necessitated this?

The admittedly long commit message has an admittedly subtle hint:
```
 - configure api/pylintrc to allow classes without public methods (default was 2)
```

In other words, all the abstract classes in `serialization/__init__.py ` are one method short of the pylint ideal.

> Do we want to make this decision for the entirety of tuf/api? Or should we disable this on a case-by-case basis (at the class or module level)?

The check seems a bit controversial (see stackoverflow and pylint issue tracker). So I think we can just disable it.

tuf/api/serialization/json.py
lukpueh: I know. I guess I just kept using the convention that we already used in this file, which we started before fully embracing our new style guide... and because PEP 8 says it's okay.

tuf/api/serialization/json.py
lukpueh: I just adopted what was already there:
https://github.com/theupdateframework/tuf/blob/74b1549c0041fd6e13c0277713803dd61d2baca4/tuf/repository_lib.py#L444-L445

tuf/api/serialization/util.py
```diff
@@ -0,0 +1,149 @@
...
+        "targets": obj.targets,
+        "delegations": obj.delegations,
+    })
+    return _dict
+
+def signed_to_dict(obj: Signed) -> Dict[str, Any]:
+    """Returns dict representation of 'Signed'-subclass object. """
+    # Dispatch to '*_to_dict'-function based on 'Signed' subclass type.
+    # TODO: Use if/else cascade, if easier to read!
+    return {
```
trishankatdatadog: Whoa. Clever, but maybe a bit too much magic?

lukpueh: Yes probably. :'(

tuf/api/metadata.py
lukpueh: I must admit that I follow this secret personal convention:

Based on the guide I use "single blank lines as [I] judge appropriate within functions or methods",  but then I use two blank lines between methods, to better distinguish from the single ones, and then three blank lines between top-level definitions, to distinguish from the two blank lines.

I should probably either strictly adhere to the guide, or propose a style guide amendment.

tuf/api/metadata.py
lukpueh: See https://github.com/theupdateframework/tuf/pull/1279#discussion_r582149810

tuf/api/metadata.py
lukpueh: See https://github.com/theupdateframework/tuf/pull/1279#discussion_r582149810

tuf/api/metadata.py
lukpueh: See https://github.com/theupdateframework/tuf/pull/1279#discussion_r582149810

lukpueh:
Thanks for the detailed review and great suggestions, @joshuagl! I left a couple of replies inline and will address the rest in code. Regarding...

> New code in tuf.api is fairly consistently importing individual classes and functions from modules. Should we consider whether we want to make a style guide modification at this point?

Maybe we should ... also in the light of #1261. Do you think we should get this straight before landing this PR?

> Thank you for the detailed commit messages. I wonder whether we might want to capture some of this detail in the ADR?

I tried to capture the gist without making the ADR too verbose. But I'll try to revise the ADR once more. :) 

> Having reviewed this PR, I am in favour of keeping `[from|to]_dict` as metadata class methods. Curious to read what others think.

Cool. I'll wait just a bit longer to see if others still want to weigh in.

trishankatdatadog:
> Having reviewed this PR, I am in favour of keeping `[from|to]_dict` as metadata class methods. Curious to read what others think.

Can someone enlighten me as to why? Cos I don't see it yet.

MVrachev:
> > Having reviewed this PR, I am in favour of keeping `[from|to]_dict` as metadata class methods. Curious to read what others think.
> 
> Can someone enlighten me as to why? Cos I don't see it yet.

My personal feeling is that `from_dict`  functions are logical to be located inside the classes because they create a new instance.
I see them as a second constructor. 
About `to_dict` methods I see them as `casting` which again feels like it's a functionality logically connected with the classes.

trishankatdatadog:
@MVrachev but are they useful to other (de)serializers?

MVrachev:
> @MVrachev but are they useful to other (de)serializers?

I am not sure where are the `[from|to]_dict` functions located relate to the other (de)serializers.
Could you elaborate more on this @trishankatdatadog?

jku:
> > @MVrachev but are they useful to other (de)serializers?
> 
> I am not sure where are the `[from|to]_dict` functions located relate to the other (de)serializers.
> Could you elaborate more on this @trishankatdatadog?

The point Trishank is making is that they look like implementation details of the json de/serializer hidden inside metadata.py. And I think we should admit that that's what they are -- e.g. `from_dict()` is not a generic constructor it's clearly a helper for deserializing json.

On the other hand, the reason Lukas is proposing we leave them in is that it makes the code simpler -- I agree with this as well: the compromise seems reasonable and should not make other de/serializers any harder to implement.

lukpueh:
FYI, just left another comment on https://github.com/theupdateframework/tuf/pull/1270#issuecomment-785965015

tuf/api/metadata.py
joshuagl: Agreed, we should strictly adhere to the guide or propose an amendment. I'm neither for or against this personal convention, but I do think amendments to the style guide which contradict the "upstream" Google guide should be kept to a minimum.

tuf/api/serialization/util.py
joshuagl: Agreed, the if/else cascade is uglier but less magical and easier to follow.

joshuagl:
> Thanks for the detailed review and great suggestions, @joshuagl! I left a couple of replies inline and will address the rest in code. Regarding...
> 
> > New code in tuf.api is fairly consistently importing individual classes and functions from modules. Should we consider whether we want to make a style guide modification at this point?
> 
> Maybe we should ... also in the light of #1261. Do you think we should get this straight before landing this PR?

No need to block this PR. However, I think we should figure this out and ensure the new code conforms to the style ASAP. Otherwise we risk similar incidents unintentionally contradicting the style guide in future. At the same time as we ensure new code conforms to the style, it would be good to add style checking to CI – and perhaps have a [black](https://black.readthedocs.io/en/stable/) configuration for folks to ensure their code matches before submission.

> The point Trishank is making is that they look like implementation details of the json de/serializer hidden inside metadata.py. And I think we should admit that that's what they are -- e.g. from_dict() is not a generic constructor it's clearly a helper for deserializing json.

> On the other hand, the reason Lukas is proposing we leave them in is that it makes the code simpler -- I agree with this as well: the compromise seems reasonable and should not make other de/serializers any harder to implement.

This matches my understanding, and opinion, also. The compromise is pragmatic and reasonable, but let's be sure the comments indicate that those methods are a helper primarily for conversion to/from JSON.

tuf/api/metadata.py
lukpueh: You are right. Let's play by the rules we chose. :) Sorry for the noise.

tox.ini
```diff
@@ -42,6 +42,9 @@ commands =
 
 [testenv:lint]
 commands =
-    pylint {toxinidir}/tuf --ignore={toxinidir}/tuf/api
+    # Use different pylint configs for legacy and new (tuf/api) code
+    # NOTE: Contrary to what the pylint docs suggest, ignoring full paths does
+    # work, unfortunately each subdirectory has to be ignored explicitly.
+    pylint {toxinidir}/tuf --ignore={toxinidir}/tuf/api,{toxinidir}/tuf/api/serialization
```
MVrachev, line 48: Is there a way to ignore all files under a certain folder?
That way you would be able to ignore everything under `tuf/api`.

tuf/api/serialization/json.py
MVrachev: I agree with Joshua on this one. 
We don't have a linter that would automatically catch those mistakes, but it's important we correct each other.
Also, it seems logical to forbid importing of functions and classes.
This way we prevent future problems with naming and namespaces.

tox.ini
lukpueh: > Is there a way to ignore all files under a certain folder?

Nope (see my comment). I even looked at the implementation of  `--ignore` and `--ignore-patterns`.

lukpueh: IIRC there's a corresponding ticket on the pylint issue tracker.

lukpueh:
This could use another review. :)

- Following discussions here, in #1270, and elsewhere I reverted the commit that moved to/from_dict methods to a separate module (8e9afc96).
- Unfortunately, I was wrong that this frees us from cyclic imports, but I at least updated the corresponding code comments (88582800).
- Given that to/from_dict remain on the class, I made an effort to demystify the design  of the subclassed 'Signed' class by adding more explicit helpers (c254fd04).
- And I tried to largely remove JSON-bias from the metadata class model so that users understand that JSON is only (the default) one of many TUF wireline formats (f421cdce).

- Apart from that I addressed code review comments regarding license boilerplate (2f57eb8e), the Optional type hint  (71a684b5), a named argument (0b92da91), argument renames (b934dedd) and blank lines (17c25e29), and a linter warning about inconsistent returns (8be8ec1c).

tuf/api/pylintrc
joshuagl: Thanks for explaining.

tuf/api/serialization/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+- Metadata de/serializers are used to convert to and from wireline formats.
+- Signed serializers are used to canonicalize data for cryptographic signatures
+  generation and verification.
+
+"""
+import abc
+
+class MetadataDeserializer():
+    """Abstract base class for deserialization of Metadata objects. """
+    __metaclass__ = abc.ABCMeta
```
joshuagl: Note: now that we are only supporting Python3 we can switch to less confusing metaclass syntax, either
```Python
import abc

class MetadataDeserializer(ABC):
```
or
```Python
import abc

class MetadataDeserializer(metaclass=ABCMeta):
```
we can handle this in a future PR, though.

tuf/api/serialization/json.py
```diff
@@ -0,0 +1,55 @@
...
+        """Serialize Metadata object into utf-8 encoded JSON bytes. """
+        indent = (None if self.compact else 1)
+        separators=((',', ':') if self.compact else (',', ': '))
+        return json.dumps(metadata_obj.to_dict(),
+                          indent=indent,
+                          separators=separators,
+                          sort_keys=True).encode("utf-8")
+
+
+class CanonicalJSONSerializer(SignedSerializer):
```
joshuagl, line 70: Do you think it is worth clarifying, via the class name, that this is the OLPC canonical JSON?
You mention it in the header above, but as we have had several adopters surprised that our Canonical JSON isn't compatible with go-tuf's Canonical JSON, I wonder whether it is worth being very explicit?

Counter-argument, I do not like `OLPCCanonicalJSONSerializer` as a class name...

tuf/api/serialization/json.py
```diff
@@ -7,6 +7,7 @@
 
 """
 import json
+import six
```
joshuagl: Why are we using six in code that was never intended to support Python2?

tuf/api/serialization/json.py
```diff
@@ -41,18 +48,26 @@ def __init__(self, compact: bool = False) -> None:
...
+        try:
+            indent = (None if self.compact else 1)
+            separators=((',', ':') if self.compact else (',', ': '))
+            return json.dumps(metadata_obj.to_dict(),
+                              indent=indent,
+                              separators=separators,
+                              sort_keys=True).encode("utf-8")
+
+        except Exception as e: # pylint: disable=broad-except
+            six.raise_from(SerializationError, e)
```
joshuagl: Python3 syntax for this appears to be 
```Python
            raise SerializationError from e
```

tuf/api/serialization/json.py
```diff
@@ -41,18 +48,26 @@ def __init__(self, compact: bool = False) -> None:
...
     def serialize(self, signed_obj: Signed) -> bytes:
         """Serialize Signed object into utf-8 encoded Canonical JSON bytes. """
-        signed_dict = signed_obj.to_dict()
-        return encode_canonical(signed_dict).encode("utf-8")
+        try:
+            signed_dict = signed_obj.to_dict()
+            return encode_canonical(signed_dict).encode("utf-8")
+
+        except Exception as e: # pylint: disable=broad-except
+            six.raise_from(SerializationError, e)
```
joshuagl: Let's not use six here
```Python
            raise SerializationError from e
```

tuf/api/metadata.py
```diff
@@ -206,10 +206,10 @@ def sign(self, key: JsonDict, append: bool = False,
             A securesystemslib-style signature object.
 
         """
-        if serializer is None:
-            # Function-scope import to avoid circular dependency. Yucky!!!
-            # TODO: At least move to a _get_default_signed_serializer helper.
-            from tuf.api.serialization.json import CanonicalJSONSerializer # pylint: disable=import-outside-toplevel
+        if signed_serializer is None:
```
joshuagl, line 206: I think we're missing some of the `serializer` -> `signed_serializer` rename in this patch (885828007dd5b5a68826470dce3dbe123d065aab)? The argument name and the variable assigned below are still `serializer`

tuf/api/metadata.py
```diff
@@ -301,31 +301,31 @@ def __init__(
...
-        # super().from_dict.
-
-        # NOTE: cls might be a subclass of Signed, if 'from_dict' was called on
-        # that subclass (see e.g. Metadata.from_dict).
-        return cls(**signed_dict)
+        # is implemented in '_common_fields_to_dict'.
+        expires = tuf.formats.expiry_string_to_datetime(expires_str)
+        return [_type, version, spec_version, expires]
 
+    def _common_fields_to_dict(self) -> Dict[str, Any]:
```
joshuagl, line 320: Should this have the `@staticmethod` decorator also? Throughout the rest of this patch it is not being called on an instance of the class, but on the `super()` proxy object, much like `_common_fields_from_dict()`.

tuf/api/serialization/json.py
```diff
@@ -0,0 +1,83 @@
...
+    """Provides JSON-to-Metadata deserialize method. """
+
+    def deserialize(self, raw_data: bytes) -> Metadata:
+        """Deserialize utf-8 encoded JSON bytes into Metadata object. """
+        try:
+            json_dict = json.loads(raw_data.decode("utf-8"))
+            metadata_obj = Metadata.from_dict(json_dict)
+
+        except Exception as e: # pylint: disable=broad-except
+            six.raise_from(DeserializationError, e)
```
joshuagl: Python3 syntax for this appears to be 
```Python
            raise SerializationError from e
```

joshuagl (changes requested):
Thank you for the updates Lukas. I have a couple of minor nits and one surprise, why are we using six in code that was never intended to run on Python2?

tuf/api/serialization/json.py
lukpueh: I agree with both arguments. :D

tuf/api/serialization/json.py
lukpueh: For `raise_from` below, which I will change to `raise ... from`.

tuf/api/metadata.py
lukpueh: It's the other way around, this line should have gone into b934deddd. Apologies for the bad `git add --patch`ing. Happy to rebase. Maybe after a final LGTM?

tuf/api/metadata.py
lukpueh: Can't be static because it needs `self`, but thanks for pointing out that those two are called on `super()`. I guess it's less confusing to call them on `self`, given that they are only called in the parent class. 🤷

tuf/api/serialization/json.py
lukpueh: Another way to avoid ambiguity would be to not call it `CanonicalJSONSerializer` at all, but something like `DefaultSignedSerializer` or so? Then people are forced to read the docstring to get more info. At any rate, I will mention OLPC in the class and function docstrings.

tuf/api/metadata.py
joshuagl: WFM

tuf/api/metadata.py
joshuagl: Sounds good, thanks.

lukpueh:
Thanks for the review! I (hope I) addressed all your comments in the last 3 commits. I chose to leave `CanonicalJSONSerializer` as is but update the docstrings. Let me know if that works for you.

I'm happy to also fix the minor git history hiccup from https://github.com/theupdateframework/tuf/pull/1279#discussion_r590493044 once you have approved the PR.

joshuagl (approved):
Thank you for quickly addressing the review comments.

> Thanks for the review! I (hope I) addressed all your comments in the last 3 commits. I chose to leave `CanonicalJSONSerializer` as is but update the docstrings. Let me know if that works for you.

Last 3 commits look great, thanks. I'm happy to leave `CanonicalJSONSerializer`. Let's see if we can come up with a 'better' name before we release with an API stability promise.

> I'm happy to also fix the minor git history hiccup from [#1279 (comment)](https://github.com/theupdateframework/tuf/pull/1279#discussion_r590493044) once you have approved the PR.

That would be good, thanks!

lukpueh:
> > I'm happy to also fix the minor git history hiccup from [#1279 (comment)](https://github.com/theupdateframework/tuf/pull/1279#discussion_r590493044) once you have approved the PR.
> 
> That would be good, thanks!

Done. Given that the [overall diff of the PR didn't change](https://github.com/theupdateframework/tuf/compare/a223961..ef91964) I suppose your approval is still valid. Merging...

PR #1282, opened 2021-02-16
Bump Cryptography, pin the py2 version separately

jku:
On Python3 bump cryptography from 3.3.1 to 3.4.5.

On python2 bump from 3.3.1 to 3.3.2 (3.3-branch is the last branch
with python2 support).

---

I think this is enough in TUF? -- once securesystemslib releases we can change TUF setup.py to require the new securesystemslib and that will bring in the correct cryptography for the binary packages?


joshuagl (approved):
This looks good to me, thank you!

PR #1283, opened 2021-02-16
CI: Limit build-on-push to develop branch only

jku:
Dependabot pushes to main repository and ends up triggering two builds
every time (one for PR, one for push): limit the rule for build-on-push
to apply to develop branch only.

If release branches are used later on they should be added to list here.


joshuagl (approved):
Thank you for the speedy submission!

PR #1289, opened 2021-02-22
ci: Remove workaround for pip resolver issue

### LINKED ISSUE #1290: Ugly pip resolver workaround in tox configuration
We're currently enabling the legacy pip resolver in our tox.ini. The reason is working around https://github.com/pypa/pip/issues/9215 that lead to hundreds of unnecessary package downloads from pypi

I'm not 100% sure the problem has otherwise disappeared but newest pip has added more info about package conflicts (this happens in the with-sslib-master tox configuration if the resolver hack is removed):

```
[20 lines of normal pip download logging removed here ...]
INFO: pip is looking at multiple versions of requests to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of pynacl to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of pycparser to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of idna to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of cryptography to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of chardet to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of cffi to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of certifi to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of tuf to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of securesystemslib to determine which version is compatible with other requirements. This could take a while.ERROR: Cannot install securesystemslib 0.19.0 (from git+http://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib), securesystemslib[crypto,pynacl]==0.19.0 and tuf==0.16.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested securesystemslib 0.19.0 (from git+http://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib)
    tuf 0.16.0 depends on securesystemslib>=0.18.0
    securesystemslib[crypto,pynacl] 0.19.0 depends on securesystemslib 0.19.0 (from https://files.pythonhosted.org/packages/d6/a3/1d02d78e44531f512b4b2778acec11e67be105a180afd3ef591983c7097d/securesystemslib-0.19.0-py2.py3-none-any.whl#sha256=f32cbc84e77d910425a64eb956541d1173c6471fa5feeb267523e1f2f99da940 (from https://pypi.org/simple/securesystemslib/) (requires-python:>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4))

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies

```

So I guess the is that requirements-pinned.txt requires `securesystemslib[crypto,pynacl]==0.19.0` and that resolves to the 0.19.0 wheel version? and then we explicitly also ask for the git version as a special case but that does not specify th extras?

jku:
* Remove the workaround that was added for pip issue 9215 and does not _seem_ to appear anymore with pip 21.0.1.
* Add the "extras" requirements for the with-sslib-master case (which pip now requires)
* do not accept plain http for anything

Fixes #1290

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,9 +24,7 @@ deps =
     # installation (see `skipsdist`), to get relative paths in coverage reports
     --editable {toxinidir}
 
-# FIXME: use legacy resolver because https://github.com/pypa/pip/issues/9215
-install_command = pip install --use-deprecated=legacy-resolver --pre {opts} {packages}
-
+install_command = pip install --pre {opts} {packages}
 
 # Develop test env to run tests against securesystemslib's master branch
 # Must to be invoked explicitly with, e.g. `tox -e with-sslib-master`
--- a/tox.ini
+++ b/tox.ini
@@ -32,7 +32,7 @@ install_command = pip install --use-deprecated=legacy-resolver --pre {opts} {pac
 # Must to be invoked explicitly with, e.g. `tox -e with-sslib-master`
 [testenv:with-sslib-master]
 deps =
-    git+http://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib
+    git+https://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib[crypto,pynacl]
     -r{toxinidir}/requirements-test.txt
     --editable {toxinidir}
 


jku:
The test failure is a coveralls upload 504 


(╯°□°）╯︵ sllɐɹǝʌoƆ

joshuagl:
We still appear to have an issue with testing against securesystemslib master:

> ERROR: Cannot install securesystemslib 0.20.0 (from git+https://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib[crypto,pynacl]) [crypto, pynacl] and securesystemslib[crypto and pynacl]==0.19.0 because these package versions have conflicting dependencies.
> 
> The conflict is caused by:
>     The user requested securesystemslib 0.20.0 (from git+https://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib[crypto,pynacl]) [crypto, pynacl]
>     The user requested securesystemslib[crypto and pynacl]==0.19.0
> 
> To fix this you could try to:
> 1. loosen the range of package versions you've specified
> 2. remove package versions to allow pip attempt to solve the dependency conflict
> 
> ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies

jku:
I'm rebasing since the securesystemslib update leads to this failing on CI: no changes, just a rebase.

So in the future the `with-sslib-master` test will fail whenever securesystemslib master contains a newer version than our requirements-pinned (I'm not sure if this happened with the old resolver).

joshuagl (approved):
> I'm rebasing since the securesystemslib update leads to this failing on CI: no changes, just a rebase.

👍 

> So in the future the `with-sslib-master` test will fail whenever securesystemslib master contains a newer version than our requirements-pinned (I'm not sure if this happened with the old resolver).

Ah, interesting. It did not happen with the old resolver. I don't think it is a major issue, dependabot tends to pick up updates quickly and we can always manually bump the version as part of a PR if we need to ensure all checks are green.

jku:
> Ah, interesting. It did not happen with the old resolver. I don't think it is a major issue, dependabot tends to pick up updates quickly and we can always manually bump the version as part of a PR if we need to ensure all checks are green.

Right. The alternative is that tox deps just list the normal requirements and then tox commands would replace securesystemslib with the git version by calling "pip install" (at least I assume that would work) ...

rzr:
This change is desirable:
https://github.com/theupdateframework/tuf/pull/1289/commits/2de2758393883ae0f96c30b43cc6b2b2107bf921

I picked it to build a snapshot debian package of development branch:
 
https://build.opensuse.org/package/show/home:rzrfreefr:devel/python-tuf

joshuagl (approved):
LGTM, thanks for not just adding a FIXME and forgetting about it.

PR #1293, opened 2021-03-02
Begin dropping support for Python 2.7

### LINKED ISSUE #1276: Make a final release supporting 2.7, then stop supporting EOL versions of Python

Cryptography is the latest one of our dependencies to [drop support for Python 2.7](https://cryptography.io/en/latest/changelog.html#v3-4), and one that seems unwise to try and pin. Given that the Python ecosystem, including many of our dependencies, continue to drop support for Python 2.7 it seems prudent to make a 'final' Python 2.7 supporting release so that we can stop worrying about whether dependency updates affect our ability to support Python 2.7, and instead can focus on [moving the codebase forwards](https://github.com/theupdateframework/tuf/projects/2).

**Current behavior**:

We continue to try and support Python 2.7 with the current codebase, despite many of our dependencies dropping support and that version of Python being EOL. This often results in pinning dependencies to Python 2.7 supported versions, which will prevent us from updating for security, and other bug, fixes.

**Expected behavior**:

Make one final release supporting Python 2.7, then drop support in future releases. We can create a branch from the tag of this final 2.7 supporting release should we want to integrate updates to that release in future.

**Tasks**
- [x] make a release of the current code
- [ ] drop Python 2.7 from `setup.py`, GitHub Actions workflows, requirements files, dependabot config, and other places.
- [ ] update documentation (including ADR0001?)

joshuagl:
Fixes #1276

This is a first pass at removing Python 2.7 from the codebase, focused on ensuring our next release does not claim to support Python 2.7 when we have no intention of doing so.

There are many more improvements that can be made in future, including (I think) to cleanups to the GitHub Actions Workflow file.

requirements**:


MVrachev:
Do we plan to remove the `python2` test files in this pr?
It seems like a logical thing to do given that you are removing `python2` as a test environment for tox.

From the top of my head the python2 tests we should remove are: 
- `test_proxy_use.py`
- `proxy_server.py`
- a unit  test in `test.utils.py` testing startup of `proxy_server.py`

joshuagl:
Thanks for taking a look at the PR Martin!

I hadn't removed the Python2 specific test files in this PR because I was aiming for the low hanging fruit that would prevent folks concluding that we continue to support Python2. However, now that you have pointed them out I will be happy to update the PR to also remove them.

setup.py
```diff
@@ -112,15 +110,12 @@
     'Source': 'https://github.com/theupdateframework/tuf',
     'Issues': 'https://github.com/theupdateframework/tuf/issues'
   },
-  python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4",
```
mnm678, line 115: Do we still want the `<4`?

mnm678 (approved):
One question, but this looks great!

MVrachev:
> Thanks for taking a look at the PR Martin!
> 
> I hadn't removed the Python2 specific test files in this PR because I was aiming for the low hanging fruit that would prevent folks concluding that we continue to support Python2. However, now that you have pointed them out I will be happy to update the PR to also remove them.

Glad to hear that! 

If you want, another check you can remove is `IS_PY_VERSION_SUPPORTED` in `test_api.py`.

setup.py
joshuagl: Good question! We retain this behaviour with `python_requires ="~=3.6"`, because the ~= clause indicates compatible releases. As 4 would be a major version change, that will not be compatible with 3.6.

See [[1](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires)] and [[2](https://www.python.org/dev/peps/pep-0440/#version-specifiers)].

jku (approved):
Some remaining things are 
* removing `from __future__ import ...` and 
* removing six dependency

After that there's various comments referencing python version issues and probably loads of little details that only made sense because of of python2.

None of the above should prevent this iteration from happening: LGTM!

... except, oops: I just merged teodoras conflicting changes

joshuagl:
Thank you for the review @mnm678 and @jku. I've filed issues for six (#1296) and `from __future__` (#1297) and rebased the branch on develop. Once CI is happy, I will merge.

PR #1302, opened 2021-03-11
Fix unclosed socket warning caused by RequestsFetcher

sechkova:

Close the Response object when an HTTP error is raised.

requirements**:


sechkova:
I also opened #1303 as a requests for further improvement.

jku (approved):
Agreed, this is correct. Would be great if we could somehow use the response as a contextmanager but ... I don't think it's possible considering we want the chunks()-generator to own and finally close the response. 

So LGTM.

PR #1322, opened 2021-03-24
Remove additional version settup in Signed

MVrachev:

There is no need for this assignment given that we are doing the same below after the if statement.


lukpueh (approved):
Great find! :)

PR #1327, opened 2021-03-31
test: Start coverage from python module

rzr:
Runing module from python will help
portability between systems.

Those scripts may be renamed by distros,
for instance Debian provides python3-converage.

Signed-off-by: Philippe Coval <rzr@users.sf.net>

that is provided, the better.

Fixes #<ISSUE NUMBER>

requirements**:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,8 +15,8 @@ changedir = tests
 
 commands =
     python --version
-    coverage run aggregate_tests.py
-    coverage report -m --fail-under 97
+    python -m coverage run aggregate_tests.py
+    python -m coverage report -m --fail-under 97
 
 deps =
     -r{toxinidir}/requirements-test.txt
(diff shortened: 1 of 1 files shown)


jku:
> Those scripts may be renamed by distros, for instance Debian provides python3-converage.

Alternatively Debian could just stop being weird when it's no longer necessary (Bullseye has no conflicts, /usr/bin/coverage could just exist) :)

Seriously speaking I think this change looks reasonable. That said, I think the tox configuration might not run on Debian even after this: calling "python" (without a version) does not work unless you've done some extra steps, right?

jku (approved):

rzr:
yes I have added a patch that add "3" suffix:
https://salsa.debian.org/python-team/packages/tuf/-/tree/debian/master/debian/patches

maybe this could be forwarded upstream because python2 is not more supported

note that there is Debian package python-is-python3 that add the symlink.

Issue #1230, opened 2020-12-01
New API: Revise "update"  methods in tuf/api/metadata.py


MVrachev:
**Description of issue or feature request**:
Currently, we are using a method called `update` to update certain fields of the metadata represented by our classes.
In `Timestamp` we are updating the information about the `meta` part from `signed` or more precisely the information about `snapshot.json`
In `Snapshot` we are updating the information about target.json or a specific delegation which are again in `meta` part from `signed` portion of the metadata file.
In `Targets` we are updating information about a specific target which is in `targets` part from `signed`  portion of the metadata file.

It's clear that by `update` the different classes are doing different stuff and it's a little ambiguous. One could easily ask the following questions:
In `Snapshot` with `update` do we update one or all files from `meta`?
In `Targets` with `update`  do we update the information about one or many target files?
In `Timestamp` with `update`  what are we updating? The `spec version`, the `expiry date`?

@lukpueh gave more insight with the following comment: 
```
I would say the name "update" in {Targets, Snapshot, Timestamp}.update(...) does not stand for 
"update attributes of an object" but rather for "update attributes of an object in the course 
of a repository update". 
I am talking about the Update in The Update Framework.

Right now a TUF integrator needs to call exactly these update methods explicitly 
whenever target files are added, thus the above meaning of update made sense to me.
```
here: https://github.com/theupdateframework/tuf/pull/1223#discussion_r532665409

We need to revise the need for the update functions, their naming, and purpose.

**Important Consideration**
We know that Warehouse is using our current version of `tuf/api/metadata` which means
they are using the `update` methods.
Resolving this issue will be a breaking change and we should be careful when to do it.

**Solution**
Revise the need of the `update` functions as a whole.
Maybe we can replace them with `add/remove` functions?

jku:
> in the future we will add a controller class which will handle the update.

Following is my gut feeling on this, not going to fight on these:
I think it makes sense to keep the metadata API unaware of the complexities of the client update process (and just include things that are truly common to all of the tools that need to handle metadata) -- but do agree that a client metadata controller is required _somewhere_: I estimate that that component will include most of the tricky bits of the updater code so I'd be inclined to make it an implementation detail of the updater (instead of part of the public API) at least in the beginning

jku:
> It's clear that by update the different classes are doing different stuff and it's little ambigious.

Yes, agreed 100%

[...]

PR #1329, opened 2021-03-31
New metadata API: add MetaFile and TargetFile classes

MVrachev:
This pr is a continuation from pr https://github.com/theupdateframework/tuf/pull/1223

This is the final pr planned for https://github.com/theupdateframework/tuf/issues/1139, 

In the top-level metadata classes, there are complex attributes such as
"meta" in Targets and Snapshot, "key" and "roles" in Root etc.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -29,6 +29,7 @@ from tuf.api.metadata import (
     Targets,
     Key,
     Role,
+    MetaFile,
     Delegations,
     DelegatedRole,
 )
@@ -246,28 +247,31 @@ class TestMetadata(unittest.TestCase):
                 self.repo_dir, 'metadata', 'snapshot.json')
         snapshot = Metadata.from_file(snapshot_path)
 
-        # Create a dict representing what we expect the updated data to be
-        fileinfo = copy.deepcopy(snapshot.signed.meta)
+        # Create a MetaFile instance representing what we expect
+        # the updated data to be.
         hashes = {'sha256': 'c2986576f5fdfd43944e2b19e775453b96748ec4fe2638a6d2f32f1310967095'}
-        fileinfo['role1.json']['version'] = 2
-        fileinfo['role1.json']['hashes'] = hashes
-        fileinfo['role1.json']['length'] = 123
+        fileinfo = MetaFile(2, 123, hashes)
 
-
-        self.assertNotEqual(snapshot.signed.meta, fileinfo)
+        self.assertNotEqual(
+            snapshot.signed.meta['role1.json'].to_dict(), fileinfo.to_dict()
+        )
         snapshot.signed.update('role1', 2, 123, hashes)
-        self.assertEqual(snapshot.signed.meta, fileinfo)
+        self.assertEqual(
+            snapshot.signed.meta['role1.json'].to_dict(), fileinfo.to_dict()
+        )
 
         # Update only version. Length and hashes are optional.
         snapshot.signed.update('role2', 3)
-        fileinfo['role2.json'] = {'version': 3}
-        self.assertEqual(snapshot.signed.meta, fileinfo)
+        fileinfo = MetaFile(3)
+        self.assertEqual(
+            snapshot.signed.meta['role2.json'].to_dict(), fileinfo.to_dict()
+        )
 
         # Test from_dict and to_dict without hashes and length.
         snapshot_dict = snapshot.to_dict()
-        test_dict = snapshot_dict['signed'].copy()
-        del test_dict['meta']['role1.json']['length']
-        del test_dict['meta']['role1.json']['hashes']
+        del snapshot_dict['signed']['meta']['role1.json']['length']
+        del snapshot_dict['signed']['meta']['role1.json']['hashes']
+        test_dict = copy.deepcopy(snapshot_dict['signed'])
         snapshot = Snapshot.from_dict(test_dict)
         self.assertEqual(snapshot_dict['signed'], snapshot.to_dict())
 
@@ -295,28 +299,33 @@ class TestMetadata(unittest.TestCase):
         timestamp.signed.bump_expiration(delta)
         self.assertEqual(timestamp.signed.expires, datetime(2036, 1, 3, 0, 0))
 
+        # Create a MetaFile instance representing what we expect
+        # the updated data to be.
         hashes = {'sha256': '0ae9664468150a9aa1e7f11feecb32341658eb84292851367fea2da88e8a58dc'}
-        fileinfo = copy.deepcopy(timestamp.signed.meta['snapshot.json'])
-        fileinfo['hashes'] = hashes
-        fileinfo['version'] = 2
-        fileinfo['length'] = 520
+        fileinfo = MetaFile(2, 520, hashes)
 
-        self.assertNotEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
+        self.assertNotEqual(
+            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+        )
         timestamp.signed.update(2, 520, hashes)
-        self.assertEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
+        self.assertEqual(
+            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+        )
 
         # Test from_dict and to_dict without hashes and length.
         timestamp_dict = timestamp.to_dict()
-        test_dict = timestamp_dict['signed'].copy()
-        del test_dict['meta']['snapshot.json']['length']
-        del test_dict['meta']['snapshot.json']['hashes']
+        del timestamp_dict['signed']['meta']['snapshot.json']['length']
+        del timestamp_dict['signed']['meta']['snapshot.json']['hashes']
+        test_dict = copy.deepcopy(timestamp_dict['signed'])
         timestamp_test = Timestamp.from_dict(test_dict)
         self.assertEqual(timestamp_dict['signed'], timestamp_test.to_dict())
 
         # Update only version. Length and hashes are optional.
         timestamp.signed.update(3)
-        fileinfo = {'version': 3}
-        self.assertEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
+        fileinfo = MetaFile(version=3)
+        self.assertEqual(
+            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+        )
 
     def test_key_class(self):
         keys = {
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -306,13 +306,6 @@ class TestMetadata(unittest.TestCase):
             snapshot.signed.meta['role1.json'].to_dict(), fileinfo.to_dict()
         )
 
-        # Update only version. Length and hashes are optional.
-        fileinfo = MetaFile(3)
-        snapshot.signed.update('role2', fileinfo)
-        self.assertEqual(
-            snapshot.signed.meta['role2.json'].to_dict(), fileinfo.to_dict()
-        )
-
         # Test from_dict and to_dict without hashes and length.
         snapshot_dict = snapshot.to_dict()
         del snapshot_dict['signed']['meta']['role1.json']['length']
@@ -365,13 +358,6 @@ class TestMetadata(unittest.TestCase):
         test_dict = copy.deepcopy(timestamp_dict['signed'])
         timestamp_test = Timestamp.from_dict(test_dict)
         self.assertEqual(timestamp_dict['signed'], timestamp_test.to_dict())
-
-        # Update only version. Length and hashes are optional.
-        fileinfo = MetaFile(version=3)
-        timestamp.signed.update(fileinfo)
-        self.assertEqual(
-            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
-        )
 
     def test_key_class(self):
         keys = {
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -30,6 +30,7 @@ from tuf.api.metadata import (
     Key,
     Role,
     MetaFile,
+    TargetFile,
     Delegations,
     DelegatedRole,
 )
@@ -540,22 +541,23 @@ class TestMetadata(unittest.TestCase):
             "sha512": "ef5beafa16041bcdd2937140afebd485296cd54f7348ecd5a4d035c09759608de467a7ac0eb58753d0242df873c305e8bffad2454aa48f44480f15efae1cacd0"
         },
 
-        fileinfo = {
-            'hashes': hashes,
-            'length': 28
-        }
+        fileinfo = TargetFile(length=28, hashes=hashes)
 
         # Assert that data is not aleady equal
-        self.assertNotEqual(targets.signed.targets[filename], fileinfo)
+        self.assertNotEqual(
+            targets.signed.targets[filename].to_dict(), fileinfo.to_dict()
+        )
         # Update an already existing fileinfo
-        targets.signed.update(filename, fileinfo)
+        targets.signed.update(filename, fileinfo.to_dict())
         # Verify that data is updated
-        self.assertEqual(targets.signed.targets[filename], fileinfo)
+        self.assertEqual(
+            targets.signed.targets[filename].to_dict(), fileinfo.to_dict()
+        )
 
         # Test from_dict/to_dict Targets without delegations
         targets_dict = targets.to_dict()
         del targets_dict["signed"]["delegations"]
-        tmp_dict = targets_dict["signed"].copy()
+        tmp_dict = copy.deepcopy(targets_dict["signed"])
         targets_obj = Targets.from_dict(tmp_dict)
         self.assertEqual(targets_dict["signed"], targets_obj.to_dict())
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -243,6 +243,51 @@ class TestMetadata(unittest.TestCase):
         self.assertFalse(is_expired)
         md.signed.expires = expires
 
+
+    def test_metafile_class(self):
+        # Test from_dict and to_dict with all attributes.
+        data = {
+            "hashes": {
+                "sha256": "8f88e2ba48b412c3843e9bb26e1b6f8fc9e98aceb0fbaa97ba37b4c98717d7ab"
+            },
+            "length": 515,
+            "version": 1
+        }
+        metafile_obj = MetaFile.from_dict(copy.copy(data))
+        self.assertEqual(metafile_obj.to_dict(), data)
+
+        # Test from_dict and to_dict without length.
+        del data["length"]
+        metafile_obj = MetaFile.from_dict(copy.copy(data))
+        self.assertEqual(metafile_obj.to_dict(), data)
+
+        # Test from_dict and to_dict without length and hashes.
+        del data["hashes"]
+        metafile_obj = MetaFile.from_dict(copy.copy(data))
+        self.assertEqual(metafile_obj.to_dict(), data)
+
+
+    def test_targetfile_class(self):
+        # Test from_dict and to_dict with all attributes.
+        data = {
+            "custom": {
+                "file_permissions": "0644"
+            },
+            "hashes": {
+                "sha256": "65b8c67f51c993d898250f40aa57a317d854900b3a04895464313e48785440da",
+                "sha512": "467430a68afae8e9f9c0771ea5d78bf0b3a0d79a2d3d3b40c69fde4dd42c461448aef76fcef4f5284931a1ffd0ac096d138ba3a0d6ca83fa8d7285a47a296f77"
+            },
+            "length": 31
+        }
+        targetfile_obj = TargetFile.from_dict(copy.copy(data))
+        self.assertEqual(targetfile_obj.to_dict(), data)
+
+        # Test from_dict and to_dict without custom.
+        del data["custom"]
+        targetfile_obj = TargetFile.from_dict(copy.copy(data))
+        self.assertEqual(targetfile_obj.to_dict(), data)
+
+
     def test_metadata_snapshot(self):
         snapshot_path = os.path.join(
                 self.repo_dir, 'metadata', 'snapshot.json')
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -301,14 +301,14 @@ class TestMetadata(unittest.TestCase):
         self.assertNotEqual(
             snapshot.signed.meta['role1.json'].to_dict(), fileinfo.to_dict()
         )
-        snapshot.signed.update('role1', 2, 123, hashes)
+        snapshot.signed.update('role1', fileinfo)
         self.assertEqual(
             snapshot.signed.meta['role1.json'].to_dict(), fileinfo.to_dict()
         )
 
         # Update only version. Length and hashes are optional.
-        snapshot.signed.update('role2', 3)
         fileinfo = MetaFile(3)
+        snapshot.signed.update('role2', fileinfo)
         self.assertEqual(
             snapshot.signed.meta['role2.json'].to_dict(), fileinfo.to_dict()
         )
@@ -353,7 +353,7 @@ class TestMetadata(unittest.TestCase):
         self.assertNotEqual(
             timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
         )
-        timestamp.signed.update(2, 520, hashes)
+        timestamp.signed.update(fileinfo)
         self.assertEqual(
             timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
         )
@@ -367,8 +367,8 @@ class TestMetadata(unittest.TestCase):
         self.assertEqual(timestamp_dict['signed'], timestamp_test.to_dict())
 
         # Update only version. Length and hashes are optional.
-        timestamp.signed.update(3)
         fileinfo = MetaFile(version=3)
+        timestamp.signed.update(fileinfo)
         self.assertEqual(
             timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
         )
@@ -593,7 +593,7 @@ class TestMetadata(unittest.TestCase):
             targets.signed.targets[filename].to_dict(), fileinfo.to_dict()
         )
         # Update an already existing fileinfo
-        targets.signed.update(filename, fileinfo.to_dict())
+        targets.signed.update(filename, fileinfo)
         # Verify that data is updated
         self.assertEqual(
             targets.signed.targets[filename].to_dict(), fileinfo.to_dict()
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -598,6 +598,60 @@ class Root(Signed):
         del self.keys[keyid]
 
 
+class MetaFile:
+    """A container with information about a particular metadata file.
+
+    Attributes:
+        version: An integer indicating the version of the metadata file.
+        length: An optional integer indicating the length of the metadata file.
+        hashes: An optional dictionary mapping hash algorithms to the
+            hashes resulting from applying them over the metadata file
+            contents.::
+
+                'hashes': {
+                    '<HASH ALGO 1>': '<METADATA FILE HASH 1>',
+                    '<HASH ALGO 2>': '<METADATA FILE HASH 2>',
+                    ...
+                }
+
+        unrecognized_fields: Dictionary of all unrecognized fields.
+
+    """
+
+    def __init__(
+        self,
+        version: int,
+        length: Optional[int] = None,
+        hashes: Optional[Dict[str, str]] = None,
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
+    ) -> None:
+        self.version = version
+        self.length = length
+        self.hashes = hashes
+        self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
+
+    @classmethod
+    def from_dict(cls, meta_dict: Dict[str, Any]) -> "MetaFile":
+        """Creates MetaFile object from its dict representation."""
+        version = meta_dict.pop("version")
+        length = meta_dict.pop("length", None)
+        hashes = meta_dict.pop("hashes", None)
+        # All fields left in the meta_dict are unrecognized.
+        return cls(version, length, hashes, meta_dict)
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Returns the dictionary representation of self."""
+        res_dict = {"version": self.version, **self.unrecognized_fields}
+
+        if self.length is not None:
+            res_dict["length"] = self.length
+
+        if self.hashes is not None:
+            res_dict["hashes"] = self.hashes
+
+        return res_dict
+
+
 class Timestamp(Signed):
     """A container for the signed part of timestamp metadata.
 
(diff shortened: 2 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -463,22 +463,64 @@ def remove_key(self, role: str, keyid: str) -> None:
             del self.keys[keyid]
 
 
+class MetadataInfo:
+    """A container with information about a particular metadata file.
+    Instances of MetadataInfo are used as values in a dictionary called
+    "meta" in Timestamp and Snapshot.
```
sechkova: Can you add a new line before the 'Attributes' section?

tuf/api/metadata.py
```diff
@@ -463,22 +463,64 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+        """Compare objects by their values instead of by their identities."""
+        return (
+            self.version == other.version
+            and self.length == other.length
+            and self.hashes == other.hashes
+        )
+
+    def to_dict(self) -> Dict:
+        """Returns the dictionary representation of self. """
+        json_dict = {"version": self.version}
```
sechkova: After abstracting the (de)serialization, other formats than `json` are possible and although the default is `json` it shouldn't be hardcoded.

tuf/api/metadata.py
```diff
@@ -463,22 +463,64 @@ def remove_key(self, role: str, keyid: str) -> None:
...
-                'snapshot.json': {
-                    'version': <SNAPSHOT METADATA VERSION NUMBER>,
-                    'length': <SNAPSHOT METADATA FILE SIZE>, // optional
-                    'hashes': {
-                        '<HASH ALGO 1>': '<SNAPSHOT METADATA FILE HASH 1>',
-                        '<HASH ALGO 2>': '<SNAPSHOT METADATA FILE HASH 2>',
-                        ...
-                    } // optional
-                }
+            'meta': {
```
sechkova: 'meta' is already listed as class attribute above.

tuf/api/metadata.py
```diff
@@ -489,38 +531,39 @@ def __init__(
...
-        # TODO: Add class for meta
         self.meta = meta
 
     @classmethod
     def from_dict(cls, timestamp_dict: Mapping[str, Any]) -> "Timestamp":
         """Creates Timestamp object from its dict representation. """
         common_args = cls._common_fields_from_dict(timestamp_dict)
-        meta = timestamp_dict.pop("meta")
+        meta = {
+            "snapshot.json": MetadataInfo(
```
sechkova: Same comment as above, related to hardcoding file formats. In this case if an implementation does not use `.json` files this will probably raise a `KeyError`.
Possibly the filename/format should be read from `timestamp_dict` instead.

tuf/api/metadata.py
```diff
@@ -628,13 +622,7 @@ def update(
     ) -> None:
         """Assigns passed (delegated) targets role info to meta dict. """
         metadata_fn = f"{rolename}.json"
```
sechkova, line 759: This is not your change but again, hardcoded `.json`.

tuf/api/metadata.py
```diff
@@ -463,22 +463,64 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+        self,
+        version: int,
+        length: Optional[int] = None,
+        hashes: Optional[Dict] = None,
+    ) -> None:
+        self.version = version
+        self.length = length
+        self.hashes = hashes
+
+    def __eq__(self, other: "MetadataInfo") -> bool:
```
sechkova: The `__eq__` implementation is needed only for the tests, right?
I guess it doesn't hurt having it :shrug: Maybe it will come handy in future.

tuf/api/metadata.py
```diff
@@ -625,22 +625,64 @@ def update(
...
+        """Compare objects by their values instead of by their identities."""
+        return (
+            self.length == other.length
+            and self.hashes == other.hashes
+            and self.custom == other.custom
+        )
+
+    def to_dict(self) -> Dict:
+        """Returns the JSON-serializable dictionary representation of self. """
+        json_dict = {"length": self.length, "hashes": self.hashes}
```
sechkova: Same comment about `json`.

sechkova (commented):
In the PR description you say that it will:
> Rename a few instances of the update method in our classes and use a more verbose name.

but there is only one commit with marked TODOs.

My general comment is about removing hardcoded file extensions and mentions of "json" throughout the code. After the abstract (de)serialization was implemented, TUF supposedly supports other file formats too.

tuf/api/metadata.py
MVrachev: Yes, currently `__eq__` is needed only for tests.

tuf/api/metadata.py
MVrachev: That's a really good catch!
Will make it work with other extensions as well!

MVrachev:
I rebased and fixed your two small comments @sechkova about other issues besides the hard-coded file names.

I fixed the problem of hardcoded `json` strings partially in the places I modified already.
To fully fix this problem we either have to use a more complicate solution or wait and consider other alternatives
which will give us the ability to set metadata extensions as a configuration option.

Created an issue about the hardcoded filename problems: https://github.com/theupdateframework/tuf/issues/1333

tuf/api/metadata.py
```diff
@@ -534,22 +534,41 @@ def __init__(
         super().__init__(_type, version, spec_version, expires)
         self.meta = meta
 
+    @classmethod
+    def _get_snapshot_filename(cls, meta_dict):
```
sechkova: This is a reasonable check but aren't we just assuming that input data is correct everywhere else? Since input validation is still a WIP? Anyway, probably better having it than not.

sechkova (approved):

tuf/api/metadata.py
MVrachev: Well the problem is that `meta_dict` sometimes is passed from the user - when calling `Timestamp.from_dict()`
and consequently when calling the `cls. _get_snapshot_filename()` method we haven't validated the `meta_dict` field.

At least, that is the case before implementing any new validation option as described in ADR 007.

tuf/api/metadata.py
```diff
@@ -458,22 +460,65 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+                    ...
+                }
+
+    """
+
+    def __init__(
+        self,
+        version: int,
+        length: Optional[int] = None,
+        hashes: Optional[Dict] = None,
```
jku: `Optional[Dict[str, str]]` I suppose

(much of the API uses Mapping in place of Dict though... I think that might be a mistake but maybe I've missed the reasons?)

tuf/api/metadata.py
```diff
@@ -484,35 +529,61 @@ def __init__(
...
     def from_dict(cls, timestamp_dict: Mapping[str, Any]) -> "Timestamp":
         """Creates Timestamp object from its dict representation. """
         common_args = cls._common_fields_from_dict(timestamp_dict)
-        meta = timestamp_dict.pop("meta")
+        snapshot_filename = cls._get_snapshot_filename(timestamp_dict["meta"])
+        meta = {
+            snapshot_filename: MetadataInfo(
+                **timestamp_dict["meta"][snapshot_filename]
+            )
+        }
```
jku: I think `get_snapshot_filename()` is not useful -- it does not do much and the ValueError is now hidden in an unexpected function. Overall we end up doing many (4?) dictionary lookups when I think one would be enough (it's not a performance issue or that important here but ... just more code than should be needed).

tuf/api/metadata.py
```diff
@@ -484,35 +529,61 @@ def __init__(
         version: int,
         spec_version: str,
         expires: datetime,
-        meta: Mapping[str, Any],
+        meta: Mapping[str, MetadataInfo],
```
jku: We could just have `snapshot_filename` and `snapshot_info` attributes instead of the dictionary... 

That way the API would be more usable as well: `timestamp.signed.meta["snapshot.json"].version` is ugly and won't autocomplete whereas `timestamp.signed.snapshot_info.version` is ... well, harder to typo at least.

tuf/api/metadata.py
jku: Something like this (assuming the attribute changes I proposed above, and popping being acceptable):
```
meta_dict = timestamp_dict["meta"]
# TODO check length
snapshot_filename, snapshot_dict = meta_dict.popitem()
snapshot_info = MetadataInfo(**snapshot_dict)
return cls(*common_args, snapshot_filename, snapshot_info)
```

tuf/api/metadata.py
```diff
@@ -548,23 +606,29 @@ def __init__(
...
-        # TODO: Add class for meta
         self.meta = meta
 
     @classmethod
     def from_dict(cls, snapshot_dict: Mapping[str, Any]) -> "Snapshot":
         """Creates Snapshot object from its dict representation. """
         common_args = cls._common_fields_from_dict(snapshot_dict)
-        meta = snapshot_dict.pop("meta")
+        meta = {}
+        for meta_path in snapshot_dict["meta"].keys():
```
jku: `for meta_path, meta_dict in snapshot_dict["meta"].items()` -- no need to do the lookup later

tuf/api/metadata.py
```diff
@@ -484,35 +529,61 @@ def __init__(
...
         return timestamp_dict
 
     # Modification.
     def update(
-        self, version: int, length: int, hashes: Mapping[str, Any]
+        self,
+        version: int,
+        length: Optional[int] = None,
+        hashes: Optional[Mapping[str, Any]] = None,
     ) -> None:
```
jku: I don't know where this is used so this may be dumb, but is this function needed?
Would `timestamp.signed.snapshot_info = MetadataInfo(version, length, hashes)` be equivalent to to calling this ?

(this assumes the snapshot_info change proposed above)

tuf/api/metadata.py
```diff
@@ -650,28 +752,38 @@ def __init__(
...
         self.delegations = delegations
 
     @classmethod
     def from_dict(cls, targets_dict: Mapping[str, Any]) -> "Targets":
         """Creates Targets object from its dict representation. """
         common_args = cls._common_fields_from_dict(targets_dict)
-        targets = targets_dict.pop("targets")
-        delegations = targets_dict.pop("delegations")
+        targets = {}
+        for target_path in targets_dict["targets"].keys():
```
jku: `for target_path, target_dict in targets_dict["targets"].items():`

tuf/api/metadata.py
```diff
@@ -680,4 +792,8 @@ def to_dict(self) -> Dict[str, Any]:
     # Modification.
     def update(self, filename: str, fileinfo: Mapping[str, Any]) -> None:
```
jku: Same question here, is this a useful function vs caller modifying targets.signed.targets directly

tuf/api/metadata.py
```diff
@@ -484,35 +529,61 @@ def __init__(
...
         timestamp_dict = self._common_fields_to_dict()
-        timestamp_dict.update({"meta": self.meta})
+        snapshot_filename = self._get_snapshot_filename(self.meta)
+        timestamp_dict.update(
+            {
+                "meta": {
+                    snapshot_filename: self.meta[snapshot_filename].to_dict()
+                }
+            }
+        )
```
jku: This works but is there a reason to use update over the simpler index access when you only have a single item?
```
 timestamp_dict["meta"] = {
     ...
}
```

jku (commented):
Apologies for not looking at this before. I've gone through most, but maybe not quite processed all of it.

Martin knows of this but for others information: the changes to `.pop()` usage are going to clash with another PR #1345: they need to be coordinated. In general I think this PR could have been several smaller instead -- I get that it means annoying depending PRs... but this way means I feel bad for leaving so many comments :grin:  

Would be great to see a bit more prose about the design and usability of the specific API classes somewhere: If we go through the trouble of creating objects for everything, we should try to make them more usable than the dictionaries they replace. As examples, see question in code about snapshot_info attribute and ideas like _'should Targets maybe implement abc.Mapping so caller could access targets.signed["target_name"] instead of targets.signed.targets["target_name"]'_ -- that last one may not be a good idea but it would be good to see some ideas and discussion before we start maintaining more code.

This feels like a lot to ask on a big PR but would be nice to see how the code now copes without broken input in tests (missing fields, extra fields, etc) -- to confirm the benefits that we should be getting from this work.

jku:
> Apologies for not looking at this before. I've gone through most, but maybe not quite processed all of it.

Also grumpy yesterday me sounds like he is unsatisfied with this PR but that's not true so I'll say it now:
PR looks correct to me (no bugs that I can find), good work, thank you

jku:
> Would be great to see a bit more prose about the design and usability of the specific API classes somewhere: If we go through the trouble of creating objects for everything, we should try to make them more usable than the dictionaries they replace

After further discussion, this opinion seems to go against previous decisions to keep the API as close to file format as possible -- so possibly we need to discuss further or implement this as "faithful reproduction of spec" (disregarding my comments) and think about potential usability improvements afterwards...

tuf/api/metadata.py
jku: the above comment was done when I wasn't aware/thinking of the earlier decision to try to map the API to file format as close as possible... in that light you might want to disregard this (I think I still want the usable version but that can be added later as a property)

MVrachev:
There were a lot of discussions after my last update on this pr.
I would prefer if we wait for #1345 and #1360 to be merged first.
Then, I will update/split this pr.

tuf/api/metadata.py
MVrachev: Agree, this could be subject to another set of changes.

tuf/api/metadata.py
MVrachev: I removed `get_snapshot_filename` from the latest rebased version.
I will think more about how to fix this approach when working on #1333

tuf/api/metadata.py
MVrachev: It will be the same I suppose, but I hope we will use this function to make some validation before doing the assignment.

tuf/api/metadata.py
MVrachev: It will be the same I suppose, but I hope we will use this function to make some validation before doing the assignment.

MVrachev:
I have rebased and made changes in this pr after "develop" has evolved a lot.

Some of the changes I made:
- rename `MetadataInfo` to `MetaFile` and `TargetInfo` to `TargetFile`. Those names seem better in my opinion because they communicate better what the classes represent.
- added support for unrecognized fields in `MetaFile` and `TargetFile`. I didn't add a test, because #1382 will add an automatic test for that for dictionaries and classes.
- dropped a couple of the planned commits after #1367 was merged.
- dropped a couple more commits, because I realized they were not needed - removing the side effect of destroying the dictionaries when calling `from_dict()` from the classes (this is needed for ADR 8) and adding a todo comment about the `update()` functions naming (there is an issue for that #1230)
- added `from_dict()` for `MetaFile` and `TargetFile`
- added `MetaFile` and `TargetFile` specific tests
- disabled the `C0302: Too many lines in module` warning for tuf/api/metadata.py.

@jku and @sechkova pr is again ready for a fresh review.

tuf/api/metadata.py
```diff
@@ -698,25 +741,31 @@ def __init__(
...
 
     @classmethod
     def from_dict(cls, snapshot_dict: Dict[str, Any]) -> "Snapshot":
         """Creates Snapshot object from its dict representation."""
         common_args = cls._common_fields_from_dict(snapshot_dict)
-        meta = snapshot_dict.pop("meta")
+        meta_dict = snapshot_dict.pop("meta")
+        meta = {}
+        for meta_path in meta_dict.keys():
+            meta[meta_path] = MetaFile.from_dict(meta_dict[meta_path])
```
jku: `for meta_path, metafile_dict in meta_dict.items():` or something like it maybe?

tuf/api/metadata.py
```diff
@@ -698,25 +741,31 @@ def __init__(
...
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self."""
         snapshot_dict = self._common_fields_to_dict()
-        snapshot_dict.update({"meta": self.meta})
+        meta_dict = {}
+        for meta_path, meta_info in self.meta.items():
+            meta_dict[meta_path] = meta_info.to_dict()
+
+        snapshot_dict.update({"meta": meta_dict})
```
jku: nit: `snapshot_dict["meta"] = meta_dict` seems more normal

tuf/api/metadata.py
```diff
@@ -897,35 +1002,42 @@ def __init__(
...
+        res_dict = self._common_fields_to_dict()
         if self.delegations:
-            targets_dict["delegations"] = self.delegations.to_dict()
-        return targets_dict
+            res_dict["delegations"] = self.delegations.to_dict()
+        targets = {}
+        for target_path, target_file_obj in self.targets.items():
+            targets[target_path] = target_file_obj.to_dict()
+
+        res_dict.update({"targets": targets})
```
jku: `res_dict["targets"] = targets`

tuf/api/metadata.py
```diff
@@ -897,35 +1002,42 @@ def __init__(
...
+        return res_dict
 
     # Modification.
     def update(self, filename: str, fileinfo: Dict[str, Any]) -> None:
         """Assigns passed target file info to meta dict."""
-        self.targets[filename] = fileinfo
+        self.targets[filename] = TargetFile.from_dict(fileinfo)
```
jku: This function should have TargetFile as argument, no?

tuf/api/metadata.py
```diff
@@ -861,22 +904,84 @@ def to_dict(self) -> Dict[str, Any]:
...
+        self.hashes = hashes
+        self.custom = custom
+        self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
+
+    def __eq__(self, other: "TargetFile") -> bool:
+        """Compare objects by their values instead of by their identities."""
+        return (
+            self.length == other.length
+            and self.hashes == other.hashes
+            and self.custom == other.custom
```
jku: missing unrecognised_fields

tuf/api/metadata.py
```diff
@@ -861,22 +904,84 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        unrecognized_fields: Dictionary of all unrecognized fields.
+
+    """
+
+    def __init__(
+        self,
+        length: int,
+        hashes: Dict[str, str],
+        custom: Optional[Dict[str, Any]] = None,
```
jku: It could be argued we don't need to handle custom at all in any of these functions: it's just an unrecognised field and already supported via that mechanism.

If we want to match the file format exactly we'd then need a "custom" property getter though (that refers to `unrecognised_fields.get("custom")`)

No strong opinions, maybe I lean a bit towards the property :shrug:

tuf/api/metadata.py
```diff
@@ -655,13 +717,7 @@ def update(
...
     ) -> None:
         """Assigns passed info about snapshot metadata to meta dict."""
-        self.meta["snapshot.json"] = {"version": version}
-
-        if length is not None:
-            self.meta["snapshot.json"]["length"] = length
-
-        if hashes is not None:
-            self.meta["snapshot.json"]["hashes"] = hashes
+        self.meta["snapshot.json"] = MetaFile(version, length, hashes)
```
jku: If we have this function it should take a Metafile as argument, IMO.

some other opinions on update():
* if we keep it a function, it should be renamed: "update" tells me nothing
* on the other hand I kind of think this should be a property instead, maybe "snapshot_meta" so you could do things like :
`version = timestamp.signed.snapshot_meta.version
`
or 
`timestamp.signed.snapshot_meta = MetaFile(ver, len, hashes)`

tuf/api/metadata.py
jku: I think a property checks all boxes: I'll leave a comment in the new code

tuf/api/metadata.py
jku: good point, thanks

tuf/api/metadata.py
jku: there may be an argument _against_ using helper properties like the one I handwaved: that it does not separate the attributes-that-map-the-file-format and the helpers enough (and that functions like `update()` are better in that regard). 

I think that's an issue we should consider but it does not outweigh the advantages (such as having a single read-and-write property with a descriptive name).

jku:
A question on the `__eq__` implementations: do we need them? None of the other classes seem to have them -- I assumed you used them in the tests but that does not seem the case (or maybe I just missed it)

tuf/api/metadata.py
MVrachev: About the `update()` naming there is an issue already opened capturing this #1230.

I think we should restrain ourselves from changing the API here (like changing the argument to `MetaFile` or adding properties), but focus more on the changes included in the pr. 

Instead, those could be discussed in the issue I referenced above and the issue area of focus could widen to discuss the `update()` functions design as a whole.

tuf/api/metadata.py
MVrachev: I like the idea of using custom as `unrecognized fields` through using a property. 
Will do it and will add a comment that custom will replace `unrecognized fields` for this particular case.

MVrachev: Wondering, if we use `custom` replacing the need of `unrecognized fields` for this class, then do you test for this in #1382?
Do you add an unrecognized field under the `custom` dictionary?

tuf/api/metadata.py
MVrachev: I still think this is not in the scope of this pr.
See my comment https://github.com/theupdateframework/tuf/pull/1329#discussion_r631716043.

MVrachev:
Updated the pr addressing almost all of the comments @jku made with the only exceptions when commenting on the `update` functions.
There I feel the changes Jussi proposed are out of scope for this pr as I described here https://github.com/theupdateframework/tuf/pull/1329#discussion_r631716043.

Additionally, I added a new commit where I remove the `__eq__` methods.

tuf/api/metadata.py
jku: I'm fine with not re-designing update here.

I disagree with leaving the argument as a dict. Why are we going through the trouble of adding objects and then not using them in the API?

tuf/api/metadata.py
jku: I'm fine with not completely re-designing update() methods in this PR.

I disagree with leaving the argument as a dict. Why are we going through the trouble of adding objects and then not using them in the API? 

I think your test code makes the case well: it does:
`targets.signed.update(filename, fileinfo.to_dict())`
and then in update you do 
` TargetFile.from_dict(fileinfo)`
it seems the argument should be TargetFile here

tuf/api/metadata.py
jku: Leaving update() re-design for later is fine.

I see no reason (so far) not to make the argument a MetaFile.

tuf/api/metadata.py
jku: I was trying to suggest dropping "custom", and just letting the already existing unrecognised_fields functionality handle it -- that basically handles de/serializing both "custom" and any other unrecognised fields.

Then I was saying that we might still want to add a property "custom" (to conform with the API design principle that the object API would copy the file format) but the property would just access `self.unrecognised_fields.get("custom")` and de/ser still would not handle "custom" at all -- it's handled like other unrecognised fields.

What you implemented is close but not quite the same... now TargetFile cannot have unrecognised fields at all (it can only have a "custom" field): as you correctly point out that means deciding that TargetFile is not allowed to have unrecognised fields and updating the test... which seems like the implementation detail guiding how we read the specification. 

I don't think this is a major issue, but I also think what I suggested would cover all the cases without any more complexity... opinions?

tuf/api/metadata.py
MVrachev: Okay, will change the argument type for the `update` functions, but won't rename it or add additional attributes as you suggested above.

tuf/api/metadata.py
MVrachev: I commented about the same question here:
https://github.com/theupdateframework/tuf/pull/1329#discussion_r632555437

tuf/api/metadata.py
MVrachev: Well, `custom` basically is the same as `unrecognised_fields`.
I feel like we should encourage our users to use `custom` as described in the spec when adding additional fields instead of relying on `unrecognized_fields` at least in this scenario.
Maybe we can achieve that by adding a `custom` property, but then if we want to treat it as a regular attribute maybe we would have to add a property setter as well?

On another hand, if we don't have `unrecognized_fields` in TargetFile, target.json file with unrecognized fields outside custom won't work...

jku: > Well, `custom` basically is the same as `unrecognised_fields`.
> I feel like we should encourage our users to use `custom` as described in the spec when adding additional fields instead of relying on `unrecognized_fields` at least in this scenario.
> Maybe we can achieve that by adding a `custom` property, but then if we want to treat it as a regular attribute maybe we would have to add a property setter as well?
> 
> On another hand, if we don't have `unrecognized_fields` in TargetFile, target.json file with unrecognized fields outside custom won't work...

that's what I'm saying: if we read the spec as "you need to be able to add new fields to any dictionary where it's technically possible" (which was the consensus I believe), then  implementing custom is not enough for spec-compliance

I don't think it's a major thing but ... if we can easily achieve spec compliance in a trivial matter then why not do it?

MVrachev: Okay, will update the class, so that `unrecognized fields` is documented and added as an attribute, but I am not sure should we document `custom` when `custom` will be just a property and we retrieve with `self.unrecognised_fields.get("custom")`?

MVrachev: I didn't define it in my last update. Do you think that's okay?

tuf/api/metadata.py
```diff
@@ -624,42 +675,33 @@ def __init__(
...
         # All fields left in the timestamp_dict are unrecognized.
         return cls(*common_args, meta, timestamp_dict)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self."""
         timestamp_dict = self._common_fields_to_dict()
-        timestamp_dict.update({"meta": self.meta})
+        timestamp_dict.update(
+            {"meta": {"snapshot.json": self.meta["snapshot.json"].to_dict()}}
+        )
```
jku: please, no need to use update when a simple `timestamp_dict["meta"] = ...` works

tuf/api/metadata.py
```diff
@@ -859,22 +882,69 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Returns the JSON-serializable dictionary representation of self."""
+        res_dict = {
+            "length": self.length,
+            "hashes": self.hashes,
+        }
+
+        if self.custom is not None:
+            res_dict["custom"] = self.custom
```
jku: Should you have unrecognised_fields handling here instead of handling custom?

tests/test_api.py
```diff
@@ -241,33 +243,83 @@ def test_metadata_base(self):
...
+
+
     def test_metadata_snapshot(self):
         snapshot_path = os.path.join(
                 self.repo_dir, 'metadata', 'snapshot.json')
         snapshot = Metadata.from_file(snapshot_path)
 
         # Create a dict representing what we expect the updated data to be
-        fileinfo = copy.deepcopy(snapshot.signed.meta)
+        fileinfo = copy.deepcopy(snapshot.signed.meta['role1.json'])
```
jku: This is not a dict like comment says but a MetaFile object, I think?

Is there a reason for this deepcopy and then modifying data -- couldn't you just create the object with MetaFile() using your arguments to make it clear what it is?

tests/test_api.py
```diff
@@ -241,33 +243,83 @@ def test_metadata_base(self):
...
+        )
 
         # Update only version. Length and hashes are optional.
-        snapshot.signed.update('role2', 3)
-        fileinfo['role2.json'] = {'version': 3}
-        self.assertEqual(snapshot.signed.meta, fileinfo)
+        fileinfo = MetaFile(3)
+        snapshot.signed.update('role2', fileinfo)
+        self.assertEqual(
+            snapshot.signed.meta['role2.json'].to_dict(), fileinfo.to_dict()
```
jku: Just a nit but mentioning in case there's something missing: This test doesn't really test much anything anymore (it used to compare two separate dicts, now it's pretty much comparing an object to itself)

tests/test_api.py
```diff
@@ -297,26 +349,32 @@ def test_metadata_timestamp(self):
...
-        fileinfo['hashes'] = hashes
-        fileinfo['version'] = 2
-        fileinfo['length'] = 520
-
-        self.assertNotEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
-        timestamp.signed.update(2, 520, hashes)
-        self.assertEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
+        fileinfo.hashes = hashes
+        fileinfo.version = 2
+        fileinfo.length = 520
```
jku: nit, This would be much more understandable as `fileinfo = MetaInfo(...)` as well

tests/test_api.py
```diff
@@ -297,26 +349,32 @@ def test_metadata_timestamp(self):
...
 
         # Update only version. Length and hashes are optional.
-        timestamp.signed.update(3)
-        fileinfo = {'version': 3}
-        self.assertEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
+        fileinfo = MetaFile(version=3)
+        timestamp.signed.update(fileinfo)
+        self.assertEqual(
+            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+        )
```
jku: same nit as before, this doesn't test much anything anymore

jku (changes requested):
Everything else is a style comment (feel free to disregard) but I think there's a bug in the "custom" handling

tuf/api/metadata.py
MVrachev: I realized that I don't need to handle custom specifically anymore.
"custom", if exist, will be part of `unrecognized fields and my `to_dict` function could be just:
```
        """Returns the JSON-serializable dictionary representation of self."""
        return {
            "length": self.length,
            "hashes": self.hashes,
            **self.unrecognized_fields,
        }
```

tests/test_api.py
MVrachev: Yes, I understand this. The test by itself makes sure that we are actually supporting optional `hashes` and `length`.
Will remove the test in the `fix update argument` commit. Then, we won't need this, because length and hashes will be optional because of the `MetaFile` implementation, no the update function itself.

MVrachev:
Addressed all of @jku commets, and additionally I wanted to make sure that each of the commits passes the CI and makes sense.
That's why I also made the following changes:
1. Merge changes in "Replace usage of dict.copy with copy.deepcopy()" in "Add MetaFile" and "Add TargetFile". Then drop "Replace usage of dict.copy with copy.deepcopy()"
2. Simplified how we handle "custom" in `TargetFile.to_dict()`
3. Squash "New API: represent custom with unrecognized_fields" into "Add TargetFile"
4. Added a new commit removing two tests testing "update with only version" because they are obsolete.

tuf/api/metadata.py
jku: I see `__eq__()` is still here (which is fine -- maybe you use it in tests) but it's still not complete: it should compare unrecognized_fields instead of custom

jku (approved):
I think my comment in-source won't show up since it's in a resolved discussion but:

The` __eq__()` implementation is still not complete: if it is there, it should compare unrecognized_fields instead of custom.

Altogether this looks good, thanks. The docstrings are very verbose but let's trim those later... I'm marking this approved already, but please fix `__eq__()` still in this PR.

MVrachev:
> I think my comment in-source won't show up since it's in a resolved discussion but:
> 
> The` __eq__()` implementation is still not complete: if it is there, it should compare unrecognized_fields instead of custom.
> 
> Altogether this looks good, thanks. The docstrings are very verbose but let's trim those later... I'm marking this approved already, but please fix `__eq__()` still in this PR.

I thought I had removed the `__eq__` methods.
It was a mistake that they are left there.
Removed them and rebased.

jku:
I was about to merge but now it has conflicts with a bug fix I merged yesterday. Sorry. Could you rebase?

MVrachev:
Done.

PR #1330, opened 2021-04-01
fix the outdate link in docs/tuf-spec.0.9.txt

maoyangLiu:
I find the link in docs/tuf-spec.0.9.txt has outdate,so I fix it


jku (changes requested):
The issue is valid, thanks!

The fix is not complete though: the tuf-spec.0.9.txt is trying to link to the _historical version 0.9 of the spec_, not the current one as the PR does. https://github.com/theupdateframework/specification/blob/master/historical/tuf-spec.0.9.txt might be the correct url. 

Also 
* I'd like to see just a single commit, not five (force pushing your PR branch is fine: it'll update here)
* we use a developer certificate of origin and would like to see a signed-off-by in the commits, see https://github.com/theupdateframework/tuf/blob/develop/docs/CONTRIBUTORS.rst

jku (approved):
LGTM, thanks!

Issue #1297, opened 2021-03-03
Remove uses of `from __future__ import` 


joshuagl:
**Description of issue or feature request**:

Following the release of tuf 0.17.0 (see https://github.com/theupdateframework/tuf/pull/1284) we are no longer supporting Python2, and have started to remove uses of Python2 in #1293 (see also #1276).

Various incompatible changes were made available to Python 2.7 via the [`__future__`](https://docs.python.org/3/library/__future__.html), where importing from the module would ensure that features worked the same across Python 2 and 3.

Now that we are no longer supporting Python 2 we do not need to import from the future module.

**Current behavior**:

`from __future__ import` pattern is used in 168 places in the tuf codebase.

**Expected behavior**:

`from __future__ import` pattern is not used in the code.

PR #1331, opened 2021-04-06
WIP: Mirrors redesign

sechkova:
Addresses: #1307

Regrouping of client-related modules:

- moved components related to file download inside the `client_rework` directory
- applied black and isort on the modules with the old code style
-  `download.py` is deleted and its functionality is merged with `mirrors.py` which is now `mirrors_download.py`
- code is wrapped in `Mirrors` class

requirements**:


tuf/client_rework/mirrors.py
```diff
@@ -22,174 +22,162 @@
   of the file with respect to the base url.
 """
 
+
 # Help with Python 3 compatibility, where the print statement is a function, an
 # implicit relative import is invalid, and the '/' operator performs true
 # division.  Example:  print 'hello world' raises a 'SyntaxError' exception.
```
MVrachev: Do we want to copy those comments and the future imports too?
I know we are going to remove them when resolving this issue https://github.com/theupdateframework/tuf/issues/1297,
but do we want to add them in new files as well?

tuf/client_rework/mirrors.py
```diff
@@ -22,174 +22,162 @@
...
-                   file_mirror_errors)
+_SUPPORTED_FILE_TYPES = ["meta", "target"]
+
+
+class Mirrors:
+    def __init__(self, mirrors_dict: Dict):
+        tuf.formats.MIRRORDICT_SCHEMA.check_match(mirrors_dict)
+        self._config = mirrors_dict
+
+    def _get_list_of_mirrors(self, file_type, file_path):
```
MVrachev: Do you plan to add type annotations to the old functions as well which were moved or only on the newly added ones?

tuf/client_rework/mirrors.py
```diff
@@ -34,25 +34,33 @@
...
-from typing import BinaryIO, Dict, TextIO
+from typing import BinaryIO, Dict, Optional, TextIO
 
 import securesystemslib
 import six
 
 import tuf
 import tuf.client_rework.download as download
 import tuf.formats
+from tuf.requests_fetcher import RequestsFetcher
```
MVrachev: I think you should import the `FetcherInterface` as well, because later on in the `Mirrors.__init__` you use the "FetcherInterface" as a type annotation.
Locally I see warnings because of this.

tuf/client_rework/mirrors_download.py
```diff
@@ -164,7 +164,10 @@ def meta_download(self, filename: str, upper_length: int) -> TextIO:
                 temp_obj.seek(0)
                 yield temp_obj
 
-            except Exception as exception:
+            # pylint cannot figure out that we store the exceptions
+            # in a dictionary to raise them later so we disable
+            # the warning. This should be reviewed in the future still.
```
MVrachev, line 169: Maybe add a `TODO` if you expect us to review this in the future?

MVrachev (changes requested):
The only questionable choice I see is the one with the naming of the class `Mirrors`.

Of the 6 methods in the `Mirrors` class, 4 of them have the word `download` in them.
I think this is a good sign that the class name is better to be just `Downloader`.
Yes, for download were using mirrors, but that seems like an API detail giving
that we don't have a non-mirrors downloader class.

One would say that `Fetcher` and `Downloader` are close names but calling the class
`Mirrors` just hides that proximity.  

Other than that, I have some small comments, but nothing too serious.

tuf/client_rework/fetcher.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2021, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Provides an interface for network IO abstraction.
+"""
+
+# Imports
```
jku, line 7: Nit: I think these _Imports/Classes_ comments are not useful.

sechkova:
Closing this one in favour of #1341 and #1352

PR #1340, opened 2021-04-08
Remove use of six

### LINKED ISSUE #1296: Remove use of six

Following the release of tuf 0.17.0 (see https://github.com/theupdateframework/tuf/pull/1284) we are no longer supporting Python2, and have started to remove uses of Python2 in #1293 (see also #1276).

[Six](https://six.readthedocs.io) is a Python 2 and 3 compatibility library, which is clearly no longer required.

**Current behavior**:

Six is marked as a dependency in setup.py and our requirements files.
Six is imported and used in multiple modules, tests, and scripts.

**Expected behavior**:

six is no longer a dependency and is not imported, or used, in any modules, tests, or scripts.

avelichka:
Removes the use of six so it is no longer a dependency and is not imported and used in any modules, tests, or scripts.

Fixes #1296

requirements**:


jku:
Just two comments (unfortunately one touches a lot of lines):
* I believe `import urllib` does not bring `urllib.request` or `urllib.parse` into scope -- I think some dependency (requests?) maybe happens to import those so we don't notice this? If I'm correct, I'd be fine with `import urllib.request` style in tests so you don't need to modify all calls again but in the actual code some form of `from urllib import x` would probably be preferred for vendoring reasons 
* `for key, value in dict.items(mydict):` is not wrong... but wouldn't `for key, value in mydict.items():` be the normal way to do this? I hope there is a nice regex to do that replace if you end up doing it.

avelichka:
Thank you for the review, @jku! Hopefully, I've managed to address your comments in the two new commits.

jku (approved):

jku:
Thanks, LGTM! 

If you want to squash these into one commit and force-push feel free to do so, otherwise I'll do a "squash and merge" (either way is fine by me but in the latter case I can only mark you as co-author as the new commit will then be mine).

avelichka:
I've squashed them into one commit. Thank you, @jku.

Issue #1334, opened 2021-04-06
Use lazy (old style %-format) log formatting in the new pylint config


sechkova:
**Description of issue or feature request**:
Using f-strings (and all types of string formatting except %-syle) in logging messages is not lazy evaluated so linters "recommend" %-style formatting.

Should we disable the related pylint warnings and use f-strings:
```
logging-not-lazy
logging-format-interpolation
logging-fstring-interpolation
```
or adhere to the recommendation.

**Current behavior**:
Top-level `pylintrc` disables logging that recommend lazy formatting globally.
New `api/pylintrc` does not.

**Expected behavior**:
Decide on a common approach.

sechkova:
An example approach to avoid the warnings while using f-strings which feels more like a workaround: 438c0a0

jku:
It's annoying to use a different formatting method for logging than anything else but the reasoning is solid... Most of our logging is not in a "hot loop" but some debug logs really might be used quite a lot (and the workaround only hides the warning, not the potential performance issue in those cases).

I think using %-style is fine for logging: `logger.debug("processing, a=%s", a)` but it's not a strong opinion -- in the opposite case we may want to do a quick review to identify hottest logging calls though.

sechkova:
And then there is [this issue](https://github.com/PyCQA/pylint/issues/3362) when trying to use %-style formatting but you've set up `pylint` to `logging-format-style=new`.

jku:
I had understood that using %-style implies means we would need to use `logging-format-style=old` anyway (or rather not set it as that's the default)? Pylint docs really could have more examples.

jku:
I'm proposing we use the default logging style, meaning: 
* use `logging-format-style=old` (which is default value)
* stop disabling `logging-not-lazy` (note: this is done only in some experimental client branches -- metadata api itself hasn't needed this because there's zero logging so far :grimacing: )

It looks a bit uglier (other string formatting is f-strings but logging is then based on %-formatting) but:
* it's a common style
* it's suggested by the google style guide
* it means we don't have to think about performance when writing logging code

joshuagl:
Proposal above makes sense to me, if it's what's suggested by the style guide we probably don't even need to capture anything for this? Let's just do what our style guide suggests.

PR #1341, opened 2021-04-08
Client refactor: code reformatting

sechkova:

- Move file download related modules inside client directory
- Remove (un)safe_download functions
- Apply black and isort to the newly moved files
- Apply the new pylint config from api/pylintrc and fix the various linter issues
- Remove the use of six and future and make the rest of the imports vendoring-compatible

requirements**:

### DIFF
--- a/tuf/client_rework/download.py
+++ b/tuf/client_rework/download.py
@@ -131,11 +131,10 @@ def download_file(url, required_length, fetcher, strict_required_length=True):
                 )
                 break
 
-            else:
-                logger.debug(
-                    "The average download speed has not dipped below the"
-                    " minimum average download speed set in tuf.settings.py."
-                )
+            logger.debug(
+                "The average download speed has not dipped below the"
+                " minimum average download speed set in tuf.settings.py."
+            )
 
         # Does the total number of downloaded bytes match the required length?
         _check_downloaded_length(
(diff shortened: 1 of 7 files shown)


tuf/client_rework/updater_rework.py
```diff
@@ -147,71 +148,22 @@ def download_target(self, target: Dict, destination_directory: str):
         The file is saved to the 'destination_directory' argument.
         """
 
-        try:
-            for temp_obj in self._mirror_target_download(target):
+        for temp_obj in mirrors._mirror_target_download(
+            target, self._mirrors, self._fetcher
+        ):
+
+            try:
```
jku: commit message says it only moves code but this obviously changes things... We already have an issue on figuring out the download methods design (and fixing the bugs it has with multiple mirrors) so I'm not expecting this commit to fix everything -- but I also don't understand  what this change does?

tuf/client_rework/download.py
```diff
@@ -99,7 +99,8 @@ def download_file(url, required_length, fetcher, STRICT_REQUIRED_LENGTH=True):
     # encoded as %5c in the url, which should also be replaced with a forward
     # slash.
     url = six.moves.urllib.parse.unquote(url).replace("\\", "/")
-    logger.info("Downloading: " + repr(url))
+    msg = f"Downloading: {url}"
+    logger.info(msg)
```
jku: I disagree with the whole logging-not-lazy commit: Working around warning like this is not useful.
* if we want to use f-strings with logging, let's silence the warning in the linter config
* alternatively, let's just use %-style strings (I lean this way)

tuf/client_rework/mirrors.py
```diff
@@ -110,7 +99,7 @@ def get_list_of_mirrors(file_type, file_path, mirrors_dict):
     in_confined_directory = securesystemslib.util.file_in_confined_directories
 
     list_of_mirrors = []
-    for dummy, mirror_info in six.iteritems(mirrors_dict):
+    for dummy, mirror_info in mirrors_dict.items():
```
jku: nit: dummy is not even required `for mirror_info in mirrors_dict.values()`

tuf/client_rework/download.py
```diff
@@ -23,24 +23,13 @@
...
-    absolute_import,
-    division,
-    print_function,
-    unicode_literals,
-)
-
 import logging
 import tempfile
 import timeit
+import urllib
```
jku: `from urllib import parse` maybe -- and then use parse.unquote() etc. Current code only works by accident since some other module happens to import the right thing

jku (changes requested):
Thanks, looks mostly good to me:
* disagree with logging-not-lazy (we can discuss though)
* The *_target_download() functions still make me itch -- whatever we do here let's next concentrate on getting those fixed

tuf/client_rework/updater_rework.py
sechkova: Thanks for catching this, it was a mistake after cherry-picking this commit from another branch.
I had to rebase and force push to fix this. This is the new commit: 35587c4

sechkova:
Force pushed the new commits. The differences are in 
- 35587c4, thanks for catching the error
- 2f0bbd0
- Dropped the "logging-not-lazy" commit and simply ignored the warning with 101ab3d 

I chose the lazy way of just ignoring the logging warning now, since there is already an issue capturing the need of a decision: #1334

jku (approved):
LGTM, thanks. 

The issues left IMO are:
* download implementation is fishy
* the internal file locations may not be final (I think they should be in _internal/ or something)
but we have these in writing already.

So let's go with this and see where we end up :)

PR #1345, opened 2021-04-09
New metadata API: add support for ADR 0008

MVrachev:

In order to support ADR 0008 we would want to accept unrecognized
fields in all metadata classes.
Input that contains unknown fields in the 'signed' dictionary should
successfully deserialize into a Metadata object, and that object should
successfully serialize with the unknown fields intact.

Also, we should test that we support unrecognized fields when adding
new classes or modifying existing ones to make sure we support
ADR 0008.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -330,6 +330,37 @@ class TestMetadata(unittest.TestCase):
         # Verify that data is updated
         self.assertEqual(targets.signed.targets[filename], fileinfo)
 
+    def setup_dict_with_unrecognized_field(self, file_path, field, value):
+        json_dict = {}
+        with open(file_path) as f:
+            json_dict = json.loads(f.read())
+        # We are changing the json dict without changing the signature.
+        # This could be a problem if we want to do verification on this dict.
+        json_dict["signed"][field] = value
+        return json_dict
+
+    def test_support_for_unrecognized_fields(self):
+        for metadata in ["root", "timestamp", "snapshot", "targets"]:
+            path = os.path.join(self.repo_dir, "metadata", metadata + ".json")
+            dict1 = self.setup_dict_with_unrecognized_field(path, "f", "b")
+            # Test that the metadata classes store unrecognized fields when
+            # initializing and passes them when casting the instance to a dict.
+
+            temp_copy = copy.deepcopy(dict1)
+            metadata_obj = Metadata.from_dict(temp_copy)
+
+            self.assertEqual(dict1["signed"], metadata_obj.signed.to_dict())
+
+            # Test that two instances of the same class could have different
+            # unrecognized fields.
+            dict2 = self.setup_dict_with_unrecognized_field(path, "f2", "b2")
+            temp_copy2 = copy.deepcopy(dict2)
+            metadata_obj2 = Metadata.from_dict(temp_copy2)
+            self.assertNotEqual(
+                metadata_obj.signed.to_dict(), metadata_obj2.signed.to_dict()
+            )
+
+
 # Run unit test.
 if __name__ == '__main__':
     utils.configure_test_logging(sys.argv)
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -300,6 +300,7 @@ class Signed:
         spec_version: The TUF specification version number (semver) the
             metadata format adheres to.
         expires: The metadata expiration datetime object.
+        unrecognized_fields: Dictionary of all unrecognized fields.
 
     """
 
@@ -307,7 +308,12 @@ class Signed:
     # we keep it to match spec terminology (I often refer to this as "payload",
     # or "inner metadata")
     def __init__(
-        self, _type: str, version: int, spec_version: str, expires: datetime
+        self,
+        _type: str,
+        version: int,
+        spec_version: str,
+        expires: datetime,
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
 
         self._type = _type
@@ -318,6 +324,7 @@ class Signed:
         if version < 0:
             raise ValueError(f"version must be >= 0, got {version}")
         self.version = version
+        self.unrecognized_fields = unrecognized_fields or {}
 
     @staticmethod
     def _common_fields_from_dict(signed_dict: Mapping[str, Any]) -> list:
@@ -349,6 +356,7 @@ class Signed:
             "version": self.version,
             "spec_version": self.spec_version,
             "expires": self.expires.isoformat() + "Z",
+            **self.unrecognized_fields,
         }
 
     # Modification.
@@ -409,8 +417,11 @@ class Root(Signed):
         consistent_snapshot: bool,
         keys: Mapping[str, Any],
         roles: Mapping[str, Any],
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
-        super().__init__(_type, version, spec_version, expires)
+        super().__init__(
+            _type, version, spec_version, expires, unrecognized_fields
+        )
         # TODO: Add classes for keys and roles
         self.consistent_snapshot = consistent_snapshot
         self.keys = keys
@@ -423,7 +434,8 @@ class Root(Signed):
         consistent_snapshot = root_dict.pop("consistent_snapshot")
         keys = root_dict.pop("keys")
         roles = root_dict.pop("roles")
-        return cls(*common_args, consistent_snapshot, keys, roles)
+        # All fields left in the root_dict are unrecognized.
+        return cls(*common_args, consistent_snapshot, keys, roles, root_dict)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self. """
@@ -485,8 +497,11 @@ class Timestamp(Signed):
         spec_version: str,
         expires: datetime,
         meta: Mapping[str, Any],
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
-        super().__init__(_type, version, spec_version, expires)
+        super().__init__(
+            _type, version, spec_version, expires, unrecognized_fields
+        )
         # TODO: Add class for meta
         self.meta = meta
 
@@ -495,7 +510,8 @@ class Timestamp(Signed):
         """Creates Timestamp object from its dict representation. """
         common_args = cls._common_fields_from_dict(timestamp_dict)
         meta = timestamp_dict.pop("meta")
-        return cls(*common_args, meta)
+        # All fields left in the timestamp_dict are unrecognized.
+        return cls(*common_args, meta, timestamp_dict)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self. """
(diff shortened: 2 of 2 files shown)


tests/test_api.py
```diff
@@ -330,6 +333,32 @@ def test_metadata_targets(self):
         # Verify that data is updated
         self.assertEqual(targets.signed.targets[filename], fileinfo)
 
+    def setup_json_dict_with_unrecognized_fields(self, file_path):
+        storage_backend = FilesystemBackend()
+        json_dict = {}
+        with storage_backend.get(file_path) as file_obj:
+            raw_data = file_obj.read()
+            json_dict = json.loads(raw_data.decode("utf-8"))
```
jku: does using FileSystemBackend give us something or is this same as:
```
with open(file_path) as f:
   json_dict = json.loads(f.read())
```

tuf/api/metadata.py
```diff
@@ -361,6 +361,20 @@ def bump_version(self) -> None:
         self.version += 1
 
 
+def _get_unrecognized_fields(
+    all_fields: Mapping[str, Any], recognized_fields: list
+) -> Mapping[str, Any]:
+    """Utility function to get the additional unrecognized fields from a dict.
+    Read ADR 0008."""
```
jku: ADRs are hopefully nice for checking "why was this done the way it was done" in the future but I don't want to have to read any to understand the code -- I think mentioning the ADR in the commit messages might be enough?

This applies to all mentions of the adr: if this functionality needs explaining I think we should explain it in a comment.

tuf/api/metadata.py
```diff
@@ -409,12 +426,14 @@ def __init__(
...
         keys: Mapping[str, Any],
         roles: Mapping[str, Any],
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
         super().__init__(_type, version, spec_version, expires)
         # TODO: Add classes for keys and roles
         self.consistent_snapshot = consistent_snapshot
         self.keys = keys
         self.roles = roles
+        self.unrecognized_fields = unrecognized_fields
```
jku: this really feels like it should be in Signed... what if 
* Signed had a class variable known_fields (that already includes the common fields) and the derived classes can redefine the variable to match their known fields
* _get_unrecognised_fields() is a class method in Signed (and now does not need expected_fields because previous point)
* Signed constructor takes unrecognised_fields as argument

does that make sense at all?

tuf/api/metadata.py
```diff
@@ -485,22 +523,35 @@ def __init__(
...
+        unrecognized_fields = _get_unrecognized_fields(
+            timestamp_dict, expected_fields
+        )
+        return cls(*common_args, meta, unrecognized_fields)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self. """
         timestamp_dict = self._common_fields_to_dict()
-        timestamp_dict.update({"meta": self.meta})
+        timestamp_dict.update({"meta": self.meta, **self.unrecognized_fields})
```
jku: I think this will fail if you create Timestamp using the constructor with unrecognized_fields unset. Maybe the instance variable should not be Optional (set to {} if argument is None)

jku: should we add a test for pure constructor as well (as in not using `from_dict()`)?

jku (commented):
I'm a bit out of my depth with metadata API still but let's try:
* Please don't expect commit message reader to read the ADR: mentioning it is good but especially the title should explain what the commit does
* I wonder if the repetition could be fixed by making the implementation part of Signed -- I left more details in the code

Other comments are more trivial

tests/test_api.py
MVrachev: Well, it doesn't give me anything.
I just forgot I can just use the above.

MVrachev:
Thank you @jku for your great review!
I amended my last commit to fix the commit message and used the opportunity to apply your suggestions.

I particularly found [that comment](https://github.com/theupdateframework/tuf/pull/1345#discussion_r614083992) really useful which simplified the code a lot!
Also, it seems to me that [this comment](https://github.com/theupdateframework/tuf/pull/1345#discussion_r614096817) is no longer applicable because I don't have `unrecognized_fields` as an argument for the `Timestamp` constructor.

jku (commented):
Some code comments to get them out of the way:
* The derived classes can just set the known_fields class variable in the class definition, no need to have the new code in `_get_unrecognized_fields()`
* `_get_unrecognized_fields` should probably be `_unrecognised_fields_from_dict()` to match the existing function for common fields.

But before you get to those: there may be an alternative solution that means we don't need to track known fields... If the code uses pop() consistently when looking up fields from dict  (as it seems to) then whatever is left after collecting all known fields is unrecognised_fields.

MVrachev:
> But before you get to those: there may be an alternative solution that means we don't need to track known fields... If the code uses pop() consistently when looking up fields from dict (as it seems to) then whatever is left after collecting all known fields is unrecognised_fields.

Now, it makes sense why Lukas decided to use `pop` versus just accessing the fields.
I used that suggestion in my latest changes squashed everything into one commit.
It does look better now, even though I am not such a fan of the side effects when using `from_dict`.

tuf/api/metadata.py
```diff
@@ -300,14 +300,22 @@ class Signed:
         spec_version: The TUF specification version number (semver) the
             metadata format adheres to.
         expires: The metadata expiration datetime object.
+        unrecognized_fields: An optional dictionary storing all unrecognized
+            fields. Used for backward compatibility. If None is provided, an
+            empty dictionary will be created.
```
```suggestion
        unrecognized_fields: Dictionary of all unrecognized fields.
```
The attribute is not optional I think (the constructor argument is). Also this isn't specifically for backwards (forwards?) compat: IIRC Spec requires this.

tuf/api/metadata.py
```diff
@@ -423,7 +438,11 @@ def from_dict(cls, root_dict: Mapping[str, Any]) -> "Root":
         consistent_snapshot = root_dict.pop("consistent_snapshot")
         keys = root_dict.pop("keys")
         roles = root_dict.pop("roles")
-        return cls(*common_args, consistent_snapshot, keys, roles)
+        # All fields left in the root_dict and unrecognized.
```
jku: typo and/are?  Same comment applies to other `from_dict()` methods

jku (commented):
Looks fine to me, some small comments inline. Also this part of the commit message: 

> including the classes that would be
> added representing a subportion of a role like "meta", "delegations"
> and "roles".

I'm not quite sure what this means... maybe something like "Input that contains unknown fields in the 'signed' dictionary should succesfully deserialize into a Metadata object, and that object should succesfully serialize with the unknown fields intact".

The only other thing I'm wondering is if we need more substantial tests (more complex content?) I don't know what this would accomplish so I guess I'm fine with the tests as they are...

MVrachev:
Fixed all of the comments you mentioned and the commit message.

jku (approved):
Thanks for working on the iterations even when feedback came in dripping... result looks nice to me.

I still feel like the testing is pretty light... but I'm also not sure what to suggest so :heavy_check_mark:

MVrachev:
> Thanks for working on the iterations even when feedback came in dripping... result looks nice to me.
> 
> I still feel like the testing is pretty light... but I'm also not sure what to suggest so ✔️

Not sure what other tests should I add.
I make sure that between calling `from_dict` and `to_dict` there is no change and I thought this is the idea right?
Is there another property we want to check for?

jku:
> Not sure what other tests should I add.

Just to clarify: this is just fine for me.

jku:
So after I was schooled yesterday on how the unrecognised fields are planned to operate (that unrecognised fields are allowed to appear at many levels of the file format hierarchy -- e.g. tap15 adds a field into delegations) I've started thinking the testing could reflect that from the beginning: the testing proposed here is very focused on a specific place in the file format

Is it possible to add a test that inserts extra fields at every possible location in our file format and does a basic serialize/deserialize test? The naive attempt would be to go through the json and add a bogus field to every dictionary... but there are exceptions: METAFILES in timestamp.json is explicitly not allowed to have extra fields. But maybe you could add extra fields to every dict except a hand-crafted list of exceptions?

This could be a new issue if it looks like a major change... but implementing this might make the future object creation PRs easier (as test would exist already)

tuf/api/metadata.py
```diff
@@ -318,6 +324,9 @@ def __init__(
         if version < 0:
             raise ValueError(f"version must be >= 0, got {version}")
         self.version = version
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+        self.unrecognized_fields = unrecognized_fields
```
joshuagl: We can reduce this to a one line statement with `self.unrecognized_fields = unrecognized_fields or {}`, but I don't know whether we _should_?

```
>>> foo = {'a':3, 'b':2, 'c':1}
>>> bar = None
>>> baz = foo or {}
>>> baz
{'a': 3, 'b': 2, 'c': 1}
>>> baz = bar or {}
>>> baz
{}
```

tuf/api/metadata.py
```diff
@@ -423,7 +436,11 @@ def from_dict(cls, root_dict: Mapping[str, Any]) -> "Root":
         consistent_snapshot = root_dict.pop("consistent_snapshot")
         keys = root_dict.pop("keys")
         roles = root_dict.pop("roles")
-        return cls(*common_args, consistent_snapshot, keys, roles)
+        # All fields left in the root_dict are unrecognized.
+        unrecognized_fields = root_dict
```
joshuagl: Given the comment, do we need the assignment to `unrecognized_fields` (here and in the other constructors)?

joshuagl (approved):
I really appreciate the iteration here, the end result looks good.

I left a couple of optional suggestions that would reduce the lines of code in metadata.py by ~7LOC. Nothing major.

joshuagl:
> Is it possible to add a test that inserts extra fields at every possible location in our file format and does a basic serialize/deserialize test? The naive attempt would be to go through the json and add a bogus field to every dictionary... but there are exceptions: METAFILES in timestamp.json is explicitly not allowed to have extra fields. But maybe you could add extra fields to every dict except a hand-crafted list of exceptions?
> 
> This could be a new issue if it looks like a major change... but implementing this might make the future object creation PRs easier (as test would exist already)

This sounds like a good test to add indeed. I'd be happy to see it in a future PR, does not have to be this one.

MVrachev:
> So after I was schooled yesterday on how the unrecognised fields are planned to operate (that unrecognised fields are allowed to appear at many levels of the file format hierarchy -- e.g. tap15 adds a field into delegations) I've started thinking the testing could reflect that from the beginning: the testing proposed here is very focused on a specific place in the file format
> 
> Is it possible to add a test that inserts extra fields at every possible location in our file format and does a basic serialize/deserialize test? The naive attempt would be to go through the json and add a bogus field to every dictionary... but there are exceptions: METAFILES in timestamp.json is explicitly not allowed to have extra fields. But maybe you could add extra fields to every dict except a hand-crafted list of exceptions?
> 
> This could be a new issue if it looks like a major change... but implementing this might make the future object creation PRs easier (as test would exist already)

Yes, that's true and I am adding unrecognized fields and tests for each of the subfields I am replacing with a new class.
See #1360.

Also, I addressed the two comments by @jku. 
Thank you both for your patience and detail reviews!

jku:
> Yes, that's true and I am adding unrecognized fields and tests for each of the subfields I am replacing with a new class.
> See #1360.

So just to make it clear: I'm trying to propose that we don't try to cover each case individually, but instead do it in an automated manner. No need to block this PR though.

PR #1347, opened 2021-04-12
Add a method of Signed metadata class returning information about metadata expiration 

### LINKED ISSUE #1305: Add 'expired' method to Signed class
Add a method of `Signed` metadata class returning information about metadata expiration given a reference time.

**Current behavior**:
`Signed` class already contains expires, a metadata expiration datetime object. 

**Expected behavior**:
A method exists that checks the 'expires' attribute against a reference time.
`def is_expired(self, reference_time: float) -> bool`

avelichka:
Adds the **is_expired** method to the **Signed** metadata class that checks the **expires** attribute against a reference time and returns information about metadata expiration 

that is provided, the better.

Fixes #1305

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -219,7 +219,25 @@ class TestMetadata(unittest.TestCase):
         md.signed.bump_expiration(timedelta(days=365))
         self.assertEqual(md.signed.expires, datetime(2031, 1, 2, 0, 0))
 
+        # Test is_expired with reference_time provided
+        is_expired = md.signed.is_expired(md.signed.expires)
+        self.assertTrue(is_expired)
+        is_expired = md.signed.is_expired(md.signed.expires + timedelta(days=1))
+        self.assertTrue(is_expired)
+        is_expired = md.signed.is_expired(md.signed.expires - timedelta(days=1))
+        self.assertFalse(is_expired)
 
+        # Test is_expired without reference_time, 
+        # manipulating md.signed.expires
+        expires = md.signed.expires
+        md.signed.expires = datetime.utcnow()
+        is_expired = md.signed.is_expired()
+        self.assertTrue(is_expired)
+        md.signed.expires = datetime.utcnow() + timedelta(days=1)
+        is_expired = md.signed.is_expired()
+        self.assertFalse(is_expired)
+        md.signed.expires = expires
+        
     def test_metadata_snapshot(self):
         snapshot_path = os.path.join(
                 self.repo_dir, 'metadata', 'snapshot.json')
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -351,6 +351,22 @@ class Signed:
             "expires": self.expires.isoformat() + "Z",
         }
 
+    def is_expired(self, reference_time: datetime = None) -> bool:
+        """Checks metadata expiration against a reference time.
+
+        Args:
+            reference_time: Optional; The time to check expiration date against.
+                A naive datetime in UTC expected.
+                If not provided, checks against the current UTC date and time.
+
+        Returns:
+            True if expiration time is less than the reference time.
+        """
+        if reference_time is None:
+            reference_time = datetime.utcnow()
+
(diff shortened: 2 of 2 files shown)


tests/test_api.py
```diff
@@ -219,6 +219,13 @@ def test_metadata_base(self):
         md.signed.bump_expiration(timedelta(days=365))
         self.assertEqual(md.signed.expires, datetime(2031, 1, 2, 0, 0))
 
+        # Test is_expired
+        is_exired = md.signed.is_expired(md.signed.expires)
+        self.assertEqual(is_exired, True)
```
jku: just a nit but: typo in is_expired

tuf/api/metadata.py
```diff
@@ -351,6 +351,10 @@ def _common_fields_to_dict(self) -> Dict[str, Any]:
             "expires": self.expires.isoformat() + "Z",
         }
 
+    def is_expired(self, reference_time=datetime.now()) -> bool:
```
jku: reference_time should be type annotated, no? Also more importantly this looks like a [mutable default argument](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) and may be a bug

jku (changes requested):

tuf/api/metadata.py
sechkova: In addition, `datetime.now()` returns local time and date. The [recommended way](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) to get UTC time seems to be:
`datetime.now(timezone.utc)`

jku: This is a good point ...  but I _think_ our expiry datetimes are timezone-naive so datetime.utcnow() might actually be what we want (or alternatively we should be creating expiry datetime as timezone-aware)

jku: well the tests seem happy so maybe I was wrong and our expiry datetime is not naive?

avelichka: Ops, I've just pushed a commit with what @sechkova suggested but I will correct it if needed.
@jku, I've addressed your comments in separate commits that I'm going to squash as soon as you review and approve the changes.

tuf/api/metadata.py
```diff
@@ -351,6 +351,26 @@ def _common_fields_to_dict(self) -> Dict[str, Any]:
...
+        <Arguments>
+            reference_time:
+                The time to check the metadata expiration date against.
+                A naive datetime in UTC expected.
+                If not provided, checks against the current UTC date and time.
+
+        <Returns>
+            True if the reference time is equal or
+            greater than the metadata expiration date.
+        """
```
jku: Thanks this helps... but it's the old style doc format which we're hoping to not use in api/

```suggestion
        """Checks metadata expiration against a reference time.

        Args:
            reference_time: Optional; The time to check expiration date against.
                A naive datetime in UTC expected.
                If not provided, checks against the current UTC date and time.

        Returns:
            True if expiration time is less than the reference time.
        """
```

I think this matches the current style (https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) ... but I've never used this style yet either so maybe @MVrachev or someone more familiar can confirm if this is what we want for new code?
In any case let's try to keep it to as few lines as possible.

jku (changes requested):
Looks good, except for the docstring format (TUF is full of the format you used but we're trying to move to standard docstrings). I _think_ I have a suggestion that's correct below... 

Feel free to squash all of this into a single commit (btw, commit message titles preferred under 50 characters).

tuf/api/metadata.py
sechkova: Looks good, should be ok with the new style.

avelichka: Thank you all for the comments! I believe I addressed all issues and pushed a squashed commit. I promise to carefully go through the guidelines again and to do better with observing best practices and recommendations next time :)

jku (approved):
Thank you

Issue #1339, opened 2021-04-08
experimental client: download functions review


jku:
Current state of _mirror_target_download() and _mirror_meta_download() in experimental client is a bit bad:
* they are actually iterators that keep returning unverified file objects to the caller, this seems unusual and surprising: I was expecting them to return a verified file object
* caller is expected to verify the contents --- this makes some sense as the verification is different for e.g. root metadata and other metadata and targets... but it should be made very clear that the file object is potentially malicious data at this point
* none of the callers currently iterate these function results (they just raise if first result is invalid) so it's not clear if mirrors  actually work
* NoWorkingMirrorError can't be raised from these iterators like it now happens (on every iteration that fails), it does not make any sense 
* _mirror_target_download type annotation is wrong

I would suggest we revert to a known working design if possible (`_mirror_target_download()` could just be almost 1:1 the current clients `_get_target_file()` function but the metadata verification needs something else).

I'll try to come up with something...

jku:
this is very much related to #1318 -- the tricky file object handling is in these two functions and in the callers

jku:
I think this is a good place to document the discussion from yesterday (@sechkova can fix mistakes): We've come to the conclusion that separating the download/mirror functionality as I suggested does not work better than the "old way".

There's two things interacting here: the need to download a file from possibly multiple mirrors and the validation and verififcation of that file -- we don't know if we need to download from other mirrors until we've fully validated/verified the file from the first mirror. This is especially an issue for metadata which needs to not just be internally validated but also verified by other current metadata... This means that there are really two choices:
* mirror-downloader can verify the files (maybe via a verifier object it gets or something): this means the updater loop is now slightly easier to read but obfuscates the logic as it's hard to see when/where the verification happens
* the updater loop knows about all of the candidate downloads (as current updater does)
* (extra 3rd choice: metadata should not support mirrors. This would make things simpler and I think no-one wants to mirror metadata anyway... but we've not considered this seriously for now)

The second option just seems better: the updater download loops are monsters but it seems taking this specific for-loop out is not helpful. experimental-client sort of does the second option but with iterators that I think are not an improvement -- we've agreed to take a step back to something even more similar to current updater (no iterators, just a list of mirror-urls : code already known to work).

To make the very long updater loops more readable by other means I'm experimenting with #1313  (taking the metadata validity enforcing elsewhere).

trishankatdatadog:
[...]

Issue #1318, opened 2021-03-18
client refactor: review file object handling


jku:
There's a lot more manual file closing (`temp_obj.close()`) than I'd like and I can already see bugs: the same file object gets closed multiple times on some exceptions.

Let's do a comprehensive review and aim to A) get rid of as many manual `close()`s as we can and B) make the remaining ones more obviously correct

jku:
I'll take this as an excuse to get more familiar with the code

jku:
This one is done already: file object handling has been vastly reduced in the experimental client

Issue #1313, opened 2021-03-12
Keep track of currently valid set of metadata in Updater


sechkova:
**Description of issue or feature request**:

Define and init sequence for Updater.
Define what is the currently valid set of metadata and is  refresh() the only wait to achieve it.
Is keeping current/previous metadata enough for this?
Do we need a VCS-like functionality?

See https://github.com/theupdateframework/tuf/pull/1291#issuecomment-793607482, https://github.com/theupdateframework/tuf/pull/1291#issuecomment-792612877

Sync with #1308

jku:
So what I've been playing with is a "metadata bundle" that tracks clients current valid set of metadata. Purpose is to:
 * Make it easy to verify that the TUF spec is being followed while making the mostrous update loops readable
 * Make sure clients whole set of metadata is consistent when new metadata is added
 * Persist the metadata on disk / load metadata from disk

Explicitly purpose is not to handle:
 * downloading
 * file-like objects (from downloads)
 * traversing delegation tree
 * modifying individual metadata in any way 

I hope this abstraction will make it easier for a reader to go through the metadata verification steps, and for a test system to test it (as downloads etc are no longer needed).  This is definitely influenced by @trishankatdatadog's Repository class in tuf-on-a-plane (with some differences). 
 
Hypothetical/simplified usage example in an update loop:

```
# load local root, raises if root not present
bundle = MetadataBundle("/my/local/repo")

for version in range(bundle.root.version + 1, upper_bound):
    # raises if download fails
    tempfile = download_next_root_file()

    # raises if metadata is invalid
    unverified_root = Metadata(tempfile.read())

    # raises if metadata fails to verify with current metadata 
    bundle.root = unverified_root
```

I'll update when I have something that runs

jku:
After some quick experiments these properties seem to make sense and would work:
 * Verified metadata is always persisted on disk ASAP
 * there is a single definition of "now": this decides all expiry checks. This moment is likely decided by the final root metadata update (this is the first time an expiry check is made during the process)
 * hard-coded "now" means there is no "re-loading" of stale metadata: If new files are needed, another bundle should be created 
 * Bundle is never in an inconsistent state wrt metadata (it is never possible to have a delegate in bundle that is not signed by the delegator)
 * Possible states are:
   - local root loaded
     - new roots can be loaded from remote
   - root load finished
     - other local metadata (if valid) is also loaded
     - timestamp can now be loaded from remote
   - timestamp loaded
[...]

PR #1352, opened 2021-04-15
client refactor: simplify mirrors loop

sechkova:
Addresses: #1318, #1339 

This set of changes tries to simplify the code and make it easy to follow as a result it:

- Removes  mirror_*_download functions and moves the mirror's loop inside the Updater
- Replaces file objects used as function arguments for metadata files, which are read in memory right after download, anyway. Temporary files are now handled in a single function.
- Does small code cleanup

requirements**:

### DIFF
--- a/tuf/client_rework/updater_rework.py
+++ b/tuf/client_rework/updater_rework.py
@@ -10,7 +10,7 @@ TODO
 import fnmatch
 import logging
 import os
-from typing import BinaryIO, Dict, Optional, TextIO
+from typing import Dict, Optional
 
 from securesystemslib import exceptions as sslib_exceptions
 from securesystemslib import hash as sslib_hash
@@ -159,9 +159,9 @@ class Updater:
                 temp_obj = download.download_file(
                     file_mirror, target["fileinfo"]["length"], self._fetcher
                 )
-
+                _check_file_length(temp_obj, target["fileinfo"]["length"])
                 temp_obj.seek(0)
-                self._verify_target_file(temp_obj, target)
+                _check_hashes_obj(temp_obj, target["fileinfo"]["hashes"])
                 break
 
             except Exception as exception:  # pylint:  disable=broad-except
@@ -308,7 +308,7 @@ class Updater:
                 )
 
                 temp_obj.seek(0)
-                intermediate_root = self._verify_root(temp_obj)
+                intermediate_root = self._verify_root(temp_obj.read())
                 # When we reach this point, a root file has been successfully
                 # downloaded and verified so we can exit the loop.
                 break
@@ -356,7 +356,7 @@ class Updater:
                 )
 
                 temp_obj.seek(0)
-                verified_timestamp = self._verify_timestamp(temp_obj)
+                verified_timestamp = self._verify_timestamp(temp_obj.read())
                 break
 
             except Exception as exception:  # pylint:  disable=broad-except
@@ -410,7 +410,7 @@ class Updater:
                 )
 
                 temp_obj.seek(0)
-                verified_snapshot = self._verify_snapshot(temp_obj)
+                verified_snapshot = self._verify_snapshot(temp_obj.read())
                 break
 
             except Exception as exception:  # pylint:  disable=broad-except
@@ -465,7 +465,7 @@ class Updater:
 
                 temp_obj.seek(0)
                 verified_targets = self._verify_targets(
-                    temp_obj, targets_role, parent_role
+                    temp_obj.read(), targets_role, parent_role
                 )
                 break
 
@@ -487,12 +487,12 @@ class Updater:
             self._get_full_meta_name(targets_role, extension=".json")
         )
 
-    def _verify_root(self, temp_obj: TextIO) -> RootWrapper:
+    def _verify_root(self, file_content: bytes) -> RootWrapper:
         """
         TODO
         """
 
-        intermediate_root = RootWrapper.from_json_object(temp_obj)
+        intermediate_root = RootWrapper.from_json_object(file_content)
 
         # Check for an arbitrary software attack
         trusted_root = self._metadata["root"]
@@ -505,7 +505,6 @@ class Updater:
 
         # Check for a rollback attack.
         if intermediate_root.version < trusted_root.version:
-            temp_obj.close()
             raise exceptions.ReplayedMetadataError(
                 "root", intermediate_root.version(), trusted_root.version()
             )
@@ -514,11 +513,11 @@ class Updater:
 
         return intermediate_root
 
-    def _verify_timestamp(self, temp_obj: TextIO) -> TimestampWrapper:
+    def _verify_timestamp(self, file_content: bytes) -> TimestampWrapper:
         """
         TODO
         """
-        intermediate_timestamp = TimestampWrapper.from_json_object(temp_obj)
+        intermediate_timestamp = TimestampWrapper.from_json_object(file_content)
 
         # Check for an arbitrary software attack
         trusted_root = self._metadata["root"]
@@ -532,7 +531,6 @@ class Updater:
                 intermediate_timestamp.signed.version
                 <= self._metadata["timestamp"].version
             ):
-                temp_obj.close()
                 raise exceptions.ReplayedMetadataError(
                     "root",
                     intermediate_timestamp.version(),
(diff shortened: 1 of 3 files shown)


jku:
One of the reasons for using a file object (a streaming object if you will) is that we have not needed to read the file in memory in all cases. With metadata that's a little moot as we do have to read it to verify and process... but with targets _not_ loading the whole file is a very good idea. So I think the the file object commit is not good. This also means the updater_rework _check_hashes() is wrong.

jku:
This is not really a review (sorry) but maybe background for future discussion:

Moving the download loop back to updater looks like the right move (in other words my original idea was bad): I can understand this code :) . However I'm not sure the idea of separate loops is better than what the current/old updater does. The download loop is not trivial (we've had multiple bugs with file objects etc) so is multiplying it going to help?

I think the underlying idea with your design is that 
*  in current updater is very hard to follow how e.g. specific metadata gets verified (what actually gets checked before new root metadata is accepted)
* this becomes easier to understand when the root download code is in its own function and not intermingled with e.g. snapshot download code

This is valid... but I think the cost of duplicating the download code  is too high -- this is visible in root download where you already had to refactor that one to another function. I think having a generic metadata download function is fine we just have to figure out how to make it clear what is verified when: I'm hoping #1313 will help there -- my vision of what would happen is something like this:
 * root download loop starts going through versions N+1
   * it calls generic metadata download loop that start downloading files from mirrors 
     * for each candidate file, it tries to create a new metadata object: this verifies the internal state
     * and tries to insert the object into the current valid metadata bundle (aka local repository controller) as new root: this does most(?) of the required tuf spec checks in the root update part
     * if anything fails we continue the loop,  otherwise our metadata bundle is now using the new root and we can break the loop



Another observation:
* the Updater._verify_*() functions are named confusingly: they create a Metadata object, then they do additional verification and **return the object**. It's not a verifying method, it's a factory method to create a Metadata object that meets some criteria

tuf/client_rework/updater_rework.py
```diff
@@ -213,97 +209,157 @@ def _load_root(self) -> None:
...
+            # than the expiration timestamp in the trusted root metadata file
+            # TODO define which exceptions are part of the public API
+            intermediate_root.expires()
+
+            # 1.9. If the timestamp and / or snapshot keys have been rotated,
+            # then delete the trusted timestamp and snapshot metadata files.
+            if self._metadata["root"].keys(
+                "timestamp"
+            ) != intermediate_root.keys("timestamp"):
+                # FIXME: use abstract storage
```
MVrachev, line 267: Could we use TODO instead of FIXME?
I did a quick search in the tuf repo and found only 1 place in the code, besides this file, where we are using those.
It seems we aren't using FIXME in our code and if (hopefully) one day we decide to fix our TODOs this will be left out.

MVrachev:
> This is valid... but I think the cost of duplicating the download code is too high -- this is visible in root download where you already had to refactor that one to another function. I think having a generic metadata download function is fine we just have to figure out how to make it clear what is verified when: I'm hoping #1313 will help there -- my vision of what would happen is something like this:

I was going to write the same thing and then I saw your comment.
I also feel like there is too much repetition and I hope we find another way to do this.
Will follow your `Metadata bundle` suggestion.

sechkova:
Agree that now when mirror download code is moved back to Updater the code is more readable but also makes the repetition obvious. However replacing it with a single function with the current state of the Updater code does not result in the desired clean implementation. I propose keeping this changes as a base to test the work proposed in #1313, see how it fits in and at the same time reduce the code repetition.
I force pushed, fixed a bug that @jku noticed in the mirrors loop and also split the `_check_hashes` function in two: one correctly dealing with file objects for target files download and one using file content read in memory for metadata files.

tuf/client_rework/updater_rework.py
```diff
@@ -147,22 +147,30 @@ def download_target(self, target: Dict, destination_directory: str):
...
-            # TODO: do something with exceptions
-            raise
+                _check_file_length(temp_obj, target["fileinfo"]["length"])
+                temp_obj.seek(0)
+                _check_hashes_obj(temp_obj, target["fileinfo"]["hashes"])
+                break
+
+            except Exception as exception:
+                if temp_obj:
+                    temp_obj.close()
```
jku, line 174: this applies to every case with close(): current client does `temp_obj = None` here (inside the if-clause) and it makes sense I think: otherwise temp_obj might be defined on next iteration even if it should not be, leading to us calling close twice... 

Calling close() twice should not break things so this probably does not fail anything but it looks fishy

In fact now that you've made the metadata handling bytes-based (and not file object), all of the metadata download loops could just use a download method that returns bytes -- but it does not need to happen in this PR. Maybe adding temp_obj = None is more straight forward for now?

tuf/client_rework/updater_rework.py
```diff
@@ -147,22 +147,30 @@ def download_target(self, target: Dict, destination_directory: str):
...
+                raise exceptions.NoWorkingMirrorError({file_mirror: exception})
```
jku: this should store the exception like the other functions, no?

tuf/client_rework/updater_rework.py
```diff
@@ -759,19 +834,34 @@ def _check_file_length(file_object, trusted_file_length):
         )
 
 
-def _check_hashes(file_object, trusted_hashes):
+def _check_hashes_obj(file_object, trusted_hashes):
+    """
+    TODO
+    """
```
jku, line 855: just comment: I've filed https://github.com/theupdateframework/tuf/issues/1361 for an easy to use solution for both cases here

tuf/client_rework/updater_rework.py
```diff
@@ -759,19 +834,34 @@ def _check_file_length(file_object, trusted_file_length):
...
+    for algorithm, trusted_hash in trusted_hashes.items():
+        digest_object = sslib_hash.digest_fileobject(file_object, algorithm)
+
+        computed_hash = digest_object.hexdigest()
+
+        # Raise an exception if any of the hashes are incorrect.
+        if trusted_hash != computed_hash:
+            raise sslib_exceptions.BadHashError(trusted_hash, computed_hash)
```
jku: we have a BadHashError of our own and should not raise sslib errors

jku (commented):
So previous comment still stands 

> I think the cost of duplicating the download code is too high

... but we have a plan:
 
* get this code into a readable form again (this PR) even if it keeps duplicating a lot of code
* reduce support for mirrors (as it really is what generates all of this complexity, and no-one needs most of it) : Teodora is going to document that plan in https://github.com/theupdateframework/tuf/issues/1307
* after that we will hopefully be able to simplify this code drastically
* at this point also metadata bundle can be simplified https://github.com/theupdateframework/tuf/pull/1355

After these steps we will be able to evaluate whether the branch has value or if we should just consider it lessons learned and start over...

So with this idea in mind I've reviewed again: I think I found one actual issue in target_download and then some comments...

tuf/client_rework/updater_rework.py
sechkova: You're right, thanks for noticing. I added `temp_obj = None` for now, since changing the download function would mean a difference between metadata and targets. I'll keep this in mind.

tuf/client_rework/updater_rework.py
sechkova: Yes, fixed, thanks

sechkova:
Oops my "fixup" commit is not signed off ... but I apparently want to rebase and squash this one after your review. 
On the other hand, do I add the linter fix by running the latest black version in this PR?

jku:
Looks good enough for me -- let's get this merged so we can see if streamlining mirrors helps...

I'll merge develop into experimental-client now (#1365): After that if you'd like to rebase this on experimental-client (to get the linter fix) and also rebase --autosquash (to squash the fixup commit), that works for me.

jku (approved):

sechkova:
Thanks, it should be ready now.

tuf/client_rework/updater_rework.py
jku: I don't really want to use abstract storage in the client (or make it a priority, even if Metadata supports it) but as this part of code is hopefully going to move to bundle anyway, I'm not looking at this sort of things too closely now

PR #1354, opened 2021-04-16
api: Add Metadata.from_bytes()

### LINKED ISSUE #1336: Implement from_object/from_string method in Metadata class
**Description of feature request**:
Implement a class method in `Metadata` class, similar to the existing `from_file` method but taking a file object as an argument.

See #1304 for context.

jku:
This is essentially short-hand for

    JSONDeserializer().deserialize(data)

but seems easier for the API user so may be worth it.

Metadata.from_file() now uses Metadata.from_bytes() internally.

--- 
_(message and title edited later to replace string with bytes)_

Fixes #1336

I think adding this makes sense in that we're trying to "hide" serialization details from the default use case and currently that wasn't possible if you just have a string (and not a file).

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -105,8 +105,7 @@ class TestMetadata(unittest.TestCase):
             path = os.path.join(self.repo_dir, 'metadata', metadata + '.json')
             metadata_obj = Metadata.from_file(path)
             with open(path, 'rb') as f:
-                metadata_str = f.read()
-            metadata_obj2 = JSONDeserializer().deserialize(metadata_str)
+                metadata_obj2 = Metadata.from_bytes(f.read())
 
             # Assert that both methods instantiate the right inner class for
             # each metadata type and ...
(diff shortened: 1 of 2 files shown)


avelichka (commented):
LGTM :)

MVrachev (approved):
LGTM!

tuf/api/metadata.py
```diff
@@ -128,20 +128,40 @@ def from_file(
...
+    @staticmethod
+    def from_string(
+        data: str,
+        deserializer: Optional[MetadataDeserializer] = None,
+    ) -> "Metadata":
+        """Loads TUF metadata from string.
+
+        Arguments:
+            data: metadata content as string.
+            deserializer: Optional; A MetadataDeserializer instance that
```
joshuagl, line 146: is this notation standard? Would we be better off just relying on the typing annotations to indicate optional arguments?

joshuagl (approved):
LGTM. Minor question inline.

tuf/api/metadata.py
jku: I saw that in a google style guide example -- could leave that out as well, it does duplicate the annotation...

jku:
Let's not merge just yet, I'm still going to doublecheck whether str is the correct type -- the Metadata API seems to use bytes in some places

jku:
I had another look and decided `bytes` makes more sense than `str`:
* a hypothetical serializer could take a binary format
* Existing code already uses bytes (see api/serialization/)

@joshuagl want to have another look? (mostly this was a literal replace of string with bytes)

tests/test_api.py
```diff
@@ -119,15 +118,17 @@ def test_generic_read(self):
             self.assertDictEqual(
                     metadata_obj.to_dict(), metadata_obj2.to_dict())
 
-
         # Assert that it chokes correctly on an unknown metadata type
         bad_metadata_path = 'bad-metadata.json'
         bad_metadata = {'signed': {'_type': 'bad-metadata'}}
+        bad_string = json.dumps(bad_metadata).encode('utf-8')
```
joshuagl, line 124: Is indentation off here, or is it just the GitHub app?

tuf/api/metadata.py
joshuagl: Let's leave it for now. Could you create an issue to ensure we review docstrings and make sure they are consistent before we cut a release.

joshuagl (approved):

joshuagl:
Looks good to me, thanks for the str->bytes change.

jku:
_GitHub Actions has encountered an internal error when running your job._ Seems to be https://www.githubstatus.com/incidents/cj7gzzj30411
Re-running...

Issue #1306, opened 2021-03-12
Implement verification by a threshold of keys 


sechkova:
**Description of issue or feature request**:
A verification of signed metadata by a threshold of keys is needed both for client and repository. 

`Metadata` already has vanilla one-signature verify and could potentially be the right location for this implementation. See https://github.com/theupdateframework/tuf/pull/1060#issuecomment-660056376 for thoughts about that distinction.

joshuagl:
When computing whether the threshold of signatures has been met, we need to be sure that only a single valid signature is counted for each keyid in the signatures list.

This prevents a malicious key owner from signing metadata multiple times to meet the threshold.

For more details see: https://github.com/theupdateframework/tuf/security/advisories/GHSA-pwqf-9h7j-7mv8 and https://github.com/theupdateframework/tuf/commit/83ac7be525b733f79a7e9bc573ec580ec835f179

mnm678:
Good catch @joshuagl. Depending on how keyids are computed or validated, we may also need to ensure that each actual key is only counted once (and not, for example use both the sha256 and sha512 keyids for the same key). There's some discussion about this in [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md)

joshuagl:
Indeed, I considered multiple keyids for the same key but decided not to mention it as the current specification requires keyids be the sha256 hash of the key's canonical form. I agree it would be worth trying to implement threshold computation in  a forward-thinking way if it doesn't complicate things too much, that probably means ensuring that signatures are unique, keyids are unique, and that each keyid maps to a unique key?

FYI I filed a PR against the spec to try and describe what's required of implementations here https://github.com/theupdateframework/specification/pull/154

jku:
So the problem space is this:
* Root contains keys and thresholds for top level metadata
* delegating (targets) roles contain keys and thresholds for their delegated roles
* The "metadata controller" (client code or some repository tool) wants to verify all metadata using the correct keys and thresholds from the delegating role in a way that makes it clear what is happening
* "key uniqueness" wrt threshold needs to be upheld

Some options I see are (examples for a hypothetical client that wants to verify a timestamp file it has just downloaded):
* Root and Targets provide a function that verifies their delegates (note that this one turns the verifier/verifiee around -- root verifies timestamp instead of timestamp verifying itself with data from root):
    ```
    root.signed.verify_delegate_with_threshold("timestamp", unverified_metadata)
    ```
[...]

PR #1355, opened 2021-04-20
experimental client: Add MetadataBundle

jku:
MetadataBundle keeps track of current valid set of metadata for the
client, and handles almost every step of the "Detailed client workflow"
in the TUF specification (the remaining steps are I/O related).

It also
verifies any new metadata (local or downloaded from remote repository).

**EDIT** this bit is no longer true in the newest version: _The bundle takes care of persisting valid metadata on disk, loading
local metadata from disk and deleting invalid local metadata._

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -450,21 +450,23 @@ class TestMetadata(unittest.TestCase):
             with self.assertRaises(ValueError):
                 DelegatedRole.from_dict(role.copy())
 
-            # Test creating DelegatedRole only with "path_hash_prefixes"
+            # Test creating DelegatedRole only with "path_hash_prefixes" (an empty one)
             del role["paths"]
-            DelegatedRole.from_dict(role.copy())
-            role["paths"] = "foo"
+            role["path_hash_prefixes"] = []
+            role_obj = DelegatedRole.from_dict(role.copy())
+            self.assertEqual(role_obj.to_dict(), role)
 
-            # Test creating DelegatedRole only with "paths"
+            # Test creating DelegatedRole only with "paths" (now an empty one)
             del role["path_hash_prefixes"]
-            DelegatedRole.from_dict(role.copy())
-            role["path_hash_prefixes"] = "foo"
+            role["paths"] = []
+            role_obj = DelegatedRole.from_dict(role.copy())
+            self.assertEqual(role_obj.to_dict(), role)
 
             # Test creating DelegatedRole without "paths" and
             # "path_hash_prefixes" set
             del role["paths"]
-            del role["path_hash_prefixes"]
-            DelegatedRole.from_dict(role)
+            role_obj = DelegatedRole.from_dict(role.copy())
+            self.assertEqual(role_obj.to_dict(), role)
 
 
     def test_delegation_class(self):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -496,6 +496,21 @@ class TestMetadata(unittest.TestCase):
         delegations = Delegations.from_dict(copy.deepcopy(delegations_dict))
         self.assertEqual(delegations_dict, delegations.to_dict())
 
+        # empty keys and roles
+        delegations_dict = {"keys":{}, "roles":[]}
+        delegations = Delegations.from_dict(delegations_dict.copy())
+        self.assertEqual(delegations_dict, delegations.to_dict())
+
+        # Test some basic missing or broken input
+        invalid_delegations_dicts = [
+            {},
+            {"keys":None, "roles":None},
+            {"keys":{"foo":0}, "roles":[]},
+            {"keys":{}, "roles":["foo"]},
+        ]
+        for d in invalid_delegations_dicts:
+            with self.assertRaises((KeyError, AttributeError)):
+                Delegations.from_dict(d)
 
     def test_metadata_targets(self):
         targets_path = os.path.join(
--- a/tuf/client_rework/metadata_bundle.py
+++ b/tuf/client_rework/metadata_bundle.py
@@ -1,0 +1,349 @@
+# Copyright the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""TUF client bundle-of-metadata
+
+MetadataBundle keeps track of current valid set of metadata for the client,
+and handles almost every step of the "Detailed client workflow" in the TUF
+specification (the remaining steps are download related). The bundle takes
+care of persisting valid metadata on disk, loading local metadata from disk
+and deleting invalid local metadata.
+
+New metadata (downloaded from a remote repository) can be loaded using
+'update_metadata()'. The type of accepted metadata depends on bundle state
+(states are "root"/"timestamp"/"snapshot"/"targets"/). Bundle states advances
+to next state on every successful metadata update, except for "root" where state
+only advances when 'root_update_finished()' is called. Exceptions will be thrown
+if metadata fails to load in any way.
+
+Example (with hypothetical download function):
+
+>>> # Load local root
+>>> bundle = MetadataBundle("path/to/metadata")
+>>>
+>>> # state: "root", load more root versions from remote
+>>> with download("root", bundle.root.signed.version + 1) as f:
+>>>     bundle.load_metadata(f.read())
+>>> with download("root", bundle.root.signed.version + 1) as f:
+>>>     bundle.load_metadata(f.read())
+>>>
+>>> # Finally, no more root from remote
+>>> bundle.root_update_finished()
+>>>
+>>> # state: "timestamp", load timestamp
+>>> with download("timestamp") as f:
+>>>     bundle.load_metadata(f.read())
+>>>
+>>> # state: "snapshot", load snapshot (consistent snapshot not shown)
+>>> with download("snapshot") as f:
+>>>     bundle.load_metadata(f.read())
+>>>
+>>> # state: "targets", load targets
+>>> version = bundle.snapshot.signed.meta["targets.json"]["version"]
+>>> with download("snapshot", version + 1) as f:
+>>>     bundle.load_metadata(f.read())
+>>>
+>>> # Top level metadata is now fully loaded and verified
+
+
+TODO:
+ * Delegated targets not implement yet
+ * exceptions are all over the place and not thought out at all
+ * a bit of repetition
+ * No tests!
+ * Naming maybe not final?
+ * some metadata interactions might work better in Metadata itself
+ * Progress through Specification update process should be documented
+   (not sure yet how)
+"""
+
+from collections import abc
+from datetime import datetime
+import logging
+import os
+from typing import Dict
+
+from securesystemslib import keys as sslib_hash
+from securesystemslib import keys as sslib_keys
+
+from tuf import exceptions
+from tuf.api.metadata import Metadata
+
+logger = logging.getLogger(__name__)
+
+# This is a placeholder until ...
+# TODO issue 1306: implement this in Metadata API
+def verify_with_threshold(root: Metadata, role: str, unverified: Metadata):
+    unique_keys = set()
+    for keyid in root.signed.roles[role]["keyids"]:
+        key_metadata = root.signed.keys[keyid]
+        key, _ = sslib_keys.format_metadata_to_key(key_metadata)
+
+        try:
+            if unverified.verify(key):
+                unique_keys.add(key["keyval"]["public"])
+        except:
+            pass
+
+    return len(unique_keys) >= root.signed.roles[role]["threshold"]
+
+
+# TODO issue 1336: implement in metadata api
+from tuf.api.serialization.json import JSONDeserializer
+
+
+def from_string(data: str) -> Metadata:
+    return JSONDeserializer().deserialize(data)
+
+
+class MetadataBundle(abc.Mapping):
+    def __init__(self, path: str):
+        """Initialize by loading existing metadata from disk
+
+        This includes root, timestamp, snapshot and _top-level_ targets .
+        """
+        self._path = path
+        self._bundle = {}  # type: Dict[str: Metadata]
+        self._state = "root"
+        self.reference_time = None
+
+        if not os.path.exists(path):
+            # TODO try to create dir instead?
+            raise exceptions.RepositoryError("Repository does not exist")
+
+        # Load and validate the local root metadata
+        # Valid root metadata is required (but invalid files are not removed)
+        try:
+            with open(os.path.join(self._path, "root.json"), "rb") as f:
+                self._load_intermediate_root(f.read())
+            logger.debug("Loaded local root.json")
+        except:
+            raise exceptions.RepositoryError("Failed to load local root metadata")
+
+    def update_metadata(self, metadata_str: str):
+        logger.debug("Updating %s", self._state)
+        if self._state == "root":
+            self._load_intermediate_root(metadata_str)
+            self.root.to_file(os.path.join(self._path, "root.json"))
+        elif self._state == "timestamp":
+            self._load_timestamp(metadata_str)
+            self.timestamp.to_file(os.path.join(self._path, "timestamp.json"))
+            self._state = "snapshot"
+        elif self._state == "snapshot":
+            self._load_snapshot(metadata_str)
+            self.snapshot.to_file(os.path.join(self._path, "snapshot.json"))
+            self._state = "targets"
+        elif self._state == "targets":
+            self._load_targets(metadata_str)
+            self.targets.to_file(os.path.join(self._path, "targets.json"))
+            self._state = ""
+        else:
+            raise NotImplementedError
+
+    def root_update_finished(self):
+        if self._state != "root":
+            # bundle does not support this order of ops
+            raise exceptions.RepositoryError
+
+        self._make_root_permanent(self)
+        self._state = "timestamp"
+
+    # Implement Mapping
+    def __getitem__(self, key: str):
+        return self._bundle[key]
+
+    def __len__(self):
+        return len(self._bundle)
+
+    def __iter__(self):
+        return iter(self._bundle)
+
+    # Helper properties for top level metadata
+    @property
+    def root(self):
+        return self._bundle.get("root")
+
+    @property
+    def timestamp(self):
+        return self._bundle.get("timestamp")
+
+    @property
+    def snapshot(self):
+        return self._bundle.get("snapshot")
+
+    @property
+    def targets(self):
+        return self._bundle.get("targets")
+
+    def _load_intermediate_root(self, data: str):
+        """Verify the new root using current root (if any) and use it as current root
+
+        Raises if root fails verification
+        """
+        new_root = from_string(data)
+        if new_root.signed._type != "root":
+            raise exceptions.RepositoryError
+
+        if self.root is not None:
+            if not verify_with_threshold(self.root, "root", new_root):
+                raise exceptions.UnsignedMetadataError(
+                    "New root is not signed by root", new_root.signed
+                )
+
+            if new_root.signed.version != self.root.signed.version + 1:
+                # TODO not a "Replayed Metadata attack": the version is just not what we expected
+                raise exceptions.ReplayedMetadataError(
+                    "root", new_root.signed.version, self.root.signed.version
+                )
+
+        if not verify_with_threshold(new_root, "root", new_root):
+            raise exceptions.UnsignedMetadataError(
+                "New root is not signed by itself", new_root.signed
+            )
+
+        self._bundle["root"] = new_root
+
+    def _make_root_permanent(self):
+        # Store our reference "now", verify root expiry
+        self.reference_time = datetime.utcnow()
+        if self.root.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError
+
+        logger.debug("Verified final root.json")
+
+        # Load remaning local metadata: this ensures invalid
+        # metadata gets wiped from disk
+        try:
+            with open(os.path.join(self._path, "timestamp.json"), "rb") as f:
+                self._load_timestamp(f.read())
+            logger.debug("Loaded local timestamp.json")
+        except Exception as e:
+            # TODO only handle specific errors
+            logger.debug("Failed to load local timestamp.json")
+            # TODO delete local file
+
+        try:
+            with open(os.path.join(self._path, "snapshot.json"), "rb") as f:
+                self._load_snapshot(f.read())
+            logger.debug("Loaded local snapshot.json")
+        except Exception as e:
+            # TODO only handle specific errors
+            logger.debug("Failed to load local snapshot.json")
+            # TODO delete local file
+
+        try:
+            with open(os.path.join(self._path, "targets.json"), "rb") as f:
+                self._load_targets(f.read())
+            logger.debug("Loaded local targets.json")
+        except Exception as e:
+            # TODO only handle specific errors
+            logger.debug("Failed to load local targets.json")
+            # TODO delete local file
+
+    def _load_timestamp(self, data: str):
+        """Verifies the new timestamp and uses it as current timestamp
+
+        Raises if verification fails
+        """
+        new_timestamp = from_string(data)
+        if new_timestamp.signed._type != "timestamp":
+            raise exceptions.RepositoryError
+
+        if not verify_with_threshold(self.root, "timestamp", new_timestamp):
+            raise exceptions.UnsignedMetadataError(
+                "New timestamp is not signed by root", new_timestamp.signed
+            )
+
+        if self.timestamp is not None:
+            # Prevent rolling back timestamp version
+            if new_timestamp.signed.version < self.timestamp.signed.version:
+                raise exceptions.ReplayedMetadataError(
+                    "timestamp",
+                    new_timestamp.signed.version,
+                    self.timestamp.signed.version,
+                )
+            # Prevent rolling back snapshot version
+            if (
+                new_timestamp.signed.meta["snapshot.json"]["version"]
+                < self.timestamp.signed.meta["snapshot.json"]["version"]
+            ):
+                # TODO not sure about the
+                raise exceptions.ReplayedMetadataError(
+                    "snapshot",
+                    new_timestamp.signed.meta["snapshot.json"]["version"],
+                    self.timestamp.signed.meta["snapshot.json"]["version"],
+                )
+
+        if new_timestamp.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError
+
+        self._bundle["timestamp"] = new_timestamp
+
+    def _load_snapshot(self, data: str):
+        # Verify against the hashes in timestamp, if any
+        meta = self.timestamp.signed.meta["snapshot.json"]
+        hashes = meta.get("hashes") or {}
+        for algo, _hash in meta["hashes"].items():
+            digest_object = sslib_hash.digest(algo)
+            digest_object.update(data)
+            if digest_object.hexdigest() != _hash:
+                raise exceptions.BadHashError()
+        new_snapshot = from_string(data)
+        if new_snapshot.signed._type != "snapshot":
+            raise exceptions.RepositoryError
+
+        if not verify_with_threshold(self.root, "snapshot", new_snapshot):
+            raise exceptions.UnsignedMetadataError(
+                "New snapshot is not signed by root", new_snapshot.signed
+            )
+
+        if (
+            new_snapshot.signed.version
+            != self.timestamp.signed.meta["snapshot.json"]["version"]
+        ):
+            raise exceptions.BadVersionNumberError
+
+        if self.snapshot:
+            for filename, fileinfo in self.snapshot.signed.meta.items():
+                new_fileinfo = new_snapshot.signed.meta.get(filename)
+
+                # Prevent removal of any metadata in meta
+                if new_fileinfo is None:
+                    raise exceptions.ReplayedMetadataError
+
+                # Prevent rollback of any metadata versions
+                if new_fileinfo["version"] < fileinfo["version"]:
+                    raise exceptions.ReplayedMetadataError
+
+        if new_snapshot.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError
+
+        self._bundle["snapshot"] = new_snapshot
+
+    def _load_targets(self, data: str):
+        # Verify against the hashes in snapshot, if any
+        meta = self.snapshot.signed.meta["targets.json"]
+
+        hashes = meta.get("hashes") or {}
+        for algo, _hash in hashes.items():
+            digest_object = sslib_hash.digest(algo)
+            digest_object.update(data)
+            if digest_object.hexdigest() != _hash:
+                raise exceptions.BadHashError()
+
+        new_targets = from_string(data)
+        if new_targets.signed._type != "targets":
+            raise exceptions.RepositoryError
+
+        if not verify_with_threshold(self.root, "targets", new_targets):
+            raise exceptions.UnsignedMetadataError(
+                "New targets is not signed by root", new_targets.signed
+            )
+
+        if new_targets.signed.version != meta["version"]:
+            raise exceptions.BadVersionNumberError
+
+        if new_targets.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError
+
+        self._bundle["targets"] = new_targets
--- a/tuf/client_rework/metadata_bundle.py
+++ b/tuf/client_rework/metadata_bundle.py
@@ -10,38 +10,43 @@ care of persisting valid metadata on disk, loading local metadata from disk
 and deleting invalid local metadata.
 
 New metadata (downloaded from a remote repository) can be loaded using
-'update_metadata()'. The type of accepted metadata depends on bundle state
-(states are "root"/"timestamp"/"snapshot"/"targets"/). Bundle states advances
-to next state on every successful metadata update, except for "root" where state
-only advances when 'root_update_finished()' is called. Exceptions will be thrown
-if metadata fails to load in any way.
+'update_metadata()'. The type of accepted metadata depends on bundle state:
+ * Metadata is loadable only if all metadata it depends on is loaded
+ * Metadata is immutable if any metadata depending on it has been loaded 
+
+Exceptions are raised if metadata fails to load in any way (except in the
+case of local loads -- see locad_local_metadata()).
 
 Example (with hypothetical download function):
 
 >>> # Load local root
 >>> bundle = MetadataBundle("path/to/metadata")
 >>>
->>> # state: "root", load more root versions from remote
+>>> # load more root versions from remote
 >>> with download("root", bundle.root.signed.version + 1) as f:
->>>     bundle.load_metadata(f.read())
+>>>     bundle.update_metadata(f.read())
 >>> with download("root", bundle.root.signed.version + 1) as f:
->>>     bundle.load_metadata(f.read())
+>>>     bundle.update_metadata(f.read())
 >>>
->>> # Finally, no more root from remote
+>>> # Finally, no more roots from remote
 >>> bundle.root_update_finished()
 >>>
->>> # state: "timestamp", load timestamp
+>>> # load local timestamp, then update it
+>>> bundle.load_local_metadata("timestamp")
 >>> with download("timestamp") as f:
->>>     bundle.load_metadata(f.read())
+>>>     bundle.update_metadata(f.read())
 >>>
->>> # state: "snapshot", load snapshot (consistent snapshot not shown)
->>> with download("snapshot") as f:
->>>     bundle.load_metadata(f.read())
+>>> # load local snapshot, then update it if needed
+>>> if not bundle.load_local_metadata("snapshot"):
+>>>     # load snapshot (consistent snapshot not shown)
+>>>     with download("snapshot") as f:
+>>>         bundle.update_metadata(f.read())
 >>>
->>> # state: "targets", load targets
->>> version = bundle.snapshot.signed.meta["targets.json"]["version"]
->>> with download("snapshot", version + 1) as f:
->>>     bundle.load_metadata(f.read())
+>>> # load local targets, then update it if needed
+>>> if not bundle.load_local_metadata("targets"):
+>>>     version = bundle.snapshot.signed.meta["targets.json"]["version"]
+>>>     with download("snapshot", version + 1) as f:
+>>>         bundle.update_metadata(f.read())
 >>>
 >>> # Top level metadata is now fully loaded and verified
 
@@ -102,7 +107,6 @@ class MetadataBundle(abc.Mapping):
         """
         self._path = path
         self._bundle = {}  # type: Dict[str: Metadata]
-        self._state = "root"
         self.reference_time = None
 
         if not os.path.exists(path):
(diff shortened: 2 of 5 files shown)


jku:
From discussion with @sechkova: We do need to expose local metadata loading (instead of doing it automatically) for two reasons:
* delegate handling is going to need it
* spec "requires" downloading top-level metadata but in fact we can avoid it if we know the current local metadata is valid (valid for snapshot and targets I suppose)

For delegate handle we also need to have an optional delegator role.

So instead of just `update_metadata(metadata: str)` we'll likely have something like
```
def load_local_metadata(role: str, delegator_role: Optional[str]) -> bool
    """Tries to load 'role' metadata from disk
    Returns True if valid, up-to-date metadata was found and loaded
    """

def update_metadata(metadata: str, delegator_role: Optional[str])
    """Loads and persists given metadata on disk
    Raises if metadata is not valid or could not be verified
    """
```

jku:
Initial draft was very neat and clean but after some impacts with reality (thanks Teodora for early review), it currently looks like this:
 * ~50 lines of code that should in the end be in metadata
 * 160 lines of code for internal state tracking (so e.g. root can't be updated when targets is already loaded), and quality of life improvements like properties
 * ~160 lines of code for the actual spec checks (these _should_ be complete now)
 
I still think this is a good direction. The code is not ready by any means but I'm making this pr now so that Teodora can start using it early if we decide this is a path experimental-client should take. I suggest reviewing the final form and not individual commits -- there were false starts in there.

The second bullet point in the list above is a potential issue: is the internal state tracking understandable and clear (or can it be improved to be that).

The other potential issue is delegated metadata. Delegates can be added and they are validated on insert but they are not currently under the same guarantees that top-level metadata is: if there's a top-level "targets" in the bundle it's guaranteed to be valid, but nothing currently stops caller from updating  "role1" so that it no longer delegates to "role2" that is also in the bundle... This can definitely be fixed (I think just reloading all metadata delegated by the just updated metadata should do it and might even simplify the internal state tracking : I just haven't had time to try yet).

Third things to mention might be exceptions: I believe one is thrown where needed now... but the actual exceptions are not chosen at all: My plan is that all issues that arise from metadata content will be under one base Exception. Client will likely just handle the base exception. Other exceptions will be internal errors and user errors.

jku:
On the API and naming:

```
MetadataBundle
  # Implements Mapping so individual metadata can be access with []-indexing

  # methods
  def load_local_metadata(...)
  def update_metadata(...)
  def root_update_finished(...)

  # helper properties
  root
  snapshot
  timestamp
  targets
```
MetadataBundle itself is not a great name but works -- I know Teodora is fond of "trusted" or some variation of it in the name to make that part clear.

the loader function names could be `load_local()`, `update_from_remote()` to make the distinction clear but names shorter.
 `root_update_finished(...)` is not actually strictly speaking required -- the root verification could safely be made on first timestamp load -- but I've kept it there since it matches the specification better (it's the marker for root update loop having finished).

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,349 @@
...
+
+    return len(unique_keys) >= root.signed.roles[role]["threshold"]
+
+
+# TODO issue 1336: implement in metadata api
+from tuf.api.serialization.json import JSONDeserializer
+
+
+def from_string(data: str) -> Metadata:
+    return JSONDeserializer().deserialize(data)
```
MVrachev: You can start using `Metadata.from_bytes()` after it was added by you, right?

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+class MetadataBundle(abc.Mapping):
+    def __init__(self, path: str):
```
MVrachev: Nitpick: Maybe it's a good idea to use a little more descriptive name here.

It's a little unclear: is this path to specific metadata and all others are in the same folder or it's a path to the parent folder of all expected metadata files?
Probably something like `repository_path` or if you want to keep it short: `repo_path`.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,349 @@
...
+        """Initialize by loading existing metadata from disk
+
+        This includes root, timestamp, snapshot and _top-level_ targets .
+        """
+        self._path = path
+        self._bundle = {}  # type: Dict[str: Metadata]
+        self._state = "root"
+        self.reference_time = None
+
+        if not os.path.exists(path):
```
MVrachev: Nitpick: I would expect function argument checks to be before attribute assignment.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,349 @@
...
+            # TODO try to create dir instead?
```
MVrachev: Yes, I will expect that to happen if there is no repository in the path I have passed.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,450 @@
...
+# TODO issue 1306: implement this in Metadata API
+def verify_with_threshold(delegator: Metadata, role_name: str, unverified: Metadata):
+    if delegator.signed._type == "root":
+        keys = delegator.signed.keys
+        role = delegator.signed.roles.get(role_name)
+    elif delegator.signed._type == "targets":
+        keys = delegator.signed.delegations["keys"]
+        # role names are unique: first match is enough
+        roles = delegator.signed.delegations["roles"]
+        role = next((role for role in roles if role["name"] == role_name), None)
```
MVrachev: Not sure why do you take the `next` one?
Don't you want exactly the role which equals the `role_name`?

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,450 @@
...
+>>>
+>>> # Top level metadata is now fully loaded and verified
+
+
+TODO:
+ * Delegated targets are implemented but they are not covered
+   by same immutability guarantees: the top-level metadata is handled
+   by hard-coded rules (can't update root if snapshot is loaded)
+   but delegations would require storing the delegation tree ...
+ * exceptions are all over the place and not thought out at all
```
MVrachev: I would add  `Update docstrings with what exceptions are raised`, so that you won't forget.
Right now, you are documenting `Raises if X` in the functions docstrings.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,450 @@
...
+                    "timestamp",
+                    new_timestamp.signed.version,
+                    self.timestamp.signed.version,
+                )
+            # Prevent rolling back snapshot version
+            if (
+                new_timestamp.signed.meta["snapshot.json"]["version"]
+                < self.timestamp.signed.meta["snapshot.json"]["version"]
+            ):
+                # TODO not sure about the
```
MVrachev: Do you remember what you meant here? 
Probably not sure about the exception?

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,450 @@
...
+>>> # Top level metadata is now fully loaded and verified
+
+
+TODO:
+ * Delegated targets are implemented but they are not covered
+   by same immutability guarantees: the top-level metadata is handled
+   by hard-coded rules (can't update root if snapshot is loaded)
+   but delegations would require storing the delegation tree ...
+ * exceptions are all over the place and not thought out at all
+ * usefulness of root_update_finished() can be debated: it could be done
```
MVrachev, line 62: Yea, I agree and I don't see where are you using it now?

MVrachev (changes requested):
Overall, the code is easy to read and follow. The few complicated places (like verifying hashes) have issues that those things can be added in the Metadata API.
Also, I agree that helper functions could be useful like `get_snap_version()` in timestamp.

One additional suggestion.
Right now, you are calling `self._raise_on_unsupported_state(role_name)` when loading and updating, and you do checks at the beginning of each of `_load_*` functions.
That way it's easy to forget or duplicate checks.

I would suggest using `self._raise_on_unsupported_state(role_name)` to handle all state checks
and call it at the begging of each of the `_load_*` functions.

tuf/client_rework/metadata_bundle.py
sechkova: Updater calls it, it is a public method.
 If we want to be strict with the spec and do what it calls "protection from fast forward attacks" , I think this is the place. Although I admit, you already got me here because I forgot to call `root_update_finished()` and got an error :)

tuf/client_rework/metadata_bundle.py
jku: I'm looking for the role in roles that has the name `role_name`.

the list comprehension works as a filter that is also an iterator, and next() gives me the first value from that iterator (or None if no match).

API could maybe offer `Targets.get_delegation(role_name)` -- but let's not move too hastily: this calling function should IMO live in the API itself (see #1306).

tuf/client_rework/metadata_bundle.py
jku: yes almost certainly :)

tuf/client_rework/metadata_bundle.py
jku: it's a public method that client should currently call after final root has been loaded -- according to spec we verify root expiry at this point... Practically we could verify it when the next metadata (timestamp) is about to be loaded but it looks a bit goofy: caller is trying to load timestamp but gets an error saying root is suddenly not valid.

jku:
> Right now, you are calling `self._raise_on_unsupported_state(role_name)` when loading and updating, and you do checks at the beginning of each of `_load_*` functions.
> That way it's easy to forget or duplicate checks.

Yeah this is true

> I would suggest using `self._raise_on_unsupported_state(role_name)` to handle all state checks
> and call it at the begging of each of the `_load_*` functions.

I think I want to only do it early in the public functions and am 99% sure I can just remove the checks in the actual `_load_*()` methods... just haven't gone through the trouble of verifying that yet.

tuf/client_rework/metadata_bundle.py
jku: On second thought it's not actually _argument_ validation: we're checking whether the path exists on filesystem and might even create it in this code path... so I think the order is fine

trishankatdatadog:
There's a lot of if-else inside generic-sounding functions. Is that a good idea? Should we perhaps use specific functions for specific purposes? Not only improves readability, but also detecting errors.

jku:
> There's a lot of if-else inside generic-sounding functions. Is that a good idea? Should we perhaps use specific functions for specific purposes? 

You're totally correct (I debated myself about writing about this but ... I wrote so many comments already :) ).

Let me drill down into this a bit. There are two classes of if-elses:
* state checks in _raise_on_unsupported_state(): this does not so much make program flow decisions but errors out if a logical error (a programming error) is about to happen. This code can possibly be simplified but I argue it has its place... It could also be removed completely, if we just assume caller never makes mistakes :)
* The other case is the public functions `load_local_metadata()` and `update_metadata()` that operate as dispatchers to the specific internal functions -- here I completely agree with you that specific functions would be better (instead of a dispatcher we would have more public methods like `load_local_root`, `load_local_timestamp` etc that would be self explanatory and harder to make mistakes with). However this ties to the mirrors discussion (more below)

This relates to mirrors in the following way:
* we recognised the download/mirrors code is what makes this tricky -- this is what Teodora tried to solve in some earlier PRs but ultimately failed to create something that improves on the current code _and still supports mirrors_...
* as long as the mirror downloading works as it does, it means there is going to be a "generic download loop" that is going to download files, parse the content, verify the metatadata and then decide if it needs to download another version of the file. Anything else is going to lead to a lot of repetition in the updater code.
* This means that if bundle does not offer generic dispatcher functions, then the calling updater code is going to have to do the exact same thing instead

So I believe the dispatching code (the if-clauses that look at role name to decide what to call) is going to exist _somewhere_ anyway (again, assuming current mirroring support)


So I guess this brings us to: **Re-designing mirror support (essentially limiting it drastically) would make this, and the updater download code, much cleaner**. We've been really trying to not solve too many issues at once but maybe we do need to look at that as well...


(marking draft until the mirrors discussion happens)

tuf/client_rework/metadata_bundle.py
jku: Since bundle does not support any "root bootstrapping" at this point creating directories would be pointless so I removed this whole check

jku:
Updating this based on the mirrors/downloads changes:
* functions modifying top-level metadata are now named so `update_root(data)` instead of `update_metadata(data, "root")`
* all of the type-dispatching was removed as result
* minimal internal state checking is now done in `_load_*()` (the functions that actually do the spec checks)

I'm quite happy with this: it's not any shorter than the previous version but it's simple and straight forward. I've tested using the bundle from the client and mostly it feels like a clear improvement: the client code can concentrate on deciding what should happen (_"Now load snapshot from local storage: if that one's not valid anymore, then load snapshot from remote"_) while the bundle verifies that metadata state is valid according to spec at every step.

It's definitely not complete (most importantly missing tests and exceptions are still WIP, and the spec validation needs to be reviewed carefully by someone who's not me) but I think it might be good enough for the experimental-client? Opinions as well as naming suggestions are welcome. On naming: is MetadataBundle ok? load_local_*()? update_*()?

TLDR documentation:
> MetadataBundle is an internal class for the Updater. It's purpose is to keep track of valid metadata, ensure that this metadata is always valid, and to provide easy ways to update the metadata with the caller making decisions on what is updated.
>
> ## Modifying metadata
> `load_local_*()`: Load metadata from local cache
> `update_*()`: Update and persist metadata using data from remote repository
> `root_update_finished()`: Mark root metadata as final (to verify it's not expired)
> 
> ## Accessing loaded metadata
> Use the Mapping index access: `bundle['root']`, `bundle.get('root')`
> For top-level metadata use the helpers: `bundle.root`
> 
> ## Internal
> `_load_*()`: Actual spec checks, these are called by both `load_local_*()` and `update_*()`

(EDIT: rebasing this to top of the branch brings out more linting issues: I'll still be fixing those)

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,482 @@
...
+   generic RepositoryError that covers every issue that server provided metadata
+   could inflict (other errors would be user errors), but this is not yet the case
+ * usefulness of root_update_finished() can be debated: it could be done
+   in the beginning of load_timestamp()...
+ * there are some divergences from spec:
+   * 5.3.11: timestamp and snapshot are not deleted right away (only on next load):
+     the load functions will refuse to load the files when they are not signed by
+     current root keys. Deleting at the specified point is possible but means additional
+     code with some quirks..
+   * in general local metadata files are not deleted (they just won't succesfully load)
```
jku: This might be worth a discussion. I don't like deleting local metadata files when they're invalid for two reasons:
* it doesn't provide security (if the metadata is invalid it cannot be loaded -- if it could then our program has a bug that should be fixed: deleting the metadata won't really help)
* it prevents all debugging: "oh the app says update failed because of invalid metadata... well, that file doesn't exist anymore :shrug:"

jku:
Had to rebase on current experimental-client to get the linting up-to-date: of course the newer linting revealed lots of new issues that I've fixed or documented :)

Even more than before I suggest not reading individual commits: there's a lot of less-than-useful history there.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+from tuf import exceptions
+from tuf.api.metadata import Metadata
+from tuf.api.serialization import SerializationError
+
+# TODO: Either enaable old-style logging in pylintc (issue #1334)
+# or change this file to use f-strings for logging
+# pylint: disable=logging-too-many-args
+
+# TODO: signed._type really does not work issue #1375:
+# pylint: disable=protected-access
```
joshuagl: For the whole module? 😱 Let's rebase after the merge of #1376

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+
+logger = logging.getLogger(__name__)
+
+# This is a placeholder until ...
+# TODO issue 1306: implement this in Metadata API
+def verify_with_threshold(
+    delegator: Metadata, role_name: str, unverified: Metadata
+) -> bool:
+    """Verify 'unverified' with keys and treshold defined in delegator"""
```
joshuagl:
```suggestion
    """Verify 'unverified' with keys and threshold defined in delegator"""
```

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+            new_root = Metadata.from_bytes(data)
+        except SerializationError as e:
+            raise exceptions.RepositoryError("Failed to load root") from e
+
+        if new_root.signed._type != "root":
+            raise exceptions.RepositoryError(
+                f"Expected 'root', got '{new_root.signed._type}'"
+            )
+
+        if self.root is not None:
```
sechkova, line 214: A comment will be useful to show that `if root is None` means that we are loading the local trusted root, otherwise we are updating it.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+        """Mark root metadata as final."""
+        if self.reference_time is not None:
+            raise ValueError("Root update is already finished")
+
+        # Store our reference "now", verify root expiry
+        self.reference_time = datetime.utcnow()
+        if self.root.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("New root.json is expired")
+
+        logger.debug("Verified final root.json")
```
sechkova: Not sure what would be the right place to put this but I think we need a text somewhere saying that the by design the bundle avoids fast-forward attacks (it does not load previous snapshot and timestamp on init) and does not need to perform explicitly step 5.3.11 from the spec.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
+# Copyright the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""TUF client bundle-of-metadata
+
+MetadataBundle keeps track of current valid set of metadata for the client,
+and handles almost every step of the "Detailed client workflow" in the TUF
```
joshuagl: Worth adding a link? https://theupdateframework.github.io/specification/latest/#detailed-client-workflow

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+Loaded metadata can be accessed via the index access with rolename as key
+or, in the case of top-level metadata using the helper properties like
+'MetadataBundle.root'
+
+The rules for top-level metadata are
+ * Metadata is loadable only if metadata it depends on is loaded
+ * Metadata is immutable if any metadata depending on it has been loaded
+ * Caller must load/update these in order:
+   root -> timestamp -> snapshot -> targets -> (other delegated targets)
+ * Caller should try loading local file before updating metadata from remote
```
joshuagl: Why? What happens if they don't? We should somehow distinguish, at least in our documentation, between initial/trusted root metadata – which [must exist](https://theupdateframework.github.io/specification/latest/#load-trusted-root) – and other local metadata.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+>>> bundle.root_update_finished()
+>>>
+>>> # load timestamp, then update from remote
+>>> bundle.load_local_timestamp()
+>>> with download("timestamp") as f:
+>>>     bundle.update_timestamp(f.read())
+>>>
+>>> # load snapshot, update from remote if needed
+>>> if not bundle.load_local_snapshot():
+>>>     # TODO get version from timestamp
```
joshuagl: For this and the other TODO entries in the example, are we waiting for the `MetadataInfo` class to merge #1329?

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+from securesystemslib import hash as sslib_hash
+from securesystemslib import keys as sslib_keys
+
+from tuf import exceptions
+from tuf.api.metadata import Metadata
+from tuf.api.serialization import SerializationError
+
+# TODO: Either enaable old-style logging in pylintc (issue #1334)
+# or change this file to use f-strings for logging
+# pylint: disable=logging-too-many-args
```
joshuagl: Per discussion in #1334, let's enable "old-style" logging.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+    def targets(self) -> Optional[Metadata]:
+        return self._bundle.get("targets")
+
+    # Public methods
+    def update_root(self, data: bytes):
+        """Update root metadata with data from remote repository."""
+        logger.debug("Updating root")
+
+        self._load_intermediate_root(data)
+        self.root.to_file(os.path.join(self._path, "root.json"))
```
joshuagl: Should we be persisting intermediate root metadata with its versioned prefix? Not necessarily from a spec compliance perspective, but more in-line with your argument about not deleting files as it aides debugging.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+
+    def root_update_finished(self):
+        """Mark root metadata as final."""
+        if self.reference_time is not None:
+            raise ValueError("Root update is already finished")
+
+        # Store our reference "now", verify root expiry
+        self.reference_time = datetime.utcnow()
+        if self.root.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("New root.json is expired")
```
joshuagl: Neat trick. The specification suggests the update start time before updating root. Arguably we should assign `self.reference_time` earlier in the `MetadataBundle` logic. What do you think?

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+
+        logger.debug("Verified final root.json")
+
+    def load_local_timestamp(self) -> bool:
+        """Load cached timestamp metadata from local storage.
+
+        Returns True if timestamp was succesfully loaded"""
```
joshuagl: This was a surprise. Especially as there were not any signs of checking return type of the `load_local_*` functions in the example snippet near the top of the file. Wouldn't the Pythonic way to do this be to throw an exception if the `load_local_*()` functions do not succeed? Or was this intentional, because only root metadata **must** exist locally?

joshuagl (commented):
This is very nice, easy to follow even when attention is depleted. Nice work @jku !

The class level docstring suggests that invalid local metadata is deleted. I think you made a conscious decision _not_ to do that? It looks like the current implementation only ever preserves the most recently verified metadata file for each role? If we did start storing intermediate files on disk, we would need to add a mechanism to prevent the local metadata directory getting too large. But I don't think that's necessary in the current state.

There's a lot of LBYL style code `load_local_*`, `is_expired()` and `if not verify_with_threshold()`, whereas I think we want to lean into the Pythonic EAFP. Can we reasonably have each of those, especially the latter two, raise Exceptions that would be reasonable to surface to the user? Or would we need to catch them and re-raise to provide meaningful errors?

It will be nice to see this without the repeated logic, such as the big block for checking hashes in meta.

Also, this appears to be internal API – did we figure the best way to indicate that? Or is that TBD?

tuf/client_rework/metadata_bundle.py
jku: Root is handled separately: local root load happens during instance initialization. I should make that clear in this text.

For all others I wanted to leave the decision to caller: it would be possible to keep track of local loads within the bundle and to do it automatically when needed, but it's more code and the caller is going to do it anyway (because in several cases caller wants to know if the local metadata is valid before it decides whether to even try downloading remote metadata)...

tuf/client_rework/metadata_bundle.py
jku: it's just that currently it ends up being about five lines of code more which makes the example a bit heavy... This is roughly what my test branch looks like:

```
        # load local snapshot, update from remote only if local is not valid
        if not self._bundle.load_local_snapshot():
            # Use length if available, use versioned name if consistent_snapshot
            filename = "snapshot.json"
            metainfo = self._bundle.timestamp.signed.meta[filename]
            length = metainfo.get("length") or DEFAULT_SNAPSHOT_MAX_LENGTH
            if self._bundle.root.signed.consistent_snapshot:
                filename = f"{metainfo['version']}.{filename}"

            data = self._download_metadata(filename, length)
            self._bundle.update_snapshot(data)
``` 

Maybe TODO is the wrong word: I should just mention that the example doesn't use consistent_snapshot

tuf/client_rework/metadata_bundle.py
jku: IMO spec compliance explicitly does not require this, btw: the part that talks about writing consistent snapshots seems to talk about the _remote_ repository to me, and the part that talks about client writes in general explicitly says files must be written as FILENAME.EXT.

Now, we will likely want to write versioned root files at some point anyway (to handle the bootstrap case) but I'm fine with adding that later on and keeping the code as simple as possible for now.

tuf/client_rework/metadata_bundle.py
jku: Maybe yes. The spec does say that...

This choice made sense to me as it is (very close to) the time that client last checks for root.json -- so theoretically  there's actual value in comparing it to e.g. server publish times if issues are debugged...

But this could be a situation where following the spec to letter is a better choice. That just means I need an extra variable for `root_is_final` :)

tuf/client_rework/metadata_bundle.py
jku: Examples do do things like `if not bundle.load_local_snapshot()`.

This was intentional design ... but I'm not sure it was correct. The logic is that failing to load non-root local metadata for any reason (it does not exist, it is expired, it is no longer signed by correct keys, ...) is completely normal for the caller: in all cases the caller can only assume things are working as expected and that remote metadata should now be downloaded.

Would it be more pythonic to raise? I guess so, it doesn't look horrible anyway (simplified example but still...):

```python
try:
    bundle.local_local_timestamp()
except: LocalLoadError as e:
    logger.info("Failed to load local timestamp: %s", e)

# Always load remote timestamp, even if local is valid
data = download_metadata(timestamp_filename)
bundle.update_timestamp(data)

try:
    bundle.local_local_snapshot()
except: LocalLoadError as e:
    logger.info("Failed to load local snapshot: %s", e)

    # Load remote snapshot only if local is not valid
    data = download_metadata(snapshot_filename)
    bundle.update_snapshot(data)
```

Maybe I need to try this

jku:
Thanks, good to get "fresh" eyes on this.

> The class level docstring suggests that invalid local metadata is deleted.

Oops.

> There's a lot of LBYL style code load_local_*, is_expired() and if not verify_with_threshold(), whereas I think we want to lean into the Pythonic EAFP. Can we reasonably have each of those, especially the latter two, raise Exceptions that would be reasonable to surface to the user? Or would we need to catch them and re-raise to provide meaningful errors?

I've gone into detail about load_local_* in a code comment (you may be right).  `is_expired()` and `verify_with_threshold()` are Metadata API questions (yes the latter is currently implemented here but it should be in Metadata API in my opinion): I don't object to exceptions but suggest we handle them elsewhere: I recently filed #1377 (which I see you've already seen).

> It will be nice to see this without the repeated logic, such as the big block for checking hashes in meta.

Yes, hash checks should also be part of Metadata API (#1377) in my opinion.

> Also, this appears to be internal API – did we figure the best way to indicate that? Or is that TBD?

My suggestion was to move everything except updater.py and fetcherinterface.py into a tuf/client/_internal/ directory. Would make sense to do before we merge to develop.

tuf/client_rework/metadata_bundle.py
sechkova: I wanted to suggest adding an additional flag marking `root_update_finished`. I see that reference time serves exactly like that but something like `if root_update_finished:` is instantly comprehensible.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,505 @@
...
+        self, role_name: str, delegator_name: str
+    ) -> bool:
+        """Load cached metadata for 'role_name' from local storage.
+
+        Metadata for 'delegator_name' must be loaded already.
+
+        Returns True if metadata was succesfully loaded"""
+        if self.get(role_name):
+            logger.debug("Local %s already loaded", role_name)
+            return True
```
sechkova: Why is this lookup done only for targets? I guess the intention is making the delegations tree traversal more efficient?

sechkova (commented):
I have one doubt about file system access. 
On one hand, by doing it inside the bundle, as it is now, the bundle itself is in control of persisting only verified metadata files that passed all checks which seems less error prone. Also keeps the `Updater` code simpler.

But on the other hand, if the caller (`Updater`) handles both access to the file system and the network download, MetadataBundle will implement only the actual metadata verification steps from the client workflow, will be I/O agnostic and will provide a single public method per metadata type (basically the now private `_load_*` methods).

Other than that, I tried to follow the specification and I think all steps that do not relate to the network download are implemented (besides consistent_snapshot and some TODOs already in the code). But for sure another look is needed when all client components are combined together.

tuf/client_rework/metadata_bundle.py
jku: Yeah it's true we could do it for others too -- it's just that the issue should not come up for top-level metadata but definitely can come up for delegated targets.

jku:
> But on the other hand, if the caller (`Updater`) handles both access to the file system and the network download, MetadataBundle will implement only the actual metadata verification steps from the client workflow, will be I/O agnostic and will provide a single public method per metadata type (basically the now private `_load_*` methods).

These are valid points. I guess the main idea was to get (almost) all spec steps into the same file so it would be easy to follow. But I can definitely be persuaded on this.

sechkova:
> > But on the other hand, if the caller (`Updater`) handles both access to the file system and the network download, MetadataBundle will implement only the actual metadata verification steps from the client workflow, will be I/O agnostic and will provide a single public method per metadata type (basically the now private `_load_*` methods).
> 
> These are valid points. I guess the main idea was to get (almost) all spec steps into the same file so it would be easy to follow. But I can definitely be persuaded on this.

We can come back to this idea when the different PRs are pieced together.

jku:
> But on the other hand, if the caller (`Updater`) handles both access to the file system and the network download, MetadataBundle will implement only the actual metadata verification steps from the client workflow, will be I/O agnostic and will provide a single public method per metadata type (basically the now private `_load_*` methods).

I've been hinking about this and I think I agree with you: this makes the bundle really easy to understand (and makes bundle really do one thing well), and I think the Updater will still be manageable -- it's main function during refresh() is to do I/O, either network or filesystem. I'll do this change as part of this PR

tuf/client_rework/metadata_bundle.py
jku: Exceptions work fine, I'll use them throughout

jku:
First of all, I've rebased so I can use some fixes from newer experimental-client: those who reviewed already can take commit 
800b088 (MetadataBundle: Fix loads of linting issues) as the previous review point, just rebased. So new changes are 
800b088...f2cff95

The major changes are:
* As the last commit, dropped all file I/O as Teodora suggested (this is the last commit on purpose so it's easy to revert if needed, but I do think it makes sense): this means the API is now really simple: one method per metadata type
* API now consistently raises, no bool return values
* Initial tests were added
* reference time is now recorded at initialization

All of the smaller comments should be handled now. This PR also includes 2 commits of PR #1390 to make things work.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,403 @@
...
+
+
+class MetadataBundle(abc.Mapping):
+    """Internal class to keep track of valid metadata in Updater
+
+    MetadataBundle ensures that metadata is valid. It provides easy ways to
+    update the metadata with the caller making decisions on what is updated.
+    """
+
+    def __init__(self, data: bytes):
```
sechkova: Can we do something to concretize that data has to be the initial trusted root data?
Making it clear that  `MetadataBundle` is initialized always from a trusted root.
Yet I'm aware that this happens just a couple of lines below so my comment should be taken as a suggestion only.

jku:
See https://github.com/jku/tuf/blob/experimental-client-use-bundle/tuf/client_rework/updater_rework.py#L241 for a working example of Updater using this code (I'll make that into a PR once this one is merged).

In this version we end up with following lines-of-code counts for the complete client (as counted by sloccount so comments not counted):
```
client_rework   700
api             500
exceptions.py   120
```
This looks pretty good to me

sechkova (approved):
I like this simpler version. I left one suggestion so that I don't leave this review blank :)

tuf/client_rework/metadata_bundle.py
jku: Yeah, maybe I should do one more commit with better docstring coverage.

jku: :heavy_check_mark:

jku:
oops the type hint fix I included in the last commit depended on a python3.9 feature: I removed that, the commit is now just docstrings based onTeodoras last comments

tuf/client_rework/metadata_bundle.py
MVrachev: Wondering is there a possible situation where you can fill up the storage by providing let's say thousands of invalid metadata files? 

Other than that your arguments are sensible.

tests/test_api.py
```diff
@@ -496,6 +496,21 @@ def test_delegation_class(self):
         delegations = Delegations.from_dict(copy.deepcopy(delegations_dict))
         self.assertEqual(delegations_dict, delegations.to_dict())
 
+        # empty keys and roles
+        delegations_dict = {"keys":{}, "roles":[]}
+        delegations = Delegations.from_dict(delegations_dict.copy())
+        self.assertEqual(delegations_dict, delegations.to_dict())
```
joshuagl, line 502: Note: `assertDictEqual()` gives more meaningful messages when this fails.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,464 @@
...
+            # TODO specify the Exceptions (see issue #1351)
+            logger.info("verify failed: %s", e)
+
+    return len(unique_keys) >= role.threshold
+
+
+class MetadataBundle(abc.Mapping):
+    """Internal class to keep track of valid metadata in Updater
+
+    MetadataBundle ensures that the collection of metadata in the bundle valid.
```
joshuagl:
```suggestion
    MetadataBundle ensures that the collection of metadata in the bundle is valid.
```

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,464 @@
...
+    It provides easy ways to update the metadata with the caller making
+    decisions on what is updated.
+    """
+
+    def __init__(self, data: bytes):
+        """Initialize bundle by loading trusted root metadata
+
+        Args:
+            data: Trusted root metadata as bytes. Note that this metadata will
+                will only be verified by itself: it is the source of trust for
```
joshuagl, line 135:
```suggestion
               only be verified by itself: it is the source of trust for
```

tuf/client_rework/metadata_bundle.py
```diff
@@ -142,23 +125,16 @@ class MetadataBundle(abc.Mapping):
     update the metadata with the caller making decisions on what is updated.
     """
 
-    def __init__(self, repository_path: str):
-        """Initialize by loading root metadata from disk"""
-        self._path = repository_path
+    def __init__(self, data: bytes):
```
MVrachev: I would suggest renaming `data` to `root_data` to make it even more descriptive.

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,464 @@
...
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+        self._bundle = {}  # type: Dict[str: Metadata]
+        self.reference_time = datetime.utcnow()
+        self._root_update_finished = False
+
+        # Load and validate the local root metadata
+        # Valid root metadata is required
```
joshuagl:
```suggestion
        # Valid initial trusted root metadata is required
```

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,464 @@
...
+            )
+
+        if self.root is not None:
+            # We are not loading initial trusted root: verify the new one
+            if not verify_with_threshold(self.root, "root", new_root):
+                raise exceptions.UnsignedMetadataError(
+                    "New root is not signed by root", new_root.signed
+                )
+
+            if new_root.signed.version != self.root.signed.version + 1:
```
joshuagl, line 221: Per the spec, this comes _after_ verifying with a threshold of keys from the new root metadata. The code is certainly nicer if it happens in this same `if`... do you think we should move it?

MVrachev (approved):
Overall, it looks like a pretty good implementation.
I left only one suggestion.
Of course, we will iterate over time on this code, but now it looks good to me!

tuf/client_rework/metadata_bundle.py
```diff
@@ -0,0 +1,464 @@
...
+        """
+        if self._root_update_finished:
+            raise RuntimeError("Root update is already finished")
+
+        if self.root.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("New root.json is expired")
+
+        # We skip specification step 5.3.11: deleting timestamp and snapshot
+        # with rotated keys is not needed as they will be invalid, are not
+        # loaded and cannot be loaded
```
joshuagl: This comment feels redundant now that the `MetadataBundle` isn't doing any file handling?

joshuagl (approved):
The final result here looks good. I noticed a couple of docstring issues and a comment that may not make much sense given the removal of disk and network I/O. I did not do a comprehensive review against the spec, but did notice one place where we do things in a slightly different order than specified. Curious to hear your thoughts about that.

eb648d19bc340ebbe3a60d57eae5ddcff76f6b26 makes me a bit anxious – why do we have `Metadata.to_file()` if we can not rely on it to persist metadata which can be verified? Perhaps a topic for a new Issue?

joshuagl:
> [eb648d1](https://github.com/theupdateframework/tuf/commit/eb648d19bc340ebbe3a60d57eae5ddcff76f6b26) makes me a bit anxious – why do we have `Metadata.to_file()` if we can not rely on it to persist metadata which can be verified? Perhaps a topic for a new Issue?

I see we already have https://github.com/theupdateframework/tuf/issues/1392 to capture this. Thanks!

tuf/client_rework/metadata_bundle.py
jku: You're correct... 

I think I'm happy with it being here (this feels like the spec doing things in the non-optimal order and the code happens to look nicer this way as well) but no strong opinions.

jku:
> [eb648d1](https://github.com/theupdateframework/tuf/commit/eb648d19bc340ebbe3a60d57eae5ddcff76f6b26) makes me a bit anxious – why do we have `Metadata.to_file()` if we can not rely on it to persist metadata which can be verified?

This does feel like an oversight in the spec as well: it talks about using canonical forms (and canonical json in particular) for signatures... but hashes aren't talked about (and requiring the implementer to parse canonical json before checking hashes would be wrong  anyway). Even the test files in our repo are not in canonical form WRT whitespace.

On the other hand, it's probably not the end of the world: for the client this is not an issue anyway: writing the original bytes seems a lot safer anyway.

joshuagl:
> > [eb648d1](https://github.com/theupdateframework/tuf/commit/eb648d19bc340ebbe3a60d57eae5ddcff76f6b26) makes me a bit anxious – why do we have `Metadata.to_file()` if we can not rely on it to persist metadata which can be verified?
> 
> This does feel like an oversight in the spec as well: it talks about using canonical forms (and canonical json in particular) for signatures... but hashes aren't talked about (and requiring the implementer to parse canonical json before checking hashes would be wrong anyway). Even the test files in our repo are not in canonical form WRT whitespace.

Some clarifications in the spec, or the oft suggested [secondary literature](https://github.com/theupdateframework/specification/issues/91), would probably help here.

> On the other hand, it's probably not the end of the world: for the client this is not an issue anyway: writing the original bytes seems a lot safer anyway.

Agreed. That's what the current client does and what I believe the spec is suggesting when it states what name to download the file as.

PR #1360, opened 2021-04-21
New API: Add Key and Role classes

### LINKED ISSUE #1139: Add classes for complex metadata fields

MVrachev:
Addresses, but doesn't fix: #1139

  In the top level metadata classes, there are complex attributes such as
  `meta` in Targets and Snapshot, `key` and `roles` in Root etc.

The changes in this pr include:
- addition of a new `Key` class  integrated into Root
- addition of a new `Roles` class integrated into 
- added support for unrecognized fields
- added tests for unrecognized fields in `Key` and `Roles`

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -306,21 +306,21 @@ class TestMetadata(unittest.TestCase):
             root_key2['keytype'], root_key2['scheme'], root_key2['keyval'])
 
         # Assert that root does not contain the new key
-        self.assertNotIn(keyid, root.signed.roles['root']['keyids'])
+        self.assertNotIn(keyid, root.signed.roles['root'].keyids)
         self.assertNotIn(keyid, root.signed.keys)
 
         # Add new root key
         root.signed.add_key('root', keyid, key_metadata)
 
         # Assert that key is added
-        self.assertIn(keyid, root.signed.roles['root']['keyids'])
+        self.assertIn(keyid, root.signed.roles['root'].keyids)
         self.assertIn(keyid, root.signed.keys)
 
         # Remove the key
         root.signed.remove_key('root', keyid)
 
         # Assert that root does not contain the new key anymore
-        self.assertNotIn(keyid, root.signed.roles['root']['keyids'])
+        self.assertNotIn(keyid, root.signed.roles['root'].keyids)
         self.assertNotIn(keyid, root.signed.keys)
 
 
@@ -369,6 +369,9 @@ class TestMetadata(unittest.TestCase):
             if metadata == "root":
                 for keyid in dict1["signed"]["keys"].keys():
                     dict1["signed"]["keys"][keyid]["d"] = "c"
+                for role_str in dict1["signed"]["roles"].keys():
+                    dict1["signed"]["roles"][role_str]["e"] = "g"
+
             temp_copy = copy.deepcopy(dict1)
             metadata_obj = Metadata.from_dict(temp_copy)
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -365,6 +365,10 @@ class TestMetadata(unittest.TestCase):
             # Test that the metadata classes store unrecognized fields when
             # initializing and passes them when casting the instance to a dict.
 
+            # Add unrecognized fields to all metadata sub (helper) classes.
+            if metadata == "root":
+                for keyid in dict1["signed"]["keys"].keys():
+                    dict1["signed"]["keys"][keyid]["d"] = "c"
             temp_copy = copy.deepcopy(dict1)
             metadata_obj = Metadata.from_dict(temp_copy)
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -381,6 +381,11 @@ class TestMetadata(unittest.TestCase):
         self.assertIn(keyid, root.signed.roles['root'].keyids)
         self.assertIn(keyid, root.signed.keys)
 
+        # Try adding the same key again and assert its ignored.
+        pre_add_keyid = root.signed.roles['root'].keyids.copy()
+        root.signed.add_key('root', keyid, key_metadata)
+        self.assertEqual(pre_add_keyid, root.signed.roles['root'].keyids)
+
         # Remove the key
         root.signed.remove_key('root', keyid)
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -25,7 +25,9 @@ from tuf.api.metadata import (
     Metadata,
     Snapshot,
     Timestamp,
-    Targets
+    Targets,
+    Key,
+    Role
 )
 
 from tuf.api.serialization import (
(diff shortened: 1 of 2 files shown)


MVrachev:
There is an interesting reason for the CI [failure](https://github.com/theupdateframework/tuf/pull/1360/checks?check_run_id=2401922288): 
`/home/runner/work/tuf/tuf/tuf/api/metadata.py:532:4: R0914: Too many local variables (18/15) (too-many-locals)`

I used many local variables to make it clear what their purpose is.
Any suggestions? Should I disable this warning with an inline comment or?

tuf/api/metadata.py
```diff
@@ -481,10 +522,14 @@ def from_dict(cls, root_dict: Mapping[str, Any]) -> "Root":
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self. """
         root_dict = self._common_fields_to_dict()
+        keys = {}
+        for keyid, key in self.keys.items():
+            keys[keyid] = key.to_dict()
```
jku: Do we have a stance on comprehension? These would be the places to use it... This would translate to I think
`keys = {keyid: key.to_dict() for (keyid, key) in self.keys.items()}`

That might still fit the line length :)

tuf/api/metadata.py
```diff
@@ -407,6 +407,49 @@ def bump_version(self) -> None:
...
+            "scheme": self.scheme,
+            "keyval": self.keyval,
+            **self.unrecognized_fields,
+        }
+
+        return res_dict
+
+    def get_public_val(self):
+        """Helper function to access the public value of the key."""
+        return self.keyval["public"]
```
jku: I wonder if we should add these helpers only when we know there are users for them?

For example we won't need this in the client as long as the API provides a verify_with_threshold() function...

tuf/api/metadata.py
```diff
@@ -450,6 +450,41 @@ def get_public_val(self):
...
+        keyids: A set of strings each of which represents a given key.
+        threshold: An integer representing the required number of keys for that
+            particular role.
+        unrecognized_fields: Dictionary of all unrecognized fields.
+
+    """
+
+    def __init__(
+        self,
+        keyids: set,
```
jku: `Set[str]` maybe?

That said the input you give when you call this constructor isn't a set I think.

I'm not sure how to handle this right now...

tuf/api/metadata.py
```diff
@@ -513,6 +544,13 @@ def from_dict(cls, root_dict: Mapping[str, Any]) -> "Root":
             unrecognized_key_fields = key
             keys[keyid] = Key(keytype, scheme, keyval, unrecognized_key_fields)
 
+        for role_str, role_dict in roles.items():
+            keyids = role_dict.pop("keyids")
+            threshold = role_dict.pop("threshold")
+            # All fields left in the role_dict are unrecognized.
+            unrecognized_role_fields = role_dict
+            roles[role_str] = Role(keyids, threshold, unrecognized_role_fields)
```
jku: keyids here probably isn't a set but a list (and Role() hopes it is a set)?

tuf/api/metadata.py
```diff
@@ -525,12 +563,15 @@ def to_dict(self) -> Dict[str, Any]:
         keys = {}
         for keyid, key in self.keys.items():
             keys[keyid] = key.to_dict()
+        roles = {}
+        for role_str, role in self.roles.items():
+            roles[role_str] = role.to_dict()
```
jku: _maybe_ this would still be readable with dict comprehension (I may be stretching it now)?

tuf/api/metadata.py
```diff
@@ -397,6 +407,84 @@ def bump_version(self) -> None:
...
+
+    def to_dict(self) -> Dict:
+        """Returns the dictionary representation of self."""
+        res_dict = {
+            "keyids": self.keyids,
+            "threshold": self.threshold,
+            **self.unrecognized_fields,
+        }
+
+        return res_dict
```
jku: nit: could just return the dict without a local variable

mnm678:
> There is an interesting reason for the CI [failure](https://github.com/theupdateframework/tuf/pull/1360/checks?check_run_id=2401922288):
> `/home/runner/work/tuf/tuf/tuf/api/metadata.py:532:4: R0914: Too many local variables (18/15) (too-many-locals)`
> 
> I used many local variables to make it clear what their purpose is.
> Any suggestions? Should I disable this warning with an inline comment or?

This is interesting. I think local variables generally make code more readable, so I'm in favor of disabling this warning.

jku:
> I used many local variables to make it clear what their purpose is.
> Any suggestions? Should I disable this warning with an inline comment or?

I don't think the function looks horrible but ... you could make `Role.from_dict()` and `Key.from_dict()` -- this would fit the existing style and would probably remove half the variables.

tuf/api/metadata.py
```diff
@@ -407,6 +407,49 @@ def bump_version(self) -> None:
...
+    def __init__(
+        self,
+        keytype: str,
+        scheme: str,
+        keyval: Mapping[str, str],
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
+    ) -> None:
+        self.keytype = keytype
+        self.scheme = scheme
+        self.keyval = keyval
```
jku, line 432: I'm fine with storing keyval as is, but should we validate it here (basically check that there is a `keyval["public"]`)? Or are you planning to do all that later?

jku (commented):
Thanks Martin. I have one potential bug (the keyids type issue which I'm not yet sure how to handle), the rest are just nits/suggestions/questions

jku:
There are currently no Key/Role specific tests: We probably should have at least very basic ones if we now have objects?

One thing that might be nice to check for is invalid data -- what does the failure look like if the metadata is not valid (say a missing required field). I don't think this needs to be exhaustively tested but would be nice to see at least an example: I assume with this code we mostly get KeyErrors?

tuf/api/metadata.py
MVrachev: Yes, you are right.
I thought it would be useful, but I am not sure about the use case.
Will remove it.

tuf/api/metadata.py
MVrachev: The Google-style guideliness [here](https://google.github.io/styleguide/pyguide.html#27-comprehensions--generator-expressions) says that:
`Okay to use for simple cases.`
This is simple enough and I will try it.

tuf/api/metadata.py
MVrachev: I tried it and it doesn't look good, because it doesn't fit in the line.

tests/test_api.py
```diff
@@ -365,6 +424,13 @@ def test_support_for_unrecognized_fields(self):
             # Test that the metadata classes store unrecognized fields when
             # initializing and passes them when casting the instance to a dict.
 
+            # Add unrecognized fields to all metadata sub (helper) classes.
+            if metadata == "root":
+                for keyid in dict1["signed"]["keys"].keys():
+                    dict1["signed"]["keys"][keyid]["d"] = "c"
+                for role_str in dict1["signed"]["roles"].keys():
+                    dict1["signed"]["roles"][role_str]["e"] = "g"
```
jku, line 443: this looks like it will become hard to maintain when more tests are added but I'm fine with it for now: can you make an issue about doing this in a more structured way (like automate the dictionary poisoning so that ever dictionary is injected, not just ones we list here)?

tuf/api/metadata.py
```diff
@@ -405,6 +405,96 @@ def bump_version(self) -> None:
...
+        self,
+        keyids: Set[str],
+        threshold: int,
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
+    ) -> None:
+        keyids_set = set(keyids)
+        if len(keyids_set) != len(keyids):
+            raise ValueError(
+                f"keyids should be a set, instead got a list with dublicates!"
+            )
```
jku: typo on duplicate.

More importantly, there's still a mismatch with input and the type hints. If your code expects a set here, then the calling code must make the length check. Alternatively if the check is left here, then the constructor argument type is actually a list. Currently from_dict() already creates a set _without doing the length check_ and then calls this function -- so this error is likely never raised.

jku (changes requested):
Just one real issue: the keyids uniqueness checks and creation of the set still have a smell around them: I think the error isn't currently raised (when using `from_dict()`).

tests/test_api.py
MVrachev: Yes, you are right. We should automate this.
In a discussion, you showed a prototype automating this process.
Will appreciate it if you propose these changes and we can discuss them, but I don't think this is in the scope of this pr.

MVrachev:
Updated the pr by:
- changing the type of `keyds` to a list inside the `Role.__init__` 
-  don't make create a set in `Role.from_dict` and left the `Role.__init__` to make that check
- added a unit test case for creating a `Role` instance with duplicates in `keyid`.

MVrachev:
Rebased after #1363 was merged.

tuf/api/metadata.py
```diff
@@ -490,17 +577,17 @@ def add_key(
         self, role: str, keyid: str, key_metadata: Mapping[str, Any]
     ) -> None:
         """Adds new key for 'role' and updates the key store."""
-        if keyid not in self.roles[role]["keyids"]:
-            self.roles[role]["keyids"].append(keyid)
+        if keyid not in self.roles[role].keyids:
+            self.roles[role].keyids.add(keyid)
```
jku: only realized this now but the check here is not needed with a set

jku (approved):
I think this looks good. I did notice one unneeded "if" that I had missed earlier (adding duplicates to a set is safe).

PR #1364, opened 2021-04-27
tox: Don't use pre-release tools

### LINKED ISSUE #1350: Let's not use pre-release versions of tools in tox environments
We do `pip install --pre ...` in tox.ini. This means our testing tools like coverage, black, isort etc (and I suppose their dependencies) may be pre-releases.

This has been the case since 2015 but it sounds like a bad idea. Should we drop the `--pre`?

jku:
Keep using newest versions of build tools, but don't use pre-releases:
* less downloading (with chances of failures) happens in the CI because
  versions change less often
* tools are less likely to break the build (or at least it happens less
  often)

This change should not affect the software under test as we install
pinned versions of those.

Note that this also doesn't affect black: tox still ends up with a
pre-release version of black because that project has never made an
actual release.

Fixes #1350

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ deps =
     # installation (see `skipsdist`), to get relative paths in coverage reports
     --editable {toxinidir}
 
-install_command = pip install --pre {opts} {packages}
+install_command = pip install {opts} {packages}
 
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
Thank you

PR #1366, opened 2021-04-28
Prototype descriptors and validation functions

MVrachev:

  This experiment tries to help us envision how validation will look like
  if we decide to use validation functions + descriptors.
  
  What I found with this experiment is that python descriptors can't
  handle dictionary keys validation as well as dictionary values.
  More precisely, we can't validate that each of the "roles" in Root is
  one of the 4 metadata types or that Root role keyids are unique.
  
  The reason is that for example if we are to run this:
  "root.roles = 3" this will invoke validation for root.roles, but
  if we invoke: "root.roles["dwad"] = 3" then we are getting an element
  of the dictionary roles and then assigning it a new value.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -382,6 +382,8 @@ class TestMetadata(unittest.TestCase):
             )
 
 
+class TestValidation(unittest.TestCase):
+
     def test_validate_signed_attr(self):
         # spec_version validation
         for val in [None, True, 111, 1.1]:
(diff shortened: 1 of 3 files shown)


jku:
Writing down some thoughts on this (wall of text sorry).


The task at hand is not "write validation code" -- or if it was that I want to now change it to _"1. improve security against malicious json, and 2. make API easy to use correctly and hard to use incorrectly (with least amount of most obvious code)"_. This is abstract but I do have practical suggestions below. The reason I say this is that I fear we are concentrating too much on adding validation when there are other solutions (like modifying the implementation) to achieve the same goals.

How should we proceed with this task? I believe we need to one by one analyze individual items in the API (starting with the small bits like strings and continuing to larger items). For each item we should consider and document
 * what can go wrong with this item
 * where is this used? where is it set? Should it be used in these places?
 * what can we do to improve the safety/usability?
 * if we choose validation, define what exactly do we want to validate
 
and only then implement chosen improvement, if any. Many of these improvements will include validation but _we should not add validation if we can achieve the goals without it_ and validation may be only part of the solution to a specific issue. Also what _exactly_ is validated should be spelled out somewhere -- otherwise it's impossible to tell if the validation implementation is correct or not.

This is hard work but I believe it's the path to better code. Trying to implement a lot validation at once will not lead to good quality.

As examples I went through analysis for _type and spec_version:
* _type should not be user-settable at all (it does not make sense), and deserialization validation of _type happens in Metadata already (the fact that type is a Signed constructor argument is IMO a bug that we should fix: each Signed implementation should know its type without it being told what it is): no validation should be required here
*  spec_version: deserialization should be validated but there does not seem any reason for spec_version attribute to be writable: we probably want to prevent modifying that instead of validating the new value. Calling a validation function from constructor only seems fine to me (if modifying the public attribute is not possible) but other options do exist. Validation should include 
   * value can be parsed as three ints separated by dots (specified by semver)
   * _maybe_ comparison to library supported version? -- this is implementation defined, not specified so we can choose. Updater currently raises on major version mismatch and logs on minor
 
I would start with individual PRs with that sort of analysis and proposed fixes. I would probably start by just adding the validation calls in constructors and maybe making the attributes properties with setters and see how it goes from there -- but I can see the value in trying other integration methods as well (descriptors/decorations)... I'm assuming a single solution is not going to cover all cases and starting simple is good (so we can change our minds).

More detail comments:
* keeping validation functions outside of metadata.py seems fine to me
* looking at the example, descriptors IMO hide the validation and are still magic: When I'm reading metadata.py I can't see what gets validated and when, and I think that's bad. I don't need to know details of the validation but I would like to see e.g. `validate_spec_version(spec_version)` in the constructor or a property setter (or maybe it's a `@validate_spec_version` instead) -- anything that makes it easy to see what is validated, and easy to follow through to the validation code.
* I don't like the type checks -- they may be useful in specific situations but generally speaking I suggest letting the validation fail with TypeErrors or AttributeErrors instead 
* I do like the idea of testing how validation reacts to completely wrong types though (I suggest only covering types that json can contain) -- this should give us a good idea where we might need better type checks

MVrachev:
I agree with the general idea to rethink how we store and operate with the different metadata fields.

I plan to follow this process for each of the metadata attributes:
1. How do we use it? 
2. What might go wrong? 
3. Rethink how we store it?
4. If, after the changes in step 4 there are concerns, then add a validation function.
5. Make the necessary changes for points 1 - 5., think and propose a way to integrate it into the class/classes (decorators, descriptors, etc.)
6. create a pr for all additional changes on top of the validation functions.
7. document my findings in a commit message and/or somewhere else

MVrachev:
Closing this one as it's no longer applicable.
We decided we will research each attribute individually and this decide on its validation.

Issue #1333, opened 2021-04-06
New API: Don't hardcode metadata files extension


MVrachev:
**Description of issue or feature request**:
In a review of a pr adding additional classes in the metadata API, @sechkova noticed that we have hardcoded
the usage of JSON files representing our metadata files after the implementation of abstract (de)serializer  [here](https://github.com/theupdateframework/tuf/pull/1279).

For context: [Teodoras comment](https://github.com/theupdateframework/tuf/pull/1329#discussion_r607214167)

**Current behavior**:
Hardcoded `*.json` file names in a lot of places: https://github.com/theupdateframework/tuf/blob/52207862008cc54db8a43e1543a653be0d0333f7/tuf/api/metadata.py#L468

**Expected behavior**:
Don't hardcode the file names representing our metadata files with their file extensions.

TODO:

- [ ] Remove hardcoded `json` strings from the code in all classes.
- [ ] Fix our documentation to not give examples with JSON files or have a sentence like `For JSON metadata files, here is how it will look like...`
- [ ] Add support for other metadata file types inside the metadata classes. 
This could be achieved with a configuration option in a `Repository` class or helper functions for each of the metadata classes.

MVrachev:
Added commit https://github.com/theupdateframework/tuf/pull/1329/commits/15723859dfefd6bc9412f7bbc172fe1f2570297b which removed part of the places where we used
hardcoded `JSON string and added support for other file extensions in the `Timestamp` class.

PR #1367, opened 2021-04-28
Make hashes, length and delegations optional + improvements

### LINKED ISSUE #1362: Metadata API: Targets.delegations should be optional
The delegation field in all targets type metadata seems to be optional by specification (see current reference implementation and https://github.com/theupdateframework/specification/pull/157)

This is a potentially breaking change for a client but fortunately no-one's built one with Metadata API yet.

**Current behavior**:
In Metadata API deserializing target metadata currently fails if delegations field is missing.
The Target constructor argument is not optional.

**Expected behavior**:
"delegations" field should not be required in Metadata API deserialization and.  delegations=None should lead to serialized format that does not include the field "delegations".

MVrachev:
Fixes: #1362

**Please note: this pr contains one breaking change:**
1. make `delegations` optional for `Targets`

The idea of this pr came from improvements proposed/discussed in pr #1329.
Those improvements are proposed together (because they are small enough to be bundled) in a
separate pr and with that to simplify the review on pr #1329.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -23,6 +23,7 @@ from tests import utils
 import tuf.exceptions
 from tuf.api.metadata import (
     Metadata,
+    Root,
     Snapshot,
     Timestamp,
     Targets,
@@ -98,6 +99,7 @@ class TestMetadata(unittest.TestCase):
 
     def test_generic_read(self):
         for metadata, inner_metadata_cls in [
+                ('root', Root),
                 ('snapshot', Snapshot),
                 ('timestamp', Timestamp),
                 ('targets', Targets)]:
@@ -144,7 +146,7 @@ class TestMetadata(unittest.TestCase):
 
 
     def test_read_write_read_compare(self):
-        for metadata in ['snapshot', 'timestamp', 'targets']:
+        for metadata in ['root', 'snapshot', 'timestamp', 'targets']:
             path = os.path.join(self.repo_dir, 'metadata', metadata + '.json')
             metadata_obj = Metadata.from_file(path)
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -445,6 +445,14 @@ class TestMetadata(unittest.TestCase):
         # Verify that data is updated
         self.assertEqual(targets.signed.targets[filename], fileinfo)
 
+        # Test from_dict/to_dict Targets without delegations
+        targets_dict = targets.to_dict()
+        del targets_dict["signed"]["delegations"]
+        tmp_dict = targets_dict["signed"].copy()
+        targets_obj = Targets.from_dict(tmp_dict)
+        tar_d = targets_obj.to_dict()
+        self.assertEqual(targets_dict["signed"], targets_obj.to_dict())
+
     def setup_dict_with_unrecognized_field(self, file_path, field, value):
         json_dict = {}
         with open(file_path) as f:
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -260,6 +260,18 @@ class TestMetadata(unittest.TestCase):
         snapshot.signed.update('role1', 2, 123, hashes)
         self.assertEqual(snapshot.signed.meta, fileinfo)
 
+        # Update only version. Length and hashes are optional.
+        snapshot.signed.update('role2', 3)
+        fileinfo['role2.json'] = {'version': 3}
+        self.assertEqual(snapshot.signed.meta, fileinfo)
+
+        # Test from_dict and to_dict without hashes and length.
+        snapshot_dict = snapshot.to_dict()
+        test_dict = snapshot_dict['signed'].copy()
+        del test_dict['meta']['role1.json']['length']
+        del test_dict['meta']['role1.json']['hashes']
+        snapshot = Snapshot.from_dict(test_dict)
+        self.assertEqual(snapshot_dict['signed'], snapshot.to_dict())
 
     def test_metadata_timestamp(self):
         timestamp_path = os.path.join(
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -336,6 +336,8 @@ def __init__(
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
 
+        # "_type" is not actually meant to be "protected" in the OOP sense
+        # but rather is a workaround to not shadow the built-in "type" name.
```
jku: I've been wondering, is this actually true for the object attribute?  As in, I get that arguments/variables shouldn't be named `type` but why can't the attribute be `Signed.type` -- it doesn't shadow anything?

I don't think `Signed.type` would be confusing to user (in fact I'd argue it's in line with imitating the file format) and I don't think there's actually anything preventing us from calling it that as `type` is not a reserved keyword... Maybe I'm missing a tool that complains about this?

tests/test_api.py
```diff
@@ -258,6 +258,10 @@ def test_metadata_snapshot(self):
         snapshot.signed.update('role1', 2, 123, hashes)
         self.assertEqual(snapshot.signed.meta, fileinfo)
 
+        # Update only version. Length and hashes are optional.
+        snapshot.signed.update('role1', 3)
+        fileinfo['role1.json'] = {'version': 3}
+        self.assertEqual(snapshot.signed.meta, fileinfo)
```
jku: A de/serialize test with json without length/hashes would be good for future

tests/test_api.py
```diff
@@ -293,6 +297,10 @@ def test_metadata_timestamp(self):
         timestamp.signed.update(2, 520, hashes)
         self.assertEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
 
+        # Update only version. Length and hashes are optional.
+        timestamp.signed.update(3)
+        fileinfo = {'version': 3}
+        self.assertEqual(timestamp.signed.meta['snapshot.json'], fileinfo)
```
jku, line 321: A de/serialize test with json without length/hashes would be good for future

tuf/api/metadata.py
```diff
@@ -557,14 +557,19 @@ def to_dict(self) -> Dict[str, Any]:
...
-            "length": length,
-            "hashes": hashes,
-        }
+        self.meta["snapshot.json"] = {"version": version}
+
+        if length is not None:
+            self.meta["snapshot.json"]["length"] = length
+
+        if hashes is not None:
+            self.meta["snapshot.json"]["hashes"] = hashes
```
jku, line 666: nit: Not a huge fan of three separate lookups of "snapshot.json" key... it's not a major deal here, just pointing it out

jku (changes requested):
Thanks. The actual API changes look fine to me
* length+hashes commit message is misleading in that this does not seem like a breaking change to me (looks like de/serialize probably works fine right now). Am I missing something?
* I prefer no TODO comments for existing issues (without a very good reason). If the issue is closed as wontfix, who fixes the TODOs?
* is there a benefit to using restructuredtext in docstrings -- do you (or are we planning to) run this through sphinx somewhere?

As a process comment I think multiple PRs instead of a single one might have increased the throughput here: now we may be stuck discussing a comment about _type when we could just merge the API changes.

tuf/api/metadata.py
MVrachev: I searched for that and one relevant response from Stackoverflow: https://stackoverflow.com/a/10568115/7058471.
I am not against or for renaming.

joshuagl: One other reason to use `_type` rather than `type` is that it better matches the spec. 

Historical context: the use of `_type` in the spec was to match the implementation, which did not want to shadow the builtin `type()` method. Shadowing `type()` was more problematic in a procedural implementation, and will likely be less of an issue in the OOP metadata API – but the name in the spec is `_type` until we do something with https://github.com/theupdateframework/specification/issues/149 and a possible 2.0 of the specification.

tuf/api/metadata.py
MVrachev: I imagine this would be something we would change when we fix #1333.

MVrachev:
Rebased because of conflicts with develop.
Updates include:
- added additional tests for (de)serialization without hashes and length for Timestamp and Snapshot
- removed the commit with the TODO comments
- changed the length and hashes commit message

tuf/api/metadata.py
jku: I don't think that's related: you don't need more than a single lookup even now but it is  a small detail

tuf/api/metadata.py
jku: oof I forgot the spec uses _type . Thanks

tuf/api/metadata.py
```diff
@@ -504,16 +504,33 @@ class Root(Signed):
             supports consistent snapshots.
         keys: A dictionary that contains a public key store used to verify
             top level roles metadata signatures::
-            {
-                '<KEYID>': <Key instance>,
-                ...
-            },
```
jku, line 517: is this change of keyid content intentional? I think the opposite change was just done a week ago

jku:
I think it looks good but I  wanted to check if the documentation change is intentional? it's not mentioned in the commit message and the opposite change was done quite recently I think

tuf/api/metadata.py
MVrachev: You are right, this is an unintentional change coming from the time when I had those changes before we added Key and Role classes.
Noticed the same for Roles.
Will fix it.
Great catch!

jku (approved):

PR #1368, opened 2021-04-29
client refactor: remove mirrors

sechkova:
Addresses #1307 

Drop "mirrors" support. 
Updater no longer supports downloading each metadata file from a list of user-provided mirrors. All downloads are performed from a single metadata url. 
Target files are downloaded either from a single default url or an optional url is passed for each target file on download.

More info in: [comment](https://github.com/theupdateframework/tuf/issues/1307#issuecomment-827359322), [comment](https://github.com/theupdateframework/tuf/issues/1307#issuecomment-827435529
)

API is changed the following way:

```
class Updater:
     def __init__(
         self,
         repository_name: str,
-        repository_mirrors: Dict,
+        metadata_url: str,
+        default_target_url: Optional[str] = None,
     ):



    def download_target(
         self,
         targetinfo: Dict,
         destination_directory: str,
+        target_url: Optional[str] = None,
     ):
```

_Edit:_ The exact `metdata_url` and `target_url` format needs either better documentation for the user or better normalization functions provided by TUF.

requirements**:


tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,26 @@ class Updater:
...
-        consistent_snapshot:
+    TODO
     """
 
     def __init__(
         self,
         repository_name: str,
-        repository_mirrors: Dict,
+        metadata_url: str,
+        default_target_url: Optional[str] = None,
```
joshuagl, line 46: nit: these are prefixes or bases, not full urls – should we call them `metadata_base_url` and `default_target_base_url`?
Or perhaps `default_base_url` and `targets_base_url` ?

tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,26 @@ class Updater:
...
         fetcher: Optional[FetcherInterface] = None,
     ):
 
         self._repository_name = repository_name
-        self._mirrors = repository_mirrors
+        self._metadata_url = metadata_url
+        # Should we accept metadata url as a default for targets or
+        # targets_url should be provided either in this constructor
+        # or as a download_target parameter?
```
joshuagl, line 54: I think it makes sense to default to using the same host/url combination for all downloads, unless told otherwise via the `targets_base_url` or the optional function parameters.

tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,26 @@ class Updater:
...
+        if default_target_url is None:
+            self._default_target_url = metadata_url
+        else:
+            self._default_target_url = default_target_url
```
joshuagl, line 58: Nit: I'd probably write this as 
```suggestion
        self._default_target_url = metadata_url
        if default_target_url is not None:
            self._default_target_url = default_target_url
```

tuf/client_rework/updater_rework.py
```diff
@@ -142,44 +142,38 @@ def updated_targets(targets: Dict, destination_directory: str) -> Dict:
 
         return updated_targets
 
-    def download_target(self, target: Dict, destination_directory: str):
+    def download_target(
+        self,
+        targetinfo: Dict,
+        destination_directory: str,
+        target_url: Optional[str] = None,
```
joshuagl, line 149: Nit: `target_url_base` ?

tuf/client_rework/updater_rework.py
```diff
@@ -142,44 +142,38 @@ def updated_targets(targets: Dict, destination_directory: str) -> Dict:
...
+        full_url = _build_full_url(target_url, targetinfo["filepath"])
 
-        filepath = os.path.join(destination_directory, target["filepath"])
+        temp_obj = None
+        try:
+            temp_obj = download.download_file(
+                full_url, targetinfo["fileinfo"]["length"], self._fetcher
+            )
+            _check_file_length(temp_obj, targetinfo["fileinfo"]["length"])
+            temp_obj.seek(0)
```
joshuagl, line 166: I wonder whether `_check_file_length()` should do the `seek(0)`, what do you think?

tuf/client_rework/updater_rework.py
```diff
@@ -142,44 +142,38 @@ def updated_targets(targets: Dict, destination_directory: str) -> Dict:
...
+            _check_hashes_obj(temp_obj, targetinfo["fileinfo"]["hashes"])
+
+        except Exception as e:
+            if temp_obj:
+                temp_obj.close()
+            # TODO: do we reraise a NoWorkingMirrorError or just
+            # let exceptions propagate?
+            raise exceptions.NoWorkingMirrorError({full_url: e}) from e
```
joshuagl, line 174: (here and below) does `NoWorkingMirrorError` still make sense, given we're removing mirrors?

tuf/client_rework/updater_rework.py
```diff
@@ -903,11 +831,17 @@ def _get_target_hash(target_filepath, hash_function="sha256"):
...
+    a relative file path.
     """
-    if isinstance(mirror_error, exceptions.FetcherHTTPError):
-        if mirror_error.status_code in {403, 404}:
-            return False
-    return True
+    # Are these steps enough? Or too much? Is this the right place?
+    filepath = parse.quote(filepath)
+    # Assuming that base_url ends with a '/' character, otherwise parse.urljoin
+    # omits (correcly) the last part of the base URL path
```
joshuagl, line 842:
```suggestion
    # omits (correctly) the last part of the base URL path
```

joshuagl (commented):
Apologies for quick review, but I wanted to share some quick impressions. This is a pleasing diff stat. I think we might be able to reduce it further: the `_build_full_url()` / `try:`, `download()`, `verify()` loop is repeated a few times. Might it generalise to a helper function?

tuf/client_rework/updater_rework.py
jku: I don't think it works though: metadata base url directory should contain metadata (it's like current `url_prefix+metadata_path`). I would be surprised if someone wants to store targets in the same directory. I think it's reasonable to expect either a default target url prefix or the individual target download url prefixes

jku:
> This is a pleasing diff stat. I think we might be able to reduce it further

Absolutely: e.g. we can now get rid of almost all manual file object handling

tuf/client_rework/updater_rework.py
trishankatdatadog: Jussi, I agree with you, but I think Joshua simply means the following: there is a `metadata_base_url` parameter that must always be explicitly set. However, there is also a `targets_base_url` that defaults to `metadata_base_url` unless also explicitly set to something else.

jku: I think I understood: I'm just wondering if anyone would want it to default to metadata_base_url?

jku:
Since Teodora is away for the week I'll take this and continue with suggested fixes (and removing loads of temp file handling). I think I'll do a new PR and close this (but will keep these commits so review can "continue")

jku:
superseded  by #1373

PR #1369, opened 2021-04-29
CI: Allow failure when publishing on coveralls

sechkova:

A failure during publishing of the coverage results on coveralls should not fail the whole build job.
Allow the step to fail.

A side effect of this change is that failures in this step will be unnoticed by PR authors unless intentionally looking at the CI run log.

requirements**:


jku (approved):
Agreed: Noticing a 500-error from Coverage is less important than reliable testing on CI (this is IIRC the same setup we had on travis)

PR #1370, opened 2021-04-29
New API: Add DelegationRole and Delegations classes

MVrachev:
Addresses #1139

  In the top-level metadata classes, there are complex attributes such as
  "meta" in Targets and Snapshot, "key" and "roles" in Root etc.
  
  DelegationRole shares a couple of fields with the Role class and that's
  why it inherits it.
  I decided to use a separate Delegations class because I thought it will
  make it easier to read, verify and add additional helper functions.
  Also, I tried to make sure that I test each level of the delegations
  representation for support of storing unrecognized fields.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -477,6 +477,15 @@ class TestMetadata(unittest.TestCase):
                     dict1["signed"]["keys"][keyid]["d"] = "c"
                 for role_str in dict1["signed"]["roles"].keys():
                     dict1["signed"]["roles"][role_str]["e"] = "g"
+            elif metadata == "targets" and dict1["signed"].get("delegations"):
+                for keyid in dict1["signed"]["delegations"]["keys"].keys():
+                    dict1["signed"]["delegations"]["keys"][keyid]["d"] = "c"
+                new_roles = []
+                for role in dict1["signed"]["delegations"]["roles"]:
+                    role["e"] = "g"
+                    new_roles.append(role)
+                dict1["signed"]["delegations"]["roles"] = new_roles
+                dict1["signed"]["delegations"]["foo"] = "bar"
 
             temp_copy = copy.deepcopy(dict1)
             metadata_obj = Metadata.from_dict(temp_copy)
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -28,7 +28,9 @@ from tuf.api.metadata import (
     Timestamp,
     Targets,
     Key,
-    Role
+    Role,
+    Delegations,
+    DelegatedRole,
 )
 
 from tuf.api.serialization import (
@@ -328,7 +330,7 @@ class TestMetadata(unittest.TestCase):
                     "public": "edcd0a32a07dce33f7c7873aaffbff36d20ea30787574ead335eefd337e4dacd"
                 },
                 "scheme": "ed25519"
-                        },
+            },
         }
         for key_dict in keys.values():
             # Testing that the workflow of deserializing and serializing
@@ -421,6 +423,76 @@ class TestMetadata(unittest.TestCase):
 
         with self.assertRaises(KeyError):
             root.signed.remove_key('root', 'nosuchkey')
+
+    def test_delegated_role_class(self):
+        roles = [
+            {
+                "keyids": [
+                    "c8022fa1e9b9cb239a6b362bbdffa9649e61ad2cb699d2e4bc4fdf7930a0e64a"
+                ],
+                "name": "role1",
+                "paths": [
+                    "file3.txt"
+                ],
+                "terminating": False,
+                "threshold": 1
+            }
+        ]
+        for role in roles:
+            # Testing that the workflow of deserializing and serializing
+            # a delegation role dictionary doesn't change the content.
+            key_obj = DelegatedRole.from_dict(role.copy())
+            self.assertEqual(role, key_obj.to_dict())
+
+            # Test creating a DelegatedRole object with both "paths" and
+            # "path_hash_prefixes" set.
+            role["path_hash_prefixes"] = "foo"
+            with self.assertRaises(ValueError):
+                DelegatedRole.from_dict(role.copy())
+
+            # Test creating DelegatedRole only with "path_hash_prefixes"
+            del role["paths"]
+            DelegatedRole.from_dict(role.copy())
+            role["paths"] = "foo"
+
+            # Test creating DelegatedRole only with "paths"
+            del role["path_hash_prefixes"]
+            DelegatedRole.from_dict(role.copy())
+            role["path_hash_prefixes"] = "foo"
+
+            # Test creating DelegatedRole without "paths" and
+            # "path_hash_prefixes" set
+            del role["paths"]
+            del role["path_hash_prefixes"]
+            DelegatedRole.from_dict(role)
+
+
+    def test_delegation_class(self):
+        roles = [
+                {
+                    "keyids": [
+                        "c8022fa1e9b9cb239a6b362bbdffa9649e61ad2cb699d2e4bc4fdf7930a0e64a"
+                    ],
+                    "name": "role1",
+                    "paths": [
+                        "file3.txt"
+                    ],
+                    "terminating": False,
+                    "threshold": 1
+                }
+            ]
+        keys = {
+                "59a4df8af818e9ed7abe0764c0b47b4240952aa0d179b5b78346c470ac30278d":{
+                    "keytype": "ed25519",
+                    "keyval": {
+                        "public": "edcd0a32a07dce33f7c7873aaffbff36d20ea30787574ead335eefd337e4dacd"
+                    },
+                    "scheme": "ed25519"
+                },
+            }
+        delegations_dict = {"keys": keys, "roles": roles}
+        delegations = Delegations.from_dict(copy.deepcopy(delegations_dict))
+        self.assertEqual(delegations_dict, delegations.to_dict())
 
 
     def test_metadata_targets(self):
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -728,6 +728,135 @@ def update(
...
+        self.keys = keys
+        self.roles = roles
+        self.unrecognized_fields = unrecognized_fields or {}
+
+    @classmethod
+    def from_dict(cls, delegations_dict: Mapping[str, Any]) -> "Delegations":
+        """Creates Delegations object from its dict representation."""
+        keys = delegations_dict.pop("keys")
+        for keyid, key_dict in keys.items():
+            keys[keyid] = Key.from_dict(key_dict.copy())
```
jku: modifying a dictionary while iterating through it is traditionally a red flag... maybe worth building a new dict?

tuf/api/metadata.py
```diff
@@ -728,6 +728,135 @@ def update(
...
+        roles = delegations_dict.pop("roles")
+        roles_res = []
+        for role_dict in roles:
+            new_role = DelegationRole.from_dict(role_dict.copy())
+            roles_res.append(new_role)
```
jku: I don't think you need a copy() here?

tuf/api/metadata.py
```diff
@@ -728,6 +728,135 @@ def update(
...
+            keys[keyid] = key.to_dict()
+        roles = []
+        for role_obj in self.roles:
+            roles.append(role_obj.to_dict())
+        res_dict = {
+            "keys": keys,
+            "roles": roles,
+            **self.unrecognized_fields,
+        }
+        return res_dict
```
jku: nit: could just return dict without variable

tuf/api/metadata.py
```diff
@@ -808,17 +907,18 @@ def from_dict(cls, targets_dict: Mapping[str, Any]) -> "Targets":
         """Creates Targets object from its dict representation."""
         common_args = cls._common_fields_from_dict(targets_dict)
         targets = targets_dict.pop("targets")
-        delegations = targets_dict.pop("delegations")
+        delegations_dict = targets_dict.pop("delegations")
```
jku: if delegations is optional then this is clearly a bug.

This is properly fixed in #1367

tuf/api/metadata.py
```diff
@@ -808,17 +907,18 @@ def from_dict(cls, targets_dict: Mapping[str, Any]) -> "Targets":
...
+        return cls(*common_args, targets, delegations_obj, targets_dict)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self."""
         targets_dict = self._common_fields_to_dict()
         targets_dict.update(
             {
                 "targets": self.targets,
-                "delegations": self.delegations,
+                "delegations": self.delegations.to_dict(),
```
jku: bug when delegations is None: This is properly fixed in #1367

tuf/api/metadata.py
```diff
@@ -728,6 +728,135 @@ def update(
...
+        return cls(keys, roles_res, delegations_dict)
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Returns the dict representation of self."""
+        keys = {}
+        for keyid, key in self.keys.items():
+            keys[keyid] = key.to_dict()
+        roles = []
+        for role_obj in self.roles:
+            roles.append(role_obj.to_dict())
```
jku: could consider dict/list comprehension again but up to you really

tuf/api/metadata.py
```diff
@@ -728,6 +728,135 @@ def update(
...
+        self.terminating = terminating
+        if paths and path_hash_prefixes:
+            raise ValueError(
+                "Only one of the attributes 'paths' and"
+                "'path_hash_prefixes' can be set!"
+            )
+        if paths:
+            self.paths = paths
+        elif path_hash_prefixes:
+            self.path_hash_prefixes = path_hash_prefixes
```
jku: if-else leads to undefined object attributes. Some variant of this is not tested.

tests/test_api.py
```diff
@@ -441,6 +503,15 @@ def test_support_for_unrecognized_fields(self):
...
                     dict1["signed"]["roles"][role_str]["e"] = "g"
+            elif metadata == "targets" and dict1["signed"].get("delegations"):
+                for keyid in dict1["signed"]["delegations"]["keys"].keys():
+                    dict1["signed"]["delegations"]["keys"][keyid]["d"] = "c"
+                new_roles = []
+                for role in dict1["signed"]["delegations"]["roles"]:
+                    role["e"] = "g"
+                    new_roles.append(role)
+                dict1["signed"]["delegations"]["roles"] = new_roles
+                dict1["signed"]["delegations"]["foo"] = "bar"
```
MVrachev, line 559: I know this is ugly, but until @jku proposes his test where he automatically validates support for unrecognized fields on each level of the dictionary, we would have to stick to this.

tuf/api/metadata.py
```diff
@@ -728,6 +728,135 @@ def update(
...
+    """
+
+    def __init__(
+        self,
+        name: str,
+        keyids: list,
+        threshold: int,
+        terminating: bool,
+        paths: Optional[list] = None,
+        path_hash_prefixes: Optional[list] = None,
```
jku: three list type hints without list-item type hints

tuf/api/metadata.py
```diff
@@ -728,6 +728,135 @@ def update(
...
+        roles: A list of DelegationRole instances containing information about
+            all delegated roles.
+        unrecognized_fields: Dictionary of all unrecognized fields.
+
+    """
+
+    def __init__(
+        self,
+        keys: Mapping[str, Key],
+        roles: list,
```
jku: missing list-item hint

jku (changes requested):
I'll take another look after #1367 fixes are done and this is rebased: currently it looks buggy

tests/test_api.py
jku: the reason I haven't done it is that while there seemingly was a consensus that "unrecognised keys should be allowed everywhere in the file format" ... it's not really true. There's a lot of places where that would be weird and unsafe (what if someone inserts random data into "keys" or "roles") and we should not allow that. The real rule is probably "unreconised keys should be allowed at every _object_ root level": so you should be allowed to insert a key into Signed, Key or Role ... but not into any dictionary where we cannot know the recognised key names before hand (like "keys") or where the allowed keys are strictly defined (like "meta").

Poisoning everything except a hard-coded list of dictionaries (instead of only poisoning specific dicts) may still be the better testing solution but it's not trivial.

jku: there's an implementation in #1382 now

tuf/api/metadata.py
MVrachev: I don't think there is a mistake here.
From Joshua's comment [here](https://github.com/theupdateframework/specification/issues/156#issuecomment-823447318), I understand we can have one of them or either of them. The checks I made are:
- we can't have both of them (first if)
- we can have only `paths` (second if)
- we can have only `path_hash_prefixes` (the elif)
- we can have an object without any of them.

What I miss is a specific test for creating `DelegationRole` only with one of `paths` and `path_hash_prefixes`.

tuf/api/metadata.py
MVrachev: I am returning a new dictionary by calling `key_dict.copy()`.
Should I add a new variable to make it clearer?

tuf/api/metadata.py
jku: the instance attributes are left unset so I think `to_dict() `will fail in some of those cases (at least when you only have path_hash_prefixes). I don't think there should be any if-else here (except for the validation that raises)

tuf/api/metadata.py
MVrachev: Will try them.

tuf/api/metadata.py
MVrachev: Just tested it and you are right. We don't need the if-else here.

MVrachev:
Rebased on top of the latest develop branch and addressed all comments made by @jku.

tuf/api/metadata.py
```diff
@@ -740,6 +740,129 @@ def update(
...
+        self.path_hash_prefixes = path_hash_prefixes
+
+    @classmethod
+    def from_dict(cls, role_dict: Dict[str, Any]) -> "Role":
+        """Creates DelegationRole object from its dict representation."""
+        name = role_dict.pop("name")
+        keyids = role_dict.pop("keyids")
+        threshold = role_dict.pop("threshold")
+        terminating = role_dict.pop("terminating")
+        paths = paths = role_dict.pop("paths", None)
```
```suggestion
        paths = role_dict.pop("paths", None)
```

jku (commented):
This looks good to me, thanks.

I left one typo comment and have a naming question but it's basically :heavy_check_mark: for me.

Naming: Would DelegatedRole be better than DelegationRole -- maybe it conveys a little bit more information (in that this clearly is the role that is delegated, not the one that is delegating)? I don't feel strongly about it though, just want to make sure we try to find the best names...

MVrachev:
> This looks good to me, thanks.
> 
> I left one typo comment and have a naming question but it's basically ✔️ for me.
> 
> Naming: Would DelegatedRole be better than DelegationRole -- maybe it conveys a little bit more information (in that this clearly is the role that is delegated, not the one that is delegating)? I don't feel strongly about it though, just want to make sure we try to find the best names...

Yes, I agree `DelegatedRole` sounds better. I was never in love with the `DelegationRole` naming but wasn't sure what other options there are.

jku (approved):
Thanks!

MVrachev:
I rebased to fix one type-annotation I have forgotten.

MVrachev:
Rebased on top of the latest master branch.

PR #1372, opened 2021-05-04
API: Make _type read-only

### LINKED ISSUE #1371: API: Make _type read-only
The discussion in #1366 lead me to realize that 
* _type is not useful as Signed constructor argument (it's not an instance variable: it's a class constant)
* _type should not be modifiable via the API

If we fix these, there's no longer a need to validate _type (except in the Metadata from_dict() dispatcher).

I have a patch.

jku:
Also remove `_type` from Signed constructor arguments: the value is in a
class attribute. This way `_type` never needs to be validated (except
in the dispatcher in Metadata). There is a double-check in
`_common_fields_from_dict()` just to be sure.

This makes the API easier to use correctly as the public property is
immutable.

This is an API break as all Signed constructors change -- this could be
avoided but seems like the correct choice.

--- 

Fixes #1371

The discussion-worthy aspects of this PR are:
* it's a breaking change: It changes all Signed-derived constructors API (although calling Signed constructors isn't a common thing to do -- it usually happens in from_dict())
* it uses a property to make an attribute read-only: this is not done so far in the API but I believe we should do that more for attributes that should not be changed or that should be changed via some setter function. 

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -324,19 +324,23 @@ class Signed:
 
     """
 
+    # Signed implementations are expected to override this
+    _signed_type = None
+
+    @property
+    def _type(self):
+        return self._signed_type
+
     # NOTE: Signed is a stupid name, because this might not be signed yet, but
     # we keep it to match spec terminology (I often refer to this as "payload",
     # or "inner metadata")
     def __init__(
         self,
-        _type: str,
         version: int,
         spec_version: str,
         expires: datetime,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
This looks reasonable to me. I like removing `_type` from the constructor and making the property read-only.

Given that this is an in-flux API, I'm not worried about the API break. Let's get this right before we promise something stable.

@woodruffw is not using the constructors directly in the PyPI/warehouse integration, IIRC. (However I think he already has changes to make as that WIP PR is using `Metadata.from_json_file()`, which has been renamed `Metadata.from_file()`)

Issue #1018, opened 2020-04-08
Add stricter TARGETPATH and PATHPATTERN checks


lukpueh:
**Description of issue or feature request**:
#1008 removes ambiguity in interfaces that take [TARGETPATH](https://github.com/theupdateframework/specification/blame/69b8456614796b57c6076d3894568a83b3f7ce1a/tuf-spec.md#L808-L815) and/or [PATHPATTERN](https://github.com/theupdateframework/specification/blame/69b8456614796b57c6076d3894568a83b3f7ce1a/tuf-spec.md#L881-L891) arguments, by only accepting paths and path patterns that use "/" as directory separator and don't start with "/".

To make the interface even stronger, we should only accept paths and path patterns that completely adhere to the recommended format.

Note that, although the specification only *recommends* a format, we agreed that an implementation can be more assertive than the spec (see [spec#63](https://github.com/theupdateframework/specification/issues/63) and [spec#67](https://github.com/theupdateframework/specification/issues/67)).

**Current behavior**:
- Functions that take TARGETPATH and/or PATHPATTERN arguments only accept strings that use "/" as directory separator and don't start with "/".

**Expected behavior**:
- Functions that take TARGETPATH arguments only accept [path-relative-scheme-less-URL](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string) strings.

- Functions that take PATHPATTERN arguments only accept [path-relative-scheme-less-URL](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string) strings that may include shell-style glob characters. I suggest to align this with whatever the client updater supports.

jku:
I wonder if there's anything more to do here?

lukpueh:
I think the big question is how much input validation should be performed in the Metadata API, which still remains unanswered (#1130)

The checks requested in this ticket are currently not performed in the relevant interfaces:

TARGETPATH not checked at all
https://github.com/theupdateframework/python-tuf/blob/8ac7167f8d27f60955b8d984c65b5cdefa19d987/tuf/api/metadata.py#L1503

PATHPATTERN only checked to be strings
https://github.com/theupdateframework/python-tuf/blob/8ac7167f8d27f60955b8d984c65b5cdefa19d987/tuf/api/metadata.py#L1278-L1279

I lean towards implementing the requested checks, also to be consistent with places where we do check more strictly eg.

[...]

PR #1373, opened 2021-05-04
experimental client: Remove mirrors

jku:
(This is a continuation of and replaces #1368.)

**Drop "mirrors" support**

New Updater no longer supports downloading each metadata/target file from a list of user-provided mirrors. All downloads are performed from a single url: Default URL prefixes for both metadata and targets can be set in Updater(), and individual target downloads can override the default.

I believe this addresses the comments in #1368 in one way or another. One specific response:

>  `_build_full_url()` / `try:`, `download()`, `verify()` loop is repeated a few times. Might it generalise to a helper function?

I've removed a lot of cruft from these functions (including every manual file object close) but have not made a helper except for a more usable download function: I think there will be no need since the functions will be quite tight once all the verification moves to another component that handles the metadata consistency.

Another noteworthy thing is removing some url quoting/checking:
* We definitely have to go through the `fs path <-> url` conversions with a fine comb but this does not seem like the place: the "relative paths" in this case come from metadata and our code so there should be no need to quote here (as long as `get_one_valid_targetinfo()` argument is valid)?
* The current client has this weird interaction where `get_list_of_mirrors()` quotes parts of the url... and `_download_file()` then unquotes the whole url -- I'm not quite sure what the objective is
* Possibly we just want to document clearly that e.g. `get_one_valid_targetinfo()` argument must be a valid "path-relative-url" https://url.spec.whatwg.org/#path-relative-url-string (and then test what actually works): I don't want to handle filesystem and OS specific paths in the updater in the places where it can be avoided

If this is controversial I can leave this commit out


tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,19 @@ class Updater:
...
-        repository_mirrors: Dict,
+        metadata_base_url: str,
+        target_base_url: Optional[str] = None,
         fetcher: Optional[FetcherInterface] = None,
     ):
-
         self._repository_name = repository_name
-        self._mirrors = repository_mirrors
+        self._metadata_base_url = _ensure_trailing_slash(metadata_base_url)
+        self._target_base_url = _ensure_trailing_slash(target_base_url)
```
jku: well that's not going to work for `None`

jku:
> Note that this means we're now exposing the Updater caller to all kinds
> of new exceptions (as NoWorkingMirrorError is no longer an excuse we can
> use).

This should be called out as "requires future work"

tuf/client_rework/updater_rework.py
```diff
@@ -142,46 +143,38 @@ def updated_targets(targets: Dict, destination_directory: str) -> Dict:
 
         return updated_targets
 
-    def download_target(self, target: Dict, destination_directory: str):
+    def download_target(
```
joshuagl, line 158: Is this private API, and needs an underscore prefix? Or public API, and needs an args section in the docstring?

tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,26 @@ class Updater:
...
-        consistent_snapshot:
+    TODO
     """
 
     def __init__(
         self,
         repository_name: str,
-        repository_mirrors: Dict,
+        metadata_url: str,
+        default_target_url: Optional[str] = None,
```
joshuagl: I know this makes the variable names _really_ long, but should we better indicate that these are base urls? Or will we rely on the docstring for that?

tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,22 @@ class Updater:
...
-        consistent_snapshot:
+    TODO
     """
 
     def __init__(
         self,
         repository_name: str,
-        repository_mirrors: Dict,
+        metadata_base_url: str,
+        target_base_url: Optional[str] = None,
```
joshuagl, line 46: Instead of a separate base URL for metadata and targets, should we imitate the mirrors implementation we are removing which has a base URL + 'metadata' for metadata downloads and base URL + 'targets' for target file downloads? Something like:
```
# constructor parameters
repository_base_url: str,
default_metadata_dir: Optional[str] = None,
default_targets_dir: Optional[str] = None

# ensure sane defaults
if default_metadata_dir is None:
  default_metadata_dir = 'metadata'
self.metadata_base_url = parse.urljoin(repository_base_url, default_metadata_dir)
# etc
```
This gets slightly less intuitive when it comes to `download_target()`, where I think we want to provide a replacement for repository_base_url + default_targets_dir ? Still, I think having a base URL for the repository plus default paths joined to that base for targets and metadata makes sense. What do you think?

tuf/client_rework/updater_rework.py
```diff
@@ -83,13 +79,18 @@ def refresh(self) -> None:
...
-        Returns the target information for a specific file identified by its
-        file path.  This target method also downloads the metadata of updated
-        targets.
+        Returns the target information for a target identified by target_path.
+
+        As a side-effect this method downloads all the metadata it needs to
+        return the target information.
+
+        Args:
+            target_path: A path-relative-URL string
```
joshuagl: relative to?

tuf/client_rework/updater_rework.py
```diff
@@ -142,46 +143,38 @@ def updated_targets(targets: Dict, destination_directory: str) -> Dict:
 
         return updated_targets
 
-    def download_target(self, target: Dict, destination_directory: str):
+    def download_target(
+        self,
+        targetinfo: Dict,
+        destination_directory: str,
+        target_base_url: Optional[str] = None,
```
joshuagl, line 162: Following on from above, I think this makes sense as an argument to replace `repository_base_url` + `default_targets_dir` – i.e. in the case where targets files are on an entirely different host.

tuf/client_rework/updater_rework.py
joshuagl: `default_[metadata|targets]_path` may be a better name?

joshuagl (commented):
As expected, this simplifies the code a lot. Nice.
I have a question inline on the way we are configuring repositories, proposing we match the semantics of the outgoing mirrors configuration by having a base url to the repository with default relative paths to targets and metadata directories. This seems to work with `download_target()` being able to replace the entire base url + targets directory join with an optional `target_base_url` – curious to hear what you think?

We have some issues filed on the fs path <-> URL conversion already (https://github.com/theupdateframework/tuf/issues/1018 and https://github.com/theupdateframework/tuf/issues/1077). Let's make sure this work is part of the client refactor milestone please, either as a new issue or one (or both) of those issues being moved into the milestone.

There are quite a few TODO docstrings, can we file an issue to complete those and make it part of the appropriate milestone? (Is that 'Experimental client' or 'Client Refactor'?)

All that said, nice clean up. Good work @sechkova and @jku!

tuf/client_rework/updater_rework.py
jku: There's two cases to optimize for I guess: the reference "example" implementation and code that gets used in production.

For the first "example code" case: I'd argue this is easy to understand (and communicates the possibility of using separate hosts or whatever directories as a bonus):
```
updater = Updater(
    "/path/to/local/metadata/",
    "http://localhost:8080/metadata/", 
    "http://localhost:8080/targets/"
)
````
Just seeing that allows the new developer to open a browser and see what's actually in http://localhost:8080/metadata/, unlike when urls and directories are provided separately ...

note that the current PR still uses repository_name as first arg so the example above is from the future -- I'm demonstrating what the end result should be IMO: the only things the client decides are
 * local metadata directory
 * remote metadata base url
 * (optional) remote targets base url

that said the example code is definitely confusing right now: I pinky swear I'm going to improve that in coming PRs.

For the second case (production use), using a single URL and two paths makes one case more complex: if only one target host is used (so overriding the base url would not be needed in `download_target()`) but the host is not the same as metadata host. We don't have a lot of statistical data but this feels like it's not an edge case: We already know that hosting metadata of a regularly timestamping TUF instance sort of requires "actively running code on and maintaining a server" but hosting the target files is a matter of dumping them on a file server. I would expect a "two host" model to be a logical choice for production setups.

File: tuf/client_rework/updater_rework.py (hunk 5 above)
Comment (jku) [reply tuf/client_rework/updater_rework.py]: A fantastic question that I'd love to have an answer to... This unholy merger of `target identifier` and `relative local path` and `url fragment` is one of the sharp corners of TUF. For this function, target_path is just an identifier that's used to find the relevant pieces of metadata  ... but it is expected to somehow also transform nicely into a relative local filesystem path AND a part of a URL. Depending on the use case the user might be thinking in terms of URLs (like pip fetching a specific index.html) or local file paths (a system updater that wants to make sure the newest firmware file is on local disk).

In a way the url-fragment is not necessarily relative to anything as the hash-prefixing might mean $TARGETSBASEURL + $TARGETPATH isn't actually the url we want...  I'll try to think of a way to word this.

Comment (jku) [reply tuf/client_rework/updater_rework.py]: A realization now that I wrote it down: Is the idea that target_paths are directly/uniquely transferable to paths on local filesystem actually from the spec at all, or is it just the way the reference client was written? Could we just as well have decided to download `http://localhost:8080/targets/a/b/c.txt` into just `c.txt` (without path prefix) or asked the user to provide the local filename?

If this is purely a client decision, is it good one?

Comment (jku) [reply tuf/client_rework/updater_rework.py]: > client MUST write the file to non-volatile storage as FILENAME.EXT

Looks like path handling is not in spec, filename is (even that might be just copy-paste from metadata handling rules where it makes sense)

File: tuf/client_rework/updater_rework.py (hunk 4 above)
Comment (joshuagl) [reply tuf/client_rework/updater_rework.py:46]: > We already know that hosting metadata of a regularly timestamping TUF instance sort of requires "actively running code on and maintaining a server" but hosting the target files is a matter of dumping them on a file server. I would expect a "two host" model to be a logical choice for production setups.

This is a well reasoned argument, I'm convinced. Thanks for taking the time to explore it with me.

**Comment by joshuagl** [comment]
There's some issues to file (and move into appropriate milestones), then this should be good to merge. We can address the minor items (docstrings, variable names, public vs. private API) here or file issues to tackle them before experimental-client is merged with develop.

**Comment by jku** [comment]
I've marked issues resolved if I think they are now resolved in this PR. 
I've filed #1385 for documenting the module, and updated #1312 for exceptions
All of the other TODOs in updater_rework.py will be handled by #1355.

(The CI failure is sslib master build since we're a bit behind current develop)

File: tuf/client_rework/updater_rework.py (hunk 5 above)
Comment (jku) [reply tuf/client_rework/updater_rework.py]: I've tried improving the docstring a bit... not sure how to proceed other than making a note in #1385

Comment (joshuagl) [reply tuf/client_rework/updater_rework.py]: thanks for tackling this, cleaned up doctoring and adding a note to #1385 feels sufficient.

File: tests/test_updater_rework.py [hunk 7]
```diff
@@ -123,14 +123,13 @@ def setUp(self):
     # directory copied from the original repository files.
     tuf.settings.repositories_directory = self.client_directory
 
-    self.repository_mirrors = {'mirror1': {'url_prefix': url_prefix,
-                                           'metadata_path': 'metadata',
-                                           'targets_path': 'targets'}}
-
+    metadata_url = os.path.join(url_prefix, 'metadata/')
+    targets_url = os.path.join(url_prefix, 'targets/')
```
joshuagl: Nit: should this be `urllib.parse.urljoin()`?

tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,32 @@ class Updater:
...
-        metadata:
-
-        repository_name:
-
-        mirrors:
-
-        fetcher:
-
-        consistent_snapshot:
+    TODO
```
joshuagl, line 39: Is this covered by #1385?

tuf/client_rework/updater_rework.py
```diff
@@ -35,27 +36,32 @@ class Updater:
...
-        repository_mirrors: Dict,
+        metadata_base_url: str,
+        target_base_url: Optional[str] = None,
         fetcher: Optional[FetcherInterface] = None,
     ):
-
+        """
+        Args:
+            repository_name: directory name (within a local directory
+                defined by 'tuf.settings.repositories_directory')
```
joshuagl, line 52: This took me a couple of attempts to parse, I'm not sure my suggestion is any better though.
```suggestion
            repository_name: name of the directory used for storing local copies of working files
                for this repository. Created as a child of 'tuf.settings.repositories_directory'.
```

joshuagl (approved):
LGTM, with a few minor comments/questions/suggestions in-line.

I'd like to see some more testing around this, the `seek(0)` calls especially feel like they require some reassurance that they are matching user expectations.

tuf/client_rework/updater_rework.py
jku: This is better but I'm not changing it as I intend to
* not use tuf.settings for anything at all in near future
* use local_repo_dir as the argument directly -- there's no need for a system wide repositories_directory (at least not one managed by Updater)

this will happen after we have a MetadataBundle #1355 we can use and should be much easier to understand and document

tuf/client_rework/updater_rework.py
jku: Yeah that was the idea: The public API will change a bit as we start using the MetadataBundle so let's document after that

tests/test_updater_rework.py
jku: yeah, I'll change that 

Fact is that the test needs to be rewritten, it's copied from test_updater.py and is horribly complex and full of issues like this for (I assume) Historical Reasons... but I'm trying to limit scope so not fixing anything else here

joshuagl:
Thanks for responding to feedback, LGTM for a merge to experimental-client.

PR #1376, opened 2021-05-07
Metadata API: Provide type as well as _type

### LINKED ISSUE #1375: Metadata API: linting _type is an issue for API users
Any code that uses the Metadata API is going to get linting results like this: 
```
file.py:299:11: W0212: Access to a protected member _type of a client class (protected-access)
file.py:301:41: W0212: Access to a protected member _type of a client class (protected-access)
file.py:336:11: W0212: Access to a protected member _type of a client class (protected-access)
file.py:338:46: W0212: Access to a protected member _type of a client class (protected-access)
file.py:396:11: W0212: Access to a protected member _type of a client class (protected-access)
file.py:398:45: W0212: Access to a protected member _type of a client class (protected-access)
```

`Signed._type` is named like it is to copy the file format 1-to-1 which makes some sense...  but this is also pretty unacceptable. Maybe we should make another property `type` that works exactly like `_type` ?

jku:
Using Metadata APIs '_type' attribute (from outside metadata.py) currently
results in linter errors.

Add a duplicate 'type': this way the API users can avoid linter errors
but '_type' is still available in case the strict file format
compatibility is needed.

Fixes #1375

---

I think the only alternative to this that I would consider acceptable is just replacing `_type` with `type`.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -315,20 +315,24 @@ class Signed:
     are common for all TUF metadata types (roles).
 
     Attributes:
-        _type: The metadata type string.
+        _type: The metadata type string. Also available without underscore.
         version: The metadata version number.
         spec_version: The TUF specification version number (semver) the
             metadata format adheres to.
         expires: The metadata expiration datetime object.
         unrecognized_fields: Dictionary of all unrecognized fields.
-
     """
 
     # Signed implementations are expected to override this
     _signed_type = None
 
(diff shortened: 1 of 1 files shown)


MVrachev (approved):
Given that `_type` is read-only that's okay.
LGTM!

joshuagl (approved):
This seems reasonable as a convenience API to avoid having to disable classes of linter warnings.

PR #1379, opened 2021-05-07
Metadata minor fixes

jku:
Some minor improvements to metadata.py:
 * type hint fixes
 * Root.remove_key(): avoid unneeded lookups, also raise if key is not found

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -589,16 +589,18 @@ class Root(Signed):
         self.roles[role].keyids.add(keyid)
         self.keys[keyid] = key_metadata
 
-    # Remove key for a role.
     def remove_key(self, role: str, keyid: str) -> None:
-        """Removes key for 'role' and updates the key store."""
-        if keyid in self.roles[role].keyids:
-            self.roles[role].keyids.remove(keyid)
-            for keyinfo in self.roles.values():
-                if keyid in keyinfo.keyids:
-                    return
+        """Removes key from 'role' and updates the key store.
 
-            del self.keys[keyid]
+        Raises:
+            KeyError: If 'role' does not include the key
+        """
+        self.roles[role].keyids.remove(keyid)
+        for keyinfo in self.roles.values():
+            if keyid in keyinfo.keyids:
+                return
+
+        del self.keys[keyid]
 
 
 class Timestamp(Signed):
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -17,7 +17,7 @@ available in the class model.
 """
 import tempfile
 from datetime import datetime, timedelta
-from typing import Any, Dict, Mapping, Optional
+from typing import Any, Dict, List, Mapping, Optional
 
 from securesystemslib.keys import verify_signature
 from securesystemslib.signer import Signature, Signer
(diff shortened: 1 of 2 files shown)


jku:
Oops a mistake push here, don't pay any attention yet (this includes the next PR as well)

EDIT: all good now, ready for review

tuf/api/metadata.py
```diff
@@ -42,25 +42,17 @@ class Metadata:
...
-                },
-                ...
-            ]
+        signatures: A list of Securesystemslib Signature objects, each signing
+            the canonical serialized representation of 'signed'.
 
     """
 
-    def __init__(self, signed: "Signed", signatures: list) -> None:
+    def __init__(self, signed: "Signed", signatures: List[Signature]) -> None:
```
MVrachev, line 50: I agree we should start using `List` instead of `list` in type annotations mostly because we can annotate what's on the list too. I have been thinking to make that change for a while!

MVrachev (approved):
Overall LGTM, but you will have to resolve the conflicts.

jku:
rebased on current develop

tuf/api/metadata.py
```diff
@@ -429,8 +421,8 @@ def __init__(
         self,
         keytype: str,
         scheme: str,
-        keyval: Mapping[str, str],
-        unrecognized_fields: Optional[Mapping[str, Any]] = None,
+        keyval: Dict[str, str],
+        unrecognized_fields: Optional[Dict[str, Any]] = None,
```
MVrachev: In your commit message you said
`Stop using Mapping where we actually mean Dict:
  Mapping means "we only need a read-only dict" and most of the
  time this is not really the case.
`
which hints the read-only dictionaries in our classes should be `Mapping`?
Because `unrecognized_fields` is read-only for each of the classes.

jku: this is probably true... I'll have to try it

jku: Yeah Mapping works for unrecognised_fields... it's not very ergonomical (as `unrecognised_fields or {}` needs extra type hints or the types are useless) but maybe it still makes sense

jku:
Latest change: Left `unrecognised_fields` as a Mapping as suggested. We don't intend to modify unrecognised fields so let's make that clear in the type hints.

tuf/api/metadata.py
```diff
@@ -348,10 +340,10 @@ def __init__(
         if version <= 0:
             raise ValueError(f"version must be > 0, got {version}")
         self.version = version
-        self.unrecognized_fields = unrecognized_fields or {}
+        self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
```
joshuagl, line 343: Interesting. I think this is the first time we've added an inline type annotation? Is this the only class property which is not set by an annotated constructor argument?

joshuagl (approved):
LGTM, thank you.

tuf/api/metadata.py
jku: The issue here is that `{}` has type "dict" so then mypy figures that our instance attribute type must be `Mapping[str, Any] | dict[str, Any]` which... does not break anything at all but also makes the whole idea of using Mapping a bit useless

PR #1382, opened 2021-05-10
tests: API: Test unrecognized fields everywhere

jku:
Currently we test adding unrecognised fields in specific places in our
file format structure. Instead add the fields everywhere (except
specific locations where that would not be valid).

--- 

The idea here is to test inserting custom items into every dictionary (except where spec doesn't actually allow it), instead of testing only specific dictionaries.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -17,6 +17,7 @@ import copy
 
 from datetime import datetime, timedelta
 from dateutil.relativedelta import relativedelta
+from typing import Any, Dict
 
 from tests import utils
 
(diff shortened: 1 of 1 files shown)


jku:
rebased on develop

tests/test_api.py
```diff
@@ -526,52 +527,40 @@ def test_metadata_targets(self):
...
+
+            # Process all child dicts. spec does not contain any
+            # dicts inside arrays so no need to check those
+            for valname, value in container.items():
+                if isinstance(value, dict):
+                    named_dicts.append((valname, value))
+
+            # if dicts keys are limited in spec or if dicts values are
+            # strictly defined, skip it
+            if name in ["hashes", "keys", "meta", "roles", "targets"]:
```
sechkova, line 545: Not really a review but how did we come up with this list? I guess you did it by going through the spec. Can we note it somewhere so that we don't do it again?

jku: Yes you guessed correctly. I'll try to phrase that somehow clearly

MVrachev: Wondering why do you include `meta` here?

MVrachev (commented):
Overall, it checks what I expected from such a test.
I had a look into the spec if you missed a dictionary where we don't want to have `unrecognized_fields` and it seems there are no other.

Only one unclear thing for me which I asked in a comment.

tests/test_api.py
jku: because it seems to me the meta content is strictly defined: keys must be metadata filenames (in the timestamp case there is basically only one allowed value). values must be MetaFile dict representations. I don't think allowing something else is reasonable.

MVrachev:
I will suggest combing this pr with changes in ADR 8 clarifying that some of the fields won't support unrecognized fields.
We want to have that documented to minimize the chance of including unrecognized fields where it's sensitive.

jku:
The alternative to this PR would be to include a test case "unrecognised fields" in every test dataset in #1416 ... That would mean none of this code is needed (good) but does mean we'd have to be diligent about really including the cases everywhere (bad)

We could leave this as draft until we see if we get #1416 implemented in a nice way instead?

MVrachev:
Yes, I suppose let's agree on 1416 and then decide about this one.

jku:
made this a draft until we see if serialization tests can/should cover this without code

MVrachev:
The Pr using the table testing for tests connected with unrecognized fields out in the wild: #1466.

PR #1386, opened 2021-05-12
Fix Root.add_key() argument's type

### LINKED ISSUE #1378: Metadata API: Root.add_key() is broken
Recent Key object addition broke `Root.add_key()`: it now adds a dictionary into the internal dict but it should add a Key. This can be tested by adding a `metadata.to_dict()` after calling `Root.add_key()` in the tests.

I didn't fix it since I don't know how this is used -- should the function accept a Key as arg?

sechkova:
Fixes #1378

Since the existence of a `Key` class, taking `Key` as an argument type seems like the logical change.
The test case is updated accordingly.

Since there are no "users" of this functionality yet, I can't say if further improvements are
needed for now. 

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -48,10 +48,6 @@ from securesystemslib.interface import (
     import_ed25519_privatekey_from_file
 )
 
-from securesystemslib.keys import (
-    format_keyval_to_metadata
-)
-
 from securesystemslib.signer import (
     SSlibSigner
 )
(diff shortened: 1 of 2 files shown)


sechkova:
Force-pushed after I changed my mind over the test case a bit

jku (commented):
looks ok to me but could you add a regression test so root metadata with an added key serializes correctly (a test that fails with current develop and succeeds with your fix)?

sechkova:
> looks ok to me but could you add a regression test so root metadata with an added key serializes correctly (a test that fails with current develop and succeeds with your fix)?

I was wondering whether to add such tests case because the bug was in our test using incorrectly the class ... and in the type annotation but types in Python ... 
Anyway, since this comes up as a request, I guess adding it was the better decision. I'll update the PR.

jku:
> I was wondering whether to add such tests case because the bug was in our test using incorrectly the class

yeah you do have a point here -- thanks anyway.

jku (approved):

PR #1390, opened 2021-05-14
Avoid using falsy tests for None

jku:
Fix a bug in DelegatedRole serialization and change all "falsy" object comparisons to something else.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -497,6 +497,21 @@ class TestMetadata(unittest.TestCase):
         delegations = Delegations.from_dict(copy.deepcopy(delegations_dict))
         self.assertEqual(delegations_dict, delegations.to_dict())
 
+        # empty keys and roles
+        delegations_dict = {"keys":{}, "roles":[]}
+        delegations = Delegations.from_dict(delegations_dict.copy())
+        self.assertEqual(delegations_dict, delegations.to_dict())
+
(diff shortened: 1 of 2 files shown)


jku:
The fact that these issues are found tells me our testing is not adequate. I'll file an issue on that, if one does not exist: #1391

tests/test_api.py
```diff
@@ -495,6 +497,14 @@ def test_delegation_class(self):
...
         self.assertEqual(delegations_dict, delegations.to_dict())
 
+        # Test some basic missing data
+        invalid_delegations_dicts = [
+            {},
+            {"keys":{}, "roles":[]},
+        ]
+        for d in invalid_delegations_dicts:
+            with self.assertRaises(KeyError):
+                Targets.from_dict(d)
```
sechkova: I'm confused about your intentions here. 
Either` Delegations.from_dict(d)` instead of ` Targets.from_dict(d)` or the invalid input dictionaries should better represent a `Targets` one.

sechkova (commented):
One comment about a test case, otherwise LGTM.

tests/test_api.py
jku: Oops, I created the test code in the targets test case and just moved it...

sechkova (approved):

joshuagl (approved):
Nice fixes, thanks!

PR #1394, opened 2021-05-14
Metadata API: Make consistent_snapshot optional

### LINKED ISSUE #1374: Metadata API: Root.consistent_snapshot should be optional
Spec:
> Finally, the root metadata should write the Boolean "consistent_snapshot" attribute at the root level of its keys of attributes. If consistent snapshots are not written by the repository, then the attribute may either be left unspecified or be set to the False value. Otherwise, it must be set to the True value.

Metadata API (tuf/api/metadata.py) seems like it would fail to read json that does not contain "consistent_snapshot"

MVrachev:
Fixes #1374

From chapter 7 in the spec (version 1.0.17)
```
Finally, the root metadata should write the Boolean
"consistent_snapshot" attribute at the root level of its keys of
attributes.
If consistent snapshots are not written by the repository,
then the attribute may either be left unspecified or be set to the
False value. Otherwise, it must be set to the True value.
```

We want to make sure we support repositories
without consistent_snapshot set.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -425,6 +425,12 @@ class TestMetadata(unittest.TestCase):
         with self.assertRaises(KeyError):
             root.signed.remove_key('root', 'nosuchkey')
 
+        # Test serializing and deserializing without consistent_snapshot.
+        root_dict = root.to_dict()
+        del root_dict["signed"]["consistent_snapshot"]
+        root = Root.from_dict(copy.deepcopy(root_dict["signed"]))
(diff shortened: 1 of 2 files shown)


jku (approved):
Re-arranging constructor arguments changes API, but I think you made the right choice.

LGTM, I'm approving but you could mention the API break in the commit message.

MVrachev:
It's not a matter of choice where the `consistent_snapshot` should be.
It's not allowed to have arguments without default values after arguments with:
image

Will update clarify in the commit that this is a breaking change.

Issue #1384, opened 2021-05-12
Metadata API: stop using formats.py


jku:
Currently we have one use of tuf/formats.py in tuf/api/metadata.py. I think we should try to keep the two implementations separate.

**Current behavior**:
`formats.expiry_string_to_datetime()` is used in metadata.py

**Expected behavior**:
We should do the conversion in metadata.py (and I think we should not raise securesystemslib errors like current implementation does).

avelichka:
I'd like to work on that :)

PR #1395, opened 2021-05-17
Build: Initial mypy integration

### LINKED ISSUE #1335: lint/CI: Use mypy for static type checking
We should start running mypy in CI (probably in tox "lint" environment). This would enforce type annotation use and would check types at compile time.

Possibly we only want to do this to "new code" like `api/` but it might be worth it to add annotations to everything (I'm not sure how much work this is).

**Current behavior**:
* type annotation use is not enforced
* types are not checked at compile time

**Expected behavior**:
* new code without type annotations fails CI
* types are statically checked: type check failure leads to CI failure

jku:
Goal is to start using mypy to statically check our source code:
* Add mypy configuration: start by just checking `tuf/api`, and ignoring securesystemslib imports
* Use mypy in tox "lint" environment.  
* Refactor tox lint env a little
* Use all cores for pylint to make it faster (pylint is the slowest part of the lint by far)
* Fix all kinds of typing issues in tuf/api/ (either actual issues or just things that prevent mypy from doing its job)

How mypy is used after this PR:
* CI will run it as part of lint
* developers can run `tox -e lint` or just `mypy`

Fixes #1335

### DIFF
--- a/tuf/api/serialization/__init__.py
+++ b/tuf/api/serialization/__init__.py
@@ -15,6 +15,11 @@ base classes defined in this __init__.py module.
 
 """
 import abc
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # pylint: disable=cyclic-import
+    from tuf.api.metadata import Metadata, Signed
 
 
 # TODO: Should these be in tuf.exceptions or inherit from tuf.exceptions.Error?
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -11,4 +11,5 @@ coverage
 black
 isort
 pylint
+mypy
 bandit
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,3 +6,10 @@ ignore =
   requirements-dev.txt
   .travis.yml
   .coveragerc
+
+[mypy]
+warn_unused_configs = True
+files = tuf/api/
+
+[mypy-securesystemslib.*]
+ignore_missing_imports = True
--- a/tox.ini
+++ b/tox.ini
@@ -39,15 +39,18 @@ commands =
     python -m coverage report -m
 
 [testenv:lint]
+changedir = {toxinidir}
 commands =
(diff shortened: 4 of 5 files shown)


jku:
rebased on develop

jku:
rebased on develop, added fix for an issue in new code

tuf/api/serialization/__init__.py
```diff
@@ -15,6 +15,11 @@
 
 """
 import abc
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # pylint: disable=cyclic-import
+    from tuf.api.metadata import Metadata, Signed
```
sechkova, line 22: So mypy recognizes that the quoted strings "Metadata" and "Signed" are the classes in this import?
Cool trick!

sechkova (approved):
Do you think stricter options should be tried next? For example "disallow-untyped-defs" on api/metadata.py.
I see that now this results to errors coming from `formats.py `and `exceptions.py`.

Or maybe this is too strict and we can add type annotations to the coding guidelines for now?

tuf/api/serialization/__init__.py
jku: Yes, AFAIK for mypy the type-names-as-string are just as valid.

jku:
> Do you think stricter options should be tried next? For example "disallow-untyped-defs" on api/metadata.py.
> I see that now this results to errors coming from `formats.py `and `exceptions.py`.

We probably should start experimenting with stricter options: once it's all set up I imagine it's not hard to maintain... But probably the most valuable improvement would be to gradually start supporting securesystemslib modules.

On the specific cases you mentioned: formats.py should go IMO: #1384. Fixing exceptions.py should be a few minutes job.

joshuagl (approved):
LGTM, thanks. Might take a while getting used to inline type annotations, but should be worth it 🤞 

Will be great to include ngclient in the lint test Tox environment, once it has merged.

Issue #1312, opened 2021-03-12
review/document client exceptions


sechkova:
**Description of issue or feature request**:
Revise and implements client exceptions handling

**Current behavior**:
Not all thrown exceptions are documented or handled by the client.

**Expected behavior**:
Implement proper exception handling and document the API

Sync with #1131

jku:
The major items here are probably:
* bundle (once it's in) should only raise RepositoryError (or anyway a single base error representing failures because remote provided metadata we can't cope with) and user errors like ValueError
* since we removed mirrors, we can longer hide behind NoWorkingMirror -- need to figure out a nice solution. This probably still involves at least one wrapper error as we may get random errors from whatever Fetcher is being used

jku:
After doing the metadata bundle integration, I think our client exceptions should be grouped like this:
* RepositoryErrors (various): something is wrong with metadata. In some cases this can be handled within client (like in the case of local metadata being no longer signed by correct keys), but in many cases this is an external exception
* OSErrors (various): Failures to read or write local metadata
* RuntimeErrors/ValueErrors/etc: errors that mean API misuse (code should be fixed, exception should not be handled)
* some kind of generic download error: We should define an error that download functions return, wrapping whatever error the FetcherInterface implementation raises. This way the updater public API has a specific error for download failures

jku:
My impression is that ngclient exceptions are in a pretty good state by now, but review would  be good:
* see what exceptions are handled in code using ngclient (like tests)
* review ngclient/trustedmetadataset and see if exceptions seem to make sense

In addition to a review, we should document how ngclient uses exceptions: mostly that RepositoryError is used as a catch all for  fatal update errors -- a metadata state that clients cannot recover from, but where a later update (after the remote updates metadata) might succeed.

jku:
Updated title: I think we're mostly missing proper documentation:
* document our own exceptions
* document OSErrors where necessary (anytime we read or write files)

PR #1396, opened 2021-05-18
Experimental client: use bundle

### LINKED ISSUE #1304: Remove `MetadataWrapper` class
I find it a bit paradox that `MetadataWrapper` is another container on top of `Metadata`, adding another level of nesting, when most of the methods/properties are actually shortcuts to attributes contained somewhere in `Metadata`. If we want a client-variant of `Metadata`, I'd maybe use inheritance instead of composition. 🤷 

But maybe we don't even need it. Let's quickly go through the individual methods/properties of here:

On `MetadataWrapper`:

- `from_json_file` -> available as `Metadata.from_file` with json as default (#1279)
- `from_json_object` -> could be a format-agnostic method on `Metadata` just like `from_file`. #1336 
- `persist` ->  available as `Metadata.to_file` with json as default (#1279)
- `expires` -> could be a method on `Signed`, I'd call it 'is_expired' #1305 
- `verify` -> This is definitely needed for the repository tool as well. So it should be somewhere where both client and repository tool can access it. Maybe even on `Metadata` in addition to the vanilla  one-signature `verify` we currently have. See https://github.com/theupdateframework/tuf/pull/1060#issuecomment-660056376 for thoughts about that distinction. #1306

All the other methods here are either pure shortcuts or filters of attributes contained in Metadata objects (or contained objects).

- Re **shortcuts**: https://github.com/theupdateframework/tuf/pull/1060/files#r452906629 has arguments against shortcuts.
- Re **filters**: Maybe it will become easier to access individual properties once we have classes for all complex attributes as planned in https://github.com/theupdateframework/tuf/issues/1139? But similarly to the arguments against shortcuts, I'd be careful when implementing behavior that relates to contained classes on container classes.

_Originally posted by @lukpueh in https://github.com/theupdateframework/tuf/pull/1291#discussion_r590391070_

### LINKED ISSUE #1319: client refactor: review _get_*_meta_name() usage
Simplify and remove duplication from metadata filename handling (_get_relative_meta_name(),_get_full_meta_name())

* remove duplication
* probably no need to make the extension modifiable: .json is fine
* there are cases of _get_full_meta_name('timestamp.json') -- this seems like a bug?

### LINKED ISSUE #1320: Remove temporary commits

Remove commits that:

- temporary reduce code coverage percentage: 5c58c1e
- temporary disable linter checks: d252818, 062bda4

jku:
Use the MetadataBundle in the new Updater:
* Remove MetadataWrapper
* Update and expand tests
* Stop using tuf.settings, store the constants in updater_rework.py
* Use MetadataBundle to verify current set of metadata, remove all verification code from Updater
* load local metadata from disk
* persist new verified metadata to disk

Fixes #1304 
Fixes #1320 
Fixes #1319


jku:
The constants could use some thought: We definitely don't want to keep using tuf.settings, but I guess the question is, should these be instance variables in Updater instead of module level ones:

```
MAX_ROOT_ROTATIONS = 32
MAX_DELEGATIONS = 32
DEFAULT_ROOT_MAX_LENGTH = 512000  # bytes
DEFAULT_TIMESTAMP_MAX_LENGTH = 16384  # bytes
DEFAULT_SNAPSHOT_MAX_LENGTH = 2000000  # bytes
DEFAULT_TARGETS_MAX_LENGTH = 5000000  # bytes
```
Maybe it does not hurt if they were instance variables instead. They don't even need to be optional constructor arguments (as initialization does not use any of them: caller can initialize the Updater and then modify the values).
I guess the fanciest option is that constructor takes an optional argument dataclass UpdaterOptions with those attributes ...

tuf/client_rework/updater_rework.py
```diff
@@ -516,7 +331,7 @@ def _preorder_depth_first_walk(self, target_filepath) -> Dict:
         target = None
         role_names = [("targets", "root")]
         visited_role_names = set()
-        number_of_delegations = settings.MAX_NUMBER_OF_DELEGATIONS
+        number_of_delegations = MAX_DELEGATIONS
 
         # Ensure the client has the most up-to-date version of 'targets.json'.
         # Raise 'exceptions.NoWorkingMirrorError' if the changed metadata
```
jku: This is not actually the case: NoWorkingMirrors don't happen anymore, there's TODOs about download errors

tests/test_updater_rework.py
```diff
@@ -119,50 +114,25 @@ def setUp(self):
...
-    # Setting 'tuf.settings.repository_directory' with the temporary client
-    # directory copied from the original repository files.
-    tuf.settings.repositories_directory = self.client_directory
-
     metadata_url = f"{url_prefix}/metadata/"
     targets_url = f"{url_prefix}/targets/"
     # Creating a repository instance.  The test cases will use this client
     # updater to refresh metadata, fetch target files, etc.
-    self.repository_updater = updater.Updater(self.repository_name,
+    self.repository_updater = updater.Updater(self.client_directory,
```
sechkova, line 119: `self.repository_name` is not used anymore.

tests/test_updater_rework.py
```diff
@@ -187,60 +157,16 @@ def test_refresh(self):
 
     self.assertListEqual(updated_targets, [])
 
+  def test_refresh_with_only_local_root(self):
```
sechkova, line 158: But of course ... files are already copied in the directory. Can we mention this in `test_refresh`, that it actually first loads all metadata files from the file system.

tuf/client_rework/updater_rework.py
```diff
@@ -542,14 +357,13 @@ def _preorder_depth_first_walk(self, target_filepath) -> Dict:
 
             # The metadata for 'role_name' must be downloaded/updated before
             # its targets, delegations, and child roles can be inspected.
-            # self._metadata['current'][role_name] is currently missing.
             # _refresh_targets_metadata() does not refresh 'targets.json', it
```
sechkova: The rest of the comment is also out-of-date as it references functions not existing in this client version.

tuf/client_rework/updater_rework.py
```diff
@@ -610,10 +424,8 @@ def _preorder_depth_first_walk(self, target_filepath) -> Dict:
...
         ):
             msg = (
-                f"{len(role_names)}  roles left to visit, ",
-                "but allowed to visit at most ",
-                f"{settings.MAX_NUMBER_OF_DELEGATIONS}",
-                " delegations.",
+                f"{len(role_names)} roles left to visit, but allowed to ",
+                f"visit at most {MAX_DELEGATIONS} delegations.",
             )
             logger.debug(msg)
```
sechkova, line 417: This trick of mine with logging needs to be reverted now that we agreed on %s-strings in the logging calls. 
But not in this PR. I'll note it for myself.

tuf/client_rework/updater_rework.py
```diff
@@ -195,318 +215,113 @@ def download_target(
...
-
-        intermediate_targets.verify(
-            parent_role.keys(filename), parent_role.threshold(filename)
-        )
-
-        intermediate_targets.expires()
-
-        return intermediate_targets
+            data = self._load_local_metadata(role)
+            self._bundle.update_delegated_targets(data, role, parent_role)
```
sechkova, line 310: Using `self._bundle.update_delegated_targets` here leaves `MetadataBundle.update_targets()` unused.
Not a big deal but mentioning it.

MVrachev:
After I made `consistent_snapshot` optional for root [here](#1394), I realized that the `MetadataBundle` was created before that.
Can we check that if `consistent_snapshot` is not provided everything will work with the update process?

sechkova (commented):
> The constants could use some thought: We definitely don't want to keep using tuf.settings, but I guess the question is, should these be instance variables in Updater instead of module level ones:
> 
> ```
> MAX_ROOT_ROTATIONS = 32
> MAX_DELEGATIONS = 32
> DEFAULT_ROOT_MAX_LENGTH = 512000  # bytes
> DEFAULT_TIMESTAMP_MAX_LENGTH = 16384  # bytes
> DEFAULT_SNAPSHOT_MAX_LENGTH = 2000000  # bytes
> DEFAULT_TARGETS_MAX_LENGTH = 5000000  # bytes
> ```
> 
> Maybe it does not hurt if they were instance variables instead. They don't even need to be optional constructor arguments (as initialization does not use any of them: caller can initialize the Updater and then modify the values).
> I guess the fanciest option is that constructor takes an optional argument dataclass UpdaterOptions with those attributes ...

Given the proposed structure in https://github.com/theupdateframework/tuf/issues/1397#issuecomment-843103638  where the idea is to expose only the `Updater` class in the client `__init__` , probably the constants need to be inside the class (or the fancier options).  However this can happen when solving #1397. 

I left some minor comments but I in general I like how Updater looks now.

jku:
Addressed the resolved discussions. Also checked that consistent_snapshots=None seems to work fine (using a falsy test)

tuf/client_rework/updater_rework.py
sechkova: Captured in #1400

sechkova (approved):

tuf/client_rework/updater_rework.py
```diff
@@ -195,318 +196,111 @@ def download_target(
...
+
+            metainfo = self._bundle.timestamp.signed.meta["snapshot.json"]
+            length = metainfo.get("length") or DEFAULT_SNAPSHOT_MAX_LENGTH
+            version = None
+            if self._bundle.root.signed.consistent_snapshot:
+                version = metainfo["version"]
+
+            data = self._download_metadata("snapshot", length, version)
+            self._bundle.update_snapshot(data)
+            self._persist_metadata("snapshot", data)
```
joshuagl, line 304: I was confused at first by the downloading of new metadata all happening in the `except` block. It took me a second read to realise that the `_bundle.update_snapshot()` call fails if the snapshot version is out of date with a `RepositoryError`, and then the logic beneath the `except` clause is executed.

joshuagl, line 304: Here, and other `_load_*()` functions, the exceptions raised by `MetadataBundle.update_snapshot` and the `open` and `write` calls in `_persist_metadata` become errors the user must handle. The former seems reasonable enough, I'm less certain about the latter. Especially when  `get_one_valid_targetinfo()` -> `_preorder_depth_first_walk()` could result in a myriad of errors (download, file open/write, ???).

Probably a topic for https://github.com/theupdateframework/tuf/issues/1312 ?

joshuagl (approved):
This looks very concise, nice. It took me a couple of reads to figure out why the fetching of metadata only happened in the exception handler of a load. That might want some attention in terms of comments and/or function names? 🤔 

I also noticed there are some unhandled exceptions which I am not sure whether it is reasonable to leave those to a caller to handle or not. Would welcome thoughts on that.

Overall, very solid. LGTM.

tuf/client_rework/updater_rework.py
jku: yeah, basically the only case where we don't want to download snapshot is if we have a local snapshot file AND it is valid -- but with exceptions programming this is what it looks like...

Better comments wouldn't hurt

jku: yes I vote for handling in 1312.  I think the three classes of exceptions (repository / download / file IO) are logical and also the granularity that I could see a serious client wanting to handle (or not handle as might be reasonable with e.g. file IO ).

PR #1401, opened 2021-05-20
Experimental client: renaming

### LINKED ISSUE #1397: Client refactor: directories and naming
Move the packages under the client directory in a way that makes it easy to distinguish repository components from internal client components and client public API.

Proper naming is needed for the client refactor directory which is planned to land in the main branch in parallel with the current client.

sechkova:
Fixes #1397

~Note: this is done on top of changes introduced by #1396~

Proposes the following client naming and directory structure:

```
tuf/
    ngclient/
        __init__.py
        updater.py
        metadatabundle.py
        network/
            __init__.py
            requestsfetcher.py
            download.py
```
Most significant changes:
- `client_rework` -> `ngclient` (next-gen-client): almost what @jku suggested but trying to avoid dashes (not accepted) and underscores (not recommended) in package names.
- client public classes are exposed in `ngclient/__init__.py`
- download related modules are moved under `network`. `FetcherInterface` is defined in `network/__init__.py` instead of `fetcher.py`
- new client code is excluded from coverage until better tested

requirements**:


jku:
Note that the names here are not package names: they're just module names where underscores can be used if it improves readability... but I'm fine with ngclient, no better ideas at the moment

jku:
The reason I suggested `ngclient._internal` is that the name makes it obvious to an outsider that they should not rely on things inside this module -- we want to be able to change them when we want. `ngclient.network` does not convey that and I don't think we can assume outsiders understand that `ngclient` (the __init__.py) defines what our API is and that tings below it are off-limits (as much as I otherwise like the concept).

jku:
> Note that the names here are not package names: they're just module names where underscores can be used if it improves readability... but I'm fine with ngclient, no better ideas at the moment

I'll just correct myself here: they do seem to call any directory with `__init__.py` a package. 

I still maintain `_internal` is a fine name for a directory: this what a lot of big name python projects do (see pip for an example).

sechkova:
Agreed with @jku's review to move non-public modules to _internal.
The new structure is:
```
tuf/
    ngclient/
        __init__.py
        updater.py
        fetcher.py
        _internal/
            requests_fetcher.py
            download.py
            metadata_bundle.py
```
Also rebased on latest changes in experimental-client branch.

jku (approved):
Thanks!

There's obviously many valid choices here but I'm happy with ones made here

jku:
For some reason github claims there are conflicts here after merging 1403 -- manual rebase works without any issues though?

sechkova:
Yep, rebased.

Issue #1402, opened 2021-05-20
ngclient: client-level constants


sechkova:
The constants could use some thought: We definitely don't want to keep using tuf.settings, but I guess the question is, should these be instance variables in Updater instead of module level ones:

```
MAX_ROOT_ROTATIONS = 32
MAX_DELEGATIONS = 32
DEFAULT_ROOT_MAX_LENGTH = 512000  # bytes
DEFAULT_TIMESTAMP_MAX_LENGTH = 16384  # bytes
DEFAULT_SNAPSHOT_MAX_LENGTH = 2000000  # bytes
DEFAULT_TARGETS_MAX_LENGTH = 5000000  # bytes
```
Maybe it does not hurt if they were instance variables instead. They don't even need to be optional constructor arguments (as initialization does not use any of them: caller can initialize the Updater and then modify the values).
I guess the fanciest option is that constructor takes an optional argument dataclass UpdaterOptions with those attributes ...

_Originally posted by @jku in https://github.com/theupdateframework/tuf/issues/1396#issuecomment-842985740_

sechkova:
I am in favour of the dataclass option because it groups all the configuration at one place and I drafted a proposal based on it:  #1470.
There is also `RequestsFetcher` which needs to be configured on init so I added the configuration as an optional parameter to `Updater`.  However I am not totally sure if this solution makes it convenient enough to customise the configuration and I will be happy to hear some opinions about it.

trishankatdatadog:
> I am in favour of the dataclass option because it groups all the configuration at one place and I drafted a proposal based on it: #1470.

[Me too](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/config.py) 🙂

Issue #1400, opened 2021-05-20
ngclient: Use %-style formatting in logging


jku:
(this is about code currently in experimental-client/ngclient branch, hopefully soon in develop)

We currently have some f-strings in logging:

```
                        msg = (
                            f"Adding child role {child_role_name}.\n",
                            "Not backtracking to other roles.",
                        )
                        logger.debug(msg)
```

Let's use %-style instead in logging (even if f-strings are great for other string formatting, see the google python style guide)

avelichka:
I'd like to give it a try, please?

trishankatdatadog:
Curious: why does Google recommend % over f-strings? I think the latter is far more readable, no?

jku:
I agree, it's just that f-strings get greedily evaluated before logger can check if they are needed: with %-style the logger can check if the string is needed (and usually it isn't) before building the string. Obviously 99% of our logging is nowhere near a critical path where this would matter ... but I'd still like to be able to log inside e.g. a download loop without thinking about it.

The other reason would be that our linters complain :) We could of course modify our linter config (and document the style guide difference) but that feels like work too...

trishankatdatadog:
My 0.02 BTC is that we should keep developers not Google/linters/speedtests happy, but that's just me. I will let you guys decide 🙂

Issue #1399, opened 2021-05-19
ngclient: review download code


jku:
(this is about code currently in experimental-client branch, hopefully soon in develop)

After we removed mirrors support, we should look through the download code 
* Add type annotations
* replace old style docstrings
* see if code can be simplified
* consider removing average download speed check (it is useless in practice, will trigger false positives if the limit was usefully low and makes the code complex). It also prevents setting chunk size to a reasonable value (it's currently way too large but I'm afraid to lower it because the average speed check then becomes even more volatile for the first chunks -- if the limit is even close to reasonable, it could trigger for the first chunk much easier than later chunks just because that's how statistics works)
* consider getting rid of strict_required_length: I do not see the value, and it makes the code more complex (if on the other hand there is actual value in length checks, we should implement them in MetadataBundle: the spec does not require them but they would be easy to add)

jku:
Additional reasoning for length check opinion:

> if on the other hand there is actual value in length checks, we should implement them in MetadataBundle

* MetadataBundle ensures new metadata is correct: it checks hashes, signatures and data validity. checking that length is correct fits right in there (even though spec does not require it)
* the download code on the other hand should worry about A) getting the requested bytes B) not downloading more than is needed

sechkova:
> * consider removing average download speed check (it is useless in practice, will trigger false positives if the limit was usefully low and makes the code complex). It also prevents setting chunk size to a reasonable value (it's currently way too large but I'm afraid to lower it because the average speed check then becomes even more volatile for the first chunks -- if the limit is even close to reasonable, it could trigger for the first chunk much easier than later chunks just because that's how statistics works)

Given that defense against slow retrieval attack is considered optional and hard to specify (https://github.com/theupdateframework/specification/issues/124), this sounds like a fair decision.

> * consider getting rid of strict_required_length: I do not see the value, and it makes the code more complex (if on the other hand there is actual value in length checks, we should implement them in MetadataBundle: the spec does not require them but they would be easy to add)

I agree about strict length check. I think we can still keep an upper boundary length check without much cost.

sechkova:
In addition we can consider creating a `Download` class with a `Download` object owned by `Updater` and instantiated with a `Fetcher` (default or user-provided).
We can probably keep Fetcher as it is now (the plug-in networking) and Download will add the file object handling.

Issue #1398, opened 2021-05-19
ngclient: review _preorder_depth_first_walk()


jku:
This is for the new client that's currently in experimental-client branch: but I'm hoping this issue is handled after the experimental client is merged to develop

We've not touched `_preorder_depth_first_walk()` or `_visit_child_role()` yet: they need some love:
First things:
* type annotations would make the whole thing a lot easier to read and reason about
* docstrings are old style
* evaluate if some functionality could/should be in API (e.g. in DelegatedRole)

Then maybe:
* maybe the code can be made simpler?
* review the use of fnmatch(): it seems a bit suspicious (again using filesystem path functions for what I think are url fragments)

joshuagl:
See also a request for clarification on the algorithm in the specification https://github.com/theupdateframework/specification/issues/164

PR #1408, opened 2021-05-21
Merge ngclient: a new client library implementation

jku:
This is something we've been working in experimental-client branch so far and would like to merge to develop now. It's a new client library implementation:
 * Uses Metadata API
 * Is partly based on current tuf/client/, but no longer shares any modules with it (except exceptions)
 * There's a major simplification in the way downloads are handled due to removing mirrors support
 * Tracking of valid metadata has been separated into another component
 * There's no MultiRepoUpdater or mirrors support

The implementation is roughly feature complete and ready for a test-drive. 

There are still TODO-items that we'd like to handle before this could replace the current client, but we feel this would be a good time to merge so future work can be shared and is visible to everyone. Teodora has promised to post a list of things currently on our TODO-list here. 
 
On the PR itself:
 * Review is very welcome, but I hope we can consider ngclient directory a work-in-progress, in that we can file issues instead of fixing all review items here
 * I've decided to squash the commits in this branch (the content matches experimental-client) The reasons were
   * there were a number of file moves while we were deciding what to do
   * we made decisions that we decided to back away from in the branch (but due to other changes did not use revert) so quite a few commits are just red herrings
   * there are ~100 commits and multiple merges of develop in experimental-client so history becomes a little hard to rack

If others would rather see the full history, I can make this PR from the experimental-client branch too.

### DIFF
--- a/tests/test_metadata_bundle.py
+++ b/tests/test_metadata_bundle.py
@@ -1,0 +1,124 @@
+import json
+import logging
+import os
+import shutil
+import sys
+import tempfile
+import unittest
+
+from tuf import exceptions
+from tuf.api.metadata import Metadata
+from tuf.ngclient._internal.metadata_bundle import MetadataBundle
+
+from tests import utils
+
+logger = logging.getLogger(__name__)
+
+class TestMetadataBundle(unittest.TestCase):
+
+    def test_update(self):
+        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
+
+        with open(os.path.join(repo_dir, "root.json"), "rb") as f:
+            bundle = MetadataBundle(f.read())
+        bundle.root_update_finished()
+
+        with open(os.path.join(repo_dir, "timestamp.json"), "rb") as f:
+            bundle.update_timestamp(f.read())
+        with open(os.path.join(repo_dir, "snapshot.json"), "rb") as f:
+            bundle.update_snapshot(f.read())
+        with open(os.path.join(repo_dir, "targets.json"), "rb") as f:
+            bundle.update_targets(f.read())
+        with open(os.path.join(repo_dir, "role1.json"), "rb") as f:
+            bundle.update_delegated_targets(f.read(), "role1", "targets")
+        with open(os.path.join(repo_dir, "role2.json"), "rb") as f:
+            bundle.update_delegated_targets(f.read(), "role2", "role1")
+
+    def test_out_of_order_ops(self):
+        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
+        data={}
+        for md in ["root", "timestamp", "snapshot", "targets", "role1"]:
+            with open(os.path.join(repo_dir, f"{md}.json"), "rb") as f:
+                data[md] = f.read()
+
+        bundle = MetadataBundle(data["root"])
+
+        # Update timestamp before root is finished
+        with self.assertRaises(RuntimeError):
+            bundle.update_timestamp(data["timestamp"])
+
+        bundle.root_update_finished()
+        with self.assertRaises(RuntimeError):
+            bundle.root_update_finished()
+
+        # Update snapshot before timestamp
+        with self.assertRaises(RuntimeError):
+            bundle.update_snapshot(data["snapshot"])
+
+        bundle.update_timestamp(data["timestamp"])
+
+        # Update targets before snapshot
+        with self.assertRaises(RuntimeError):
+            bundle.update_targets(data["targets"])
+
+        bundle.update_snapshot(data["snapshot"])
+
+        #update timestamp after snapshot
+        with self.assertRaises(RuntimeError):
+            bundle.update_timestamp(data["timestamp"])
+
+        # Update delegated targets before targets
+        with self.assertRaises(RuntimeError):
+            bundle.update_delegated_targets(data["role1"], "role1", "targets")
+
+        bundle.update_targets(data["targets"])
+        bundle.update_delegated_targets(data["role1"], "role1", "targets")
+
+    def test_update_with_invalid_json(self):
+        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
+        data={}
+        for md in ["root", "timestamp", "snapshot", "targets", "role1"]:
+            with open(os.path.join(repo_dir, f"{md}.json"), "rb") as f:
+                data[md] = f.read()
+
+        # root.json not a json file at all
+        with self.assertRaises(exceptions.RepositoryError):
+            MetadataBundle(b"")
+        # root.json is invalid
+        root = Metadata.from_bytes(data["root"])
+        root.signed.version += 1
+        with self.assertRaises(exceptions.RepositoryError):
+            MetadataBundle(json.dumps(root.to_dict()).encode())
+
+        bundle = MetadataBundle(data["root"])
+        bundle.root_update_finished()
+
+        top_level_md = [
+            (data["timestamp"], bundle.update_timestamp),
+            (data["snapshot"], bundle.update_snapshot),
+            (data["targets"], bundle.update_targets),
+        ]
+        for metadata, update_func in top_level_md:
+            # metadata is not json
+            with self.assertRaises(exceptions.RepositoryError):
+                update_func(b"")
+            # metadata is invalid
+            md = Metadata.from_bytes(metadata)
+            md.signed.version += 1
+            with self.assertRaises(exceptions.RepositoryError):
+                update_func(json.dumps(md.to_dict()).encode())
+
+            # metadata is of wrong type
+            with self.assertRaises(exceptions.RepositoryError):
+                update_func(data["root"])
+
+            update_func(metadata)
+
+
+    # TODO test updating over initial metadata (new keys, newer timestamp, etc)
+    # TODO test the actual specification checks
+
+
+if __name__ == '__main__':
+  utils.configure_test_logging(sys.argv)
+  unittest.main()
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -7,33 +7,15 @@
 """
 
 import os
-import time
 import shutil
-import copy
 import tempfile
 import logging
-import errno
 import sys
 import unittest
-import json
-import tracemalloc
-
-if sys.version_info >= (3, 3):
-  import unittest.mock as mock
-else:
-  import mock
-
-import tuf
-import tuf.exceptions
-import tuf.log
-import tuf.repository_tool as repo_tool
 import tuf.unittest_toolbox as unittest_toolbox
 
 from tests import utils
-from tuf.api import metadata
 from tuf import ngclient
-
-import securesystemslib
 
 logger = logging.getLogger(__name__)
 
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python
+
+# Copyright 2021, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Test Updater class
+"""
+
+import os
+import time
+import shutil
+import copy
+import tempfile
+import logging
+import errno
+import sys
+import unittest
+import json
+import tracemalloc
+
+if sys.version_info >= (3, 3):
+  import unittest.mock as mock
+else:
+  import mock
+
+import tuf
+import tuf.exceptions
+import tuf.log
+import tuf.repository_tool as repo_tool
+import tuf.unittest_toolbox as unittest_toolbox
+
+from tests import utils
+from tuf.api import metadata
+from tuf import ngclient
+
+import securesystemslib
+
+logger = logging.getLogger(__name__)
+
+
+class TestUpdater(unittest_toolbox.Modified_TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    # Create a temporary directory to store the repository, metadata, and target
+    # files.  'temporary_directory' must be deleted in TearDownModule() so that
+    # temporary files are always removed, even when exceptions occur.
+    cls.temporary_directory = tempfile.mkdtemp(dir=os.getcwd())
+
+    # Needed because in some tests simple_server.py cannot be found.
+    # The reason is that the current working directory
+    # has been changed when executing a subprocess.
+    cls.SIMPLE_SERVER_PATH = os.path.join(os.getcwd(), 'simple_server.py')
+
+    # Launch a SimpleHTTPServer (serves files in the current directory).
+    # Test cases will request metadata and target files that have been
+    # pre-generated in 'tuf/tests/repository_data', which will be served
+    # by the SimpleHTTPServer launched here.  The test cases of 'test_updater.py'
+    # assume the pre-generated metadata files have a specific structure, such
+    # as a delegated role 'targets/role1', three target files, five key files,
+    # etc.
+    cls.server_process_handler = utils.TestServerProcess(log=logger,
+        server=cls.SIMPLE_SERVER_PATH)
+
+
+
+  @classmethod
+  def tearDownClass(cls):
+    # Cleans the resources and flush the logged lines (if any).
+    cls.server_process_handler.clean()
+
+    # Remove the temporary repository directory, which should contain all the
+    # metadata, targets, and key files generated for the test cases
+    shutil.rmtree(cls.temporary_directory)
+
+
+
+  def setUp(self):
+    # We are inheriting from custom class.
+    unittest_toolbox.Modified_TestCase.setUp(self)
+
+    # Copy the original repository files provided in the test folder so that
+    # any modifications made to repository files are restricted to the copies.
+    # The 'repository_data' directory is expected to exist in 'tuf.tests/'.
+    original_repository_files = os.path.join(os.getcwd(), 'repository_data')
+    temporary_repository_root = \
+      self.make_temp_directory(directory=self.temporary_directory)
+
+    # The original repository, keystore, and client directories will be copied
+    # for each test case.
+    original_repository = os.path.join(original_repository_files, 'repository')
+    original_keystore = os.path.join(original_repository_files, 'keystore')
+    original_client = os.path.join(original_repository_files, 'client', 'test_repository1', 'metadata', 'current')
+
+    # Save references to the often-needed client repository directories.
+    # Test cases need these references to access metadata and target files.
+    self.repository_directory = \
+      os.path.join(temporary_repository_root, 'repository')
+    self.keystore_directory = \
+      os.path.join(temporary_repository_root, 'keystore')
+
+    self.client_directory = os.path.join(temporary_repository_root, 'client')
+
+    # Copy the original 'repository', 'client', and 'keystore' directories
+    # to the temporary repository the test cases can use.
+    shutil.copytree(original_repository, self.repository_directory)
+    shutil.copytree(original_client, self.client_directory)
+    shutil.copytree(original_keystore, self.keystore_directory)
+
+    # 'path/to/tmp/repository' -> 'localhost:8001/tmp/repository'.
+    repository_basepath = self.repository_directory[len(os.getcwd()):]
+    url_prefix = 'http://' + utils.TEST_HOST_ADDRESS + ':' \
+        + str(self.server_process_handler.port) + repository_basepath
+
+    metadata_url = f"{url_prefix}/metadata/"
+    targets_url = f"{url_prefix}/targets/"
+    # Creating a repository instance.  The test cases will use this client
+    # updater to refresh metadata, fetch target files, etc.
+    self.repository_updater = ngclient.Updater(self.client_directory,
+                                              metadata_url,
+                                              targets_url)
+
+  def tearDown(self):
+    # We are inheriting from custom class.
+    unittest_toolbox.Modified_TestCase.tearDown(self)
+
+    # Logs stdout and stderr from the sever subprocess.
+    self.server_process_handler.flush_log()
+
+  def test_refresh(self):
+    # All metadata is in local directory already
+    self.repository_updater.refresh()
+
+    # Get targetinfo for 'file1.txt' listed in targets
+    targetinfo1 = self.repository_updater.get_one_valid_targetinfo('file1.txt')
+    # Get targetinfo for 'file3.txt' listed in the delegated role1
+    targetinfo3= self.repository_updater.get_one_valid_targetinfo('file3.txt')
+
+    destination_directory = self.make_temp_directory()
+    updated_targets = self.repository_updater.updated_targets([targetinfo1, targetinfo3],
+                                                      destination_directory)
+
+    self.assertListEqual(updated_targets, [targetinfo1, targetinfo3])
+
+    self.repository_updater.download_target(targetinfo1, destination_directory)
+    updated_targets = self.repository_updater.updated_targets(updated_targets,
+                                                      destination_directory)
+
+    self.assertListEqual(updated_targets, [targetinfo3])
+
+
+    self.repository_updater.download_target(targetinfo3, destination_directory)
+    updated_targets = self.repository_updater.updated_targets(updated_targets,
+                                                  destination_directory)
+
+    self.assertListEqual(updated_targets, [])
+
+  def test_refresh_with_only_local_root(self):
+    os.remove(os.path.join(self.client_directory, "timestamp.json"))
+    os.remove(os.path.join(self.client_directory, "snapshot.json"))
+    os.remove(os.path.join(self.client_directory, "targets.json"))
+    os.remove(os.path.join(self.client_directory, "role1.json"))
+
+    self.repository_updater.refresh()
+
+    # Get targetinfo for 'file3.txt' listed in the delegated role1
+    targetinfo3= self.repository_updater.get_one_valid_targetinfo('file3.txt')
+
+if __name__ == '__main__':
+  utils.configure_test_logging(sys.argv)
+  unittest.main()
--- a/tuf/ngclient/README.md
+++ b/tuf/ngclient/README.md
@@ -7,16 +7,16 @@ described in the TUF specification (see
 https://theupdateframework.github.io/specification/latest/#detailed-client-workflow)
 
 **tuf.ngclient.FetcherInterface** is an abstract class that client
-implementers can optionally use to integrate with their own
-network/download infrastructure -- a Requests-based implementation is
+implementers can implement a concrete class of in order to reuse their
+own networking/download libraries -- a Requests-based implementation is
 used by default.
 
 This package:
 * Aims to be a clean, easy-to-validate reference client implementation
   written in modern Python
-* At the same time aims to be the library choice for anyone
+* At the same time aims to be the library of choice for anyone
   implementing a TUF client in Python: light-weight, easy to integrate
   and with minimal required dependencies
-* Is still under development but planned to become the default client
-  in this code base (as in the older tuf.client will be deprecated in
-  the future)
+* Is still under development but is planned to become the default client
+  in this implementation (i.e., the older `tuf.client` will be deprecated
+  in the future)
--- a/tuf/ngclient/README.md
+++ b/tuf/ngclient/README.md
@@ -1,0 +1,22 @@
+## Next-gen TUF client for Python
+
+This package provides modules for TUF client implementers.
+
+**tuf.ngclient.Updater** is a class that implements the client workflow
+described in the TUF specification (see
+https://theupdateframework.github.io/specification/latest/#detailed-client-workflow)
+
+**tuf.ngclient.FetcherInterface** is an abstract class that client
+implementers can optionally use to integrate with their own
+network/download infrastructure -- a Requests-based implementation is
+used by default.
+
+This package:
+* Aims to be a clean, easy-to-validate reference client implementation
+  written in modern Python
+* At the same time aims to be the library choice for anyone
+  implementing a TUF client in Python: light-weight, easy to integrate
+  and with minimal required dependencies
+* Is still under development but planned to become the default client
+  in this code base (as in the older tuf.client will be deprecated in
+  the future)
--- a/tuf/ngclient/_internal/download.py
+++ b/tuf/ngclient/_internal/download.py
@@ -1,0 +1,243 @@
+#!/usr/bin/env python
+
+# Copyright 2012 - 2017, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""
+<Program Name>
+  download.py
+
+<Started>
+  February 21, 2012.  Based on previous version by Geremy Condra.
+
+<Author>
+  Konstantin Andrianov
+  Vladimir Diaz <vladimir.v.diaz@gmail.com>
+
+<Copyright>
+  See LICENSE-MIT OR LICENSE for licensing information.
+
+<Purpose>
+  Download metadata and target files and check their validity.  The hash and
+  length of a downloaded file has to match the hash and length supplied by the
+  metadata of that file.
+"""
+
+import logging
+import tempfile
+import timeit
+from urllib import parse
+
+from securesystemslib import formats as sslib_formats
+
+import tuf
+from tuf import exceptions, formats
+
+# See 'log.py' to learn how logging is handled in TUF.
+logger = logging.getLogger(__name__)
+
+
+def download_file(url, required_length, fetcher, strict_required_length=True):
+    """
+    <Purpose>
+      Given the url and length of the desired file, this function opens a
+      connection to 'url' and downloads the file while ensuring its length
+      matches 'required_length' if 'STRICT_REQUIRED_LENGH' is True (If False,
+      the file's length is not checked and a slow retrieval exception is raised
+      if the downloaded rate falls below the acceptable rate).
+
+    <Arguments>
+      url:
+        A URL string that represents the location of the file.
+
+      required_length:
+        An integer value representing the length of the file.
+
+      strict_required_length:
+        A Boolean indicator used to signal whether we should perform strict
+        checking of required_length. True by default. We explicitly set this to
+        False when we know that we want to turn this off for downloading the
+        timestamp metadata, which has no signed required_length.
+
+    <Side Effects>
+      A file object is created on disk to store the contents of 'url'.
+
+    <Exceptions>
+      exceptions.DownloadLengthMismatchError, if there was a
+      mismatch of observed vs expected lengths while downloading the file.
+
+      securesystemslib.exceptions.FormatError, if any of the arguments are
+      improperly formatted.
+
+      Any other unforeseen runtime exception.
+
+    <Returns>
+      A file object that points to the contents of 'url'.
+    """
+    # Do all of the arguments have the appropriate format?
+    # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
+    sslib_formats.URL_SCHEMA.check_match(url)
+    formats.LENGTH_SCHEMA.check_match(required_length)
+
+    # 'url.replace('\\', '/')' is needed for compatibility with Windows-based
+    # systems, because they might use back-slashes in place of forward-slashes.
+    # This converts it to the common format.  unquote() replaces %xx escapes in
+    # a url with their single-character equivalent.  A back-slash may be
+    # encoded as %5c in the url, which should also be replaced with a forward
+    # slash.
+    url = parse.unquote(url).replace("\\", "/")
+    logger.info("Downloading: %s", url)
+
+    # This is the temporary file that we will return to contain the contents of
+    # the downloaded file.
+    temp_file = tempfile.TemporaryFile()  # pylint: disable=consider-using-with
+
+    average_download_speed = 0
+    number_of_bytes_received = 0
+
+    try:
+        chunks = fetcher.fetch(url, required_length)
+        start_time = timeit.default_timer()
+        for chunk in chunks:
+
+            stop_time = timeit.default_timer()
+            temp_file.write(chunk)
+
+            # Measure the average download speed.
+            number_of_bytes_received += len(chunk)
+            seconds_spent_receiving = stop_time - start_time
+            average_download_speed = (
+                number_of_bytes_received / seconds_spent_receiving
+            )
+
+            if average_download_speed < tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED:
+                logger.debug(
+                    "The average download speed dropped below the minimum"
+                    " average download speed set in tuf.settings.py."
+                    " Stopping the download!"
+                )
+                break
+
+            logger.debug(
+                "The average download speed has not dipped below the"
+                " minimum average download speed set in tuf.settings.py."
+            )
+
+        # Does the total number of downloaded bytes match the required length?
+        _check_downloaded_length(
+            number_of_bytes_received,
+            required_length,
+            strict_required_length=strict_required_length,
+            average_download_speed=average_download_speed,
+        )
+
+    except Exception:
+        # Close 'temp_file'.  Any written data is lost.
+        temp_file.close()
+        logger.debug("Could not download URL: %s", url)
+        raise
+
+    else:
+        temp_file.seek(0)
+        return temp_file
+
+
+def download_bytes(url, required_length, fetcher, strict_required_length=True):
+    """Download bytes from given url
+
+    Returns the downloaded bytes, otherwise like download_file()
+    """
+    with download_file(
+        url, required_length, fetcher, strict_required_length
+    ) as dl_file:
+        return dl_file.read()
+
+
+def _check_downloaded_length(
+    total_downloaded,
+    required_length,
+    strict_required_length=True,
+    average_download_speed=None,
+):
+    """
+    <Purpose>
+      A helper function which checks whether the total number of downloaded
+      bytes matches our expectation.
+
+    <Arguments>
+      total_downloaded:
+        The total number of bytes supposedly downloaded for the file in
+        question.
+
+      required_length:
+        The total number of bytes expected of the file as seen from its metadata
+        The Timestamp role is always downloaded without a known file length, and
+        the Root role when the client cannot download any of the required
+        top-level roles.  In both cases, 'required_length' is actually an upper
+        limit on the length of the downloaded file.
+
+      strict_required_length:
+        A Boolean indicator used to signal whether we should perform strict
+        checking of required_length. True by default. We explicitly set this to
+        False when we know that we want to turn this off for downloading the
+        timestamp metadata, which has no signed required_length.
+
+      average_download_speed:
+       The average download speed for the downloaded file.
+
+    <Side Effects>
+      None.
+
+    <Exceptions>
+      securesystemslib.exceptions.DownloadLengthMismatchError, if
+      strict_required_length is True and total_downloaded is not equal
+      required_length.
+
+      exceptions.SlowRetrievalError, if the total downloaded was
+      done in less than the acceptable download speed (as set in
+      tuf.settings.py).
+
+    <Returns>
+      None.
+    """
+
+    if total_downloaded == required_length:
+        logger.info("Downloaded %d bytes as expected.", total_downloaded)
+
+    else:
+        # What we downloaded is not equal to the required length, but did we ask
+        # for strict checking of required length?
+        if strict_required_length:
+            logger.info(
+                "Downloaded %d bytes, but expected %d bytes",
+                total_downloaded,
+                required_length,
+            )
+
+            # If the average download speed is below a certain threshold, we
+            # flag this as a possible slow-retrieval attack.
+            if average_download_speed < tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED:
+                raise exceptions.SlowRetrievalError(average_download_speed)
+
+            raise exceptions.DownloadLengthMismatchError(
+                required_length, total_downloaded
+            )
+
+        # We specifically disabled strict checking of required length, but
+        # we will log a warning anyway. This is useful when we wish to
+        # download the Timestamp or Root metadata, for which we have no
+        # signed metadata; so, we must guess a reasonable required_length
+        # for it.
+        if average_download_speed < tuf.settings.MIN_AVERAGE_DOWNLOAD_SPEED:
+            raise exceptions.SlowRetrievalError(average_download_speed)
+
+        logger.debug(
+            "Good average download speed: %f bytes per second",
+            average_download_speed,
+        )
+
+        logger.info(
+            "Downloaded %d bytes out of upper limit of %d bytes.",
+            total_downloaded,
+            required_length,
+        )
--- a/tuf/ngclient/_internal/metadata_bundle.py
+++ b/tuf/ngclient/_internal/metadata_bundle.py
@@ -240,6 +240,11 @@ class MetadataBundle(abc.Mapping):
         if self.root.signed.is_expired(self.reference_time):
             raise exceptions.ExpiredMetadataError("New root.json is expired")
 
+        # No need to recover from fast-forward attack here since
+        # timestamp and snapshot are not loaded at this point and
+        # when loaded later will be verified with the new rotated
+        # keys.
+
         self._root_update_finished = True
         logger.debug("Verified final root.json")
 
--- a/tuf/ngclient/_internal/metadata_bundle.py
+++ b/tuf/ngclient/_internal/metadata_bundle.py
@@ -356,7 +356,7 @@ class MetadataBundle(abc.Mapping):
 
         # If an existing trusted snapshot is updated,
         # check for a rollback attack
-        if self.snapshot:
+        if self.snapshot is not None:
             for filename, fileinfo in self.snapshot.signed.meta.items():
                 new_fileinfo = new_snapshot.signed.meta.get(filename)
 
--- a/tuf/ngclient/_internal/metadata_bundle.py
+++ b/tuf/ngclient/_internal/metadata_bundle.py
@@ -1,0 +1,458 @@
+# Copyright the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""TUF client bundle-of-metadata
+
+MetadataBundle keeps track of current valid set of metadata for the client,
+and handles almost every step of the "Detailed client workflow" (
+https://theupdateframework.github.io/specification/latest#detailed-client-workflow)
+in the TUF specification: the remaining steps are related to filesystem and
+network IO which is not handled here.
+
+Loaded metadata can be accessed via the index access with rolename as key
+(bundle["root"]) or, in the case of top-level metadata using the helper
+properties (bundle.root).
+
+The rules for top-level metadata are
+ * Metadata is loadable only if metadata it depends on is loaded
+ * Metadata is immutable if any metadata depending on it has been loaded
+ * Metadata must be loaded/updated in order:
+   root -> timestamp -> snapshot -> targets -> (other delegated targets)
+
+
+Exceptions are raised if metadata fails to load in any way.
+
+Example of loading root, timestamp and snapshot:
+
+>>> # Load local root (RepositoryErrors here stop the update)
+>>> with open(root_path, "rb") as f:
+>>>     bundle = MetadataBundle(f.read())
+>>>
+>>> # update root from remote until no more are available
+>>> with download("root", bundle.root.signed.version + 1) as f:
+>>>     bundle.update_root(f.read())
+>>> # ...
+>>> bundle.root_update_finished()
+>>>
+>>> # load local timestamp, then update from remote
+>>> try:
+>>>     with open(timestamp_path, "rb") as f:
+>>>         bundle.update_timestamp(f.read())
+>>> except (RepositoryError, OSError):
+>>>     pass # failure to load a local file is ok
+>>>
+>>> with download("timestamp") as f:
+>>>     bundle.update_timestamp(f.read())
+>>>
+>>> # load local snapshot, then update from remote if needed
+>>> try:
+>>>     with open(snapshot_path, "rb") as f:
+>>>         bundle.update_snapshot(f.read())
+>>> except (RepositoryError, OSError):
+>>>     # local snapshot is not valid, load from remote
+>>>     # (RepositoryErrors here stop the update)
+>>>     with download("snapshot", version) as f:
+>>>         bundle.update_snapshot(f.read())
+
+TODO:
+ * exceptions are not final: the idea is that client could just handle
+   a generic RepositoryError that covers every issue that server provided
+   metadata could inflict (other errors would be user errors), but this is not
+   yet the case
+ * usefulness of root_update_finished() can be debated: it could be done
+   in the beginning of load_timestamp()...
+ * some metadata interactions might work better in Metadata itself
+ * Progress through Specification update process should be documented
+   (not sure yet how: maybe a spec_logger that logs specification events?)
+"""
+
+import logging
+from collections import abc
+from datetime import datetime
+from typing import Dict, Iterator, Optional
+
+from securesystemslib import hash as sslib_hash
+
+from tuf import exceptions
+from tuf.api.metadata import Metadata, Root, Targets
+from tuf.api.serialization import DeserializationError
+
+logger = logging.getLogger(__name__)
+
+# This is a placeholder until ...
+# TODO issue 1306: implement this in Metadata API
+def verify_with_threshold(
+    delegator: Metadata, role_name: str, unverified: Metadata
+) -> bool:
+    """Verify 'unverified' with keys and threshold defined in delegator"""
+    role = None
+    keys = {}
+    if isinstance(delegator.signed, Root):
+        keys = delegator.signed.keys
+        role = delegator.signed.roles.get(role_name)
+    elif isinstance(delegator.signed, Targets):
+        if delegator.signed.delegations:
+            keys = delegator.signed.delegations.keys
+            # role names are unique: first match is enough
+            roles = delegator.signed.delegations.roles
+            role = next((r for r in roles if r.name == role_name), None)
+    else:
+        raise ValueError("Call is valid only on delegator metadata")
+
+    if role is None:
+        raise ValueError(f"Delegated role {role_name} not found")
+
+    # verify that delegate is signed by correct threshold of unique keys
+    unique_keys = set()
+    for keyid in role.keyids:
+        key = keys[keyid]
+        try:
+            key.verify_signature(unverified)
+            unique_keys.add(key.keyval["public"])
+        except Exception as e:  # pylint: disable=broad-except
+            # TODO specify the Exceptions (see issue #1351)
+            logger.info("verify failed: %s", e)
+
+    return len(unique_keys) >= role.threshold
+
+
+class MetadataBundle(abc.Mapping):
+    """Internal class to keep track of valid metadata in Updater
+
+    MetadataBundle ensures that the collection of metadata in the bundle is
+    valid. It provides easy ways to update the metadata with the caller making
+    decisions on what is updated.
+    """
+
+    def __init__(self, root_data: bytes):
+        """Initialize bundle by loading trusted root metadata
+
+        Args:
+            root_data: Trusted root metadata as bytes. Note that this metadata
+                will only be verified by itself: it is the source of trust for
+                all metadata in the bundle.
+
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+        self._bundle = {}  # type: Dict[str: Metadata]
+        self.reference_time = datetime.utcnow()
+        self._root_update_finished = False
+
+        # Load and validate the local root metadata. Valid initial trusted root
+        # metadata is required
+        logger.debug("Updating initial trusted root")
+        self.update_root(root_data)
+
+    def __getitem__(self, role: str) -> Metadata:
+        """Returns current Metadata for 'role'"""
+        return self._bundle[role]
+
+    def __len__(self) -> int:
+        """Returns number of Metadata objects in bundle"""
+        return len(self._bundle)
+
+    def __iter__(self) -> Iterator[Metadata]:
+        """Returns iterator over all Metadata objects in bundle"""
+        return iter(self._bundle)
+
+    # Helper properties for top level metadata
+    @property
+    def root(self) -> Optional[Metadata]:
+        """Current root Metadata or None"""
+        return self._bundle.get("root")
+
+    @property
+    def timestamp(self) -> Optional[Metadata]:
+        """Current timestamp Metadata or None"""
+        return self._bundle.get("timestamp")
+
+    @property
+    def snapshot(self) -> Optional[Metadata]:
+        """Current snapshot Metadata or None"""
+        return self._bundle.get("snapshot")
+
+    @property
+    def targets(self) -> Optional[Metadata]:
+        """Current targets Metadata or None"""
+        return self._bundle.get("targets")
+
+    # Methods for updating metadata
+    def update_root(self, data: bytes):
+        """Verifies and loads 'data' as new root metadata.
+
+        Note that an expired intermediate root is considered valid: expiry is
+        only checked for the final root in root_update_finished().
+
+        Args:
+            data: unverified new root metadata as bytes
+
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+        if self._root_update_finished:
+            raise RuntimeError(
+                "Cannot update root after root update is finished"
+            )
+        logger.debug("Updating root")
+
+        try:
+            new_root = Metadata.from_bytes(data)
+        except DeserializationError as e:
+            raise exceptions.RepositoryError("Failed to load root") from e
+
+        if new_root.signed.type != "root":
+            raise exceptions.RepositoryError(
+                f"Expected 'root', got '{new_root.signed.type}'"
+            )
+
+        if self.root is not None:
+            # We are not loading initial trusted root: verify the new one
+            if not verify_with_threshold(self.root, "root", new_root):
+                raise exceptions.UnsignedMetadataError(
+                    "New root is not signed by root", new_root.signed
+                )
+
+            if new_root.signed.version != self.root.signed.version + 1:
+                raise exceptions.ReplayedMetadataError(
+                    "root", new_root.signed.version, self.root.signed.version
+                )
+
+        if not verify_with_threshold(new_root, "root", new_root):
+            raise exceptions.UnsignedMetadataError(
+                "New root is not signed by itself", new_root.signed
+            )
+
+        self._bundle["root"] = new_root
+        logger.debug("Updated root")
+
+    def root_update_finished(self):
+        """Marks root metadata as final and verifies it is not expired
+
+        Raises:
+            ExpiredMetadataError: The final root metadata is expired.
+        """
+        if self._root_update_finished:
+            raise RuntimeError("Root update is already finished")
+
+        if self.root.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("New root.json is expired")
+
+        self._root_update_finished = True
+        logger.debug("Verified final root.json")
+
+    def update_timestamp(self, data: bytes):
+        """Verifies and loads 'data' as new timestamp metadata.
+
+        Args:
+            data: unverified new timestamp metadata as bytes
+
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+        if not self._root_update_finished:
+            # root_update_finished() not called
+            raise RuntimeError("Cannot update timestamp before root")
+        if self.snapshot is not None:
+            raise RuntimeError("Cannot update timestamp after snapshot")
+
+        try:
+            new_timestamp = Metadata.from_bytes(data)
+        except DeserializationError as e:
+            raise exceptions.RepositoryError("Failed to load timestamp") from e
+
+        if new_timestamp.signed.type != "timestamp":
+            raise exceptions.RepositoryError(
+                f"Expected 'timestamp', got '{new_timestamp.signed.type}'"
+            )
+
+        if not verify_with_threshold(self.root, "timestamp", new_timestamp):
+            raise exceptions.UnsignedMetadataError(
+                "New timestamp is not signed by root", new_timestamp.signed
+            )
+
+        if self.timestamp is not None:
+            # Prevent rolling back timestamp version
+            if new_timestamp.signed.version < self.timestamp.signed.version:
+                raise exceptions.ReplayedMetadataError(
+                    "timestamp",
+                    new_timestamp.signed.version,
+                    self.timestamp.signed.version,
+                )
+            # Prevent rolling back snapshot version
+            if (
+                new_timestamp.signed.meta["snapshot.json"].version
+                < self.timestamp.signed.meta["snapshot.json"].version
+            ):
+                # TODO not sure about the correct exception here
+                raise exceptions.ReplayedMetadataError(
+                    "snapshot",
+                    new_timestamp.signed.meta["snapshot.json"].version,
+                    self.timestamp.signed.meta["snapshot.json"].version,
+                )
+
+        if new_timestamp.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("New timestamp is expired")
+
+        self._bundle["timestamp"] = new_timestamp
+        logger.debug("Updated timestamp")
+
+    # TODO: remove pylint disable once the hash verification is in metadata.py
+    def update_snapshot(self, data: bytes):  # pylint: disable=too-many-branches
+        """Verifies and loads 'data' as new snapshot metadata.
+
+        Args:
+            data: unverified new snapshot metadata as bytes
+
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+
+        if self.timestamp is None:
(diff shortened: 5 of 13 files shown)


sechkova:
As mentioned in the description, we don't consider this PR as a fully complete client and here is the list of our current TODOs which very likely will be extended further:

- review download code #1399
- review _preorder_depth_first_walk() #1398
- consider API changes #1317
- revise exceptions #1312
- add tests #1309
- use %-style formatting in logging #1400
- proper public module documentation #1385

jku:
I'll add a few more potential TODOs that I dug from my notes:
* experiment with separating local target caching from updater #1412
* use MetaFile/TargetFile verification API when it exists #1410
* type annotations review and mypy integration #1409 
* Use Metadata APIs verify_with_threshold() when possible #1413

trishankatdatadog:
Thanks for the great work, and would love to review! Give me a deadline, please...

joshuagl:
> Thanks for the great work, and would love to review! Give me a deadline, please...

Thanks @trishankatdatadog! The team are focused on supporting changes targeting develop (with no dependency on this PR) for the [next two weeks](https://github.com/theupdateframework/tuf/milestone/12), but after that it would be really good to keep moving this forward. Would a deadline of Wed 9th work?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+>>>     bundle.update_root(f.read())
+>>> # ...
+>>> bundle.root_update_finished()
+>>>
+>>> # load local timestamp, then update from remote
+>>> try:
+>>>     with open(timestamp_path, "rb") as f:
+>>>         bundle.update_timestamp(f.read())
+>>> except (RepositoryError, OSError):
+>>>     pass # failure to load a local file is ok
```
MVrachev: From what I understand, the `RepositoryError` exception will be caught when`BadVersionNumberError` or `ExpiredMetadataError` is thrown (commit https://github.com/theupdateframework/tuf/pull/1408/commits/91935fa2c46eb39e4e6a61719ef7b9e905d6bca2).
So, I don't think this is a simple `fail to load a local file`.

Maybe we shouldn't give examples where you would want to just `pass` on `RepositoryError` when this exception could be a serious one?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+TODO:
+ * exceptions are not final: the idea is that client could just handle
+   a generic RepositoryError that covers every issue that server provided
+   metadata could inflict (other errors would be user errors), but this is not
+   yet the case
+ * usefulness of root_update_finished() can be debated: it could be done
+   in the beginning of load_timestamp()...
+ * some metadata interactions might work better in Metadata itself
+ * Progress through Specification update process should be documented
+   (not sure yet how: maybe a spec_logger that logs specification events?)
```
MVrachev: Do we have an issue for this one?
Seems useful for debugging. 
The suggestion for `spec_logger.log` or something sounds reasonable.

tuf/ngclient/_internal/metadata_bundle.py
jku: the question is, when would you want to stop the update process here? I don't think those cases exist: I don't want to stop if the local metadata is expired, corrupted or not signed with current keys for example.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+        if delegator.signed.delegations:
+            keys = delegator.signed.delegations.keys
+            # role names are unique: first match is enough
+            roles = delegator.signed.delegations.roles
+            role = next((r for r in roles if r.name == role_name), None)
+    else:
+        raise ValueError("Call is valid only on delegator metadata")
+
+    if role is None:
+        raise ValueError(f"Delegated role {role_name} not found")
```
MVrachev: Do you allow calling `verify_with_threshold` when `delegator.signed` is `Targets`, but there are no delegations?
Or do you make that check somewhere else?
Then, you would end up with `role` which is None, and throw that exception.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+
+    # verify that delegate is signed by correct threshold of unique keys
+    unique_keys = set()
+    for keyid in role.keyids:
+        key_dict = keys[keyid].to_dict()
+        key, dummy = sslib_keys.format_metadata_to_key(key_dict)
```
MVrachev: Nit: maybe you can just `key, _ = sslib_keys.format_metadata_to_key(key_dict)` instead of giving a name to the second value.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+
+        if not verify_with_threshold(new_root, "root", new_root):
+            raise exceptions.UnsignedMetadataError(
+                "New root is not signed by itself", new_root.signed
+            )
+
+        self._bundle["root"] = new_root
+        logger.debug("Updated root")
+
+    def root_update_finished(self):
```
MVrachev: Do we want to be able to reset `root_update_finished` back to `False` at some point? 
Most likely it's reasonable to assume that a `MetadataBundle` instance won't live beyond one root version, but I wanted to make sure I understand the assumption. 
If this assumption is true, then do we want to document that somewhere?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+            raise RuntimeError("Cannot update timestamp before root")
+        if self.snapshot is not None:
+            raise RuntimeError("Cannot update timestamp after snapshot")
+
+        try:
+            new_timestamp = Metadata.from_bytes(data)
+        except DeserializationError as e:
+            raise exceptions.RepositoryError("Failed to load timestamp") from e
+
+        if new_timestamp.signed.type != "timestamp":
```
MVrachev: I just realized that we are checking `Timestamp.type` a couple of times:
1. once when we call `Metadata.from_bytes()` -> `deserializer.deserialize()` -> `Metadata.from_dict()` and there we check that `_type` is one from the four metadata types.
2. from `Metadata.from_dict()` ->  `Timestamp.from_dict()` -> `Signed._common_fields_from_dict()` and there we check `if _type != cls._signed_type:`
3. is this check

I understand why you made that check here, I am just wondering do we need all of them? 
PS: This is true for all of the top metadata classes.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+                    "timestamp",
+                    new_timestamp.signed.version,
+                    self.timestamp.signed.version,
+                )
+            # Prevent rolling back snapshot version
+            if (
+                new_timestamp.signed.meta["snapshot.json"].version
+                < self.timestamp.signed.meta["snapshot.json"].version
+            ):
+                # TODO not sure about the correct exception here
```
MVrachev: Do we want another exception here?
I thought that this is the correct one?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+        if self.timestamp is None:
+            raise RuntimeError("Cannot update snapshot before timestamp")
+        if self.targets is not None:
+            raise RuntimeError("Cannot update snapshot after targets")
+        logger.debug("Updating snapshot")
+
+        meta = self.timestamp.signed.meta["snapshot.json"]
+
+        # Verify against the hashes in timestamp, if any
+        hashes = meta.hashes or {}
```
MVrachev: Nit: This line made me question the default value for `MetaFile.hashes`.
Maybe we can use an empty dictionary instead of `None`?
Then, you wouldn't have to use or.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+        for algo, stored_hash in hashes.items():
+            digest_object = sslib_hash.digest(algo)
+            digest_object.update(data)
+            observed_hash = digest_object.hexdigest()
+            if observed_hash != stored_hash:
+                # TODO: Error should derive from RepositoryError
```
MVrachev: I think this is fixed in https://github.com/theupdateframework/tuf/pull/1408/commits/91935fa2c46eb39e4e6a61719ef7b9e905d6bca2.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+>>>     # local snapshot is not valid, load from remote
+>>>     # (RepositoryErrors here stop the update)
+>>>     with download("snapshot", version) as f:
+>>>         bundle.update_snapshot(f.read())
+
+TODO:
+ * exceptions are not final: the idea is that client could just handle
+   a generic RepositoryError that covers every issue that server provided
+   metadata could inflict (other errors would be user errors), but this is not
+   yet the case
```
MVrachev: I am not sure if this is not the case anymore.
Do you still throw exceptions that aren't `RepositoryError` or don't inherit `RepositoryError`? 
The only one I see is `RuntimeError` and `ValueError`, but we don't want to document them right?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,575 @@
...
+        """
+        return self._preorder_depth_first_walk(target_path)
+
+    @staticmethod
+    def updated_targets(targets: Dict, destination_directory: str) -> Dict:
+        """
+        After the client has retrieved the target information for those targets
+        they are interested in updating, they would call this method to
+        determine which targets have changed from those saved locally on disk.
+        All the targets that have changed are returns in a list.  From this
```
MVrachev:
```suggestion
        All the targets that have changed are returned in a list.  From this
```

tests/test_updater_ng.py
```diff
@@ -0,0 +1,171 @@
...
+import copy
+import tempfile
+import logging
+import errno
+import sys
+import unittest
+import json
+import tracemalloc
+
+if sys.version_info >= (3, 3):
```
MVrachev: I don't think we need this check given that we support Python3.6+.

MVrachev (changes requested):
I am sorry I left two reviews according to GitHub. 
I thought I had just left comments there, but somehow I shared it as a review...

I like the `MetadataBundle`.
It's simple and easy to read. Can't wait to see what will happen when we update the rest of the code!
There are one or two important comments I want us to answer, the others are style improvements.

tuf/ngclient/_internal/metadata_bundle.py
jku: yeah, I don't have an issue... I'm still on the fence about it -- it sounds like a nice idea but it does add quite a few lines of code that maybe aren't necessary (the fact that something is an item in the spec might not mean it's worth logging)

tuf/ngclient/_internal/metadata_bundle.py
jku: I do allow that on purpose, and I think the exception is correct: caller is trying to verify a role that isn't delegated by delegator.

tuf/ngclient/_internal/metadata_bundle.py
jku: I agree but our pylint configuration does not currently allow that. Possibly we should change that (Google style guide and their pylintrc does allow it)

tuf/ngclient/_internal/metadata_bundle.py
jku: Correct, idea is that if you ever want to "refresh again", you create a new bundle and start from scratch -- or in other words: updating a specific piece of metadata is only possible until you have loaded any "depending metadata" (this means delegated metadata but also the special cases for top-level metadata where they each depend on the previous one in order _timestamp -> snapshot -> targets_)

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: But it's a valid allowed situation to have Targets without delegations and thus call  `verify_with_threshold` which will throw that exception.
Then, you should add a check inside `update_targets` which doesn't call `verify_with_threshold` on empty delegations or catch the exception.

tuf/ngclient/_internal/metadata_bundle.py
jku: Good question:
Number 1 is the dispatcher, we need to do this somewhere.
Number 2 is just a sanity check IIRC: could be removed but it doesn't hurt anyone
Number 3 we do need but you are right that it looks like we could do better: I wish I could do something like this:
```
try:
    new_timestamp = Metadata[Timestamp].from_bytes(data)
except DeserializationError as e:
    raise exceptions.RepositoryError("Failed to load timestamp") from e
```
Where metadata was a Generic type over (Root/Timestamp/Snapshot/Targets) and the constructor used above would
* fail if the type is not `Timestamp.type`
* type annotate new_timestamp.signed correctly as `Timestamp`

I have no idea if this is possible: I've done an experiment and managed to get close but not quite there.

A simpler alternative is adding an optional argument to the factory constructors: `Metadata.from_bytes(data, type=Timestamp)` which would raise if the Type is not correct (but would do nothing to fix the annotation issue)

tuf/ngclient/_internal/metadata_bundle.py
jku: I thought about it but doing that in Metadata means the content changes when you do deserialize + serialize: no hashes becomes empty hashes...

tuf/ngclient/_internal/metadata_bundle.py
jku: I think you are probably correct but I also think another close look would be good

tuf/ngclient/_internal/metadata_bundle.py
jku: yes I guess that's fine

tuf/ngclient/_internal/metadata_bundle.py
jku: btw, we should polish this code in the PR for #1306 (I plan to do one soon) -- it's here just as a placeholder

jku: can you comment in the code where we should handle the exception because I am not sure what this refers to?

If there are no delegations, then how does the client end up calling update_targets() for a delegate?

tuf/ngclient/_internal/metadata_bundle.py
jku: -> #1433

tuf/ngclient/_internal/metadata_bundle.py
jku: btw, Teodora noticed that this call should not be done: this is being fixed in #1417

tuf/ngclient/_internal/metadata_bundle.py
jku: at least hash verify failure is not a repositoryerror yet

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: If your local keys are broken how you are supposed to verify the new timestamp file?
How do you recover from that?

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: Okay, well this shouldn't stop us from merging this. 
We have an issue where we document this step, so I will consider this comment as resolved.

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: I imagine logging will be useful for the bigger more time-consuming steps.
For example, logging when the update process is finished for each of the metadata classes.

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: You can call it here: https://github.com/jku/tuf/blob/f057d57e6aa2cc1bad1a153343fc6335c9c5c11c/tuf/ngclient/_internal/metadata_bundle.py#L445 line 445 in your current changes.
What's interesting is that you aren't testing if we support `targets` with no `delegations` field.

The reason is that you don't call `update_delegated_targets()` in your tests with a target file without `delegations`.
I thought you should been testing this in `test_metadata_bundle.py` when updating `role2.json` like this:
```python
        with open(os.path.join(repo_dir, "role2.json"), "rb") as f:
            bundle.update_delegated_targets(f.read(), "role2", "role1")
```
but after a quick check, I found that `role2.json` had `delegations`, but `delegations` itself consists only of empty `keys` dict and `roles` list.
I asked if that was the intention when marking `delegations` as optional [here](https://github.com/theupdateframework/specification/pull/165#issuecomment-855858990).

When I removed `delegations` from `role2.json` I got: 
```
tuf.exceptions.UnsignedMetadataError: New role2 is not signed by role1
```

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: So, do we need to merge 1417 in order to merge this?

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: Thanks for the clarifications!

tuf/ngclient/_internal/metadata_bundle.py
jku: The spec is fairly silent on what to demand from local metadata. But the basic rule is that we cannot demand local metadata to even exist (apart from the demand that one root version must exist).

I don't know what you mean by broken keys, but for some cases:
* local timestamp is signed by keys that are not the correct signing keys according to root: this is uncommon but not an error. The keys have likely been rotated and we must now download the new timestamp and cannot do the rollback check: _version number of the **trusted timestamp metadata file, if any**, MUST be less than or equal to the version number of the new timestamp metadata file_ -- we do not have a trusted timestamp file at this point in time, so we proceed without one
* local timestamp file has been corrupted on disk or tampered with: we do not have a trusted timestamp file at this point so should proceed without one. Logging something might be a good idea but I don't think failing the process makes sense as a trusted timestamp is not required
* the local software installation is broken (e.g. a dependency got uninstalled) so e.g. signature verification fails: This would be nice to handle separately but I'm not currently convinced we can meaningfully separate this case from e.g. an attacker messing with the signatures and feeding bad data to us and the signature verification failing weirdly because of that. Again, if local metadata fails to load, we should proceed without it.

tuf/ngclient/_internal/metadata_bundle.py
jku: > When I removed `delegations` from `role2.json` I got:
> ```
> tuf.exceptions.UnsignedMetadataError: New role2 is not signed by role1
> ```

Did you resign the metadata after modifying? If you didn't, this seems like the correct thing to do -- signature in role2 is no longer valid

I'm not saying there isn't a bug here (and we definitely should have better ways of modifying and resigning the data) but I just don't see the bug... The delegations in the _delegated_ metadata are not processed at all (apart from what happens in `Metadata.from_bytes()`) during `update_delegated_targets()`. 

The delegations in the _delegator_ metadata are processed -- but in that case calling `update_delegated_targets()` on metadata that does not have delegations would be a bug and would result in a ValueError.

tuf/ngclient/_internal/metadata_bundle.py
jku: Personally I consider the implementation of verify_with_threshold in this PR a placeholder: I don't mind it if it's not perfect, and I don't think waiting on the replacement is necessary. 

I suggest we polish #1423 and #1436 as much as we want and use the result here when they are ready, but don't block this PR on those two

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+        if new_delegate.signed.type != "targets":
+            raise exceptions.RepositoryError(
+                f"Expected 'targets', got '{new_delegate.signed.type}'"
+            )
+
+        if not verify_with_threshold(delegator, role_name, new_delegate):
+            raise exceptions.UnsignedMetadataError(
+                f"New {role_name} is not signed by {delegator_name}",
+                new_delegate,
+            )
```
jku: this might be more readable if we called it `new_delegate_name` instead of `role_name` -- so the connection between the name and new_delegate metadata is clear.

Of course, if Metadata object included the name (#1425), this would be even clearer:
```
if not verify_with_threshold(delegator, new_delegate):
    raise exceptions.UnsignedMetadataError(
        f"New {new_delegate.name} is not signed by {delegator.name}",
        new_delegate,
    )
```

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
+# Copyright the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""TUF client bundle-of-metadata
+
+MetadataBundle keeps track of current valid set of metadata for the client,
```
trishankatdatadog: As I've mentioned before, I think it is a fantastic idea to use a state machine to keep track of the update workflow, but I do think MetadataBundle is not the best name for this. Could we please rename to something much more descriptive? Maybe something like Workflow? Not the best, but you get the idea.

tuf/ngclient/README.md
```diff
@@ -0,0 +1,22 @@
+## Next-gen TUF client for Python
+
+This package provides modules for TUF client implementers.
+
+**tuf.ngclient.Updater** is a class that implements the client workflow
+described in the TUF specification (see
+https://theupdateframework.github.io/specification/latest/#detailed-client-workflow)
+
+**tuf.ngclient.FetcherInterface** is an abstract class that client
+implementers can optionally use to integrate with their own
```
trishankatdatadog:
```suggestion
implementers can implement a concrete class of in order to reuse their own
```

tuf/ngclient/README.md
```diff
@@ -0,0 +1,22 @@
...
+network/download infrastructure -- a Requests-based implementation is
```
trishankatdatadog:
```suggestion
networking/download libraries -- a Requests-based implementation is
```

tuf/ngclient/README.md
```diff
@@ -0,0 +1,22 @@
...
+used by default.
+
+This package:
+* Aims to be a clean, easy-to-validate reference client implementation
+  written in modern Python
+* At the same time aims to be the library choice for anyone
```
trishankatdatadog:
```suggestion
* At the same time aims to be the library of choice for anyone
```

tuf/ngclient/README.md
```diff
@@ -0,0 +1,22 @@
...
+  implementing a TUF client in Python: light-weight, easy to integrate
+  and with minimal required dependencies
+* Is still under development but planned to become the default client
```
trishankatdatadog:
```suggestion
* Is still under development but is planned to become the default client
```

tuf/ngclient/README.md
```diff
@@ -0,0 +1,22 @@
...
+  in this code base (as in the older tuf.client will be deprecated in
```
trishankatdatadog:
```suggestion
  in this implementation (i.e., the older `tuf.client` will be deprecated in
```

tuf/ngclient/fetcher.py
```diff
@@ -0,0 +1,41 @@
...
+
+    By providing a concrete implementation of the abstract interface,
+    users of the framework can plug-in their preferred/customized
+    network stack.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def fetch(self, url, required_length):
```
trishankatdatadog, line 23: Could we annotate with typing here to define the exact contract?

tests/test_metadata_bundle.py
```diff
@@ -0,0 +1,124 @@
...
+        #update timestamp after snapshot
+        with self.assertRaises(RuntimeError):
+            bundle.update_timestamp(data["timestamp"])
+
+        # Update delegated targets before targets
+        with self.assertRaises(RuntimeError):
+            bundle.update_delegated_targets(data["role1"], "role1", "targets")
+
+        bundle.update_targets(data["targets"])
+        bundle.update_delegated_targets(data["role1"], "role1", "targets")
```
trishankatdatadog: Very cool. Is it clever enough to figure out which delegation should be checked next, and if not checked, raise an exception? Should be possible, I think.

tests/test_metadata_bundle.py
```diff
@@ -0,0 +1,124 @@
...
+        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
+        data={}
+        for md in ["root", "timestamp", "snapshot", "targets", "role1"]:
+            with open(os.path.join(repo_dir, f"{md}.json"), "rb") as f:
+                data[md] = f.read()
+
+        bundle = MetadataBundle(data["root"])
+
+        # Update timestamp before root is finished
+        with self.assertRaises(RuntimeError):
```
trishankatdatadog: Why RuntimeError? Should we use something specific instead?

tests/test_metadata_bundle.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def test_update_with_invalid_json(self):
+        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
+        data={}
+        for md in ["root", "timestamp", "snapshot", "targets", "role1"]:
+            with open(os.path.join(repo_dir, f"{md}.json"), "rb") as f:
+                data[md] = f.read()
+
+        # root.json not a json file at all
+        with self.assertRaises(exceptions.RepositoryError):
```
trishankatdatadog: Why RepositoryError? Should we use something specific instead?

tuf/ngclient/_internal/download.py
```diff
@@ -0,0 +1,243 @@
...
+      improperly formatted.
+
+      Any other unforeseen runtime exception.
+
+    <Returns>
+      A file object that points to the contents of 'url'.
+    """
+    # Do all of the arguments have the appropriate format?
+    # Raise 'securesystemslib.exceptions.FormatError' if there is a mismatch.
+    sslib_formats.URL_SCHEMA.check_match(url)
```
trishankatdatadog, line 79: In general, I'm not a fan of runtime-type-checking for a dynamically and strongly typed language. It's one thing if we're type-checking untrusted input using a data encoding format that does not inherently describe types (like JSON). But why here in our own internal functions? It just makes for unhelpful reading.

tuf/ngclient/_internal/download.py
```diff
@@ -0,0 +1,243 @@
...
+        url, required_length, fetcher, strict_required_length
+    ) as dl_file:
+        return dl_file.read()
+
+
+def _check_downloaded_length(
+    total_downloaded,
+    required_length,
+    strict_required_length=True,
+    average_download_speed=None,
```
trishankatdatadog, line 160: Why would this ever be None?

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -0,0 +1,187 @@
...
+            response.raise_for_status()
+        except requests.HTTPError as e:
+            response.close()
+            status = e.response.status_code
+            raise exceptions.FetcherHTTPError(str(e), status)
+
+        # Define a generator function to be returned by fetch. This way the
+        # caller of fetch can differentiate between connection and actual data
+        # download and measure download times accordingly.
+        def chunks():
```
trishankatdatadog, line 92: This makes the function a bit too long to follow. Can you think of something simpler?

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: > - local timestamp is signed by keys that are not the correct signing keys according to root: this is uncommon but not an error. The keys have likely been rotated and we must now download the new timestamp and cannot do the rollback check: version number of the trusted timestamp metadata file, if any, MUST be less than or equal to the version number of the new timestamp metadata file -- we do not have a trusted timestamp file at this point in time, so we proceed without one
> - local timestamp file has been corrupted on disk or tampered with: we do not have a trusted timestamp file at this point so should proceed without one. Logging something might be a good idea but I don't think failing the process makes sense as a trusted timestamp is not required
> - the local software installation is broken (e.g. a dependency got uninstalled) so e.g. signature verification fails: This would be nice to handle separately but I'm not currently convinced we can meaningfully separate this case from e.g. an attacker messing with the signatures and feeding bad data to us and the signature verification failing weirdly because of that. Again, if local metadata fails to load, we should proceed without it.

You mean the updating process will start all over again and we will download from version 1 of the `timestamp` right?

tuf/ngclient/_internal/metadata_bundle.py
MVrachev: Yea, I had forgotten to resign the metadata. I did the experiment again. 

> The delegations in the delegated metadata are not processed at all (apart from what happens in Metadata.from_bytes()) during update_delegated_targets().
The delegations in the delegator metadata are processed -- but in that case calling update_delegated_targets() on metadata that does not have delegations would be a bug and would result in a ValueError.

Yes, you are right and I verify that after debugging it.
Probably there is no bug here and it's me spending too much time on optional attributes. :D

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+    keys = {}
+    if isinstance(delegator.signed, Root):
+        keys = delegator.signed.keys
+        role = delegator.signed.roles.get(role_name)
+    elif isinstance(delegator.signed, Targets):
+        if delegator.signed.delegations:
+            keys = delegator.signed.delegations.keys
+            # role names are unique: first match is enough
+            roles = delegator.signed.delegations.roles
+            role = next((r for r in roles if r.name == role_name), None)
```
trishankatdatadog: What does this line do? Looks like magic

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+    else:
+        raise ValueError("Call is valid only on delegator metadata")
```
trishankatdatadog: Why ValueError?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+
+from tuf import exceptions
+from tuf.api.metadata import Metadata, Root, Targets
+from tuf.api.serialization import DeserializationError
+
+logger = logging.getLogger(__name__)
+
+# This is a placeholder until ...
+# TODO issue 1306: implement this in Metadata API
+def verify_with_threshold(
```
trishankatdatadog: Why does this method exist separately as a public module-wide function?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+        Args:
+            root_data: Trusted root metadata as bytes. Note that this metadata
+                will only be verified by itself: it is the source of trust for
+                all metadata in the bundle.
+
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+        self._bundle = {}  # type: Dict[str: Metadata]
```
trishankatdatadog: Is there a reason why these single-underscore variables are not really private, and can be accessed by subclasses?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+        self.reference_time = datetime.utcnow()
```
trishankatdatadog: Should we provide an option for users to [delay what now is, to provide some grace period for expiration](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/config.py#L45)?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+
+        Args:
+            data: unverified new root metadata as bytes
+
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+        if self._root_update_finished:
+            raise RuntimeError(
```
trishankatdatadog: Should we use a more specific error?

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+            if unverified.verify(key):
+                unique_keys.add(key["keyval"]["public"])
+        except Exception as e:  # pylint: disable=broad-except
+            # TODO specify the Exceptions (see issue #1351)
+            logger.info("verify failed: %s", e)
+
+    return len(unique_keys) >= role.threshold
+
+
+class MetadataBundle(abc.Mapping):
```
trishankatdatadog: Very nice API overall. I like it.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+
+        Args:
+            data: unverified new timestamp metadata as bytes
+
+        Raises:
+            RepositoryError: Metadata failed to load or verify. The actual
+                error type and content will contain more details.
+        """
+        if not self._root_update_finished:
+            # root_update_finished() not called
```
trishankatdatadog: This comment looks like you forgot to delete old code, when it is not. Also consider comment getting outdated when you refactor.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+            raise exceptions.RepositoryError(
+                f"Expected 'timestamp', got '{new_timestamp.signed.type}'"
+            )
+
+        if not verify_with_threshold(self.root, "timestamp", new_timestamp):
+            raise exceptions.UnsignedMetadataError(
+                "New timestamp is not signed by root", new_timestamp.signed
+            )
+
+        if self.timestamp is not None:
```
trishankatdatadog: Why would it be None? Not obvious.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+                error type and content will contain more details.
+        """
+
+        if self.timestamp is None:
+            raise RuntimeError("Cannot update snapshot before timestamp")
+        if self.targets is not None:
+            raise RuntimeError("Cannot update snapshot after targets")
+        logger.debug("Updating snapshot")
+
+        meta = self.timestamp.signed.meta["snapshot.json"]
```
trishankatdatadog: Hmm, not related to this PR, but rather to the Metadata API. I don't think we should be indexing things by filenames. There is only one meaningful entry in timestamp, and that's snapshot. Can be hardcoded and doesn't need indexing.

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+            new_snapshot.signed.version
+            != self.timestamp.signed.meta["snapshot.json"].version
+        ):
+            raise exceptions.BadVersionNumberError(
+                f"Expected snapshot version"
+                f"{self.timestamp.signed.meta['snapshot.json'].version},"
+                f"got {new_snapshot.signed.version}"
+            )
+
+        if self.snapshot:
```
trishankatdatadog: Who sets this? Not obvious...

tuf/ngclient/_internal/metadata_bundle.py
```diff
@@ -0,0 +1,461 @@
...
+
+        # Verify against the hashes in snapshot, if any
+        meta = self.snapshot.signed.meta.get(f"{role_name}.json")
+        if meta is None:
+            raise exceptions.RepositoryError(
+                f"Snapshot does not contain information for '{role_name}'"
+            )
+
+        hashes = meta.hashes or {}
+        for algo, stored_hash in hashes.items():
```
trishankatdatadog: This code snippet is repeated often here. DRY?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,575 @@
...
+
+from securesystemslib import exceptions as sslib_exceptions
+from securesystemslib import hash as sslib_hash
+from securesystemslib import util as sslib_util
+
+from tuf import exceptions
+from tuf.ngclient._internal import download, metadata_bundle, requests_fetcher
+from tuf.ngclient.fetcher import FetcherInterface
+
+# Globals
```
trishankatdatadog, line 25: Hmm, I feel these should be in a separate configuration class in a separate module, [like so](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/config.py). Otherwise, harder to make for safe concurrency.

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,575 @@
...
+class Updater:
+    """
+    An implemetation of the TUF client workflow.
+    Provides a public API for integration in client applications.
+    """
+
+    def __init__(
+        self,
+        repository_dir: str,
+        metadata_base_url: str,
```
trishankatdatadog, line 45: Great idea to separate metadata vs targets.

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,575 @@
...
+        raise exceptions.FormatError(
+            repr(child_role_name) + " "
+            'has neither a "paths" nor "path_hash_prefixes".  At least'
+            " one of these attributes must be present."
+        )
+
+    return None
+
+
+def _check_file_length(file_object, trusted_file_length):
```
trishankatdatadog, line 524: Hmm. I get that these are helper functions, but any reason not to fold them into the class?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,575 @@
...
+            filename = f"{version}.{rolename}.json"
+        url = parse.urljoin(self._metadata_base_url, filename)
+        return download.download_bytes(
+            url,
+            length,
+            self._fetcher,
+            strict_required_length=False,
+        )
+
+    def _load_local_metadata(self, rolename: str) -> bytes:
```
trishankatdatadog, line 237: This feels [familiar](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/readers.py#L60) :)

tuf/ngclient/_internal/metadata_bundle.py
trishankatdatadog: Where do you [check](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/repository.py#L209) for revoking timestamp/snapshot to recover from fast-forward attacks?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,575 @@
...
+            length = metainfo.length or DEFAULT_TARGETS_MAX_LENGTH
+            version = None
+            if self._bundle.root.signed.consistent_snapshot:
+                version = metainfo.version
+
+            data = self._download_metadata(role, length, version)
+            self._bundle.update_delegated_targets(data, role, parent_role)
+            self._persist_metadata(role, data)
+
+    def _preorder_depth_first_walk(self, target_filepath) -> Dict:
```
trishankatdatadog, line 329: Stack-based preorder DFS. Nice, but too long. Could we refactor for simplicity and to check for correctness?

trishankatdatadog (commented):
Very nice work, everyone, I love how the new Updater is a few hundred lines, like my PoC! 👏🏽 

A few overall comments:
1. Could use more typing hints.
2. I feel the Metadata API could be more intuitive, but that's a different PR.
3. A few functions are a bit too long. Please consider refactoring.
4. Please double-check against the spec and another implementation (e.g., my PoC) to check whether you missed anything.

Thank you very much again, great work, the ngclient really lives up to its name! Very nice and easy to read like Joshua promised.

tuf/ngclient/_internal/metadata_bundle.py
jku: there is nothing to start over. The whole timestamp loading process is:
* load local timestamp (if it's available, readable and valid)
* load remote timestamp (latest version), validating against the already loaded timestamp if it was loaded. 
* If the remote timestamp was not valid for any reason, abort

tuf/ngclient/_internal/metadata_bundle.py
jku: yeah we can, I've not been totally happy with the name. A lot of the code is indeed about implementing the update workflow so Workflow makes some sense... 

The core idea (or at least the big change) I was starting from is that the currently valid set of metadata must be considered a unit. Single metadata file cannot usually be considered "valid": validity can only be decided for the whole set, and the set must be valid through the update process, at every step. I was trying to figure out a name for this set (that could contain just root metadata or possibly dozens of individual metadata files)

Would be nice to get more opinions, I've been looking at this for way too long to be able to switch viewpoints now...

tuf/ngclient/updater.py
jku: This is issue #1410.

In short: I think the Metadata API should offer these sort of functions (so that there is a "canonical" implementation) and Teodora is currently implementing those in #1437 (so these functions are here as placeholders)

tuf/ngclient/_internal/metadata_bundle.py
jku: We don't: 
* because we are strict about the order of loading different metadata, we cannot ever have timestamp or snapshot loaded at this point
* after this point, we cannot load local timestamp or snapshot if they are no longer signed by the correct (rotated) keys 

So I believe there's just nothing to revoke. The spec does also say _delete the trusted timestamp and snapshot metadata files_. We don't do that either currently:
* the request is a bit odd: the metadata files _by definition_ are not trusted anymore: the keys have been rotated and the old timestamp and snapshot are no longer valid
* There's just no way to load those files into the bundle anymore

We could add the file removal of course... and if we don't we should definitely document why. I've been reluctant to add file removing unless really needed because it feels like removing evidence just as something maybe going wrong

Something else we don't make clear is the difference to current implementation: we have no "current" and "previous" metadata: Currently we store a single copy of each roles metadata (new metadata gets written as soon as it's validated by the bundle). There's probably value in storing some versioned metadata (like every Root) but nothing in the spec seems to require storing "current" and "previous".

tuf/ngclient/updater.py
jku: #1398 : This is still an almost direct copy-paste from current client. We've been avoiding touching it so far since it's so well isolated from the code that we did touch and also because having type annotations will make following the code so much easier.

tuf/ngclient/fetcher.py
jku: We definitely should. There are a few more places where code that we haven't touched much that are still without annotations (side note: this file is the prime example as it's almost complete copy from elsewhere in the code base -- we did not want to import files from the old client to keep everythin clean and easy to follow but we might want to re-evaluate that: maybe we do want to import ngclient code from the older code base, as long as they are compatible).

Anyway: annotations are  #1409: they will be much easier to get right now that mypy support is available.

jku:
Thanks a lot for review

> 1. Could use more typing hints.

Agreed, issue #1409, I suggest we do this after merge

> 2. I feel the Metadata API could be more intuitive, but that's a different PR.

If you have something that you manage to write down in an issue, that would be great... I feel like there is room for improvement but personally the things I am able to find and express are fairly small details. If there are bigger ideas and/or complaints I think now is a great time to look at them: We don't have too much code built yet on top of it but do have some experience already.

> 3. A few functions are a bit too long. Please consider refactoring.

For the bundle: I agree the update_* methods are long but I don't want to split them. There are improvements we can make though (like #1433 removing the type check and making the hash/length checks tighter by implementing in Metadata API).

For the Updater, I think the longer functions are the ones we have not refactored yet, full ack here.

> 4. Please double-check against the spec and another implementation (e.g., my PoC) to check whether you missed anything.

:+1:

tuf/ngclient/updater.py
jku: I fee like we have an issue or at least have planned _something_ here... @sechkova better memory than me?

Totally agreed on the reasoning

joshuagl: Maybe https://github.com/theupdateframework/tuf/issues/1402? Also https://github.com/theupdateframework/tuf/issues/1273 for context.

tests/test_metadata_bundle.py
jku: the bundle makes sure that the new metadata (role "role1" and it's metadata) can be loaded into the bundle and validated -- this means all delegators in the delegation chain must be loaded and must be valid: It is callers responsibility to load metadata in the correct order

tests/test_metadata_bundle.py
jku: I guess it could be more specific: but this is a programming error, not a problem with data. The caller should not be doing this and cannot realistically handle the exception: so making the exception more specific does not help much since it should never be caught.

tests/test_metadata_bundle.py
jku: the repositoryerrors do need a review:  #1312

The core idea with using RepositoryError in general is that we should have one base error that covers all the issues that may come from broken/malicious metadata: this is what clients will really want to handle -- there's nothing they can do about it except show an error message. 

This case results in this chain of exceptions:
```
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
tuf.api.serialization.DeserializationError:
tuf.exceptions.RepositoryError: Failed to load root

```
for debugging purposes that chain is pretty perfect already: making the exception more specific does not help there. Making the error more specific could help users of MetadataBundle though (as they could have separate except-cases for different errors): I'd just like to add the complexity when we know what that use case is... Making exceptions more specific is easy, going the other way after release is practically impossible.

tuf/ngclient/_internal/download.py
jku: Sorry if that was unclear but this is mostly a copy of the download.py in current code base: we absolutely want to refactor this whole file: #1399

My feeling is that after the refactor we should have gotten rid of most of this code: it's needlessly complex and as you say the validation we do should happen in deserialization.

tuf/ngclient/_internal/metadata_bundle.py
jku: it returns the first role in delegations with name `role_name` (or None if no match).

I agree it's not pretty and a bit too clever.
I could do something like this instead:
```
role = None
for delegated_role in roles:
    if delegated_role.name == role_name:
        role = delegated_role
        break
```
but that's 5 lines and _initialize-then-modify_ antipattern: not the end of the world but not perfect either.

the real fix is probably #1426 though (make roles a OrderedDict). Then this code becomes readable and efficient:
`role = roles.get(role_name)`

tuf/ngclient/_internal/metadata_bundle.py
jku: hmm, good question. This is again a case of programming error so the type of error is not terribly important (no-one should be handling this error) but ValueError does seem wrong as arguments are not the problem...

tuf/ngclient/_internal/metadata_bundle.py
jku: #1306: this is something the Metadata API should provide (so there's one canonical implementation): this is a placeholder until then

tuf/ngclient/_internal/metadata_bundle.py
jku: python :shrug: 

attributes with double underscores do get name mangled (by many implementations at least, not sure if the language actually requires it) but even that is not hard to work around IIRC. I think it's appropriate to consider all definitions of "private" in python only polite suggestions: if someone _really_ wants to, they will be able to modify private data.

tuf/ngclient/_internal/metadata_bundle.py
sechkova: Same as @jku's comment below, this is meant to be part of Metadata API https://github.com/theupdateframework/tuf/issues/1410 and then these repetitions will be avoided.

tuf/ngclient/_internal/metadata_bundle.py
jku: hmm, maybe? 

Why is there a grace period? I can think of two potential cases
* clients with really bad clocks 
* buggy repositories that do publish new metadata but do it after old one has expired

This still sounds like something that isn't part of this PR (ngclient is already better than current code since we at least have a reference time). We should file an issue if this is relevant

tuf/ngclient/_internal/metadata_bundle.py
jku: I talked about handling programming errors like this elsewhere

tuf/ngclient/_internal/metadata_bundle.py
jku: yes, makes sense to try to explain that having a valid timestamp is not required to load new timestamp (but if we have one we do additional validation)

tuf/ngclient/_internal/metadata_bundle.py
jku: I agree with this. I get the idea of providing an API that tries to mimic the file format to a point... but this goes far beyond that point.

tuf/ngclient/_internal/metadata_bundle.py
jku: these properties are shorthand for `self._bundle.get("snapshot", None)` (they exist for the top level metadata only):
I like the properties because there is no indexing by name, and top-level metadata needs to be accessed a lot, but it does create another layer of abstraction for the reader ...

tuf/ngclient/_internal/metadata_bundle.py
jku: Yes, agreed: #1437

tuf/ngclient/_internal/metadata_bundle.py
jku: will be handled for real in https://github.com/theupdateframework/tuf/pull/1436/files#r649773993

tuf/ngclient/_internal/metadata_bundle.py
jku: #1443

tuf/ngclient/updater.py
jku: yes that's it, thank you

tuf/ngclient/_internal/metadata_bundle.py
sechkova: Why no one likes my `TrustedMetadata` option :smile: 
This is how the specification refers to metadata files that have been verified by the client and persisted and `MetadataBundle` is implemented as a mapping class rather than a workflow.

trishankatdatadog: I like the name `TrusteMetadata`, but `MetadataBundle` is both a mapping and a workflow class, no?

joshuagl:
There's a lot of great comments here, much of which has been filed as issues to tackle after the code has landed. Are there any blockers to merging this in its current state? Obviously a rebase and an approving review are required.

trishankatdatadog:
> There's a lot of great comments here, much of which has been filed as issues to tackle after the code has landed. Are there any blockers to merging this in its current state? Obviously a rebase and an approving review are required.

Nope, provided all the tests for the old Updater continue to pass for the new Updater, and we make it clear it's not production-ready.

tuf/ngclient/_internal/metadata_bundle.py
jku: Yeah _Trusted_ would be nice in the name but I don't like TrustedMetadata because that sounds like a type of Metadata (a class we already have). The component does contain a lot of the update workflow, but in a way that workflow is just a side effect of keeping track of our collection of trusted metadata...

jku:
Potential work left:
- [x] fix some smaller issues mentioned in review (Teodora has a branch which I just merged here)
- [x] The Big Naming Issue (_MetadataBundle, TrustedMetadata, Workflow_ have been suggested -- all acceptable, none overwhelmingly popular)
- [x] finally, rebase on develop

The test issue is a bit tricky: the existing tests are not blackbox tests (in some places not even close) and we have changed API slightly (e.g. with mirrors config) so getting the old tests to run is not trivial (Teodora tried this so can comment with details if needed). I suggest we merge with the tests we have included, acknowledge that they are inadequate, and start improving them as a priority.

trishankatdatadog:
> The test issue is a bit tricky: the existing tests are not blackbox tests (in some places not even close) and we have changed API slightly (e.g. with mirrors config) so getting the old tests to run is not trivial (Teodora tried this so can comment with details if needed).

About mirrors: is it possible to ignore it in a backwards-compatible way? When only one mirror, just go with the one-repo workflow, but with multiple mirrors, then thrown an error with a message.

jku:
I don't think I'm interested in trying to preserve compatibility like that: 
* It's more code that can break
* the changes required from the client will be minimal (assuming they aren't actually using mirrors)
* I think door should also be kept open for further tweaks to the API #1317

trishankatdatadog:
> I don't think I'm interested in trying to preserve compatibility like that:
> 
> * It's more code that can break
> * the changes required from the client will be minimal (assuming they aren't actually using mirrors)
> * I think door should also be kept open for further tweaks to the API #1317

Fair, no need to keep the old API happy. We are preserving the old codebase for now anyway.

tuf/ngclient/_internal/metadata_bundle.py
joshuagl: There's an issue filed against the spec to [add the notion of slack time](https://github.com/theupdateframework/specification/issues/137). I propose we hold off on adding this feature to the reference implementation until it is part of the spec.

jku:
> * The Big Naming Issue (_MetadataBundle, TrustedMetadata, Workflow_ have been suggested -- all acceptable, none overwhelmingly popular)

Some more brainstorming produced the option `TrustedMetadataSet`:
* it's getting a bit long but should still be acceptable as it isn't a class name that will be needed a lot
* contains "Trusted" which improves on the current name
* does not have the implication of being about a single Metadata (like TrustedMetadata does)

trishankatdatadog:
> Some more brainstorming produced the option `TrustedMetadataSet`:

But it still doesn't hint at any state-keeping about the client update workflow.

jku:
This is true. I think I mentioned this earlier (but it's buried in the nested comments so repeating):
I think the core innovation, if there is any, in this architecture is that the important thing is ensuring that the _collection of Metadata is trusted at all points in time_ (this is in contrast to e.g. current python-tuf that might have metadata in memory that is no longer signed by the delegating metadata, and must handle this situation somehow). Yes, the component does implement most of the steps in the detailed client workflow but the progress is driven by updater: updater decides what is done, the bundle just ensures the collection of metadata is trusted at every step.

Based on that I'd rather emphasize _"trusted collection of metadata"_ than _"client workflow"_, even though both are correct and descriptive.

jku:
Two changes since last comment:
* rebased on develop: I intentionally did not use any new Metadata functionality, this was just a rebase with minimal changes to enable the merge
* Added a new commit that renames MetadataBundle to TrustedMetadataSet (and variable "bundle" to "trusted_set")

From my POV all open issues are now handled or filed as issues (I thought we had one for the tests but filed #1462 since I could not find one).

tuf/ngclient/_internal/metadata_bundle.py
joshuagl: 👍 completely agree with using variable names that make the connection clearer.

tuf/ngclient/_internal/requests_fetcher.py
joshuagl: There's some WIP simplification of this code in #1448

tuf/ngclient/_internal/download.py
joshuagl: Addressed in #1448

tuf/ngclient/_internal/download.py
joshuagl: `average_download_speed` is removed in #1448 (it was a copy/paste from download.py, which #1448 overhauls significantly and ultimately deletes the this copy as the code is moved into other files in ngclient),.

tuf/ngclient/_internal/metadata_bundle.py
joshuagl: > We could add the file removal of course... and if we don't we should definitely document why.

This is a good discussion, we need some action to come from this. Even if it's just an issue capturing this intent.

tuf/ngclient/updater.py
joshuagl: and there's a draft PR to improve things here at #1470

tuf/ngclient/fetcher.py
joshuagl: Annotations added in draft PR #1448

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+
+TrustedMetadataSet keeps track of current valid set of metadata for the client,
+and handles almost every step of the "Detailed client workflow" (
+https://theupdateframework.github.io/specification/latest#detailed-client-workflow)
+in the TUF specification: the remaining steps are related to filesystem and
+network IO which is not handled here.
+
+Loaded metadata can be accessed via the index access with rolename as key
+(trusted_set["root"]) or, in the case of top-level metadata using the helper
+properties (trusted_set.root).
```
joshuagl: Feel free to ignore this suggestion:
```suggestion
TrustedMetadataSet keeps track of the current valid set of metadata for the client,
and handles almost every step of the "Detailed client workflow" (
https://theupdateframework.github.io/specification/latest#detailed-client-workflow)
in the TUF specification: the remaining steps are related to filesystem and
network IO, which are not handled here.

Loaded metadata can be accessed via index access with rolename as key
(trusted_set["root"]) or, in the case of top-level metadata, using the helper
properties (trusted_set.root).
```

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+
+The rules for top-level metadata are
+ * Metadata is loadable only if metadata it depends on is loaded
+ * Metadata is immutable if any metadata depending on it has been loaded
```
joshuagl:
```suggestion
 * Metadata can not be (re-)loaded if any metadata depending on it has been
    loaded
```
Immutable feels like a strong claim?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+>>> # ...
+>>> trusted_set.root_update_finished()
+>>>
+>>> # load local timestamp, then update from remote
+>>> try:
+>>>     with open(timestamp_path, "rb") as f:
+>>>         trusted_set.update_timestamp(f.read())
+>>> except (RepositoryError, OSError):
+>>>     pass # failure to load a local file is ok
+>>>
```
joshuagl, line 42:
```suggestion
```
loading a local copy of timestamp isn't required by the spec, should we simplify this example by removing that step?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+>>> # load local snapshot, then update from remote if needed
+>>> try:
+>>>     with open(snapshot_path, "rb") as f:
+>>>         trusted_set.update_snapshot(f.read())
+>>> except (RepositoryError, OSError):
+>>>     # local snapshot is not valid, load from remote
+>>>     # (RepositoryErrors here stop the update)
+>>>     with download("snapshot", version) as f:
+>>>         trusted_set.update_snapshot(f.read())
```
joshuagl, line 54:
```suggestion
>>> # update snapshot from remote
>>> with download("snapshot", version) as f:
>>>     trusted_set.update_snapshot(f.read())
```
Same suggestion here, does it make sense to simplify the example by only fetching from the remote?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+            )
+
+        if (
+            new_snapshot.signed.version
+            != self.timestamp.signed.meta["snapshot.json"].version
+        ):
+            raise exceptions.BadVersionNumberError(
+                f"Expected snapshot version"
+                f"{self.timestamp.signed.meta['snapshot.json'].version},"
+                f"got {new_snapshot.signed.version}"
```
joshuagl: I think this is missing whitespace in the error message?
```suggestion
                f"Expected snapshot version "
                f"{self.timestamp.signed.meta['snapshot.json'].version}, "
                f"got {new_snapshot.signed.version}"
```

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+
+    def update_delegated_targets(
+        self, data: bytes, role_name: str, delegator_name: str
+    ):
+        """Verifies and loads 'data' as new metadata for target 'role_name'.
+
+        Args:
+            data: unverified new metadata as bytes
+            role_name: The role name of the new metadata
+            delegator_name: The name of the role delegating the new metadata
```
joshuagl:
```suggestion
            delegator_name: The name of the role delegating to the new metadata
```

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+        if not verify_with_threshold(delegator, role_name, new_delegate):
+            raise exceptions.UnsignedMetadataError(
+                f"New {role_name} is not signed by {delegator_name}",
+                new_delegate,
+            )
+
+        if new_delegate.signed.version != meta.version:
+            raise exceptions.BadVersionNumberError(
+                f"Expected {role_name} version"
+                f"{meta.version}, got {new_delegate.signed.version}"
```
joshuagl:
```suggestion
                f"Expected {role_name} version "
                f"{meta.version}, got {new_delegate.signed.version}."
```

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -0,0 +1,464 @@
...
+                if new_fileinfo is None:
+                    raise exceptions.RepositoryError(
+                        f"New snapshot is missing info for '{filename}'"
+                    )
+
+                # Prevent rollback of any metadata versions
+                if new_fileinfo.version < fileinfo.version:
+                    raise exceptions.BadVersionNumberError(
+                        f"Expected {filename} version"
+                        f"{new_fileinfo.version}, got {fileinfo.version}"
```
joshuagl:
```suggestion
                        f"Expected {filename} version "
                        f"{new_fileinfo.version}, got {fileinfo.version}."
```

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,579 @@
...
+MAX_ROOT_ROTATIONS = 32
+MAX_DELEGATIONS = 32
+DEFAULT_ROOT_MAX_LENGTH = 512000  # bytes
+DEFAULT_TIMESTAMP_MAX_LENGTH = 16384  # bytes
+DEFAULT_SNAPSHOT_MAX_LENGTH = 2000000  # bytes
+DEFAULT_TARGETS_MAX_LENGTH = 5000000  # bytes
+
+logger = logging.getLogger(__name__)
+
+# Classes
```
joshuagl:
```suggestion
```
Seems an unnecessary comment?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,579 @@
...
+
+        if fetcher is None:
+            self._fetcher = requests_fetcher.RequestsFetcher()
+        else:
+            self._fetcher = fetcher
+
+    def refresh(self) -> None:
+        """
+        This method downloads, verifies, and loads metadata for the top-level
+        roles in a specific order (root -> timestamp -> snapshot -> targets)
```
joshuagl:
```suggestion
        roles the specified order (root -> timestamp -> snapshot -> targets)
```
?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,579 @@
...
+        self._load_timestamp()
+        self._load_snapshot()
+        self._load_targets("targets", "root")
+
+    def get_one_valid_targetinfo(self, target_path: str) -> Dict:
+        """
+        Returns the target information for a target identified by target_path.
+
+        As a side-effect this method downloads all the metadata it needs to
+        return the target information.
```
joshuagl: All metadata it needs? Or only metadata for delegated targets roles? If this method downloads _all_ metadata, should `refresh()`'s docstring say that `refresh()` " should be called by the client before any target requests"?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,579 @@
...
+        self,
+        targetinfo: Dict,
+        destination_directory: str,
+        target_base_url: Optional[str] = None,
+    ):
+        """
+        Download target specified by 'targetinfo' into 'destination_directory'.
+
+        Args:
+            targetinfo: data received from get_one_valid_targetinfo()
```
joshuagl: 🤔 but `updated_targets()` docstring tells me I can feed the items in the list there to this method.

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,579 @@
...
+                        )
+                        logger.debug(msg)
+                        role_names = []
+                        child_roles_to_visit.append(
+                            (child_role_name, role_name)
+                        )
+                        break
+
+                    if child_role_name is None:
+                        msg = f"Skipping child role {child_role_name}"
```
joshuagl, line 393: AFAICS this will only ever be "Skipping child role `None`"?

tuf/ngclient/updater.py
joshuagl: Draft PR to improve this code in #1463

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,579 @@
...
+                f"visit at most {MAX_DELEGATIONS} delegations.",
+            )
+            logger.debug(msg)
+
+        return {"filepath": target_filepath, "fileinfo": target}
+
+
+def _visit_child_role(child_role: Dict, target_filepath: str) -> str:
+    """
+    <Purpose>
```
joshuagl, line 429: Old style docstring here.

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,579 @@
...
+            raise ValueError(
+                "target_base_url must be set in either download_target() or "
+                "constructor"
+            )
+        if target_base_url is None:
+            target_base_url = self._target_base_url
+        else:
+            target_base_url = _ensure_trailing_slash(target_base_url)
+
+        full_url = parse.urljoin(target_base_url, targetinfo["filepath"])
```
joshuagl, line 208: Per [5.7.3](https://theupdateframework.github.io/specification/v1.0.19/#fetch-target) of the detailed client workflow, if consistent snapshots are enabled for the repository then we should prefix the target's filename with the hash(es) when downloading.

joshuagl (approved):
Apologies for the delay, but I finally gave this a complete review. Great work folks!

I started out by reading the existing discussion threads and checking to see whether all discussion items are resolved. I believe they are, apart from one where @trishankatdatadog [noticed](https://github.com/theupdateframework/tuf/pull/1408#discussion_r648810281) the code does not implement [5.3.11](https://theupdateframework.github.io/specification/v1.0.19/#update-root) which @jku [indicated was intentional](https://github.com/theupdateframework/tuf/pull/1408#discussion_r649133679), but that it needs to be documented _why_ it's not implemented per spec or implemented per spec.

I then did a review of the code, which looks great 👍 . I made only a few minor suggestions on strings/comments/docstrings.

Finally, I did a side-by-side comparison of the [detailed client workflow](https://theupdateframework.github.io/specification/v1.0.19/#detailed-client-workflow) to the new code, which wasn't an enormous task 🎉 – another indication of the solid design here. When doing so I noticed that:
1. When we compare versions for rollback attack checks, the spec says "version number of the [...] metadata file, if any, MUST be less than or equal to the version number of [...]", but the code checks `<`, not `<=`. I don't think this is a major issue, but it's something I was able to notice thanks to the quality of the new implementation. Perhaps worth a comment at least?
2. The current updater implementation does not handle consistent snapshots in the target file downloading (`Updater.download_target()`).
3. Step 5.3.11 is not implemented, as discussed above.

I'm OK with handling all of these as follow-ons, so long as we have issues (or PRs) created before merging this.

tuf/ngclient/_internal/trusted_metadata_set.py
jku: the spec is basically silent about loading anything local (except the trusted root), as far as I can tell: 
* it still expects us to compare new files to "trusted metadata" for rollback protection. I think loading a local copy is the logical way to find "trusted metadata".
* I think there's possibly even a case to be made for loading _expired_ local metadata for purposes of rollback prevention but since the spec explicitly mentions in the root case that expired metadata is fine, and does not mention that in other cases, I didn't seriously consider that. Still this may be a discussion topic ...

tuf/ngclient/updater.py
jku: targetinfo in this case can come from either `get_one_valid_targetinfo()` or `updated_targets()`. I'll update the docstring

btw, updated_targets() annotations are very incorrect, I'll fix those too...

tuf/ngclient/updater.py
jku: True. Teodora is fixing this in #1463 already so not touching here.

tuf/ngclient/updater.py
jku: This gets refactored in #1463

tuf/ngclient/updater.py
jku: oops! Filed #1479

jku:
> When we compare versions for rollback attack checks, the spec says "version number of the [...] metadata file, if any, MUST be less than or equal to the version number of [...]", but the code checks <, not <=. I don't think this is a major issue, but it's something I was able to notice thanks to the quality of the new implementation. Perhaps worth a comment at least?

I believe the checks are actually as specified: it's just that spec defines it as

    if trusted_version <= new_version:
        # all good, continue 

and we check for 

    if new_version < trusted_version:
        # NOT good, raise error

> The current updater implementation does not handle consistent snapshots in the target file downloading (Updater.download_target()).

Good call , filed #1479

> Step 5.3.11 is not implemented, as discussed above.

There was a comment explaining it in `root_update_finished()` (where 5.3.11 logically would happen): I tried to make it clearer.

jku:
changes since last push:
* added two commits to improve docstrings and annotations based on review comments
* rebased to fix conflicts in tuf/exceptions.py (so the commit _exceptions: Make more errors RepositoryErrors_ had minor changes)

jku:
changes since last push:
* added three commits to improve docstrings, error messages and annotations based on review comments
* rebased to fix conflicts in tuf/exceptions.py (so the commit _exceptions: Make more errors RepositoryErrors_ had minor changes)

All discussions fixed here have been marked resolved

tuf/ngclient/_internal/trusted_metadata_set.py
joshuagl: I changed my mind on this while doing the review and should have deleted this comment.

> * it still expects us to compare new files to "trusted metadata" for rollback protection. I think loading a local copy is the logical way to find "trusted metadata".

Exactly this. The spec implies there's loading of snapshot and timestamp to check for rollback attacks and targets to check for freeze attacks. I made a note while reviewing this PR to do _something_ about this for the spec.

You have a good point that expired metadata is still valid as a source of information for protection against rollback attacks.

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,582 @@
...
+        self._load_root()
+        self._load_timestamp()
+        self._load_snapshot()
+        self._load_targets("targets", "root")
+
+    def get_one_valid_targetinfo(self, target_path: str) -> Dict:
+        """
+        Returns the target information for a target identified by target_path.
+
+        As a side-effect this method downloads all the additional (delegate
```
joshuagl:
```suggestion
        As a side-effect this method downloads all the additional (delegated
```

joshuagl:
> > When we compare versions for rollback attack checks, the spec says "version number of the [...] metadata file, if any, MUST be less than or equal to the version number of [...]", but the code checks <, not <=. I don't think this is a major issue, but it's something I was able to notice thanks to the quality of the new implementation. Perhaps worth a comment at least?
> 
> I believe the checks are actually as specified: it's just that spec defines it as
> 
> ```
> if trusted_version <= new_version:
>     # all good, continue 
> ```
> 
> and we check for
> 
> ```
> if new_version < trusted_version:
>     # NOT good, raise error
> ```

🤦 whoops, this LGTM.

>  > The current updater implementation does not handle consistent snapshots in the target file downloading (Updater.download_target()).
> 
> Good call , filed #1479

Thanks.

> > Step 5.3.11 is not implemented, as discussed above.
> 
> There was a comment explaining it in `root_update_finished()` (where 5.3.11 logically would happen): I tried to make it clearer.

Updated comment looks good, thanks for doing that. Given the discussion about trusting expired metadata for rollback attack prevention, we may just want to go ahead and delete the old files anyway?

jku:
> Updated comment looks good, thanks for doing that. Given the discussion about trusting expired metadata for rollback attack prevention, we may just want to go ahead and delete the old files anyway?

The spec 5.3.11 talks about deleting "trusted metadata" -- what it really means is _metadata that is not signed by current root_: We'll never load metadata like that. What we might want to use for rollback prevention is _correctly signed, but expired metadata_

I just don't see the point in deleting the files -- our local files may get mangled by random bit flips, malware, users, cats sitting on keyboards: we just need to accept that they may not be valid.

joshuagl:
> > Updated comment looks good, thanks for doing that. Given the discussion about trusting expired metadata for rollback attack prevention, we may just want to go ahead and delete the old files anyway?
> 
> The spec 5.3.11 talks about deleting "trusted metadata" -- what it really means is _metadata that is not signed by current root_: We'll never load metadata like that. What we might want to use for rollback prevention is _correctly signed, but expired metadata_

I believe you are right, but it took me a bit of time to reason it through. 🤔 
This further points to the need to specify how trusted metadata is loaded, and in what scenarios it remains trusted (as in https://github.com/theupdateframework/taps/pull/128).

Issue #1391, opened 2021-05-14
Metadata API: comprehensive de/serialization testing


jku:
It's really hard to avoid de/serialization issues like #1389 and #1386 

Can we improve the testing so a large number of different json representatations are deserialized (and then serialized), so that things like missing fields and e.g. empty containers are tested -- I'm not even talking about testing broken data (that would be nice but not first priority), just valid data in all possible forms.

This should be much more doable now that we have the smaller objects: testing the "complete API" of small classes is a much more reasonable goal than testing the whole metadata.

I don't have practical advice right now how this can be implemented without a lot of inline dictionaries...

joshuagl:
Is this a place where fuzzing could add some confidence that would be harder to achieve with designed test cases? Cc @sechkova

jku:
pasting a related comment from chat:
> the way we currently write the tests makes seeing what is tested very difficult: the test data is constantly built in the middle of the test so you need to read the whole test very carefully to see if a specific case is handled

parameterized testing could help now that we have the smaller classes. There are test dependencies for that but even plain unittest (on python>3.4) includes SubTest that could be used:
```python
testdata = {
    "with attr": "{'attr': 'value'}",
    "missing attr": "{}",
}
def test_data(self):
    for case, data in testdata.items():
        with self.subTest(case=case):
            # deserialize/serialize data here
```
The value is that A) error message will now include `case="missing attr"` and B) all cases are tested even on sub-test failure. 

An additional decision here is what checks need to be made:
* at least do a round of deserialize + serialize and compare the result
* maybe also compare to an object created with constructor and not deser? This requires more test data

MVrachev:
I started working on tests for that.
The way I imagined is:
1. The tests will focus on testing `truthy cases` with valid data.
2. Each test will test deserialization and serialization.
3. Every class should be tested.
4. Every attribute in each class should be tested against a couple of variations of valid data. 
4.1 If there is a limited number of options for an attribute then test with all valid options.
4.2 If there are no limits on what value an attribute can be (for example 64-character string) then test only for a couple of them.
5. Each time testing an attribute with many valid options, generate random valid data.
For example to generate 64 character string containing lower case letters and digits:
```python
''.join(random.choice(string.ascii_lowercase + string.digits) for i in range(len))
```

jku:
My advice is to not get into detailed tests where different variations of attributes are tested, at least not yet: it's so easy to end up with more and more test code where no-one is really sure what actually gets tested.

[...]

PR #1415, opened 2021-05-21
new API: make sure targets in Targets can be empty

MVrachev:
This change is relevant to the new metadata class Targets.

In the specification, when describing the Targets metadata file format
and more precisely "TARGETPATH" (or targets containing the actual
target files) it's said:

```
It is allowed to have a TARGETS object with no TARGETPATH elements.
This can be used to indicate that no target files are available.
```

If there are no "TARGETPATH" keys for the dictionary "targets", this
would mean that "Targets.targets" is {}.
Make sure we test for that.

See: https://theupdateframework.github.io/specification/latest/#targetpath

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -585,6 +585,13 @@ class TestMetadata(unittest.TestCase):
             targets.signed.targets[filename].to_dict(), fileinfo.to_dict()
         )
 
+        # Test from_dict/to_dict Targets with empty targets.
+        targets_dict = copy.deepcopy(targets.to_dict())
+        targets_dict["signed"]["targets"] = {}
+        tmp_dict = copy.deepcopy(targets_dict["signed"])
+        targets_obj = Targets.from_dict(tmp_dict)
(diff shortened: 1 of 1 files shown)


jku:
> It is allowed to have a TARGETS object with no TARGETPATH elements.

This doesn't mean the spec allows dropping the "targets" key altogether, it just means the value (the array/list) can be empty.

MVrachev:
@jku you are right.
If there is no `TARGETPATH` this would mean that `Targets.targets` will be `{}`.
Then, the purpose of this pr changes to provide a unit test that makes sure we support `Targets.targets` equal to {}. 
I updated the pr and commit description and titles as well as the code.

jku:
I marked this draft but I think I didn't see the latest changes when I did -- so I maybe did the wrong thing accidentally: I'm un-drafting it on that assumption

jku (approved):
I'm hoping #1391 would make tests like this possible without code... but since that's not possible now, this looks fine.

PR #1416, opened 2021-05-25
New API: Comprehensive serialization testing

### LINKED ISSUE #1391: Metadata API: comprehensive de/serialization testing

MVrachev:
Fixes #1391 

The idea of this pr is to separate (de)serialization testing outside
`test_api.py` and make sure we are testing from_dict/to_dict for all
possible valid data for all classes.

As a result of this pr, there will be a single place where
one can have a quick look at what use-cases are we testing for
(de)serialization.

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -1,0 +1,203 @@
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+""" Unit tests testing tuf/api/metadata.py classes
+serialization and deserialization.
+
+"""
+
+import json
+import sys
+import logging
+import unittest
+import copy
+
+from typing import Dict, Callable
+
+from tests import utils
+
+from tuf.api.metadata import (
+    Root,
+    Snapshot,
+    Timestamp,
+    Targets,
+    Key,
+    Role,
+    MetaFile,
+    TargetFile,
+    Delegations,
+    DelegatedRole,
+)
+
+logger = logging.getLogger(__name__)
+
+# DataSet is only here so type hints can be used:
+# It is a dict of name to test dict
+DataSet = Dict[str, str]
+
+# Test runner decorator: Runs the test as a set of N SubTests,
+# (where N is number of items in dataset), feeding the actual test
+# function one test case at a time
+def run_sub_tests_with_dataset(dataset: DataSet):
+    def real_decorator(function: Callable[["TestSerialization", str], None]):
+        def wrapper(test_cls: "TestSerialization"):
+            for case, data in dataset.items():
+                with test_cls.subTest(case=case):
+                    function(test_cls, data)
+        return wrapper
+    return real_decorator
+
+
+class TestSerialization(unittest.TestCase):
+
+    valid_keys: DataSet = {
+        "all": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", \
+            "keyval": {"public": "foo"}}',
+    }
+
+    @run_sub_tests_with_dataset(valid_keys)
+    def test_key_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        key = Key.from_dict("id", copy.copy(case_dict))
+        self.assertDictEqual(case_dict, key.to_dict())
+
+
+    valid_roles: DataSet = {
+        "all": '{"keyids": ["keyid"], "threshold": 3}'
+    }
+
+    @run_sub_tests_with_dataset(valid_roles)
+    def test_role_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        role = Role.from_dict(copy.deepcopy(case_dict))
+        self.assertDictEqual(case_dict, role.to_dict())
+
+
+    valid_roots: DataSet = {
+        "all": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
+            "expires": "2030-01-01T00:00:00Z", "consistent_snapshot": false, \
+            "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": { "targets": {"keyids": ["keyid"], "threshold": 3}} \
+            }',
+        "no consistent_snapshot": '{ "_type": "root", "spec_version": "1.0.0", "version": 1, \
+            "expires": "2030-01-01T00:00:00Z", \
+            "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"} }}, \
+            "roles": { "targets": {"keyids": ["keyid"], "threshold": 3} } \
+            }',
+    }
+
+    @run_sub_tests_with_dataset(valid_roots)
+    def test_root_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        root = Root.from_dict(copy.deepcopy(case_dict))
+        self.assertDictEqual(case_dict, root.to_dict())
+
+    valid_metafiles: DataSet = {
+        "all": '{"hashes": {"sha256" : "abc"}, "length": 12, "version": 1}',
+        "no length": '{"hashes": {"sha256" : "abc"}, "version": 1 }',
+        "no hashes": '{"length": 12, "version": 1}'
+    }
+
+    @run_sub_tests_with_dataset(valid_metafiles)
+    def test_metafile_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        metafile = MetaFile.from_dict(copy.copy(case_dict))
+        self.assertDictEqual(case_dict, metafile.to_dict())
+
+
+    valid_timestamps: DataSet = {
+        "all": '{ "_type": "timestamp", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
+            "meta": {"snapshot.json": {"hashes": {"sha256" : "abc"}, "version": 1}}}'
+    }
+
+    @run_sub_tests_with_dataset(valid_timestamps)
+    def test_timestamp_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        timestamp = Timestamp.from_dict(copy.deepcopy(case_dict))
+        self.assertDictEqual(case_dict, timestamp.to_dict())
+
+
+    valid_snapshots: DataSet = {
+        "all": '{ "_type": "snapshot", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
+            "meta": { "file.txt": { "hashes": {"sha256" : "abc"}, "version": 1 }}}'
+    }
+
+    @run_sub_tests_with_dataset(valid_snapshots)
+    def test_snapshot_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        snapshot = Snapshot.from_dict(copy.deepcopy(case_dict))
+        self.assertDictEqual(case_dict, snapshot.to_dict())
+
+
+    valid_delegated_roles: DataSet = {
+        "no hash prefix attribute":
+            '{"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], \
+            "terminating": false, "threshold": 1}',
+        "no path attribute":
+            '{"keyids": ["keyid"], "name": "a", "terminating": false, \
+            "path_hash_prefixes": ["h1", "h2"], "threshold": 99}',
+        "no hash or path prefix":
+            '{"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3}',
+    }
+
+    @run_sub_tests_with_dataset(valid_delegated_roles)
+    def test_delegated_role_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        deserialized_role = DelegatedRole.from_dict(copy.copy(case_dict))
+        self.assertDictEqual(case_dict, deserialized_role.to_dict())
+
+
+    valid_delegations: DataSet = {
+        "all": '{"keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ {"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3} ]}'
+    }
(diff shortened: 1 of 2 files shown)


jku:
I decided to leave a longer comment in the issue since it's not really a review. The core points are
* we should have the ability to scale, to increase coverage _without always adding more code_
* it should be easy to understand what exactly gets tested

MVrachev:
I updated the pr to use decorators as proposed by @jku in his comment here https://github.com/theupdateframework/tuf/issues/1391#issuecomment-849390669.
While working on it, I wanted to showcase how it will look to hardcode the top-level metadata classes and I ended up changing the whole test module.
As a positive side effect of using decorators, we don't need to create a new temporary directory and copy metadata files to it.

tests/test_metadata_serilization.py
```diff
@@ -0,0 +1,201 @@
...
+class TestSerialization(unittest.TestCase):
+
+    KEY = '{"keytype": "rsa", "scheme": "a", "keyval": {"public": "foo"}}'
+    KEYS = f'"keys": {{"id" : {KEY} }}'
+    HASHES = '"hashes": {"sha256" : "abc"}'
+    SIGNED_COMMON = '"spec_version": "1.0.0", "version": 1, \
+        "expires": "2030-01-01T00:00:00Z"'
+
+    valid_keys = {
+        "all": KEY,
```
jku: no reason not to write the key content here: this is the place where we want to test the key data so all the variants that we want to test should be right here IMO (one case is fine as a start, but it should be defined here).

KEY constant may well be useful in other tests (where we are not specifically interested in testing the key data, we just want something that serializes successfully). The fact that it may be a duplicate of a single case in the key test here is fine in my opinion.

On the actual data for key: I think using bogus data in serialization tests in general is fine (it pretty clearly shows what we are _not_ testing here) but `scheme` is probably something we are going to validate (?) so that should maybe be real data? We can fix that when the validation gets added though, I don't really care.

jku (commented):
I like this a lot more!
* much less code, more data :+1: 
* no setup and teardown: lovely
* the test data for the smaller helper classes looks pretty good (signed classes are the trickiest as expected but even they are close to being readable: we can try to improve that later on)

There are still things I'd like to tweak (we can consider leaving the work for future PRs, but I think this code is going to be a model for a lot of other new test code so it may be worth polishing this one):
* I really dislike referring to test data of _another_ test in the test data: I want to be able to modify a test cases data without worrying about other test cases using that data -- in other words test dataset + test should be an atomic unit that does not affect other tests, if possible. The common shared data should be minimal an clearly marked as shared (like you do with the class 'constants')
* we have a line length limit in our style guide: I think this is a great place to admit that the pirate code is more what you'd call "guidelines" than actual rules. If the test data becomes more readable (or you can avoid using variables in the dataset) I think using longer lines is a good choice here
* the class constants should be chosen carefully: I want to understand what they mean when I read the test case (without seeing the actual constant definition). I'd avoid using ambiguous ones like KEYS and HASHES, and instead use objects (like KEY) or types (like SHA256 string if we need one). SIGNED_COMMON might be an exception to this that can't be avoided
* class constants should not be used when they can be avoided -- see e.g. using KEY when testing key serialization.

jku:
> test dataset + test should be an atomic unit that does not affect other tests, if possible

Just to continue on this: I even considered suggesting this:
```
    @run_sub_tests_with_dataset({
        "all": '{"keyids": ["id"], "threshold": 3}'
     })
     def test_role_serialization(self, test_case_data: str):
         # test code
```
to make it clear that the test data is part of the test, not something others should use -- but this maybe scales badly and moves the decorator name so far from the method name if there are many test cases?

tests/test_metadata_serilization.py
```diff
@@ -0,0 +1,201 @@
...
+
+class TestSerialization(unittest.TestCase):
+
+    KEY = '{"keytype": "rsa", "scheme": "a", "keyval": {"public": "foo"}}'
+    KEYS = f'"keys": {{"id" : {KEY} }}'
+    HASHES = '"hashes": {"sha256" : "abc"}'
+    SIGNED_COMMON = '"spec_version": "1.0.0", "version": 1, \
+        "expires": "2030-01-01T00:00:00Z"'
+
+    valid_keys = {
```
jku: Would it make this more clear if we annotated?
```suggestion
    valid_keys: Dataset = {
```

MVrachev:
> > test dataset + test should be an atomic unit that does not affect other tests, if possible
> 
> Just to continue on this: I even considered suggesting this:
> 
> ```
>     @run_sub_tests_with_dataset({
>         "all": '{"keyids": ["id"], "threshold": 3}'
>      })
>      def test_role_serialization(self, test_case_data: str):
>          # test code
> ```
> 
> to make it clear that the test data is part of the test, not something others should use -- but this maybe scales badly and moves the decorator name so far from the method name if there are many test cases?

The syntax looks necessarily complicated that way.
This will be especially true for the top-level metadata classes requiring bigger strings to be hardcoded.

tests/test_metadata_serilization.py
MVrachev: Most likely most of the bogus data should be changed to something meaningful.
I will use a real scheme here.

MVrachev:
I updated this pr and used as few as possible constants and removed usage of other tests valid data.
Also, in some cases, I used more than 80 characters per line to make it readable.

I rebased again to fix a typo in the test file name.

tests/test_metadata_serialization.py
```diff
@@ -0,0 +1,206 @@
...
+    TargetFile,
+    Delegations,
+    DelegatedRole,
+)
+
+logger = logging.getLogger(__name__)
+
+# DataSet is only here so type hints can be used:
+# It is a dict of name to test dict
+DataSet = Dict[str, Dict[str, Any]]
```
jku: oof, I just suggested using more of this annotation but now that I review the actual content, it doesn't actually make sense...

This type was true in one of my original examples when I still had the test case data as dict... but here (and already in my later examples) the tests case data is a string. In which case having a custom DataSet type isn't useful at all we could just say `Dict[str, str]` instead, right?

(Would be nice to get mypy running on the test files as well so we didn't have to review these details)

jku (commented):
There's  a lot to like I think:
* there's very little test specific code
* The test data for all of the helper classes (MetaFile, Role, TargetFile, Key) looks great I think: expanding those tests will be very easy
* tests and their data are completely independent: you can modify one test case without worrying it will change what gets tested elsewhere
 
The only annoying bit for me is:
* signed-classes test data is still a bit hard to read (because of the combo of multiple lines, constants and the f-string formatting)

I think it's pretty good overall though: this feels like a test set I feel I could start expanding without it becoming unmaintainable. So LGTM after fixing my DataSet stupidity (see code comment). 

Would be great to get a comment from someone who's not been looking at this yet: @sechkova maybe: does this need advice on how the test cases work or something?

tests/test_metadata_serialization.py
MVrachev: I decided to not remove the `DataSet` type because it saved me from splitting an existing line into two. 

I still, updated the definitions for it though.

tests/test_metadata_serialization.py
```diff
@@ -0,0 +1,206 @@
...
+        "keyval": {"public": "foo"}}'
+    SIGNED_COMMON = '"spec_version": "1.0.0", "version": 1, \
+        "expires": "2030-01-01T00:00:00Z"'
+
+    valid_keys: DataSet = {
+        "all": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", \
+            "keyval": {"public": "foo"}}',
+    }
+
+    @run_sub_tests_with_dataset(valid_keys)
```
sechkova, line 58: This turned out to be a very clever decorator which avoids calling `subTest` each time and yet all functions in the class repeat exactly the same code, except for the different metadata class. It feels imbalanced. It was not worth figuring out how this decorator works to only see repeated code everywhere else.

I understand that Key.from_dict is the only function with a different signature which is very annoying and I couldn't come up with a good alternative yet too.

sechkova (commented):
I agree with the general idea and I like the [example in the original issue](https://github.com/theupdateframework/tuf/issues/1391#issuecomment-844836997). What I don't like is that you seem to be diverging from it.

- the "driver" code for all test sub-cases is hidden in the decorator outside the class
- meanwhile there is the same test code repeated for each metadata class
- I understand that the goal was to group the test data and the test case but iterating constants and code through the class does not give me this feeling. 

I realize that this review is pure complaining without any good suggestions which should show that I agree, it is not an easy task.  

If I try to summarize, the implementation could benefit from some better structure:
- common tests driver code 
- tests code
- tests data

jku:
Thanks Teodora, maybe we should still try to improve...

> * the "driver" code for all test sub-cases is hidden in the decorator outside the class

Just to make sure this is not just a misunderstanding because we lack comments: the decorator is intentionally not doing _anything_ that relates to TUF, serialization or even json: the only thing it does is enable the "table testing style" where the test function gets a single test case data at a time from the hashtable of test cases. We could have taken a new test dependency to do this for us but this still felt more light weight.

So a test writer really does not need to understand the decorator further than: _"decorator ensures test function will be called once for each of the values in the 'dataset' dictionary"_: Would a better comment stating this next to the decorator help? Or should we move the decorator to a different file just to make it clear it's not a serialization thing per se?

> * meanwhile there is the same test code repeated for each metadata class

Yeah the three lines obviously could be refactored to a function... That would cut down on lines of code but do you think it would make the tests more readable?

> * I understand that the goal was to group the test data and the test case but iterating constants and code through the class does not give me this feeling.

I hear you. I'm happy with the test data for most of the helper classes but the signed-derivatives are ugly. Some options:
* could declare failure for Signed-classes and use another solution to test them. I'm not convinced they are better options
* could just use more space for test case data and avoid using constants: E.g. Delegations would arguably become easier to read with KEY just inlined (as it wouldn't need to be an f-string anymore at all). In fact, with current test data getting rid of the constants completely would mean a max of +10 lines... and probably less in practice
* could squeeze the test data tighter to account for that by breaking the line length limit even more -- I think this would not be a an issue in this case


> If I try to summarize, the implementation could benefit from some better structure:
> * common tests driver code
> * tests code
> * tests data

I firmly believe in this instance each test case should have it's own data: if the data gets bundled with other data then over time the same data gets used in multiple tests... and then no-one touches the test data anymore because who knows what it will break. That said I'm not against a different structure if that issue is kept in mind.

sechkova:
Since I really don't have better suggestions, maybe simply "beautifying" the "data sets" format will make them look nice and readable.
The decorator already has an explanatory comment but could be:

> feeding the actual test function one ~~data item~~ test case at a time

MVrachev:
I decided to experiment and create commits for both:
1. when we remove the `KEY` constants and some of the `f-strings`
2. when we remove the last one constant `SIGNED_COMMON` and remove all of the `f-strings`.

Which one of the three situations is best:
- with `KEY` and `SIGNED_COMMON`
- without `KEY` with `SIGNED_COMMON`
- no constants and `f-strings`?

jku:
I think I like the last version: sure there's a lot of text but I can just _read_ it and it's literally working json (and not f-strings with "double-curly encoding")... 

I had a closer look at the decorator and I think I probably messed up the type annotations in my examples at some point while making a couple of different versions: @MVrachev can you have a look at last commit in https://github.com/jku/tuf/commits/comprehensive-testing and include it if it seems correct to you

MVrachev:
@sechkova can you have another look and give us your opinion?

sechkova:
Yes, it looks better formatted, I can't say I strongly like putting data right next to tests instead of a separate "data set class" but I also don't insist on changing it as long as you are happy with this approach.

MVrachev:
I rebased, fixed the merge conflict, and squashed all of the commits together.

Additionally, I noticed I can 
1. remove `Metafile` class serialization tests
2. remove the test for  `from_dict/to_dict Targets with empty targets`
3. remove the test for `from_dict/to_dict Targets without delegations`

because I already test for them in `test_metadata_serialization.py` and so I did.

> Yes, it looks better formatted, I can't say I strongly like putting data right next to tests instead of a separate "data set class" but I also don't insist on changing it as long as you are happy with this approach.

I understand your point of view, but I prefer the way it's now, so you can see the test and the related dataset together.

@jku you agree with me or you prefer if I move the hardcoded strings to a separate class?
If you are okay the way it is now, can we merge this, or we want to add something more in this pr?

jku:
> @jku you agree with me or you prefer if I move the hardcoded strings to a separate class?

I agree: moving the test data somewhere else removes the connection between the two, and then the data gets used for some other test as well (and I really think each test should have its own data if its practically possible)

> If you are okay the way it is now, can we merge this, or we want to add something more in this pr?

From my previous comment:
_I had a closer look at the decorator and I think I probably messed up the type annotations in my examples at some point while making a couple of different versions: @MVrachev can you have a look at last commit in https://github.com/jku/tuf/commits/comprehensive-testing and include it if it seems correct to you_

It's just two annotation changes, feel free to squash them in if they look right to you. The commit is this https://github.com/jku/tuf/commit/a03c8c79456d53dae4dd97bd575930a58d1f895e

MVrachev:
> > @jku you agree with me or you prefer if I move the hardcoded strings to a separate class?
> 
> I agree: moving the test data somewhere else removes the connection between the two, and then the data gets used for some other test as well (and I really think each test should have its own data if its practically possible)
> 
> > If you are okay the way it is now, can we merge this, or we want to add something more in this pr?
> 
> From my previous comment:
> _I had a closer look at the decorator and I think I probably messed up the type annotations in my examples at some point while making a couple of different versions: @MVrachev can you have a look at last commit in https://github.com/jku/tuf/commits/comprehensive-testing and include it if it seems correct to you_
> 
> It's just two annotation changes, feel free to squash them in if they look right to you. The commit is this [jku@a03c8c7](https://github.com/jku/tuf/commit/a03c8c79456d53dae4dd97bd575930a58d1f895e)

Sorry, I didn't notice that comment.
Updated the annotations and removed two typing imports for `Any` and `Type`,
I think it should be ready for merging now.

jku (approved):
Thanks! Let's go with this and see what it looks like after we start adding more test cases and new tests.

Issue #1422, opened 2021-05-28
Metadata API: enforce signature uniqueness by keyid


jku:
I don't want to test in verify_signature() that only one signature exists for keyid: this should be enforced by Metadata class.

This requirement is not in the spec but I think it makes sense: if there are multiple signatures for the same keyid, something is wrong and failing early seems reasonable.

this should work:
* store signatures internally in `Dict[str, Signature]`: this enforces uniqueness for API users
* in from_dict(), error out if duplicate keyids are found: this enforces uniqueness in serialization
* when serializing (or otherwise returning the signatures as list), remember to make the output deterministic. Alternatively use OrderedDict in the first place to get determinism

jku:
As trishank mentioned on slack: from python 3.7 Dict itself guarantees insert order so is deterministic -- for our uses OrderedDict still makes sense (and also informs the reader that the order matters)

joshuagl:
> _Changed in version 3.7_: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

from [here](https://docs.python.org/3/library/stdtypes.html#dict). That final part is notable because Python 3.6 is our baseline Python version, per [ADR-0001](https://github.com/theupdateframework/tuf/blob/develop/docs/adr/0001-python-version-3-6-plus.md).

jku:
This is the same issue for the spec btw: https://github.com/theupdateframework/specification/issues/155 (only found it now :) )

MVrachev:
> * in from_dict(), error out if duplicate keyids are found: this enforces uniqueness in serialization

In the `Role` class we are storing `keyids` as a list. Maybe we should store them in a set or it's better to loudly warn about this?

Issue #1351, opened 2021-04-15
Exceptions in metadata API, especially verify()


jku:
I've just tried verifying metadata using the new API. I wanted to do this:
```
if md.verify(key):
    print("yay")
```
but it looks like I'll end up with 

```
from securesystemslib import exceptions as sslib_exceptions
from tuf import exceptions
from tuf.api import serialization

try:
    if md.verify(key):
        print("yay")
except (
    exceptions.Error, 
    sslib_exceptions.FormatError,
    serialization.SerializationError, 
    sslib_exceptions.CryptoError,
    sslib_exceptions.UnsupportedAlgorithmError
):
    # 🤷
    pass        
```
It feels like we could do better.

This issue is about exceptions handling in general but I'm fine if the solutions are case-by-case (like maybe verify specifically just does not need to throw at all). 

Some specific issues:
* It feels like in most cases bleeding these securesystemslib errors through the API is wrong: I don't really want to handle them in my client code and I don't think we should expect client developers to know how to do that
* why are serialization errors defined in a different place than other errors? should serialization errors at least derive from generic exceptions.Error so I could pass all TUF errors with a single line if I want to (like I think makes sense with verify())

jku:
I think a way forward is to just review all case individually and afterwards try to come up with generic rules.

I think there are cases where we do want to throw but do not want to document it: these are "internal  error" cases that should not happen and that users cannot realistically prepare for. These should be documented in code though

The other group of errors here are the ones that are a result of malformed or unexpected metadata: I think these should all fall under one top-level exception so at least clients can handle that by outright refusing to handle the metadata.


* **SerializationError**: 
  * should not require me to import a file I don't otherwise need but maybe that's fine because ...
  * I think this is an internal error of the Serializer (as in we do not know of a case where it should happen: this is very different from e.g. Metadata.from_dict() where SerializationError is what happens if metadata is broken in any way). Maybe we should not list this in verify() documentation
  * Proposal: Remove mentions from Metadata API docs as internal errors

* **Error**:
  * this happens if zero signatures were found for key. Is this an error? I feel like this is just `return false` -- or at least similar to the case where none of the signatures match (in which case we `return false`)
  * this also happens if there are multiple keyids but that would mean spec violation...
  * Proposal: 
    * zero signatures for keyid should result in `return false`
    * we should file an issue if Metadata does not enforce keyid uniqueness.
    * we should consider multiple keyids case in verify() an internal error: throw but don't document?

[...]

PR #1423, opened 2021-05-28
Metadata API: implement sig verification in Key, store id in key

### LINKED ISSUE #1417: Metadata API: implement verify in Key, store id in key
This is a potential API improvement I'm considering (both ideas originally from @sechkova i think):
* Implement verify_signature() in Key class
* Store keyid in Key object

Implementing verify() in Key would make sense because:
* having both `verify()` and `verify_with_threshold()/verify_delegate()` in Metadata is tricky because former verifies a sig on itself, and latter would verifiy threshold of sigs on a delegated metadata)
* verify() is not actually usually needed by the API users -- they are only interested in threshold verification
* Asking key to verify a signature makes sense and makes the SSLib integration more a Key class implementation detail

The only reason against this is that now `sign()` and `verify()` are in different places -- I don't think this is a major issue, and I think the solution maybe to move `sign()` away from Metadata as well (not sure if we can implement a MetadataSigner based on SSLib.Signer?) -- this would make some sense because signing is not needed by client and may require specific dependencies.

Storing keyid with key makes sense since when keys are used (e.g. if we have `Key.verify_signature()`), the id is needed. This means `from_dict()` now has to take an extra argument

jku:
Fixes #1417 

Adding id to Key class simplifies life for API users as usually a key needs its identifier: this is visible in how Root.add_key() becomes simpler and in how `verify_signature()` is now possible to implement in key without an additional id argument.

Moving verification to Key arguably also makes the API cleaner because including both "verify myself" and "verify a delegate with threshold" can look awkward in Metadata, and because the somewhat ugly Securesystemslib integration is now Key class implementation detail (see e.g. call to `format_metadata_to_key()`).

Returning bool on verify failure was found to confuse API users (both me and Teodora misused it the first time) and was arguably not a pythonic way to handle it: Now an exception is raised.

There are two issues in this code that are still unresolved:
* #1351: exceptions story is now better but low-level exceptions still bleed through API
* #1422: Metadata should enforce that signatures are unique per key id

I intend to fix these in the next weeks

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -373,18 +373,18 @@ class TestMetadata(unittest.TestCase):
             # Testing that the workflow of deserializing and serializing
             # a key dictionary doesn't change the content.
             test_key_dict = key_dict.copy()
-            key_obj = Key.from_dict(test_key_dict)
+            key_obj = Key.from_dict("id", test_key_dict)
             self.assertEqual(key_dict, key_obj.to_dict())
             # Test creating an instance without a required attribute.
             for key in key_dict.keys():
                 test_key_dict = key_dict.copy()
                 del test_key_dict[key]
                 with self.assertRaises(KeyError):
-                    Key.from_dict(test_key_dict)
+                    Key.from_dict("id", test_key_dict)
             # Test creating a Key instance with wrong keyval format.
             key_dict["keyval"] = {}
             with self.assertRaises(ValueError):
-                Key.from_dict(key_dict)
+                Key.from_dict("id", key_dict)
 
 
     def test_role_class(self):
@@ -413,7 +413,7 @@ class TestMetadata(unittest.TestCase):
                 test_role_dict = role_dict.copy()
                 del test_role_dict[role_attr]
                 with self.assertRaises(KeyError):
-                    Key.from_dict(test_role_dict)
+                    Key.from_dict("id", test_role_dict)
             # Test creating a Role instance with keyid dublicates.
             # for keyid in role_dict["keyids"]:
             role_dict["keyids"].append(role_dict["keyids"][0])
@@ -433,7 +433,7 @@ class TestMetadata(unittest.TestCase):
 
 
         keyid = root_key2['keyid']
-        key_metadata = Key(root_key2['keytype'], root_key2['scheme'],
+        key_metadata = Key(keyid, root_key2['keytype'], root_key2['scheme'],
             root_key2['keyval'])
 
         # Assert that root does not contain the new key
@@ -441,7 +441,7 @@ class TestMetadata(unittest.TestCase):
         self.assertNotIn(keyid, root.signed.keys)
 
         # Add new root key
-        root.signed.add_key('root', keyid, key_metadata)
+        root.signed.add_key('root', key_metadata)
 
         # Assert that key is added
         self.assertIn(keyid, root.signed.roles['root'].keyids)
@@ -453,7 +453,7 @@ class TestMetadata(unittest.TestCase):
 
         # Try adding the same key again and assert its ignored.
         pre_add_keyid = root.signed.roles['root'].keyids.copy()
-        root.signed.add_key('root', keyid, key_metadata)
+        root.signed.add_key('root', key_metadata)
         self.assertEqual(pre_add_keyid, root.signed.roles['root'].keyids)
 
         # Remove the key
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -81,13 +81,10 @@ class TestMetadata(unittest.TestCase):
         # Load keys into memory
         cls.keystore = {}
         for role in ['delegation', 'snapshot', 'targets', 'timestamp']:
-            cls.keystore[role] = {
-                'private': import_ed25519_privatekey_from_file(
-                        os.path.join(cls.keystore_dir, role + '_key'),
-                        password="password"),
-                'public': import_ed25519_publickey_from_file(
-                        os.path.join(cls.keystore_dir, role + '_key.pub'))
-            }
+            cls.keystore[role] = import_ed25519_privatekey_from_file(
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -417,6 +417,7 @@ class Key:
     """A container class representing the public portion of a Key.
 
     Attributes:
+        id: An identifier string
```
sechkova: "An identifier string" is true from the Key class perspective but I think some hints about it being the digest of  an ambiguous canonical key is needed.

tests/test_api.py
```diff
@@ -162,50 +159,46 @@ def test_read_write_read_compare(self):
 
 
     def test_sign_verify(self):
+        root_path = os.path.join(self.repo_dir, 'metadata', 'root.json')
+        root:Root = Metadata.from_file(root_path).signed
+
+        # Locate the public keys we need from root
+        targets_keyid = next(iter(root.roles["targets"].keyids))
+        targets_key = root.keys[targets_keyid]
```
sechkova, line 167: I noticed now this line in the spec saying

> ... Clients MUST calculate each KEYID to verify this is correct for the associated key ...

Calculating the `KEYID` is now done during `sslib.format_metadata_to_key` but is never compared to the one read from the metadata files. I can't decide is there a benefit in explicitly doing this check. Or is it done somewhere and I am missing it?

sechkova (commented):
As already mentioned in the issue, I am in favour of these changes. 
Left some thoughts/questions as comments.

tuf/api/metadata.py
jku: yeah, makes sense -- I'm not fond of documenting the (probably untrue) idea that we could calculate the id from the key content but I should document that the id must be unique within the keys in this Metadata

tests/test_api.py
jku: * it's not done anywhere
* I've read through the sslib implementation and I don't think the keyid calculation is actually reproducible -- things like securesystemslib.settings.HASH_ALGORITHMS value affect it.
* I think specifying how the unique key must be created in the spec is a mistake: it serves no purpose. The spec should just demand the keyid uniquely identifies the key within this metadata file
* I also think that's how we should handle keyid in python-tuf, even if spec is what it is: IMO we can't do anything else

jku: see also https://github.com/theupdateframework/taps/blob/master/tap12.md

tuf/api/metadata.py
jku:
```
        id: An identifier string that must uniquely identify a key within
            the metadata it is used in. This implementation does not verify
            that the id is the hash of a specific representation of the key.
```
this work for you?

sechkova: works for me :)

tuf/api/metadata.py
```diff
@@ -458,6 +408,51 @@ def to_dict(self) -> Dict[str, Any]:
...
+                metadata.signed,
+            ) from None
+
+        if signed_serializer is None:
+            # pylint: disable=import-outside-toplevel
+            from tuf.api.serialization.json import CanonicalJSONSerializer
+
+            signed_serializer = CanonicalJSONSerializer()
+
+        sslib_key, dummy = sslib_keys.format_metadata_to_key(self.to_dict())
```
sechkova: Given the latest discovery of the correct usage of `sslib_keys.format_metadata_to_key` in this use case, it seems that it needs to receive `default_keyid` so that it does not recalculate the `keyid`.

sechkova: or not using it at all ... :roll_eyes:

tests/test_api.py
sechkova: Thanks for the explanation. Feel like I am convinced now!

tuf/api/metadata.py
jku: yeah good find: I think I'll try dropping the whole call, that function does not look like something I'd like to call

jku:
Changes based on Teodoras review:
* Added a bit of docstring on key id
* Removed the call to `securesystemslib.keys.format_metadata_to_key()`, added `Key.to_securesystemslib_key()` and used that instead

sechkova (approved):

tests/test_api.py
joshuagl: > I noticed now this line in the spec saying
> 
> > ... Clients MUST calculate each KEYID to verify this is correct for the associated key ...
> 
> Calculating the `KEYID` is now done during `sslib.format_metadata_to_key` but is never compared to the one read from the metadata files. I can't decide is there a benefit in explicitly doing this check. Or is it done somewhere and I am missing it?

FYI https://github.com/theupdateframework/tuf/pull/1014, specifically https://github.com/theupdateframework/tuf/commit/9fa92e22865b2ba4e164493564042cf3ca778bee

joshuagl (commented):
I know that it feels a bit repetitive/redundant in the API, but given the consistent use of the term `keyid` in the specification and in the code (i.e. `Key.__init__()` and `Key.from_dict()` both take a `keyid` parameter), would it make sense to rename `Key.id` -> `Key.keyid`?

jku:
> I know that it feels a bit repetitive/redundant in the API, but given the consistent use of the term `keyid` in the specification and in the code (i.e. `Key.__init__()` and `Key.from_dict()` both take a `keyid` parameter), would it make sense to rename `Key.id` -> `Key.keyid`?

I would have called the arguments id as well but that shadows the builtin `id()` which feels wrong... so I guess with the spec-consistency argument that means you have a point: consistent use of keyid would make most sense

joshuagl (approved):
Other than id -> keyid, this LGTM.

jku:
oh forgot to metnion: the only changes in the latest force push are renaming "key.id" to "key.keyid"

PR #1430, opened 2021-06-01
New metadata API: spec_version attribute validation

### LINKED ISSUE #1419: Metadata Attribute research: spec_version
This issue aims to document thoughts about the `spec_version` attribute from the `Signed` class in order
to understand how we use that attribute, what might go wrong with it and how we can validate it.

1. How do we use it?
2. What might go wrong?
3. Rethink how we store it?

MVrachev:
Fixes #1419

Even though version strings like `2.0.0-rc.2` or `1.0.0-beta` are
valid strings in semantic versioning format, in TUF we never needed
to add letters for our specification number.

That's why I validate that: spec_version is a `.` separated string
and when split it has a length of 3 and that each of the
three elements is a number.

Also, I remove one comment about the `version` validation place, because it
seems irrelevant anymore.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -293,7 +293,6 @@ class Signed(metaclass=abc.ABCMeta):
         self.spec_version = spec_version
         self.expires = expires
 
-        # TODO: Should we separate data validation from constructor?
         if version <= 0:
             raise ValueError(f"version must be > 0, got {version}")
         self.version = version
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -37,6 +37,10 @@ from tuf.api.serialization import (
 # and currently, we are above 1000 lines by a small margin.
 # pylint: disable=C0302
 
+# We aim to support SPECIFICATION_VERSION and require the input metadata
+# files to have the same major version (the first number) as ours.
+SPECIFICATION_VERSION = ["1", "0", "19"]
+
 
 class Metadata:
     """A container for signed TUF metadata.
@@ -290,6 +294,16 @@ class Signed(metaclass=abc.ABCMeta):
         expires: datetime,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
+        spec_list = spec_version.split(".")
+        if (
+            len(spec_list) != 3
+            or not all(el.isdigit() for el in spec_list)
+            or spec_list[0] != SPECIFICATION_VERSION[0]
+        ):
+            raise ValueError(
+                f"Unsupported spec_version, got {spec_list}, "
+                f"supported {'.'.join(SPECIFICATION_VERSION)}"
+            )
         self.spec_version = spec_version
         self.expires = expires
 


jku (approved):
I'm approving, just commenting: 
* referring to semver in error message is maybe a bit misleading (when we only accept a small but common subset of semver )
* I'm fine with being more strict than the spec when it's useful, I just don't quite get why we do that here

anyway, those are minor quibbles: LGTM, thanks

MVrachev:
> I'm approving, just commenting:
> 
> * referring to semver in error message is maybe a bit misleading (when we only accept a small but common subset of semver )

Yea, you are right, but if I wanted to be 100% correct, I will have to use more words and make this message longer...
Considering we don't need letters for our spec version, for now, I think this message is okay even if a little misleading. 

> * I'm fine with being more strict than the spec when it's useful, I just don't quite get why we do that here

I think it's better to be more conservative when we do checks than liberal and given that we will use almost the same amount of code if we check that the first element in the string is a number or all of them I decided the latter.
Also, the semver format specifically mentions that all of the three dot-separated strings are positive integers: https://semver.org/#spec-item-2 and if there is a need one could add an additional string to the third one.

MVrachev:
@jku I realized that probably before merging we would want to create some tests actually testing `spec_version` invalid examples?
That applies to `version` and `expiry` even though we haven't made changes to their code.
That way when we finish with the initial discussion for each of the attributes we would know that we have added some testing as well.

jku:
> I think it's better to be more conservative when we do checks than liberal

We're spending time writing code that prevents future spec writers from using something like "2.0.0.beta1" without getting any benefits that I can see -- but it's fine let's move on.

jku:
Oh I forgot: Did you actively decide not to verify the actual version number against version supported by the implementation (I may have missed the discussion)? 

Maybe we do what the old implementation did: raise if major version does not match the supported major version (which I think should probably be defined in `tuf/api/metadata.py` and not `tuf/__init__.py` considering we have essentially two implementations)?

MVrachev:
> Oh I forgot: Did you actively decide not to verify the actual version number against version supported by the implementation (I may have missed the discussion)?
> 
> Maybe we do what the old implementation did: raise if major version does not match the supported major version (which I think should probably be defined in `tuf/api/metadata.py` and not `tuf/__init__.py` considering we have essentially two implementations)?

Great catch!
I updated the pr.

tuf/api/metadata.py
```diff
@@ -343,10 +346,26 @@ def __init__(
...
+            raise exceptions.UnsupportedSpecificationError(
+                f"Unsupported major spec_version, got {spec_list}, "
+                f"supported {supported_ver}"
+            )
+        if spec_list[1] != supported_ver[1]:
+            logger.info(
+                "Different minor version than supported, got %s, supported %s",
+                spec_version,
+                supported_ver,
+            )
```
jku: Changing logging behaviour based on content (when all forms of the content are valid) seems like code we don't absolutely need. Using using six lines to do it also seems overkill -- 1-2 lines really should do it.

How about we log once per initialized Metadata (or signed), and always log the basic things there (type, version, spec version, etc)?

tuf/api/metadata.py
```diff
@@ -343,10 +346,26 @@ def __init__(
...
+        if len(spec_list) != 3 or not all(el.isdigit() for el in spec_list):
+            raise ValueError(
+                f"spec_version must be in semver format, got {spec_version}"
+            )
+        supported_ver = SPECIFICATION_VERSION.split(".")
+        if spec_list[0] != supported_ver[0]:
+            raise exceptions.UnsupportedSpecificationError(
+                f"Unsupported major spec_version, got {spec_list}, "
+                f"supported {supported_ver}"
+            )
```
jku: This isn't a major thing and I get that you use different errors because they exist already but will this bring actual value versus code that raises a ValueError -- I don't see anyone ever handling UnsupportedSpecificationError (especially since it'll be caught in serialization code and a DeserializationError will be thrown)?
 
I think this code could just as well be a single if-clause that makes all the checks and a single raise if any of the checks fail -- I think there's just no need to add many separate error messages for something like spec version.

tuf/api/metadata.py
```diff
@@ -25,7 +26,7 @@
 from securesystemslib.storage import FilesystemBackend, StorageBackendInterface
 from securesystemslib.util import persist_temp_file
 
-from tuf import exceptions
+from tuf import SPECIFICATION_VERSION, exceptions
```
jku: I've mentioned already that I think this is not correct -- we have two separate implementations in our code base and I don't see why this variable would the one thing that they share after sharing almost nothing else.

Using tuf.SPECIFICATION_VERSION also means you end up writing code you don't need: `SPECIFICATION_VERSION.split(".")` every time a signed is initialized. Instead you could define the spec version in metadata.py as a tuple of strings which you an then directly compare (`spec_list[0] != SPEC_VERSION[0]`)

jku (commented):
I'm a bit grumpy about +19 lines of code for something like spec_version but I guess the comments are mostly suggestions -- I'm not seeing bugs or demanding changes.

The SPECIFICATION_VERSION issue I can live with as well, I just don't understand why we would want to do that.

tuf/api/metadata.py
MVrachev: Yea, I decided to use `UnsupportedSpecificationError` mostly because it's a specific error.
You have a point that probably nobody takes notice of `UnsupportedSpecificationError`.
I will merge this check into a single if.

tuf/api/metadata.py
MVrachev: I agree with you, the only reason I reused `tuf. SPECIFICATION_VERSION` is that in the python-tuf slack channel when I brought that question I got a mixed answer.
There were opinions we shouldn't bump the `tuf. SPECIFICATION_VERSION` to `1.0.19` for example, because we haven't confirmed we support it. See the channel for details.

I will add a constant(s) in `tuf.api` and we will discuss should we bump it later.

jku:
code looks great to me (thanks) but I have reservations on the comment, placement and value of SPECIFICATION_VERSION: 
* I don't understand why we would want to specify 1.0.0 as version specifically (I could understand either specifying MAJOR only or specifying the latest specific version we think we're up-to-date with -- specifying 1.0.0 definitely gives the wrong message to me)
* what the comment about "conforming" actually means and does it really need six lines
* why the value couldn't be in metadata.py (which is likely the only place that will need it).

MVrachev:
Updated where I placed the `SPECIFICATION_VERSION` constant, used `1.0.19` as supported, and shorten the comment.

jku (approved):
LGTM, thank you

PR #1432, opened 2021-06-02
Stop using formats.py in metadata.py

### LINKED ISSUE #1384: Metadata API: stop using formats.py

avelichka:
Currently, we have one use of tuf/formats.py in tuf/api/metadata.py.
If we do the conversion of the expires string in metadata.py,
we can keep the two implementations separate.

Fixes #1384

The use of tuf/formats.py is removed from tuf/api/metadata.py and 
the **expires** string conversion to datetime is done directly in metadata.py

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -25,7 +25,7 @@ from securesystemslib.signer import Signature, Signer
 from securesystemslib.storage import FilesystemBackend, StorageBackendInterface
 from securesystemslib.util import persist_temp_file
 
-from tuf import exceptions, formats
+from tuf import exceptions
 from tuf.api.serialization import (
     MetadataDeserializer,
     MetadataSerializer,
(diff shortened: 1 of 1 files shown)


jku (changes requested):
I think this is good. We're now not doing:
* securesystemslib input validation: I think this is fine -- let's let datetime handle it
* error handling: also fine, this ends up a deserialization error

A test to confirm the deserialization error would probably be nice -- but leaving that until #1416 exists probably makes sense, otherwise we just keep adding testing code in unmaintainable ways. 

One request: Can you get rid of the unneeded merge commit? A rebase on develop should work fine.

avelichka:
Thank you for the review and the feedback, @jku! 
I got rid of the merge commit :)

jku (approved):

PR #1435, opened 2021-06-04
Handle exceptions in verify

### LINKED ISSUE #1351: Exceptions in metadata API, especially verify()

jku:
#1417 has been merged so this is good to go.

We want to handle exceptions in verify_signature(): Callers are not going to be interested in the various details of why verification failed (or rather, good error messages are important but no-one is going to write code to handle the different cases). 

This call is typically used on the client: the metadata comes from remote repository, and we're trying to verify it. We might try to handle a specific error situation by saying something specific like "we can't verify this sig because we're missing dependency X"... but this could also just be a mistake in the remote end. I feel like `UnsignedMetadataError` with descriptive message and exception history is probably the best we can do.

Still, discussing each error case here probably makes sense... I'll write a list.

Fixes #1351

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -181,6 +181,11 @@ class TestMetadata(unittest.TestCase):
         with self.assertRaises(exceptions.UnsignedMetadataError):
             snapshot_key.verify_signature(metadata_obj)
 
+        # Test verifying with explicitly set serializer
+        targets_key.verify_signature(metadata_obj, CanonicalJSONSerializer())
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            targets_key.verify_signature(metadata_obj, JSONSerializer())
+
         sslib_signer = SSlibSigner(self.keystore['snapshot'])
         # Append a new signature with the unrelated key and assert that ...
         metadata_obj.sign(sslib_signer, append=True)
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -205,6 +205,25 @@ class TestMetadata(unittest.TestCase):
         with self.assertRaises(exceptions.UnsignedMetadataError):
             targets_key.verify_signature(metadata_obj)
 
+        # Test failure on broken public key data (securesystemslib CryptoError)
+        public = timestamp_key.keyval["public"]
+        timestamp_key.keyval["public"] = "ffff"
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            timestamp_key.verify_signature(metadata_obj)
+        timestamp_key.keyval["public"] = public
+
+        # Test failure with invalid signature (securesystemslib FormatError)
+        sig = metadata_obj.signatures[timestamp_keyid]
+        correct_sig = sig.signature
+        sig.signature = "foo"
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            timestamp_key.verify_signature(metadata_obj)
+
+        # Test failure with valid but incorrect signature
+        sig.signature = "52af76354db3403242e1437b1fbf1c7edc4e66b81dfd63b3026ff681d57e88e11a697cca78061a376a9dd8d7fde5777b14d4e6d8e75f976101cbc61321642f06"
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            timestamp_key.verify_signature(metadata_obj)
+        sig.signature = correct_sig
 
     def test_metadata_base(self):
         # Use of Snapshot is arbitrary, we're just testing the base class features
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -205,6 +205,13 @@ class TestMetadata(unittest.TestCase):
         with self.assertRaises(exceptions.UnsignedMetadataError):
             targets_key.verify_signature(metadata_obj)
 
+        # Test failure on unknown scheme (securesystemslib UnsupportedAlgorithmError)
+        scheme = timestamp_key.scheme
+        timestamp_key.scheme = "foo"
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -453,15 +452,25 @@ def verify_signature(
...
+                    metadata.signed,
+                )
+        except (
+            sslib_exceptions.CryptoError,
+            sslib_exceptions.FormatError,
+            sslib_exceptions.UnsupportedAlgorithmError,
+        ) as e:
             raise exceptions.UnsignedMetadataError(
-                f"Failed to verify {self.id} signature for metadata",
+                f"Failed to verify {self.id} signature",
```
sechkova: Can we use "e" to add these descriptive messages that are promised in the description? 
Or I should assume that this is to be added since this is still a draft?

jku: Well, I said "error message" but mean the whole traceback that includes the securesystemslib error if e.g. key is malformed.

jku:
The error cases that I know of are listed below. All lead to a UnsignedMetadataError in the PR:
* This key does not have a signature at all in the metadata
* UnsupportedAlgorithmError (signature algorithm is not supported)
* signature fails to verify
* CryptoError (key may be malformed)
* FormatError (signature is not hex encoded hash, or key fails a format check)

From a clients perspective I think these are all the same: the error content may be interesting but the client can only regard the signature as invalid and cannot "handle" the situations specifically -- and cannot tell when the error might be a broken client install or broken metadata from repository. I can see how the opposing argument could possibly be made at least for the first two items: but I'd like to see someone actually make those cases before we add separate handling for them.

It's also noteworthy that client code is unlikely to call this directly: they will call the "verify delegate with threshold of keys" function once it exists, and that function will make multiple calls to this one: so forwarding specific errors to client becomes a major pain even if we wanted to do that.

tests/test_api.py
```diff
@@ -200,6 +205,25 @@ def test_sign_verify(self):
...
+        timestamp_key.keyval["public"] = public
+
+        # Test failure with invalid signature (securesystemslib FormatError)
+        sig = metadata_obj.signatures[timestamp_keyid]
+        correct_sig = sig.signature
+        sig.signature = "foo"
+        with self.assertRaises(tuf.exceptions.UnsignedMetadataError):
+            timestamp_key.verify_signature(metadata_obj)
+
+        # Test failure with valid but incorrect signature
```
MVrachev, line 229: Which exception does this test case test?
Also, I am not sure I see a test case when `sslib_keys.verify_signature` will call `UnsupportedAlgorithmError`.
Is this a test case for that?

jku: Signature verification failure is not an exception in Securesystemslib.

UnsupportedAlgorithm question is relevant, I wonder if there was a reason I didn't include it: I'll have to check.

jku: I did not see a reason not to so I added the test... we could still test many more possible cases but AFAIK they will all lead to exceptions that we now handle (and test) so I think this might be good enough

jku:
Since last comment I've added some testing on Martins suggestions, and rebased on develop

joshuagl (approved):
Current `repository_lib` verifies signatures in `_generate_and_write_metadata()` _before_ writing metadata to ensure that the required threshold of signatures is met. I don't know that this is desirable behaviour to replicate moving forward, because it's easy to imagine a scenario where one might want to incrementally sign metadata, but it does point to a possible use-case for more detailed exceptions in verification failures.

That being said, I'm all for adding these later when we know what we need and how we want to use them. It might be as simple as having more complex exceptions derived from `UnsignedMetadataError`?

jku:
> Current `repository_lib` verifies signatures in `_generate_and_write_metadata()` _before_ writing metadata to ensure that the required threshold of signatures is met. I don't know that this is desirable behaviour to replicate moving forward, because it's easy to imagine a scenario where one might want to incrementally sign metadata, but it does point to a possible use-case for more detailed exceptions in verification failures.

Even with a repository tool, I can't imagine what the code could do differently based on the error type... but the possibility does exist of course.

As a side note, there's a lot of functionality that is "adjacent" to writing Metadata (version bumps, signing, choosing the filename (versioned or not), updating related MetaFiles). This sounds like a possible repository library component that is the equivalent of MetadataBundle in the ngclient code: Something that tracks a collection of metadata and makes sure things are written to disk in the correct order while handling all of the related changes.

> That being said, I'm all for adding these later when we know what we need and how we want to use them. It might be as simple as having more complex exceptions derived from `UnsignedMetadataError`?

I believe this is correct.

Issue #1429, opened 2021-05-31
Metadata API: (maybe) Enforce public key content uniqueness in keys


jku:
Root and targets have a `keys` dictionary with `keyid` as key. It would be nice to know for sure that these keys are actually unique keys (and not e.g. the same key content with two different keyids).

One solution would be to validate that the keyid is correct according to spec:

> **KEYID**: ... a hexdigest of the SHA-256 hash of the canonical form of the key.

Having looked at the securesystemslib implementation I don't think we can consider keyid reproducible... what SSLib considers canonical form depends on the settings that were used on the machine that generated the key, and those settings are not stored on the key itself.

The other solution would be to validate key content uniqueness in the keys dict ourselves. This might make sense: we could just not accept a keys dictionary that contained two keys with identical key content -- I don't think there's a non-malicious, non-error case where that would happen.

mnm678:
This would be a nice check to have. I thought about something similar while working on [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md) (see #1084). This would require using a standardized representation of keys, like the modulus and exponent for rsa, as the representation of the key can differ for the same key material.

jku:
I think in the end the exact method of producing the keyids is uninteresting (and the spec probably should not specify it): the one thing that matters is that the keyid must be unique for the key -- or in other words the same public key content must not be present under multiple keyids.

jku:
> This would require using a standardized representation of keys, like the modulus and exponent for rsa, as the representation of the key can differ for the same key material.

I'm really not familiar with this but are you saying the current key representation in tuf allows two seemingly different keys to actually be representations of the same key? I was assuming we could just compare the keytype/scheme/keyval of each key and accept the keys as unique if the combination of  keytype/scheme/keyval is unique, but maybe I was being naive.

mnm678:
The issue is that the keyval can differ for the same key. As an example, PEM formatting supports additional whitespace, optional headers, etc.

mnm678:
That being said, de-duplicating on just the keytype/scheme/keyval would probably still prevent accidental key reuse.

jku:
Thanks. So to recap:
* we could check for key content uniqueness
* it would be a minor usability/sanitycheck feature, to prevent accidentally using the same key with two keyids
* it would not prevent maliciously using the same key in two forms (but if attacker is able to define the key content, they are typically also able to e.g. change thresholds or add new keys so this is not a huge issue)

[...]

Issue #1426, opened 2021-05-31
Metadata API: enforce role name uniqueness in delegations


jku:
The spec does not say anything about role name uniqueness in a delegations object but I believe we cannot safely allow multiple roles with same role name in the roles array of a delegations object. If we did then the roles could have different keyids, and then we would end up in a situation where a metadata may be both a valid delegation and an invalid delegation at the same time, depending on how the role gets chosen and that does not seem like the intention of the design.

I don't think there are real uses cases for non-unique role names in delegations either -- and I assume that this situation is just an unintentional side-effect of making roles an ordered list -- so the spec should possibly clarify that uniqueness is required (although I guess theoretically some complex ordering issue could be solved by multiple roles per role name).

Regardless, metadata API should IMO enforce role name uniqueness on Delegations.roles unless a good case is made against that.

jku:
Potentially we could use `roles: OrderedDict[str, DelegatedRole]` to store the roles (and raise in from_dict() if input contains duplicate role names)?

As trishank mentioned on slack: from python 3.7 onwards Dict itself guarantees insert order so is deterministic -- for our uses OrderedDict still makes sense (and also informs the reader that the order matters)

MVrachev:
I would suggest discussing and solving this issue in the spec repo first so that we make sure we are on the same page with the spec.
If there is a use case where role name duplicates are needed (which I doubt) then I hope they will be shared there.

Otherwise, a good catch, and I agree with the sentiment that from my perspective I see only problems in allowing duplicate role names in delegations.

sechkova:
> I don't think there are real uses cases for non-unique role names in delegations either -- and I assume that this situation is just an unintentional side-effect of making roles an ordered list -- so the spec should possibly clarify that uniqueness is required (although I guess theoretically some complex ordering issue could be solved by multiple roles per role name).

The same is implied by the spec itself it just needs to be written explicitly I guess:

> In order to accommodate prioritized delegations, the "roles" key in the DELEGATIONS object above points to an array of delegated roles, rather than to a hash table.

MVrachev:
I want to work on this and will assign myself.
If somebody else is already working on it, please let me know.

Issue #1425, opened 2021-05-31
Metadata API: Consider if Metadata class should contain role name


jku:
I'm not sure if this is worth the trouble, writing it out anyway...

**Description of issue or feature request**:

I've been looking at our data structures lately, trying to improve their usability... and one theme that comes up is that objects should include their identifier if they have one -- the json file format does not do this because it would be wasteful in a wire format but the same does not apply to python objects. There's a PR for Key already but I think the same logic applies to Metadata itself: Our Metadata objects always have a role name that is available when the Metadata is constructed, the name is immutable during lifetime of the object and often needed with the Metadata object (e.g. verify_delegate(), writing to file)

Quick estimates of what this means to API:
* `Metadata.__init__()`: Add argument `role_name: str`
* `Metadata.from_bytes()`: Add argument `role_name: str`
* `Metadata.from_dict()`: Add argument `role_name: str`
* `Metadata.from_file()`: No change -- the role name can be deciphered from file name. Alternatively the argument could be split to two (to match to_file() below): directory and role name
* `Metadata.to_dict()`: No change
* `Metadata.to_file()`: Hmm... not sure how to do this in a nice way: there's three components to the path: directory, basename and extension. Maybe the argument should be just the directory, the basename is then the role name and extension should be a class variable of Serializer?
* `verify_delegate()`: (once it exists) remove argument role_name -- this is included in metadata already

jku:
So I've done a rough conversion... and I'm not convinced:
* changes are quite invasive to the API: the changes in Metadata itself are documented fairly well above but it also ends up modifying Serialization
* the gains are not as impressive as they seemed at first glance: most times when we deal with rolenames we may not have the metadata object yet
  * e.g. when updating a delegated targets we want to use the rolename for snapshot hashes/length check before we need to parse the metadata bytes
  * same in the updater when figuring out delegations: there's no metadata object yet

Conclusion: Considered adding role name to Metadata: suggestion is to not implement it at this time.

joshuagl:
Thanks for doing the research. If you have your WIP changes somewhere, might be good to link them here?
Shall we close this issue?

jku:
let me rebase before pushing a branch so we see at least the the verify_delegate() improvement...

[...]

PR #1436, opened 2021-06-04
Metadata API: Implement threshold verification

### LINKED ISSUE #1306: Implement verification by a threshold of keys

jku:
#1417 is merged so this is good to go

The delegating Metadata (root or targets) verifies that the delegated metadata is signed by required threshold of keys for the delegated role.

This implementation raises UnsignedMetadataError if threshold is not met. Other possible exceptions are programming errors.

Fixes #1306

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -333,6 +333,55 @@ class TestMetadata(unittest.TestCase):
         )
 
 
+    def test_metadata_verify_delegate(self):
+        root_path = os.path.join(self.repo_dir, 'metadata', 'root.json')
+        root = Metadata.from_file(root_path)
+        snapshot_path = os.path.join(
+                self.repo_dir, 'metadata', 'snapshot.json')
+        snapshot = Metadata.from_file(snapshot_path)
+        targets_path = os.path.join(
+                self.repo_dir, 'metadata', 'targets.json')
+        targets = Metadata.from_file(targets_path)
+        role1_path = os.path.join(
+                self.repo_dir, 'metadata', 'role1.json')
+        role1 = Metadata.from_file(role1_path)
+        role2_path = os.path.join(
+                self.repo_dir, 'metadata', 'role2.json')
+        role2 = Metadata.from_file(role2_path)
+
+        # test the expected delegation tree
+        root.verify_delegate('root', root)
+        root.verify_delegate('snapshot', snapshot)
+        root.verify_delegate('targets', targets)
+        targets.verify_delegate('role1', role1)
+        role1.verify_delegate('role2', role2)
+
+        # only root and targets can verify delegates
+        with self.assertRaises(ValueError):
+            snapshot.verify_delegate('snapshot', snapshot)
+        # verify fails for roles that are not delegated by delegator
+        with self.assertRaises(ValueError):
+            root.verify_delegate('role1', role1)
+        with self.assertRaises(ValueError):
+            targets.verify_delegate('targets', targets)
+
+        # verify fails when delegate content is modified
+        expires = snapshot.signed.expires
+        snapshot.signed.bump_expiration()
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            root.verify_delegate('snapshot', snapshot)
+        snapshot.signed.expires = expires
+
+        # verify fails if roles keys do not sign the metadata
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            root.verify_delegate('timestamp', snapshot)
+
+        # verify fails if threshold of signatures is not reached
+        root.signed.roles['snapshot'].threshold = 2
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            root.verify_delegate('snapshot', snapshot)
+
+        # TODO test successful verify with higher thresholds
 
     def test_key_class(self):
         keys = {
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -51,7 +51,8 @@ from securesystemslib.interface import (
 )
 
 from securesystemslib.signer import (
-    SSlibSigner
+    SSlibSigner,
+    Signature
 )
 
 logger = logging.getLogger(__name__)
@@ -357,7 +358,7 @@ class TestMetadata(unittest.TestCase):
         role1.verify_delegate('role2', role2)
 
         # only root and targets can verify delegates
-        with self.assertRaises(ValueError):
+        with self.assertRaises(TypeError):
             snapshot.verify_delegate('snapshot', snapshot)
         # verify fails for roles that are not delegated by delegator
         with self.assertRaises(ValueError):
@@ -376,12 +377,25 @@ class TestMetadata(unittest.TestCase):
         with self.assertRaises(exceptions.UnsignedMetadataError):
             root.verify_delegate('timestamp', snapshot)
 
+        # Add a key to snapshot role, make sure the new sig fails to verify
+        ts_keyid = next(iter(root.signed.roles["timestamp"].keyids))
+        root.signed.add_key("snapshot", root.signed.keys[ts_keyid])
+        snapshot.signatures[ts_keyid] = Signature(ts_keyid, "ff"*64)
+
+        # verify succeeds if threshold is reached even if some signatures
+        # fail to verify
+        root.verify_delegate('snapshot', snapshot)
+
         # verify fails if threshold of signatures is not reached
         root.signed.roles['snapshot'].threshold = 2
         with self.assertRaises(exceptions.UnsignedMetadataError):
             root.verify_delegate('snapshot', snapshot)
 
-        # TODO test successful verify with higher thresholds
+        # verify succeeds when we correct the new signature and reach the
+        # threshold of 2 keys
+        snapshot.sign(SSlibSigner(self.keystore['timestamp']), append=True)
+        root.verify_delegate('snapshot', snapshot)
+
 
     def test_key_class(self):
         keys = {
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -365,6 +365,9 @@ class TestMetadata(unittest.TestCase):
             root.verify_delegate('role1', role1)
         with self.assertRaises(ValueError):
             targets.verify_delegate('targets', targets)
+        # verify fails when delegator has no delegations
+        with self.assertRaises(ValueError):
+            role2.verify_delegate('role1', role1)
 
         # verify fails when delegate content is modified
         expires = snapshot.signed.expires
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -271,16 +271,16 @@ class Metadata:
 
     def verify_delegate(
         self,
-        role_name: str,
-        delegate: "Metadata",
+        delegated_role: str,
+        delegated_metadata: "Metadata",
         signed_serializer: Optional[SignedSerializer] = None,
-    ):
-        """Verifies that 'delegate' is signed with the required threshold of
-        keys for the delegated role 'role_name'.
+    ) -> None:
+        """Verifies that 'delegated_metadata' is signed with the required
+        threshold of keys for the delegated role 'delegated_role'.
 
         Args:
-            role_name: Name of the delegated role to verify
-            delegate: The Metadata object for the delegated role
+            delegated_role: Name of the delegated role to verify
+            delegated_metadata: The Metadata object for the delegated role
             signed_serializer: Optional; serializer used for delegate
                 serialization. Default is CanonicalJSONSerializer.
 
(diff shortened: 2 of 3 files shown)


sechkova (approved):
Maybe I've become biased after looking at the code in the next-gen client for a while because I couldn't point out any issues :) 

I was searching for a way to avoid adding `role_name` as an argument of `verify_delegate` but I saw you already opened https://github.com/theupdateframework/tuf/issues/1425 about it so +1 from me on that.

tests/test_api.py
```diff
@@ -352,6 +352,56 @@ def test_metadata_timestamp(self):
...
+        root.verify_delegate('targets', targets)
+        targets.verify_delegate('role1', role1)
+        role1.verify_delegate('role2', role2)
+
+        # only root and targets can verify delegates
+        with self.assertRaises(ValueError):
+            snapshot.verify_delegate('snapshot', snapshot)
+        # cannot verify with non-existing role
+        with self.assertRaises(ValueError):
+            root.verify_delegate('role1', role1)
```
MVrachev, line 365: Nit: If you use a role you haven't used before in the test it will be easier to understand.
Even though it's logical you can't delegate trust from `role1` to `role1`.

MVrachev (approved):
I didn't found any issues either.
LGTM!

tuf/api/metadata.py
```diff
@@ -250,59 +250,61 @@ def sign(
...
+            raise ValueError(f"No delegation found for {role_name}")
+
+        # verify that delegate is signed by required threshold of unique keys
+        signing_keys = set()
+        for keyid in role.keyids:
+            key = keys[keyid]
+            try:
+                key.verify_signature(delegate, signed_serializer)
+                # keyids are unique. Try to make sure the public keys are too
+                signing_keys.add(key.keyval["public"])
```
MVrachev: Maybe making sure that we have `keyval["public"]` uniqueness can be done instead in `Root` and `Targets`?

tests/test_api.py
```diff
@@ -359,6 +352,56 @@ def test_metadata_timestamp(self):
...
+        # different delegation keys should fail verification
+        with self.assertRaises(tuf.exceptions.UnsignedMetadataError):
+            root.verify_delegate('timestamp', snapshot)
+
+        # Higher threshold should fail verification
+        root.signed.roles['snapshot'].threshold += 1
+        with self.assertRaises(tuf.exceptions.UnsignedMetadataError):
+            root.verify_delegate('snapshot', snapshot)
+
+        # TODO test higher thresholds
```
MVrachev: Wondering why `higher thresholds` will make a difference?

jku: it's just that we don't have a test here that actually uses a threshold of more than one key successfully. It should work but... it still feels like we're testing the edge case (1 key) when the real problem is N keys.

tuf/api/metadata.py
jku: yeah possible but I'd say it's a separate issue: #1429

tests/test_api.py
jku: The fact that I did not add a test case but just the comment does point to a problem in our testing infra: we probably need to be able to produce new signed metadata more easily in the test itself: this is something we should put some effort into (we could use repository_tool... but I don't think we want to even write the metadata to a file, we just want to be able to generate it in memory and test it)

tests/test_api.py
jku: This tries to verify that root delegates to role1 (and it does not so ValueError). The comment could be better though
`# try verifying a role that is not delegated by the delegator `

MVrachev: Yes, I didn't read it right that it's the root that's calling it.🤦
Still, I agree that the comment can be improved.

tuf/api/metadata.py
```diff
@@ -250,59 +250,61 @@ def sign(
...
+            keys = self.signed.keys
+            role = self.signed.roles.get(role_name, None)
+        elif isinstance(self.signed, Targets):
+            if self.signed.delegations:
+                keys = self.signed.delegations.keys
+                # Assume role names are unique in delegations.roles: #1426
+                roles = self.signed.delegations.roles
+                role = next((r for r in roles if r.name == role_name), None)
+        else:
+            raise ValueError("Call is valid only on delegator metadata")
```
jku: from Trishank: "Why ValueError?"

Valid question as arguments are not the issue here (except if we consider "self" an argument, which it is in a way). 

In any case it's not terribly important as this is a programming error and should not be handled by anyone

jku: I left it a ValueError: I'm not going to create a custom error for something that isn't meant to be caught and can't find a better one in the builtin exceptions.

tuf/api/metadata.py
```diff
@@ -256,6 +256,61 @@ def sign(
...
+        role = None
+        if isinstance(self.signed, Root):
+            keys = self.signed.keys
+            role = self.signed.roles.get(role_name)
+        elif isinstance(self.signed, Targets):
+            if self.signed.delegations:
+                keys = self.signed.delegations.keys
+                # Assume role names are unique in delegations.roles: #1426
+                roles = self.signed.delegations.roles
+                role = next((r for r in roles if r.name == role_name), None)
```
jku: Also from trishank: "What does this line do? Looks like magic"
My response was:
> it returns the first role in delegations with name role_name (or None if no match).
> 
> I agree it's not pretty and a bit too clever.
> I could do something like this instead:
> 
> ```
> role = None
> for delegated_role in roles:
>     if delegated_role.name == role_name:
>         role = delegated_role
>         break
> ```
> 
> but that's 5 lines of _initialize-then-modify_ antipattern: not the end of the world but not perfect either.
> 
> the real fix is probably #1426 though (make roles a OrderedDict). Then this code becomes readable and efficient:
> `role = roles.get(role_name)`

I think I'd leave it as is ... but if others think that's too clever I'm happy to change to the more verbose one

tests/test_api.py
jku: I think the right way forward here is issue #1444

MVrachev:
In our discussion with @jku  in issue #1438 we agreed that we should make sure that the metadata is accepted even if it has a couple of invalid keys, but it accommodates the requirement placed by threshold.
I think I don't see a test for that. Do you plan to add one @jku?

jku:
> In our discussion with @jku in issue #1438 we agreed that we should make sure that the metadata is accepted even if it has a couple of invalid keys, but it accommodates the requirement placed by threshold.
> I think I don't see a test for that. Do you plan to add one @jku?

I think there are two things that we should definitely test:
* test that Key.verify_signature() raises UnsignedMetadataError (or potentially a derived error) if keys are in any way invalid. This  should happen in #1435, I've added the tests that I think make sense there
* test that verify_delegate() handles UnsignedMetadataErrors coming from Key.verify_signature(): catching that error is fine as long as threshold is eventually reached: this should happen in this PR. It does look like there is no test case with failing signatures that should eventually succeed (because threshold is still reached). That would make sense to add (but it does not need to be about invalid keys in particular).

Soo, yes I guess I plan to add one now...

tuf/api/metadata.py
sechkova: Could be `TypeError` since what `isinstance` checks is the type but is it much better than `ValueError` .. I guess not.

tests/test_api.py
jku: I did end up adding tests for threshold of 2.

jku:
rebased on develop, added a commit to add a tests for 
* signature that fails to verify, but overall signatures reach threshold
* threshold that actually contains multiple signatures

This also changes one error type as suggested by teodora

tuf/api/metadata.py
jku: I originally did not pick that because the object type the call is made on (Metadata) is correct ... but I think you're right, it's not perfect but it's better.

MVrachev (approved):
Thanks for adding the tests!
Seems good to me!

tuf/api/metadata.py
jku: I only very recently realised the TUF spec does not require the existence of "public" at all -- the three keys defined in the spec all happen to have it but it isn't required.

So this line needs to change. The spec does require that KEYIDs are unique so `verify_delegate()` will just respect that and will not try to poke into the internals.

jku:
Added a commit that removes the attempt to enforce key contents uniqueness:
* I only very recently realised the TUF spec does not require the existence of "public" key within the Key.keyval dictionary at all -- the three keys defined in the spec all happen to have it but it isn't required.
* We know that the attempt is not protection against maliciously crafted keys anyway: creating multiple representations of the same key is not difficult 

Luckily the spec requires that KEYIDs are unique so verify_delegate() will just respect that and will not try to poke into the internals of the keys.

tuf/api/metadata.py
```diff
@@ -273,6 +276,60 @@ def sign(
...
+        role_name: str,
+        delegate: "Metadata",
+        signed_serializer: Optional[SignedSerializer] = None,
+    ):
+        """Verifies that 'delegate' is signed with the required threshold of
+        keys for the delegated role 'role_name'.
+
+        Args:
+            role_name: Name of the delegated role to verify
+            delegate: The Metadata object for the delegated role
```
joshuagl: My reading of this is that `verify_delegate` takes an identifier for the delegated role (role_name) and an instance of the delegated role (delegate) – I would expect these two named parameters to have similar names.
* `delegated_role` and `delegated_metadata`?
* `role_name` and `role_metadata`? (handily avoiding the delegate, delegated, delegatee, delegator confusion)

tuf/api/metadata.py
joshuagl: Perhaps add a one-liner comment for less familiar readers?
> Find the first role in delegations with name role_name (or None if no match)

joshuagl (commented):
This looks good to me. I wonder if the parameter names in `verify_delegate` could be renamed to more clearly indicate that role_name and delegate are related (an identifier and an instance)?

jku:
I chose the more verbose names: it's easy to lose track of where the keys and thresholds really come from and what they are applied on so more verbose seems reasonable.

jku:
Someone made mypy rules stricter in the mean time :grimacing: so had to rebase and fix the annotation (no changes in existing commits, only in two new commit).

tuf/api/metadata.py
```diff
@@ -266,6 +269,63 @@ def sign(
...
+                role = next(
+                    (r for r in roles if r.name == delegated_role), None
+                )
+        else:
+            raise TypeError("Call is valid only on delegator metadata")
+
+        if role is None:
+            raise ValueError(f"No delegation found for {delegated_role}")
+
+        # verify that delegate is signed by required threshold of unique keys
```
joshuagl: Nit:
```suggestion
        # verify that delegated_metadata is signed by required threshold of unique keys
```

tuf/api/metadata.py
```diff
@@ -266,6 +269,63 @@ def sign(
...
+                threshold of keys for 'role_name'
+        """
+
+        # Find the keys and role in delegator metadata
+        role = None
+        if isinstance(self.signed, Root):
+            keys = self.signed.keys
+            role = self.signed.roles.get(delegated_role)
+        elif isinstance(self.signed, Targets):
+            if self.signed.delegations:
```
joshuagl: Nit: I wonder if this would be slightly clearer if we did something like:
```Python
if not self.signed.delegations:
    raise TypeError("Targets does not delegate to any other roles")
keys = ...
...
```
?

tuf/api/metadata.py
```diff
@@ -266,6 +269,63 @@ def sign(
...
         return signature
 
+    def verify_delegate(
+        self,
+        delegated_role: str,
+        delegated_metadata: "Metadata",
+        signed_serializer: Optional[SignedSerializer] = None,
+    ) -> None:
+        """Verifies that 'delegate' is signed with the required threshold of
+        keys for the delegated role 'role_name'.
```
```suggestion
        """Verifies that 'delegated_metadata' is signed with the required
        threshold of keys for the 'delegated_role'.
```

joshuagl (approved):
I really like how easy this is to follow, nice work.

jku:
Yes, agree with all suggestions. I've combined all the improvements from suggestions into one commit.

The changes since last push are
* use `if self.signed.delegations is None` to make that path clearer as suggested
* Add a test for this specific case (this required modifying the test data)
* tweak comments as suggested

It's minor enough that I don't think this absolutely needs another approval (so I will merge with the one I have) but ... there's not a lot of code that's more critical for TUF so feel free to have another look

MVrachev (approved):
I looked at the new changes and it seems good to me!

joshuagl (approved):
LGTM, thanks!

PR #1437, opened 2021-06-04
Add hash and length verification to MetaFile and TargetFile

### LINKED ISSUE #1361: API : Add hash verification for timestamp/snapshot/targets
Feature request: Would be nice to have hash verification builtin to the API: it's not a massive amount of code but a single call to a function with a descriptive name is a lot better than 6 lines of sslib-calling code...

I can see a couple of options:

```python
Timestamp.verify_snapshot_hashes(data: bytes)
Snapshot.verify_hashes(role_name: str, data: bytes)
```
or, as alternative to both of those, with MetadataInfo Martin is building in #1329 :
```python
MetadataInfo.verify_hashes(data: bytes)
```
Last one is probably the correct place.

I'm not sure what should happen on failure though: Exception probably makes sense but then we probably want to change `Metadata.verify()` to do that as well

sechkova:
that is provided, the better.

Fixes #1361

`MetaFile` and `TargetFile` need exactly the same code for length and hash verification but with different exceptions
for the user API.
- Add a `HashVerificationMixIn` class implementing the common functionality.
- Add public methods to `MetaFile` and `TargetFile`

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -16,10 +16,23 @@ available in the class model.
 
 """
 import abc
+import io
 import tempfile
 from datetime import datetime, timedelta
-from typing import Any, ClassVar, Dict, List, Mapping, Optional, Tuple, Type
+from typing import (
+    Any,
+    BinaryIO,
+    ClassVar,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+)
 
+from securesystemslib import hash as sslib_hash
 from securesystemslib import keys as sslib_keys
 from securesystemslib.signer import Signature, Signer
 from securesystemslib.storage import FilesystemBackend, StorageBackendInterface
@@ -622,7 +635,53 @@ class Root(Signed):
         del self.keys[keyid]
 
 
-class MetaFile:
+class BaseFile:
+    """A base class of MetaFile and TargetFile.
+
+    Encapsulates common static methods for length and hash verification.
+    """
+
+    @staticmethod
+    def _verify_hashes(
+        data: Union[bytes, BinaryIO], expected_hashes: Dict[str, str]
+    ) -> None:
+        """Verifies that the hash of 'data' matches 'expected_hashes'"""
+        is_bytes = isinstance(data, bytes)
+        for algo, exp_hash in expected_hashes.items():
+            if is_bytes:
+                digest_object = sslib_hash.digest(algo)
+                digest_object.update(data)
+            else:
+                # if data is not bytes, assume it is a file object
+                digest_object = sslib_hash.digest_fileobject(data, algo)
+
+            observed_hash = digest_object.hexdigest()
+            if observed_hash != exp_hash:
+                raise exceptions.LengthOrHashMismatchError(
+                    f"Observed hash {observed_hash} does not match"
+                    f"expected hash {exp_hash}"
+                )
+
+    @staticmethod
+    def _verify_length(
+        data: Union[bytes, BinaryIO], expected_length: int
+    ) -> None:
+        """Verifies that the length of 'data' matches 'expected_length'"""
+        if isinstance(data, bytes):
+            observed_length = len(data)
+        else:
+            # if data is not bytes, assume it is a file object
+            data.seek(0, io.SEEK_END)
+            observed_length = data.tell()
+
+        if observed_length != expected_length:
+            raise exceptions.LengthOrHashMismatchError(
+                f"Observed length {observed_length} does not match"
+                f"expected length {expected_length}"
+            )
+
+
+class MetaFile(BaseFile):
     """A container with information about a particular metadata file.
 
     Attributes:
@@ -677,6 +736,22 @@ class MetaFile:
             res_dict["hashes"] = self.hashes
 
         return res_dict
+
+    def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]):
+        """Verifies that the length and hashes of "data" match expected
+            values.
+        Args:
+            data: File object or its content in bytes.
+        Raises:
+            LengthOrHashMismatchError: Calculated length or hashes do not
+                match expected values.
+        """
+        if self.length is not None:
+            self._verify_length(data, self.length)
+
+        # Skip the check in case of an empty dictionary too
+        if self.hashes:
+            self._verify_hashes(data, self.hashes)
 
 
 class Timestamp(Signed):
@@ -905,7 +980,7 @@ class Delegations:
         }
 
 
-class TargetFile:
+class TargetFile(BaseFile):
     """A container with information about a particular target file.
 
     Attributes:
@@ -923,12 +998,6 @@ class TargetFile:
 
     """
 
-    @property
-    def custom(self):
-        if self.unrecognized_fields is None:
-            return None
-        return self.unrecognized_fields.get("custom", None)
-
     def __init__(
         self,
         length: int,
@@ -938,6 +1007,12 @@ class TargetFile:
         self.length = length
         self.hashes = hashes
         self.unrecognized_fields = unrecognized_fields or {}
+
+    @property
+    def custom(self):
+        if self.unrecognized_fields is None:
+            return None
+        return self.unrecognized_fields.get("custom", None)
 
     @classmethod
     def from_dict(cls, target_dict: Dict[str, Any]) -> "TargetFile":
@@ -954,6 +1029,18 @@ class TargetFile:
             "hashes": self.hashes,
             **self.unrecognized_fields,
         }
+
+    def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]):
+        """Verifies that the length and hashes of "data" match expected
+            values.
+        Args:
+            data: File object or its content in bytes.
+        Raises:
+            LengthOrHashMismatchError: Calculated length or hashes do not
+                match expected values.
+        """
+        self._verify_length(data, self.length)
+        self._verify_hashes(data, self.hashes)
 
 
 class Targets(Signed):
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -719,6 +719,13 @@ class MetaFile(BaseFile):
         version = meta_dict.pop("version")
         length = meta_dict.pop("length", None)
         hashes = meta_dict.pop("hashes", None)
+
+        # Do some basic input validation
+        if version <= 0:
+            raise ValueError(f"Metafile version must be > 0, got {version}")
+        if length is not None and length <= 0:
+            raise ValueError(f"Metafile length must be > 0, got {length}")
+
         # All fields left in the meta_dict are unrecognized.
         return cls(version, length, hashes, meta_dict)
 
@@ -1019,6 +1026,13 @@ class TargetFile(BaseFile):
         """Creates TargetFile object from its dict representation."""
         length = target_dict.pop("length")
         hashes = target_dict.pop("hashes")
+
+        # Do some basic validation checks
+        if length <= 0:
+            raise ValueError(f"Targetfile length must be > 0, got {length}")
+        if not hashes:
+            raise ValueError("Missing targetfile hashes")
+
         # All fields left in the target_dict are unrecognized.
         return cls(length, hashes, target_dict)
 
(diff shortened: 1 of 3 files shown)


tuf/api/metadata.py
```diff
@@ -617,7 +630,55 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+    def _verify_length(self, data: Union[bytes, BinaryIO]):
+        """Verifies that the length of 'data' matches an expected value"""
+        if isinstance(data, bytes):
+            observed_length = len(data)
+        else:
+            # if data is not bytes, assume it is a file object
+            data.seek(0, io.SEEK_END)
+            observed_length = data.tell()
+
+        if self.length != observed_length:
```
jku: this uses self.length but it's marked Optional (see other comments about related things though)... not a bug I guess but a bit weird

tuf/api/metadata.py
```diff
@@ -673,6 +734,34 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """Verifies that the hash of "data" matches an expected value.
+        Args:
+            data: File object or its content in bytes.
+        Raises:
+            ValueError: Expected hash value is missing.
+            BadHashError: Calculated hash does not match
+                expected hash.
+        """
+        if self.hashes is None:
+            raise ValueError("Missing expected hashes")
```
jku: _verify_hashes() does this already

tuf/exceptions.py
```diff
@@ -88,8 +88,30 @@ def __repr__(self):
     #     self.__class__.__name__ + '(' + repr(self.expected_hash) + ', ' +
     #     repr(self.observed_hash) + ')')
 
+# TODO: LengthMismatchError fully duplicates DownloadLengthMismatchError
+# except that it is not derived from Download.  Revise and remove duplicates
+# with the progress of the client refactor.
+class LengthMismatchError(Error):
```
jku: I see why you're doing this but let's see if the other questions change the situation at all, maybe we can figure out something else... If I call verify_length_and_hashes() I don't think I would ever want to handle the two errors separately: that implies this maybe should be just a single error.

Let's at least not add the functions: I have never seen the data in TUF custom errors being used for anything, I don't expect anyone to use these either -- so a plain message should be fine

tuf/api/metadata.py
```diff
@@ -617,7 +630,55 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+
+            def get_digest_object(algorithm):
+                digest_object = digest(algorithm)
+                digest_object.update(data)
+                return digest_object
+
+        else:
+            # if data is not bytes, assume it is a file object
+            def get_digest_object(algorithm):
+                return digest_fileobject(data, algorithm)
```
jku: this is clever... but it looks to me like the straight forward implementation (without any internal methods) would not be complex either? Maybe I'm missing something?

jku (commented):
I was left wondering about two things:

do we actually want to raise on missing length/hashes? and related, do we need separate checks at all? I'm asking because it seems that the repository decides whether there are lengths/hashes in the metadata or not, and (assuming the spec doesn't demand them in this situation) it seems to me the client shouldn't really do anything about missing length/hashes. So if client shouldn't do anything, why do we raise?

Sure, I can imagine someone implementing TUF in a way that requires hashes for metadata files... but that's not what the spec says so I don't see why the metadata API should implement that case.

Second thing is, I wonder if you experimented with a common parent class instead of a Mixin? It would probably work better with type annotations (defining a class variable of same name seems like a hack that has some runtime effects as well, I think the proper way to do Mixins with type annotations is Protocols: https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes, but no idea if they're worth the trouble)

tuf/api/metadata.py
sechkova: you mean duplicating the loop below for each case?

tuf/api/metadata.py
sechkova: I had to add one of those checks to silence `mypy` .. but I was planning to revise them :)

tuf/api/metadata.py
jku: I mean something like 
```
is_bytes = isinstance(data, bytes)
for algo, stored_hash in self.hashes.items():
    if is_bytes:
        digest_object = digest(algo)
        digest_object.update(data)
    else:
        digest_object = digest_fileobject(data, algo)
    # check digest_object here
```
I think that's easier to read maybe?

jku: (I have no idea if `isinstance()` can be expensive so left it out of the loop: it's probably insignificant so could be in as well)

sechkova:
> do we actually want to raise on missing length/hashes? and related, do we need separate checks at all? I'm asking because it seems that the repository decides whether there are lengths/hashes in the metadata or not, and (assuming the spec doesn't demand them in this situation) it seems to me the client shouldn't really do anything about missing length/hashes. So if client shouldn't do anything, why do we raise?

Sounds right that the repository makes the decision and the client cannot do anything about it. Based on this I think an API like this should look neater and it'd be easier to use:

```
BaseClass or MixIn:
    _verify_hashes(self, data: Union[bytes, BinaryIO]):
         ....
         raise BadHashError # should we define a new error covering both

     _verify_length(self, data: Union[bytes, BinaryIO]):
          ....
         raise ? # should we define a new error 
       
TargetFile:
def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]):
      self._verify_hashes(data)
      self._verify_length(data)


MetaFile:
def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]):
       if self.length is not None:
           self._verify_length(data)
       if self.hashes is not None:
           self._verify_hashes(data)
```

sechkova:
> Second thing is, I wonder if you experimented with a common parent class instead of a Mixin? It would probably work better with type annotations (defining a class variable of same name seems like a hack that has some runtime effects as well, I think the proper way to do Mixins with type annotations is Protocols: https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes, but no idea if they're worth the trouble)

I agree that I started including too many "hackish" lines of code because of type checking and the mixin. I guess I have to define the BaseClass as an ABC to help mypy? I will try and see if the members being optional in only one of the classes causes troubles again.

jku:
> I agree that I started including too many "hackish" lines of code because of type checking and the mixin. I guess I have to define the BaseClass as an ABC to help mypy? I will try and see if the members being optional in only one of the classes causes troubles again.

On the other hand, I'm sure the Python annotation story still has some holes in it (we're definitely seeing them in python < 3.7): so if annotation is preventing a good design, let's also consider disabling annotation for that bit...

But in this case I think the baseclass would probably work fine. Some options: 
* method in baseclass that checks self.length/self.hashes if they are non-None: this way the child classes don't even need to define methods of their own (but TargetFile must make sure it never exists without length/hashes)
* class method(s) in baseclass that take the length and hashes as argument: then the baseclass is just a way to provide an encapsulated method and does not require any data to exist in child instance

sechkova:
>  Some options:
> 
>     * method in baseclass that checks self.length/self.hashes if they are non-None: this way the child classes don't even need to define methods of their own (but TargetFile must make sure it never exists without length/hashes)

I hesitated because the first option reduces duplicated code shared between the two classes (`MetaFile`, `TargetFile`). On the other hand optional parameters lead to slight differences between the two. Silently ignoring missing length and hashes when calling the base class function from a `TargetFile` object just doesn't feel right (even if we made sure that this should never happen). 

Same for declaring length/hashes in the base class as optional. It feels misleading for `TargetFile`.

>     * class method(s) in baseclass that take the length and hashes as argument: then the baseclass is just a way to provide an encapsulated method and does not require any data to exist in child instance

I think I will try this one which is also close to my initial idea but now the verification methods can be just class/static methods of the base class.

sechkova:
I believe I took into account all comments from the discussion and it should be ready for a review now.

tuf/api/metadata.py
```diff
@@ -655,6 +714,13 @@ def from_dict(cls, meta_dict: Dict[str, Any]) -> "MetaFile":
         version = meta_dict.pop("version")
         length = meta_dict.pop("length", None)
         hashes = meta_dict.pop("hashes", None)
+
+        # Do some basic input validation
+        if version <= 0:
+            raise ValueError(f"Metafile version must be > 0, got {version}")
+        if length is not None and length <= 0:
+            raise ValueError(f"Metafile length must be > 0, got {length}")
```
jku, line 727: I guess this is where we get to decide what to do with input data that we don't have a "natural" check for: length and version get validated "enough" like this IMO but hashes could still be anything...

For the values we could run 
```
if not all(isinstance(h, str) for h in hashes.values()):
    raise ValueError("Invalid hash value")
```
But maybe we don't need to? Is it enough that the hash verification will fail when it's tried on completely bogus hash value?

For the dict keys: Can we make SSLib check that the algorithms are known to SSLib?

tuf/api/metadata.py
```diff
@@ -934,11 +1010,24 @@ def __init__(
         self.hashes = hashes
         self.unrecognized_fields = unrecognized_fields or {}
 
+    @property
+    def custom(self):
+        if self.unrecognized_fields is None:
+            return None
+        return self.unrecognized_fields.get("custom", None)
+
```
jku, line 1023: is this move some style guide thing?

jku (commented):
I like it. I think you could have squeezed the exceptions to a line shorter but that's so minor I'm not asking to change that.

So the only real question is about validation: since you made the mistake of adding some, I'm of course asking for more... I'm fine with filing an issue on that though, and getting this merged without solving hashes validation.

tuf/api/metadata.py
sechkova: It was pushed by mistake after reformatting my commits but then I thought this _is_ the expected order of class methods anyway ...

sechkova: I don't think it is formally stated in a guideline though

tuf/api/metadata.py
sechkova: Yeah ... I avoided intentionally the downward validation spiral here ... 

Hash values: there is even a regex about correct hash value in formats.py now but I wasn't sure if it isn't an overkill.
Dict keys: Probably we can pre-define somewhere supported algorithms by sslib because now the check is done runtime during hash calculation.

I will file an issue about it, especially hash values.

sechkova: Here it is: #1441

jku:
there's a conflict here with my changes to verify_signature()... In particular I ended making the import `from securesystemslib import keys as sslib_keys`.
I don't have strong opinions about this (and think there's probably no "right way" to import) so either way works, but I felt that importing individual functions is less useful: it's harder in the code to notice where actual library calls are made.

Otherwise LGTM

sechkova:
Rebased on develop and changed the import to `sslib_hashes` to match `sslib_keys`.

tuf/api/metadata.py
```diff
@@ -939,11 +1015,24 @@ def __init__(
...
     def from_dict(cls, target_dict: Dict[str, Any]) -> "TargetFile":
         """Creates TargetFile object from its dict representation."""
         length = target_dict.pop("length")
         hashes = target_dict.pop("hashes")
+
+        # Do some basic validation checks
+        if length <= 0:
+            raise ValueError(f"Targetfile length must be > 0, got {length}")
+        if not hashes:
+            raise ValueError("Missing targetfile hashes")
```
jku, line 1034: this is actually a bit misleading as None is already handled (pop() will raise), so this actually handles the "empty hashes dict" case... but since the end result is correct and since we have an issue on properly validating hashes, let's just go with this.

jku (approved):

tuf/api/metadata.py
sechkova: It was meant to handle existing but empty dictionary:
`hashes: {}`

jku:
So I approved this already but I'm now thinking if we should handle the unsupported algorithm and format error cases in this PR as well... I guess a separate issue  works too?

Basically as discussed in #1438 and #1441: I don't think we should try to deeply validate e.g. our hash algorithms at deserialize-time. This means when they are used (in this PR), we need to handle every error, and turn the errors into the hash mismatch errors

jku:
I'm making an executive decision and merging, work continues in:
* #1451 validation
* #1452 exception handling

PR #1445, opened 2021-06-14
Metadata API: snapshot_meta property in Timestamp

### LINKED ISSUE #1443: Metadata API: Timestamp should provide easy access to snapshot MetaFile
Quoth @trishankatdatadog
> I don't think we should be indexing things by filenames. There is only one meaningful entry in timestamp, and that's snapshot. Can be hardcoded and doesn't need indexing.

that's in context of accessing the snapshot metafile object in timestamp with `snapshot_meta = timestamp.signed.meta["snapshot.json"]`. I agree and think we should have at least a helper property that allows accessing without string lookups: `snapshot_meta = timestamp.signed.snapshot_meta`.

In fact I might suggest going further and removing the meta dictionary from the Timestamp object and just having snapshot_meta attribute instead: yes, it moves API further from file format but it's safe and easy to understand.

MVrachev:
Fixes #1443 

In Timestamp, the only valid "meta" value is the dictionary representing
meta information for the snapshot file.
That's why we want to provide an access to it without changing the
timestamp representation through the Timestamp class.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -343,11 +343,11 @@ class TestMetadata(unittest.TestCase):
         fileinfo = MetaFile(2, 520, hashes)
 
         self.assertNotEqual(
-            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+            timestamp.signed.snapshot_meta.to_dict(), fileinfo.to_dict()
         )
         timestamp.signed.update(fileinfo)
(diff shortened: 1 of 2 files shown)


MVrachev:
Ohh.. I closed it by accident.
Can you reopen it @jku, because it seems I don't have the permissions?

PR #1446, opened 2021-06-14
Metadata API: change meta type in Timestamp

### LINKED ISSUE #1443: Metadata API: Timestamp should provide easy access to snapshot MetaFile

MVrachev:
Fixes #1443 

In Timestamp, the only valid "meta" value is the dictionary representing
meta information for the snapshot file. This makes the API unnecessarily
complicated and requires validation that only information about snapshot
is available inside "meta".
Together with the python-tuf maintainers we decided that snapshot meta
information will not be represented by a "meta" dictionary but instead
by a MetaFile instance and with this it will diverge from the
specification.
This decision is coherent with ADR9 and the rationale
behind it is to provide easier, safer, and direct access to the
snapshot meta information.

**History of modifications made on the pr**:
- Proposing `snapshot_meta` as a property with a setter and getter.
- Adding annotations
- Suggestion to remove the `update` function for `Timestamp`, will be handled in https://github.com/theupdateframework/tuf/issues/1230
- Documenting that `snapshot_meta as the "API for interacting with snapshot meta information"`
and that `meta is an implementation detail that mimics the file format`.
- Decided with the python-tuf maintainers that we will change the approach to this problem
and we will change the `timestamp.meta` type to `MetaFile`.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -350,11 +350,11 @@ class TestMetadata(unittest.TestCase):
         fileinfo = MetaFile(2, 520, hashes)
 
         self.assertNotEqual(
-            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+            timestamp.signed.snapshot_meta.to_dict(), fileinfo.to_dict()
         )
         timestamp.signed.update(fileinfo)
         self.assertEqual(
-            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+            timestamp.signed.snapshot_meta.to_dict(), fileinfo.to_dict()
         )
 
 
@@ -563,7 +563,7 @@ class TestMetadata(unittest.TestCase):
         timestamp_path = os.path.join(
             self.repo_dir, 'metadata', 'timestamp.json')
         timestamp = Metadata[Timestamp].from_file(timestamp_path)
-        snapshot_metafile = timestamp.signed.meta["snapshot.json"]
+        snapshot_metafile = timestamp.signed.snapshot_meta
 
         snapshot_path = os.path.join(
             self.repo_dir, 'metadata', 'snapshot.json')
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -923,11 +923,11 @@ class MetaFile(BaseFile):
 class Timestamp(Signed):
     """A container for the signed part of timestamp metadata.
 
-    Timestamp contains information about the snapshot Metadata file.
+    TUF file format uses a dictionary to contain the snapshot information:
+    this is not the case with Timestamp.snapshot_meta which is a MetaFile.
 
     Attributes:
-        meta: A dictionary of filenames to MetaFiles. The only valid key value
-            is the snapshot filename, as defined by the specification.
+        snapshot_meta: MetaFile instance with the snapshot meta information.
     """
 
     _signed_type = "timestamp"
@@ -937,33 +937,31 @@ class Timestamp(Signed):
         version: int,
         spec_version: str,
         expires: datetime,
-        meta: Dict[str, MetaFile],
+        snapshot_meta: MetaFile,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
         super().__init__(version, spec_version, expires, unrecognized_fields)
-        self.meta = meta
+        self.snapshot_meta = snapshot_meta
 
     @classmethod
     def from_dict(cls, signed_dict: Dict[str, Any]) -> "Timestamp":
         """Creates Timestamp object from its dict representation."""
         common_args = cls._common_fields_from_dict(signed_dict)
         meta_dict = signed_dict.pop("meta")
-        meta = {"snapshot.json": MetaFile.from_dict(meta_dict["snapshot.json"])}
+        snapshot_meta = MetaFile.from_dict(meta_dict["snapshot.json"])
         # All fields left in the timestamp_dict are unrecognized.
-        return cls(*common_args, meta, signed_dict)
+        return cls(*common_args, snapshot_meta, signed_dict)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self."""
         res_dict = self._common_fields_to_dict()
-        res_dict["meta"] = {
-            "snapshot.json": self.meta["snapshot.json"].to_dict()
-        }
+        res_dict["meta"] = {"snapshot.json": self.snapshot_meta.to_dict()}
         return res_dict
 
     # Modification.
     def update(self, snapshot_meta: MetaFile) -> None:
-        """Assigns passed info about snapshot metadata to meta dict."""
-        self.meta["snapshot.json"] = snapshot_meta
+        """Assigns passed info about snapshot metadata."""
+        self.snapshot_meta = snapshot_meta
 
 
 class Snapshot(Signed):
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -233,13 +233,13 @@ class TrustedMetadataSet(abc.Mapping):
                 )
             # Prevent rolling back snapshot version
             if (
-                new_timestamp.signed.meta["snapshot.json"].version
-                < self.timestamp.signed.meta["snapshot.json"].version
+                new_timestamp.signed.snapshot_meta.version
+                < self.timestamp.signed.snapshot_meta.version
             ):
                 raise exceptions.ReplayedMetadataError(
                     "snapshot",
-                    new_timestamp.signed.meta["snapshot.json"].version,
-                    self.timestamp.signed.meta["snapshot.json"].version,
+                    new_timestamp.signed.snapshot_meta.version,
+                    self.timestamp.signed.snapshot_meta.version,
                 )
 
         # expiry not checked to allow old timestamp to be used for rollback
@@ -286,11 +286,11 @@ class TrustedMetadataSet(abc.Mapping):
         # Snapshot cannot be loaded if final timestamp is expired
         self._check_final_timestamp()
 
-        meta = self.timestamp.signed.meta["snapshot.json"]
+        snapshot_meta = self.timestamp.signed.snapshot_meta
 
         # Verify against the hashes in timestamp, if any
         try:
-            meta.verify_length_and_hashes(data)
+            snapshot_meta.verify_length_and_hashes(data)
         except exceptions.LengthOrHashMismatchError as e:
             raise exceptions.RepositoryError(
                 "Snapshot length or hashes do not match"
@@ -345,14 +345,10 @@ class TrustedMetadataSet(abc.Mapping):
         assert self.timestamp is not None  # nosec
         if self.snapshot.signed.is_expired(self.reference_time):
             raise exceptions.ExpiredMetadataError("snapshot.json is expired")
-
-        if (
-            self.snapshot.signed.version
-            != self.timestamp.signed.meta["snapshot.json"].version
-        ):
+        snapshot_meta = self.timestamp.signed.snapshot_meta
+        if self.snapshot.signed.version != snapshot_meta.version:
             raise exceptions.BadVersionNumberError(
-                f"Expected snapshot version "
-                f"{self.timestamp.signed.meta['snapshot.json'].version}, "
+                f"Expected snapshot version {snapshot_meta.version}, "
                 f"got {self.snapshot.signed.version}"
             )
 
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -352,11 +352,11 @@ class Updater:
             logger.debug("Local snapshot not valid as final: %s", e)
 
             assert self._trusted_set.timestamp is not None  # nosec
-            metainfo = self._trusted_set.timestamp.signed.meta["snapshot.json"]
-            length = metainfo.length or self.config.snapshot_max_length
+            snapshot_meta = self._trusted_set.timestamp.signed.snapshot_meta
+            length = snapshot_meta.length or self.config.snapshot_max_length
             version = None
             if self._trusted_set.root.signed.consistent_snapshot:
-                version = metainfo.version
+                version = snapshot_meta.version
 
             data = self._download_metadata("snapshot", length, version)
             self._trusted_set.update_snapshot(data)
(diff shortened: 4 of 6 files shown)


tuf/api/metadata.py
```diff
@@ -720,15 +728,13 @@ def from_dict(cls, signed_dict: Dict[str, Any]) -> "Timestamp":
...
-            "snapshot.json": self.meta["snapshot.json"].to_dict()
-        }
+        res_dict["meta"] = {"snapshot.json": self.snapshot_meta.to_dict()}
         return res_dict
 
     # Modification.
     def update(self, snapshot_meta: MetaFile) -> None:
         """Assigns passed info about snapshot metadata to meta dict."""
-        self.meta["snapshot.json"] = snapshot_meta
+        self.snapshot_meta = snapshot_meta
```
jku, line 964: I think this pretty clearly shows that the update() function has no utility: your property setter now does the actual work with better ergonomics.

I think we should remove update()

tuf/api/metadata.py
```diff
@@ -697,6 +697,14 @@ class Timestamp(Signed):
 
     _signed_type = "timestamp"
 
+    @property
+    def snapshot_meta(self):
+        return self.meta["snapshot.json"]
+
+    @snapshot_meta.setter
+    def snapshot_meta(self, value):
```
jku: missing annotation

(we should start tightening mypy config so it complains about this)

jku (commented):
There may still be a discussion on _"is this the way we want to solve the issue"_, but the patch is small so why not finish anyway...

Left a couple of small comments, also note that there is a third reference to `meta['snapshot.json']` in the tests you could change (it probably wasn't there when you made this patch)

jku:
> There may still be a discussion on "is this the way we want to solve the issue"

Expanding on this a bit: The context is that validating something like meta is a bit tricky and is affected by the API we provide:
* providing a helper is already useful and makes safe API use easier but...
* If we leave meta (the plain dictionary) easily editable (and documented as being part of the API), then we have to assume it can contain just about anything. Maybe it's fine and we just assume deserialization fails gracefully when that happens (it does seem to)?
* If we don't want that to advertize meta as being editable, then we'd want to either 
  * hide meta attribute (diverging from file format) and make snapshot_meta an actual instance variable or 
  * make meta a property that is generated from the actual snapshot_meta variable or 
  * annotate/document it as read-only

The last two options of course mean more code in the API itself -- and don't provide any real value. The real options are 
1. this pr
2. remove meta, and only provide snapshot_meta attribute

Externally snapshot_meta will look the same in both cases so I guess this is good to go whatever we decide in the future?

MVrachev:
Thanks, @jku  for all of your comments.

> I think this pretty clearly shows that the update() function has no utility: your property setter now does the actual work with better ergonomics.
I think we should remove update()

I agree with this, but I want to remove this one and change the other `update()` functions in a separate pr where we can
discuss those things. The issue related to that: https://github.com/theupdateframework/tuf/issues/1230

I  agree that `meta` shouldn't contain whatever the user wants and we should add validation.
I proposed a possible solution that won't change the metadata structure and will perform some validation when calling
`__init__` or directly changing `timestamp.meta`.

If we decide to remove `meta` and not follow strictly the specification that's another situation and this validation won't be required.

What do you think of my proposal?

jku:
> If we decide to remove meta and not follow strictly the specification

Not following the spec is not an option on the table, we do want to implement it. The decisions on 
* whether we follow the spec and 
* whether we implement our internal data structures in a way that mimics the file format

are not related to each other and should not be mixed.


I'm not a big fan of adding multiple seemingly supported APIs for one purpose (as I think you are doing with meta property and snapshot_meta property). Also, validating dictionaries like that feels a bit futile as there are so many other ways to modify them:

    timestamp.meta = {"snapshot.json": MetaFile(1)}
    timestamp.meta.clear()
    # oops, we now have a invalid timestamp even with the added validation

Validating standard containers is probably only possible at serialization time or by hiding the container from the caller behind a more restrictive API (like add_key/remove_key): this is what I meant when I said the other day that the more interesting validating discussions are related to the containers, not the individual attributes or classes.

But in this specific case I think we don't need to go that far: documenting `timestamp.snapshot_meta` as the _"API for interacting with snapshot meta information"_ and documenting `timestamp.meta` as an implementation detail that mimics the file format is fine: Let's make it easy to do the right thing, instead of trying to prevent the wrong thing from happening (former is easier and leads to a better API as a side effect).

I think the only reasonable option would be to _not keep meta dict in the API at all_ (in other words stop mimicking the file format) but there does not seem to be consensus on doing that.

MVrachev:
Okay, I documented `meta` and `snapshot_meta` the way you suggested.
Do you think that's a suitable version?

MVrachev:
I think I have clarified that the user should use `snapshot_meta` preferably even more.

tuf/api/metadata.py
MVrachev: Will do that change in a separate pr where we can change all of the update functions
and propose a better solution. 
The issue related to that: https://github.com/theupdateframework/tuf/issues/1230

MVrachev:
I have updated the pr description describing the pr modifications that were made in the pr.
The pr is ready for another review.

sechkova:
Can you also fix the mypy errors in the CI?

MVrachev:
Yes, I forgot to do it.
It's done.

sechkova:
Keeping both a property and an internal dictionary for snapshot seem like an overkill to me (in this particular case). 
I share the opinion that we should drop the dictionary.

sechkova (commented):
Keeping both a property and an internal dictionary for snapshot seems like an overkill to me (in this particular case).
I share the opinion that we should drop the dictionary.

MVrachev:
This issue is blocked until we come up with an agreement what to do with `meta` in Timestamp
after adding `snapshot_meta` property.

MVrachev:
After #1547 got merged this pr will be unblocked.

MVrachev:
This pr is no longer blocked.
After a discussion with the python-tuf maintainers, we decide that:
```
Snapshot meta-information will not be represented by a "meta" dictionary,
but by a MetaFile instance, and with this, it diverges from the specification.
This decision is coherent with ADR9 and the rationale behind it is to provide easier, safer,
and direct access to the snapshot meta information.
```
That's why I will update the pr naming and description.

tuf/api/metadata.py
```diff
@@ -905,10 +905,13 @@ class Timestamp(Signed):
     """A container for the signed part of timestamp metadata.
 
     Timestamp contains information about the snapshot Metadata file.
+    Snapshot meta information is not represented by a "meta" dictionary but by
+    a MetaFile instance and with this, it diverges from the specification.
+    This decision is coherent with ADR9 and the rationale behind it is to
+    provide easier, safer, and direct access to the snapshot meta information.
```
jku: This sounds like a commit message, not API reference documentation:
* API reference should only document the API, not defend the decisions made
* ADRs are likely meaningless to reader, should not be mentioned
* "diverge from specification" sounds like we 're not spec compliant which is not the case

I'm not opposed to noting that the implementation does not quite follow file format here (if you want to rephrase) ... but it also looks quite understandable without the note so maybe this isn't needed at all?

tuf/api/metadata.py
```diff
@@ -918,7 +921,7 @@ def __init__(
         version: int,
         spec_version: str,
         expires: datetime,
-        meta: Dict[str, MetaFile],
+        meta: MetaFile,
```
jku: I think meta is not a great name as it doesn't document the connection to snapshot and it's easily confusing to someone reading the spec and implementation. How about `snapshot_meta`?

jku (changes requested):

tuf/api/metadata.py
jku: basically, API reference is read by people trying to use the code _"how do I use this API to solve this problem X I have"_: text should be written with that person in mind

tuf/api/metadata.py
sechkova: tuf-on-a-plane refers to it as `timestamp.snapshot` which can lead to some confusion but is shorter and has never really bothered me as a reader of that code. We can give this option a chance too.

MVrachev: Given that the main purpose of the `timestamp` role is to provide information for `snapshot` probably you are right it makes sense to just call it `snapshot`.
Will change it.

MVrachev:
I reword the doc strings for the `Timestamp` class and renamed `meta` to `snapshot`.

tuf/api/metadata.py
```diff
@@ -918,33 +918,31 @@ def __init__(
...
-        self.meta = meta
+        self.snapshot = snapshot
 
     @classmethod
     def from_dict(cls, signed_dict: Dict[str, Any]) -> "Timestamp":
         """Creates Timestamp object from its dict representation."""
         common_args = cls._common_fields_from_dict(signed_dict)
         meta_dict = signed_dict.pop("meta")
-        meta = {"snapshot.json": MetaFile.from_dict(meta_dict["snapshot.json"])}
+        snapshot = MetaFile.from_dict(meta_dict["snapshot.json"])
```
sechkova:
```suggestion
        snapshot_meta = MetaFile.from_dict(meta_dict["snapshot.json"])
```

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -286,11 +286,11 @@ def update_snapshot(self, data: bytes) -> None:
         # Snapshot cannot be loaded if final timestamp is expired
         self._check_final_timestamp()
 
-        meta = self.timestamp.signed.meta["snapshot.json"]
+        snapshot = self.timestamp.signed.snapshot
```
sechkova:
```suggestion
        snapshot_meta = self.timestamp.signed.snapshot
```

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -345,14 +345,11 @@ def _check_final_snapshot(self) -> None:
...
-        if (
-            self.snapshot.signed.version
-            != self.timestamp.signed.meta["snapshot.json"].version
-        ):
+        snapshot_meta = self.timestamp.signed.snapshot
+        if self.snapshot.signed.version != snapshot_meta.version:
             raise exceptions.BadVersionNumberError(
                 f"Expected snapshot version "
-                f"{self.timestamp.signed.meta['snapshot.json'].version}, "
+                f"{snapshot_meta.version}, "
```
sechkova: Can the f-string fit into one line less now when it is shortened?

tuf/ngclient/updater.py
```diff
@@ -352,11 +352,11 @@ def _load_snapshot(self) -> None:
             logger.debug("Local snapshot not valid as final: %s", e)
 
             assert self._trusted_set.timestamp is not None  # nosec
-            metainfo = self._trusted_set.timestamp.signed.meta["snapshot.json"]
-            length = metainfo.length or self.config.snapshot_max_length
+            snapshot = self._trusted_set.timestamp.signed.snapshot
```
```suggestion
            snapshot_meta = self._trusted_set.timestamp.signed.snapshot
```

sechkova (dismissed):
I know we agreed on `snapshot` as the name of the `Timestamp` class member but when using it as a single variable I've suggested `snapshot_meta`, to avoid confusion.

The rest LGTM.

tests/test_api.py
```diff
@@ -350,11 +350,11 @@ def test_metadata_timestamp(self):
         fileinfo = MetaFile(2, 520, hashes)
 
         self.assertNotEqual(
-            timestamp.signed.meta['snapshot.json'].to_dict(), fileinfo.to_dict()
+            timestamp.signed.snapshot.to_dict(), fileinfo.to_dict()
```
jku: this is an example of code that looks a bit weird... we have a timestamp that is not a `Timestamp`, and a snapshot that is not a `Snapshot` -- the first one is just local variable naming of course, but naming that our API design easily leads to...

jku (approved):
Looks good I think. Timestamp.update() is now even weirder than before but you've said you'd rather remove it in a separate change: this sounds good to me.

The names in  the attribute chains do end up looking a bit goofy when the name is just "snapshot" (and data type is not actually Snapshot).... @joshuagl  do you have an opinion on the naming (looking at test code with e.g. `self.timestamp.snapshot` and `timestamp.signed.snapshot`). I think I can live with this but wouldn't object to calling this attribute snapshot_meta either :shrug:

joshuagl:
> The names in  the attribute chains do end up looking a bit goofy when the name is just "snapshot" (and data type is not actually Snapshot).... @joshuagl  do you have an opinion on the naming (looking at test code with e.g. `self.timestamp.snapshot` and `timestamp.signed.snapshot`). I think I can live with this but wouldn't object to calling this attribute snapshot_meta either :shrug:
> 

While I appreciate that it's a long variable name, I am fond of `snapshot_meta`. It retains some symmetry with the file formats in the spec without adding the extra layer of indirection.

MVrachev:
> > The names in  the attribute chains do end up looking a bit goofy when the name is just "snapshot" (and data type is not actually Snapshot).... @joshuagl  do you have an opinion on the naming (looking at test code with e.g. `self.timestamp.snapshot` and `timestamp.signed.snapshot`). I think I can live with this but wouldn't object to calling this attribute snapshot_meta either 🤷
> 
> While I appreciate that it's a long variable name, I am fond of `snapshot_meta`. It retains some symmetry with the file formats in the spec without adding the extra layer of indirection.

For sure `snapshot_meta` is clearer than just `snapshot`.  We should shorten our names only if we are sure it won't cause confusion and as it seems that both @jku and @joshuagl have some concerns and that's why it seems better to stick to the slightly longer and descriptive `snapshot_meta`.
Will rename it now.

MVrachev:
Done, renamed `snapshot` to `snapshot_meta`.
It seems there are newer versions of pylint which require fixing `global-variable-not-assigned` error.
Do you want me to fix that here or in another pr?

jku:
> Do you want me to fix that here or in another pr?

no conflicts expected so separate PR seems like it would not be much more work?

MVrachev:
> > Do you want me to fix that here or in another pr?
> 
> no conflicts expected so separate PR seems like it would not be much more work?

It won't be. Will post a pr today.

MVrachev:
> > Do you want me to fix that here or in another pr?
> 
> no conflicts expected so separate PR seems like it would not be much more work?

Fixed in #1584.

jku (approved):

jku:
Can you rebase please (branch protection does not allow merging with failing checks)

Issue #338, opened 2016-06-27
high cpu usage on download


julianhille:
Im trying to have tuf on a device which is not very powerful.
It is an arm v7 with max 1ghz.

The cpu load was very high for the time the repository updates i found the cause to be the downloading while loop.

Normal update of a repository with one mirror. And nothing to update.

```
real	0m13.169s
user	0m9.796s
sys	0m0.600s
```

After adding [here](https://github.com/theupdateframework/tuf/blob/develop/tuf/download.py#L318) 
    
    time.sleep(0.05)

The time where nearly the same but my CPU had enough power to process other things.

```
real	0m13.686s
user	0m2.840s
sys	0m0.352s
```

SantiagoTorres:
> ```
> 
> After adding [here](https://github.com/theupdateframework/tuf/blob/develop/tuf/download.py#L318) 
>     
>     time.sleep(0.05)

Also, it seems that there are some logger calls commented out?

https://github.com/theupdateframework/tuf/blob/develop/tuf/download.py#L324
https://github.com/theupdateframework/tuf/blob/develop/tuf/download.py#L345-L346

I think it's worth cleaning up now that we're at it.

Thanks,
-Santiago.

vladimir-v-diaz:
What happens if you increase the chunk size?
https://github.com/theupdateframework/tuf/blob/develop/tuf/conf.py#L79-L80

Is there a substantial improvement in this case?

julianhille:
I didn't expect to see an improvement and i didn't.
I doubled the CHUNK_SIZE and had these:

```
real	0m13.250s
user	0m9.984s
sys	0m0.656s
```

i also halve it:

```
real	0m12.762s
user	0m9.676s
sys	0m0.552s
```

So this does not seem to change anything.

vladimir-v-diaz:
I'll take a closer look and try to determine why that while loop might be slow.
I tested an update example on my machine (Intel® Core™ i5-4210U CPU @ 1.70GHz × 4 ) and got the following times:

> real	0m3.166s
> user	0m3.142s
> sys	0m0.024s

However, I used [this](https://github.com/vladimir-v-diaz/tuf/tree/keyids_multihash_algorithms/tests/repository_data/repository/metadata) repository.  My client script performed a refresh() of top-level metadata (only timestamp.json was downloaded, verified, and found to be up to date).

SantiagoTorres:
[...]

PR #1448, opened 2021-06-15
Review download code

### LINKED ISSUE #1399: ngclient: review download code

sechkova:
Fixes #1399

_This PR builds on top of #1408 and will remain a draft until its merge._

- simplify download code, remove average download speed check
- wrap the download functionality in `FileDownload` class
- simplify `RequestsFetcher.fetch()`
- add type annotations

requirements**:


jku:
Looks great at first glance. 

This maybe a dumb question but is there something that prevents us from putting the new FileDownloader functionality directly into FetcherInterface?

Are the RequestsFetcher improvements dependent on the other changes or just included in same PR for convenience?

sechkova:
> This maybe a dumb question but is there something that prevents us from putting the new FileDownloader functionality directly into FetcherInterface?
> 
Only me not thinking enough about it ... :roll_eyes: 

> Are the RequestsFetcher improvements dependent on the other changes or just included in same PR for convenience?

No, they are not dependent, I just consider fetcher as part of the download code.

I do like the idea of simplifying further and keeping only a single download/fetcher module instead of the two. How should we do it? Build on top of this PR, a separate PR or I can rewrite these commits here and directly move download.py functions into FetcherInterface?

jku:
> I do like the idea of simplifying further and keeping only a single download/fetcher module instead of the two. How should we do it? Build on top of this PR, a separate PR or I can rewrite these commits here and directly move download.py functions into FetcherInterface?

I think merging a PR that adds a new object would be weird if we intend to remove it soon in a followup PR... otherwise whatever works for you I think

sechkova:
Rewrote part of the commits here. Now the simplified download functions are made part of `FetcherInterface` and the `download.py` module is removed.

sechkova:
I don't know what is blocking GitHub actions here?

jku:
> I don't know what is blocking GitHub actions here?

quite strange. It never actually started the action (https://github.com/theupdateframework/tuf/actions) so I can't even restart it.

Looks really nice btw... Net result of removing 200 lines and ending up with a cleaner implementation is not too bad.

tuf/ngclient/fetcher.py
```diff
@@ -40,3 +48,51 @@ def fetch(self, url: str, required_length: int) -> Iterator[bytes]:
...
+        Yields:
+          A file object that points to the contents of 'url'.
+        """
+        # 'url.replace('\\', '/')' is needed for compatibility with
+        # Windows-based systems, because they might use back-slashes in place
+        # of forward-slashes. This converts it to the common format.
+        # unquote() replaces %xx escapes in a url with their single-character
+        # equivalent. A back-slash may beencoded as %5c in the url, which
+        # should also be replaced with a forward slash.
+        url = parse.unquote(url).replace("\\", "/")
```
jku, line 74: let's get rid of this: this component should expect to get a real URL and should not do this.

Do we need a new issue to handle possible issues in higher level code?

tuf/ngclient/fetcher.py
```diff
@@ -40,3 +48,51 @@ def fetch(self, url: str, required_length: int) -> Iterator[bytes]:
             A bytes iterator
         """
         raise NotImplementedError  # pragma: no cover
+
+    @contextmanager
+    def download_file(self, url: str, required_length: int) -> Iterator[IO]:
```
jku: required_length is actually more like max_length, right?

tuf/ngclient/fetcher.py
```diff
@@ -40,3 +48,51 @@ def fetch(self, url: str, required_length: int) -> Iterator[bytes]:
...
+          url: a URL string that represents the location of the file.
+          required_length: an integer value representing the length of
+              the file or an upper boundary.
+
+        Raises:
+          DownloadLengthMismatchError: a mismatch of observed vs expected
+              lengths while downloading the file.
+
+        Yields:
+          A file object that points to the contents of 'url'.
```
jku: I wonder if it's worth mentioning that it's a TemporaryFile?

tuf/ngclient/fetcher.py
```diff
@@ -40,3 +48,51 @@ def fetch(self, url: str, required_length: int) -> Iterator[bytes]:
...
+                temp_file.write(chunk)
+                number_of_bytes_received += len(chunk)
+            if number_of_bytes_received > required_length:
+                raise exceptions.DownloadLengthMismatchError(
+                    required_length, number_of_bytes_received
+                )
+            temp_file.seek(0)
+            yield temp_file
+
+    def download_bytes(self, url: str, required_length: int) -> bytes:
```
jku: (same comment on required_length naming)

tuf/ngclient/fetcher.py
```diff
@@ -39,3 +48,51 @@ def fetch(self, url, required_length):
...
+        up to 'required_length'.
+
+        Args:
+          url: a URL string that represents the location of the file.
+          required_length: an integer value representing the length of
+              the file or an upper boundary.
+
+        Raises:
+          DownloadLengthMismatchError: a mismatch of observed vs expected
+              lengths while downloading the file.
```
jku:
```suggestion
          DownloadLengthMismatchError: File length was larger than <name of argument>
```

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -86,58 +87,62 @@ def fetch(self, url, required_length):
...
+                # large file in one shot. Before beginning the round, sleep
+                # (if set) for a short amount of time so that the CPU is not
+                # hogged in the while loop.
+                if settings.SLEEP_BEFORE_ROUND:
+                    time.sleep(settings.SLEEP_BEFORE_ROUND)
+
+                read_amount = min(
+                    settings.CHUNK_SIZE,
+                    required_length - bytes_received,
+                )
```
jku: let's get rid of settings: 
* CHUNK_SIZE: this is a RequestsFetcher internal detail. make it an instance attribute (maybe with a smaller default than the current huge number)?
* SLEEP_BEFORE_ROUND: if this truly is an issue we should fix the issue -- there is no reason for us to be running a hot while loop for downloading things, that's just crazy.  #338 documents that that is indeed how the code used to work... but that was a long time ago, and the code looked entirely different. I'd say let's spend a bit of time making sure the loop is sane (and doesn't hog cpu) and let's remove SLEEP_BEFORE_ROUND

jku (commented):
I like it a lot: this makes download code much easier to understand (and I agree with the decisions made)! Left several comments but the only one that I think needs to be fixed here is the argument name: _required_length_ does not seem to represent what it is.

Feel free to file issues for the other things

tuf/ngclient/_internal/requests_fetcher.py
jku: for reference, pip and requests use default chunk size of 1024*10

tuf/ngclient/fetcher.py
sechkova: yes, it is the upper boundary.

tuf/ngclient/_internal/requests_fetcher.py
jku: this is alread being handled 1470 so maybe forget the comment here -- we should get rid of sleep before round but doesn't need to happen here.

tuf/ngclient/fetcher.py
sechkova: Opened a new issue although there are several ones related to paths and urls in Updater already: https://github.com/theupdateframework/tuf/issues/1483

tuf/ngclient/_internal/requests_fetcher.py
sechkova: I opened https://github.com/theupdateframework/tuf/issues/1484 for it and we can continue discussing settings in the #1470 thread.

tuf/ngclient/fetcher.py
jku: oh you might be right about multiple bugs -- I probably file one every time I see this code

sechkova:
I think all suggestions are added with 959ab5a.

sechkova:

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -53,7 +53,7 @@ def __init__(self):
         self.chunk_size: int = 400000  # bytes
         self.sleep_before_round: Optional[int] = None
 
-    def fetch(self, url, required_length):
+    def fetch(self, url: str, required_length: int) -> Iterator[bytes]:
```
jku: I guess this is more `max_length` than `required_length` as well?

the docstring could be better as well

tuf/ngclient/fetcher.py
```diff
@@ -20,7 +29,7 @@ class FetcherInterface:
     __metaclass__ = abc.ABCMeta
 
     @abc.abstractmethod
-    def fetch(self, url, required_length):
+    def fetch(self, url: str, required_length: int) -> Iterator[bytes]:
```
jku: maybe should be max_length (also docstring could be better)?

This is sort of API already but the change maybe worth it as required_length really is misleading...

jku (approved):
Left comments/question about naming of the argument to Fetcher.fetch() as well... but I'm approving as is, this is a very good improvement on download code IMO

sechkova:
Amended the last commit 0eff004 with the suggestion.

MVrachev (approved):
Great job!
I liked all of the simplifications.
It's easier to read now.
It looks good to me as well.

PR #1449, opened 2021-06-15
Metadata API: Add Key attributes types validation

### LINKED ISSUE #1438: Metadata Attribute research: KEY
This issue aims to document thoughts about the `KEY` attribute from the `Root` and `Delegations` classes.
The goal is to understand how we use that attribute, what might go wrong with it and how we can validate it.

When thinking about how we use `KEY` worth reading what is implemented in pr #1408 and pr #1436 implementing threshold validation. 

1. How do we use it?
2. What might go wrong?
3. Rethink how we store it?

MVrachev:
Fixes #1438

In our discussion with @jku , we come to the conclusion that we want
to verify that all Key attributes contain values in the expected types,
but at the same time, we don't want to focus on validating the semantics
behind them.
The reason is that having a Key instance with invalid attributes is
possible and supported by the spec.
That's why we have a "threshold" for the roles meaning we can have up to
a certain number of invalid Keys until we satisfy
the required threshold.

Also, for deeper semantic validation it's better to be done in
securesystemslib which does the actual work with keys.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -376,6 +376,8 @@ class Signed(metaclass=abc.ABCMeta):
 
 class Key:
     """A container class representing the public portion of a Key.
+    Please note that "Key" instances are not semanticly validated during
+    initialization. We consider this as responsibility of securesystemslib.
 
     Attributes:
         keyid: An identifier string that must uniquely identify a key within
@@ -398,15 +400,9 @@ class Key:
         keyval: Dict[str, str],
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
-        public_val = keyval.get("public")
-        if not public_val or not isinstance(public_val, str):
-            raise ValueError("keyval doesn't follow the specification format!")
-        if not isinstance(scheme, str):
-            raise ValueError("scheme should be a string!")
-        if not isinstance(keytype, str):
-            raise ValueError("keytype should be a string!")
-        if not isinstance(keyid, str):
-            raise ValueError("keyid should be a string!")
+        val = keyval["public"]
+        if not all(isinstance(at, str) for at in [keyid, keytype, scheme, val]):
+            raise ValueError("Unexpected Key attributes types!")
         self.keyid = keyid
         self.keytype = keytype
         self.scheme = scheme
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -398,8 +398,15 @@ class Key:
         keyval: Dict[str, str],
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
-        if not keyval.get("public"):
+        public_val = keyval.get("public")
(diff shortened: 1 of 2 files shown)


sechkova:
Should I assume from this PR that we want to implement the validation checks of the various metadata classes in the constructor and not in the deserialization methods (`from_dict()`)?

MVrachev:
> Should I assume from this PR that we want to implement the validation checks of the various metadata classes in the constructor and not in the deserialization methods (`from_dict()`)?

From my understanding, yes. The reason is that this way no matter how we are creating Key objects -  directly through `__init__` like `Key()` or calling `Key.from_dict()` which after that will call `Key.__init__` we have validation.

tuf/api/metadata.py
```diff
@@ -398,8 +400,15 @@ def __init__(
         keyval: Dict[str, str],
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
-        if not keyval.get("public"):
+        public_val = keyval.get("public")
+        if not public_val or not isinstance(public_val, str):
```
jku: We're relying on KeyErrors from index access elsewhere in serialization (and this code also relies on TypeError if keyval is not a dict). I think letting index access fail with keyerror is simpler and works just as well:
`if not isinstance(keyval["public"], str):`

jku (approved):
This seems like the approach we have to take: explicit type checking is ewww, but we don't want to / cannot try to understand the semantics of our input at serialization time so it probably makes sense.

The repetition of if-raise could easily be avoided (as the unique error messages are not that important here) so the whole thing would be just 3 lines but I'm fine with this as well, your choice. Left one simplification suggestion though

MVrachev:
I experimented with a shorter variation.
In this case, we will receive `KeyError` when accessing the public key value `val = keyval["public"]` and the exception will not point to a long if a line with checks and of course the code is shorter and at least to me looks readable :D.

tuf/api/metadata.py
```diff
@@ -398,8 +400,9 @@ def __init__(
         keyval: Dict[str, str],
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
-        if not keyval.get("public"):
-            raise ValueError("keyval doesn't follow the specification format!")
+        val = keyval["public"]
+        if not all(isinstance(at, str) for at in [keyid, keytype, scheme, val]):
+            raise ValueError("Attributes don't follow the type annotations!")
```
jku: LGTM, I agree getting rid of the repetition outweighs the slightly more complex code.

The message is really not terribly important but maybe something like "Unexpected Key attributes" might be better than referring to annotations?

approved in any case.

jku (approved):

tuf/api/metadata.py
MVrachev: Updated with `Unexpected Key attributes types!` message.

PR #1450, opened 2021-06-15
Metadata API: Add simple threshold validation

### LINKED ISSUE #1439: Metadata Attribute research: Threshold
This issue aims to document thoughts about the `Threshold` attribute from the `root` and `targets` roles and more precisely in our implementation `Role` and `DelegatedRole` classes.
The goal is to understand how we use that attribute, what might go wrong with it and how we can validate it.

1. How do we use it?
2. What might go wrong?
3. Rethink how we store it?

MVrachev:
Fixes #1439

Probably there could be future API calls that modify "threshold"
to a new value, but the problem is we don't have a clear idea
if they would exist and what exactly they will do.
That's why it makes sense to validate against the potential problems
we can imagine - in this case, is passing a threshold below 1.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -506,6 +506,8 @@ class Role:
                 f"keyids should be a list of unique strings,"
                 f" instead got {keyids}"
             )
+        if threshold < 1:
+            raise ValueError("threshold should be at least 1!")
         self.keyids = keyids_set
         self.threshold = threshold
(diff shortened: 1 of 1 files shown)


jku (approved):
looks reasonable to me (spec doesn't very explicitly make this requirement but it seem logical). Doing a value check like this also implicitly checks that we have a type we can compare with <>. This seems like good enough typing to me.

PR #1451, opened 2021-06-16
api/metadata input validation: length and hashes

### LINKED ISSUE #1441: api/metadata input validation: hashes
Implement input validation for `TargetFile` and `MetaFile` `hashes` attribute.

**Current behavior**:
The new api/metadata code does not perform any input validation on `hashes`.
`formats.py`  has a defined `HASHDICT_SCHEMA` that is not used in the new code.

**Expected behavior**:
Define allowed values for hashes. 
Implement the verification in `metadata.py`

sechkova:
Fixes #1441

_Draft since it is dependent on #1437_ 

Adds input validation checks to `MetaFile` and `TargetFile` common fields:

- valid length: greater than zero
- valid hashes: a non-empty dictionary

Checking the validity of hash algorithms is not part of the metadata input validation and is done
by `securesystemslib` during  hash verification.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -702,6 +702,19 @@ class BaseFile:
                 f"expected length {expected_length}"
             )
 
+    @staticmethod
+    def _validate_hashes(hashes: Dict[str, str]) -> None:
+        if not hashes:
+            raise ValueError("Hashes must be a non empty dictionary")
+        for key, value in hashes.items():
+            if not (isinstance(key, str) and isinstance(value, str)):
+                raise TypeError("Hashes items must be strings")
+
+    @staticmethod
+    def _validate_length(length: int) -> None:
+        if length <= 0:
+            raise ValueError(f"Length must be > 0, got {length}")
+
 
 class MetaFile(BaseFile):
     """A container with information about a particular metadata file.
@@ -730,6 +743,14 @@ class MetaFile(BaseFile):
         hashes: Optional[Dict[str, str]] = None,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
+
+        if version <= 0:
+            raise ValueError(f"Metafile version must be > 0, got {version}")
+        if length is not None:
+            self._validate_length(length)
+        if hashes is not None:
+            self._validate_hashes(hashes)
+
         self.version = version
         self.length = length
         self.hashes = hashes
@@ -741,12 +762,6 @@ class MetaFile(BaseFile):
         version = meta_dict.pop("version")
         length = meta_dict.pop("length", None)
         hashes = meta_dict.pop("hashes", None)
-
-        # Do some basic input validation
-        if version <= 0:
-            raise ValueError(f"Metafile version must be > 0, got {version}")
-        if length is not None and length <= 0:
-            raise ValueError(f"Metafile length must be > 0, got {length}")
 
         # All fields left in the meta_dict are unrecognized.
         return cls(version, length, hashes, meta_dict)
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -708,6 +720,14 @@ def __init__(
         hashes: Optional[Dict[str, str]] = None,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
+
+        if version <= 0:
```
MVrachev, line 747: Nit: these days I had a thought that it's probably easier to read if we use `if version < 1` when doing this check.
What do you think?

MVrachev:
Even though we don't want to focus on validating the semantic behind `hashes` probably it's good to validate that
the keys and values in this dictionaries are strings.
If they aren't how are we going to compare the hashes against what we expect?
We did a similar choice for #1449 where we don't validate the semantics, but we make sure the types are as expected
and we can do the required comparisons.

jku:
> ... probably it's good to validate that the keys and values in this dictionaries are strings.

I've started to agree with type checks _in this situation_: 
* we explicitly do not want to check the semantics of the keys (algorithms) and checking value content (hashes) is a bit pointless
* this means the keys/values could be any json object after deserialization...
* which should not be an issue for our code (sslib will raise in _verify_hashes() and we will handle that in #1452), but ...
* **we will happily serialize the invalid metadata**, if asked to

The last point feels quite wrong so maybe we should check that they are str at object intialization

tuf/api/metadata.py
sechkova: I find `<= 0` very readable

sechkova:
Rebased on develop and added type checks for the dictionary key-values based on your suggestions.

tuf/api/metadata.py
```diff
@@ -702,6 +702,21 @@ def _verify_length(
                 f"expected length {expected_length}"
             )
 
+    @staticmethod
+    def _validate_hashes(hashes: Dict[str, str]) -> None:
+        if not isinstance(hashes, dict):
+            raise TypeError(f"Hashes must be a dictionary, got {type(hashes)}")
```
jku: This check could now be dropped -- the error we get from calling `hashes.items()` on something that isn't a dict is not as good but it does raise an exception without extra code...

jku (dismissed):
looks fine to me. Left a comment about a check that isn't strictly necessary anymore but if you want to keep it, that's fine too.

tuf/api/metadata.py
sechkova: I dropped it and force pushed, we raise build-in errors anyway.

tuf/api/metadata.py
```diff
@@ -702,6 +702,19 @@ def _verify_length(
                 f"expected length {expected_length}"
             )
 
+    @staticmethod
+    def _validate_hashes(hashes: Dict[str, str]) -> None:
+        if not hashes:
+            raise ValueError("Hashes must be a non empty dictionary")
+        for key, value in hashes.items():
+            if not (isinstance(key, str) and isinstance(value, str)):
```
jku, line 710: on a more careful re-read, this is a bug -- the order of operations is not what you expect here

sechkova: maybe I'm missing something because I don't see the bug?

jku: oops I'm the one that needs new glasses, somehow managed to read this without the enclosing parens

jku (approved):

MVrachev (approved):
LGTM!

PR #1453, opened 2021-06-17
Add 'ecdsa' to the list of supported key types

### LINKED ISSUE #1388: Metadata file contains a key with an invalid keytype.
ECDSA keys are generated with a keytype of `ecdsa` using the securesystemslib https://github.com/secure-systems-lab/securesystemslib/blob/2d33d91bef92f53b5e9489d57033bb9d046a6d23/securesystemslib/keys.py#L1678.

However, if you load a repository that's created with a key from `import_ecdsakey_from_public_pem`, you run a TUF error that the keytype is unsupported because of:

https://github.com/theupdateframework/tuf/blob/30ba6e9f9ab25e0370e29ce574dada2d8809afa0/tuf/keydb.py#L48

Should the keytype instead be `ecdsa`, rather than `ecdsa-sha2-nistp256`, which is the scheme type?

avelichka:
Adds a new "ecdsa" key type to the list of the supported key types

Fixes #1388

requirements**:


sechkova (approved):

trishankatdatadog:
Might interest @asraa

asraa:
Yay! Yes, this should fix that invalid key type issue! might be good to add a test where you import a key from the securesystemslib func `pubkey = import_ecdsakey_from_pem(pem)`

avelichka:
I've changed tests/repository_data/keystore/root_key3* to be an ecdsa key so that the test_updater_root_rotation_integration.py test tests both repotool and updater as @jku suggested. 
I created the new ecdsa key with securesystemslib.keys' generate_ecdsa_key and encrypted it with securesystemslib.keys' encrypt_key. 
Hope that works :fingers_crossed:

jku (approved):
Looks good to me, thank you. 

This does not handle any deprecation of 'ecdsa-sha2-nistp256' as keytype, but... I think TUF should not really look at the keytypes at all in the future so I don't think that's necessary and it definitely does not have be in this PR.

PR #1454, opened 2021-06-17
BaseFile._verify_hashes: handle sslib errors

### LINKED ISSUE #1452: MetaFile/TargetFile: ensure securesystemslib errors are all handled
Filing this so we can merge #1437 (_Add hash and length verification to MetaFile and TargetFile_) and can continue with this afterwards.

We need to review the errors that maybe thrown by securesystemslib in in `Basefile._verify_hashes()`: 
* do we want to handle all of them? 
* should we just raise LengthOrHashMismatchError or something else?

My first instinct is that this is like verify_signatures(): the error messages may be interesting but from client POV what matters is that we cannot guarantee that hashes match.

cc @sechkova

sechkova:
Fixes #1452

BaseFile._verify_hashes handles the following sslib errors raised during hash verification:

- securesystemslib.exceptions.UnsupportedAlgorithmError
- securesystemslib.exceptions.FormatError

and reraises them as tuf.exceptions.UnsupportedAlgorithmError.

securesystemslib.exceptions.FormatError should not be raised if the Target/MetaFile object is
created correctly through its constructor but is possible if Target/MetaFile.hashes is modified
directly so catching it just in case. 

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -707,6 +707,15 @@ class TestMetadata(unittest.TestCase):
             self.assertRaises(exceptions.LengthOrHashMismatchError,
                 snapshot_metafile.verify_length_and_hashes, data)
 
+            snapshot_metafile.hashes = {'unsupported-alg': "8f88e2ba48b412c3843e9bb26e1b6f8fc9e98aceb0fbaa97ba37b4c98717d7ab"}
+            self.assertRaises(exceptions.LengthOrHashMismatchError,
+            snapshot_metafile.verify_length_and_hashes, data)
+
+            # Test wrong algorithm format (sslib.FormatError)
+            snapshot_metafile.hashes = { 256: "8f88e2ba48b412c3843e9bb26e1b6f8fc9e98aceb0fbaa97ba37b4c98717d7ab"}
+            self.assertRaises(exceptions.LengthOrHashMismatchError,
+            snapshot_metafile.verify_length_and_hashes, data)
+
             # test optional length and hashes
             snapshot_metafile.length = None
             snapshot_metafile.hashes = None
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -679,12 +679,20 @@ class BaseFile:
         """Verifies that the hash of 'data' matches 'expected_hashes'"""
         is_bytes = isinstance(data, bytes)
         for algo, exp_hash in expected_hashes.items():
-            if is_bytes:
-                digest_object = sslib_hash.digest(algo)
-                digest_object.update(data)
-            else:
-                # if data is not bytes, assume it is a file object
-                digest_object = sslib_hash.digest_fileobject(data, algo)
+            try:
+                if is_bytes:
+                    digest_object = sslib_hash.digest(algo)
+                    digest_object.update(data)
+                else:
+                    # if data is not bytes, assume it is a file object
+                    digest_object = sslib_hash.digest_fileobject(data, algo)
+            except (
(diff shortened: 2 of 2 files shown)


sechkova:
Depends on #1451 so opened as a draft.

jku:
This is the core question:

>  reraises them as tuf.exceptions.UnsupportedAlgorithmError

this is the obvious choice... yet I'm left wondering if it's actually useful for the API users.
* in my usage I've only been interested in "have all hashes been verified to match the content?" and now I need to handle another exception to reliably get an answer
* Is there a reasonable use for UnsupportedAlgorithmError? What might an application do in this case that it does not want to do in the case of LengthOrHashMismatchError?


Follow up in the case that there is a real use case:
* if there is a reasonable use case, should the error we use derive from LengthOrHashMismatchError so that users who do not care about unsupported algorithms could just handle that one?

tuf/api/metadata.py
```diff
@@ -670,12 +671,20 @@ def _verify_hashes(
...
+                    digest_object = sslib_hash.digest(algo)
+                    digest_object.update(data)
+                else:
+                    # if data is not bytes, assume it is a file object
+                    digest_object = sslib_hash.digest_fileobject(data, algo)
+            except (
+                sslib_exceptions.UnsupportedAlgorithmError,
+                sslib_exceptions.FormatError,
+            ) as e:
+                raise exceptions.UnsupportedAlgorithmError(
```
jku: assuming we go with this, it needs to be documented in at least the public methods

sechkova:
What I think is worth doing is giving the interested users, if any, information about the error. Following this thought maybe your proposal of deriving UnsupportedAlgorithmError from LengthOrHashMismatchError is a good compromise making everyone relatively happy.

jku:
> What I think is worth doing is giving the interested users, if any, information about the error.

This is absolutely not contested: we should have actionable error messages.

But I would like to only add errors if they have a purpose: to understand which component might handle the error (the component may be a hypothetical one: I'd just like to understand the use case). Adding more complex errors in later releases is easy, removing existing ones is really hard.

sechkova:
Kept only `LengthOrHashMismatchError` since this is easily extendable later and for consistency with #1435.
Rebased on develop, not a draft anymore.

tests/test_api.py
```diff
@@ -707,6 +707,15 @@ def  test_length_and_hash_validation(self):
...
                 snapshot_metafile.verify_length_and_hashes, data)
 
+            snapshot_metafile.hashes = {'unsupported-alg': "8f88e2ba48b412c3843e9bb26e1b6f8fc9e98aceb0fbaa97ba37b4c98717d7ab"}
+            self.assertRaises(exceptions.LengthOrHashMismatchError,
+            snapshot_metafile.verify_length_and_hashes, data)
+
+            # Test wrong algorithm format (sslib.FormatError)
+            snapshot_metafile.hashes = { 256: "8f88e2ba48b412c3843e9bb26e1b6f8fc9e98aceb0fbaa97ba37b4c98717d7ab"}
+            self.assertRaises(exceptions.LengthOrHashMismatchError,
+            snapshot_metafile.verify_length_and_hashes, data)
```
jku, line 717: this should get prevented by MetaFile but does not hurt to test...

jku (approved):
lgtm, thanks

tests/test_api.py
sechkova: yes, you are right but since nothing forbids updating directly `hashes` like this ... just in case :shrug:

PR #1455, opened 2021-06-17
Harden mypy configuration

jku:
We can almost enable "strict = True" already but
* we should decide if we really want to do that (what strict means may change when mypy updates)
* we're not quite there yet...

So fix a few issues and harden the mypy configuration where possible. 

The one remaining possibly interesting option (that is currently included in "strict") is: 
* `no_any_return`: seems to require securesystemslib.formats annotations (or specifying some types in tuf code that calls securesystemslib)

### DIFF
--- a/tuf/exceptions.py
+++ b/tuf/exceptions.py
@@ -24,6 +24,8 @@
 
 from urllib import parse
 
+from typing import Any, Dict
+
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


sechkova (approved):
So the only way to find out what will be applied with the `strict` flag is to check `mypy --strict `on the current version ... 

Anyway, I approve extending the list and see where this will lead us to.

jku:
rebased on develop, no changes

jku:
rebased on develop, no changes

tuf/exceptions.py
```diff
@@ -252,16 +253,17 @@ class InvalidNameError(Error):
 class UnsignedMetadataError(Error):
   """Indicate metadata object with insufficient threshold of signatures."""
 
-  def __init__(self, message, signable):
+  # signable is not used but kept in method signature for backwards compat
```
joshuagl, line 256: Do we want a FIXME or similar here? So that we are reminded to remove this unused variable before we tag 1.0?

joshuagl (approved):
Looks good to me, thanks!

> We can almost enable "strict = True" already but
> 
> we should decide if we really want to do that (what strict means may change when mypy updates)

We should notice when the definition of strict changes thanks to CI, but could always pin mypy version and review updates for changes to the definition of strict?

PR #1456, opened 2021-06-21
Metadata API: Rewrite comments

jku:
Try to keep dostrings and comments to the point, avoid mentioning
details if they are not necessary or are likely to become outdated
and try to minimize number of comment lines.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -46,10 +46,7 @@ from tuf.api.serialization import (
     SignedSerializer,
 )
 
-# Disable the "C0302: Too many lines in module" warning which warns for modules
-# with more 1000 lines, because all of the code here is logically connected
-# and currently, we are above 1000 lines by a small margin.
-# pylint: disable=C0302
+# pylint: disable=too-many-lines
 
 # We aim to support SPECIFICATION_VERSION and require the input metadata
 # files to have the same major version (the first number) as ours.
@@ -65,7 +62,6 @@ class Metadata:
     Attributes:
         signed: A subclass of Signed, which has the actual metadata payload,
             i.e. one of Targets, Snapshot, Timestamp or Root.
-
         signatures: An ordered dictionary of keyids to Signature objects, each
             signing the canonical serialized representation of 'signed'.
     """
@@ -92,8 +88,8 @@ class Metadata:
 
         Returns:
             A TUF Metadata object.
-
         """
+
         # Dispatch to contained metadata class on metadata _type field.
         _type = metadata["signed"]["_type"]
 
@@ -148,8 +144,8 @@ class Metadata:
 
         Returns:
             A TUF Metadata object.
-
         """
+
         if storage_backend is None:
             storage_backend = FilesystemBackend()
 
@@ -202,20 +198,18 @@ class Metadata:
 
         Arguments:
             filename: The path to write the file to.
-            serializer: A MetadataSerializer subclass instance that implements
-                the desired wireline format serialization. Per default a
-                JSONSerializer is used.
-            storage_backend: An object that implements
-                securesystemslib.storage.StorageBackendInterface. Per default
-                a (local) FilesystemBackend is used.
+            serializer: A MetadataSerializer instance that implements the
+                desired serialization format. Default is JSONSerializer.
+            storage_backend: A StorageBackendInterface implementation. Default
+                is FilesystemBackend (i.e. a local file).
 
         Raises:
             tuf.api.serialization.SerializationError:
                 The metadata object cannot be serialized.
             securesystemslib.exceptions.StorageError:
                 The file cannot be written.
-
         """
+
         if serializer is None:
             # Use local scope import to avoid circular import errors
             # pylint: disable=import-outside-toplevel
@@ -237,14 +231,12 @@ class Metadata:
         """Creates signature over 'signed' and assigns it to 'signatures'.
 
         Arguments:
-            signer: An object implementing the securesystemslib.signer.Signer
-                interface.
+            signer: A securesystemslib.signer.Signer implementation.
             append: A boolean indicating if the signature should be appended to
                 the list of signatures or replace any existing signatures. The
                 default behavior is to replace signatures.
-            signed_serializer: A SignedSerializer subclass instance that
-                implements the desired canonicalization format. Per default a
-                CanonicalJSONSerializer is used.
+            signed_serializer: A SignedSerializer that implements the desired
+                serialization format. Default is CanonicalJSONSerializer.
 
         Raises:
             tuf.api.serialization.SerializationError:
@@ -255,8 +247,8 @@ class Metadata:
 
         Returns:
             A securesystemslib-style signature object.
-
         """
+
         if signed_serializer is None:
             # Use local scope import to avoid circular import errors
             # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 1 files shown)


tuf/api/metadata.py
```diff
@@ -891,18 +852,14 @@ class DelegatedRole(Role):
 
     Attributes:
         name: A string giving the name of the delegated role.
-        keyids: A set of strings each of which represents a given key.
-        threshold: An integer representing the required number of keys for that
-            particular role.
```
jku, line 896: these were removed because they are Role attributes, not DelegatedRole attributes (and we don't document parent attributes in other classes)

tuf/api/metadata.py
```diff
@@ -46,10 +46,7 @@
     SignedSerializer,
 )
 
-# Disable the "C0302: Too many lines in module" warning which warns for modules
-# with more 1000 lines, because all of the code here is logically connected
-# and currently, we are above 1000 lines by a small margin.
-# pylint: disable=C0302
+# pylint: disable=too-many-lines
```
MVrachev, line 49: Great, didn't know we can reference warning explicitly!

tuf/api/metadata.py
```diff
@@ -407,8 +399,9 @@ def bump_version(self) -> None:
 
 class Key:
     """A container class representing the public portion of a Key.
+
     Please note that "Key" instances are not semanticly validated during
-    initialization. We consider this as responsibility of securesystemslib.
+    initialization: this only happens at signature verification time.
```
MVrachev, line 404: Do we plan to use the `Key` class besides signature verification?
If no other cases appear in mind then okay, let's be more precise, but if there are other cases we envision it could be better to be a little more abstract.

tuf/api/metadata.py
```diff
@@ -512,15 +504,15 @@ def verify_signature(
 
 
 class Role:
-    """A container class containing the set of keyids and threshold associated
-    with a particular role.
+    """Container for data that defines which keys are required to sign metadata
```
```suggestion
    """Container for data that defines which keys are required to sign metadata.
```

MVrachev (approved):
LGTM with the exception of one comment in the `Key` class.

tuf/api/metadata.py
jku: I can't think of anything else we would want to expose... we probably want to add a `Key.from_securesystemslib_key()` still but even there we likely will just assign the attributes and that's it

jku:
Last push: Modified Role docstring based on review

tuf/api/metadata.py
```diff
@@ -512,15 +504,15 @@ def verify_signature(
...
+
+    Role defines how many keys are required to successfully sign the roles
+    metadata, and which keys are accepted.
 
     Attributes:
-        keyids: A set of strings each of which represents a given key.
-        threshold: An integer representing the required number of keys for that
-            particular role.
+        keyids: A set of strings representing signing keys for this role.
+        threshold: Number of keys required to sign this roles metadata.
```
```suggestion
        threshold: Number of keys required to sign this role's metadata.
```

tuf/api/metadata.py
```diff
@@ -786,13 +754,11 @@ def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]):
...
-        meta: A dictionary that contains information about snapshot metadata::
-
-            {
-                'snapshot.json': <MetaFile INSTANCE>
-            }
+    Timestamp contains information about the snapshot Metadata file.
 
+    Attributes:
+        meta: A dictionary of filenames to MetaFiles. Only valid key value is
+            the snapshot filename, as defined by specification.
```
```suggestion
        meta: A dictionary of filenames to MetaFiles. The only valid key value
            is the snapshot filename, as defined by the specification.
```
IMNSHO these additional words increase ease of reading without increasing line count, but feel free to ignore.

tuf/api/metadata.py
```diff
@@ -834,15 +800,10 @@ def update(self, snapshot_meta: MetaFile) -> None:
...
-
-            {
-                'targets.json': <MetaFile INSTANCE>,
-                '<DELEGATED TARGETS ROLE 1>.json': <MetaFile INSTANCE>,
-                '<DELEGATED TARGETS ROLE 2>.json': <MetaFile INSTANCE>,
-            }
+    Snapshot contains information about all target Metadata files.
 
+    Attributes:
+        meta: A dictionary of target metadata filenames to MetaFiles.
```
```suggestion
        meta: A dictionary of target metadata filenames to MetaFile objects.
```
Wanted to preserve the keyword MetaFile

tuf/api/metadata.py
```diff
@@ -891,18 +852,14 @@ class DelegatedRole(Role):
...
         terminating: A boolean indicating whether subsequent delegations
-            should be considered.
+            should be considered during a target lookup.
         paths: An optional list of strings, where each string describes
-            a path that the role is trusted to provide.
+            a path pattern that the role is trusted to provide.
         path_hash_prefixes: An optional list of HEX_DIGESTs used to succinctly
             describe a set of target paths. Only one of the attributes "paths"
-            and "path_hash_prefixes" is allowed to be set.
+            and "path_hash_prefixes" is allowed within a DelegatedRole.
```
joshuagl: I wonder if there's a clearer way to document that paths and paths_hash_prefixes are mutually exclusive?

tuf/api/metadata.py
```diff
@@ -965,12 +922,11 @@ class Delegations:
...
 
     Attributes:
-        keys: A dictionary of keyids and key objects containing information
-            about the corresponding key.
-        roles: A list of DelegatedRole instances containing information about
-            all delegated roles.
+        keys: Dictionary of keyids to Keys. Defines the keys used in 'roles'.
+        roles: List of DelegatedRoles that define which keys are required to
+            sign the metadata for a specific role. The roles order also
+            defines the order the role delegations are considered.
```
```suggestion
            defines the order that role delegations are considered in.
```
?

I wanted to say "in which the role delegations are considered" but thought you might object to the character count increase.

tuf/api/metadata.py
```diff
@@ -1082,25 +1030,18 @@ def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]):
...
-            {
-                '<TARGET FILE NAME>': <TargetFile INSTANCE>,
-                ...
-            }
-
-        delegations: An optional object containing a list of delegated target
-            roles and public key store used to verify their metadata
-            signatures.
+    Targets contains verifying information about target files but also
+    delegates responsibility to other Targets files.
```
```suggestion
    Targets contains verifying information about target files and also
    delegates responsibility to other Targets roles.
```

joshuagl (commented):
Excellent work, thanks for doing this. ~60 fewer LOC and increases the clarity of the code comments.

I made a few minor suggestions, what do you think?

tuf/api/metadata.py
jku: I guess if we consider actual code changes we could just have a single list[str] attribute and then a way to say "does this DelegatedRole use hash prefixes Y/N?", but however you implement it (a flag attribute or even two separate classes) it it's going to look quite different compared to the file format ... so does not seem worth it

I think this might be worth pointing out in the body text of this class docstring... I'll try that

jku:
squashed joshuas suggestions (thanks!), then added another commit to improve the DelegatedRole docstring: It explains all of the options that exist in broad terms and expects reader to lookup the details (like what exactly is a "target path hash") elsewhere, and tries to make it clearer that the attributes are mutually exclusive.

joshuagl (approved):
LGTM, thanks!

PR #1457, opened 2021-06-21
Improve signed typing

### LINKED ISSUE #1433: Metadata API: Improve metadata.signed "typing"
There's two case I'd like to see improved in metadata signed handling:
1. Usually I call Metadata factory constructors (e.g. `Metadata.from_bytes()`) knowing what signed type should be, yet I can't enforce this
2. Using metadata.signed in a type annotated fashion (so that it's annotated as the actual type: Root/Snapshot/Timestamp/Targets) would be fantastic.

Item 1 could be fixed with just adding an optional type argument to the constructors (`Metadata.from_bytes(data, type=Root)`) and raising if it does not match.

But I wonder if some Python wizard could fix both by making Metadata generic over Root/Snapshot/Timestamp/Targets:
```
try:
    root = Metadata[Root].from_bytes(data)
except DeserializationError as e:
    # this happens (among other things) if signed type does not match Root
    raise exceptions.RepositoryError("Failed to load root") from e 

for key in root.signed.keys.items():
    # MAGIC!: keys, key and keytype are all annotated here because root
    # is Metadata[Root] and so root.signed is Root
    print(key.keytype) 
```

jku:
_EDIT: the current proposal only contains generic annotations, not the runtime enforcing that is also described below. I'm leaving the original text here for posterity, but see the commit message to see current state._

---

This is a PR but I invite discussion about whether it is a good idea: I implemented it first because I did not know what it would end up looking like so discussion before implementation seemed futile.

Please see[ comment below](https://github.com/theupdateframework/tuf/pull/1457#issuecomment-866588434) for detailed reasoning and a comparison with alternatives.

copying description from one of the commits:

    The purpose is two-fold:
    1. When we deserialize metadata, we usually know what signed type we
       expect: make it easy to enforce that
    2. When we use Metadata, it is helpful if the specific signed type (and
       all of the classes attribute types are correctly annotated
    
    Making Metadata Generic over T, where
    
        T = TypeVar("T", "Root", "Timestamp", "Snapshot", "Targets")
    
    allows both of these cases to work. Using Generics is completely
    optional so all existing code still works. For case 1, the following
    calls will now raise a Deserialization error if the expected type is
    incorrect:
    
        md = Metadata.from_bytes(data, signed_type=Snapshot)
        md = Metadata.from_file(filename, signed_type=Snapshot)
    
    For case 2, the return value md of those calls is now of type
    "Metadata[Snapshot]", and md.signed is now of type "Snapshot" allowing
    IDE annotations and static type checking.
    
    Adding a type argument is an unconventional way to do this: the reason
    for it is that the specific type (e.g. Snapshot) is not otherwise
    available at runtime. A call like this works fine and md is annotated:
    
        md = Metadata[Snapshot].from_bytes(data)
    
    but it's not possible to validate that "data" contains a "Snapshot",
    because the value "Snapshot" is not defined at runtime at all, it is
    purely an annotation. So an actual argument is needed.
    
Fixes #1433

requirements**:

- [] Docs have been added for the bug fix or new feature

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -26,12 +26,15 @@ from typing import (
     BinaryIO,
     ClassVar,
     Dict,
+    Generic,
     List,
     Mapping,
     Optional,
     Tuple,
     Type,
+    TypeVar,
     Union,
+    cast,
 )
 
 from securesystemslib import exceptions as sslib_exceptions
@@ -56,12 +59,30 @@ logger = logging.getLogger(__name__)
 # files to have the same major version (the first number) as ours.
 SPECIFICATION_VERSION = ["1", "0", "19"]
 
+# T is a Generic type constraint for Metadata.signed
+T = TypeVar("T", "Root", "Timestamp", "Snapshot", "Targets")
 
-class Metadata:
+
+class Metadata(Generic[T]):
     """A container for signed TUF metadata.
 
     Provides methods to convert to and from dictionary, read and write to and
     from file and to create and verify metadata signatures.
+
+    Metadata[T] is a generic container type where T can be any one type of
+    [Root, Timestamp, Snapshot, Targets]. The purpose of this is to allow
+    static type checking of the signed attribute in code using Metadata::
+
+        root_md = Metadata[Root].from_file("root.json")
+        # root_md type is now Metadata[Root]. This means signed and its
+        # attributes like consistent_snapshot are now statically typed and the
+        # types can be verified by static type checkers and shown by IDEs
+        print(root_md.signed.consistent_snapshot)
+
+    Using a type constraint is not required but not doing so means T is not a
+    specific type so static typing cannot happen. Note that the type constraint
+    "[Root]" is not validated at runtime (as pure annotations are not available
+    then).
 
     Attributes:
         signed: A subclass of Signed, which has the actual metadata payload,
@@ -70,10 +91,8 @@ class Metadata:
             signing the canonical serialized representation of 'signed'.
     """
 
-    def __init__(
-        self, signed: "Signed", signatures: "OrderedDict[str, Signature]"
-    ):
-        self.signed = signed
+    def __init__(self, signed: T, signatures: "OrderedDict[str, Signature]"):
+        self.signed: T = signed
         self.signatures = signatures
 
     @classmethod
@@ -119,7 +138,8 @@ class Metadata:
             signatures[sig.keyid] = sig
 
         return cls(
-            signed=inner_cls.from_dict(metadata.pop("signed")),
+            # Specific type T is not known at static type check time: use cast
+            signed=cast(T, inner_cls.from_dict(metadata.pop("signed"))),
             signatures=signatures,
         )
 
@@ -129,7 +149,7 @@ class Metadata:
         filename: str,
         deserializer: Optional[MetadataDeserializer] = None,
         storage_backend: Optional[StorageBackendInterface] = None,
-    ) -> "Metadata":
+    ) -> "Metadata[T]":
         """Loads TUF metadata from file storage.
 
         Arguments:
@@ -156,11 +176,12 @@ class Metadata:
         with storage_backend.get(filename) as file_obj:
             return cls.from_bytes(file_obj.read(), deserializer)
 
-    @staticmethod
+    @classmethod
     def from_bytes(
+        cls,
         data: bytes,
         deserializer: Optional[MetadataDeserializer] = None,
-    ) -> "Metadata":
+    ) -> "Metadata[T]":
         """Loads TUF metadata from raw data.
 
         Arguments:
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -183,7 +183,7 @@ class TestMetadata(unittest.TestCase):
 
     def test_sign_verify(self):
         root_path = os.path.join(self.repo_dir, 'metadata', 'root.json')
-        root:Root = Metadata.from_file(root_path).signed
+        root = Metadata[Root].from_file(root_path).signed
 
         # Locate the public keys we need from root
         targets_keyid = next(iter(root.roles["targets"].keyids))
@@ -302,7 +302,7 @@ class TestMetadata(unittest.TestCase):
     def test_metadata_snapshot(self):
         snapshot_path = os.path.join(
                 self.repo_dir, 'metadata', 'snapshot.json')
-        snapshot = Metadata.from_file(snapshot_path)
+        snapshot = Metadata[Snapshot].from_file(snapshot_path)
 
         # Create a MetaFile instance representing what we expect
         # the updated data to be.
@@ -321,7 +321,7 @@ class TestMetadata(unittest.TestCase):
     def test_metadata_timestamp(self):
         timestamp_path = os.path.join(
                 self.repo_dir, 'metadata', 'timestamp.json')
-        timestamp = Metadata.from_file(timestamp_path)
+        timestamp = Metadata[Timestamp].from_file(timestamp_path)
 
         self.assertEqual(timestamp.signed.version, 1)
         timestamp.signed.bump_version()
@@ -358,19 +358,19 @@ class TestMetadata(unittest.TestCase):
 
     def test_metadata_verify_delegate(self):
         root_path = os.path.join(self.repo_dir, 'metadata', 'root.json')
-        root = Metadata.from_file(root_path)
+        root = Metadata[Root].from_file(root_path)
         snapshot_path = os.path.join(
                 self.repo_dir, 'metadata', 'snapshot.json')
-        snapshot = Metadata.from_file(snapshot_path)
+        snapshot = Metadata[Snapshot].from_file(snapshot_path)
         targets_path = os.path.join(
                 self.repo_dir, 'metadata', 'targets.json')
-        targets = Metadata.from_file(targets_path)
+        targets = Metadata[Targets].from_file(targets_path)
         role1_path = os.path.join(
                 self.repo_dir, 'metadata', 'role1.json')
-        role1 = Metadata.from_file(role1_path)
+        role1 = Metadata[Targets].from_file(role1_path)
         role2_path = os.path.join(
                 self.repo_dir, 'metadata', 'role2.json')
-        role2 = Metadata.from_file(role2_path)
+        role2 = Metadata[Targets].from_file(role2_path)
 
         # test the expected delegation tree
         root.verify_delegate('root', root)
@@ -468,7 +468,7 @@ class TestMetadata(unittest.TestCase):
     def test_metadata_root(self):
         root_path = os.path.join(
                 self.repo_dir, 'metadata', 'root.json')
-        root = Metadata.from_file(root_path)
+        root = Metadata[Root].from_file(root_path)
 
         # Add a second key to root role
         root_key2 =  import_ed25519_publickey_from_file(
@@ -530,7 +530,7 @@ class TestMetadata(unittest.TestCase):
     def test_metadata_targets(self):
         targets_path = os.path.join(
                 self.repo_dir, 'metadata', 'targets.json')
-        targets = Metadata.from_file(targets_path)
+        targets = Metadata[Targets].from_file(targets_path)
 
         # Create a fileinfo dict representing what we expect the updated data to be
         filename = 'file2.txt'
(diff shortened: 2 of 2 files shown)


jku:
To get a real-world impression of how this would help us, see e.g. https://github.com/theupdateframework/tuf/blob/74fd891677817320a2f5701d436ac9f98161bc18/tuf/ngclient/_internal/metadata_bundle.py#L432
* Many of the variables are currently not statically typed (so can't be type checked)
* there's five lines in every method that raises if `new_delegate.signed.type != "targets"`

tuf/api/metadata.py
```diff
@@ -183,7 +194,14 @@ def from_bytes(
 
             deserializer = JSONDeserializer()
 
-        return deserializer.deserialize(data)
+        md = deserializer.deserialize(data)
+
+        # Ensure deserialized signed type matches the requested type
+        if signed_type is not None and signed_type != type(md.signed):
```
sechkova: Is `Generic` needed for the `signed_type` check to work or they serve two different purposes (independent of each other)?
- making Metadata class Generic over T for type checking
- signed_type argument for runtime check

jku: the runtime check certainly works without any generics, yes.

They are still related in that providing `signed_type` also defines what the specific type T is: if the signed_type argument is not used, some other means needs to be used to get a specific type for static typing (like the syntax `md = Metadata[Snapshot].from_bytes(data)`)

joshuagl:
Few observations to stimulate discussion, as I'm heading AFK for a few days.
1. This is just complex and controversial enough that it might be worth creating an ADR
2. `T` is not very descriptive, can we do better? Should we?
3. Are generics available in all of the Python versions we want to support?

jku:
> 1. This is just complex and controversial enough that it might be worth creating an ADR

This is probably true...

> 2. `T` is not very descriptive, can we do better? Should we?

We can (I don't object to it) but
* we can't call it Signed since we already have a class of that name so it would have to be SignedType or something
* Single letter type aliases are very common (I guess because usually there is no good name for them)

> 3. Are generics available in all of the Python versions we want to support?

They were included along with other annotations in 3.5. I don't see anything important in the API that has changed since then. 

I believe the performance of some operations in `typing` has dramatically improved over time (as the implementation has moved more into CPython core) but our uses cases are unlikely to ever notice that in any way.

jku:
So I wrote an ADR, but then started thinking is this really appropriate? It's a one-off decision as far as I can tell, not something that needs to be referred to to later on...

Anyway, I think the text may be useful so here it is:

# ADR: Use Generics in Metadata API to enable better static typing

Technical Story:
    https://github.com/theupdateframework/tuf/issues/1433
    https://github.com/theupdateframework/tuf/pull/1457

## Context and Problem Statement

Metadata API is type annotated quite well. The annotations are a major benefit to both users and developers of the API as they enable:
* static type checking run by developers and by CI: this prevents errors from getting integrated into the code base and improves the code quality
* productivity improving IDE integrations (live type checking in IDE)

The Metadata API design naturally leads to usages like this:

    root_md = Metadata.from_file("root.json")
    for keyid in root_md.signed.role.keyids:
        key = root_md.signed.keys[keyid]
        key.verify_signature(unverified_metadata)

The issue here is that while the developer knows from context that signed type is (or should be) Root, static typing tools cannot know this (and it is not verified at runtime). Because signed type is not known, almost no types in above example are known, and the protections of static type checking are not available.

## Considered Options

* Document other usage patterns
* Add More Metadata-class derivatives
* Make Metadata a Generic container

## Decision Outcome

Chosen option: **Make Metadata a Generic container**

The slight increase in code complexity is outweighed by the advantages:
* Existing code continues to work
* static typing starts working with minimal code changes
* deserialized metadata types can be checked at runtime as well
* API surface does not grow meaningfully

For users of the API the change is either invisible (as the feature is completely optional) or easy to use: It typically requires either
* Deserializing with e.g. `Root.metadata_from_*()` instead of `Metadata.from_*()`
* if Metadata needs to be annotated, using the format familiar from standard library containers: e.g. `Metadata[Root]`

## Pros and Cons of the Options

### Document other usage patterns

We could tell users to do this instead:

    root_md = Metadata.from_file("root.json")
    root:Root = root_md.signed
    for keyid in root.role.keyids:
        key = root.keys[keyid]
        key.verify_signature(unverified_metadata)

This does let static type checker see the types here, but unfortunately the types are completely based on the type specified by the programmer. Also, the type chosen by the programmer cannot be verified at runtime in any way.

### Add more Metadata-class derivatives

We could add four new classes (RootMetadata, SnapshotMetadata, ...) that derive from Metadata and that define their signed attributes with the correct type. This would allow creating objects that are type checkable:

    root_md = RootMetadata.from_file("root.json")
    for keyid in root_md.signed.role.keyids:
        key = root_md.signed.keys[keyid]
        key.verify_signature(unverified_metadata)

This also allows adding code in the new classes deserialization paths that would raise if the type in "root.json" is not what was expected.

The downside here is adding four new classes to the most visible part of the API for practically no runtime purpose at all.

### Make Metadata a Generic container

Making Metadata a Generic container (much like List or Dict) allows static typing to automatically work. While Metadata is "generic" it is also constrained to contain only one of the four types we know are valid. Types in the example below are fully understood by static type checker:

    md = Root.metadata_from_file("root.json")
    for keyid in md.signed.role.keyids:
        key = md.signed.keys[keyid]
        key.verify_signature(unverified_metadata)

This also allows the `Signed.metadata_from_*()` constructors to verify the type at runtime.

The downsides are 
* using an advanced Python feature, generics, in a code base that is mostly quite simple: while generics is well used in the Python standard library (all containers are Generic) it's not very common outside it.
* The `Signed.metadata_from_*()` constructors do not add new code but there is a lot of boilerplate (signatures and docstrings)

jku:
Added an explanation in Metadata docstring

jku:
Based on discussion on slack, I added a commit that moves the "typed constructors" to Signed, so

    root_md = Root.metadata_from_file("root.json")

Functionality is the same as before but now we do not need the ugly signed_type variable. Using the API now looks quite good IMO (and Metadata.from_file() still exists so "unknown" metadata can be loaded if the need arises)

There's not any more code but unfortunately adding two abstract methods to Signed means a lot of boilerplate (10 method signatures and docstrings) ... I think this is still better than adding new Metadata-derivatives

I also edited the ADR-lookalike comment above to reflect the current style

sechkova (commented):
I like how Generics help with determining the exact "Signed" type and I also prefer the changes from the latest commit rather than having an additional argument.

```
Before:
    root_md = Metadata._from_file(filename, signed_type=Root)
after:
    root_md = Root.metadata_from_file(filename)
```

Unfortunately it seems like we have already started abusing the call stack when constructing a Metadata object. It now roughly looks like this:

```
Root.metadata_from_bytes()
-> Metadata.from_bytes()
    -> deserializer.deserialize()
        -> Metadata.from_dict()
            ...
            _type = metadata["signed"]["type"]
            inner_cls = Root
            -> Root.from_dict()
             # Here we do another check of _type

       # return Metadata object
  
   # back to Root.metadata_from_bytes()
   if not isinstance (metadata.signed, cls)
       raise
   return Metadata
```

One option is to revert to the initial form of `metadata` proposed while it was still under development and have all methods creating metadata objects as part of `Signed`. The call stack will remain similar but we get rid of the hoping from `Signed` to `Metadata` and back. This means an inconsistency in the to/from methods pairs, though:

```
class Metadata(Generic[T]):
     sign()
     verify_delegate()
     to_dict()
     to_file()
    
class Signed:
     metadata_from_dict() -> Metadata
     metadata_from_file() -> Metadata
     metadata_from_bytes() -> Metadata
     ...
```
```
Root.metadata_from_bytes()
    -> deserializer.deserialize() # need to pass cls type to deserializer
        -> Root.from_dict()
           # Signed type check here
   return Metadata
```
I haven't actually tested this proposal so maybe there are pitfalls that I haven't spotted.

tuf/api/metadata.py
```diff
@@ -127,7 +124,8 @@ def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata":
             signatures[sig.keyid] = sig
 
         return cls(
-            signed=inner_cls.from_dict(metadata.pop("signed")),
+            # Specific type T is not known at static type check time: cast
+            signed=cast(T, inner_cls.from_dict(metadata.pop("signed"))),
```
MVrachev, line 142: From what I understand, you are using `cast` here to signal the type checker that you expect the type of the returned metadata.signed to be of type `T`
Wondering when does `T` gets actually assigned?

jku:
> Root.metadata_from_bytes()
>    -> deserializer.deserialize() # need to pass cls type to deserializer

Jumping over `Metadata.from_bytes()` like this is totally possible, but it does mean that there are now 5 (or 9?) places that initialize a default `JsonDeserializer()` instead of one place.

> Root.metadata_from_bytes()
>     -> deserializer.deserialize() # need to pass cls type to deserializer
>         -> Root.from_dict()
>            # Signed type check here
>    return Metadata

You still can't remove the last type check in `Root.metadata_from_bytes()`: deserializer returns any type -- whatever the data contains (could be Root, but could be something else). The check in `_common_fields_from_dict()` is just a sanity check that the Signed type and _type match, it does not check that the type is what we wanted.

MVrachev (commented):
I liked that we were able to remove the `signed_type` argument.
Still, I agree we have soo many `_type` checks...
I just remembered we added additional  `_type` checks in the `TrustedMetadataSet`...
Maybe this is a good time to rethink how we do initialization through `Metadata_from_bytes()` as Teodora suggested.

tuf/api/metadata.py
jku: well, it's useful to remember that `T` and `cast()` have zero runtime effect: cast() is me promising to _mypy_ (not cpython) that signed will be one of the types T once the code runs, and telling mypy to not worry about it. The cpython runtime doesn't know anything about T or that promise.

For the static typing to work `__init__() / from_dict()` themselves do not need to define what type T is: calling methods like `Root.metadata_from_file()` do define it (as the return value is `Metadata[Root]`).

jku: You can see this in action if you use the current constructor with this branch: `Metadata.from_file()` still works but return value is not statically typed with the correct type (it's just Metadata[T]) because there is nothing the static type check could use to figure it out.

conveniently at least VS Code offers me all the possible completions (Root,Targets,Snapshot,Timestamp) when T is not defined so that's a win-win.

jku:
The issues in Teodoras mypy branch are a pretty good example of what this is about:
https://github.com/theupdateframework/tuf/pull/1489#issuecomment-878063847

It also gives us a place to see the alternatives. There are surprisingly few places where this is an issue in ngclient after all: maybe 10-15 well placed calls like `assert(isinstance(self._trusted_set.snapshot.signed, Snapshot))` would make the checks work... it just looks so ugly

joshuagl:
> The issues in Teodoras mypy branch are a pretty good example of what this is about:
> [#1489 (comment)](https://github.com/theupdateframework/tuf/pull/1489#issuecomment-878063847)
> 
> It also gives us a place to see the alternatives. There are surprisingly few places where this is an issue in ngclient after all: maybe 10-15 well placed calls like `assert(isinstance(self._trusted_set.snapshot.signed, Snapshot))` would make the checks work... it just looks so ugly

Given that the alternative (`assert()`s) offloads responsibility to the API user, I am not a fan of that. The fact that it looks ugly makes the option even easier to argue against.

The work in this PR to use generics feels like the best solution we have to a problem we want to solve, that is – ensuring that the API is as easy to use as possible. I'm not entirely fond of the contained object constructing the container, but I don't yet have a better alternative to propose.

I will try to spend some time with this PR and give it a more thorough review next week.

sechkova:
`bandit` doesn't fully agree with asserts too:
https://github.com/theupdateframework/tuf/pull/1489#issuecomment-886744534

jku:
New attempt: leave out the runtime type checks, just include the minimal generics:

* this allows using static typing using multiple annotation methods:
`md = Metadata[Root].from_bytes(data)`
`md:Metadata[Root] = Metadata.from_bytes(data)`
* this does not enforce that the type constraint `[Root]` is correct at runtime.

If this seems reasonable I can make another PR with the runtime-type-checking Metadata constructors

joshuagl (approved):
Thanks for narrowing scope to make the PR easier to reason about. LGTM.

PR #1460, opened 2021-06-22
Metadata API: test invalid input values for attributes

### LINKED ISSUE #1434: Metadata API: Tests (de)serialization with invalid arguments
I started creating issues documenting some research I do for how we use our metadata attributes.
For example #1418, #1419 and #1420. 
The end goal of this research is to decide how we are going to validate each of them, make the necessary changes and add tests ensuring the validation works.

Together with @jku we discussed that to achieve the last step - adding tests ensuring the validation works
it will be better if we wait for #1391 to be resolved.
That way we would come up with a better-refined approach of running a test case against multiple inputs. 

Also, I wanted to point out that there was a suggestion to use a fuzzing tool for this.
Maybe it's worth exploring.

We will want to test cases when our attributes are:

- [ ] empty values
- [ ] values from the wrong type
- [ ] non existent - they don't appear in the input dictionary when calling `from_dict()`

MVrachev:
Fixes #1434

A while ago we decided that it's best to research each of the individuals
attributes one by one and identify what level of validation it needs
compared to how we use it:
#1366 (comment).

This work is ongoing and there are a couple of commits already merged
for this:
- 6c5d970
- f20664d
- 41afb1e

We want to be able to test the validation we add for attributes against known bad
values. The way we want to do that is with table testing we have added
using decorators for our metadata classes defined in New API:
#1416. 
This gives us an easy way to add new cases for each of the attributes and
not depend on external files.

_**It's important to note that I haven't added invalid tests for all attributes, because
we haven't finished with the metadata research.**_

When new validation for a certain attribute is added we should make sure to add
invalid tests in `TestInvalidSerialization` as well.

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -12,11 +12,13 @@ import logging
 import unittest
 import copy
 
-from typing import Dict, Callable
+from typing import Dict, Callable, Optional, Mapping, Any
+from datetime import datetime
 
 from tests import utils
 
 from tuf.api.metadata import (
+    Signed,
     Root,
     Snapshot,
     Timestamp,
@@ -50,6 +52,44 @@ def run_sub_tests_with_dataset(dataset: DataSet):
 
 class TestSerialization(unittest.TestCase):
 
+    # Snapshot instances with meta = {} are valid, but for a full valid
+    # repository it's required that meta has at least one element inside it.
+    invalid_signed: DataSet = {
+        "no _type": '{"spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no spec_version": '{"_type": "signed", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no version": '{"_type": "signed", "spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no expires": '{"_type": "signed", "spec_version": "1.0.0", "version": 1, "meta": {}}',
+        "empty str _type":
+            '{"_type": "", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "empty str spec_version":
+            '{"_type": "signed", "spec_version": "", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "_type wrong type":
+            '{"_type": "foo", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "version wrong type":
+            '{"_type": "signed", "spec_version": "1.0.0", "version": "a", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "invalid spec_version str":
+            '{"_type": "signed", "spec_version": "abc", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "two digit spec_version":
+            '{"_type": "signed", "spec_version": "1.2.a", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no digit spec_version":
+            '{"_type": "signed", "spec_version": "a.b.c", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "different major spec_version":
+            '{"_type": "signed", "spec_version": "0.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "version 0":
+            '{"_type": "signed", "spec_version": "1.0.0", "version": 0, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "version below 0":
+            '{"_type": "signed", "spec_version": "1.0.0", "version": -1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "wrong datetime string":
+            '{"_type": "signed", "spec_version": "1.0.0", "version": 1, "expires": "abc", "meta": {}}',
+    }
+
+    @run_sub_tests_with_dataset(invalid_signed)
+    def test_invalid_signed_serialization(self, test_case_data: Dict[str, str]):
+        case_dict = json.loads(test_case_data)
+        with self.assertRaises((KeyError, ValueError, TypeError)):
+            Snapshot.from_dict(copy.deepcopy(case_dict))
+
+
     valid_keys: DataSet = {
         "all": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", \
             "keyval": {"public": "foo"}}',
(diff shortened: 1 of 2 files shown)


MVrachev:
This pr is ready for review after the merge of #1416.

jku (changes requested):
This seems very fine grained. There's no benefit from having a separate test for each attribute (when tests are practically identical) but it makes the patch massive. I would expect less granularity would mean at least 50% reduction in commit size?

Also, currently it's hard to see what is being tested: I need to read deep into the test code to find what is the actual class that is being tested by an individual test.

I don't see why we would not do the exact same thing we already do for valid data tests: one invalid-data dataset and one new test per class-under-test. I also think the invalid test+data for a class should be next to the valid test+data for the same class -- If the tests need to be split by for some reason, I expect it to be more useful to split by class-under-test (so e.g. TestKeySerialization) than splitting by TestSerialization/TestInvalidSerialization: I often want to run tests for a specific class, but I have never wanted to run tests based whether success or failure is expected.

MVrachev:
> This seems very fine grained. There's no benefit from having a separate test for each attribute (when tests are practically identical) but it makes the patch massive. I would expect less granularity would mean at least 50% reduction in commit size?
> 
> Also, currently it's hard to see what is being tested: I need to read deep into the test code to find what is the actual class that is being tested by an individual test.
> 
> I don't see why we would not do the exact same thing we already do for valid data tests: one invalid-data dataset and one new test per class-under-test. I also think the invalid test+data for a class should be next to the valid test+data for the same class -- If the tests need to be split by for some reason, I expect it to be more useful to split by class-under-test (so e.g. TestKeySerialization) than splitting by TestSerialization/TestInvalidSerialization: I often want to run tests for a specific class, but I have never wanted to run tests based whether success or failure is expected.

I did it more granularly because we have more cases per attribute.
Will create a commit making test cases per class and we will comment it again.

MVrachev:
In a discussion with Jussi, he proposed that I should group invalid and valid tests close to each other in the same test class.
Added a commit doing that.

tests/test_metadata_serialization.py
```diff
@@ -62,6 +117,41 @@ def test_key_serialization(self, test_case_data: str):
         self.assertDictEqual(case_dict, key.to_dict())
 
 
+    invalid_key: DataSet = {
+        "no keyid": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "abc"}}',
+        "no keytype": '{"keyid": "id", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}',
+        "no scheme": '{"keyid": "id", "keytype": "rsa", "keyval": {"public": "foo"}}',
+        "no keyval": '{"keyid": "id", "keytype": "rsa", "scheme": "rsassa-pss-sha256"}',
+        "no public in keyval": '{"keyid": "id", "keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {}}',
```
jku: I believe this should not be invalid (I mean rsa keys do require a "public" but key in general should not be required to have one, this is my reading of the spec)

Maybe file an issue on this?

tests/test_metadata_serialization.py
```diff
@@ -92,6 +182,25 @@ def test_root_serialization(self, test_case_data: str):
...
         self.assertDictEqual(case_dict, root.to_dict())
 
+
+    invalid_metafile_attr: DataSet = {
+        "wrong length type": '{"version": 1, "length": "a", "hashes": {"sha256" : "abc"}}',
+        "length 0": '{"version": 1, "length": 0, "hashes": {"sha256" : "abc"}}',
+        "length below 0": '{"version": 1, "length": -1, "hashes": {"sha256" : "abc"}}',
+        "empty hashes dict": '{"version": 1, "length": 1, "hashes": {}}',
+        # Hashes keys cannot be strings because the JSON specification.
+        # Then, do we want to test for that?
```
jku: this does not seem to belong in the test data (but I also don't understand the comment)

tests/test_metadata_serialization.py
```diff
@@ -62,6 +117,41 @@ def test_key_serialization(self, test_case_data: str):
...
+        "no keyid": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "abc"}}',
+        "no keytype": '{"keyid": "id", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}',
+        "no scheme": '{"keyid": "id", "keytype": "rsa", "keyval": {"public": "foo"}}',
+        "no keyval": '{"keyid": "id", "keytype": "rsa", "scheme": "rsassa-pss-sha256"}',
+        "no public in keyval": '{"keyid": "id", "keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {}}',
+        "keyid wrong type": '{"keyid": 1, "keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "abc"}}',
+        "keytype wrong type": '{"keyid": "id", "keytype": 1, "scheme": "rsassa-pss-sha256", "keyval": {"public": "abc"}}',
+        "scheme wrong type": '{"keyid": "id", "keytype": "rsa", "scheme": 1, "keyval": {"public": "abc"}}',
+        "keyval wrong type": '{"keyid": "id", "keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": 1}',
+        "keyval public wrong type": '{"keyid": "id", "keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": 1}}',
```
jku: same as above: this should not be invalid for key in general

tests/test_metadata_serialization.py
```diff
@@ -48,8 +50,61 @@ def wrapper(test_cls: "TestSerialization"):
...
+        return cls(*common_args, signed_dict)
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Returns the dict representation of self."""
+        return super()._common_fields_to_dict()
+
+
 class TestSerialization(unittest.TestCase):
 
+    invalid_signed: DataSet = {
```
jku, line 57: there's a lot of test cases here that are unlikely to be useful (e.g. 5 invalid spec versions  is just overkill) but since the tests are so fast... I guess no harm done?

tests/test_metadata_serialization.py
```diff
@@ -48,8 +50,61 @@ def wrapper(test_cls: "TestSerialization"):
     return real_decorator
 
 
+class TestSigned(Signed):
```
jku: I think creating a class just for testing it feels wrong. I would prefer using an existing Signed-derivative for testing these fields, but it's your choice.

jku (commented):
I would still really like it if the variable and function names were consistent

> invalid_key vs valid_key**s**
> invalid_role vs valid_role**s**
> invalid_metafile_attr vs valid_metafiles
> invalid_targetfile_hashes_length vs valid_targetfiles

With some of those I really need to squint to figure out that they are supposed to be tests for the same class

Other comments are just nitpicks I think

tests/test_metadata_serialization.py
MVrachev: I added it because this is the current behavior of the code.
Will remove it and create a pr about it.

tests/test_metadata_serialization.py
MVrachev: Oh... Sorry, it was a comment to myself and forgot to remove it.

tests/test_metadata_serialization.py
MVrachev: I removed one of the cases `one digit spec_version`, the others seem relevant to the checks we are making.

tests/test_metadata_serialization.py
MVrachev: I removed one of the cases one digit spec_version, the others seem relevant to the checks we are making.

tests/test_metadata_serialization.py
MVrachev: I will remove it, but as a side effect, the `invalid_signed` tests will become longer.

tests/test_metadata_serialization.py
MVrachev: https://github.com/theupdateframework/tuf/pull/1480

tests/test_metadata_serialization.py
MVrachev: Removed it and filed a pr: https://github.com/theupdateframework/tuf/pull/1480

MVrachev:
@jku squashed my commits and addressed all of your comments (including the inconsistencies) except for the ones
about `keyval` which are addressed in another pr.

tests/test_metadata_serialization.py
```diff
@@ -159,6 +247,20 @@ def test_delegation_serialization(self, test_case_data: str):
         self.assertDictEqual(case_dict, delegation.to_dict())
 
 
+    invalid_targetfiles: DataSet = {
+        "no hashes": '{"length": 1}',
+        "no length": '{"hashes": {"sha256": 1}}'
```
jku: that's an invalid hashes value type

tests/test_metadata_serialization.py
```diff
@@ -50,18 +52,87 @@ def wrapper(test_cls: "TestSerialization"):
...
 
+    invalid_signed: DataSet = {
+        "no _type": '{"spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {"f.txt": {"version": 1}}}',
+        "no spec_version": '{"_type": "signed", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {"f.txt": {"version": 1}}}',
+        "no version": '{"_type": "signed", "spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {"f.txt": {"version": 1}}}',
+        "no expires": '{"_type": "signed", "spec_version": "1.0.0", "version": 1, "meta": {"f.txt": {"version": 1}}}',
+        "empty str _type": '{"_type": "", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
+            "meta": {"f.txt": {"version": 1}}}',
+        "empty str spec_version": '{"_type": "signed", "spec_version": "", "version": 1, "expires": "2030-01-01T00:00:00Z", \
+            "meta": {"f.txt": {"version": 1}}}',
```
jku: Snapshot meta isn't required to contain anything: `"meta": {}` is enough if you're going for the minimal serialised form

nitpick on the style: I think this would be easier to read if you just barely need two lines (also allows names to be as long as you want):
```
        "empty str spec_version":
            '{"_type": "signed", "spec_version": "", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
```

tests/test_metadata_serialization.py
jku: ok, but do keep in mind that the idea isn't to test the lines of code that we happen to have in there right now: the idea is to have test data that covers as much of the possible inputs as practically useful -- since we can't cover 100% of the possibilities we inevitably end up making decisions on what's important: we test the inputs we think could be interesting.

jku (approved):
LGTM, left a couple of comments that are up to you

jku:
(that last comment is a reply to earlier discussion on spec_version test cases, for some reason github shows it independently here)

tests/test_metadata_serialization.py
MVrachev: Okay, will do that, but will add a comment that per spec having a snapshot object like this in a repository is considered invalid.

MVrachev:
@jku I addressed your comments.

MVrachev:
Rebased on top of the latest changes in develop branch and fixed conflicts.

PR #1463, opened 2021-06-23
ngclient: review preorder dfs code

### LINKED ISSUE #1398: ngclient: review _preorder_depth_first_walk()

sechkova:
Fixes #1398

 - Reviews and simplifies `Updater._preorder_depth_first_walk` code.
 - makes `_visit_child_role` a member of `DelegatedRole`

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -242,10 +242,9 @@ class TestSerialization(unittest.TestCase):
         "no path attribute":
             '{"keyids": ["keyid"], "name": "a", "terminating": false, \
             "path_hash_prefixes": ["h1", "h2"], "threshold": 99}',
-        "no hash or path prefix":
-            '{"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3}',
         "unrecognized field":
-            '{"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3, "foo": "bar"}',
+            '{"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], \
+            "terminating": true, "threshold": 3, "foo": "bar"}',
     }
 
     @run_sub_tests_with_dataset(valid_delegated_roles)
@@ -255,12 +254,27 @@ class TestSerialization(unittest.TestCase):
         self.assertDictEqual(case_dict, deserialized_role.to_dict())
 
 
+    invalid_delegated_roles: DataSet = {
+        "missing hash prefixes and paths":
+            '{"name": "a", "keyids": ["keyid"], "threshold": 1, "terminating": false}',
+        "both hash prefixes and paths":
+            '{"name": "a", "keyids": ["keyid"], "threshold": 1, "terminating": false, \
+            "paths": ["fn1", "fn2"], "path_hash_prefixes": ["h1", "h2"]}',
+    }
+
+    @run_sub_tests_with_dataset(invalid_delegated_roles)
+    def test_invalid_delegated_role_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        with self.assertRaises(ValueError):
+            DelegatedRole.from_dict(copy.copy(case_dict))
+
+
     valid_delegations: DataSet = {
         "all": '{"keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
-            "roles": [ {"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3} ]}',
+            "roles": [ {"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": true, "threshold": 3} ]}',
         "unrecognized field":
             '{"keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
-            "roles": [ {"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3} ], \
+            "roles": [ {"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": true, "threshold": 3} ], \
             "foo": "bar"}',
     }
 
@@ -305,13 +319,13 @@ class TestSerialization(unittest.TestCase):
             "targets": { "file.txt": {"length": 12, "hashes": {"sha256" : "abc"} } }, \
             "delegations": {"keys": {"keyid" : {"keytype": "rsa", \
                     "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"} }}, \
-                "roles": [ {"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3} ]} \
+                "roles": [ {"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": true, "threshold": 3} ]} \
             }',
         "empty targets": '{"_type": "targets", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
             "targets": {}, \
             "delegations": {"keys": {"keyid" : {"keytype": "rsa", \
                     "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"} }}, \
-                "roles": [ {"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3} ]} \
+                "roles": [ {"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": true, "threshold": 3} ]} \
             }',
         "no delegations": '{"_type": "targets", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
             "targets":  { "file.txt": {"length": 12, "hashes": {"sha256" : "abc"} } } \
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1038,7 +1038,12 @@ class DelegatedRole(Role):
         the paths that DelegatedRole is trusted to provide"""
 
         if self.path_hash_prefixes is not None:
-            target_filepath_hash = _get_filepath_hash(target_filepath)
+            # Calculate the hash of the filepath
+            # to determine in which bin to find the target.
+            digest_object = sslib_hash.digest(algorithm="sha256")
+            digest_object.update(target_filepath.encode("utf-8"))
+            target_filepath_hash = digest_object.hexdigest()
+
             for path_hash_prefix in self.path_hash_prefixes:
                 if target_filepath_hash.startswith(path_hash_prefix):
                     return True
@@ -1056,21 +1061,6 @@ class DelegatedRole(Role):
                     return True
 
         return False
-
-
-def _get_filepath_hash(target_filepath, hash_function="sha256"):
-    """
-    Calculate the hash of the filepath to determine which bin to find the
-    target.
-    """
-    # The client currently assumes the repository (i.e., repository
-    # tool) uses 'hash_function' to generate hashes and UTF-8.
-    digest_object = sslib_hash.digest(hash_function)
-    encoded_target_filepath = target_filepath.encode("utf-8")
-    digest_object.update(encoded_target_filepath)
-    target_filepath_hash = digest_object.hexdigest()
-
-    return target_filepath_hash
 
 
 class Delegations:
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -16,8 +16,10 @@ available in the class model.
 
 """
 import abc
+import fnmatch
 import io
 import logging
+import os
 import tempfile
 from collections import OrderedDict
 from datetime import datetime, timedelta
@@ -1030,6 +1032,64 @@ class DelegatedRole(Role):
         elif self.path_hash_prefixes is not None:
             res_dict["path_hash_prefixes"] = self.path_hash_prefixes
         return res_dict
+
+    def visit_child_role(self, target_filepath: str) -> str:
+        """Determines whether the given 'target_filepath' is an
+        allowed path of DelegatedRole"""
+
+        if self.path_hash_prefixes is not None:
+            target_filepath_hash = _get_filepath_hash(target_filepath)
+            for path_hash_prefix in self.path_hash_prefixes:
+                if not target_filepath_hash.startswith(path_hash_prefix):
+                    continue
+
+                return self.name
+
+        elif self.paths is not None:
+            for path in self.paths:
+                # A child role path may be an explicit path or glob pattern (Unix
+                # shell-style wildcards).  The child role 'child_role_name' is
+                # returned if 'target_filepath' is equal to or matches
+                # 'child_role_path'. Explicit filepaths are also considered
+                # matches. A repo maintainer might delegate a glob pattern with a
+                # leading path separator, while the client requests a matching
+                # target without a leading path separator - make sure to strip any
+                # leading path separators so that a match is made.
+                # Example: "foo.tgz" should match with "/*.tgz".
+                if fnmatch.fnmatch(
+                    target_filepath.lstrip(os.sep), path.lstrip(os.sep)
+                ):
+
+                    return self.name
+
+                continue
+
+        else:
+            # 'role_name' should have been validated when it was downloaded.
+            # The 'paths' or 'path_hash_prefixes' fields should not be missing,
+            # so we raise a format error here in case they are both missing.
+            raise exceptions.FormatError(
+                repr(self.name) + " "
+                'has neither a "paths" nor "path_hash_prefixes".  At least'
+                " one of these attributes must be present."
+            )
+
+        return None
+
+
+def _get_filepath_hash(target_filepath, hash_function="sha256"):
+    """
+    Calculate the hash of the filepath to determine which bin to find the
+    target.
+    """
+    # The client currently assumes the repository (i.e., repository
+    # tool) uses 'hash_function' to generate hashes and UTF-8.
+    digest_object = sslib_hash.digest(hash_function)
+    encoded_target_filepath = target_filepath.encode("utf-8")
+    digest_object.update(encoded_target_filepath)
+    target_filepath_hash = digest_object.hexdigest()
+
+    return target_filepath_hash
 
 
 class Delegations:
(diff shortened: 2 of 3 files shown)


sechkova:
Opened as a draft not only because it is dependent on #1408 but also it is not well tested yet.

jku:
I've had a first look and the ideas look fine. Some smaller comments:
* some issues in commit splitting (like visit_child_role() being annotated as returning bool and then returning something else)
* Likewise terminating child roles are added to child_roles_to_visit twice, then duplication is removed in a "simplification" commit
* IMO we should avoid doing e.g. `pathpattern.lstrip(os.sep)` -- these are the sort of workarounds that make life hard at some point... either the path pattern is valid or it isn't valid and we should just refuse to load the metadata (we don't have to support every weird case spec defines as valid). That said, leaving it as is and filing an issue is fine as well
* more type annotations would be nice: Dict without contained types is not great, `role_metadata` in `_preorder_depth_first_walk` should be manually annotated as Targets: without this the whole function is a black box as far as static typing is concerned.

I'm suspicious about how this operates when target is not found:
*  old code does not seem to raise if delegation limit is not reached? I agree the old code return value seems nonsense but is raising really the right thing?
* new code does not refer to delegation limit in the error message if it is reached

Also, is_in_trusted_paths() returns False if both paths and path_hash_prefixes are unset. Is this correct? I can see how that's the "safe" assumption but it also makes the whole idea of unset paths and unset path_hash_prefixes meaningless: the delegation has no reason to exist...

trishankatdatadog:
I really think it would be a lot simpler as a recursive function...

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,430 @@
...
+            # Skip any visited current role to prevent cycles.
+            if (role_name, parent_role) in visited_role_names:
+                logger.debug("Skipping visited current role %s", role_name)
+                continue
+
+            # The metadata for 'role_name' must be downloaded/updated before
+            # its targets, delegations, and child roles can be inspected.
+            self._load_targets(role_name, parent_role)
+
+            role_metadata = self._bundle[role_name].signed
```
jku: this implies we should make bundle (or trusted_set as it maybe now) return the loaded metadata to avoid a lookup immediately afterwards. Can you file an issue please?

sechkova:
> I really think it would be a lot simpler as a recursive function...

I agree, it does make the code look much simpler, let me test implementing those terminating delegations correctly with it :)

sechkova:
> Also, is_in_trusted_paths() returns False if both paths and path_hash_prefixes are unset. Is this correct? I can see how that's the "safe" assumption but it also makes the whole idea of unset paths and unset path_hash_prefixes meaningless: the delegation has no reason to exist...

~~Both paths and path_hash_prefixes being None should be impossible since there is a check in the constructor which raises an exception but if you think it will be better/safer to duplicate it again here, I can add it.~~
https://github.com/theupdateframework/tuf/blob/24fa1128116429dc18e185047fbffb009ca509b0/tuf/api/metadata.py#L914

Forget what I wrote, the check is for both being set ... 
Which leads me to the question, should the input validation include the check whether both are _not_ set?

sechkova:
Summarizing the changes made to this PR:
- rebased on  develop
- improved the commits 
- added more type annotations
- changed the function to return None when a targetfile is not found (instead of raising an exception)
- added one possible recursive implementation of the same function   e23e64b

Still there are issues pending and I am keeping this a draft to have a chance of reviewing it again myself.
- File an issue so that `self._load_targets(role_name, parent_role)` -> `TrustedMetadataSet` returns the loaded metadata to avoid lookup
- Avoid doing e.g. `pathpattern.lstrip(os.sep) `or file an issue about it
- Revise return values and maybe throw exceptions in corner cases
- The correct behaviour when both `paths` and `path_hash_prefixes` are unset is still unclear.
- Using `fnmatch` does not do exactly what is promised in the documentation in terms of UNIX filename pattern matching

tuf/ngclient/updater.py
```diff
@@ -316,76 +319,63 @@ def _load_targets(self, role: str, parent_role: str) -> None:
             self._persist_metadata(role, data)
 
     def _preorder_depth_first_walk(
-        self, target_filepath: str
-    ) -> Union[Dict[str, Any], None]:
+        self,
+        target_filepath: str,
+        visited_role_names: Set[str],
+        current_role_pair: List[Tuple[str, ...]],
```
MVrachev: What does `...` mean in `List[Tuple[str, ...]]`?

tuf/ngclient/updater.py
```diff
@@ -316,76 +319,63 @@ def _load_targets(self, role: str, parent_role: str) -> None:
...
-                self.config.max_delegations,
-            )
-
-        # If this point is reached then target is not found, return None
-        return None
+        if number_of_delegations <= 0:
+            return targetinfo, terminated
+
+        # Pop the role name from the top of the stack.
+        role_name, parent_role = current_role_pair
```
MVrachev: From the type annotation of `current_role_pair` I see it's a `List[Tuple[str, ...]`.
Don't you need to get an exact element from `current_role_pair` which will contain `role_name` and `parent_role` pair?

MVrachev (changes requested):
> Forget what I wrote, the check is for both being set ...
> Which leads me to the question, should the input validation include the check whether both are _not_ set?

I think we should check for that.
I created an issue where we can discuss that https://github.com/theupdateframework/tuf/issues/1497.

I like that it's much simpler and easier to read now and agree that the recursive version simplifies the code.

I think that before making this pr non-draft it's good to have tests for the ngclient updater.
I am a little worried about merging this before confirming we are not missing use-cases with this simplification.

tuf/ngclient/updater.py
jku: annotations for both visited_role_names and current_role_pair are wrong

jku:
> * added one possible recursive implementation of the same function   [e23e64b](https://github.com/theupdateframework/tuf/commit/e23e64b67e85e388e967eb45e4b72acd71ad357b)

I have some reservations against recursion: when you use it, you then have to think about the call stack depth limitations and how that affects your code... In our case maximum number of delegations now depends on call stack depth. 

Many hundreds of delegations is **not**  a use case we should explicitly design for, I accept that. Still, I don't like setting an artificial limit when we don't event know what that limit is: stack depth is implementation specific, cpython uses 1000,  but we also don't know how much of that we can use. The application developer might love recursion and may have used half the stack already, and maybe the network stack is special as well and needs a few hundred stack levels...

Basically I think _expected call stack depth_  is larger than maximum reasonable number of delegations... but it's only maybe 5-10 times as large and that sounds uncomfortably close to me.

trishankatdatadog:
> I have some reservations against recursion: when you use it, you then have to think about the call stack depth limitations and how that affects your code... In our case maximum number of delegations now depends on call stack depth.

As you said, I don't think any reasonable use case has that many delegations to worry about. Furthermore, it makes for more readable code, which is more important, I think.

sechkova:
Rebased on develop + some fixes of stupid errors in last commit.

tuf/ngclient/updater.py
sechkova: Done #1507

sechkova:
Thanks for the reviews and suggestions. I think it is time for this PR to stop being a draft and get a full review. I've intentionally kept the diffs in multiple smaller commits hoping that it may ease the review but you tell me how valuable it is for the history. I do mean to squash whatever feels redundant.

Some issues that were raised meanwhile:

- Recursive vs stack implementation:
    - I reverted the recursive implementation commit (intentionally reverted and not deleted to keep the history). I agree with @trishankatdatadog that the code looks visually better but at the same time function parameters increase and when we start propagating terminating flags up the stack, it doesn't feel any simpler. @jku gave reasonable arguments against recursion too. 

- I opened follow up issues for the items from the previous [list](https://github.com/theupdateframework/tuf/pull/1463#issuecomment-877212836): 
    - #1507
    - #1506
    - Revise return values and maybe throw exceptions in corner cases: the function now returns a targetinfo dictionary or None if a match is not found.
    - #1497 and 2159b88
    - Using fnmatch does not do exactly what is promised in the documentation in terms of UNIX filename pattern matching #1505

- Not an issue but `DelegatedRole.is_in_trusted_paths` feels a bit clumsy for a method name. Suggestions are welcome.
- Update if needed when https://github.com/theupdateframework/specification/issues/177 is resolved.

MVrachev (changes requested):
All of the changes seem reasonable and the history indeed is easier to follow when there are multiple smaller commits. Thanks for that @sechkova!

The only missing part in this pr is validation through more tests.
The `_preorder_depth_first_walk` is not an easy to understand function and we should
make sure we aren't introducing bugs.

We have a couple of tests using `get_one_valid_targetinfo()` in `test_updater_ng.py`, but
there are a couple of code paths not tested yet (according to coverage all):
- delegation role with `terminating` option turn on
- test when a target is not found
- reaching the max limit of visiting delegations, but there are more delegations left

Additionally, we never test `DelegatedRole.is_in_trusted_paths()`  when `path_hash_prefixes` is set and thus never actually test calculating the hashes.

sechkova:

sechkova:
> All of the changes seem reasonable and the history indeed is easier to follow when there are multiple smaller commits. Thanks for that @sechkova!
> 
> The only missing part in this pr is validation through more tests.

No two ways about that but I don't know whether this PR is the right way given that this is a common issue for the whole Updater: https://github.com/theupdateframework/tuf/issues/1499

tuf/ngclient/updater.py
```diff
@@ -368,210 +372,80 @@ def _load_targets(self, role: str, parent_role: str) -> None:
...
                 logger.debug("Skipping visited current role %s", role_name)
                 continue
 
             # The metadata for 'role_name' must be downloaded/updated before
             # its targets, delegations, and child roles can be inspected.
+            self._load_targets(role_name, parent_role)
 
-            role_metadata = self._trusted_set[role_name].signed
+            role_metadata: Targets = self._trusted_set[role_name].signed
             target = role_metadata.targets.get(target_filepath)
```
MVrachev, line 406:
```suggestion
            target: Union[TargetFile, None] = role_metadata.targets.get(target_filepath)
```
According to the metadata API if `role_metadata` is an instance of `Targets`, then 
`role_metadata.targets.get(target_filepath)` should be a `TargetFile` if it exists.
I think it make sense to add annotation here.

tuf/ngclient/updater.py
```diff
@@ -143,7 +142,9 @@ def refresh(self) -> None:
         self._load_snapshot()
         self._load_targets("targets", "root")
 
-    def get_one_valid_targetinfo(self, target_path: str) -> Dict:
+    def get_one_valid_targetinfo(
+        self, target_path: str
+    ) -> Union[Dict[str, Any], None]:
```
```suggestion
    ) -> Optional[Dict[str, Any]]:
```

tuf/ngclient/updater.py
```diff
@@ -368,210 +372,80 @@ def _load_targets(self, role: str, parent_role: str) -> None:
             self._trusted_set.update_delegated_targets(data, role, parent_role)
             self._persist_metadata(role, data)
 
-    def _preorder_depth_first_walk(self, target_filepath) -> Dict:
+    def _preorder_depth_first_walk(
+        self, target_filepath: str
+    ) -> Union[Dict[str, Any], None]:
```
```suggestion
    ) -> Optional[Dict[str, Any]]:
```

these mean union is then likely unused import

tuf/ngclient/updater.py
jku: (as reply to Martins comment:) since role_metadata is now a `Targets`, this does not need an annotation for target: static type check already figures out that target is a `TargetFile | None`

tuf/api/metadata.py
```diff
@@ -1031,6 +1040,36 @@ def to_dict(self) -> Dict[str, Any]:
...
 
+    def is_in_trusted_paths(self, target_filepath: str) -> bool:
+        """Determines whether the given 'target_filepath'
+        is in one of the trusted paths of DelegatedRole"""
+
+        if self.path_hash_prefixes is not None:
+            # Calculate the hash of the filepath to determine which bin
+            # to find the target. The client currently assumes the repository
+            # uses sslib's default algorithm to generate hashes.
+            digest_object = sslib_hash.digest()
```
jku: I think you can  remove the comment about assumptions and specify sha256 as algorithm here (as per spec) , no?

tuf/api/metadata.py
```diff
@@ -994,6 +996,13 @@ def __init__(
                 "Only one of the attributes 'paths' and"
                 "'path_hash_prefixes' can be set!"
             )
+
+        if paths is None and path_hash_prefixes is None:
+            raise ValueError(
+                "At least one of the attributes 'paths' and"
+                "'path_hash_prefixes' must be set!"
+            )
```
jku: I'd really appreciate if we could squeeze the exceptions and comments to the minimum number of lines that are required to convey the message... in the long term this has a major effect on readability. The content in this one seems a bit misleading as well so maybe: 

```suggestion
            raise ValueError("One of paths or path_hash_prefixes must be set")
```

tuf/ngclient/updater.py
```diff
@@ -169,6 +170,9 @@ def get_one_valid_targetinfo(self, target_path: str) -> Dict:
             OSError: New metadata could not be written to disk
             RepositoryError: Metadata failed to verify in some way
             TODO: download-related errors
+
+        Returns:
+            A targetinfo dictionary or None
```
jku, line 175: I think this is fine as long as we aim to get rid of the whole dictionary soon... the dict is just bad API. 

The goal for this method would be to just return Optional[TargetFile] but I guess we can't do that yet since 
* updated_targets() for some reason takes multiple TargetFiles and needs targetpaths for each -- this will not look good if caller needs to provide the targetpaths ... but some API redesign might help
* TargetFile does not have targetpath in it (yet at least)

tests/test_metadata_serialization.py
```diff
@@ -242,10 +242,9 @@ def test_snapshot_serialization(self, test_case_data: str):
         "no path attribute":
             '{"keyids": ["keyid"], "name": "a", "terminating": false, \
             "path_hash_prefixes": ["h1", "h2"], "threshold": 99}',
-        "no hash or path prefix":
-            '{"keyids": ["keyid"], "name": "a", "terminating": true, "threshold": 3}',
```
jku, line 246: shouldn't there now be a test that checks that deserialization fails with this input?

jku (approved):
I think this is a major improvement, much easier to follow now. `_preorder_depth_first_walk()` could maybe still be simplified a bit but this is so much better that I say let's get it in and do smaller tweaks afterwards (so the testing work can start).

A good set of test cases would be nice but I think we can leave that to #1499 -- the testing needs some serious thinking as well to prevent a mess of testing spagetti so makes sense to do in its own PR

`DelegatedRole.is_in_trusted_paths()` seems like a good idea. I tried to think of a better name but ... can't come up with anything, I guess this is fine.

the paths/path_hash_prefixes resolution seems reasonable.

I would remove both the recursive implementation commit and the revert if it's not a direction that was chosen.

I Also left some comments in code but overall this LGTM: I'm approving, but please have a look at whether you want to act on the comments.

sechkova:
> I Also left some comments in code but overall this LGTM: I'm approving, but please have a look at whether you want to act on the comments.

Thanks for the comments, let me try to address them before merging.

tuf/api/metadata.py
sechkova: Right ... it does say exactly sha256

tuf/api/metadata.py
sechkova: I tried to shorten the exception messages and also make them look consistently. It may have slightly changed the meaning though: 34f59c9

tests/test_metadata_serialization.py
sechkova: Added one with invalid DelegatedRoles: 34f59c9

tuf/ngclient/updater.py
sechkova: Hopefully we can change that with the progress of #1514

sechkova:
I think I addressed everything and removed the commits about the recursive implementation. Should be ready now.

tuf/api/metadata.py
```diff
@@ -1033,48 +1033,29 @@ def to_dict(self) -> Dict[str, Any]:
             res_dict["path_hash_prefixes"] = self.path_hash_prefixes
         return res_dict
 
-    def visit_child_role(self, target_filepath: str) -> str:
-        """Determines whether the given 'target_filepath' is an
-        allowed path of DelegatedRole"""
+    def is_in_trusted_paths(self, target_filepath: str) -> bool:
```
joshuagl: I'm not sure about use of the term "trusted" here. How about `is_described_path` (or `is_in_described_paths`)? 

At this point we are trying to determine whether the metadata's `paths` or `path_hash_prefixes` pattern might include the file at target_filepath. Of [`paths`](https://theupdateframework.github.io/specification/latest/#delegation-role-paths) and [`path_hash_prefixes`](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes) the specification states:

The specification has
> **"path_hash_prefixes"**
A list of **HEX_DIGESTs** used to succinctly describe a set of target paths.

and

> **"paths"**
A list of strings, where each string is a **PATHPATTERN** describing a path that the delegated role is trusted to provide.

*describe* and *describing* – what do you think?

joshuagl (commented):
Lovely set of clean ups, nice work @sechkova.

I'd prefer a different name for `is_in_trusted_paths()`, added a comment in-line. Let me know what you think.

tuf/api/metadata.py
jku: better. How about using "delegated" again: `DelegatedRole.is_path_delegated()` or `DelegatedRole.is_delegated_path()`?

joshuagl: I vote `is_delegated_path()`

sechkova: And `is_delegated_path()` it is!

PR #1466, opened 2021-06-23
Metadata API: Simplify unrecognized fields testing

### LINKED ISSUE #1464: Metadata API: Test unrecognized fields
Add tests for unrecognized fields support.

Read: https://github.com/theupdateframework/tuf/blob/develop/docs/adr/0008-accept-unrecognised-fields.md

Related pr: #1382

MVrachev:
Fixes #1464 

We have merged ADR 8 allowing for unrecognized fields everywhere.
After a discussion with Jussi, we realized that there are a couple of
places where we don't want to allow `unrecognized fields` because they
are sensitive or there are limited acceptable values for them.
The places where we don't want to allow unrecognized fields are
`keys`, `roles`, `meta`, `hashes` or `targets`.
If we allow unrecognized fields in them we won't follow the spec or
even open the door for possible security vulnerability.

Second, the test we added for `unrecognized fields`
added tests for that which are too specific and not scalable.
Now, I use table testing which we have used initially in #1416
to test unrecognized fields support in a cleaner and much more readable
way.

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -53,6 +53,10 @@ class TestSerialization(unittest.TestCase):
     valid_keys: DataSet = {
         "all": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", \
             "keyval": {"public": "foo"}}',
+        "unrecognized field": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", \
+            "keyval": {"public": "foo"}, "foo": "bar"}',
+        "unrecognized field in keyval": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", \
+            "keyval": {"public": "foo", "foo": "bar"}}',
     }
 
     @run_sub_tests_with_dataset(valid_keys)
@@ -63,7 +67,8 @@ class TestSerialization(unittest.TestCase):
 
 
     valid_roles: DataSet = {
-        "all": '{"keyids": ["keyid"], "threshold": 3}'
+        "all": '{"keyids": ["keyid"], "threshold": 3}',
+        "unrecognized field": '{"keyids": ["keyid"], "threshold": 3, "foo": "bar"}',
     }
 
     @run_sub_tests_with_dataset(valid_roles)
@@ -84,6 +89,11 @@ class TestSerialization(unittest.TestCase):
             "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"} }}, \
             "roles": { "targets": {"keyids": ["keyid"], "threshold": 3} } \
             }',
+        "unrecognized field": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
+            "expires": "2030-01-01T00:00:00Z", "consistent_snapshot": false, \
+            "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": { "targets": {"keyids": ["keyid"], "threshold": 3}}, \
+            "foo": "bar"}',
     }
 
     @run_sub_tests_with_dataset(valid_roots)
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
Directly relevant to https://github.com/theupdateframework/specification/pull/169

@asraa @mnm678 wdyt?

mnm678:
I think there's a difference between unrecognized and unauthorized fields. Unrecognized fields in the signature wrapper (which are also unauthorized) could be dangerous as they could add `alg=none` style vulnerabilities. However, I think there is less danger in unrecognized fields within signed metadata. Yes, and attacker could add something malicious to one of these fields, but if they can do so in the signed metadata, they could also just replace keys, etc in the existing fields of the metadata.

Therefore, I prefer the simplicity of allowing these unrecognized fields anywhere within the signed metadata to allow for extensions to the specification to be backwards compatible.

MVrachev:
> Therefore, I prefer the simplicity of allowing these unrecognized fields anywhere within the signed metadata to allow for extensions to the specification to be backwards compatible.

We cannot allow unrecognized fields everywhere if we follow the spec.
For example, `ROLE` in root has a limited number of allowed values: https://theupdateframework.github.io/specification/latest/index.html#root-role
The same can be said for `meta` in `timestamp`. 
For `hashes` there are concrete requirements
```
A dictionary that specifies one or more hashes of the target file at TARGETPATH, 
with a string describing the cryptographic hash function as key and HASH as defined for METAFILES.
```
and if a in `hashes` is a dictionary following the format described above we don't consider it as an `unrecognized field`.
Same argument for `keys`.

tests/test_metadata_serialization.py
```diff
@@ -51,8 +51,11 @@ def wrapper(test_cls: "TestSerialization"):
 class TestSerialization(unittest.TestCase):
 
     valid_keys: DataSet = {
+        # Do we want to add unrecognized fields in the Key class?
```
jku:
```suggestion
```

this may be a valid question but maybe not useful in test code?

jku (commented):
Maybe the ADR change could be another PR? I'm not as invested in tweaking the ADR text but would like to see the tests land...

Anyway for ADR text:
* I wouldn't explain the reasons further than "don't store unrecognized fields when it is not allowed by specification: keys, roles, meta, hashes and targets are actual dictionaries (vs json objects that most structures in the format are) where 'unrecognised field' is not meaningful concept". This seems to cover it?
* there's a lot of added text that is IMO not necessary and that would require review: security claims (like "possibility to inject keys, compromise trust") do not seem useful when they refer to metadata that is signed and thus by definition able to add keys and compromise trust.


For the tests:
* I think this is a good way to do this (better than writing code anyway), I have no major comments
* Question: Have you reviewed the file format and made sure we are testing every possible dictionary?
* nitpick: Always adding a trailing comma after the last item in the test dataset would make the next PR easier to read

tests/test_metadata_serialization.py
MVrachev: I will remove it, but wanted to ask raise that question.
Do we want to allow unrecognized fields in the Key class?

joshuagl: There's a desire to be able to store unrecognised fields in a [keyval](https://theupdateframework.github.io/specification/latest/#keyval), see https://github.com/theupdateframework/specification/issues/163 and https://github.com/theupdateframework/go-tuf/pull/133

MVrachev:
I updated this pr to include only the changes related to the tests.

After the @joshuagl comment, I included a test that we do support unrecognized fields in `keyval`.

> * Question: Have you reviewed the file format and made sure we are testing every possible dictionary?

Yes, I reviewed the file formats yet again and it seems okay. The only places where we don't allow unrecognized fields is where I described above.

> * nitpick: Always adding a trailing comma after the last item in the test dataset would make the next PR easier to read

Fixed that.

jku (approved):
the commented question is still in the test data?

anyway, looks fine, thanks

MVrachev:
> the commented question is still in the test data?
> 
> anyway, looks fine, thanks

Removed the question, but still - it's okay to have unrecognized fields there, right?

jku:
> Removed the question, but still - it's okay to have unrecognized fields there, right?

I don't see why not -- keyval is even designed for that, AFAICT (as the contents are not defined for keys in general).

PR #1467, opened 2021-06-25
ngclient: use MetaFile/TargetFile verification

### LINKED ISSUE #1410: ngclient: use MetaFile/TargetFile verification API when it exists
Related to #1377 and #1361: Once Metadata API offers the verification functionality, we should use it in ngclient and verify file lengths and hashes for both metadata and targets.

This would mean removal of helper functions like _check_file_length() _check_hashes_obj(), and removal of hash checking code in metadatabundle

sechkova:
Fixes #1410

Use the newly added  MetaFile/TargetFile.verify_length_and_hashes in the ngclient workflow.

requirements**:

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -205,17 +205,21 @@ class Updater:
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        full_url = parse.urljoin(target_base_url, targetinfo["filepath"])
+        target_filepath = targetinfo["filepath"]
+        target_fileinfo: "TargetFile" = targetinfo["fileinfo"]
+        full_url = parse.urljoin(target_base_url, target_filepath)
 
         with download.download_file(
-            full_url, targetinfo["fileinfo"].length, self._fetcher
+            full_url, target_fileinfo.length, self._fetcher
         ) as target_file:
-            _check_file_length(target_file, targetinfo["fileinfo"].length)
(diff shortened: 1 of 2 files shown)


sechkova:
Only the last two commits are new in this PR: 39ebe92  f8ecac8

sechkova:
Rebased on `develop`

tuf/ngclient/updater.py
```diff
@@ -205,17 +205,21 @@ def download_target(
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        full_url = parse.urljoin(target_base_url, targetinfo["filepath"])
+        target_filepath = targetinfo["filepath"]
+        target_fileinfo = targetinfo["fileinfo"]
```
```suggestion
        target_fileinfo:TargetFile = targetinfo["fileinfo"]
```
the real fix would be fixing the targetinfo argument type but this helps typing in the mean time

jku (commented):
Looks good I think. Handling the exception just to raise again is a bit annoying but ... also probably correct:
* in these cases we do want to raise a RepositoryError -- client has no way to "handle" this error
* but other cases might be different (like if client itself was testing a cached file)

just on one question: what about the targetfile check in `updated_targets()`? it's a bit different case but should be able to use the same methods

sechkova:
> just on one question: what about the targetfile check in `updated_targets()`? it's a bit different case but should be able to use the same methods

:roll_eyes: 
1b28604 <- I used them now
aeccb05 is the same commit with added `TargetFile` type

jku (approved):
:+1:

PR #1470, opened 2021-06-29
ngclient constants and configuration

### LINKED ISSUE #1402: ngclient: client-level constants

sechkova:
Fixes #1402

Adds a new public config module `ngclient/config.py`.
- containing a dataclass `UpdaterConfig` with all client settings
- Initializes updater with the default settings if no other config is provided.

requirements**:

### DIFF
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -7,6 +7,7 @@
 
 import logging
 import time
+from typing import Optional
 from urllib import parse
 
 # Imports
@@ -14,7 +15,7 @@ import requests
 import urllib3.exceptions
 
 import tuf
-from tuf import exceptions, settings
+from tuf import exceptions
 from tuf.ngclient.fetcher import FetcherInterface
 
 # Globals
@@ -47,6 +48,11 @@ class RequestsFetcher(FetcherInterface):
         # Some cookies may not be HTTP-safe.
         self._sessions = {}
 
+        # Default settings
+        self.socket_timeout: int = 4  # seconds
+        self.chunk_size: int = 400000  # bytes
+        self.sleep_before_round: Optional[int] = None
+
     def fetch(self, url, required_length):
         """Fetches the contents of HTTP/HTTPS url from a remote server.
 
@@ -75,9 +81,7 @@ class RequestsFetcher(FetcherInterface):
         # requests as:
         #  - connect timeout (max delay before first byte is received)
         #  - read (gap) timeout (max delay between bytes received)
-        response = session.get(
-            url, stream=True, timeout=settings.SOCKET_TIMEOUT
-        )
+        response = session.get(url, stream=True, timeout=self.socket_timeout)
         # Check response status.
         try:
             response.raise_for_status()
@@ -99,11 +103,11 @@ class RequestsFetcher(FetcherInterface):
                     # large file in one shot. Before beginning the round, sleep
                     # (if set) for a short amount of time so that the CPU is not
                     # hogged in the while loop.
-                    if settings.SLEEP_BEFORE_ROUND:
-                        time.sleep(settings.SLEEP_BEFORE_ROUND)
+                    if self.sleep_before_round:
+                        time.sleep(self.sleep_before_round)
 
                     read_amount = min(
-                        settings.CHUNK_SIZE,
+                        self.chunk_size,
                         required_length - bytes_received,
                     )
 
--- a/tuf/ngclient/config.py
+++ b/tuf/ngclient/config.py
@@ -1,0 +1,17 @@
+# Copyright 2021, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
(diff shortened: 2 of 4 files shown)


tuf/ngclient/config.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+@dataclass
+class UpdaterConfig:
+    MAX_ROOT_ROTATIONS: int = 32
+    MAX_DELEGATIONS: int = 32
+    DEFAULT_ROOT_MAX_LENGTH: int = 512000  # bytes
+    DEFAULT_TIMESTAMP_MAX_LENGTH: int = 16384  # bytes
+    DEFAULT_SNAPSHOT_MAX_LENGTH: int = 2000000  # bytes
+    DEFAULT_TARGETS_MAX_LENGTH: int = 5000000  # bytes
```
jku: since these are now not intended to be constants or scary we can lowercase them... that removes a pylint disable as well

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,578 @@
...
+            self._target_base_url = _ensure_trailing_slash(target_base_url)
+
+        # Read trusted local root metadata
+        data = self._load_local_metadata("root")
+        self._trusted_set = trusted_metadata_set.TrustedMetadataSet(data)
+
+        if config is None:
+            self.config = UpdaterConfig()
+        else:
+            self.config = config
```
```suggestion
        self.config = config or UpdaterConfig()
```
should work in this case?

jku (changes requested):
The RequestsFetcher changes do not belong here IMO: RequestsFetcher is only one fetcher implementation:
* If we think these values must be supported by every implementation we should make that part of the interface, not sneak them in like this
* If these are just RequestsFetcher config, then they shouldn't be in UpdaterConfig

I would just make the requestsfetcher configuration instance attributes of RequestsFetcher: The user can always instantiate one and then modify the attributes before giving the fetcher to updater.

tuf/ngclient/config.py
sechkova: yes, I did and and I think it even looks better f5b61ec276aac286d2e6b2f91955c4315124a08b Thanks for the proposal!

sechkova:
Rebased on `develop`. 
Also reworked all commits after the review suggestions so another one will be needed.

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -47,6 +48,11 @@ def __init__(self):
         # Some cookies may not be HTTP-safe.
         self._sessions = {}
 
+        # Default settings
+        self.sleep_before_round: int = 4  # seconds
```
jku: er, I think you mixed up sleep_before_round and socket_timeout: this is going to make life painful

tuf/ngclient/config.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+@dataclass
+class UpdaterConfig:
+    max_root_rotations: int = 32
+    max_delegations: int = 32
+    default_root_max_length: int = 512000  # bytes
+    default_timestamp_max_length: int = 16384  # bytes
+    default_snapshot_max_length: int = 2000000  # bytes
+    default_targets_max_length: int = 5000000  # bytes
```
jku: None one of the variables should be called "default" i think -- looks a bit odd in the code since they might be set by the user and they're still called "default"?

jku (commented):
Looks good. I'm not super into adding optional things into the constructors... but if others like the config object pattern I'm OK with it.

We should get rid of sleep_before_round: if we're "hogging CPU" as the comment says, let's fix that instead. I don't think we are though: I'm sure requests does not require us to poll constantly. Can you file an issue on that?

tuf/ngclient/_internal/requests_fetcher.py
sechkova: oh 🤦

tuf/ngclient/config.py
sechkova: Maybe default makes sense for snapshot and targets where this is the value used if the one in metadata is not se but I renamed all anyway 520b344.

jku (approved):

Issue #1471, opened 2021-06-29
Track licencing using SDPX headers


rzr:
Current double licensing can be unclear, 
latest release was rejected by debian project:

https://github.com/theupdateframework/tuf/issues/263#issuecomment-862352717

To address this problem I decided to open a new issue.

The first problem is to identify imported tor's thandy's sources, 
so double checking  "Copyright:"   "SPDX-License-Identifier:"  headers in files 
would not hurt.

The oldest  copyright header I current tree is about year 2012 while thandy is from ~2008,
it sounds like something hasn't been preserved.

Is thandy @abandonware ? if not can tuf be rebased on upstream ?

Regards

trishankatdatadog:
Thanks, Philippe!

As far as I'm aware, there's very little code, if any, left over from Thandy...

mnm678:
I opened #1472 to remove the outdated Thandy license info. I'd like to wait for @JustinCappos and others with more historical knowledge of the project to approve, but I believe that will help solve this issue.

trishankatdatadog:
> I opened #1472 to remove the outdated Thandy license info. I'd like to wait for @JustinCappos and others with more historical knowledge of the project to approve, but I believe that will help solve this issue.

Yeap, Justin is best placed to resolve this. I know, for example, I rewrote the download code, but the entire RVCS history has unfortunately been lost.

mnm678:
I just merged #1472. Is there anything else needed to address this issue @rzr?

rzr:
looks good , do we want next released  to be also MIT licensed ? 
since the file is still present:

https://github.com/theupdateframework/tuf/blob/develop/LICENSE-MIT

I'll try to submit next release to debian and see how it goes.

BTW MIT license is not encouraged since it brings some kind of ambiguity.

For dual licensing may I also suggest to use ISC instead.
which is in same vein of MIT:

https://spdx.org/licenses/ISC.html



cc: @joshuagl

jku:
So #1472 removed the mention of thandy files in TUF (and my review agrees with that assessment: there is no thandy code in TUF itself). In addition #1491 has Tor project relicensing thandy code as "MIT OR Apache2.0": this is relevant because securesystemslib _does_ have some thandy code in it (https://github.com/secure-systems-lab/securesystemslib/issues/365).

All code files (apart from empty `__init__.py` files) seem to have the correct SPDX header (`SPDX-License-Identifier: MIT OR Apache-2.0`) so things should now be in order: closing this

PR #1472, opened 2021-06-29
Remove license from Thandy files that are no longer present

mnm678:
Solves #1471 

Based on my current understanding, the Thandy license information is out of date.


trishankatdatadog:
Should get a 👍🏽 from @JustinCappos

JustinCappos (approved):
I agree this is correct.

PR #1473, opened 2021-06-30
Fix CLI repo.py

Moutix:
Fixes/Improvment on the repo.py script

## Read threshold parameters while trusting keys

We can now set a threshold of signature when trusting keys like this:

```bash
repo.py --init --bare

## Generate 3 keys for root and trust them
repo.py --key --filename root_key
repo.py --key --filename root_key2
repo.py --key --filename root_key3
repo.py --trust --threshold 2 --pubkeys tufkeystore/root_key.pub tufkeystore/root_key2.pub tufkeystore/root_key3.pub --role root
```

## Read pw parameters when signing

--sign parameters was not reading the pw parameters and was asking the password. A different behavior than for the others command which was misleading.

## Generate the delegate target file when using --delegate

When generating delegation, script was failing because the delegate file was not created (role.json). It is fix with this PR

```bash
+ repo.py --clean
+ repo.py --init
+ repo.py --key --filename my_role_key
+ repo.py --role targets --delegate 'my_role/*' --delegatee my_role --pubkeys tufkeystore/my_role_key.pub --threshold 1 --terminating
+ repo.py --sign tufkeystore/targets_key --role targets
  File "/Users/selim.menouar/.virtualenvs/tuf/bin/repo.py", line 1147, in <module>
    process_command_line_arguments(arguments)
  File "/Users/selim.menouar/.virtualenvs/tuf/bin/repo.py", line 262, in process_command_line_arguments
    sign_role(parsed_arguments)
  File "/Users/selim.menouar/.virtualenvs/tuf/bin/repo.py", line 592, in sign_role
    repository = repo_tool.load_repository(
  File "/Users/selim.menouar/.virtualenvs/tuf/lib/python3.8/site-packages/tuf/repository_tool.py", line 3111, in load_repository
    metadata_path = delegated_roles_filenames[rolename]
KeyError: 'my_role'
```

You can validate it's working by running this script

```bash
#!/bin/bash

set -ex

# Clean the repository
repo.py --clean

## Initialize a new repo
repo.py --init

# my_role
repo.py --key --filename my_role_key
repo.py --role targets --delegate "my_role/*" --delegatee my_role --pubkeys tufkeystore/my_role_key.pub --threshold 1 --terminating
repo.py --sign tufkeystore/targets_key --role targets

# my_other_role
repo.py --key --filename my_other_key
repo.py --role my_role --delegate "my_other_role/*" --delegatee my_other_role --pubkeys tufkeystore/my_role_key.pub --threshold 1 --terminating --sign tufkeystore/my_role_key
repo.py --sign tufkeystore/my_role_key --role my_role
```

### DIFF
--- a/tuf/scripts/repo.py
+++ b/tuf/scripts/repo.py
@@ -94,7 +94,7 @@
     
   --trust:
     Indicate trusted keys located in </path/to/pubkey> directory of a role.
-    
+
   --sign: 
     Sign metadata of target role(s) with keys in specified directory.
     
@@ -117,10 +117,11 @@
     
   --terminating:
     Mark delegation to --delegatee role from delegator as a terminating one.
-  
+
   --threshold:
-    Specify signature threshold of --delegatee role as the value <X>.
-    
+    Set the threshold number of signatures needed to validate a metadata file.
+    Can be used with --delegate and --trust.
+
   --revoke:
     Revoke trust of target files from delegated role (--delegatee)
     
@@ -282,24 +283,36 @@ def delegate(parsed_arguments):
       os.path.join(parsed_arguments.path, REPO_DIR))
 
   if parsed_arguments.role == 'targets':
-    repository.targets.delegate(parsed_arguments.delegatee, public_keys,
-        parsed_arguments.delegate, parsed_arguments.threshold,
-        parsed_arguments.terminating, list_of_targets=None,
-        path_hash_prefixes=None)
-
-    targets_private = import_privatekey_from_file(
-        os.path.join(parsed_arguments.path, KEYSTORE_DIR, TARGETS_KEY_NAME),
-        parsed_arguments.targets_pw)
-
-    repository.targets.load_signing_key(targets_private)
-
-  # A delegated (non-top-level-Targets) role.
+    parent_targets = repository.targets
+    # Use key from --sign if present or use the default targets_key
+    parent_targets_keys = parsed_arguments.sign or [os.path.join(KEYSTORE_DIR, TARGETS_KEY_NAME)]
   else:
-    repository.targets(parsed_arguments.role).delegate(
-        parsed_arguments.delegatee, public_keys,
-        parsed_arguments.delegate, parsed_arguments.threshold,
-        parsed_arguments.terminating, list_of_targets=None,
-        path_hash_prefixes=None)
+    parent_targets = repository.targets(parsed_arguments.role)
+    parent_targets_keys = parsed_arguments.sign
+
+  parent_targets.delegate(parsed_arguments.delegatee, public_keys,
+    parsed_arguments.delegate, parsed_arguments.threshold or 1,
+    parsed_arguments.terminating, list_of_targets=None,
+    path_hash_prefixes=None
+  )
+
+  # Load the private key for the parent targets
+  for key in parent_targets_keys:
+    targets_private = import_privatekey_from_file(
+      os.path.join(parsed_arguments.path, key),
+      parsed_arguments.targets_pw
+    )
+    parent_targets.load_signing_key(targets_private)
+
+  # Load the private keys of the delegatee targets to allow the generation of the delegatee targets file
+  for key in parsed_arguments.pubkeys:
+    # Expect the public key to be named public.key and private key public
+    delegatee_private = import_privatekey_from_file(
+      os.path.join(parsed_arguments.path, key[:-4]),
+      parsed_arguments.pw,
+    )
+    repository.targets(parsed_arguments.delegatee).load_signing_key(delegatee_private)
+
 
   # Update the required top-level roles, Snapshot and Timestamp, to make a new
   # release.  Automatically making a new release can be disabled via
@@ -514,19 +527,12 @@ def add_verification_key(parsed_arguments):
 
     if parsed_arguments.role not in ('root', 'targets', 'snapshot', 'timestamp'):
       raise exceptions.Error('The given --role is not a top-level role.')
-
-    elif parsed_arguments.role == 'root':
-      repository.root.add_verification_key(imported_pubkey)
-
-    elif parsed_arguments.role == 'targets':
-      repository.targets.add_verification_key(imported_pubkey)
-
-    elif parsed_arguments.role == 'snapshot':
-      repository.snapshot.add_verification_key(imported_pubkey)
-
-    # The timestamp role..
     else:
-      repository.timestamp.add_verification_key(imported_pubkey)
+      role = getattr(repository, parsed_arguments.role)
+      role.add_verification_key(imported_pubkey)
+      # If we ask to update the threshold update it
+      if role.threshold:
+        role.threshold = parsed_arguments.threshold
 
   consistent_snapshot = roledb.get_roleinfo('root',
       repository._repository_name)['consistent_snapshot']
@@ -592,7 +598,7 @@ def sign_role(parsed_arguments):
 
   for keypath in parsed_arguments.sign:
 
-    role_privatekey = import_privatekey_from_file(keypath)
+    role_privatekey = import_privatekey_from_file(keypath, parsed_arguments.pw)
 
     if parsed_arguments.role == 'targets':
       repository.targets.load_signing_key(role_privatekey)
@@ -1095,9 +1101,10 @@ def parse_arguments():
   parser.add_argument('-t', '--terminating', action='store_true',
       help='Set the terminating flag to True.  Can be used with --delegate.')
 
-  parser.add_argument('--threshold', type=int, default=1, metavar='<int>',
+  # By default is set to 1 on new metadata but on update, it doesn't change the threshold
+  parser.add_argument('--threshold', type=int, metavar='<int>',
       help='Set the threshold number of signatures'
-      ' needed to validate a metadata file.  Can be used with --delegate.')
+      ' needed to validate a metadata file.  Can be used with --delegate and --trust.')
 
   parser.add_argument('--pubkeys', type=str, nargs='+',
       metavar='</path/to/pubkey_file>', help='Specify one or more public keys'


tuf/scripts/repo.py
```diff
@@ -94,7 +94,8 @@
     
   --trust:
     Indicate trusted keys located in </path/to/pubkey> directory of a role.
-    
+  --expire:
```
trishankatdatadog: The `--threshold` parameter is missing here, no?

tuf/scripts/repo.py
```diff
@@ -287,11 +289,22 @@ def delegate(parsed_arguments):
...
+    # Use key from --sign if present or use the default target_key
+    for key in parsed_arguments.sign or [os.path.join(KEYSTORE_DIR, TARGETS_KEY_NAME)]:
+      targets_private = import_privatekey_from_file(
+        os.path.join(parsed_arguments.path, key),
+        parsed_arguments.targets_pw
+      )
+      repository.targets.load_signing_key(targets_private)
 
-    repository.targets.load_signing_key(targets_private)
+    # Generate the delegate targets file
```
trishankatdatadog:
```suggestion
    # Generate the delegatee targets file
```

tuf/scripts/repo.py
```diff
@@ -287,11 +289,22 @@ def delegate(parsed_arguments):
...
+    for key in parsed_arguments.pubkeys:
+      delegate_private = import_privatekey_from_file(
```
trishankatdatadog:
```suggestion
      delegatee_private = import_privatekey_from_file(
```

tuf/scripts/repo.py
```diff
@@ -287,11 +289,22 @@ def delegate(parsed_arguments):
...
+        os.path.join(parsed_arguments.path, key[:-4]),
+        parsed_arguments.pw,
+      )
+
+      repository.targets(parsed_arguments.delegatee).load_signing_key(delegate_private)
```
trishankatdatadog:
```suggestion
      repository.targets(parsed_arguments.delegatee).load_signing_key(delegatee_private)
```

tuf/scripts/repo.py
```diff
@@ -509,24 +522,20 @@ def add_verification_key(parsed_arguments):
   repository = repo_tool.load_repository(
       os.path.join(parsed_arguments.path, REPO_DIR))
 
+  expire = None
```
trishankatdatadog: Does it never expire by default then?

tuf/scripts/repo.py
trishankatdatadog: Where are the password flags documented?

trishankatdatadog (changes requested):
Hard to reason about this PR because it's trying to do many things at once, even if each of them is small. Would you please answer the questions below?

tuf/scripts/repo.py
Moutix: if not set it will take the default value (as before this PR)

tuf/scripts/repo.py
Moutix: the --threshold parameters is defined here https://github.com/theupdateframework/tuf/blob/develop/tuf/scripts/repo.py#L121 but is actually only use for delegate right now

the --pw field is documented here https://github.com/theupdateframework/tuf/blob/develop/tuf/scripts/repo.py#L107

trishankatdatadog:
@joshuagl and friends, thoughts here pls? 🙂

tuf/scripts/repo.py
```diff
@@ -1097,12 +1095,16 @@ def parse_arguments():
 
   parser.add_argument('--threshold', type=int, default=1, metavar='<int>',
       help='Set the threshold number of signatures'
-      ' needed to validate a metadata file.  Can be used with --delegate.')
+      ' needed to validate a metadata file.  Can be used with --delegate and --trust.')
```
MVrachev, line 1107: While you have started updating the `threshold` description could you update the `threshold` description in the
module documentation as well?
I think it should be line 122 from your changes.
I saw you had updated it, but can you make them be the same?

It will be great if you can have at the module documentation about the rest of the options as well and compare it
with the actual documentation when you call `--help` from the cli.

tuf/scripts/repo.py
```diff
@@ -295,6 +295,15 @@ def delegate(parsed_arguments):
...
     repository.targets.load_signing_key(targets_private)
 
+    # Generate the delegate targets file
+    for key in parsed_arguments.pubkeys:
+      delegate_private = import_privatekey_from_file(
+        os.path.join(parsed_arguments.path, key[:-4]),
+        parsed_arguments.pw,
+      )
+
+      repository.targets(parsed_arguments.delegatee).load_signing_key(delegate_private)
```
MVrachev: From what I can see we are not generating a new file in this code block.
I know we are generating a new delegate targets file later on and I am not sure if this comment
helps or confuses where it's placed.

tuf/scripts/repo.py
MVrachev: @Moutix is this resolved?

MVrachev (commented):
Some minor suggestions as I am not a maintainer.

jku (commented):
Thanks, and sorry for taking my time (this file is new code to me too). Martin already mentioned the slightly misleading comment so I just have questions/requests for commit messages:  
* Would like to see "Why?" answered in the PR message and/or the commit messages -- I assume the current code just doesn't work when you try to delegate but that would be good to say out loud (maybe with the command line included that did not work, but works after your fix?)
* I assume this fixes things specifically for top-level targets but not for delegated targets -- I can see how fixing that might get tricky the way the CLI password arguments are designed but would be good to see this mentioned in a commit or PR message?

tuf/scripts/repo.py
Moutix: Thks indeed! I've fixed it for the threshold line

tuf/scripts/repo.py
Moutix: Yeah true, indeed we load the key so that the delegate targets role can be generated, I'm updating it

Moutix:
> Thanks, and sorry for taking my time (this file is new code to me too). Martin already mentioned the slightly misleading comment so I just have questions/requests for commit messages:
> 
>     * Would like to see "Why?" answered in the PR message and/or the commit messages -- I assume the current code just doesn't work when you try to delegate but that would be good to say out loud (maybe with the command line included that did not work, but works after your fix?)
> 
>     * I assume this fixes things specifically for top-level targets but not for delegated targets -- I can see how fixing that might get tricky the way the CLI password arguments are designed but would be good to see this mentioned in a commit or PR message?

Thks! I've updated the PR description with more details.

For the 2nd point, I push something. It will require to set the password of the parent targets in the --targets_pw if needed

tuf/scripts/repo.py
```diff
@@ -282,24 +285,35 @@ def delegate(parsed_arguments):
...
+      os.path.join(parsed_arguments.path, key),
+      parsed_arguments.targets_pw
+    )
+    parent_targets.load_signing_key(targets_private)
 
-    repository.targets.load_signing_key(targets_private)
+  # Load the private keys of the delegatee targets to allow the generation of the delegatee targets file
+  for key in parsed_arguments.pubkeys:
+    delegatee_private = import_privatekey_from_file(
+      os.path.join(parsed_arguments.path, key[:-4]),
```
jku, line 311: This now loads private keys based on `--pubkeys` argument, making assumptions on filenames that are not made before this in the code. Hmm...

I have to admit I don't care too much about the maintainability of this file (that train has probably passed the station)... I'm just pointing out that that looks like a workaround.

jku (commented):
Question on the expiry setting: Does it actually do something useful? I can see that the code already has this concept of "key expiry" but ... the spec and file format do not. Does the key expiry get used in the produced repository somehow or is it just dead code?

tuf/scripts/repo.py
Moutix: Yes completely. Otherwise I would have had to update the interface

But yeah overall I feel like to have something clean a lot of things should probably be rewritten ^^'

Moutix: Actually an assumption is already done in the code that the key should be named base on the role (targets.pub/snapshot.pub) otherwise it doesn't work. Which actually make things complicated if you have multiple key and a threshold

Moutix:
> Question on the expiry setting: Does it actually do something useful? I can see that the code already has this concept of "key expiry" but ... the spec and file format do not. Does the key expiry get used in the produced repository somehow or is it just dead code?



> Question on the expiry setting: Does it actually do something useful? I can see that the code already has this concept of "key expiry" but ... the spec and file format do not. Does the key expiry get used in the produced repository somehow or is it just dead code?

It's a good point, I thought that it would be set on the field 'expires' of the metadata, but after checking it's actually not the case. The parameter is completely ignore from what I can read from the generated file.

I will remove the parameters to not add confusion

tuf/scripts/repo.py
```diff
@@ -514,19 +526,11 @@ def add_verification_key(parsed_arguments):
...
-    elif parsed_arguments.role == 'snapshot':
-      repository.snapshot.add_verification_key(imported_pubkey)
-
-    # The timestamp role..
     else:
-      repository.timestamp.add_verification_key(imported_pubkey)
+      role = getattr(repository, parsed_arguments.role)
+      role.add_verification_key(imported_pubkey)
+      if role.threshold != parsed_arguments.threshold:
+        role.threshold = parsed_arguments.threshold
```
jku: I believe if `--threshold` is not set, this resets role.threshold to 1.

tuf/scripts/repo.py
```diff
@@ -592,7 +590,7 @@ def sign_role(parsed_arguments):
 
   for keypath in parsed_arguments.sign:
 
-    role_privatekey = import_privatekey_from_file(keypath)
+    role_privatekey = import_privatekey_from_file(keypath, parsed_arguments.pw)
```
jku, line 601: Doesn't this trigger the weirdo default behaviour where if --pw is not used, the string "pw" is used as password?

I guess I again don't really object changing one strange behaviour to another :shrug:

jku (commented):
Sorry for drip-feeding you comments, but it is multiple changes in one (and in part of the code that I've been actively trying to avoid :) )

tuf/scripts/repo.py
Moutix: Yes exactly. It's the case everywhere so I think it's better to have this strange behavior everywhere otherwise it's confusing

tuf/scripts/repo.py
Moutix: Yep exactly, so by default it will always be a threshold of 1, but if people want to change they have to specify the threshold each time they trust keys

jku: That is extremely surprising and makes it easy to create a security disaster: If I do
`repo.py --trust --pubkeys key1.pub key2.pub key3.pub --role root`
I don't expect my root threshold to just be reset to 1 from whatever it was before, yet I think that's what happens now.

I think there's two choices:
* If `--threshold` is not set, threshold should not be be modified here. This might mean you'd have to change the default value for the argument and then change how `delegate()` and `add_verification_key()` handle the default
* make `--threshold` a required argument with `--trust`

Moutix: Good point, will update that

Moutix: So I kept the same behavior than before for delegate() default of 1 if not set.

For trust it will only be updated if the threshold key is set, otherwise it will keep the last value

tuf/scripts/repo.py
```diff
@@ -282,24 +283,35 @@ def delegate(parsed_arguments):
...
+    path_hash_prefixes=None
+  )
 
+  # Load the private key for the parent targets
+  for key in parent_targets_keys:
     targets_private = import_privatekey_from_file(
-        os.path.join(parsed_arguments.path, KEYSTORE_DIR, TARGETS_KEY_NAME),
-        parsed_arguments.targets_pw)
+      os.path.join(parsed_arguments.path, key),
+      parsed_arguments.targets_pw
```
trishankatdatadog, line 303: Are all the keys expected to share the same password? That's a strange assumption...

tuf/scripts/repo.py
```diff
@@ -282,24 +283,35 @@ def delegate(parsed_arguments):
...
+    )
+    parent_targets.load_signing_key(targets_private)
 
-    repository.targets.load_signing_key(targets_private)
+  # Load the private keys of the delegatee targets to allow the generation of the delegatee targets file
+  for key in parsed_arguments.pubkeys:
+    delegatee_private = import_privatekey_from_file(
+      os.path.join(parsed_arguments.path, key[:-4]),
+      parsed_arguments.pw,
```
trishankatdatadog, line 312: Again, all the delegatees are expected to share the same password?

tuf/scripts/repo.py
trishankatdatadog: It's not the worst assumption as long as we clearly document it. Can we set the expected filename in a well-named variable, along with a comment?

tuf/scripts/repo.py
```diff
@@ -514,19 +526,11 @@ def add_verification_key(parsed_arguments):
...
-
-    elif parsed_arguments.role == 'snapshot':
-      repository.snapshot.add_verification_key(imported_pubkey)
-
-    # The timestamp role..
     else:
-      repository.timestamp.add_verification_key(imported_pubkey)
+      role = getattr(repository, parsed_arguments.role)
+      role.add_verification_key(imported_pubkey)
+      if role.threshold:
```
trishankatdatadog, line 534: Why the if? This is not obvious at all.

tuf/scripts/repo.py
```diff
@@ -1095,9 +1099,9 @@ def parse_arguments():
   parser.add_argument('-t', '--terminating', action='store_true',
       help='Set the terminating flag to True.  Can be used with --delegate.')
 
-  parser.add_argument('--threshold', type=int, default=1, metavar='<int>',
+  parser.add_argument('--threshold', type=int, metavar='<int>',
```
trishankatdatadog, line 1105: Why no more default=1?

trishankatdatadog (changes requested):

tuf/scripts/repo.py
jku: since --threshold is now optionally given to --trust as well, it should not have a numeric default: otherwise using --trust without --threshold leads to threshold getting reset to 1 which is unexpected.

trishankatdatadog: Got it, so we should add this note as a comment above 🙂

tuf/scripts/repo.py
trishankatdatadog: [Explanation](https://github.com/theupdateframework/tuf/pull/1473#discussion_r672829998)

tuf/scripts/repo.py
Moutix: True good point.  It was already the previous behavior, so I did not change that. To use different password, it need to not set this variable so it will ask the password in the prompt

tuf/scripts/repo.py
Moutix: Same than above

tuf/scripts/repo.py
Moutix: yes I put a comment!

trishankatdatadog:
> It's a good point, I thought that it would be set on the field 'expires' of the metadata, but after checking it's actually not the case. The parameter is completely ignore from what I can read from the generated file.
> 
> I will remove the parameters to not add confusion

Hold on, so `--expire` does nothing? How would the user change the default expiry, then?

Moutix:
> > It's a good point, I thought that it would be set on the field 'expires' of the metadata, but after checking it's actually not the case. The parameter is completely ignore from what I can read from the generated file.
> > I will remove the parameters to not add confusion
> 
> Hold on, so `--expire` does nothing? How would the user change the default expiry, then?

Good point. Yes it's not possible to update it with the current CLI. If we want to fix it, we should probably do it in another PR

trishankatdatadog:
> Good point. Yes it's not possible to update it with the current CLI. If we want to fix it, we should probably do it in another PR

Let's file an issue and move ahead then!

trishankatdatadog:
@MVrachev @jku can we please re-review? Thanks!

tuf/scripts/repo.py
```diff
@@ -282,24 +283,36 @@ def delegate(parsed_arguments):
...
-        parsed_arguments.targets_pw)
+      os.path.join(parsed_arguments.path, key),
+      parsed_arguments.targets_pw
+    )
+    parent_targets.load_signing_key(targets_private)
 
-    repository.targets.load_signing_key(targets_private)
+  # Load the private keys of the delegatee targets to allow the generation of the delegatee targets file
+  for key in parsed_arguments.pubkeys:
+    # Expect the public key to be name public.key and private key public
```
MVrachev:
```suggestion
    # Expect the public key to be named public.key and private key public
```
I am not sure I understand the whole sentence. I get that we expect the name of the public key to be
`public.key`, but I don't understand the part for the private key: `and private key public`.

tuf/scripts/repo.py
```diff
@@ -1100,6 +1101,7 @@ def parse_arguments():
   parser.add_argument('-t', '--terminating', action='store_true',
       help='Set the terminating flag to True.  Can be used with --delegate.')
 
+  # Default is set to 1 by default on new metadata but on update it doesn't change the threshold
```
MVrachev:
```suggestion
  # By default is set to 1 on new metadata but on update, it doesn't change the threshold
```

MVrachev (changes requested):
Thank you again @Moutix for your persistence while working on this pr!
Still, I have a couple of suggestions.
Please keep in mind I don't feel confident in this part of the code and it's possible that I won't catch
if there is a logical mistake somewhere.

**First**, general advice when working with GitHub - use your own custom branches to propose changes, don't use `develop`.
The rationale is that if for whatever reason your patch is declined if you use your own custom branch you won't have to remove your commits on top of `develop`, but instead you will just delete your branch.

Additionally, it will be easier for you to rebase on top of the latest changes.
Now, if there is an update in the upstream `develop` branch and a new conflict appears between your `develop` and upstream `develop` how would you resolve it? It will be a pain.

**Second**, as @jku has pointed out I miss an explanation why do you do some of the changes in your commits.
I will suggest you do the following steps:
1. Address the comments left from the reviews.
2. Squash the commits together. You can end up with a couple, not necessarily one logically separated commits. 
3. Update the commit descriptions so that it's clear why the changes are made even when reading locally in git
without GitHub. Github provides a good way to organize logically connected commits together and propose them
through pull requests, but we try to formulate our commit messages in a way that one without the internet can understand
the commit history by reading it locally.

If you need help on any of those steps, don't hesitate to ask us for help.

Moutix:
> Thank you again @Moutix for your persistence while working on this pr!
> Still, I have a couple of suggestions.
> Please keep in mind I don't feel confident in this part of the code and it's possible that I won't catch
> if there is a logical mistake somewhere.
> 
> **First**, general advice when working with GitHub - use your own custom branches to propose changes, don't use `develop`.
> The rationale is that if for whatever reason your patch is declined if you use your own custom branch you won't have to remove your commits on top of `develop`, but instead you will just delete your branch.
> 
> Additionally, it will be easier for you to rebase on top of the latest changes.
> Now, if there is an update in the upstream `develop` branch and a new conflict appears between your `develop` and upstream `develop` how would you resolve it? It will be a pain.
> 
> **Second**, as @jku has pointed out I miss an explanation why do you do some of the changes in your commits.
> I will suggest you do the following steps:
> 
>     1. Address the comments left from the reviews.
> 
>     2. Squash the commits together. You can end up with a couple, not necessarily one logically separated commits.
> 
>     3. Update the commit descriptions so that it's clear why the changes are made even when reading locally in git
>        without GitHub. Github provides a good way to organize logically connected commits together and propose them
>        through pull requests, but we try to formulate our commit messages in a way that one without the internet can understand
>        the commit history by reading it locally.
> 
> 
> If you need help on any of those steps, don't hesitate to ask us for help.



> Thank you again @Moutix for your persistence while working on this pr!
> Still, I have a couple of suggestions.
> Please keep in mind I don't feel confident in this part of the code and it's possible that I won't catch
> if there is a logical mistake somewhere.
> 
> **First**, general advice when working with GitHub - use your own custom branches to propose changes, don't use `develop`.
> The rationale is that if for whatever reason your patch is declined if you use your own custom branch you won't have to remove your commits on top of `develop`, but instead you will just delete your branch.
> 
> Additionally, it will be easier for you to rebase on top of the latest changes.
> Now, if there is an update in the upstream `develop` branch and a new conflict appears between your `develop` and upstream `develop` how would you resolve it? It will be a pain.
> 
> **Second**, as @jku has pointed out I miss an explanation why do you do some of the changes in your commits.
> I will suggest you do the following steps:
> 
>     1. Address the comments left from the reviews.
> 
>     2. Squash the commits together. You can end up with a couple, not necessarily one logically separated commits.
> 
>     3. Update the commit descriptions so that it's clear why the changes are made even when reading locally in git
>        without GitHub. Github provides a good way to organize logically connected commits together and propose them
>        through pull requests, but we try to formulate our commit messages in a way that one without the internet can understand
>        the commit history by reading it locally.
> 
> 
> If you need help on any of those steps, don't hesitate to ask us for help.

Hey @MVrachev I think I've adressed all comment for this PR, I've as well squashed all the commits together and updated the description so it should be better

MVrachev:
@Moutix that's a lot better.
@jku is on PTO and will be back on Thursday and around then you can expect to hear from him.
@joshuagl the other maintainer is on PTO this week as well.
So, please be patient if you want to hear from them.

jku (approved):
Looks good to me, thank you for your persistence on this.

jku:
> trishankatdatadog requested changes

I think these are about the password arguments... This PR does make the repo.py password arg handling issues more visible but fixing them seems like a big ask. @trishankatdatadog did you want changes here or do you want to remove the request?

trishankatdatadog:
> I think these are about the password arguments... This PR does make the repo.py password arg handling issues more visible but fixing them seems like a big ask. @trishankatdatadog did you want changes here or do you want to remove the request?

Let's explicitly document the current behaviour, please, and also file an issue to solve it in a future PR.

lukpueh:
Thank you for you contribution! This PR is closed as it was submitted against what is now known as the legacy codebase.

**More details about legacy code removal**

Current source code (and upcoming 1.0 release) only contains the modern components
 * a low-level Metadata API (`tuf.api`) and
 * `tuf.ngclient` that implements the client workflow,

Legacy components (`tuf.client`, `tuf.repository_tool`, `tuf.repository_lib` as well as the repo and client scripts) are no longer included. See [announcement](https://github.com/theupdateframework/python-tuf/blob/develop/docs/1.0.0-ANNOUNCEMENT.md) for more information and [API reference](https://theupdateframework.readthedocs.io/en/latest/api/api-reference.html) for details about the current API.

PR #1474, opened 2021-07-01
Document ngclient

### LINKED ISSUE #1385: ngclient: Proper public module documentation
New client should have good public module/class documentation for Updater: currently it's a bit patchy.

_Plan is to get the new client merged to develop (as alternative client, not replacing the current one): this issue can be solved after merge._

jku:
* Document the ngclient.updater module
* Include the simplest possible example
* Try out some sphinx links and layout

I've tested this with sphinx (+ napoleon plugin for google docstrings) and it works.

I did not go all-in with the sphinx links as it looks quite ugly in the source code -- not sure which to optimize here.

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -1,7 +1,61 @@
 # Copyright 2020, New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-"""TUF client workflow implementation.
+"""TUF module that implements the client update workflow.
+
+This module contains the Updater class that provides an implementation of the
+`TUF client workflow
+<https://theupdateframework.github.io/specification/latest/#detailed-client-workflow>`_.
+Updater provides an API to query available targets and to download them in a
+secure manner: All downloaded files are verified by signed metadata.
+
+High-level description of Updater functionality:
+  * Initializing an :class:`~tuf.ngclient.updater.Updater` loads and validates
+    the trusted local root metadata: This root metadata is used as the source
+    of trust for all other metadata.
+  * Calling :func:`~tuf.ngclient.updater.Updater.refresh()` will update root
+    metadata and load all other top-level metadata as described in the
+    specification, using both locally cached metadata and metadata downloaded
+    from the remote repository.
+  * When metadata is up-to-date, targets can be dowloaded. The repository
+    snapshot is consistent so multiple targets can be downloaded without
+    fear of repository content changing. For each target:
+      * :func:`~tuf.ngclient.updater.Updater.get_one_valid_targetinfo()` is
+        used to find information about a specific target. This will load new
+        targets metadata as needed (from local cache or remote repository).
+      * :func:`~tuf.ngclient.updater.Updater.updated_targets()` can be used to
+        check if target files are already locally cached.
+      * :func:`~tuf.ngclient.updater.Updater.download_target()` downloads a
+        target file and ensures it is verified correct by the metadata.
+
+Below is a simple example of using the Updater to download and verify
+"file.txt" from a remote repository. The required environment for this example
+is:
+    * A webserver running on http://localhost:8000, serving TUF repository
+      metadata at "/tuf-repo/" and targets at "/targets/"
+    * Local metadata directory "~/tufclient/metadata/" is writable and contains
+      a root metadata version for the remote repository
+    * Download directory "~/target-downloads/" is writable
+
+Example::
+
+    from tuf.ngclient import Updater
+
+    # Load trusted local root metadata from client metadata cache. Define the
+    # remote repository metadata URL prefix and target URL prefix.
+    updater = Updater(
+        repository_dir="~/tufclient/metadata/",
+        metadata_base_url="http://localhost:8000/tuf-repo/",
+        target_base_url="http://localhost:8000/targets/",
+    )
+
+    # Update top-level metadata from remote
+    updater.refresh()
+
+    # Securely download a target:
+    # Update target metadata, then download and verify target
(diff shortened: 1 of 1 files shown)


jku:
I should file issues for the "implement docs build, publish on readthedocs" steps but I'm not entirely sure what that really includes...

My quick-and-dirty test method:
* install sphinx
* configure with `sphinx-quickstart`
* Add `extensions = ['sphinx.ext.napoleon']` to conf.py to get support for google style docstrings
* run docs build (e.g. `make html`)

tuf/ngclient/updater.py
```diff
@@ -1,7 +1,61 @@
...
+Updater provides an API to query available targets and to download them in a
+secure manner: All downloaded files are verified by signed metadata.
+
+High-level description of Updater functionality:
+  * Initializing an :class:`~tuf.ngclient.updater.Updater` loads and validates
+    the trusted local root metadata: This root metadata is used as the source
+    of trust for all other metadata.
+  * Calling :func:`~tuf.ngclient.updater.Updater.refresh()` will update root
+    metadata and load all other top-level metadata as described in the
+    specification, using both locally cached metadata and metadata downloaded
```
sechkova, line 18: Maybe "using either locally cached metadata or metadata downloaded"?
I'm not sure how to phrase it differently but "both cached and downloaded metadata" may sound confusing.

tuf/ngclient/updater.py
```diff
@@ -1,7 +1,61 @@
...
+        metadata_base_url="http://localhost:8000/tuf-repo/",
+        target_base_url="http://localhost:8000/targets/",
+    )
+
+    # Update top-level metadata from remote
+    updater.refresh()
+
+    # Securely download a target:
+    # Update target metadata, then download and verify target
+    targetinfo = updater.get_one_valid_targetinfo("file.txt")
```
sechkova, line 57: Is `tuf.ngclient.updater.Updater.updated_targets()` intentionally skipped in the example?

tuf/ngclient/updater.py
```diff
@@ -0,0 +1,642 @@
+# Copyright 2020, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""TUF module that implements the client update workflow.
+
+This module contains the Updater class that provides an implementation of the
+TUF client workflow specified in
+https://theupdateframework.github.io/specification/latest/#detailed-client-workflow.
```
sechkova: If we are going for the linking, this could become:

```suggestion
 `TUF client workflow <https://theupdateframework.github.io/specification/latest/#detailed-client-workflow>`_ 
```

sechkova (commented):
I think this is a good starting point for documenting the client. We can most likely find a ways to improve it while we keep tuning the client API and publishing the docs.

Yeah, the links look a bit ugly in the source but are useful in the html.

sechkova:
> My quick-and-dirty test method:
> 
>     * install sphinx
> 
>     * configure with `sphinx-quickstart`
> 
>     * Add `extensions = ['sphinx.ext.napoleon']` to conf.py to get support for google style docstrings
> 
>     * run docs build (e.g. `make html`)

I had to add some settings so that `autodoc` is able to find the module too? 
edit: probably I didn't import TUF  from the very beginning and the autodocs tweaks were not needed.

tuf/ngclient/updater.py
jku: I did intentionally make it the simplest possible example... 

The local cache implementation IMO 
* is what makes the API ugly
* is not part of the spec 
* results in some surprising side-effects (I just want to download a file but the tool creates a whole directory structure)
* creates problems that the updater does not need to solve (who maintains the cache? do all targetpath URLs actually map to a filesystem path on this specific filesystem?)

I'd really like to try and refactor the caching somewhere outside the updater.

tuf/ngclient/updater.py
jku: I mean, typically it does use both locally cached and remote metadata ...

tuf/ngclient/updater.py
```diff
@@ -1,7 +1,61 @@
...
+  * Initializing an :class:`~tuf.ngclient.updater.Updater` loads and validates
+    the trusted local root metadata: This root metadata is used as the source
+    of trust for all other metadata.
+  * Calling :func:`~tuf.ngclient.updater.Updater.refresh()` will update root
+    metadata and load all other top-level metadata as described in the
+    specification, using both locally cached metadata and metadata downloaded
+    from the remote repository.
+  * When metadata is up-to-date, targets can be dowloaded. The repository
+    snapshot is consistent so multiple targets can be downloaded without
+    fear of repository content changing. For each target:
```
MVrachev, line 22: I am not sure I understand that sentence: ```The repository  snapshot is consistent so multiple targets can be downloaded without fear of repository content changing.```
What do you mean by consistent?

jku: that even if there is a longer period of time between downloading one targets metadata and another targets metadata, they will use metadata versions from the same snapshot (defined at refresh() time): so there's no fear of accidentally downloading one file that was added into the repo after the update started. All target files that are downloaded are from one consistent snapshot.

Maybe this does not need to be here at all, not sure...

MVrachev: Ahaaa, now it makes sense.
It's a good explanation for a new reader. I am used to that idea and have never thought about it, but it makes sense to capture it for new readers.
Maybe you can say it like that:
```The repository snapshot is consistent and can be reused so multiple targets can be downloaded without fear of repository content changing.```

joshuagl (approved):
Thanks, this looks like a decent start. Let's merge and iterate.

PR #1477, opened 2021-07-01
ng client improve TrustedMetadataSet testing

### LINKED ISSUE #1465: ngclient TrustedMetadataSet: write tests to handling coverage report
This is the minimal amount of testing we should have and the first step of resolving #1462 for the MetadataBundle.
When resolving the missing lines reported by `coverage` we will decide if we want to handle all of them
or there is a sense of leaving some of them.

MVrachev:
Fixes #1465

The current situation with the TrustedMetadataSet testing is that
we don't have a minimal amount of unit tests testing the different
branches in the various API functionality in the class.

This commit proposes simple unit tests covering almost all of the
branches in the API functions and increasing the unit test coverage
(as reported from the "coverage" tool) from 74 % to 97 %.

The code could be complicated at places because the different
branches in the update_* functions depend on other metadata classes
as well.

**History of modifications made on the pr**:

1. An initial version of the pr: 
  1.1 Move the shared code between tests into the "setupClass" function.
  1.2 Add longer test functions with multiple test cases inside them
2. Split longer test functions into multiple smaller ones
  2.1 Added a couple of `setup` helper functions calling `trusted_metadata_set.update_*` 
  aiming to move the common code between tests into separate functions.
3. Addressing multiple suggestions and optimizations by @jku.
4. Experimenting with Jussi suggestion that instead of modifying internal trusted_metadata_set state
we should modify the input data we initially feed for trusted_metadata_set.
  4.1 Add `modify_snapshot`, `modify_timestamp` helper functions splitting common code that
  now can be reused. They end up with quite complicated functions because I wanted to reuse them
  in different scenarios.
5. By suggestion of @jku, replacing `modify_snapshot` and `modify_timestamp` with a 
higher order function `modify_metadata` which accepts a modification function as a parameter 
then calls it. This allows the actual modification function to be defined inside the test itself and 
be a lot simpler and still reuse a lot of the code responsible to serialization, signing and other 
operations inside the `modify_metadata` function.

requirements**:

### DIFF
--- a/tests/test_trusted_metadata_set.py
+++ b/tests/test_trusted_metadata_set.py
@@ -16,36 +16,37 @@ logger = logging.getLogger(__name__)
 
 class TestTrustedMetadataSet(unittest.TestCase):
 
-    def test_update(self):
-        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
+    @classmethod
+    def setUpClass(cls):
+        cls.repo_dir = os.path.join(
+            os.getcwd(), 'repository_data', 'repository', 'metadata'
+        )
+        cls.metadata = {}
+        for md in ["root", "timestamp", "snapshot", "targets", "role1", "role2"]:
+            with open(os.path.join(cls.repo_dir, f"{md}.json"), "rb") as f:
+                cls.metadata[md] = f.read()
 
-        with open(os.path.join(repo_dir, "root.json"), "rb") as f:
-            trusted_set = TrustedMetadataSet(f.read())
+
+    def test_update(self):
+        trusted_set = TrustedMetadataSet(self.metadata["root"])
         trusted_set.root_update_finished()
 
-        with open(os.path.join(repo_dir, "timestamp.json"), "rb") as f:
-            trusted_set.update_timestamp(f.read())
-        with open(os.path.join(repo_dir, "snapshot.json"), "rb") as f:
-            trusted_set.update_snapshot(f.read())
-        with open(os.path.join(repo_dir, "targets.json"), "rb") as f:
-            trusted_set.update_targets(f.read())
-        with open(os.path.join(repo_dir, "role1.json"), "rb") as f:
-            trusted_set.update_delegated_targets(f.read(), "role1", "targets")
-        with open(os.path.join(repo_dir, "role2.json"), "rb") as f:
-            trusted_set.update_delegated_targets(f.read(), "role2", "role1")
+        trusted_set.update_timestamp(self.metadata["timestamp"])
+        trusted_set.update_snapshot(self.metadata["snapshot"])
+        trusted_set.update_targets(self.metadata["targets"])
+        trusted_set.update_delegated_targets(
+            self.metadata["role1"], "role1", "targets"
+        )
+        trusted_set.update_delegated_targets(
+            self.metadata["role2"], "role2", "role1"
+        )
 
     def test_out_of_order_ops(self):
-        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
-        data={}
-        for md in ["root", "timestamp", "snapshot", "targets", "role1"]:
-            with open(os.path.join(repo_dir, f"{md}.json"), "rb") as f:
-                data[md] = f.read()
-
-        trusted_set = TrustedMetadataSet(data["root"])
+        trusted_set = TrustedMetadataSet(self.metadata["root"])
 
         # Update timestamp before root is finished
         with self.assertRaises(RuntimeError):
-            trusted_set.update_timestamp(data["timestamp"])
+            trusted_set.update_timestamp(self.metadata["timestamp"])
 
         trusted_set.root_update_finished()
         with self.assertRaises(RuntimeError):
@@ -53,50 +54,53 @@ class TestTrustedMetadataSet(unittest.TestCase):
 
         # Update snapshot before timestamp
         with self.assertRaises(RuntimeError):
-            trusted_set.update_snapshot(data["snapshot"])
+            trusted_set.update_snapshot(self.metadata["snapshot"])
 
-        trusted_set.update_timestamp(data["timestamp"])
+        trusted_set.update_timestamp(self.metadata["timestamp"])
 
         # Update targets before snapshot
         with self.assertRaises(RuntimeError):
-            trusted_set.update_targets(data["targets"])
+            trusted_set.update_targets(self.metadata["targets"])
 
-        trusted_set.update_snapshot(data["snapshot"])
+        trusted_set.update_snapshot(self.metadata["snapshot"])
 
-        #update timestamp after snapshot
+        # update timestamp after snapshot
         with self.assertRaises(RuntimeError):
-            trusted_set.update_timestamp(data["timestamp"])
+            trusted_set.update_timestamp(self.metadata["timestamp"])
 
         # Update delegated targets before targets
         with self.assertRaises(RuntimeError):
-            trusted_set.update_delegated_targets(data["role1"], "role1", "targets")
+            trusted_set.update_delegated_targets(
+                self.metadata["role1"], "role1", "targets"
+            )
 
-        trusted_set.update_targets(data["targets"])
-        trusted_set.update_delegated_targets(data["role1"], "role1", "targets")
+        trusted_set.update_targets(self.metadata["targets"])
+        trusted_set.update_delegated_targets(
+            self.metadata["role1"], "role1", "targets"
+        )
+
+        trusted_set.update_targets(self.metadata["targets"])
+        trusted_set.update_delegated_targets(
+            self.metadata["role1"], "role1", "targets"
+        )
 
     def test_update_with_invalid_json(self):
-        repo_dir = os.path.join(os.getcwd(), 'repository_data', 'repository', 'metadata')
-        data={}
-        for md in ["root", "timestamp", "snapshot", "targets", "role1"]:
-            with open(os.path.join(repo_dir, f"{md}.json"), "rb") as f:
-                data[md] = f.read()
-
         # root.json not a json file at all
         with self.assertRaises(exceptions.RepositoryError):
             TrustedMetadataSet(b"")
         # root.json is invalid
-        root = Metadata.from_bytes(data["root"])
+        root = Metadata.from_bytes(self.metadata["root"])
         root.signed.version += 1
         with self.assertRaises(exceptions.RepositoryError):
             TrustedMetadataSet(json.dumps(root.to_dict()).encode())
 
-        trusted_set = TrustedMetadataSet(data["root"])
+        trusted_set = TrustedMetadataSet(self.metadata["root"])
         trusted_set.root_update_finished()
 
         top_level_md = [
-            (data["timestamp"], trusted_set.update_timestamp),
-            (data["snapshot"], trusted_set.update_snapshot),
-            (data["targets"], trusted_set.update_targets),
+            (self.metadata["timestamp"], trusted_set.update_timestamp),
+            (self.metadata["snapshot"], trusted_set.update_snapshot),
+            (self.metadata["targets"], trusted_set.update_targets),
         ]
         for metadata, update_func in top_level_md:
             # metadata is not json
@@ -110,7 +114,7 @@ class TestTrustedMetadataSet(unittest.TestCase):
 
             # metadata is of wrong type
             with self.assertRaises(exceptions.RepositoryError):
-                update_func(data["root"])
+                update_func(self.metadata["root"])
 
             update_func(metadata)
 
--- a/tests/test_trusted_metadata_set.py
+++ b/tests/test_trusted_metadata_set.py
@@ -1,14 +1,22 @@
+import copy
 import json
 import logging
 import os
-import shutil
 import sys
-import tempfile
 import unittest
+from typing import Dict, Any
+from datetime import datetime
 
 from tuf import exceptions
 from tuf.api.metadata import Metadata
-from tuf.ngclient._internal.trusted_metadata_set import TrustedMetadataSet
+from tuf.ngclient._internal.trusted_metadata_set import(
+    TrustedMetadataSet
+)
+from securesystemslib.signer import SSlibSigner
+from securesystemslib.interface import(
+    import_ed25519_privatekey_from_file,
+    import_rsa_privatekey_from_file
+)
 
 from tests import utils
 
@@ -26,63 +34,91 @@ class TestTrustedMetadataSet(unittest.TestCase):
             with open(os.path.join(cls.repo_dir, f"{md}.json"), "rb") as f:
                 cls.metadata[md] = f.read()
 
+        keystore_dir = os.path.join(os.getcwd(), 'repository_data', 'keystore')
+        cls.keystore = {}
+        root_key_dict = import_rsa_privatekey_from_file(
+            os.path.join(keystore_dir, "root" + '_key'),
+            password="password"
+        )
+        cls.keystore["root"] = SSlibSigner(root_key_dict)
+        for role in ["delegation", "snapshot", "targets", "timestamp"]:
+            key_dict = import_ed25519_privatekey_from_file(
+                os.path.join(keystore_dir, role + '_key'),
+                password="password"
+            )
+            cls.keystore[role] = SSlibSigner(key_dict)
+
+    def setUp(self) -> None:
+        self.trusted_set = TrustedMetadataSet(self.metadata["root"])
+
+    def _root_update_finished_and_update_timestamp(self):
+        self.trusted_set.root_update_finished()
+        self.trusted_set.update_timestamp(self.metadata["timestamp"])
+
+    def _update_all_besides_targets(self):
+        self.trusted_set.root_update_finished()
+        self.trusted_set.update_timestamp(self.metadata["timestamp"])
+        self.trusted_set.update_snapshot(self.metadata["snapshot"])
 
     def test_update(self):
-        trusted_set = TrustedMetadataSet(self.metadata["root"])
-        trusted_set.root_update_finished()
-
-        trusted_set.update_timestamp(self.metadata["timestamp"])
-        trusted_set.update_snapshot(self.metadata["snapshot"])
-        trusted_set.update_targets(self.metadata["targets"])
-        trusted_set.update_delegated_targets(
+        self.trusted_set.root_update_finished()
+        self.trusted_set.update_timestamp(self.metadata["timestamp"])
+        self.trusted_set.update_snapshot(self.metadata["snapshot"])
+        self.trusted_set.update_targets(self.metadata["targets"])
+        self.trusted_set.update_delegated_targets(
             self.metadata["role1"], "role1", "targets"
         )
-        trusted_set.update_delegated_targets(
+        self.trusted_set.update_delegated_targets(
             self.metadata["role2"], "role2", "role1"
         )
+        # the 4 top level metadata objects + 2 additional delegated targets
+        self.assertTrue(len(self.trusted_set), 6)
 
     def test_out_of_order_ops(self):
-        trusted_set = TrustedMetadataSet(self.metadata["root"])
-
         # Update timestamp before root is finished
         with self.assertRaises(RuntimeError):
-            trusted_set.update_timestamp(self.metadata["timestamp"])
+            self.trusted_set.update_timestamp(self.metadata["timestamp"])
 
-        trusted_set.root_update_finished()
+        self.trusted_set.root_update_finished()
         with self.assertRaises(RuntimeError):
-            trusted_set.root_update_finished()
+            self.trusted_set.root_update_finished()
+
+        # Update root after a previous successful root update
+        with self.assertRaises(RuntimeError):
+            self.trusted_set.update_root(self.metadata["root"])
 
         # Update snapshot before timestamp
         with self.assertRaises(RuntimeError):
-            trusted_set.update_snapshot(self.metadata["snapshot"])
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
 
-        trusted_set.update_timestamp(self.metadata["timestamp"])
+        self.trusted_set.update_timestamp(self.metadata["timestamp"])
 
         # Update targets before snapshot
         with self.assertRaises(RuntimeError):
-            trusted_set.update_targets(self.metadata["targets"])
+            self.trusted_set.update_targets(self.metadata["targets"])
 
-        trusted_set.update_snapshot(self.metadata["snapshot"])
+        self.trusted_set.update_snapshot(self.metadata["snapshot"])
 
         # update timestamp after snapshot
         with self.assertRaises(RuntimeError):
-            trusted_set.update_timestamp(self.metadata["timestamp"])
+            self.trusted_set.update_timestamp(self.metadata["timestamp"])
 
         # Update delegated targets before targets
         with self.assertRaises(RuntimeError):
-            trusted_set.update_delegated_targets(
+            self.trusted_set.update_delegated_targets(
                 self.metadata["role1"], "role1", "targets"
             )
 
-        trusted_set.update_targets(self.metadata["targets"])
-        trusted_set.update_delegated_targets(
+        self.trusted_set.update_targets(self.metadata["targets"])
+
+        # Update snapshot after sucessful targets update
+        with self.assertRaises(RuntimeError):
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
+
+        self.trusted_set.update_delegated_targets(
             self.metadata["role1"], "role1", "targets"
         )
 
-        trusted_set.update_targets(self.metadata["targets"])
-        trusted_set.update_delegated_targets(
-            self.metadata["role1"], "role1", "targets"
-        )
 
     def test_update_with_invalid_json(self):
         # root.json not a json file at all
@@ -94,20 +130,27 @@ class TestTrustedMetadataSet(unittest.TestCase):
         with self.assertRaises(exceptions.RepositoryError):
             TrustedMetadataSet(json.dumps(root.to_dict()).encode())
 
-        trusted_set = TrustedMetadataSet(self.metadata["root"])
-        trusted_set.root_update_finished()
+        # update_root called with the wrong metadata type
+        with self.assertRaises(exceptions.RepositoryError):
+            self.trusted_set.update_root(self.metadata["snapshot"])
+
+        self.trusted_set.root_update_finished()
 
         top_level_md = [
-            (self.metadata["timestamp"], trusted_set.update_timestamp),
-            (self.metadata["snapshot"], trusted_set.update_snapshot),
-            (self.metadata["targets"], trusted_set.update_targets),
+            (self.metadata["timestamp"], self.trusted_set.update_timestamp),
+            (self.metadata["snapshot"], self.trusted_set.update_snapshot),
+            (self.metadata["targets"], self.trusted_set.update_targets),
         ]
         for metadata, update_func in top_level_md:
+            md = Metadata.from_bytes(metadata)
+            if md.signed.type == "snapshot":
+                # timestamp hashes and length intervene when testing snapshot
+                self.trusted_set.timestamp.signed.meta["snapshot.json"].hashes = None
+                self.trusted_set.timestamp.signed.meta["snapshot.json"].length = None
             # metadata is not json
             with self.assertRaises(exceptions.RepositoryError):
                 update_func(b"")
             # metadata is invalid
-            md = Metadata.from_bytes(metadata)
             md.signed.version += 1
             with self.assertRaises(exceptions.RepositoryError):
                 update_func(json.dumps(md.to_dict()).encode())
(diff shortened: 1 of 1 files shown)


MVrachev:
Ready for review after #1408 has been merged.

tests/test_trusted_metadata_set.py
```diff
@@ -118,9 +141,228 @@ def test_update_with_invalid_json(self):
...
+        with self.assertRaises(ValueError):
+            verify_with_threshold(self.targets, "foo", self.delegated_role)
+
+    def test_invalid_update_root(self):
+        trusted_set = TrustedMetadataSet(self.metadata["root"])
+        # new_root data with invalid _type
+        invalid_type_data = json.loads(self.metadata["root"])
+        invalid_type_data["signed"]["_type"] = "foo"
+        invalid_type_data = json.dumps(invalid_type_data)
+        # RepositoryError is throwned during new_root deserialization.
```
sechkova: throwned -> thrown

tests/test_trusted_metadata_set.py
```diff
@@ -118,9 +141,228 @@ def test_update_with_invalid_json(self):
...
+
+        # new_root.signed.version has the same version as old root
+        with self.assertRaises(exceptions.ReplayedMetadataError):
+            trusted_set.update_root(self.metadata["root"])
+
+        # if _root_update_finished, then fail when calling update_root
+        trusted_set.root_update_finished()
+        with self.assertRaises(RuntimeError):
+            trusted_set.update_root(self.metadata["root"])
+        trusted_set._root_update_finished = False
```
sechkova: Last line seems redundant.

tests/test_trusted_metadata_set.py
```diff
@@ -26,6 +33,19 @@ def setUpClass(cls):
             with open(os.path.join(cls.repo_dir, f"{md}.json"), "rb") as f:
                 cls.metadata[md] = f.read()
 
+        cls.root = Metadata.from_bytes(cls.metadata["root"])
+        cls.timestamp = Metadata.from_bytes(cls.metadata["timestamp"])
+        cls.snapshot = Metadata.from_bytes(cls.metadata["snapshot"])
+        cls.targets = Metadata.from_bytes(cls.metadata["targets"])
+        cls.delegated_role = Metadata.from_bytes(cls.metadata["role1"])
```
sechkova: I don't see these class variables used anywhere.

tests/test_trusted_metadata_set.py
```diff
@@ -118,9 +141,228 @@ def test_update_with_invalid_json(self):
...
+            verify_with_threshold(self.targets, "foo", self.delegated_role)
+
+    def test_invalid_update_root(self):
+        trusted_set = TrustedMetadataSet(self.metadata["root"])
+        # new_root data with invalid _type
+        invalid_type_data = json.loads(self.metadata["root"])
+        invalid_type_data["signed"]["_type"] = "foo"
+        invalid_type_data = json.dumps(invalid_type_data)
+        # RepositoryError is throwned during new_root deserialization.
+        # It's not throwned when checking new_root.signed.type != "root"
```
sechkova: If the goal is to test the case described in the comment too, there should be another test setting
`invalid_root_type_data["signed"]["_type"] = "snapshot"`

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +137,232 @@ def test_update_with_invalid_json(self):
...
+            modified_hashes[algo] = observed_hash
+        return modified_hashes
+
+    def test_update_snapshot(self):
+        trusted_set = TrustedMetadataSet(self.metadata["root"])
+        trusted_set.root_update_finished()
+        trusted_set.update_timestamp(self.metadata["timestamp"])
+        # cannot update snapshot after targets update completes or targets != None
+        targets_obj = Metadata.from_bytes(self.metadata["targets"])
+        trusted_set._trusted_set["targets"] = targets_obj
```
sechkova:
```suggestion
        trusted_set.targets = targets_obj
```

tests/test_trusted_metadata_set.py
```diff
@@ -118,9 +141,228 @@ def test_update_with_invalid_json(self):
...
+        new_timestamp_byte_data = json.dumps(timestamp.to_dict()).encode()
+        with self.assertRaises(exceptions.ExpiredMetadataError):
+            trusted_set.update_timestamp(new_timestamp_byte_data)
+
+    def _calculate_modified_hashes(
+        self, true_hashes,
+        data: bytes
+    ) -> Dict[str, str]:
+        modified_hashes = {}
+        # Calculate correct hashes to pass the first check
```
sechkova: What is "the first check"?

tests/test_trusted_metadata_set.py
```diff
@@ -118,9 +141,228 @@ def test_update_with_invalid_json(self):
...
+        trusted_set.root_update_finished()
+        trusted_set.update_timestamp(self.metadata["timestamp"])
+        # cannot update snapshot after targets update completes or targets != None
+        targets_obj = Metadata.from_bytes(self.metadata["targets"])
+        trusted_set._trusted_set["targets"] = targets_obj
+        with self.assertRaises(RuntimeError):
+            trusted_set.update_snapshot(self.metadata["snapshot"])
+        del trusted_set._trusted_set["targets"]
+
+        #  Deserialization error - failed to load new_snapshot.
```
sechkova: Maybe a better comment is needed, I'm not sure what is tested here.

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +137,232 @@ def test_update_with_invalid_json(self):
...
+        snapshot_obj = Metadata.from_bytes(self.metadata["snapshot"])
+        snapshot_obj.signed.meta = {}
+        # prepare timestamp.meta["snapshot"].hashes
+        self._sign_modified_obj("snapshot", snapshot_obj)
+        timestamp_meta = trusted_set.timestamp.signed.meta["snapshot.json"]
+        true_hashes = timestamp_meta.hashes or {}
+        modified_snapshot_data = json.dumps(snapshot_obj.to_dict()).encode()
+        modified_hashes = self._calculate_modified_hashes(
+            true_hashes, json.dumps(snapshot_obj.to_dict()).encode()
+        )
```
sechkova:
```suggestion
        modified_snapshot_data = json.dumps(snapshot_obj.to_dict()).encode()
        modified_hashes = self._calculate_modified_hashes(
            true_hashes, modified_snapshot_data
        )
```

sechkova (commented):
LGTM with small comments below.

tests/test_trusted_metadata_set.py
MVrachev: Great catch!
Modified my test to use that one instead of "foo".

tests/test_trusted_metadata_set.py
MVrachev: At the time of the assignment, targets is `None` and that's why I assign it that way because otherwise, I get `AttributeError: can't set attribute`.

tests/test_trusted_metadata_set.py
MVrachev: They were used, but only once and you are right that if that's the case, then it's better to create those objects close to where they will be used.
Removed those lines.

MVrachev:
Rebased on top of develop branch to get the latest changes related to hashes and length validation.

Also, I added two new commits:
- the first one `TrustedMetadataSet testing: start using setUp` uses `setUp` to create a trusted metadata set instance per test
- the second one splits the big test functions into smaller ones separated by a test case. This required creating a couple of extra small helpful functions.

Can you have another look @sechkova and tell me do you think it's better that way using many smaller functions?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
 
             # metadata is of wrong type
             with self.assertRaises(exceptions.RepositoryError):
-                update_func(data["root"])
+                update_func(self.metadata["root"])
 
             update_func(metadata)
 
 
+    def test_verify_with_threshold_wrong_delegator(self):
```
jku: let's not add tests for verify_with_threshold(): it's a placeholder until #1436 merges

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        with self.assertRaises(exceptions.ExpiredMetadataError):
+            self.trusted_set.root_update_finished()
+
+    def _sign_modified_obj(
+        self,
+        role:str,
+        metadata_obj: Metadata
+    ) -> Dict[str, Any]:
+        key_dict = self.keystore[role]
+        sslib_signer = SSlibSigner(key_dict)
```
jku: self.keystore could be self.signers (`Dict[str, Signer]`) instead so you didn't need to create a Signer every time

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+    def test_update_root_invalid_type(self):
+        # new_root data with invalid snapshot type
+        invalid_type_data = json.loads(self.metadata["root"])
+        invalid_type_data["signed"]["_type"] = "snapshot"
+        invalid_type_data["signed"]["meta"] = {"file1.txt": {"version": 1}}
+        invalid_type_data = json.dumps(invalid_type_data).encode()
+        # RepositoryError is thrown during new_root deserialization.
+        # It's not thrown when checking new_root.signed.type != "root"
+        with self.assertRaises(exceptions.RepositoryError):
+            self.trusted_set.update_root(invalid_type_data)
```
jku: this function can be replaced by 
```
with self.assertRaises(exceptions.RepositoryError):
    self.trusted_set.update_root(self.metadata["snapshot"])
```
correct?

This same test seems to get done in `test_update_with_invalid_json()` already but only for the other top-level metadata, not root... these probably all belong together: would it make sense to move the other "invalid type" tests here as well?

Alternatively, maybe the tests in `test_update_with_invalid_json()` don't actually test what they claim to (because of mismatching metafiles and such...)? if you know that this is the case then feel free to remove the useless ones

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+
+    def test_update_root_new_root_cannot_be_verified_with_threshold(self):
+        # new_root data with threshold which cannot be verified.
+        modified_threshold_data = copy.deepcopy(
+            json.loads(self.metadata["root"])
+        )
+        modified_threshold_data["signed"]["roles"]["root"]["threshold"] = 2
+        modified_threshold_data = json.dumps(modified_threshold_data).encode()
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            self.trusted_set.update_root(modified_threshold_data)
```
jku: the way the test is written leads me to think the threshold value is what's important -- when reality it's just that the signature is no longer valid. Maybe modify signed.version instead -- that's shorter as well?

Also, almost all other test construct a Metadata when they need to do modifications, maybe would be good to be consistent?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+    def test_update_root_new_root_ver_same_as_trusted_root_ver(self):
+        with self.assertRaises(exceptions.ReplayedMetadataError):
+            self.trusted_set.update_root(self.metadata["root"])
+
+    def test_update_root_after_successful_root_update(self):
+        # if _root_update_finished, then fail when calling update_root
+        self.trusted_set.root_update_finished()
+        with self.assertRaises(RuntimeError):
+            self.trusted_set.update_root(self.metadata["root"])
+        self.trusted_set._root_update_finished = False
```
jku: `test_out_of_order_ops()` checks for these types of errors -- if it does not check for this already, maybe you could add this in the right place there?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+
+    def test_root_update_finished_expired(self):
+        # call root_update_finished when trusted root has expired
+        expired_datetime = datetime.strptime(
+            "1970-01-01T00:00:00Z", "%Y-%m-%dT%H:%M:%SZ"
+        )
+        self.trusted_set.root.signed.expires = expired_datetime
+        with self.assertRaises(exceptions.ExpiredMetadataError):
+            self.trusted_set.root_update_finished()
```
jku: you don't need the parsing, something like this works:
`    self.trusted_set.root.signed.expires = datetime(1970, 1, 1)`

The test isn't wrong... but I think it would be better if it actually did load the expired root metadata (to show that the loading works for expired metadata), and then fails on root_update_finished(). The same comment applies to a lot of tests... so we can discuss that in general, no need to do that for this test specifically

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        with self.assertRaises(exceptions.ReplayedMetadataError):
+            self.trusted_set.update_timestamp(self.metadata["timestamp"])
+
+    def test_update_timestamp_snapshot_ver_below_trusted_snapshot_ver(self):
+        self._setup_update_snapshot_or_timestamp_test()
+        # new_timestamp has expired
+        timestamp = Metadata.from_bytes(self.metadata["timestamp"])
+        timestamp.signed.expires = datetime.strptime(
+            "1970-01-01T00:00:00Z", "%Y-%m-%dT%H:%M:%SZ"
+        )
```
jku: no need for parsing here either

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        self._sign_modified_obj("timestamp", timestamp)
+        new_timestamp_byte_data = json.dumps(timestamp.to_dict()).encode()
```
jku: I think this shows we should have a `Timestamp.to_bytes()` to complete the set of de/serialize functions (not in this PR, just commenting)

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+            modified_hashes[algo] = observed_hash
+        return modified_hashes
+
+    def test_update_snapshot_new_snapshot_invalid_type(self):
+        self._setup_update_snapshot_or_timestamp_test()
+        # new_snapshot data with invalid targets type
+        invalid_type_data = json.loads(self.metadata["snapshot"])
+        invalid_type_data["signed"]["_type"] = "targets"
+        invalid_type_data["signed"]["targets"] = {}
+        invalid_type_data = json.dumps(invalid_type_data).encode()
```
jku: is there some advantage to this over just using `self.metadata["targets"]` as the data?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        with self.assertRaises(exceptions.RepositoryError):
+            self.trusted_set.update_snapshot(invalid_type_data)
+
+    def test_update_snapshot_after_targets_updated(self):
+        self._setup_update_snapshot_or_timestamp_test()
+        # cannot update snapshot after targets update completes or targets != None
+        targets_obj = Metadata.from_bytes(self.metadata["targets"])
+        self.trusted_set._trusted_set["targets"] = targets_obj
+        with self.assertRaises(RuntimeError):
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
```
jku: I don't totally understand this:
* Do you need to modify trusted_set internals? That should be absolute last option if nothing else works
* shouldn't this logically fit in `test_out_of_order_ops()` (I thought it would test it already but maybe not)?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        data: bytes
+    ) -> Dict[str, str]:
+        modified_hashes = {}
+        # Calculate hashes on modified data in order to pass hashes verification.
+        for algo in true_hashes.keys():
+            digest_object = sslib_hash.digest(algo)
+            digest_object.update(data)
+            observed_hash = digest_object.hexdigest()
+            modified_hashes[algo] = observed_hash
+        return modified_hashes
```
jku: This method isn't needed? MetaFile does not require hashes so `None` will work fine

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        true_hashes = timestamp_meta.hashes or {}
+
+        modified_hashes = self._calculate_modified_hashes(
+            true_hashes, b'{""sig": }'
+        )
+        self.trusted_set.timestamp.signed.meta["snapshot.json"].hashes = modified_hashes
+        with self.assertRaises(exceptions.RepositoryError):
+            self.trusted_set.update_snapshot(b'{""sig": }')
+        self.trusted_set.timestamp.signed.meta["snapshot.json"].hashes = true_hashes
+
```
jku: can you explain better what this tests.

If it's testing the metafile verification in TrustedMetadataSet, I think it's easier (much easier for the reader) if you just modify metafile.version

If it's testing some specific failure within MetaFile verify, then the test probably doesn't belong in TrustedMetadataSet tests.

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        snapshot_obj.signed.meta = {}
+        # prepare timestamp.meta["snapshot"].hashes
+        self._sign_modified_obj("snapshot", snapshot_obj)
+        timestamp_meta = self.trusted_set.timestamp.signed.meta["snapshot.json"]
+        true_hashes = timestamp_meta.hashes or {}
+        modified_snapshot_data = json.dumps(snapshot_obj.to_dict()).encode()
+        modified_hashes = self._calculate_modified_hashes(
+            true_hashes, modified_snapshot_data
+        )
+        self.trusted_set.timestamp.signed.meta["snapshot.json"].hashes = modified_hashes
```
jku: this is quite difficult to understand. Any chance of simplifying (e.g. not using hashes)?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+
+        with self.assertRaises(exceptions.RepositoryError):
+            self.trusted_set.update_snapshot(modified_snapshot_data)
+
+    def test_update_snapshot_after_succesfull_update_new_snapshot_meta_version_different(self):
+        self._setup_update_snapshot_after_successful_update_test()
+        # snapshot.meta["project1"].version != new_snapshot.meta["project1"].version
+        for meta_file_path in self.trusted_set.snapshot.signed.meta.keys():
+            self.trusted_set.snapshot.signed.meta[meta_file_path].version = 2
```
jku:
```suggestion
for metafile in self.trusted_set.snapshot.signed.meta.values():
    metafile.version += 1
```
or something like it?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        with self.assertRaises(exceptions.BadVersionNumberError):
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
+
+    def test_update_snapshot_after_succesfull_expired_new_snapshot(self):
+        self._setup_update_snapshot_after_successful_update_test()
+        # new_snapshot has expired
+        snapshot_obj = Metadata.from_bytes(self.metadata["snapshot"])
+        snapshot_obj.signed.expires = datetime.strptime(
+            "1970-01-01T00:00:00Z", "%Y-%m-%dT%H:%M:%SZ"
+        )
```
jku: unneeded parsing

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+            true_hashes, modified_snapshot_data
+        )
+        self.trusted_set.timestamp.signed.meta["snapshot.json"].hashes = modified_hashes
+        # remove length so it doesn't intervene the validation.
+        self.trusted_set.timestamp.signed.meta["snapshot.json"].length = None
+        with self.assertRaises(exceptions.ExpiredMetadataError):
+            self.trusted_set.update_snapshot(modified_snapshot_data)
+
+
+    def setup_update_targets_test(self):
```
jku: * other setup function names start with an underscore
* Maybe the setup functions could also all be close to each other so it's easier to see the differences?

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+
+    def test_update_targets_hash_different_than_snapshot_meta_hash(self):
+        self.setup_update_targets_test()
+        # observed_hash != stored hash in snapshot meta for targets
+        true_hashes = {}
+        for target_path, meta_file in self.trusted_set.snapshot.signed.meta.items():
+            true_hashes[target_path] = meta_file.hashes
+            self.trusted_set.snapshot.signed.meta[target_path].hashes = {"sha256": "b"}
+        with self.assertRaises(exceptions.RepositoryError):
+            self.trusted_set.update_targets(self.metadata["targets"])
```
jku: true_hashes is unused

Also I think mentioning hashes in the test name is maybe not what we want: TrustedMetadataSet is not interested in whether the issue is hashes, version or length: we're testing that TrustedMetadataSet functions correctly when MetaFile verification fails (for any reason)

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+
+    def test_update_targets_version_different_snapshot_meta_version(self):
```
jku, line 348: I think this is the same test as above (see the comment there).

I _think_ the exception is now the same as well, after some of of Teodoras PRs...

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
...
+        with self.assertRaises(exceptions.BadVersionNumberError):
+            self.trusted_set.update_targets(self.metadata["targets"])
+
+    def test_update_targets_expired_new_target(self):
+        self.setup_update_targets_test()
+        # new_delegated_target has expired
+        targets_obj = Metadata.from_bytes(self.metadata["targets"])
+        targets_obj.signed.expires = datetime.strptime(
+            "1970-01-01T00:00:00Z", "%Y-%m-%dT%H:%M:%SZ"
+        )
```
jku: unneeded parsing

tests/test_trusted_metadata_set.py
```diff
@@ -110,13 +128,269 @@ def test_update_with_invalid_json(self):
 
             # metadata is of wrong type
             with self.assertRaises(exceptions.RepositoryError):
-                update_func(data["root"])
+                update_func(self.metadata["root"])
```
jku, line 213: Based on the tests you wrote I guess this does not actually test what I thought it would -- for snapshot and targets this is going to fail in metafile verification already, not because the type is wrong ... feel free to remove this if you have the root/timestamp cases covered elsewhere

jku (changes requested):
Sooo many comments, sorry about that... but it is a big patch and again the first of it's kind.

I like the smaller function, much easier to read.

many tests finish by modifying some trusted_set metadata internals: that should never be needed, or am I missing something?

There's a lot of hash calculating that I think isn't needed: MetaFile hashes are optional, you can just remove the hashes if they are there.

Many tests follow this pattern:
* update TrustedMetadataSet using valid metadata
* modify TrustedMetadataSet internal metadata
* do the test

In some cases that's what you have to do... but in a lot of cases I think it might be beneficial if we could avoid modifying internal state and instead could do:
* modify metadata before feeding it to TrustedMetadataSet
* Update TrustedMetadataSet with modified metadata (triggering the test)

It's possible that this change is not worth the trouble or should be in another PR... but let's consider at least

tests/test_trusted_metadata_set.py
MVrachev: I added this test because it was not tested by `test_update_with_invalid_json()`.
At least that's what coverage missing lines tells me.
Will investigate what's the reason.

MVrachev: The reason was that we have forgotten to add this test there.
Will add it there with the rest of the tests testing for invalid root.

tests/test_trusted_metadata_set.py
MVrachev: I moved this test inside `test_out_of_order_ops` and removed this test function.
It was not added initially in `test_out_of_order_ops`.

tests/test_trusted_metadata_set.py
MVrachev: This won't work only for snapshot where there are hashes and length verification.
I will add commit to handing this case.

tests/test_trusted_metadata_set.py
MVrachev: I think we don't have `to_bytes()` for the rest of the classes as well?
Maybe we can add `Metadata.to_bytes()`?

tests/test_trusted_metadata_set.py
MVrachev: Yes, I used what you suggested - assigning hashes and length to None and removed this method.

tests/test_trusted_metadata_set.py
MVrachev: I moved this test into `test_out_of_order_ops`.
It's different and easier to understand now.

tests/test_trusted_metadata_set.py
MVrachev: Yea, I simplified it without calculating hashes.

tests/test_trusted_metadata_set.py
MVrachev: Yes, I will group them together just after `setUp()`.

tests/test_trusted_metadata_set.py
MVrachev: Yes, but I want to be specific about what's wrong with this particular test because we have around 15 of them.

tests/test_trusted_metadata_set.py
MVrachev: From what I understand, they are different. There is a separate exception raised for bad versions.

MVrachev:
I addressed your comments @jku.

tests/test_trusted_metadata_set.py
jku: oh right, that's what I meant (I tried to say `timestamp.to_bytes()` in the example meaning `Metadata.to_bytes()`)

tests/test_trusted_metadata_set.py
jku: you are right: I mistook these for testing length and hashes, but it's version and hashes (and version does get tested much later in the process)

tests/test_trusted_metadata_set.py
```diff
@@ -1,122 +1,330 @@
...
+from typing import Dict, Any
+from datetime import datetime
 
 from tuf import exceptions
 from tuf.api.metadata import Metadata
-from tuf.ngclient._internal.trusted_metadata_set import TrustedMetadataSet
+from tuf.ngclient._internal.trusted_metadata_set import(
+    TrustedMetadataSet
+)
+from securesystemslib import hash as sslib_hash
```
jku: unused?

tests/test_trusted_metadata_set.py
```diff
@@ -1,122 +1,330 @@
...
+        self.trusted_set.update_timestamp(self.metadata["timestamp"])
 
+    def _setup_update_snapshot_after_successful_update_test(self):
+        self._setup_update_snapshot_or_timestamp_test()
+        self.trusted_set.update_snapshot(self.metadata["snapshot"])
+
+    def _setup_update_targets_test(self):
+        self.trusted_set.root_update_finished()
+        self.trusted_set.update_timestamp(self.metadata["timestamp"])
+        self.trusted_set.update_snapshot(self.metadata["snapshot"])
```
jku: now that I see them next to each other, these two look like the same function?

It might be more useful if the functions were named for what they do, not what is expected to happen after the function.

tests/test_trusted_metadata_set.py
```diff
@@ -1,122 +1,330 @@
...
-                data[md] = f.read()
+        # Update snapshot after sucessful targets update
+        with self.assertRaises(RuntimeError):
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
+
+        self.trusted_set.update_delegated_targets(
+            self.metadata["role1"], "role1", "targets"
+        )
+
+        self.trusted_set.update_targets(self.metadata["targets"])
```
jku: wait, what is this line trying to test for? 

How does it succeed, is this a TrustedMetadataSet bug?

tests/test_trusted_metadata_set.py
```diff
@@ -1,122 +1,330 @@
...
+        targets_obj = Metadata.from_bytes(self.metadata["targets"])
+        self.trusted_set._trusted_set["targets"] = targets_obj
+        with self.assertRaises(RuntimeError):
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
+
+    def test_update_snapshot_cannot_verify_snapshot_with_threshold(self):
+        self._setup_update_snapshot_or_timestamp_test()
+        # remove signature for snapshot from root data
+        self.trusted_set.root.signed.roles["snapshot"].keyids = []
+        # self.trusted_set.snapshot.signatures = {}
```
jku: what is going on here?

jku (commented):
I'm seeing several smaller issues in the code that I've commented on and discussion was marked resolved. I'm not sure how to interpret those. I guess I'll unresolve. Please try to comment in the discussions if you don't think fixes are needed.

Do you have an opinion on the pattern I mentioned (modifying TrustedMetadataSet internals during the test): are there solutions to avoid that, what would they look like, do we want to try them or leave for later?

tests/test_trusted_metadata_set.py
jku: > true_hashes is unused

tests/test_trusted_metadata_set.py
jku: what was the resolution here?

tests/test_trusted_metadata_set.py
jku: what was the resolution here:  is this function needed or not?

tests/test_trusted_metadata_set.py
jku: I think this is just an unneeded call...

The call succeeds (even though delegated target role1 has already been loaded) against what's said in the TrustedMetadataSet docs _"Metadata is not updatable if any metadata depending on it has been loaded"_. We ended up with a version that does not explicitly check that I think because the meta info in snapshot should ensure that the targets metadata is consistent even if a delegating metadata gets "reloaded". That said, there is now a possibility of a repository doing weird things like publishing a new "targets.json" with same version but modified delegations... maybe it's not something we need to care about, but at least the TrustedMetadataSet docs should be updated  (not in this PR)

jku: actually the quote from the docs is for top-level metadata so it is technically correct... Might make sense to mention that delegating targets consistency comes from snapshots meta information.

MVrachev: > That said, there is now a possibility of a repository doing weird things like publishing a new "targets.json" with same version but modified delegations... maybe it's not something we need to care about, but at least the TrustedMetadataSet docs should be updated (not in this PR)

I guess we don't care about that because if the file is signed from the keys trusted by root.json then everything is fine right? 

We haven't documented which roles are used for verification when calling all other `update_*` functions.
Why do we want to add this specifically here? 

PS: I will remove the redundant two updates of `update_targets` and `update_delegated_targets`.

tests/test_trusted_metadata_set.py
MVrachev: Not sure how this line slipped. It should have happened during a couple of releases I had to do.
Thanks for catching this!

tests/test_trusted_metadata_set.py
MVrachev: Thanks, I missed that.

MVrachev:
> I'm seeing several smaller issues in the code that I've commented on and discussion was marked resolved. I'm not sure how to interpret those. I guess I'll unresolve. Please try to comment in the discussions if you don't think fixes are needed.

I am sorry that I closed discussions that weren't addressed.
I rebased a couple of times locally and I was distracted.

> Do you have an opinion on the pattern I mentioned (modifying TrustedMetadataSet internals during the test): are there solutions to avoid that, what would they look like, do we want to try them or leave for later?

I have to experiment with this and will do that on Monday.
My only concern is that it could be more complicated as most of the time a test depends on the state of other metadata classes inside the set. 

So, for example I want to check if `new_snapshot.version != trusted timestamp.meta["snapshot"].version` will fail.
I will have to do the following operations:

To do it the way you propose I will have to:
1.  instantiate a timestamp object from bytes
2. change the timestamp_obj.meta["snapshot"].version
3. convert timestamp_obj to bytes again
4. call `update_timestamp`
5. perform the test

Compared to now where I do:
1. call `update_timestamp`
2. change the internal `timestamp.signed.meta["snapshot.json"].version`
3. perform the test.

PS: I have submitted https://github.com/theupdateframework/tuf/pull/1490 which I hope will simplify the code a little.

tests/test_trusted_metadata_set.py
```diff
@@ -1,122 +1,319 @@
...
+            self.trusted_set.update_timestamp(new_timestamp_byte_data)
+
+
+    def test_update_snapshot_after_targets_updated(self):
+        self._root_update_finished_and_update_timestamp()
+        # cannot update snapshot after targets update completes or targets != None
+        targets_obj = Metadata.from_bytes(self.metadata["targets"])
+        self.trusted_set._trusted_set["targets"] = targets_obj
+        with self.assertRaises(RuntimeError):
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
```
jku: I think you've added this test in test_out_of_order_ops() already?

(without any need to modify internal state, I might add)

tests/test_trusted_metadata_set.py
```diff
@@ -1,122 +1,319 @@
...
+
+    def test_update_snapshot_cannot_verify_snapshot_with_threshold(self):
+        self._root_update_finished_and_update_timestamp()
+        # remove signature for snapshot from root data
+        self.trusted_set.root.signed.roles["snapshot"].keyids = []
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            self.trusted_set.update_snapshot(self.metadata["snapshot"])
+        self.trusted_set.root.signed.roles["snapshot"].threshold = 1
```
jku: The last assignment seems unrelated.

nitpick: comment is still weird (talks about signature but removes keys from snapshot role).

tests/test_trusted_metadata_set.py
```diff
@@ -1,122 +1,319 @@
...
+            self.trusted_set.update_root(self.metadata["root"])
+
+    def _sign_modified_obj(
+        self,
+        role:str,
+        metadata_obj: Metadata
+    ) -> Dict[str, Any]:
+        sslib_signer = self.keystore[role]
+        signature = metadata_obj.sign(sslib_signer)
+        return signature.to_dict()
```
jku: unused return value.

is this function valuable? it seems replaceable by  e.g. `root_obj.sign(self.keystore["root"])` (example from a few lines below)

jku (commented):
> To do it the way you propose I will have to:
> 
> 1. instantiate a timestamp object from bytes
> 2. change the timestamp_obj.meta["snapshot"].version
> 3. convert timestamp_obj to bytes again
> 4. call `update_timestamp`
> 5. perform the test

You make it sound like this is  a lot more complex but I'm not sure it is 

For example, the current test: 

```python
    def test_update_snapshot_cannot_verify_snapshot_with_threshold(self):
        self._root_update_finished_and_update_timestamp()
        # remove signature for snapshot from root data
        self.trusted_set.root.signed.roles["snapshot"].keyids = []
            self.trusted_set.update_snapshot(self.metadata["snapshot"])
        self.trusted_set.root.signed.roles["snapshot"].threshold = 1
```

could be something like 

```python
    def test_update_snapshot_cannot_verify_snapshot_with_threshold(self):
        self._root_update_finished_and_update_timestamp()

        md = Metadata.from_bytes(self.metadata["snapshot"])
        md.signatures.clear()

            self.trusted_set.update_snapshot(JSONSerializer().serialize(md))
```
That does not look complex in comparison.

Sure, in some cases you need to sign the resulting metadata as well: I'm sure there could be a helper function that makes the whole edit process reasonable for individual tests.

I'm **really** not saying we need to do this now though -- this PR is big enough, feel free to leave it as is to move things along. 

I am saying that as general rule, tests are better when they do not modify internal state: we will likely keep refactoring the implementation and then slowly we will have less trust on the tests (as they rely on the internals that have now changed)

tests/test_trusted_metadata_set.py
MVrachev: Yes, you are right.
Now when we have assigned SSLib instances to `self.keystore` it makes sense to remove this helper function.

tests/test_trusted_metadata_set.py
MVrachev: Ohh.. I am not sure how I forgot to remove it.
Thanks!

MVrachev:
@jku I decide to experiment with updating the tests so that we don't change the internal state of the trusted metadata set.

I was wondering do I want to call `self._root_updated_and_update_timestamp(self._modify_timestamp_meta(version=3))`
directly or instead do it like I did in the tests
```
modified_timestamp = self._modify_timestamp_meta(version=3)
self._root_updated_and_update_timestamp(modified_timestamp)
```

What do you think of the way I implemented it? 
Does the new helper function look understandable?

MVrachev:
As suggested by @jku, in my last commit I removed the `modify_timestamp` and `modify_snapshot` functions and instead defined `modify_metadata(rolename: str, modification_func: Callable[["Signed"], None])` function which does the common thinks
like:
- instantiating a metadata object
- calling the modification function
- signing the modified object
- serializing back to bytes

jku:
I think that's a pretty good solution with the constraints you have... It's not immediately obvious how modify_metadata() works but reader can figure it out and it makes test code very readable.

The one suggestion I still have is about hashes and length in timestamp: I understand you don't want to modify the test json file but you could just set hashes and length to None in setUpClass() once. That would simplify code in 9 different tests.

I'll approve this as is, you decide.

jku (approved):
Approved, see comment above.

MVrachev:
@jku I applied your suggestion in my latest commit and additionally noticed there are more places where I can simplify the code or add annotations.

PR #1480, opened 2021-07-05
Metadata API: Fix keyval "public" requirement

MVrachev:

Currently, we require that the keyval attribute in the Key class
is a dictionary and has "public" as a key, otherwise, we throw
KeyError or ValueError.

This requirement is too strict given that in the spec for KEYVAL it's
only said that KEYVAL is:
"A dictionary containing the public portion of the key."
See: https://theupdateframework.github.io/specification/latest/index.html#keyva

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -424,8 +424,9 @@ class Key:
         keyval: Dict[str, str],
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
-        val = keyval["public"]
-        if not all(isinstance(at, str) for at in [keyid, keytype, scheme, val]):
+        if not all(
+            isinstance(at, str) for at in [keyid, keytype, scheme]
+        ) or not isinstance(keyval, Dict):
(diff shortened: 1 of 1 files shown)


MVrachev:
There is one additional place where we relly that `keyval` includes a `public` key in: 
https://github.com/theupdateframework/tuf/blob/745a8f7680c69ea290d48bc5a2066dd4ad9781d3/tuf/ngclient/_internal/trusted_metadata_set.py#L110

any ideas on how to fix this one?
I am not sure how can we retrieve the public portion of the key if there is no `public` key.

jku:
> any ideas on how to fix this one?
> I am not sure how can we retrieve the public portion of the key if there is no `public` key.

The value of using public key there was debatable in the first place so I remove it in #1436 (see https://github.com/theupdateframework/tuf/pull/1436#issuecomment-865688070). The ngclient code will be removed after this PR is in

jku (approved):
lgtm, thanks

PR #1481, opened 2021-07-05
Metadata API: preserve Role.keyids order

### LINKED ISSUE #1478: Metadata API: Role keyids is not ordered
We made Role.keyids a set because the keyids are supposed to be unique -- this still makes sense

However, the data should also preserve order (when deserialized and serialized) and currently it does not. This is fairly serious since writing signed data potentially modifies the data (making the signature invalid).

* We should have tests that deserialize/serialize metadata with more than one item in each container (like roles keyids) to notice issues like this -- this should be easy now that the serialization tests exist
* We should fix the issue with Role.keyids -- my assumption is that the issue is Set not being ordered but I'm not 100% on that

MVrachev:
Fixes #1478 

We made Role.keyids a set because the keyids are supposed
to be unique and this still makes sense.

However, the data should also preserve order
(when deserialized and serialized) and currently, it does not.
This is fairly serious since writing signed data potentially modifies
the data (making the signature invalid).

The simplest solution (as proposed by Jussi) is to make keyids
a property and there to enforce keyids uniqueness and preserve the
order by using a list instead of a set.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -554,7 +554,7 @@ class Role:
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dictionary representation of self."""
         return {
-            "keyids": list(self.keyids),
+            "keyids": sorted(self.keyids),
             "threshold": self.threshold,
             **self.unrecognized_fields,
         }
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -141,6 +141,7 @@ class TestSerialization(unittest.TestCase):
 
     valid_roles: DataSet = {
         "all": '{"keyids": ["keyid"], "threshold": 3}',
+        "many keyids": '{"keyids": ["a", "b", "c", "d", "e"], "threshold": 1}',
         "unrecognized field": '{"keyids": ["keyid"], "threshold": 3, "foo": "bar"}',
     }
 


tuf/api/metadata.py
```diff
@@ -626,15 +637,18 @@ def to_dict(self) -> Dict[str, Any]:
     # Update key for a role.
     def add_key(self, role: str, key: Key) -> None:
         """Adds new signing key for delegated role 'role'."""
-        self.roles[role].keyids.add(key.keyid)
-        self.keys[key.keyid] = key
+        if key.keyid not in self.roles[role].keyids:
+            self.roles[role].keyids.append(key.keyid)
+            self.keys[key.keyid] = key
```
jku: this changes what happens when keyid is in keyids already (keys does not get updated)

tuf/api/metadata.py
```diff
@@ -626,15 +637,18 @@ def to_dict(self) -> Dict[str, Any]:
...
 
     def remove_key(self, role: str, keyid: str) -> None:
         """Removes key from 'role' and updates the key store.
 
         Raises:
             KeyError: If 'role' does not include the key
         """
+        if keyid not in self.roles[role].keyids:
+            raise KeyError(f"Role.keyid doesn't include {keyid}!")
```
jku: I guess keyids.remove() doesn't raise KeyError now? 

I think we could just change the API doc instead so whatever is raised is documented

tuf/api/metadata.py
```diff
@@ -524,21 +524,32 @@ class Role:
...
+    # Preserve keyids order and enforce kids uniqueness.
+    @keyids.setter
+    def keyids(self, input_keyids: List[str]) -> None:
+        res_keyids = []
+        for keyid in input_keyids:
+            if keyid in res_keyids:
+                raise ValueError(
+                    f"keyids should be a list of unique strings,"
+                    f" instead got {input_keyids}"
+                )
```
jku: has to be a way to squeeze this into something less than four lines
```suggestion
                raise ValueError(f"Non-unique keyids: {input_keyids}"
```

jku:
@MVrachev can you please comment on Teodoras suggestion (comment was made in the issue)? 

I think the idea was that we could keep using the set if we make sure that on serialization we always sort the the output. This has some corner cases as well (like if your using the API to process keyids somehow, the order isn't guaranteed) but if that's the only issue then maybe that's a better way?

MVrachev:
Yes, I started answering, but it seems I haven't sent my response. 
It sounds reasonable and it won't add complexity with the additional property and checks that have to be made in my current suggestions.
I will update the pr with those suggestions.

MVrachev:
Yes, it looks a lot better.
Great suggestion @sechkova!

jku (approved):
I'm trying to think if there's any case where in the API we would need the ordered keyids... but I don't think we ever do. So this looks right to me, thanks!

PR #1486, opened 2021-07-07
ADR 8: change "Decision outcome"

MVrachev:

After a discussion with Jussi, we realized that there are a couple of
places where we don't want to allow unrecognized fields because the
they are sensitive dictionaries and the specification requires an items
of certain types inside them.
The places where we don't want to allow unrecognized fields are
"keys", "roles", "meta", "hashes" or "targets".

This is pr is updating ADR8 with the decision we already take in #1466.

### DIFF
--- a/docs/adr/0008-accept-unrecognised-fields.md
+++ b/docs/adr/0008-accept-unrecognised-fields.md
@@ -33,15 +33,18 @@ intermediate operations:
 
 then, the checksum (the content) of the file must not be changed.
 - Flexibility to add new fields in the spec without adding breaking changes.
+- Don't store unrecognized fields when it is not allowed by the specification.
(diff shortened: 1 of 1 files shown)


MVrachev:
@jku can you review that pr given that we have discussed unrecognized fields support with you in a couple of prs and issues.

jku (approved):
content looks fine to me. I don't know how these ADRs are supposed to work, do we just change it like this?

jku:
:shrug: well I merged it

joshuagl:
I think that was appropriate. The goal of ADRs is to provide design documentation and make implicit assumptions explicit, so updating ADRs to capture nuance like this seems sensible to me.

Issue #1411, opened 2021-05-21
Metadata API: Consider including filename in TargetFile/MetaFile


jku:
Currently TargetFile and Metafile do not contain the filename of the file they represent. The API itself does not need it but it could be useful for users of the api.

As an example the current client returns a dict for get_one_valid_targetinfo(): that dict contains a filepath field and and a targetinfo field (essentially TargetFile). I would like to keep a similar API but avoid hand-crafted dicts. It would be much nicer to return a TargetFile that would contain the full "metadata" of the targetfile.

I'm not 100% sure if there are similar advantages to MetaFile

jku:
relates to #1317

MVrachev:
I think it makes sense to implement this for the use case that we are sure about - add a path in `TargetFile` and then
if one day we see a benefit for adding the name in `MetaFile` then we should add it.
Do you agree @jku?
I can propose a pr for this.

MVrachev:
I am working on this one already.

MVrachev:
I decided to propose a pr adding target name in `TargetFile`, so that the use case we are sure about is addressed and after that, we can discuss if we want to do the same thing for `MetaFile`.

MVrachev:
I had a look in our code at how we are using `MetaFile` and I can't find a place where adding `filename` to the `MetaFile` class will make the code cleaner or easier to understand.
I think we can close this issue for now and if we see the need in the future, then we can reopen this issue or open another one.
What do you think @jku?

jku:
Sounds good.

PR #1489, opened 2021-07-09
Enable mypy for ngclient

### LINKED ISSUE #1409: ngclient: type annotations review and mypy integration
ngclient type annotations are not complete: once mypy is in use (#1395), we should
* enable mypy for ngclient
* fix any annotation issues

sechkova:
Fixes #1409 

Enable static type checking for ngclient.
- add all missing type annotations
- fix errors (partially for updater.py and trusted_metadata_set.py
- add urllib3 to ignored includes
- download suggested stub packages when running mypy

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -24,8 +24,8 @@ import tempfile
 from collections import OrderedDict
 from datetime import datetime, timedelta
 from typing import (
+    IO,
     Any,
-    BinaryIO,
     ClassVar,
     Dict,
     Generic,
@@ -753,7 +753,7 @@ class BaseFile:
 
     @staticmethod
     def _verify_hashes(
-        data: Union[bytes, BinaryIO], expected_hashes: Dict[str, str]
+        data: Union[bytes, IO[bytes]], expected_hashes: Dict[str, str]
     ) -> None:
         """Verifies that the hash of 'data' matches 'expected_hashes'"""
         is_bytes = isinstance(data, bytes)
@@ -782,7 +782,7 @@ class BaseFile:
 
     @staticmethod
     def _verify_length(
-        data: Union[bytes, BinaryIO], expected_length: int
+        data: Union[bytes, IO[bytes]], expected_length: int
     ) -> None:
         """Verifies that the length of 'data' matches 'expected_length'"""
         if isinstance(data, bytes):
@@ -867,7 +867,7 @@ class MetaFile(BaseFile):
 
         return res_dict
 
-    def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]) -> None:
+    def verify_length_and_hashes(self, data: Union[bytes, IO[bytes]]) -> None:
         """Verifies that the length and hashes of "data" match expected values.
 
         Args:
@@ -1182,7 +1182,7 @@ class TargetFile(BaseFile):
             **self.unrecognized_fields,
         }
 
-    def verify_length_and_hashes(self, data: Union[bytes, BinaryIO]) -> None:
+    def verify_length_and_hashes(self, data: Union[bytes, IO[bytes]]) -> None:
         """Verifies that length and hashes of "data" match expected values.
 
         Args:
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -98,7 +98,7 @@ class Metadata(Generic[T]):
         self.signatures = signatures
 
     @classmethod
-    def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata":
+    def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata[T]":
         """Creates Metadata object from its dict representation.
 
         Arguments:
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -141,7 +141,7 @@ class RequestsFetcher(FetcherInterface):
             )
 
         except urllib3.exceptions.ReadTimeoutError as e:
-            raise exceptions.SlowRetrievalError(str(e))
+            raise exceptions.SlowRetrievalError from e
 
         finally:
             response.close()
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -7,7 +7,7 @@
 
 import logging
 import time
-from typing import Iterator, Optional
+from typing import Dict, Iterator, Optional
 from urllib import parse
 
 # Imports
@@ -31,7 +31,7 @@ class RequestsFetcher(FetcherInterface):
             session per scheme+hostname combination.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         # http://docs.python-requests.org/en/master/user/advanced/#session-objects:
         #
         # "The Session object allows you to persist certain parameters across
@@ -46,7 +46,7 @@ class RequestsFetcher(FetcherInterface):
         # improve efficiency, but avoiding sharing state between different
         # hosts-scheme combinations to minimize subtle security issues.
         # Some cookies may not be HTTP-safe.
-        self._sessions = {}
+        self._sessions: Dict[str, requests.Session] = {}
 
         # Default settings
         self.socket_timeout: int = 4  # seconds
@@ -146,7 +146,7 @@ class RequestsFetcher(FetcherInterface):
         finally:
             response.close()
 
-    def _get_session(self, url):
+    def _get_session(self, url: str) -> requests.Session:
         """Returns a different customized requests.Session per schema+hostname
         combination.
         """
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -92,7 +92,7 @@ class TrustedMetadataSet(abc.Mapping):
             RepositoryError: Metadata failed to load or verify. The actual
                 error type and content will contain more details.
         """
-        self._trusted_set = {}  # type: Dict[str: Metadata]
+        self._trusted_set: Dict[str, Metadata] = {}
         self.reference_time = datetime.utcnow()
 
         # Load and validate the local root metadata. Valid initial trusted root
@@ -134,7 +134,7 @@ class TrustedMetadataSet(abc.Mapping):
         return self._trusted_set.get("targets")
 
     # Methods for updating metadata
-    def update_root(self, data: bytes):
+    def update_root(self, data: bytes) -> None:
         """Verifies and loads 'data' as new root metadata.
 
         Note that an expired intermediate root is considered valid: expiry is
@@ -175,7 +175,7 @@ class TrustedMetadataSet(abc.Mapping):
         self._trusted_set["root"] = new_root
         logger.debug("Updated root")
 
-    def update_timestamp(self, data: bytes):
+    def update_timestamp(self, data: bytes) -> None:
         """Verifies and loads 'data' as new timestamp metadata.
 
         Note that an expired intermediate timestamp is considered valid so it
@@ -237,7 +237,7 @@ class TrustedMetadataSet(abc.Mapping):
         self._trusted_set["timestamp"] = new_timestamp
         logger.debug("Updated timestamp")
 
-    def update_snapshot(self, data: bytes):
+    def update_snapshot(self, data: bytes) -> None:
         """Verifies and loads 'data' as new snapshot metadata.
 
         Note that intermediate snapshot is considered valid even if it is
@@ -314,7 +314,9 @@ class TrustedMetadataSet(abc.Mapping):
         self._trusted_set["snapshot"] = new_snapshot
         logger.debug("Updated snapshot")
 
-    def _check_final_snapshot(self):
+    def _check_final_snapshot(self) -> None:
+        """Check snapshot expiry and version before targets is updated"""
+
         if self.snapshot.signed.is_expired(self.reference_time):
             raise exceptions.ExpiredMetadataError("snapshot.json is expired")
 
@@ -328,7 +330,7 @@ class TrustedMetadataSet(abc.Mapping):
                 f"got {self.snapshot.signed.version}"
             )
 
-    def update_targets(self, data: bytes):
+    def update_targets(self, data: bytes) -> None:
         """Verifies and loads 'data' as new top-level targets metadata.
 
         Args:
@@ -342,7 +344,7 @@ class TrustedMetadataSet(abc.Mapping):
 
     def update_delegated_targets(
         self, data: bytes, role_name: str, delegator_name: str
-    ):
+    ) -> None:
         """Verifies and loads 'data' as new metadata for target 'role_name'.
 
         Args:
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -114,9 +114,9 @@ class TrustedMetadataSet(abc.Mapping):
 
     # Helper properties for top level metadata
     @property
-    def root(self) -> Optional[Metadata]:
-        """Current root Metadata or None"""
-        return self._trusted_set.get("root")
+    def root(self) -> Metadata:
+        """Current root Metadata"""
+        return self._trusted_set["root"]
 
     @property
     def timestamp(self) -> Optional[Metadata]:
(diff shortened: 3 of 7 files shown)


jku:
I think there's a few cases here:
* TrustedMetadataSet.root probably does not actually need to be Optional: Unlike other properties it's guaranteed to be a Metadata after constructor, right?
* mypy gets confused about target_base_url: I'm sure it's possible to simplify the code so mypy understands that target_base_url cannot be None
* cases where we know a e.g. self._trusted_set.snapshot is Metadata but code does not: I think this is a great place for `assert(self._trusted_set.snapshot is not None)` 
* many cases of Metadata.Signed typing problem...
  * I believe this would get fixed with #1457 and Generics
  * alternatively we could just use calls like `assert(isinstance(self._trusted_set.snapshot.signed, Snapshot))`. Ugly but works and there are surprisingly few places where this is needed (like 10-12 or so?)
  * even more alternatively we could add properties to TrustedMetadataSet that returned Root, Snapshot, Timestamp and Targets (but that makes a lot of the code a bit ugly as you usually need both the Metadata and the Signed subclass)...

First two-three can be fixed now. The last one we can discuss... In any case this is useful work as it shows how much (or little) the lack of signed typing affects us...

The download related the changes should maybe only be done after the cleanup is done in #1448 (also the SlowRetrievalError change looks a bit fishy -- I would expect it's used in the legacy code as well?)

joshuagl:
> * mypy gets confused about target_base_url: I'm sure it's possible to simplify the code so mypy understands that target_base_url cannot be None

It's debatable whether this simplifies things for humans reading the code (edit: though, apart from the variable name ugliness and possibility for confusion of `tgt_base_url`, I think it probably is simpler) but I was able to make mypy happy with the following:

```Python
        # locally scoped variable that is definitely a string, even if empty
        tgt_base_url: str = self._target_base_url or ""
        if target_base_url is not None:
            tgt_base_url = _ensure_trailing_slash(target_base_url)

        # move the check for a base URL being set to after the locally scoped
        # variable assignment logic is bottomed out.
        if not tgt_base_url:
            raise ValueError(
                "target_base_url must be set in either download_target() or "
                "constructor"
            )
```

see the [full diff here](https://github.com/joshuagl/tuf/commit/2aaec55ab303bf9b753c863df835d9be50c4a2bd).

jku:
this should also work -- the issue for mypy is two branches that seem unrelated (even if they aren't) so making the branches actually same should make the intent clear to computer...
```
        if target_base_url is None:
            if self._target_base_url is None:
                raise ValueError(
                    "target_base_url must be set in either download_target() or "
                    "constructor"
                )
            target_base_url = self._target_base_url
        else:
```

jku:
> The download related the changes should maybe only be done after the cleanup is done in #1448 (also the SlowRetrievalError change looks a bit fishy -- I would expect it's used in the legacy code as well?)

#1448 is now merged

sechkova:
Rebased on latest changes in develop and fixed all issues but the Metadata.Signed typing ones.

sechkova:
Actually `bandit` complains slightly about asserts:
```
>> Issue: [B101:assert_used] Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
   Severity: Low   Confidence: High
   Location: tuf/ngclient/updater.py:297
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html
296	
297	            assert self._trusted_set.snapshot is not None
298	            metainfo = self._trusted_set.snapshot.signed.meta[f"{role}.json"]
```

joshuagl:
> Actually `bandit` complains slightly about asserts

Right. We could reasonably silence those warnings, as we are only suggesting use of asserts to help mypy determine the expected type. But I think this further reinforces the 'code smell' of using asserts to solve this problem.

sechkova:
> > Actually `bandit` complains slightly about asserts
> 
> Right. We could reasonably silence those warnings, as we are only suggesting use of asserts to help mypy determine the expected type. But I think this further reinforces the 'code smell' of using asserts to solve this problem.

In this case we use assert as the [recommended way](https://mypy.readthedocs.io/en/stable/common_issues.html#unexpected-errors-about-none-and-or-optional-types) to deal with `None` errors. So we either silence bandit or tweak the TrustedMetadataSet public properties to not return optional values.

sechkova:
I used the part of #1457 that makes Metadata 'Generic' to resolve the type checking issues related to the 'signed' type.
Now there are no more mypy errors left and the PR should be ready for review.

tuf/api/metadata.py
```diff
@@ -730,7 +730,7 @@ class BaseFile:
 
     @staticmethod
     def _verify_hashes(
-        data: Union[bytes, BinaryIO], expected_hashes: Dict[str, str]
+        data: Union[bytes, IO[bytes]], expected_hashes: Dict[str, str]
```
joshuagl, line 756: `BinaryIO` appears to be a subclass of `IO[bytes]` with the addition of interfaces to support context managers (`with`). We are not actually using the passed `data` object as a context manager, so this looks fine.

tuf/ngclient/updater.py
```diff
@@ -111,12 +112,7 @@ def __init__(
         # Read trusted local root metadata
         data = self._load_local_metadata("root")
         self._trusted_set = trusted_metadata_set.TrustedMetadataSet(data)
-
-        if fetcher is None:
-            self._fetcher = requests_fetcher.RequestsFetcher()
-        else:
-            self._fetcher = fetcher
-
+        self._fetcher = fetcher or requests_fetcher.RequestsFetcher()
```
joshuagl, line 115: Nice 👍

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -152,7 +152,7 @@ def __len__(self) -> int:
 
     def __iter__(self) -> Iterator[Metadata]:
         """Returns iterator over all Metadata objects in TrustedMetadataSet"""
-        return iter(self._trusted_set)
+        return iter(self._trusted_set.values())
```
joshuagl: Nice catch! thanks to mypy?

tuf/ngclient/updater.py
```diff
@@ -143,7 +139,7 @@ def refresh(self) -> None:
         self._load_snapshot()
         self._load_targets("targets", "root")
 
-    def get_one_valid_targetinfo(self, target_path: str) -> Dict:
+    def get_one_valid_targetinfo(self, target_path: str) -> Dict[str, Any]:
```
joshuagl: I wonder if there is a better return type we could be using here? Rather than a dict which will only ever contain two key-value pairs where the values have different types?

I am not suggesting we change something here in this PR, just thinking out loud.

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -205,7 +205,7 @@ def update_root(self, data: bytes) -> None:
                 f"Expected 'root', got '{new_root.signed.type}'"
             )
 
-        if self.root is not None:
+        if self._trusted_set.get("root") is not None:
```
joshuagl: If the root property is guaranteed to be set after init, do we need this check?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -196,7 +196,7 @@ def update_root(self, data: bytes) -> None:
         logger.debug("Updating root")
 
         try:
-            new_root = Metadata[Root].from_bytes(data)
+            new_root: Metadata[Root] = Metadata[Root].from_bytes(data)
```
joshuagl: These type hints for the IDE are starting to look quite cumbersome. Can the IDE not recognise the type hint on the right-hand side of the assignment operator? Do we really need them?😕

joshuagl (commented):
Thanks Teodora. This is looking good. I have a minor comment on the change to the root property, and a slight pushback on the type hinting for IDE's that I would appreciate your thoughts on before we merge.

tuf/ngclient/_internal/trusted_metadata_set.py
sechkova: Yes, mypy gets the credit.

tuf/ngclient/_internal/trusted_metadata_set.py
sechkova: We need it because we call `update_root` during init too ... to validate the initial root.
Do you think we need to clarify it somehow?

tuf/ngclient/updater.py
sechkova: Yeah, I think there is a common "willingness" and a bunch of issues related to improving this part of the API.
The suggestion here seems to be the closest to yours: https://github.com/theupdateframework/tuf/issues/1411

tuf/ngclient/_internal/trusted_metadata_set.py
sechkova: Now that you made me have another look :roll_eyes: .... it turned out I wasn't able to annotate the return type of the `from_*` methods with the generic type 'T' because `from_bytes` was a static method. So I amended the last commit and I think this one now looks better and does the job: 6754470

sechkova:
Seems like the CI won't be triggered until I rebase and resolve the conflicts.

tuf/ngclient/_internal/trusted_metadata_set.py
sechkova: Since you expressed some worries about inconsistently using the 'root' property, the other solution I came up with is this  https://github.com/theupdateframework/tuf/pull/1489/commits/8c172af244c968b221bcbd5398a7a07a6c897408

jku:
> * add urllib3 to ignored includes

If you researched this can you document why? looking in urllib3 git the annotations seem to be there but is that just for some future 2.0 release?

> * download suggested stub packages when running mypy

Could we add the required stub packages to test requirements instead?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, root_data: bytes):
             RepositoryError: Metadata failed to load or verify. The actual
                 error type and content will contain more details.
         """
-        self._trusted_set = {}  # type: Dict[str: Metadata]
+        self._trusted_set = {}  # type: Dict[str, Metadata]
```
jku:
```suggestion
        self._trusted_set: Dict[str, Metadata] = {}
```
Since we're fixing my broken annotations, let's do it like everywhere else :grimacing:

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -141,7 +141,7 @@ def _chunks(
             )
 
         except urllib3.exceptions.ReadTimeoutError as e:
-            raise exceptions.SlowRetrievalError(str(e))
+            raise exceptions.DownloadError from e
```
jku: Reasons for this are good but now the documented exceptions (here and in Fetcher API) are incorrect... Hmm. Maybe you could make SlowRetrievals speed argument optional instead?

jku (commented):
I think it looks good (but obviously the last commits depend on a commit from generics PR #1457 so I'm not approving). 

This PR isn't using the whole Generics branch though: if it did then it could just remove all of the checks of type `if new_root.signed.type != "root"` since the constructor would do that already

The last commit is IMO not great since it has so much code in the try-block: it's not clear that the KeyError should only happen in `self.root`... I think the original code is better. Maybe the real alternative would be to not call `update_root()` from init but instead do the required steps in init itself -- that seems ok to me as well as I think there aren't that many steps here.

tuf/ngclient/_internal/requests_fetcher.py
jku: or figure out another solution -- I didn't spend a lot of time on that suggestion

sechkova:
> I think it looks good (but obviously the last commits depend on a commit from generics PR #1457 so I'm not approving).

Hmm ... I wouldn't say directly depending, I didn't cherry-pick the commit from the other PR but created a new one and added you as a co-author.
 
> This PR isn't using the whole Generics branch though: if it did then it could just remove all of the checks of type `if new_root.signed.type != "root"` since the constructor would do that already

This was done intentionally but I guess I should've tried to explain my intentions somewhere in written form. 

The "Generics" branch tries to achieve two things simultaneously: correct type annotations of Signed and a runtime check ensuring  content and type are really what we expect. Since there seems to be an agreement already on the use of Generics and an ongoing discussion around the runtime check part, I decided to include only the first part here, the one directly related to type annotations and mypy. This way we can leave the rest of the discussion for #1457 which will also make it simpler because it will try to solve only one problem. I hope you agree :)

sechkova:
> The last commit is IMO not great since it has so much code in the try-block: it's not clear that the KeyError should only happen in `self.root`... I think the original code is better. Maybe the real alternative would be to not call `update_root()` from init but instead do the required steps in init itself -- that seems ok to me as well as I think there aren't that many steps here.

I tried this alternative since @joshuagl expressed some concerns about not using the 'root' property consistently inside the class. The three options, none of them perfect, would be:
- don't call `update_root` from `init`, repeat some of the code (not than much repetition so could be acceptable)
- don't use the property but "get" safely from `self._trusted_set`  c205ac0 (concerns about not using consistently the root property)
- use the root property, enclose it in a try bock and catch the KeyError 8c172af (well you already described the disadvantages)

There is a forth one where we keep root optional as the rest of the properties and add asserts in the code but I tried and this means asserts at 4-5 places which seems to rule out this option.

jku:
> > I think it looks good (but obviously the last commits depend on a commit from generics PR #1457 so I'm not approving).
> 
> Hmm ... I wouldn't say directly depending, I didn't cherry-pick the commit from the other PR but created a new one and added you as a co-author.
> 
> > This PR isn't using the whole Generics branch though: if it did then it could just remove all of the checks of type `if new_root.signed.type != "root"` since the constructor would do that already
> 
> This was done intentionally but I guess I should've tried to explain my intentions somewhere in written form.

Yeah I think the generics change is major enough that we can't just hide it in a PR titled "enable mypy for ngclient" :) We can definitely split the PR if that makes it easier to swallow though.

joshuagl:
> Yeah I think the generics change is major enough that we can't just hide it in a PR titled "enable mypy for ngclient" :) We can definitely split the PR if that makes it easier to swallow though.

Ack, let's do it.

sechkova:
> > * add urllib3 to ignored includes
> 
> If you researched this can you document why? looking in urllib3 git the annotations seem to be there but is that just for some future 2.0 release?

Yes, the latest release is 1.26 which does not include the type annotations. I updated the commit description.

> > * download suggested stub packages when running mypy

> Could we add the required stub packages to test requirements instead?

Done:  d306370

tuf/ngclient/_internal/requests_fetcher.py
sechkova: Yep, optional seems to fit well: 7cbfd0e

jku:
Ok, I've made the #1457 PR into a "minimal generics" PR -- this PR could be based off of that one I believe.

sechkova:
Rebased on #1457 and I believe I have addressed all comments so waiting for another review.

tuf/ngclient/updater.py
```diff
@@ -348,6 +345,7 @@ def _load_snapshot(self) -> None:
             # Local snapshot does not exist or is invalid: update from remote
             logger.debug("Failed to load local snapshot %s", e)
 
+            assert self._trusted_set.timestamp is not None  # nosec
```
jku, line 344: this is unfortunate... I don't have any better ideas but it looks so strange in this code.

This seems to only be needed in Updater for getting meta-info -- which is also a sort of duplicate piece of code here and in `_load_targets()`... so there's probably some sort of refactoring possibility here but let's not do that in this PR.

tuf/api/metadata.py
```diff
@@ -98,7 +98,7 @@ def __init__(self, signed: T, signatures: "OrderedDict[str, Signature]"):
         self.signatures = signatures
 
     @classmethod
-    def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata":
+    def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata[T]":
```
jku, line 101: I don't object but I'm curious did this actually affect something? I expect from_dict() to be only called from Deserializer which does not know about specific signed types.

jku (approved):
Looks good, updater.py in VSCode is much nicer now. I think this is good to go... except this will definitely clash with #1514 which I was about to merge...

tuf/ngclient/_internal/trusted_metadata_set.py
jku: this looks good to me at least

tuf/api/metadata.py
sechkova: No, I did it just for consistency with the others.

sechkova:

PR #1490, opened 2021-07-09
Metadata API: add Metadata.to_bytes()

### LINKED ISSUE #1487: Metadata API: Add Metadata.to_bytes()
`Metadata.to_bytes()` is missing from the API and that is now becoming annoying when writing the tests. 
I think it makes sense to add: it'll complete the serializing counterparts to `from_bytes()/from_file()`

MVrachev:
Fixes #1487 

Metadata.to_bytes() is missing from the API and that is now becoming
annoying when writing the tests.
I think it makes sense to add:
it'll complete the serializing counterparts to from_bytes()/from_file().

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -185,6 +185,29 @@ class Metadata:
 
         return deserializer.deserialize(data)
 
+    def to_bytes(
+        self, serializer: Optional[MetadataSerializer] = None
+    ) -> bytes:
+        """Return the serialized TUF file format as bytes.
+
+        Arguments:
+            serializer: A MetadataSerializer instance that implements the
+                desired serialization format. Default is JSONSerializer.
+
+        Raises:
+            tuf.api.serialization.SerializationError:
+                The metadata object cannot be serialized.
+        """
+
+        if serializer is None:
+            # Use local scope import to avoid circular import errors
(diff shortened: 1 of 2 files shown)


joshuagl (commented):
Can we make use of the new `to_bytes` method from the `to_file` method? That avoids us duplicating logic which defines the default serialiser.

MVrachev:
I updated the pr so that `Metadata.to_file()` uses `Metadata.to_bytes()`.
Additionally, I added tests for the two cases for the `JsonSerializer` - compact and noncompact.

tuf/api/metadata.py
```diff
@@ -182,6 +182,20 @@ def from_bytes(
 
         return deserializer.deserialize(data)
 
+    def to_bytes(
+        self, serializer: Optional[MetadataSerializer] = None
+    ) -> bytes:
+        """Return the bytes representation of self."""
```
jku: Please improve this
* arguments
* raises
* I think even the return value is not yet well defined: it's not just "self as bytes", it's the serialized TUF file format as bytes)

MVrachev: I updated the function docstring.

jku (approved):
LGTM, thanks

Issue #1444, opened 2021-06-14
Improve testing using generated metadata


jku:
This is an umbrella issue for a plan I have:
* Improve Metadata API for the case of Metadata generation (in other words, creating Metadata not deserializing it)
* Make sure all of it can be done in-memory (so writing things to disk is not needed)
* Improve testing for Metadata API itself by addding tests with generated data (see e.g. #1436: testing signature validation with various thresholds)
* Improve testing for ngclient/MetadataBundle by addding tests with generated data

The reason  I want to do this is that including metadata files in the test data seems hard to scale: We should aim to test de/serialization with test data included with the test (as in #1391), and then generate the test data for the cases where this is impractical (like ngclient MetadataBundle tests).

In the end I believe the improvements will be beneficial for the future repository tools as well, but that's a longer term goal: the above list is something where we could have results in a few weeks.

For background below is what I came up with for generating full metadata for a repository and modifying it just a bit using the current Metadata API (it currently writes to files but just for debugging purposes: `to_file()` calls can be removed). There's a lot of boilerplate but also quite a few low hanging fruit that would improve the experience:

```python
from collections import OrderedDict
from datetime import datetime
from typing import Tuple

from securesystemslib.keys import generate_ed25519_key
from securesystemslib.signer import SSlibSigner, Signer
from tuf.api.metadata import DelegatedRole, Delegations, Key, MetaFile, Metadata, Role, Root, Snapshot, TargetFile, Targets, Timestamp

# todo should come from metadata
SPEC_VERSION = "1.0.999"

def generate_test_key() -> Tuple[Key, Signer]:
    keydict = generate_ed25519_key()
    # TODO maybe this should be Key.from_securesystemslib_key()
    key = Key(keydict["keyid"], keydict["keytype"], keydict["scheme"], {"public": keydict["keyval"]["public"]})
    signer = SSlibSigner(keydict)

    return (key, signer)

# generate keys
delegate_key, delegate_signer = generate_test_key()
targets_key, targets_signer = generate_test_key()
snapshot_key, snapshot_signer = generate_test_key()
timestamp_key, timestamp_signer = generate_test_key()
root_key, root_signer = generate_test_key()


# Create delegated targets
# TODO Metadata.new_targets_with_defaults(expires: datetime) ?
# TODO Add TargetFile.from_data(data: Union[bytes, BinaryIO]) so this works:
#     delegate = Metadata.new_targets_with_defaults(datetime(3000,1,1))
[...]

PR #1501, opened 2021-07-19
ng client: support for prefix_targets_with_hash when downloading targets

### LINKED ISSUE #1479: ngclient: support HASH.FILENAME.EXT for downloading targets
Specification:
> If consistent snapshots are not used (see § 6.2 Consistent snapshots), then the filename used to download the target file is of the fixed form FILENAME.EXT (e.g., foobar.tar.gz). Otherwise, the filename is of the form HASH.FILENAME.EXT (e.g., c14aeb4ac9f4a8fc0d83d12482b9197452f6adf3eb710e3b1e2b79e8d14cb681.foobar.tar.gz), where HASH is one of the hashes of the targets file listed in the targets metadata file found earlier in step § 5.6 Update the targets role. In either case, the client MUST write the file to non-volatile storage as FILENAME.EXT.

ngclient currently does not insert the hash when consistent_snapshot is true, but it should.

Additionally we should offer a way to not do hash prefixing (the use case is warehouse where urls are already unique and include a hash)

MVrachev:
Fixes #1479

ng client: support for prefix_targets_with_hash

Add support for prefixing targets with their hashes when downloading or
using HASH.FILENAME.EXT as target names.
The introduction of prefix_targets_with_hash was necessary, because
there are use cases like Warehouse where you could use
consistent_snapshot, but without adding a hash prefix to your targets.

When prefix_targets_with_hash is set to True, target files conforming
the format HASH.FILENAME.EXT will be downloaded from the server, but
they will be saved on the client side without their hash prefixes or
FILENAME.EXT.
This makes sure the client won't understand the usage of
prefix_targets_with_hash.

Still, if you want to use HASH.FILENAME.EXT as target names when
downloading, then additionally you need to provide consistent_snapshot
set to True in your root.json. The reason is that the specification uses
consistent_snapshot for the same purpose:
```
If consistent snapshots are not used (see § 6.2 Consistent snapshots),
then the filename used to download the target file is of the fixed form
FILENAME.EXT (e.g., foobar.tar.gz). Otherwise, the filename is of the
form HASH.FILENAME.EXT
(e.g., c14aeb4ac9f4a8fc0d83d12482b9197452f6adf3eb710e3b1e2b79e8d14cb681.foobar.tar.gz),
where HASH is one of the hashes of the targets file listed in the
targets metadata file found earlier in step § 5.6 Update the targets role.
In either case, the client MUST write the file to non-volatile
storage as FILENAME.EXT.
```

See chapter 5.7.3:
https://theupdateframework.github.io/specification/latest/index.html#fetch-target
By default, prefix_targets_with_hash is set to true to make it easier
to the user to provide uniquely identifiable targets file names by
using consistent_snapshot set to True.

The same behavior of using two flags is used in the legacy code when
calling tuf.client.updater.download_target() in a repository using
prefix_targets_with_hash and consistent_snapshot:
https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py#L1302

requirements**:

### DIFF
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -15,7 +15,10 @@ import unittest
 import tuf.unittest_toolbox as unittest_toolbox
 
 from tests import utils
+from tuf.api.metadata import Metadata
 from tuf import ngclient
+from securesystemslib.signer import SSlibSigner
+from securesystemslib.interface import import_rsa_privatekey_from_file
 
 logger = logging.getLogger(__name__)
 
@@ -94,13 +97,13 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     url_prefix = 'http://' + utils.TEST_HOST_ADDRESS + ':' \
         + str(self.server_process_handler.port) + repository_basepath
 
-    metadata_url = f"{url_prefix}/metadata/"
-    targets_url = f"{url_prefix}/targets/"
+    self.metadata_url = f"{url_prefix}/metadata/"
+    self.targets_url = f"{url_prefix}/targets/"
     # Creating a repository instance.  The test cases will use this client
     # updater to refresh metadata, fetch target files, etc.
     self.repository_updater = ngclient.Updater(self.client_directory,
-                                              metadata_url,
-                                              targets_url)
+                                              self.metadata_url,
+                                              self.targets_url)
 
   def tearDown(self):
     # We are inheriting from custom class.
@@ -109,14 +112,91 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     # Logs stdout and stderr from the sever subprocess.
     self.server_process_handler.flush_log()
 
+  def _create_consistent_target(self, targetname: str, target_hash:str) -> None:
+    """Create consistent targets copies of their non-consistent counterparts
+    inside the repository directory.
+
+    Args:
+      targetname: A string denoting the name of the target file.
+      target_hash: A string denoting the hash of the target.
+
+    """
+    consistent_target_name = f"{target_hash}.{targetname}"
+    source_path = os.path.join(self.repository_directory, "targets", targetname)
+    destination_path = os.path.join(
+      self.repository_directory, "targets", consistent_target_name
+    )
+    shutil.copy(source_path, destination_path)
+
+
+  def _make_root_file_with_consistent_snapshot_true(self) -> None:
+    """Swap the existing root file inside the client directory with a new root
+    file where the consistent_snapshot is set to true."""
+    root_path = os.path.join(self.client_directory, "root.json")
+    root = Metadata.from_file(root_path)
+    root.signed.consistent_snapshot = True
+    root_key_path = os.path.join(self.keystore_directory, "root_key")
+    root_key_dict = import_rsa_privatekey_from_file(
+            root_key_path, password="password"
+    )
+    root_signer = SSlibSigner(root_key_dict)
+    root.sign(root_signer)
+    # Remove the old root file and replace it with the newer root file.
+    os.remove(root_path)
+    root.to_file(root_path)
+
+
+  def test_refresh_on_consistent_targets(self):
+    # Generate a new root file where consistent_snapshot is set to true and
+    # replace the old root metadata file with it.
+    self._make_root_file_with_consistent_snapshot_true()
+    self.repository_updater = ngclient.Updater(self.client_directory,
+                                              self.metadata_url,
+                                              self.targets_url)
+    # All metadata is in local directory already
+    self.repository_updater.refresh()
+
+    # Get targetinfo for "file1.txt" listed in targets
+    targetinfo1 = self.repository_updater.get_one_valid_targetinfo("file1.txt")
+    # Get targetinfo for "file3.txt" listed in the delegated role1
+    targetinfo3 = self.repository_updater.get_one_valid_targetinfo("file3.txt")
+
+    # Create consistent targets with file path HASH.FILENAME.EXT
+    target1_hash = list(targetinfo1["fileinfo"].hashes.values())[0]
+    target3_hash = list(targetinfo3["fileinfo"].hashes.values())[0]
+    self._create_consistent_target("file1.txt", target1_hash)
+    self._create_consistent_target("file3.txt", target3_hash)
+
+    destination_directory = self.make_temp_directory()
+    updated_targets = self.repository_updater.updated_targets(
+      [targetinfo1, targetinfo3], destination_directory
+    )
+
+    self.assertListEqual(updated_targets, [targetinfo1, targetinfo3])
+    self.repository_updater.download_target(targetinfo1, destination_directory)
+    updated_targets = self.repository_updater.updated_targets(
+      updated_targets, destination_directory
+    )
+
+    self.assertListEqual(updated_targets, [targetinfo3])
+
+    self.repository_updater.download_target(targetinfo3, destination_directory)
+    updated_targets = self.repository_updater.updated_targets(
+      updated_targets, destination_directory
+    )
+
+    self.assertListEqual(updated_targets, [])
+
   def test_refresh(self):
+    # Test refresh without consistent targets - targets without hash prefixes.
+
     # All metadata is in local directory already
     self.repository_updater.refresh()
 
     # Get targetinfo for 'file1.txt' listed in targets
-    targetinfo1 = self.repository_updater.get_one_valid_targetinfo('file1.txt')
+    targetinfo1 = self.repository_updater.get_one_valid_targetinfo("file1.txt")
     # Get targetinfo for 'file3.txt' listed in the delegated role1
-    targetinfo3= self.repository_updater.get_one_valid_targetinfo('file3.txt')
+    targetinfo3 = self.repository_updater.get_one_valid_targetinfo("file3.txt")
 
     destination_directory = self.make_temp_directory()
     updated_targets = self.repository_updater.updated_targets([targetinfo1, targetinfo3],
@@ -146,7 +226,7 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
     self.repository_updater.refresh()
 
     # Get targetinfo for 'file3.txt' listed in the delegated role1
-    targetinfo3= self.repository_updater.get_one_valid_targetinfo('file3.txt')
+    targetinfo3 = self.repository_updater.get_one_valid_targetinfo('file3.txt')
 
 if __name__ == '__main__':
   utils.configure_test_logging(sys.argv)
--- a/tuf/ngclient/config.py
+++ b/tuf/ngclient/config.py
@@ -9,14 +9,27 @@ from dataclasses import dataclass
 
 @dataclass
 class UpdaterConfig:
+    """Used to store Updater configuration.
+
+    Arguments:
+        max_root_rotations: The maximum number of root rotations.
+        max_delegations: The maximum number of delegations.
+        root_max_length: The maxmimum length of a root metadata file.
+        timestamp_max_length: The maximum length of a timestamp metadata file.
+        snapshot_max_length: The maximum length of a snapshot metadata file.
+        targets_max_length: The maximum length of a targets metadata file.
+        prefix_targets_with_hash: When consistent snapshots are used
+            (see https://theupdateframework.github.io/specification/latest/#consistent-snapshots), #pylint: disable=line-too-long
+            target download URLs are formed by prefixing the filename with a
+            hash digest of file content by default. This can be overridden by
+            setting prefix_targets_with_hash to False.
+
+    """
+
     max_root_rotations: int = 32
     max_delegations: int = 32
     root_max_length: int = 512000  # bytes
     timestamp_max_length: int = 16384  # bytes
     snapshot_max_length: int = 2000000  # bytes
     targets_max_length: int = 5000000  # bytes
-    # We need this variable because there are use cases like Warehouse where
-    # you could use consistent_snapshot, but without adding a hash prefix.
-    # By default, prefix_targets_with_hash is set to true to use uniquely
-    # identifiable targets file names for repositories.
     prefix_targets_with_hash: bool = True
--- a/tuf/ngclient/config.py
+++ b/tuf/ngclient/config.py
@@ -15,3 +15,8 @@ class UpdaterConfig:
     timestamp_max_length: int = 16384  # bytes
     snapshot_max_length: int = 2000000  # bytes
     targets_max_length: int = 5000000  # bytes
+    # We need this variable because there are use cases like Warehouse where
+    # you could use consistent_snapshot, but without adding a hash prefix.
+    # By default, prefix_targets_with_hash is set to true to use uniquely
+    # identifiable targets file names for repositories.
+    prefix_targets_with_hash: bool = True
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -252,8 +252,12 @@ class Updater:
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        target_filepath = targetinfo["filepath"]
         target_fileinfo: "TargetFile" = targetinfo["fileinfo"]
+        target_filepath = targetinfo["filepath"]
+        consistent_snapshot = self._trusted_set.root.signed.consistent_snapshot
+        if consistent_snapshot and self.config.prefix_targets_with_hash:
+            hashes = list(target_fileinfo.hashes.values())
+            target_filepath = f"{hashes[0]}.{target_filepath}"
         full_url = parse.urljoin(target_base_url, target_filepath)
 
         with self._fetcher.download_file(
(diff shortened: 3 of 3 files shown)


MVrachev:
I will need more time investigating and when I am ready I will update this pr and make it non-draft.

tuf/ngclient/updater.py
```diff
@@ -185,8 +185,10 @@ def download_target(
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        target_filepath = targetinfo["filepath"]
         target_fileinfo: "TargetFile" = targetinfo["fileinfo"]
+        target_filepath = targetinfo["filepath"]
+        if self.config.consistent_snapshot:
+            target_filepath = target_fileinfo.hashes + "." + target_filepath
```
joshuagl: `TargetFile.hashes` is a dictionary of hash_alg:hash_digest. In the legacy client we currently only download the file prefixed with the first listed hash value in [`_get_target_file()`](https://github.com/theupdateframework/tuf/blob/69dca08b9acdf58b0fa86d7789c6c56aaac1fa78/tuf/client/updater.py#L1261).

The spec describes that we should retrieve the file as _one_ of the hash digest prefixed files (see [5.7.3](https://theupdateframework.github.io/specification/latest/#fetch-target))
> Otherwise, the filename is of the form HASH.FILENAME.EXT (e.g., c14aeb4ac9f4a8fc0d83d12482b9197452f6adf3eb710e3b1e2b79e8d14cb681.foobar.tar.gz), where HASH is one of the hashes of the targets file listed in the targets metadata file found earlier in step

but that the repository should provide copies of the file at each hash digest prefix (see [6.2.1](https://theupdateframework.github.io/specification/latest/#writing-consistent-snapshots))
> consistent target files should be written to non-volatile storage as digest.filename.ext. This means that if the referrer metadata lists N cryptographic hashes of the referred file, then there must be N identical copies of the referred file, where each file will be distinguished only by the value of the digest in its filename.

Therefore the legacy implementation would be appropriate to replicate here – get the first hash from `target_fileinfo.hashes` and prepend it to `target_filepath`.

MVrachev:
This pr is ready for review.
I have changed the way I construct the consistent target name.

More importantly, I made sure that when `consistent_snapshot` is True, in `download_target` we are
downloading target files with names conforming to the schema `HASH.FILENAME.EXT`, BUT when
they are saved on the client side they are saved with their non-consistent names or `FILENAME.EXT`.
That's what the current code does when calling `get_target_file`:
https://github.com/theupdateframework/tuf/blob/8e9677d2620c937e3a64e5420e7386f2464f26bd/tuf/client/updater.py#L3183
but the target file itself is saved with the name `destination` which is constructed without the hash prefix:
constructing destination:
https://github.com/theupdateframework/tuf/blob/8e9677d2620c937e3a64e5420e7386f2464f26bd/tuf/client/updater.py#L3161
persisting the temporary file:
https://github.com/theupdateframework/tuf/blob/8e9677d2620c937e3a64e5420e7386f2464f26bd/tuf/client/updater.py#L3186

That way the usage of `consistent_snapshot` doesn't affect the client in any way.

tuf/ngclient/updater.py
```diff
@@ -185,8 +185,12 @@ def download_target(
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        target_filepath = targetinfo["filepath"]
         target_fileinfo: "TargetFile" = targetinfo["fileinfo"]
+        target_filepath = targetinfo["filepath"]
+        if self.config.consistent_snapshot:
```
sechkova: Consistent snapshot is enabled if `self._trusted_set.root.signed.consistent_snapshot` is set to True.
Besides this, the config setting should indicate whether the repository uses hash prefixes or some other naming scheme (in a same way that `prefix_filename_with_hash` is used in the current updater)
Which means that:
1. The config value could be renamed to something similar in meaning to `prefix_filename_with_hash`
2. The condition should become:
`if self._trusted_set.root.signed.consistent_snapshot and self.config.consistent_snapshot`

tuf/ngclient/updater.py
```diff
@@ -200,6 +204,8 @@ def download_target(
                 ) from e
 
             filepath = os.path.join(destination_directory, target_filepath)
+            if self.config.consistent_snapshot:
+                filepath = os.path.join(destination_directory, orig_target_path)
```
sechkova: Storing `orig_target_path` is unnecessary, you can retrieve it from  `targetinfo["filepath"]` instead.
This way the  second `if self.config.consistent_snapshot` condition becomes redundant too.

tuf/ngclient/updater.py
```diff
@@ -185,8 +185,12 @@ def download_target(
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        target_filepath = targetinfo["filepath"]
         target_fileinfo: "TargetFile" = targetinfo["fileinfo"]
+        target_filepath = targetinfo["filepath"]
+        if self.config.consistent_snapshot:
+            orig_target_path = target_filepath
+            hashes = list(targetinfo["fileinfo"].hashes.values())
+            target_filepath = hashes[0] + "." + target_filepath
```
```suggestion
            target_filepath = f"{hashes[0]}.{target_filepath}"
```

tuf/ngclient/config.py
```diff
@@ -15,3 +15,6 @@ class UpdaterConfig:
     timestamp_max_length: int = 16384  # bytes
     snapshot_max_length: int = 2000000  # bytes
     targets_max_length: int = 5000000  # bytes
+    # By default, consistent_snapshot is set to true to provide uniquely
+    # identifiable metadata and targets file names for repositories.
+    consistent_snapshot: bool = True
```
sechkova: `consistent_snapshot` is set by the repository. The client can/should choose whether targets are hash prefixed or not. See comment below.

tests/test_updater_ng.py
```diff
@@ -109,14 +110,72 @@ def tearDown(self):
...
+    inside the repository directory.
+
+    Args:
+      delegator_file: Name of the delegator metadata file.
+      targetname: Name of the target file.
+
+    """
+
+    delegator_path = os.path.join(self.client_directory, delegator_file)
+    delegator = Metadata.from_file(delegator_path)
```
sechkova: You can use `sslib_util.get_file_details` to get the hashes, no need of delegator and `Metadata` object.

sechkova (changes requested):

tuf/ngclient/updater.py
MVrachev: Good suggestion.
There are cases like PyPi and Warehouse when one could use `consistent_snapshot` without using `prefix_filename_with_hash`.
I will only use the name `prefix_targets_with_hash` because it's clearer that it's meant for targets.

tuf/ngclient/updater.py
MVrachev: Agree, the only downside is that it will require more lines of code than just storing it in `orig_target_path`, but it will make it
easier to understand - one variable less.

tests/test_updater_ng.py
MVrachev: The problem is that I don't know what hash algorithm to use when calling `sslib.utils.get_file_details()`.
The only one with that knowledge is the delegator.

tuf/ngclient/updater.py
```diff
@@ -185,8 +185,16 @@ def download_target(
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        target_filepath = targetinfo["filepath"]
         target_fileinfo: "TargetFile" = targetinfo["fileinfo"]
+        target_filepath = targetinfo["filepath"]
+        consistent_snapshot = False
+        if hasattr(self._trusted_set.root.signed, "consistent_snapshot"):
```
sechkova: Why is this check needed? "Root" initializes the attribute `consistent_snapshot` in its constrictor?

tuf/ngclient/updater.py
```diff
@@ -200,6 +208,10 @@ def download_target(
                 ) from e
 
             filepath = os.path.join(destination_directory, target_filepath)
```
```suggestion
            filepath = os.path.join(destination_directory, targetinfo["filepath"])
```

sechkova:
Please reword the commit.

tuf/ngclient/updater.py
MVrachev: Yes, you are right this check is incorrect.
I forgot that we are always setting consistent_snapshot even if it's nonexistent in the JSON file itself.

tuf/ngclient/updater.py
MVrachev: You are right with this suggestion.
That way I will always save the file without the hash prefix - exactly what we want.

MVrachev:
I squashed my commits and made the following changes:
- found out that I am not setting `consistent_snapshot` to true in any of my tests and I am not actually testing downloading target names in the format `HASH.FILENAME.EXT`. Fixed that by modifying the root bytes, creating a new trusted metadata set and changing the updater internals. Not a great solution, but I think it's better than modifying the client directory used by the rest of the tests.
- reword my commits
- simplified a couple of places thanks to @sechkova suggestions.

tests/test_updater_ng.py
sechkova: It is not a problem, since you have all the data.

tests/test_updater_ng.py
```diff
@@ -109,14 +113,96 @@ def tearDown(self):
...
+    root_key_dict = import_rsa_privatekey_from_file(
+            root_key_path, password="password"
+    )
+    root_signer = SSlibSigner(root_key_dict)
+    root.sign(root_signer)
+    return root.to_bytes()
+
+
+  def test_refresh_on_consistent_targets(self):
+    # Test refresh with consistent snapshot.
```
sechkova: "on consistent targets", "with consistent snapshot", pick one

tests/test_updater_ng.py
```diff
@@ -109,14 +113,96 @@ def tearDown(self):
...
+
+
+  def _root_make_consistent_snapshot_true(self) -> bytes:
+    """Instantiate a root Metadata instanec, change roots consistent_snapshot to
+    true, self sign the changed root and then serialize the modified root back
+    to bytes."""
+    root_path = os.path.join(self.repository_directory, "metadata", "root.json")
+    root = Metadata.from_file(root_path)
+    root.signed.consistent_snapshot = True
+    root_key_path = os.path.join(self.keystore_directory, "root" + '_key')
```
sechkova: "root" + '_key' = "root_key"

tests/test_updater_ng.py
```diff
@@ -109,14 +113,96 @@ def tearDown(self):
...
+
+    self.repository_updater.download_target(targetinfo3, destination_directory)
+    updated_targets = self.repository_updater.updated_targets(updated_targets,
+                                                  destination_directory)
+
+    self.assertListEqual(updated_targets, [])
+
   def test_refresh(self):
+    # Test refresh without consistent snapshot.
+    self.repository_updater.config.prefix_targets_with_hash = False
```
sechkova: This config shouldn't be doing anything without consistent_snapshot being set. I'm not sure this explicit disable is needed.

tests/test_updater_ng.py
MVrachev: Yep, you are right that it's better to load the data into a dict instead of instantiating an object and doing all the things
around it.

MVrachev:
Addressed the new comments made by @sechkova.

tests/test_updater_ng.py
```diff
@@ -109,14 +114,94 @@ def tearDown(self):
...
+  def _create_consistent_target(self, delegator_filename: str, targetname: str):
+    """Create consistent targets copies of their non-consistent counterparts
+    inside the repository directory.
+
+    Args:
+      delegator_filename: Name of the delegator metadata file.
+      targetname: Name of the target file.
+
+    """
+    delegator_path = os.path.join(self.client_directory, delegator_filename)
```
sechkova: You keep using this delegator to get the hashes, I keep disagreeing. We need a second opinion.

MVrachev: I am sorry it seemed that I disagree. I am just not sure how can I get information about  hashes that needs to be calculated for a particular delegation if I don't access the delegator.
First, I thought you meant I didn't need to instantiate a Metadata instance in order to solve this problem and that's what I changed, but  it seems like you meant something else.
It looks like I am missing something.
Can you share more how can I simplify this without accessing the delegator?

MVrachev:
I removed the need of accessing the delegator file by swapping the places of  `get_one_valid_targetinfo()` and `_create_consistent_target()`.
After calling `get_one_valid_targetinfo` I have the information about the target hashes and I can pass it to `_create_consistent_target()` and that's what I did in my last commit.

tests/test_updater_ng.py
```diff
@@ -109,14 +114,94 @@ def tearDown(self):
...
+  def _create_consistent_target(self, targetname: str, target_hash:str):
+    """Create consistent targets copies of their non-consistent counterparts
+    inside the repository directory.
+
+    Args:
+      targetname: A string denoting the name of the target file.
+      target_hash: A string denoting the hash of the target.
+
+    """
+    consistent_target_name = target_hash + "." + targetname
```
sechkova:
```suggestion
    consistent_target_name = f"{target_hash}.{targetname}"
```

sechkova (commented):
Looks fine to me now.

tuf/ngclient/config.py
```diff
@@ -9,9 +9,16 @@
...
     max_root_rotations: int = 32
     max_delegations: int = 32
     root_max_length: int = 512000  # bytes
     timestamp_max_length: int = 16384  # bytes
     snapshot_max_length: int = 2000000  # bytes
     targets_max_length: int = 5000000  # bytes
+    # We need this variable because there are use cases like Warehouse where
+    # you could use consistent_snapshot, but without adding a hash prefix.
+    # By default, prefix_targets_with_hash is set to true to use uniquely
+    # identifiable targets file names for repositories.
```
jku: Let's use the class docstring if this is intended to be documentation: the Attributes sections should work I think?

I think the text is not saying what exactly happens based on the attribute. Maybe something along lines of:
```
    # When consistent snapshots are used, target download URLs are formed by
    # prefixing the filename with a hash digest of file content by default.
    # This can be overridden by setting prefix_targets_with_hash to False.
```
(and "consistent snapshots" should link to the spec explanation of the concept)

tests/test_updater_ng.py
```diff
@@ -109,14 +114,94 @@ def tearDown(self):
...
+
+  def test_refresh_on_consistent_targets(self):
+
+    # Make root consistent snapshot true.
+    root_modified_bytes = self._root_make_consistent_snapshot_true()
+    # Instantiate a new TrustedMetadataSet with changed root and modify internals.
+    # This seems better than modifying the actual root file in the client directory
+    # considering that the client directory is used from the rest of the tests.
+    trusted_metadata_set = TrustedMetadataSet(root_modified_bytes)
+    self.repository_updater._trusted_set = trusted_metadata_set
```
jku: I don't like tests modifying internal details.

I'm not sure I understand the comment here: both repository dir and client repo dir are set up in `setUp()` so it should be done for each test.

jku (commented):
Code part looks good. Could still have a look at variable use but these are nitpicks: 
* why is target_fileinfo declared but then not used four lines later
* should filepath be renamed so it's clear that it's the path for _local_ file

The test code looks a bit bad in that it's adding 80 lines of fairly complex code for a simple test  that doesn't even check that the files it downloaded are the correct ones (half of the code is copy-paste as well)... but fixing this well probably needs a redesign of the whole test architecture so I'm not asking for changes here apart from the comment WRT modifying Updater internals.

MVrachev:
I updated the pr so that:
- the test doesn't modify an internal state but instead creates a new root file with `consistent_snapshot` set to True
- changed the `UpdaterConfig` docstring.

MVrachev:
> Code part looks good. Could still have a look at variable use but these are nitpicks:
> 
> * why is target_fileinfo declared but then not used four lines later
> * should filepath be renamed so it's clear that it's the path for _local_ file
> 
> The test code looks a bit bad in that it's adding 80 lines of fairly complex code for a simple test that doesn't even check that the files it downloaded are the correct ones (half of the code is copy-paste as well)... but fixing this well probably needs a redesign of the whole test architecture so I'm not asking for changes here apart from the comment WRT modifying Updater internals.

I updated the pr so that:

- the test doesn't modify an internal state but instead creates a new root file with `consistent_snapshot` set to True
- changed the `UpdaterConfig` docstring
- used `targetinfo` where you pointed out
- renamed `filepath` to `local_filepath`
- rebased on top of `develop`

Maybe you can share what do you think could be simplified and we can submit an issue or directly the following pr?

jku:
> Maybe you can share what do you think could be simplified and we can submit an issue or directly the following pr?

I (and others) have made some notes here https://github.com/theupdateframework/tuf/issues/1444 -- that one is about generated test data in general but there are comments wrt the other end of the spectrum (vast amounts of metadata that is at least mostly static), maybe no need to file new issues? We'll need to prototype some approaches to see if they make sense...

MVrachev:
Yep, I agree with you we need prototypes and more discussions.
I added a test here even though it's not beautiful in order to make sure we don't forget to test the new use cases
we are supporting when implementing new features.

jku (approved):
Thanks. The UpdaterConfig docs should still be improved but lets not stop this PR with that either... the code looks fine to me.

PR #1509, opened 2021-07-26
ngclient: use Metadata.verify_delegate

### LINKED ISSUE #1413: ngclient: Use Metadata APIs verify_with_threshold() when possible
#1306, #1377 are about somehow implementing verify_with_threshold() in the API.

ngclient should use that instead of the helper function in `trusted_metadata_set.py` (and delete the helper in that module)

sechkova:
Fixes #1413

Replace the helper verify_with_threshold() with
the newly implemented method Metadata.verify_delegate.

requirements**:

### DIFF
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -71,46 +71,10 @@ from datetime import datetime
 from typing import Dict, Iterator, Optional
 
 from tuf import exceptions
-from tuf.api.metadata import Metadata, Root, Targets
+from tuf.api.metadata import Metadata
 from tuf.api.serialization import DeserializationError
 
 logger = logging.getLogger(__name__)
-
-# This is a placeholder until ...
-# TODO issue 1306: implement this in Metadata API
-def verify_with_threshold(
-    delegator: Metadata, role_name: str, unverified: Metadata
-) -> bool:
-    """Verify 'unverified' with keys and threshold defined in delegator"""
-    role = None
-    keys = {}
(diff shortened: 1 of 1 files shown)


MVrachev (approved):
LGTM!
Love those prs where we are simplifying the codebase!

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -408,7 +360,7 @@ def update_delegated_targets(
         if self.snapshot is None:
             raise RuntimeError("Cannot load targets before snapshot")
 
-        delegator = self.get(delegator_name)
+        delegator: Optional[Metadata] = self.get(delegator_name)
```
joshuagl, line 363: Are we annotating this as Optional so that mypy can understand that `None` is a valid value for delegator? Why wasn't this required before?

joshuagl (commented):
The switch to `Metadata.verify_delegate` look great, but I'm confused as to why we have added extra typing annotations that were not required before?

tuf/ngclient/_internal/trusted_metadata_set.py
sechkova: No, mypy was not complaining about this. I did it to help the code editor figure it out. I can separate it in another commit or explain it in the commit message if it looks weird in between the other changes.

joshuagl: It does look a bit out of place in this commit, though it's not an objectionable change by itself.

joshuagl (approved):
Thanks for the clean up!

PR #1510, opened 2021-07-27
Snapshot RSA Accumulator

mnm678:
This is an early PoC

### DIFF
--- a/tests/test_repository_lib.py
+++ b/tests/test_repository_lib.py
@@ -480,15 +480,38 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     file_path = os.path.join(temporary_directory, '1.file1-snapshot.json')
     self.assertTrue(os.path.exists(file_path))
 
-    # TODO: check against the correct root value
-    self.assertEqual(root_1, 5)
-
     test_nodes = {}
     test_nodes['targets'] = tuf.formats.make_metadata_fileinfo(1, None, None)
     test_nodes['role1'] = tuf.formats.make_metadata_fileinfo(1, None, None)
     test_nodes['role2'] = tuf.formats.make_metadata_fileinfo(1, None, None)
 
     root, leaves = repo_lib._build_rsa_acc(test_nodes)
+    repo_lib._write_rsa_proofs(root, leaves, storage_backend,
+        temporary_directory, version)
+
+    # Ensure that the paths are written to the directory
+    file_path = os.path.join(temporary_directory, 'targets-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    file_path = os.path.join(temporary_directory, '2.targets-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    file_path = os.path.join(temporary_directory, 'role1-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    file_path = os.path.join(temporary_directory, '1.role1-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    file_path = os.path.join(temporary_directory, 'role2-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    file_path = os.path.join(temporary_directory, '1.role2-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    # TODO: check against the correct root value
+    self.assertEqual(root_1, 5)
+    self.assertEqual(root, 5)
+
 
 
 
--- a/tests/test_repository_lib.py
+++ b/tests/test_repository_lib.py
@@ -458,6 +458,40 @@ class TestRepositoryToolFunctions(unittest.TestCase):
                                          False, use_existing_fileinfo=True)
 
 
+  def test_build_rsa_acc(self):
+    temporary_directory = tempfile.mkdtemp(dir=self.temporary_directory)
+    storage_backend = securesystemslib.storage.FilesystemBackend()
+    version = 1
+
+    # Test an rsa accumulator with a few nodes to verify the output
+
+    test_nodes = {}
+    test_nodes['file1'] = tuf.formats.make_metadata_fileinfo(5, None, None)
+
+
+    root_1, leaves = repo_lib._build_rsa_acc(test_nodes)
+    repo_lib._write_rsa_leaves(root_1, leaves, storage_backend,
+        temporary_directory, version)
+
+    # Ensure that the paths are written to the directory
+    file_path = os.path.join(temporary_directory, 'file1-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    file_path = os.path.join(temporary_directory, '1.file1-snapshot.json')
+    self.assertTrue(os.path.exists(file_path))
+
+    self.assertEqual(root_1, 5)
+
+    test_nodes = {}
+    test_nodes['targets'] = tuf.formats.make_metadata_fileinfo(1, None, None)
+    test_nodes['role1'] = tuf.formats.make_metadata_fileinfo(1, None, None)
+    test_nodes['role2'] = tuf.formats.make_metadata_fileinfo(1, None, None)
+
+    root, leaves = repo_lib._build_rsa_acc(test_nodes)
+
+
+
+
   def _setup_generate_snapshot_metadata_test(self):
     # Test normal case.
     temporary_directory = tempfile.mkdtemp(dir=self.temporary_directory)
--- a/tests/test_repository_lib.py
+++ b/tests/test_repository_lib.py
@@ -470,7 +470,7 @@ class TestRepositoryToolFunctions(unittest.TestCase):
 
 
     root_1, leaves = repo_lib._build_rsa_acc(test_nodes)
-    repo_lib._write_rsa_leaves(root_1, leaves, storage_backend,
+    repo_lib._write_rsa_proofs(root_1, leaves, storage_backend,
         temporary_directory, version)
 
     # Ensure that the paths are written to the directory
@@ -480,6 +480,7 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     file_path = os.path.join(temporary_directory, '1.file1-snapshot.json')
     self.assertTrue(os.path.exists(file_path))
 
+    # TODO: check against the correct root value
     self.assertEqual(root_1, 5)
 
     test_nodes = {}
@@ -533,7 +534,7 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     metadata_directory, version, expiration_date, storage_backend = \
         self._setup_generate_snapshot_metadata_test()
 
-    snapshot_metadata = \
+    snapshot_metadata, _ = \
       repo_lib.generate_snapshot_metadata(metadata_directory, version,
                                           expiration_date,
                                           storage_backend,
@@ -561,7 +562,7 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     metadata_directory, version, expiration_date, storage_backend = \
         self._setup_generate_snapshot_metadata_test()
 
-    snapshot_metadata = \
+    snapshot_metadata, _ = \
       repo_lib.generate_snapshot_metadata(metadata_directory, version,
                                           expiration_date,
                                           storage_backend,
@@ -592,7 +593,7 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     metadata_directory, version, expiration_date, storage_backend = \
         self._setup_generate_snapshot_metadata_test()
 
-    snapshot_metadata = \
+    snapshot_metadata, _ = \
       repo_lib.generate_snapshot_metadata(metadata_directory, version,
                                           expiration_date,
                                           storage_backend,
@@ -623,7 +624,7 @@ class TestRepositoryToolFunctions(unittest.TestCase):
     metadata_directory, version, expiration_date, storage_backend = \
         self._setup_generate_snapshot_metadata_test()
 
-    snapshot_metadata = \
+    snapshot_metadata, _ = \
       repo_lib.generate_snapshot_metadata(metadata_directory, version,
                                           expiration_date,
                                           storage_backend,
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -1771,6 +1771,35 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
 
 
+  def test_snapshot_rsa_acc(self):
+    # replace timestamp with an RSA accumulator timestamp and create the updater
+    rsa_acc_timestamp = os.path.join(self.repository_directory, 'metadata', 'timestamp-rsa.json')
+    timestamp = os.path.join(self.repository_directory, 'metadata', 'timestamp.json')
+
+    shutil.move(rsa_acc_timestamp, timestamp)
+
+    repository_updater = updater.Updater(self.repository_name,
+                                      self.repository_mirrors)
+    repository_updater.refresh()
+
+    # Test verify RSA accumulator proof
+    snapshot_info = repository_updater.verify_rsa_acc_proof('targets')
+    self.assertEqual(snapshot_info['version'], 1)
+
+    snapshot_info = repository_updater.verify_rsa_acc_proof('role1')
+    self.assertEqual(snapshot_info['version'], 1)
+
+    # verify RSA accumulator with invalid role
+    self.assertRaises(tuf.exceptions.NoWorkingMirrorError,
+        repository_updater.verify_rsa_acc_proof, 'foo')
+
+    # Test get_one_valid_targetinfo with snapshot RSA accumulator
+    repository_updater.get_one_valid_targetinfo('file1.txt')
+
+
+
+
+
 class TestMultiRepoUpdater(unittest_toolbox.Modified_TestCase):
 
   def setUp(self):
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1660,7 +1660,7 @@ class Updater(object):
             ". The update will continue as the major versions match.")
 
     except (ValueError, TypeError) as error:
-      six.raise_from(securesystemslib.exceptions.FormatError('Improperly'
+      raise exceptions.FormatError('Improperly'
           ' formatted spec_version, which must be in major.minor.fix format'),
           error)
 
--- a/tuf/client/updater.py
+++ b/tuf/client/updater.py
@@ -1087,8 +1087,12 @@ class Updater(object):
     # require strict checks on its required length.
     self._update_metadata('timestamp', DEFAULT_TIMESTAMP_UPPERLENGTH)
 
-    self._update_metadata_if_changed('snapshot',
-        referenced_metadata='timestamp')
+    if 'rsa_acc' not in self.metadata['current']['timestamp']:
+      # If an RSA Accumulator is defined, do not update snapshot metadata. Instead,
+      # we will download the relevant proof files later when downloading
+      # a target.
+      self._update_metadata_if_changed('snapshot',
+          referenced_metadata='timestamp')
     self._update_metadata_if_changed('targets')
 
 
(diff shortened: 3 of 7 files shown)


tests/test_repository_lib.py
```diff
@@ -458,6 +458,41 @@ def test_generate_targets_metadata(self):
...
+
+    # TODO: check against the correct root value
+    self.assertEqual(root_1, 5)
+
+    test_nodes = {}
+    test_nodes['targets'] = tuf.formats.make_metadata_fileinfo(1, None, None)
+    test_nodes['role1'] = tuf.formats.make_metadata_fileinfo(1, None, None)
+    test_nodes['role2'] = tuf.formats.make_metadata_fileinfo(1, None, None)
+
+    root, leaves = repo_lib._build_rsa_acc(test_nodes)
```
znewman01, line 488: Not sure what's going on here -- above, there's a whole build/write/check files/check for specific value cycle, whereas here you just seem to be verifying that it doesn't crash?

tests/test_updater.py
```diff
@@ -1771,6 +1771,35 @@ def test_13__targets_of_role(self):
...
+    repository_updater.refresh()
+
+    # Test verify RSA accumulator proof
+    snapshot_info = repository_updater.verify_rsa_acc_proof('targets')
+    self.assertEqual(snapshot_info['version'], 1)
+
+    snapshot_info = repository_updater.verify_rsa_acc_proof('role1')
+    self.assertEqual(snapshot_info['version'], 1)
+
+    # verify RSA accumulator with invalid role
```
znewman01, line 1792: Would like to also see a test for verifying an existing role against an invalid accumulator

tuf/client/updater.py
```diff
@@ -1087,8 +1087,12 @@ def refresh(self, unsafely_update_root_if_necessary=True):
     # require strict checks on its required length.
     self._update_metadata('timestamp', DEFAULT_TIMESTAMP_UPPERLENGTH)
 
-    self._update_metadata_if_changed('snapshot',
-        referenced_metadata='timestamp')
+    if 'rsa_acc' not in self.metadata['current']['timestamp']:
+      # If an RSA Accumulator is defined, do not update snapshot metadata. Instead,
+      # we will download the relevant proof files later when downloading
+      # a target.
```
znewman01, line 1093: The Merkle stuff isn't merged yet, right? You'd have to do the same thing there too.

Which is actually making me think there's room for abstraction here: one interface for "full manifest"/"rsa"/"merkle" so that not every user of the metadata has to interact with something called "rsa"

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
 
 
 
+  def signable_verification(self, metadata_role, file_object, expected_version):
```
znewman01, line 1623: Could use a doc comment.

Also, naming nitpick: I tend to like all functions/methods to be verbs, so it's clearer what they do (e.g., "make_signable_verification" over "signable_verification" but even better "attest", "prove", etc. (if those are accurate))

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+    # Verify 'file_object' according to the callable function.
```
znewman01, line 1624: What callable function?

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+
+
+
+
+  def _update_rsa_acc_metadata(self, proof_filename, upperbound_filelength,
+      version=None):
+    """
+    <Purpose>
+      Non-public method that downloads, verifies, and 'installs' the proof
+      metadata belonging to 'proof_filename'.  Calling this method implies
```
znewman01, line 1712: Nit: "Calling this method implies" to "This method should be called when", the wording of the former confuses me as to whether it's a precondition, postcondition, or just the intended use

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+      downloaded and verified.  This method also assumes that the store of
+      top-level metadata is the latest and exists.
+
+    <Arguments>
+      proof_filename:
+        The name of the metadata. This is an RSA accumulator proof file and should
+        not end in '.json'.  Examples: 'role1-snapshot', 'targets-snapshot'
+
+      upperbound_filelength:
+        The expected length, or upper bound, of the metadata file to be
```
znewman01, line 1725: Why? i.e., what happens if this is too low/high?

- wasted resources
- it'll error
- etc.

Also, how would I (as a caller) know this?

Also, units? Is it bytes?

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+      if metadata_spec_minor_version != code_spec_minor_version:
+        logger.info("Downloaded metadata that specifies a different minor " +
+            "spec_version. This code has version " +
+            str(tuf.SPECIFICATION_VERSION) +
+            " and the metadata lists version number " +
+            str(metadata_spec_version) +
+            ". The update will continue as the major versions match.")
+
+    except (ValueError, TypeError) as error:
+      six.raise_from(securesystemslib.exceptions.FormatError('Improperly'
```
znewman01: You're still supporting Py2? :'(

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+      proof_filename:
+        The name of the metadata. This is an RSA accumulator proof file and should
+        not end in '.json'.  Examples: 'role1-snapshot', 'targets-snapshot'
+
+      upperbound_filelength:
+        The expected length, or upper bound, of the metadata file to be
+        downloaded.
+
+      version:
+        The expected and required version number of the 'proof_filename' file
```
znewman01, line 1729: Nit: what does "required" mean? i.e., what happens if I get this wrong?

Also, maybe this should be clear to someone more familiar with the codebase, but...the idea is that the file represented by `proof_filename` has a monotonically increasing version number? Or is that the specification version?

And it looks like this is optional---why would I pass it? How would I know what to pass?

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+    <Returns>
+      None.
+    """
+
+    # Construct the metadata filename as expected by the download/mirror
+    # modules.
+    metadata_filename = proof_filename + '.json'
+
+    # Attempt a file download from each mirror until the file is downloaded and
+    # verified.  If the signature of the downloaded file is valid, proceed,
```
znewman01, line 1752: Grammar nit (the worst kind, sorry!) : `s/procced,/proceed;`

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+    # otherwise log a warning and try the next mirror.  'metadata_file_object'
```
znewman01, line 1753: Nit: move the description of the local variables next to where they're first instantiated?

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+    # is the file-like object returned by 'download.py'.  'metadata_signable'
+    # is the object extracted from 'metadata_file_object'.  Metadata saved to
+    # files are regarded as 'signable' objects, conformant to
+    # 'tuf.formats.SIGNABLE_SCHEMA'.
+    #
+    # Some metadata (presently timestamp) will be downloaded "unsafely", in the
```
znewman01, line 1759: Maybe this paragraph makes more sense next to `verification_fn`

tuf/client/updater.py
```diff
@@ -1616,6 +1620,206 @@ def _get_metadata_file(self, metadata_role, remote_filename,
...
+
+
+
+
+
+  def _update_rsa_acc_metadata(self, proof_filename, upperbound_filelength,
+      version=None):
+    """
+    <Purpose>
+      Non-public method that downloads, verifies, and 'installs' the proof
```
znewman01, line 1711: So, to be clear (this is because I don't know the codebase, there's nothing wrong with the code/docs):

- there's some kind of local state directory
- you call this when you want the "proof metadata" associated with "proof_filename"
- this method will get that from the repository
- and after you call it, the local state directory will have the (updated) proof metadata

tuf/client/updater.py
```diff
@@ -1732,6 +1936,87 @@ def _update_metadata(self, metadata_role, upperbound_filelength, version=None):
...
+        If the snapshot rsa accumulator file is invalid or the verification fails
+    <Returns>
+      A dictionary containing the snapshot information about metadata role,
+      conforming to VERSIONINFO_SCHEMA or METADATA_FILEINFO_SCHEMA
+    """
+
+    # Modulus from https://en.wikipedia.org/wiki/RSA_numbers#RSA-2048
+    # We will want to generate a new one
+    # This is duplicate code from repo lib, should live somewhere else
+    Modulus = "2519590847565789349402718324004839857142928212620403202777713783604366202070759555626401852588078" + \
```
znewman01, line 1957: Nit: CamelCase for constants? Doesn't seem inline with [Secure Systems Lab style](https://github.com/secure-systems-lab/code-style-guidelines/blob/master/python.md) or [Google Python style](https://google.github.io/styleguide/pyguide.html#316-naming)

(Ignore if you were planning to rename once you find a home.)

tuf/client/updater.py
```diff
@@ -1732,6 +1936,87 @@ def _update_metadata(self, metadata_role, upperbound_filelength, version=None):
...
+      tuf.exceptions.RepositoryError:
+        If the snapshot rsa accumulator file is invalid or the verification fails
+    <Returns>
+      A dictionary containing the snapshot information about metadata role,
+      conforming to VERSIONINFO_SCHEMA or METADATA_FILEINFO_SCHEMA
+    """
+
+    # Modulus from https://en.wikipedia.org/wiki/RSA_numbers#RSA-2048
+    # We will want to generate a new one
+    # This is duplicate code from repo lib, should live somewhere else
```
znewman01, line 1956: Nit: "TODO:"?

tuf/client/updater.py
```diff
@@ -1732,6 +1936,87 @@ def _update_metadata(self, metadata_role, upperbound_filelength, version=None):
...
+    m = int(Modulus, 10)
+
+
+    if not rsa_acc:
+      rsa_acc = self.metadata['current']['timestamp']['rsa_acc']
+
+    metadata_rolename = metadata_role + '-snapshot'
+
+    # Download RSA accumulator proof
+    upperbound_filelength = tuf.settings.MERKLE_FILELENGTH
```
znewman01, line 1972: Add TODO to add a new setting?

tuf/client/updater.py
```diff
@@ -1732,6 +1936,87 @@ def _update_metadata(self, metadata_role, upperbound_filelength, version=None):
...
+
+    # Ensure the metadata path is valid/exists, else ignore the call.
+    if not os.path.exists(metadata_filepath):
+      # No RSA accumulator proof found
+      raise tuf.exceptions.RepositoryError('No snapshot rsa accumulator proof file for ' +
+          metadata_role)
+    try:
+      snapshot_rsa_acc_proof = securesystemslib.util.load_json_file(
+          metadata_filepath)
+
```
znewman01, line 1986: Nit: remove this blank? (And maybe put one before the `try:`?)

tuf/client/updater.py
```diff
@@ -1732,6 +1936,87 @@ def _update_metadata(self, metadata_role, upperbound_filelength, version=None):
...
+    # Modulus from https://en.wikipedia.org/wiki/RSA_numbers#RSA-2048
+    # We will want to generate a new one
+    # This is duplicate code from repo lib, should live somewhere else
+    Modulus = "2519590847565789349402718324004839857142928212620403202777713783604366202070759555626401852588078" + \
+      "4406918290641249515082189298559149176184502808489120072844992687392807287776735971418347270261896375014971" + \
+      "8246911650776133798590957000973304597488084284017974291006424586918171951187461215151726546322822168699875" + \
+      "4918242243363725908514186546204357679842338718477444792073993423658482382428119816381501067481045166037730" + \
+      "6056201619676256133844143603833904414952634432190114657544454178424020924616515723350778707749817125772467" + \
+      "962926386356373289912154831438167899885040445364023527381951378636564391212010397122822120720357"
+    m = int(Modulus, 10)
```
znewman01, line 1963: You may want to use [`gmpy`](https://github.com/aleaxit/gmpy) or similar for [performance](https://jasonstitt.com/c-extension-n-choose-k) down the line.

tuf/client/updater.py
```diff
@@ -1732,6 +1936,87 @@ def _update_metadata(self, metadata_role, upperbound_filelength, version=None):
...
+    tuf.formats.SNAPSHOT_RSA_ACC_SCHEMA.check_match(snapshot_rsa_acc_proof)
+
+    # canonicalize the contents to determine the RSA accumulator prime
+    contents = snapshot_rsa_acc_proof['leaf_contents']
+    json_contents = securesystemslib.formats.encode_canonical(contents)
+
+    prime = repository_lib.hash_to_prime(json_contents)
+
+    # RSA accumulator proof
+    proof = snapshot_rsa_acc_proof['rsa_acc_proof']
```
znewman01, line 2004: I'd suggest abstracting over the actual verification here for testability, separation-of-concerns, etc.

The interface would maybe look something like:

```python
class RsaManifest:  # which would implement some Manifest interface
    MODULUS = 17  # biggest prime I know
    def is_in_manifest(self, entry, proof):
        canonical = ssl.format.encode_canonical(value)
        member = rl.hash_to_prime(canonical)
        return pow(proof, member, self.MODULUS) == self.digest
   def prove_membership(self, entry):
       raise NotImplemented('TODO: implement')
```

This would let me reason about the accumulator independently of the TUF repository etc. (and let us transparently swap for Merkle etc.)

EDIT: looks like you kinda do in `repository_lib.py` -- why not use it here?

tuf/repository_lib.py
```diff
@@ -30,6 +30,8 @@
 import shutil
 import json
 import tempfile
+from pyblake2 import blake2b
```
znewman01, line 33: FYI blake2 added to [stdlib `hashlib` in Py3.6](https://docs.python.org/3.6/library/hashlib.html?highlight=hashlib#hashlib.blake2b)

However, both limit the digest size to 64 bytes.  We really want an "extendable output function" (XOF), which is a hash function with arbitrary output size (see my comments in `hash_to_prime`). Candidates:

- SHAKE (included in [SHA3 standard](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf)) and [Python 3.6+ `hashlib`](https://docs.python.org/3.6/library/hashlib.html?highlight=hashlib#shake-variable-length-digests)
- [blake2x](https://www.blake2.net/blake2x.pdf) (blake2 team)
- [KangarooTwelve/MarsupilamiFourteen](https://keccak.team/kangarootwelve.html) (Keccak (SHA3) team) with a [pure-Python reference implementation](https://github.com/XKCP/XKCP/blob/master/Standalone/KangarooTwelve/Python/K12.py)

tuf/repository_lib.py
```diff
@@ -90,7 +92,7 @@ def _generate_and_write_metadata(rolename, metadata_filename,
   increment_version_number=True, repository_name='default',
   use_existing_fileinfo=False, use_timestamp_length=True,
   use_timestamp_hashes=True, use_snapshot_length=False,
-  use_snapshot_hashes=False):
+  use_snapshot_hashes=False, rsa_acc=False):
```
znewman01, line 95: 16 parameters---time for a refactor?  :smile:

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
 
 
 
+
+# I couldn't find a currently maintained python library for this, so
+# implementing it here. It would be better to implement this in c,
+# even better to use an existing library
```
znewman01, line 1565: As I mentioned above, gmpy/gmpy2 would be a good candidate for numberical ops, including primality testing (though adds a dependency on a C library libgmp).

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+# This is inspired by: https://www.literateprograms.org/miller-rabin_primality_test__python_.html
+def miller_rabin_round(a, s, d, n):
```
znewman01, line 1567: Didn't thoroughly review because we should definitely use someone else's implementation :)

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+
+
+
+
+# RSA Accumulator code insprired by https://github.com/ElrondNetwork/elrond-go/blob/v1.0.30/crypto/accumulator/rsa/rsaAcc.go
+def hash_to_prime(data):
+  # TODO: move constant definitions
+  basesMillerRabin = 12
+
+  h = blake2b(str(data).encode('utf-8'))
```
znewman01, line 1607: Big issue: this only gives you a 512 bit number, so we're using 1 out of 2^{2048 -512} = 2^{1536}th of the possible space!

Need to use an XOF to get >512 bits. Also, we want uniformly random values from [0, MODULUS).

See, e.g., [[Lem18]](https://arxiv.org/abs/1805.10941)

I'd recommend to implement as (approximate):

```python
def hash_to_prime(data, modulus):
  modulus_bits = math.ceil(math.log2(modulus))  # TODO: may need to round this up to the nearest byte

  # a number n has about 1/ln(n) chance of being prime so we may actually want to make this depend on our modulus
  for _ in range(MAX_ATTEMPTS):
    digest: bytes = xof(data).digest(modulus_bits)  # might be > modulus, so we need to wrap
    value = int.from_bytes(digest, 'big') # endianness arbitrary but must be the same everywhere

    # value should be in [0, modulus)
    if value >= modulus:
      # rejection sampling: this happens up to 1/2 of the time if modulus is 
      # Lemire gives a better method (alg. 5): https://arxiv.org/pdf/1805.10941.pdf
      continue

    if not miller_rabin(value, MILLER_RABIN_ROUNDS):
      continue  # not prime
  return value
```

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+  Create an RSA accululator from the snapshot fileinfo and writes it to individual snapshot files
+
+  Returns the root and leaves
+  """
+
+  # RSA accululator contants
+  # TODO: move constant definitions
+  g = 3
+  # Modulus from https://en.wikipedia.org/wiki/RSA_numbers#RSA-2048
+  # We will want to generate a new one
```
znewman01, line 1646: This might make sense as a setting and/or part of the format---not everyone should use ours.

(I think this is what you're saying, just trying to make explicit)

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+    contents["name"] = name
+    self.contents = contents
+    self.name = name
+
+  def set_proof(self, proof):
+    self.proof = proof
+
+
+
+def _build_rsa_acc(fileinfodict):
```
znewman01, line 1635: You're eventually going to want to separate out:

- initial build
- update
- removal (possibly)

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+		"962926386356373289912154831438167899885040445364023527381951378636564391212010397122822120720357"
+  m = int(Modulus, 10)
+
+
+  # We will build the accumulator starting with the leaf nodes. Each
+  # leaf contains snapshot information for a single metadata file.
+  leaves = []
+  primes = []
+  acc_exp = 1
+  for name, contents in sorted(fileinfodict.items()):
```
znewman01, line 1661: I have a preference for splitting all format-specific operations from all number-theoretic operations, which makes testing/understanding a lot easier.

So maybe a method like `build(values)` which takes a list of things to be hashed?

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+    if name.endswith(".json"):
+      name = os.path.splitext(name)[0]
+    cont = acc_contents(name, contents)
+    leaves.append(cont)
+
+    json_contents = securesystemslib.formats.encode_canonical(contents)
+    prime = hash_to_prime(json_contents)
+    primes.append(prime)
+    acc_exp = acc_exp * prime
```
znewman01, line 1670: This is a hidden O(n^2) algorithm: acc_exp is O(n) in each iteration!

I think there's ways to do this O(n lg n) but need to think about it/dredge them up.

Best trick (O(n)) is to reduce acc_exp mod the order of the group---but that only works if you know the group order (which means knowing the RSA factors---we can think about whether that works in our setting. e.g., if you sign manifests we're kinda implicitly trusting you -- *except* then we lose a little bit of auditability because you can forge proofs if you know the RSA factors. we should discuss).

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+    p = int(h.hexdigest(), 16)
+
+  return p
+
+
+
+class acc_contents(object):
+  contents = None
+  name = None
+  proof = None
```
znewman01, line 1622: Is this so we can hold precomputed proofs for each element?

tuf/repository_lib.py
```diff
@@ -1541,6 +1559,168 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
+  leaves = []
+  primes = []
+  acc_exp = 1
+  for name, contents in sorted(fileinfodict.items()):
+    if name.endswith(".json"):
+      name = os.path.splitext(name)[0]
+    cont = acc_contents(name, contents)
+    leaves.append(cont)
+
+    json_contents = securesystemslib.formats.encode_canonical(contents)
```
znewman01, line 1667: Above, you say `acc_contents` needs `name`, but here we're encoding *without* `name`!

znewman01 (commented):
Hey! Sorry for the long lag on this.

I'm a thorough (AKA annoying :smile:) code reviewer so feel free to ignore most of what I said. The comments in `repository_lib.py` are probably the important ones.

If anything's confusing, feel free to ask for clarification or we can set up a time to (video) call about it.

lukpueh:
Ping @mnm678, same questions as in https://github.com/theupdateframework/python-tuf/pull/1113#issuecomment-1043028730.

mnm678:
Closing in favor of writing this on top of the new Metadata api.

PR #1511, opened 2021-07-28
new API: test containers for zero or more elements

### LINKED ISSUE #1485: Test [de]serialization of metadata with more than one item in each container and zero items (where valid)

Testing, particularly [de]serialisation, should cover metadata containers with more than one item and, where it makes sense (per the specification), zero items.

**Current behavior**:

Testing only covers one item per container.

Ref https://github.com/theupdateframework/tuf/issues/1478

MVrachev:
Fixes #1485 

Test metadata (de)serialization with input data containing containers with zero or more elements.
    
Here is the status for the different use cases:

Root:
- many keys: added
- many roles: added

Root role keyids:
- many keids: already added in https://github.com/theupdateframework/tuf/pull/1481

MetaFile hashes:
- many hashes: already tested
- zero hashes: added. Testing as invalid test case.

Timestamp meta:
- zero elements: already tested
- many elements: added

Snapshot meta:
- zero items: added
- many items: added

Delegation:
- many keys: added
- many keyids: added

Delegation role:
- many paths: already tested
- many path_hash_path_prefixes: already tested

Delegation roles:
- zero roles: added
- multiple roles: added

Targets targets:
- zero items: already tested
- multiple items: added

Additional tests are added for containers with 0 or more elements for some specific cases.
Those tests are needed to cover use cases when syntactically as
standalone objects the metadata classes and their helper classes defined
in tuf/api/metadata.py are valid even if they cannot be verified.

An example where an object is valid, but cannot be verified is
if we have a Role instance with an empty list of "keyids".
This instance is valid and can be created, but cannot be verified
because there is a requirement that the threshold should be above
1, meaning that there should be at least 1 element inside the "keyids"
list to complete successful threshold verification.

The situation is the same for the rest of the tests I am adding to this
commit:
- Root object without keys
- Root object without roles
- DelegationRole object with empty "keyids"
- DelegationRole object with an empty list of "paths"
- DelegationRole object with an empty list of "path_hash_prefixes"
all of these objects can be instantiated, but cannot complete successfully
threshold verification.

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -155,8 +155,12 @@ class TestSerialization(unittest.TestCase):
     valid_roots: DataSet = {
         "all": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
             "expires": "2030-01-01T00:00:00Z", "consistent_snapshot": false, \
-            "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
-            "roles": { "targets": {"keyids": ["keyid"], "threshold": 3}} \
+            "keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
+                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
+            "roles": { \
+                "targets": {"keyids": ["keyid1"], "threshold": 1}, \
+                "snapshot": {"keyids": ["keyid2"], "threshold": 1}} \
             }',
         "no consistent_snapshot": '{ "_type": "root", "spec_version": "1.0.0", "version": 1, \
             "expires": "2030-01-01T00:00:00Z", \
@@ -198,6 +202,7 @@ class TestSerialization(unittest.TestCase):
         "no length": '{"hashes": {"sha256" : "abc"}, "version": 1 }',
         "no hashes": '{"length": 12, "version": 1}',
         "unrecognized field": '{"hashes": {"sha256" : "abc"}, "length": 12, "version": 1, "foo": "bar"}',
+        "many hashes": '{"hashes": {"sha256" : "abc", "sha512": "cde"}, "length": 12, "version": 1}',
     }
 
     @run_sub_tests_with_dataset(valid_metafiles)
@@ -205,6 +210,16 @@ class TestSerialization(unittest.TestCase):
         case_dict = json.loads(test_case_data)
         metafile = MetaFile.from_dict(copy.copy(case_dict))
         self.assertDictEqual(case_dict, metafile.to_dict())
+
+    invalid_timestamps: DataSet = {
+        "no metafile": '{ "_type": "timestamp", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z"}',
+    }
+
+    @run_sub_tests_with_dataset(invalid_timestamps)
+    def test_invalid_timestamp_serialization(self, test_case_data: Dict[str, str]):
+        case_dict = json.loads(test_case_data)
+        with self.assertRaises((ValueError, KeyError)):
+            Timestamp.from_dict(copy.deepcopy(case_dict))
 
 
     valid_timestamps: DataSet = {
@@ -223,7 +238,13 @@ class TestSerialization(unittest.TestCase):
 
     valid_snapshots: DataSet = {
         "all": '{ "_type": "snapshot", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
-            "meta": { "file.txt": { "hashes": {"sha256" : "abc"}, "version": 1 }}}',
+            "meta": { \
+                "file1.txt": {"hashes": {"sha256" : "abc"}, "version": 1}, \
+                "file2.txt": {"hashes": {"sha256" : "cde"}, "version": 1} \
+            }}',
+        "empty meta": '{ "_type": "snapshot", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
+            "meta": {} \
+            }',
         "unrecognized field": '{ "_type": "snapshot", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
             "meta": { "file.txt": { "hashes": {"sha256" : "abc"}, "version": 1 }}, "foo": "bar"}',
     }
@@ -243,8 +264,10 @@ class TestSerialization(unittest.TestCase):
             '{"keyids": ["keyid"], "name": "a", "terminating": false, \
             "path_hash_prefixes": ["h1", "h2"], "threshold": 99}',
         "unrecognized field":
-            '{"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], \
-            "terminating": true, "threshold": 3, "foo": "bar"}',
+            '{"keyids": ["keyid"], "name": "a", "terminating": true, "paths": ["fn1"], "threshold": 3, "foo": "bar"}',
+        "many keyids":
+            '{"keyids": ["keyid1", "keyid2"], "name": "a", "paths": ["fn1", "fn2"], \
+            "terminating": false, "threshold": 1}',
     }
 
     @run_sub_tests_with_dataset(valid_delegated_roles)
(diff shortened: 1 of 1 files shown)


tests/test_metadata_serialization.py
```diff
@@ -184,6 +203,7 @@ def test_root_serialization(self, test_case_data: str):
         "empty hashes dict": '{"version": 1, "length": 1, "hashes": {}}',
         "hashes wrong type": '{"version": 1, "length": 1, "hashes": 1}',
         "hashes values wrong type": '{"version": 1, "length": 1, "hashes": {"sha256": 1}}',
+        "zero hashes": '{"version": 1, "length": 1, "hashes": {}}',
```
jku: this is already tested above

tests/test_metadata_serialization.py
```diff
@@ -168,6 +168,25 @@ def test_role_serialization(self, test_case_data: str):
...
+            "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": { \
+                "targets": {"keyids": ["keyid1"], "threshold": 1}, \
+                "snapshot": {"keyids": ["keyid2"], "threshold": 1}} \
+            }',
+        "many role keyids": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
+            "expires": "2030-01-01T00:00:00Z", "consistent_snapshot": false, \
+            "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": {"targets": {"keyids": ["keyid1", "keyid2"], "threshold": 3}} \
+            }',
```
jku: this is already tested in role tests "many keyids" case

jku:
> According to the spec having an empty container for any of these cases ... is not allowed

is this really true? :open_mouth:

tests/test_metadata_serialization.py
```diff
@@ -168,6 +168,25 @@ def test_role_serialization(self, test_case_data: str):
...
+                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
+            "roles": { "targets": {"keyids": ["keyid"], "threshold": 3}} \
+            }',
+        "many roles": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
+            "expires": "2030-01-01T00:00:00Z", "consistent_snapshot": false, \
+            "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": { \
+                "targets": {"keyids": ["keyid1"], "threshold": 1}, \
+                "snapshot": {"keyids": ["keyid2"], "threshold": 1}} \
+            }',
```
jku: these are fine but to minimize the boilerplate you could just combine these into the "all" case as well? either way works

jku (commented):
some potential test cases
* timestamp meta with multiple items
* timestamp meta with no items
* snapshot meta with multiple items
* delegation with multiple roles
* targets targets with multiple items
* targets roles with multiple items

or do I misunderstand the scope of the PR?

tests/test_metadata_serialization.py
MVrachev: Good suggestion. We don't need to be explicit when we test about `multiple roles`.
Instead, the focus is to make sure we are testing for them no matter in which test cases.

MVrachev:
@jku I had to rebase, but https://github.com/theupdateframework/tuf/pull/1511/commits/92dc84453f2dbff97385c38390af70bd731d76e4 is unchanged.

The new commits are https://github.com/theupdateframework/tuf/pull/1511/commits/75b2c21b4ea5d8f2f93287cfffcad57e4e9a5aa7 and https://github.com/theupdateframework/tuf/pull/1511/commits/033e0696760e7664ee869efea9f8957112380016.

MVrachev:
> > According to the spec having an empty container for any of these cases ... is not allowed
> 
> is this really true? 😮

- Root keys are not allowed to be empty, because it's required that at least one element is inside it - the key used to sign the root metadata file itself.
- Root roles are not allowed to be empty, because it's required that there should be information for at least one role - the root role.
- Root role `keyids` are not allowed to be empty, because it's required that threshold should be at least 1 (see #1450) and as a consequence, there should be at least a threshold number of elements in keyids


- Delegation keys and Delegation roles 
For those two it's a little harder and probably my statement is wrong.
This use case is valid:
```
"delegations": {
     "keys": {}
     "roles": {}
```

for which by the way we are not testing. I included a test.
The problem comes if `Delegation keys` are empty, then means there shouldn't be any roles.
So, maybe my assumption is not correct.

But still, where do you think we should test for that?

- DelegationRole keyids cannot be empty because every role needs at least one key to be verified because the threshold is >= 1.

I am wrong about those two:

-  DelegationRole paths and DelegationRole path_hash_prefixes
I am already testing for them.

Do you think I should modify my commit message with more info as I did here?

jku:
I think you may be mixing _"is this a syntactically valid metadata"_ with "_can this metadata file be used successfully in every part of the update process"_. 

I mean when you say things like:

> Root role keyids are not allowed to be empty, because it's required that threshold should be at least 1

from the perspective of the file format and the containing metadata, keyids and threshold are not related. The spec does not say metadata files can't exist with a role that has less keys than the corresponding threshold. Yes, we happen to know that those keys then can't possibly verify that roles metadata when needed but the metadata that contains this role is still valid.

jku:
For Root roles specifically there is a spec mention (with a weird reference to "key list" but the intent seems clear: roles dictionary must contain these 4 or 5 items):

> A role for each of "root", "snapshot", "timestamp", and "targets" MUST be specified in the key list.  The role of "mirror" is OPTIONAL. 

This we should test for in Root construction time I guess.

tuf/api/metadata.py
```diff
@@ -878,20 +878,24 @@ def __init__(
         version: int,
         spec_version: str,
         expires: datetime,
-        meta: Dict[str, MetaFile],
```
jku: can you explain this?

MVrachev: Yes. I noticed we don't do validation on `Timestamp.meta`.
According to the spec, the value of `Timestamp.meta` called "METAFILES" is:
```
METAFILES is the same as described for the snapshot.json file.
In the case of the timestamp.json file, this MUST only include a
description of the snapshot.json file.
```
That's why in order to include validation for meta no matter if we create objects with `Timestamp.from_dict()`
or with the constructor like `Timestamp()` I moved the meta validation and object creation inside `__init__`.

MVrachev:
I had to rebase on top of develop.
Additionally I:
- added a new commit with tests for those valid cases, but which we cannot verify in a trusted set.
- squashed to similar commits together
- updated my commits and pr descriptions 
- had another look through the spec to see if we are missing something and it seems everything is okay.

After a discussion with @jku about his words

> I think you may be mixing "is this a syntactically valid metadata" with "can this metadata file be used successfully in every part of the update process".
I mean when you say things like:
> > Root role keyids are not allowed to be empty, because it's required that threshold should be at least 1

>from the perspective of the file format and the containing metadata, keyids and threshold are not related. The spec does not say metadata files can't exist with a role that has less keys than the corresponding threshold. Yes, we happen to know that those keys then can't possibly verify that roles metadata when needed but the metadata that contains this role is still valid.

I realized he is right and standalone objects in those cases can exist and are valid, but cannot be verified.
That's why I created a new commit explaining that.

> For Root roles specifically there is a spec mention (with a weird reference to "key list" but the intent seems clear: roles dictionary must contain these 4 or 5 items):
A role for each of "root", "snapshot", "timestamp", and "targets" MUST be specified in the key list. The role of "mirror" is OPTIONAL.
This we should test for in Root construction time I guess.

There is a separate issue for that #1516.

tuf/api/metadata.py
jku: I don't agree with this change. 
* it's not in line with all other constructors that do take actual objects and not json-like dicts
* It's going to make the API for creating a new Timestamp from scratch worse (caller feeds in json-like dicts instead of well defined objects). 
* It makes annotations useless: we should get rid of all `Any` that we can, not add more

I'm fine with you not doing this validation at all right now: we could just wait for the snapshot_meta discussion to reach a conclusion first

jku (changes requested):
LGTM (apart from the Timestamp constructor api change which I disagree with), thanks.

some of the targets test data are getting quite big ... but personally I still prefer parsing this sort of multiple lines of json in my head: at least every case is understandable without external context -- opinions may vary though

tuf/api/metadata.py
MVrachev: Now with the new discussion about `snapshot_meta` I agree that we shouldn't hurry and add validation.
Will remove the commit.

MVrachev:
@jku updated the pr by dropping the timestamp meta validation commit and removing multiple metafiles in timestamp test.

jku (approved):
Thanks!

Issue #1506, opened 2021-07-23
ngclient: avoid lstrip(os.sep) on target paths


sechkova:
**Description of issue or feature request**:

During matching of target file path and pathpattern in ngclient/updater.py we normalize the paths by removing the leading os separator.

```python
 # Make sure to strip any leading
 # path separators so that a match is made.
 # Example: "foo.tgz" should match with "/*.tgz".
 if fnmatch.fnmatch(
     target_filepath.lstrip(os.sep), pathpattern.lstrip(os.sep)
 ):
```

**Current behavior**:
Normalize file path and pathpattern in case they start with `os.sep`

**Expected behavior**:
Define a valid target file path and pathpattern strings and reject everything else.
Don't support corner cases.

jku:
some additional thoughts here, first for pathpattern:
* pathpattern comes from the server: the spec "recommends"  that '/' is used as the separator. I believe it is impossible for a client implementation to support arbitrary separators -- every implementation needs to choose one and stick with it. I also believe it makes no sense for even the reference implementation to even optionally use/accept anything else than '/'. We should just document that we are stricter than the spec here
*  spec "recommends" that pathpattern not start with a separator: I guess we can leave the pathpattern stripping in there, but for '/' specifically, not for os.sep -- that part is clearly a bug, the client os separator has nothing to do with pathpattern separator which is decided by the server

For target_filepath:
* this comes from API user so using os.sep _may_ make sense in some client API -- however I think there is no reason to accept file paths _starting_ with a separator? It's just complexity for no benefit, I would just remove this lstrip().

joshuagl:
[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119) states:
> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

IMNSHO the reference implementation does not have a valid reason to ignore any RECOMMENDED/SHOULD entry in the specification and doing so does not make us stricter than the spec.

MVrachev:
I am going to address this issue in my pr #1512.

MVrachev:
Fixed by #1512 with commit https://github.com/theupdateframework/tuf/pull/1512/commits/34e7546255a66f056f04e3afb8ed590ba503f5b6

PR #1512, opened 2021-07-29
Implement glob-like pattern matching

### LINKED ISSUE #1505: ngclient: implement glob-like pattern matching

According to the [recently updated version of the specification ](https://github.com/theupdateframework/specification/pull/174)the shell style wildcard matching is glob-like, and therefore a path separator in a path should not be matched by a wildcard in the PATHPATTERN.

**Current behavior**:
A path separator in a path is matched by a wildcard in the PATHPATTERN (`fnmatch.fnmatch`)

**Expected behavior**:
A path separator in a path should not be matched by a wildcard in the PATHPATTERN.

MVrachev:
Fixes #1505, #1506

**Implement glob-like pattern matching - 1505**

According to the recently updated version of the specification the shell
style wildcard matching is glob-like (see https://github.com/theupdateframework/specification/pull/174),
and therefore a path separator in a path should not be matched by a
wildcard in the PATHPATTERN.

That's not what happens with `fnmatch.fnmatch()` which doesn't
see "/" separator as a special symbol.
For example: fnmatch.fnmatch("targets/foo.tgz", "*.tgz") will return
True which is not what glob-like implementation will do.

We should make sure that target_path and the `pathpattern` contain the
same number of directories and because each part of the `pathpattern`
could include a glob pattern we should check that `fnmatch.fnmatch()` is true
on each target and `pathpattern` directory fragment separated by "/".

**Avoid lstrip(os.sep) on target paths - 1506**

For targetpath: we don't want to support corner cases such as
file paths starting with separator.
Why this case should be treated especially than any other case where
you have multiple "/" for example "foo//bar/tar.gz"?

For pathpattern: it's recommended that the separator in the pathpattern
should be "/":
see https://theupdateframework.github.io/specification/latest/#targetpath
I believe it could lead to issues for a client implementation if it
supports arbitrary separators - every implementation needs to choose one
and stick with it.
Then, if we decide that "/" is our separator using lstrip on "os.sep" is
wrong, because the os separator from the server could be different that
the one used in the client.

Because of the above arguments, it makes sense to just remove
lstrip on os separators.

Additionally, document in the public API that we only support "/" as a
separator and don't handle corner cases such as leading separators
in either pathpattern or target_filepath.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1057,6 +1057,24 @@ class DelegatedRole(Role):
             res_dict["path_hash_prefixes"] = self.path_hash_prefixes
         return res_dict
 
+    @staticmethod
+    def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
+        """Determines whether "targetname" matches the "pathpattern"."""
+        # We need to make sure that targetname and pathpattern are pointing to
+        # the same directory as fnmatch doesn't threat "/" as a special symbol.
+        target_parts = targetpath.split("/")
+        pattern_parts = pathpattern.split("/")
+        if len(target_parts) != len(pattern_parts):
+            return False
+
+        # Every part in the pathpattern could include a glob pattern, that's why
+        # each of the target and pathpattern parts should match.
+        for target_dir, pattern_dir in zip(target_parts, pattern_parts):
+            if not fnmatch.fnmatch(target_dir, pattern_dir):
+                return False
+
+        return True
+
     def is_delegated_path(self, target_filepath: str) -> bool:
         """Determines whether the given 'target_filepath' is in one of
         the paths that DelegatedRole is trusted to provide"""
@@ -1079,7 +1097,7 @@ class DelegatedRole(Role):
                 # are also considered matches. Make sure to strip any leading
                 # path separators so that a match is made.
                 # Example: "foo.tgz" should match with "/*.tgz".
-                if fnmatch.fnmatch(
+                if self._is_target_in_pathpattern(
                     target_filepath.lstrip(os.sep), pathpattern.lstrip(os.sep)
                 ):
                     return True
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -19,7 +19,6 @@ import abc
 import fnmatch
 import io
 import logging
-import os
 import tempfile
 from collections import OrderedDict
 from datetime import datetime, timedelta
@@ -1077,7 +1076,14 @@ class DelegatedRole(Role):
 
     def is_delegated_path(self, target_filepath: str) -> bool:
         """Determines whether the given 'target_filepath' is in one of
-        the paths that DelegatedRole is trusted to provide"""
+        the paths that DelegatedRole is trusted to provide.
+
+        The target_filepath and the DelegatedRole paths are expected to be in
+        their canonical forms, so e.g. "a/b" instead of "a//b" . Only "/" is
+        supported as target path separator. Leading separators are not handled
+        as special cases (see `TUF specification on targetpath
+        <https://theupdateframework.github.io/specification/latest/#targetpath>`_).
+        """
 
         if self.path_hash_prefixes is not None:
             # Calculate the hash of the filepath
@@ -1093,13 +1099,8 @@ class DelegatedRole(Role):
         elif self.paths is not None:
             for pathpattern in self.paths:
                 # A delegated role path may be an explicit path or glob
-                # pattern (Unix shell-style wildcards). Explicit filepaths
-                # are also considered matches. Make sure to strip any leading
-                # path separators so that a match is made.
-                # Example: "foo.tgz" should match with "/*.tgz".
-                if self._is_target_in_pathpattern(
-                    target_filepath.lstrip(os.sep), pathpattern.lstrip(os.sep)
-                ):
+                # pattern (Unix shell-style wildcards).
+                if self._is_target_in_pathpattern(target_filepath, pathpattern):
                     return True
 
         return False
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -13,7 +13,6 @@ import os
 import shutil
 import tempfile
 import unittest
-import copy
 
 from datetime import datetime, timedelta
 from dateutil.relativedelta import relativedelta
@@ -40,7 +39,6 @@ from tuf.api.serialization import (
 
 from tuf.api.serialization.json import (
     JSONSerializer,
-    JSONDeserializer,
     CanonicalJSONSerializer
 )
 
@@ -464,6 +462,37 @@ class TestMetadata(unittest.TestCase):
 
         with self.assertRaises(KeyError):
             root.signed.remove_key('root', 'nosuchkey')
+
+    def test_is_target_in_pathpattern(self):
+        supported_use_cases = [
+            ("foo.tgz", "foo.tgz"),
+            ("foo.tgz", "*"),
+            ("foo.tgz", "*.tgz"),
+            ("foo-version-a.tgz", "foo-version-?.tgz"),
+            ("targets/foo.tgz", "targets/*.tgz"),
+            ("foo/bar/zoo/k.tgz", "foo/bar/zoo/*"),
+            ("foo/bar/zoo/k.tgz", "foo/*/zoo/*"),
+            ("foo/bar/zoo/k.tgz", "*/*/*/*"),
+            ("foo/bar", "f?o/bar"),
+            ("foo/bar", "*o/bar"),
+        ]
+        for targetpath, pathpattern in supported_use_cases:
+            self.assertTrue(
+                DelegatedRole._is_target_in_pathpattern(targetpath, pathpattern)
+            )
+
+        invalid_use_cases = [
+            ("targets/foo.tgz", "*.tgz"),
+            ("/foo.tgz", "*.tgz",),
+            ("targets/foo.tgz", "*"),
+            ("foo-version-alpha.tgz", "foo-version-?.tgz"),
+            ("foo//bar", "*/bar"),
+            ("foo/bar", "f?/bar")
+        ]
+        for targetpath, pathpattern in invalid_use_cases:
+            self.assertFalse(
+                DelegatedRole._is_target_in_pathpattern(targetpath, pathpattern)
+            )
 
 
     def test_delegation_class(self):


tuf/ngclient/updater.py
```diff
@@ -527,8 +544,9 @@ def _visit_child_role(child_role: Dict, target_filepath: str) -> str:
             # target without a leading path separator - make sure to strip any
             # leading path separators so that a match is made.
             # Example: "foo.tgz" should match with "/*.tgz".
-            if fnmatch.fnmatch(
-                target_filepath.lstrip(os.sep), child_role_path.lstrip(os.sep)
+            if _is_target_in_pathpattern(
```
sechkova: The two arguments are swapped.

tuf/ngclient/updater.py
```diff
@@ -463,6 +463,23 @@ def _preorder_depth_first_walk(self, target_filepath) -> Dict:
         return {"filepath": target_filepath, "fileinfo": target}
 
 
+def _is_target_in_pathpattern(targetname, pathpattern):
+    """Determines whether the given "targetname" is inside the "pathpattern"."""
+    if pathpattern == "*":
```
sechkova: `"*" `should not match path separators too, meaning it matches only files under the root/main/base directory or whatever we should call it.

tuf/ngclient/updater.py
```diff
@@ -463,6 +463,23 @@ def _preorder_depth_first_walk(self, target_filepath) -> Dict:
...
+        return True
+    pathpattern_dir = os.path.dirname(pathpattern)
+    target_dir = os.path.dirname(targetname)
+    if pathpattern_dir == target_dir:
+        return fnmatch.fnmatch(targetname, pathpattern)
+    # Check that pathpattern_dir contains the target_dir and `*` is used at the
+    # end of the pattern.
+    # For example, targetname "foo/bar/zoo/k.tgz" and pattpatern "foo/bar/*".
+    if target_dir.startswith(pathpattern_dir) and pathpattern[-1] == "*":
```
sechkova: same as above for "*". It should not match '/', right?

tests/test_updater_ng.py
```diff
@@ -148,6 +149,29 @@ def test_refresh_with_only_local_root(self):
     # Get targetinfo for 'file3.txt' listed in the delegated role1
     targetinfo3= self.repository_updater.get_one_valid_targetinfo('file3.txt')
 
+
+  def test_is_target_in_pathpattern(self):
+    supported_use_cases = [
+      ("foo.tgz", "foo.tgz"),
+      ("foo.tgz", "*"),
+      ("targets/foo.tgz", "*"),
```
sechkova: This should fail.

tests/test_updater_ng.py
```diff
@@ -148,6 +149,29 @@ def test_refresh_with_only_local_root(self):
...
+      ("foo.tgz", "*.tgz"),
+      ("foo-version-a.tgz", "foo-version-?.tgz"),
+      ("targets/foo.tgz", "targets/*.tgz"),
+      ("foo/bar/zoo/k.tgz", "foo/bar/zoo/*"),
+      ("foo/bar/zoo/k.tgz", "foo/bar/*")
```
sechkova: Should fail too.

sechkova (commented):
Comparing the dirnames is a clever trick that matches the most common examples in TUF. But if we want to implement proper "globbing" it should be applied on each of the components of a pathname separately.
Should this be a valid test case too  (I don't know if someone would use it :shrug:) : `("foo/bar/zoo/k.tgz", "foo/*/zoo/*"),`?


Also IIUC, '*'  should not match path separator, I've added comments about it.

tuf/ngclient/updater.py
MVrachev: Yes, now when I think about it you are right.
No special symbol allows you to go recursively through directories if we are speaking for glob-like implementation.

tests/test_updater_ng.py
MVrachev: Yes, as you have pointed out with `*` we are considering files only in the current base directory.

tuf/ngclient/updater.py
MVrachev: Thanks, a great catch!

tests/test_updater_ng.py
MVrachev: Yes, as you have pointed out with * we are considering files only in the current base directory.

tuf/ngclient/updater.py
MVrachev: Yes, as you have pointed out with * we are considering files only in the current base directory.

MVrachev:
After the good review by @sechkova, I decided to change my approach.

As Teodora suggested, I will focus on checking the list of directories returned by `os.path.dirname(pathpattern).split("/")` 
and the list of directories returned by `os.path.dirname(targetname).split("/")`.
If both of them match,  I can use `fnmatch.fnmatch()`.
Also, that way I will be able to accept `*` as a directory name  for pathpattern directories and thus this example:
`targetname = "foo/bar/zoo/k.tgz"` and `pathpattern = foo/*/zoo/*` will be supported.

Please let me know if any of the reviewers have ideas for a combination of `targetname` and `pathpattern` that will return a false positive or false negative

MVrachev:
Added type annotations and better function description for `_is_target_in_pathpattern` as suggested by @avelichka.

jku:
This seems to be handling directories as a special case. I don't think you'll be able to make that work: as Teodora said globbing should probably be applied on each of the components of a pathname separately (unless of course the python base modules include a method to handle the whole path). 

Think of e.g. these cases to see new issues:
```
("foo/bar", "f?o/bar"),
("foo/bar", "*o/bar"),
```

Other quick comments:
* let's pay attention to os.path usage: both pathpattern and targetpath are (or should be, I think) url fragments and not local filesystem paths. So if we use a path module do we actually want to use `posixpath` and not `os.path`?
* we probably want to do `posixpath.abspath()` at some point to handle cases like "foo//bar" in input -- but I guess that should happen somewhere closer to the API level, and not in this code

MVrachev:
> This seems to be handling directories as a special case. I don't think you'll be able to make that work: as Teodora said globbing should probably be applied on each of the components of a pathname separately (unless of course the python base modules include a method to handle the whole path).
> 
> Think of e.g. these cases to see new issues:
> 
> ```
> ("foo/bar", "f?o/bar"),
> ("foo/bar", "*o/bar"),
> ```

I added a new commit in which I change my approach.
In the last commit, I call `fnmatch.fnmatch()` on each directory in the `pathpattern`
separately.
Also, I added two unit test cases testing that:
- `("foo/bar", "f?o/bar")` will pass
- `("foo//bar", "*o/bar")` will fail because of the number of slashes.
I remember we discussed before we don't want to artificially support corner cases.

> 
> Other quick comments:
> 
> * let's pay attention to os.path usage: both pathpattern and targetpath are (or should be, I think) url fragments and not local filesystem paths. So if we use a path module do we actually want to use `posixpath` and not `os.path`?
> * we probably want to do `posixpath.abspath()` at some point to handle cases like "foo//bar" in input -- but I guess that should happen somewhere closer to the API level, and not in this code

We had a discussion with @joshuagl about what targets do we expect as arguments
and his opinion was that `pathpattern` and `targetpath` should point in a local directory
(or at least that's how I interpreted his words).

Additionally, I run all of our tests and checked what do we pass to `fnmatch.fnmatch()` inside 
`_visit_child_role`and `_target_matches_path_pattern()` and none of them had test using URLs.
The old code provides an API call `get_valid_targetinfo()` for the `MultiRepoUpdater`
which seems to me that doesn't handle URLs in any specific way.

jku:
> I remember we discussed before we don't want to artificially support corner cases.

If you mean this code expects the "paths" to be canonical representation (so things like "foo//bar" or "foo/../foo/bar" are not supported) that seems correct. Or in other words: I'm fine with requiring well-formed input, I'm cautious about defining things as corner cases :) 

> We had a discussion with @joshuagl about what targets do we expect as arguments
and his opinion was that pathpattern and targetpath should point in a local directory
(or at least that's how I interpreted his words).

The spec is broken about this IMO but I don't think we can or should do that in the implementation -- how could the client know which separator the server is using? The API as it's currently designed should maybe accept filesystem paths as input but I believe the input handling should immediately make the transmogrification to a URL-path so rest of the code can assume everything is URL/posix path... 

Alternatively someone needs to explain to me how we can support random directory separators in all of the code. As an example: if pathpatterns and targetpaths were filesystem paths then you are missing all the windows path tests, and all the tests with half windows paths and half posix paths

tuf/ngclient/updater.py
```diff
@@ -465,30 +465,23 @@ def _preorder_depth_first_walk(self, target_filepath) -> Dict:
...
-            in_the_same_dir = False
-            break
-
-    if in_the_same_dir:
-        return fnmatch.fnmatch(targetname, pathpattern)
+    # Every part in the pathpattern could include a regex, that's why
+    # each of the target parts should match the corresponding pathpattern part.
+    for index, pattern_dir in enumerate(pathpattern_parts):
+        if not fnmatch.fnmatch(target_parts[index], pattern_dir):
+            return False
```
jku: this looks a lot better, thanks. You could use `for target_dir, pattern_dir in zip(target_parts, pathpattern_parts)` to avoid the index access.

Also should not call it a regex because it isn't.

tests/test_updater_ng.py
```diff
@@ -148,6 +149,35 @@ def test_refresh_with_only_local_root(self):
...
+
+    invalid_use_cases = [
+      ("targets/foo.tgz", "*.tgz"),
+      ("/foo.tgz", "*.tgz",),
+      ("targets/foo.tgz", "*"),
+      ("foo/bar/k.tgz", "foo/bar/zoo/*"),
+      ("foo/bar/zoo/k.tgz", "foo/bar/*"),
+      ("foo-version-alpha.tgz", "foo-version-?.tgz"),
+      ("foo//bar", "*o/bar"),
+    ]
```
jku: six cases out of seven actually hit `if len(target_parts) != len(pathpattern_parts)` branch so  fnmatch() only gets called single time in all of this. Please make sure there are failing tests for fnmatch() in the filename part and also in a directory part

tests/test_updater_ng.py
```diff
@@ -148,6 +149,35 @@ def test_refresh_with_only_local_root(self):
...
+      ("foo.tgz", "foo.tgz"),
+      ("foo.tgz", "*"),
+      ("foo.tgz", "*.tgz"),
+      ("foo-version-a.tgz", "foo-version-?.tgz"),
+      ("targets/foo.tgz", "targets/*.tgz"),
+      ("foo/bar/zoo/k.tgz", "foo/bar/zoo/*"),
+      ("foo/bar/zoo/k.tgz", "foo/*/zoo/*"),
+      ("foo/bar/zoo/k.tgz", "*/*/*/*"),
+      ("foo/bar", "f?o/bar"),
+    ]
```
jku: please add a case for successful * glob match in a directory part -- in other words a regression test that would have failed the previous version of this PR: `("foo/bar", "*o/bar"),`

jku (commented):
Maybe makes sense to finish this after #1463 is in but a few more specific comments:
* the matching looks about right now
* we'll maybe still want to remove the possible starting separators from pathpattern somewhere in this code (I left a longer comment in #1506) -- the separator just shouldn't be clients _os.sep_ because that makes no sense for pathpattern that comes from the server

joshuagl:
> We had a discussion with @joshuagl about what targets do we expect as arguments
> and his opinion was that `pathpattern` and `targetpath` should point in a local directory
> (or at least that's how I interpreted his words).

I think I just reiterated what the [spec suggests](https://theupdateframework.github.io/specification/v1.0.20/#targetpath) – that we should expect a relative path using UNIX path separators.

It's possible that we should add an explicit section to the spec, or the oft discussed and frequently desired secondary literature https://github.com/theupdateframework/specification/issues/91, on path handling?

jku (changes requested):
(marking "changes requested" so it's visible in PR list)

tuf/ngclient/updater.py
MVrachev: Thanks for the `zip` suggestion.
Isn't `f?o` a regex? 
Because that's what could be used as a folder name example: `f?o/bar`.

tests/test_updater_ng.py
MVrachev: I removed two of the test cases that seemed more redundant and added an additional one for a wrong directory regex.
Any other suggestions for tests?

tests/test_updater_ng.py
MVrachev: I think the test `("foo/bar", "*o/bar")` tests the same use case as line 162:
`("foo/bar/zoo/k.tgz", "*/*/*/*"),`.
Even in line 162 you make sure you use `fnmatch.fnmatch()` on all 4 directories otherwise, it cannot be successful.

MVrachev:
I had to rebase on top of develop and move `_is_target_in_pathpattern` inside `tuf/api/metadata.py`.
Additionally, I changed some of the tests and added a commit removing the `lstrip` on `os.sep`

In order to fix #1506 do I need to document that we don't use `lsstrip`, because in the issue description it's said:
`Define a valid target file path and pathpattern strings and reject everything else.`?

tests/test_updater_ng.py
jku: I'm just saying: your earlier PR passed all those tests but when I added `"*o/bar"` it did not pass... 

Yes I can see the current implementation should succeed but this is basic regression testing: if you fix a bug, add a test case so the bug won't come back.

tuf/ngclient/updater.py
jku: For TUF it is not a regex, it is a glob pattern

tests/test_updater_ng.py
MVrachev: I just added that test.

MVrachev:
I updated the comments and commit message to use `glob pattern` instead of `regex`.

tuf/ngclient/updater.py
MVrachev: I updated the comments and commit message to use glob pattern instead of regex.

jku (changes requested):
Thanks, functionality looks fine I think and moving the code to Metadata API looks good to me. Two questions/comments:
* should the function be a stand-alone one? I'm not against this sort of design in general but it looks quite out of place in Metadata API: I think this would be the first top-level function in `tuf.api.metadata` so far. As an alternative it could be a private staticmethod on DelegatedRole, expressing that this is not some generic pattern matching function but an implementation detail of DelegatedRole
* commenting feels heavy (are all comments useful, will they stay up-to-date?): the complete path pattern handling in DelegatedRole is now ~8 lines of fairly straight forward code covered by 13 lines of comments. At least the large comment in `is_delegated_path()` needs an update as it's repetitive and already outdated (as no stripping is done).

Please fix the outdated comment. I'll defer to your decision on the other raised issues.

MVrachev:
> Thanks, functionality looks fine I think and moving the code to Metadata API looks good to me. Two questions/comments:
> 
> * should the function be a stand-alone one? I'm not against this sort of design in general but it looks quite out of place in Metadata API: I think this would be the first top-level function in `tuf.api.metadata` so far. As an alternative it could be a private staticmethod on DelegatedRole, expressing that this is not some generic pattern matching function but an implementation detail of DelegatedRole

I agree to make the function a private static method in `DelegatedRole` sounds like the most logical solution.

> * commenting feels heavy (are all comments useful, will they stay up-to-date?): the complete path pattern handling in DelegatedRole is now ~8 lines of fairly straight forward code covered by 13 lines of comments. At least the large comment in `is_delegated_path()` needs an update as it's repetitive and already outdated (as no stripping is done).
> 
> Please fix the outdated comment. I'll defer to your decision on the other raised issues.

I removed the strip comments from `is_delegated_path` as well as another comment from `_is_target_in_pathpattern`  that was non essential I think.

tuf/api/metadata.py
```diff
@@ -1098,10 +1097,7 @@ def is_delegated_path(self, target_filepath: str) -> bool:
         elif self.paths is not None:
             for pathpattern in self.paths:
                 # A delegated role path may be an explicit path or glob
-                # pattern (Unix shell-style wildcards). Explicit filepaths
-                # are also considered matches. Make sure to strip any leading
-                # path separators so that a match is made.
-                # Example: "foo.tgz" should match with "/*.tgz".
+                # pattern (Unix shell-style wildcards).
```
sechkova, line 1102: I don't see any code change in this commit, only comments.

MVrachev: Fixed it.

MVrachev:
@jku do you think the changes I made in https://github.com/theupdateframework/tuf/pull/1512/commits/6ee2daceb36b40d25ebcda39aa9c855a85320a62 are enough to consider #1506 solved?

tuf/api/metadata.py
```diff
@@ -1057,9 +1056,32 @@ def to_dict(self) -> Dict[str, Any]:
...
+        return True
+
     def is_delegated_path(self, target_filepath: str) -> bool:
         """Determines whether the given 'target_filepath' is in one of
-        the paths that DelegatedRole is trusted to provide"""
+        the paths that DelegatedRole is trusted to provide.
+
+        Note: we only support "/" as a separator and don't handle corner cases
+        such as leading separators in pathpattern or target_filepath. See:
+        https://theupdateframework.github.io/specification/latest/#targetpath
```
jku: documentation should maybe not imply that there are _other unspecified_ "corner cases" that we do not handle.

```suggestion
        target_filepath (and the DelegatedRole paths) are expected to be in 
        their canonical forms, so e.g. "a/b" instead of "a//b" . Only
        "/" is supported as target path separator. Leading separators are not
        handled as special cases (see `TUF specification on targetpath
        https://theupdateframework.github.io/specification/latest/#targetpath`_).
```
Maybe that covers most of what we want to say?

Btw, I'm not sure I got the sphinx link format correct there

jku (approved):
LGTM, I left one comment about the docstring but I'll leave decision to you: current one seems ok too.

tuf/api/metadata.py
MVrachev: I used almost the same text as you suggested, but for the link, I had to read the sphinx docs:
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages.
Now, the link looks good.

Issue #1317, opened 2021-03-18
client refactor: consider API changes


jku:
We should consider if we want to  additions/changes to the API -- this would be a good time for them.

Possibilities:

**1. Namespace shortening**: currently using the client looks like this:
```
from tuf.client import updater
updater = updater.Updater() # that looks stupid and is a name clash
```
This is a small thing maybe but having both client **and** updater module does not seem useful: I think we could just as well have
```
from tuf import client
updater = client.Updater()
```
This should be possible by having a `client/__init__.py` that exports Updater (and FetcherInterface).

**2. Mirrors**: this is arguably the most complex part of the client API, probably mostly unused and is kind of hidden: I wonder if we can/should make it better? pip does need the mirrors config but even that use is not what mirrors config was designed for: pip potentially changes the config before every single new target download, switching the mirror list based on whether it's downloading index files or distribution files... 
#1307, #1143

**3. the actual updater API** do we want to improve  the _get_one_valid_targetinfo + updated_targets + download_target_  call sequence in any way? 
    * Possibly provide a helper that does the obvious thing (given a target name, just make sure it is downloaded)
    * should the actual local paths be returned? It feels odd that client has to build the paths when updater already had them

joshuagl:
1. I'm 💯 for the namespace shortening.
2. I do want us to think about the mirrors, it's a popular topic #1307 and #1143.
3. I like both of these suggestions, we also have https://github.com/theupdateframework/tuf/issues/1135 has some additional thoughts on client API

mnm678:
I'm not sure if we should design for not-yet-accepted TAPs (we probably shouldn't), but there might be a way to combine the mirrors functionality with the client configuration in TAP 13 (#1103) and possibly also TAP 4 into a more general 'configurations' API, where the client could set these options before using `client.Updater`.

jku:
Another thing I've not mentioned: currently it's not possible to download without writing to a file. Usually that's fine but the pip index files might be a use case that would benefit from doing that: the code currently opens and read()s the file immediately  after download succeeds...

jku:
So number 1 is handled in #1397 in coming days, item 2 was solved by not supporting mirrors. 

I have some more specific ideas about item 3:
* the current API is needlessly complex with a self defined dict returned from get_on_valid_targeinfo and expected as argument in otherfunctions : we should probably just return a TargetFile object from `get_one_valid_targetinfo()` or whatever we call it
[...]

PR #1514, opened 2021-08-03
Metadata API: include target target name in TargetFile

MVrachev:
Related to #1411, but it's not yet clear if it closes it. 

Currently, TargetFile instances do not contain the filename of the file
they represent. The API itself does not need it but it could be useful
for users of the API.

As an example, the current client returns a dict for
`get_one_valid_targetinfo()`: that dict contains a filepath field and
a targetinfo field (essentially TargetFile).
We would like to keep a similar API, but avoid hand-crafted dicts.
It would be much nicer to return a TargetFile that would contain the
full "metadata" of the targetfile.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1254,6 +1254,6 @@ class Targets(Signed):
         return targets_dict
 
     # Modification.
-    def update(self, filename: str, fileinfo: TargetFile) -> None:
+    def update(self, fileinfo: TargetFile) -> None:
         """Assigns passed target file info to meta dict."""
-        self.targets[filename] = fileinfo
+        self.targets[fileinfo.path] = fileinfo
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1138,6 +1138,8 @@ class TargetFile(BaseFile):
     Attributes:
         length: An integer indicating the length of the target file.
         hashes: A dictionary of hash algorithm names to hash values.
+        path: A string denoting the path to a target file relative to a base
+            URL of targets.
         unrecognized_fields: Dictionary of all unrecognized fields.
     """
 
@@ -1145,6 +1147,7 @@ class TargetFile(BaseFile):
         self,
         length: int,
         hashes: Dict[str, str],
+        path: str,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
 
@@ -1153,6 +1156,7 @@ class TargetFile(BaseFile):
 
         self.length = length
         self.hashes = hashes
+        self.path = path
         self.unrecognized_fields = unrecognized_fields or {}
 
     @property
@@ -1160,13 +1164,13 @@ class TargetFile(BaseFile):
         return self.unrecognized_fields.get("custom", None)
 
     @classmethod
-    def from_dict(cls, target_dict: Dict[str, Any]) -> "TargetFile":
+    def from_dict(cls, target_dict: Dict[str, Any], path: str) -> "TargetFile":
         """Creates TargetFile object from its dict representation."""
         length = target_dict.pop("length")
         hashes = target_dict.pop("hashes")
 
         # All fields left in the target_dict are unrecognized.
-        return cls(length, hashes, target_dict)
+        return cls(length, hashes, path, target_dict)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the JSON-serializable dictionary representation of self."""
@@ -1232,7 +1236,9 @@ class Targets(Signed):
             delegations = Delegations.from_dict(delegations_dict)
         res_targets = {}
         for target_path, target_info in targets.items():
-            res_targets[target_path] = TargetFile.from_dict(target_info)
+            res_targets[target_path] = TargetFile.from_dict(
+                target_info, target_path
+            )
         # All fields left in the targets_dict are unrecognized.
         return cls(*common_args, res_targets, delegations, signed_dict)
 
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -60,13 +60,13 @@ Example::
 
 import logging
 import os
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import List, Optional, Set, Tuple
 from urllib import parse
 
 from securesystemslib import util as sslib_util
 
 from tuf import exceptions
-from tuf.api.metadata import Targets
+from tuf.api.metadata import TargetFile, Targets
 from tuf.ngclient._internal import requests_fetcher, trusted_metadata_set
 from tuf.ngclient.config import UpdaterConfig
 from tuf.ngclient.fetcher import FetcherInterface
@@ -144,8 +144,8 @@ class Updater:
 
     def get_one_valid_targetinfo(
         self, target_path: str
-    ) -> Optional[Dict[str, Any]]:
-        """Returns target information for 'target_path'.
+    ) -> Optional[TargetFile]:
+        """Returns TargetFile instance with information for 'target_path'.
 
         The return value can be used as an argument to
         :func:`download_target()` and :func:`updated_targets()`.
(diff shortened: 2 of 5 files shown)


tuf/ngclient/updater.py
```diff
@@ -460,7 +459,8 @@ def _preorder_depth_first_walk(self, target_filepath) -> Dict:
                 self.config.max_delegations,
             )
 
-        return {"filepath": target_filepath, "fileinfo": target}
+        target.targetname = target_filepath
```
sechkova: This doesn't feel right. This is not the place to set "targetname". It should be done during the instantiation of a Targets object.

tuf/api/metadata.py
```diff
@@ -1085,13 +1085,15 @@ class TargetFile(BaseFile):
...
         hashes: A dictionary of hash algorithm names to hash values.
+        targetname: An optional string denoting the target file name.
         unrecognized_fields: Dictionary of all unrecognized fields.
     """
 
     def __init__(
         self,
         length: int,
         hashes: Dict[str, str],
+        targetname: Optional[str] = None,
```
sechkova: I'm not a fan of`TargetFile.targetname`. How about `TargetFile.filename`, `TargetFile.path` or `TargetFile.targetpath` if we insist on being verbose and stick to the spec.

jku: (this is just a drive by comment, will have to think about the justification for thw whole thing a bit still... but thanks for starting this, I think it's useful to make discussion more concrete)

What teodora said,  let's not use even more different names for the same thing... I think "targetpath" is used for this already so the attribute should probably be targetpath or path.

I believe it should not be optional though. Or is there a use case of a TargetFile without a targetpath?

MVrachev: > I'm not a fan of TargetFile.targetname. How about TargetFile.filename, TargetFile.path or TargetFile.targetpath if we insist on being verbose and stick to the spec.

I had a misconception and didn't want to call it `targetpath` thinking it's not an actual path, but a file name.
Now, I read the spec and it clearly this is called `TARGETPATH`:
https://theupdateframework.github.io/specification/latest/#targetpath.

So, from both options to call it `targetpath` or `path` I would prefer `path` given it's shorter.

> I believe it should not be optional though. Or is there a use case of a TargetFile without a targetpath?

I wasn't sure when I added it as optional. I did that because it's not required by the spec, but we can use `TargetFile`
in such a way that we would always pass the path.

So, in conclusion, I will rename the attribute to `path` and make it non-optional.

tuf/ngclient/updater.py
MVrachev: I had to do that before when `targetname` was an optional attribute.
Now, when it's mandatory there is no need to do such a thing.

jku:
I think this makes sense and implementation looks fine. You will have to do something about `Targets.update()` though (I still think removing it is correct -- at least currently it seems to have no purpose other than creating apparent complexity -- but I guess fixing it to match the new API is an option as well)

Maybe it would be good to document target path a little more: that it is a "path-relative-URL string" that will be used as part of the final download url to fetch the target

The high level question is whether anyone objects to making objects like this a bit smarter (in this case be aware of the path) even if the file format does not include that. I think this is a good way forward and seems immediately understandable by the reader -- it definitely makes the Updater API much better...

jku:
For those not following that closely: this PR changes TargetFile interface (which on it's own is a minor change in the Metadata API implementation) but also the whole ngclient API so that TargetFile is returned and expected as argument instead of an undefined dictionary.

I think this is going in the right direction and allows us to continue with something like https://github.com/theupdateframework/tuf/issues/1317#issuecomment-843872644: I believe the core Updater should not handle caching: this would make it much clearer what is actually part of the spec implementation and what is an additional thing we just wanted to implement in the same source directory

joshuagl:
> The high level question is whether anyone objects to making objects like this a bit smarter (in this case be aware of the path) even if the file format does not include that. I think this is a good way forward and seems immediately understandable by the reader -- it definitely makes the Updater API much better...

I do not object, in fact looking at it now having an object which describes a file without being able to explicitly identify the file does look odd. +1 from me.

jku:
@MVrachev can you suggest a solution for `Targets.update()`: at the very least the function arguments need to be updated.

jku (changes requested):
marking changes requested for `Targets.update()` issue

MVrachev:
I had to rebase to fix merge conflicts after the simplification of `_preorder_depth_first_walk`.
I updated the `Targets.update()` argument accordingly.
I will prefer if other changes or removal of the update method are done for all classes when resolving #1230.

tuf/ngclient/updater.py
```diff
@@ -252,27 +249,26 @@ def download_target(
         else:
             target_base_url = _ensure_trailing_slash(target_base_url)
 
-        target_fileinfo: "TargetFile" = targetinfo["fileinfo"]
-        target_filepath = targetinfo["filepath"]
+        target_filepath = targetinfo.path
```
jku, line 252: nitpick but this variable seems unneeded

tuf/ngclient/updater.py
```diff
@@ -414,7 +410,8 @@ def _preorder_depth_first_walk(
 
             if target is not None:
                 logger.debug("Found target in current role %s", role_name)
-                return {"filepath": target_filepath, "fileinfo": target}
+                target.targetname = target_filepath
```
jku: I don't understand this. Why is the updater modifying a TargetFile?

tuf/api/metadata.py
```diff
@@ -1138,13 +1138,15 @@ class TargetFile(BaseFile):
     Attributes:
         length: An integer indicating the length of the target file.
         hashes: A dictionary of hash algorithm names to hash values.
+        path: A string denoting the target file path.
```
jku: I'd like to not call it a _file path_ since at least for me it creates the impression that we are talking about local filesystem paths: target path is really a url fragment and is used to form the target download URL. Our updater does end up caching it locally using the target path as a file path, but that's an implementation detail IMO.

Maybe we could mention "path-relative-URL string" like https://theupdateframework.github.io/specification/latest/#targetpath does? not sure it's any more descriptive but it would avoid creating the impression of local path?

jku (commented):
Looks nice. I left a couple of comments and a question.

tuf/ngclient/updater.py
MVrachev: If we remove target_filepath we would have to define `full_url` before the `if` statement so it can be used
with and without `consistent_snapshots` and `prefix_targets_with_hash`.
It will look something like that
```python
        full_url = parse.urljoin(target_base_url, targetinfo.path)
        consistent_snapshot = self._trusted_set.root.signed.consistent_snapshot
        if consistent_snapshot and self.config.prefix_targets_with_hash:
            hashes = list(targetinfo.hashes.values())
            full_url = parse.urljoin(
                target_base_url, f"{hashes[0]}.{targetinfo.path}"
            )
```

Also, when raising the exception on line 265 we use the `target_filepath` which could be with HASH prefix or without and the correct target name will be saved in `target_filepath`.

tuf/ngclient/updater.py
MVrachev: You are right, that's useless.
Totally forgot that the TargetFile instance itself will have the correct target path already.
PS: I noticed I haven't changed `targetname` to `path` after we decide to rename it and no test had failed.
We really need tests for that functionality...

tuf/ngclient/updater.py
jku: thanks, makes sense

tuf/ngclient/updater.py
jku: it can't be a real test (as defining new attributes like this is legal in python) but _maybe_ mypy has an option to warn about defining object attributes outside init() :shrug:

tuf/api/metadata.py
```diff
@@ -1138,13 +1138,16 @@ class TargetFile(BaseFile):
     Attributes:
         length: An integer indicating the length of the target file.
         hashes: A dictionary of hash algorithm names to hash values.
+        path: A string denoting the path to a target file relative to a base
+            URL of targes.
```
jku: typo of targets

jku (approved):

PR #1517, opened 2021-08-09
Enable docs build with sphinx

### LINKED ISSUE #1482: Docs: Enable continuous documentation build
We want to be able to build and publish documentation (API docs for Metadata API and ngclient at least) as part of our continuous integration.

This bug is about choosing the tools, integrating with the TUF build system and filing any remaining issues as bugs.

Some notes:
* Sphinx seems the "obvious" choice as the docs build tool
* Using a service like ReadTheDocs also seems like a good choice (but integrating that can be a different issue)
* the default sphinx setup generates a lot of files/directories -- we might not want all of it (examples include Makefile and make.bat in root dir: these are not really required. Even the configuration file probably does not need to be in root dir)
* Need to decide what to do with the existing .rst and .md documentation in docs/ -- do we want to include some/all of that? We could start with just the API docs

I've done a quick test and sphinx mostly works:
* `pip install sphinx` (also `sphinx-rtd-theme` to see default RTD style)
* create a configuration file (see `sphinx-quickstart`)
* Add `extensions = ['sphinx.ext.napoleon']` to conf.py to get support for google style docstrings and `html_theme = 'sphinx_rtd_theme'` if using RTD theme
* `make html`

The only major issue I've seen is that `__init__()` docstring is not included (this seems to be a sphinx peculiarity: `autoclass_content = 'both'` might work)

sechkova:
Fixes #1482

- Adds .rst source files for building documentation with `sphinx`. The two mandatory files are `conf.py` containing the build configuration and the master doc file `index.rst`.
- Sphinx uses `autodoc` to automatically include docstrings. `autodoc` imports the modules and needs TUF installed in the environment.
- Adds a tox environment for the docs build
- Adds a CI build 

The following command will generate the documentation from the source files in an html format:
`tox -e docs`
or it can be build manually in an environment with TUF installed with: 
`sphinx-build -b html docs/sphinx/source docs/sphinx/build/html`

Configuring a build with `sphinx` is enough for the documentation to be imported built and published on `readthedocs`.
This would require a `readthedocs` account for the project which can be further configured from the UI and a `.readthedocs.yaml` file.
This is a demo project that I have created from my fork and this PR's branch:
https://tuf-sechkova.readthedocs.io/en/latest/

requirements**:


jku:
> * Adds a CI build

What's the advantage? Can the build fail somehow?

sechkova:
> > * Adds a CI build
> 
> What's the advantage? Can the build fail somehow?

It can fail in case of wrong configuration (cannot resolve paths to modules, dependencies etc.) 
I was actually expecting the build to fail in case of wrong directives or syntax but it only outputs warnings by default. So we can set a flag (- W) to treat warnings as errors unless this feels too strict? Docs are generated anyway in case of warnings, they need to be checked visually if something wasn't rendered correctly.

sechkova:
After discussing with @jku, I created this readthedocs project linked to TUF's github page and added us both as maintainers: https://theupdateframework.readthedocs.io/en/latest/
Next steps for publishing the documentation would be:
- organization-level access needed by readthedocs OAuth App (which I requested through github, so someone already has a mail in their inbox) .
- configuring the project (.readthedocs.yaml to replace the UI settings). c65b299
- adding additional maintainers

docs/sphinx/source/index.rst
```diff
@@ -0,0 +1,19 @@
...
+update systems. It is written in Python and intended to conform to version 1.0 of the
+`TUF specification <https://theupdateframework.github.io/specification/latest/>`_.
+
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+   
+   modern-implementation
+   legacy-implementation
```
jku: I suggest we call "modern-implementation" just "API reference" (with the expectation that we will add "Tutorials", "Examples" etc to this list soon) and then we have an additional explanation about the legacy code: I think that explanation can be right here on this page, I see no reason to have multiple single paragraph pages.

docs/sphinx/source/index.rst
```diff
@@ -0,0 +1,19 @@
+Welcome to TUF's documentation!
+===============================
+
+This is the documentation of the **reference implementation** of
```
jku:
```suggestion
This is the documentation of the **reference Python implementation** of
```

docs/sphinx/source/index.rst
```diff
@@ -0,0 +1,19 @@
...
+`The Update Framework (TUF) <https://theupdateframework.github.io/>`_.
+
+The reference implementation strives to be a readable guide and demonstration
+for those working on implementing TUF in their own languages, environments, or
+update systems. It is written in Python and intended to conform to version 1.0 of the
+`TUF specification <https://theupdateframework.github.io/specification/latest/>`_.
```
jku:
```suggestion
The reference implementation strives to be a safe and easy to use library as
well as a readable guide and demonstration for those working on implementing
TUF in their own languages, environments, or update systems. It is written in
Python and intended to conform to version 1.0 of the
`TUF specification <https://theupdateframework.github.io/specification/latest/>`_.
```

docs/sphinx/source/modern-implementation.rst
```diff
@@ -0,0 +1,31 @@
+Modern implementation
```
jku:
```suggestion
API refererence
```

I would remove all mentions of how we got here (refactoring, PyPI) on this page and concentrate on the API we have

docs/sphinx/source/tuf.api.rst
```diff
@@ -0,0 +1,10 @@
+Metadata API
+===============
+
+.. toctree::
+   :maxdepth: 4
+
+   tuf.api.serialization
+   tuf.api.metadata
```
jku:
```suggestion
   tuf.api.metadata
   tuf.api.serialization
```

docs/sphinx/source/tuf.ngclient.rst
```diff
@@ -0,0 +1,10 @@
+Client API
+==========
```
jku: This page seems to have very little value as is... I wonder if we should add a short explanation here already or just remove this and direct people to updater directly?

jku: Also, if you name this "Client API" then it looks like the links will be called that as well -- so the modern-implementation page will now talk about a "Client API module"...

jku (commented):
Reviewing documentation like this is really difficult... I've left some comments now,  not sure how useful.

I think the build changes look ok, but I'm not convinced the CI addition is useful: as you said the doc changes almost always need a human reviewer, it's not really possible to "test them in CI" -- I'd love to get docs generated per PR so they _could_ be reviewed but that's probably asking a lot.

jku:
Spent a moment thinking about the top-level documentation. I think as a short/mid term goal our top level page should have these links:

 * API Reference (this PR mostly)
 * Installation (use current INSTALLATION.rst?)
 * Getting Started (a document that uses example code from #1518, similar to current QUICKSTART.md)
 * Contribution Guidelines (use current CONTRIBUTORS.rst?)

joshuagl:
> * organization-level access needed by readthedocs OAuth App (which I requested through github, so someone already has a mail in their inbox) .

Granted.

docs/sphinx/source/index.rst
joshuagl: I suggest we do not include the legacy implementation. We do not plan to support it.

jku: yes exactly, I skipped over that bit in my brain :)

jku:
this is still missing the `__init__()` docs (this is some sphinx thing where it thinks classes only have class or init documentation). `autoclass_content = 'both'` may work

jku:
I've worked on this branch at https://github.com/jku/tuf/tree/configure-sphinx:
* some content improvements
* moved the sphinx root to docs/ (to be able to use the existing docs in the published ones)
* included  "Installation" and "Instructions for Contributors" in the sphinx documentation.
* mostly removed mentions of "modern implementation" and "legacy implementation": just talk about the API we have

I plan on having a final look through the content and then making a PR (or if teodora is back by then letting her merge my changes to this branch)

sechkova:
Force pushed the additional commits from @jku's branch + the CI step commit is now removed.

sechkova:
Rebased to get the pylint fix from develop.

jku (approved):
LGTM but since half of it is my work maybe someone else should approve as well :)

Let's try to get the readthedocs PR build working before merging (so it would be easier to review the work), but in case that doesn't happen building locally is easy:
`tox -e docs`
docs/build/html now contains the html docs

EDIT: readthedocs build now works: see list of checks below

.readthedocs.yaml
```diff
@@ -0,0 +1,20 @@
...
+sphinx:
+   builder: html
+   configuration: docs/conf.py
+
+# Optionally build your docs in additional formats such as PDF
+formats: []
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+   version: "3.8"
```
joshuagl: This is higher than our projects minimum Python version, but I don't know if that is a concern?

docs/INSTALLATION.rst
```diff
@@ -12,19 +12,13 @@ is available on our `website
...
-Assuming you trust `the maintainer's PGP key <MAINTAINERS.txt>`_, the detached
-ASC signature can be downloaded and verified.  For example:
-
-::
+Assuming you trust `the maintainer's PGP key
+<https://github.com/theupdateframework/tuf/blob/develop/docs/MAINTAINERS.txt>`_,
+the detached ASC signature can be downloaded and verified.  For example::
 
    $ gpg --verify securesystemslib-0.10.8.tar.gz.asc
    gpg: assuming signed data in 'securesystemslib-0.10.8.tar.gz'
```
joshuagl, line 24: While we are editing this document, should we change this example to be a tuf release and an active maintainer?

docs/api/api-reference.rst
```diff
@@ -0,0 +1,33 @@
+API Reference
+=====================
+
+
+TUF provides multiple APIs:
+
+
+* The low-level :doc:`tuf.api` provides access to a Metadata file abstraction
+  that closely follows the TUF specification and the file format: This API
```
joshuagl: We might make this more concrete by linking to the part of the specification the Metadata API models here?
```suggestion
  that closely follows the TUF specification's [document formats](https://theupdateframework.github.io/specification/latest/#document-formats): This API
```

docs/api/api-reference.rst
```diff
@@ -0,0 +1,33 @@
...
+  handles de/serialization to and from files and makes it easier to access
+  and modify metadata content safely. It is purely focused on individual
+  pieces of Metadata and provides no concepts like "repository" or "update
+  workflow".
+
+* The client update workflow is implemented in the :doc:`tuf.ngclient` module: It is
```
joshuagl: Similarly here, map directly to the spec?
```suggestion
* The [client update workflow](https://theupdateframework.github.io/specification/latest/#detailed-client-workflow) is implemented in the :doc:`tuf.ngclient` module: It is
```

docs/api/tuf.api.rst
```diff
@@ -0,0 +1,13 @@
+Metadata API
+===============
+
+The low-level Metadata API contains two modules:
+
+* :doc:`tuf.api.metadata` contains the actual Metadata abstraction that higher level libraries and application code should use to interact with TUF metadata. This abstraction provides safe reading and writing to supported file formats and helper functions for accessing and modifying the metadata contents.
```
joshuagl: This is a _very_ long line. Other rst files in the PR are wrapping.

docs/api/tuf.ngclient.fetcher.rst
```diff
@@ -0,0 +1,10 @@
+Fetcher
+============
+
+.. automodule:: tuf.ngclient.fetcher
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
```
joshuagl: Nit: Only one blank line at the end of a file.

docs/api/tuf.ngclient.rst
```diff
@@ -0,0 +1,17 @@
+ngclient
+========
+
+The ngclient module contains a complete TUF client library implementation.
+
+* :doc:`tuf.ngclient.updater` implements the client update workflow
```
joshuagl: More spec links?
```suggestion
* :doc:`tuf.ngclient.updater` implements the [detailed client workflow](https://theupdateframework.github.io/specification/latest/#detailed-client-workflow)
```

Can we create a reference so that we define a link to the spec only once, that might make this more maintainable?

i.e. in Markdown I can do [link definitions](https://daringfireball.net/projects/markdown/syntax#link):
```Markdown
[detailed client workflow][]

[detailed client workflow]: https://theupdateframework.github.io/specification/latest/#detailed-client-workflow
```

docs/api/tuf.ngclient.rst
```diff
@@ -0,0 +1,17 @@
...
+* :doc:`tuf.ngclient.config` provides optional configuration for the updater
+* :doc:`tuf.ngclient.fetcher` can be used for optional low-level network I/O control
```
joshuagl, line 8: should we mention the default fetcher implementation and be more explicit about when users might want to implement this interface?

joshuagl (approved):
This is a really great start, nice work @sechkova and @jku.

I left several comments in-line. I don't mind if we address these here or in a future PR, but please update the comments here to indicate which.

docs/api/tuf.ngclient.rst
jku: at least within one document this should work

```
See `Client workflow`_.

.. _Client workflow: https://theupdateframework.github.io/specification/latest/#detailed-client-workflow
```

docs/INSTALLATION.rst
jku: this one is from last release
```
$ gpg --verify tuf-0.17.0.tar.gz.asc 
gpg: assuming signed data in 'tuf-0.17.0.tar.gz'
gpg: Signature made Thu 25 Feb 2021 13:42:50 EET
gpg:                using RSA key 08F3409FCF71D87E30FBD3C21671F65CB74832A4
gpg: Good signature from "Joshua Lock (GPG on YubiKey) <jlock@vmware.com>" [unknown]
```
it does unfortunately continue with 
```
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 08F3 409F CF71 D87E 30FB  D3C2 1671 F65C B748 32A4
```
even though I fully trust your key. I hate gpg

.readthedocs.yaml
jku: this is the python version readthedocs build process will use when running sphinx-build ... so anything recent is ok. 

It would make sense to keep in sync with the one `tox -e docs` uses -- but currently we don't set a default python for tox I think. In that light maybe not setting the version here either would be best?

joshuagl: Yes, the Python version used by Tox isn't set explicitly. We use the version on the host and run our CI in GitHub runners with specific versions installed. Not setting here makes sense in that context, I think.

docs/INSTALLATION.rst
joshuagl: 😩

This was probably scope creep for this PR anyway.

docs/api/tuf.ngclient.rst
joshuagl: Awesome, let's do that?

.readthedocs.yaml
sechkova: Seems like by default read the docs uses the biggest version from this list:
https://docs.readthedocs.io/en/stable/config-file/v2.html#build-image meaning 3.7 and 3.8 for stable and latest
_edit: seems to be 3.7 for both_

sechkova: It is removed: 69c2270

docs/api/tuf.ngclient.rst
sechkova: I added links to the specification as you suggested. I've put them as references though I don't know if it makes sense since I couldn't figure out a way to cross reference them between docs. At least maybe it's more readable like that.

docs/api/tuf.ngclient.rst
sechkova: I will add this to another issue for further improving the documentation -> #1543

sechkova:
Closing to test RTD webhooks.

jku:
Let's see what happens :crossed_fingers:

PR #1519, opened 2021-08-10
Remove max_length parameter from fetch

### LINKED ISSUE #1502: ngclient requestsfetcher result when length > max_length
If RequestsFetcher downloads a URL and the content is longer than max_length, RequestFetcher does not raise but returns max_length bytes.

This is not a fatal issue but leads to a strange situation where the `TargetFile.verify_length_and_hashes()` actually succeeds the length check (because it got exactly max_length bytes) and only fails the hashes check.

I think RequestsFetcher should either

1. raise when the content is longer than max_length or
2. not try to return exactly max_length so that Fetcher itself can raise if length is incorrect. This would mean
   * RequestsFetcher: not doing the `read_amount = min(self.chunk_size, max_length - bytes_received)` thing
   * RequestsFetcher: not checking `if bytes_received >= max_length` in the end of the loop
   * Fetcher: moving the DownloadLengthMismatchError check in download_file() so it happens inside the loop

I only a had a quick look but 2 seems like the better solution: less code, more correct implementation, and only a maximum of chunk_size more downloading if the length is incorrect?

sechkova:
Fixes #1502

If a file has a bigger size than expected, RequestFetcher.fetch
downloads it up to max_length without any errors. Only the
consecutive hash check raises exception. A better behavior
in such case would be to raise a DownloadLengthMismatchError.

For this reason this PR:
- removes max_length from the fetch() definition,
- modifies download_file to check the downloaded length after
  each chunk and raise an error accordingly.

requirements**:

### DIFF
--- a/tuf/ngclient/fetcher.py
+++ b/tuf/ngclient/fetcher.py
@@ -29,15 +29,11 @@ class FetcherInterface:
     __metaclass__ = abc.ABCMeta
 
     @abc.abstractmethod
-    def fetch(self, url: str, max_length: int) -> Iterator[bytes]:
+    def fetch(self, url: str) -> Iterator[bytes]:
         """Fetches the contents of HTTP/HTTPS url from a remote server.
-
-        Ensures the length of the downloaded data is up to 'max_length'.
 
         Arguments:
             url: A URL string that represents a file location.
-            max_length: An integer value representing the maximum
-                number of bytes to be downloaded.
 
         Raises:
             tuf.exceptions.SlowRetrievalError: A timeout occurs while receiving
@@ -77,14 +73,22 @@ class FetcherInterface:
         number_of_bytes_received = 0
 
         with tempfile.TemporaryFile() as temp_file:
-            chunks = self.fetch(url, max_length)
+            chunks = self.fetch(url)
             for chunk in chunks:
-                temp_file.write(chunk)
                 number_of_bytes_received += len(chunk)
-            if number_of_bytes_received > max_length:
-                raise exceptions.DownloadLengthMismatchError(
-                    max_length, number_of_bytes_received
-                )
+                if number_of_bytes_received > max_length:
+                    raise exceptions.DownloadLengthMismatchError(
+                        max_length, number_of_bytes_received
+                    )
+
+                temp_file.write(chunk)
+
+            logger.debug(
+                "Downloaded %d out of %d bytes",
+                number_of_bytes_received,
+                max_length,
+            )
+
             temp_file.seek(0)
             yield temp_file
 
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python
+
+# Copyright 2021, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Unit test for RequestsFetcher.
+"""
+
+import io
+import logging
+import os
+import sys
+import unittest
+import tempfile
+import math
+
+from tests import utils
+from tuf import exceptions, unittest_toolbox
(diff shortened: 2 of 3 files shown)


tuf/ngclient/fetcher.py
```diff
@@ -77,14 +73,21 @@ def download_file(self, url: str, max_length: int) -> Iterator[IO]:
...
+                    raise exceptions.DownloadLengthMismatchError(
+                        max_length, number_of_bytes_received
+                    )
+
+                temp_file.write(chunk)
+                logger.debug(
+                    "Downloaded %d out of %d bytes",
+                    number_of_bytes_received,
+                    max_length,
                 )
```
jku: is this now logged for every chunk on purpose?

jku (commented):
This breaks FetcherInterface API but the result looks like the API we should have made in the first place: it's a cleaner separation of duties and simplifies the fetcher implementation. LGTM in general.

Is it possible to make a test for the case you are fixing?

jku:
> Is it possible to make a test for the case you are fixing?

It would make sense to add tests for ngclient fetcher: not just the ones that already exist for fetcher in legacy code but for download_bytes() and download_file() -- where testing this issue would make sense

tuf/ngclient/fetcher.py
sechkova: Not really ...

sechkova:
Added some tests, mostly copied from test_fetcher.py but modified for the implementation. There's some duplication, maybe unneeded ... 
Do we want to open another issue for testing the networking side of RequestsFetcher? I suppose we do.

jku:
> Do we want to open another issue for testing the networking side of RequestsFetcher?

What do you mean specifically -- this PR seems to test the complete RequestsFetcher surface? Obviously more tests doesn't hurt but it looks like coverage wise we're ok.

sechkova:
> What do you mean specifically -- this PR seems to test the complete RequestsFetcher surface? Obviously more tests doesn't hurt but it looks like coverage wise we're ok.

We still miss tests for `urllib3.exceptions.ReadTimeoutError` (and `sleep` but we want to remove it anyway) at least from the coverage report ... I should just add it to this PR

sechkova:
That's what I could do with the available test utils: 0f19815

jku:
This looks fine otherwise but I refuse to accept a test that takes 5 seconds to run. I can accept either not testing slow retrieval here or making the test runtime more reasonable

jku (changes requested):
This looks fine otherwise but I refuse to accept a test that takes 5 seconds to run. I can accept either not testing slow retrieval here or making the test runtime more reasonable

jku:
Actually I do have a question: is it necessary to setup and teardown the server for each individual test?

sechkova:
> Actually I do have a question: is it necessary to setup and teardown the server for each individual test?

Hmm ... no, I don't think so ... let me fix it.

sechkova:
> This looks fine otherwise but I refuse to accept a test that takes 5 seconds to run. I can accept either not testing slow retrieval here or making the test runtime more reasonable

I was kind of expecting it 😁 I'll remove the test and open an issue so that we do it properly (maybe tweak the current `slow_retreival_server.py`)

Edit: my mistake, the timeout is a socket parameter, so tweak the socket timeouts: https://github.com/theupdateframework/tuf/issues/1552

sechkova:
So test commits are squashed in a single one 35ef056:
- server creation is moved to `setUpClass`
- opened a separate issue for testing the read timeout #1552

I think the test setup can be further improved since the temp data file can be created once per class too but currently `Modified_TestCase` overrides `setUp` only and does not work in class methods.

jku (approved):
I think the Fetcher change is correct, and the tests look like a good basic set of tests for Fetcher/RequestsFetcher.

Both of the last two test changes independently improve the test runtime about 10x: that looks ok now.

PR #1520, opened 2021-08-10
 Add Key.from_securesystemslib_key

### LINKED ISSUE #1458: Metadata API: Add Key.from_securesystemslib_key()?
I've been playing with repository code / metadata creation and this is a mistake that's bit too easy to make:

```python
key = Key.from_dict(sslib_key["keyid"], sslib_key) # key for public TUF metadata
signer = SSlibSigner(sslib_key) # signing key
```

This works just fine... and results in public metadata that includes your private key (TUF key doesn't handle the private key of course but we allow unrecognised fields in the format). Oops.

We should probably have a function for this use case: The securesystemslib key format might be opaque to us (can't be 100% sure what's in there) but I think in this case making some assumptions about that is better than allowing the above to happen

So something like 
```python
from_securesystemslib_key(key_dict: Dict[str, Any]):
    return Key(
        keydict["keyid"],
        keydict["keytype"],
        keydict["scheme"],
        {"public": keydict["keyval"]["public"]}
    )
```
?

avelichka:
The securesystemslib key dictionary representation includes
the private key in keyval. TUF key doesn't handle it in any way,
but considering that we allow unrecognized symbols in the format,
we should exclude the private key.
A call to securesystemslib.keys.format_keyval_to_metadata
with the default private=False would do exactly that.

that is provided, the better.

Fixes #1458

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -54,6 +54,10 @@ from securesystemslib.signer import (
     Signature
 )
 
+from securesystemslib.keys import (
+    generate_ed25519_key
+)
+
 logger = logging.getLogger(__name__)
 
 
@@ -419,6 +423,14 @@ class TestMetadata(unittest.TestCase):
         # threshold of 2 keys
         snapshot.sign(SSlibSigner(self.keystore['timestamp']), append=True)
         root.verify_delegate('snapshot', snapshot)
+
+
+    def test_key_class(self):
+        # Test if from_securesystemslib_key removes the private key from keyval
+        # of a securesystemslib key dictionary.
+        sslib_key = generate_ed25519_key()
+        key = Key.from_securesystemslib_key(sslib_key)
+        self.assertFalse('private' in key.keyval.keys())
 
 
     def test_metadata_root(self):
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -562,6 +562,24 @@ class Key:
             "keyval": self.keyval,
         }
 
+    @classmethod
+    def from_securesystemslib_key(cls, key_dict: Dict[str, Any]) -> "Key":
+        """
+        Creates a Key object from a securesystemlib key dict representation
+        removing the private key from keyval.
+        """
+        key_meta = sslib_keys.format_keyval_to_metadata(
+            key_dict["keytype"],
+            key_dict["scheme"],
+            key_dict["keyval"],
+        )
+        return cls(
+            key_dict["keyid"],
+            key_meta["keytype"],
+            key_meta["scheme"],
+            key_meta["keyval"],
+        )
+
     def verify_signature(
         self,
         metadata: Metadata,


MVrachev:
As someone who didn't read the issue solved by this pr, it was a little unclear why we need to remove
the private part of the securesystemslib key. I had to read Jussi's issue to understand the problem.
Maybe you can modify your second sentence in your pr description and commit message like this:
```
TUF key doesn't handle it in any way,
but considering that we allow unrecognized symbols in the format,
we should exclude the private key otherwise this could lead to unexpected
behavior.

For example, those steps could lead to a mistake:
1. create a Metadata object with a root instance for Metadata.signed:
md = Metadata.from_file(<root_path>)
2. create a securesystemslib key:
3. create a TUF Key using the dictionary representing the securesystemslib key:
key = Key.from_dict(sslib_key["keyid"], sslib_key)
4. add the new TUF key to the root instance
md.signed.add_key("root", key)
5. save the metadata back to a file by calling Metadata.to_file():
md.to_file("root_test.json")

this will result in a new "root_test.json" file containing the private portion of a key.
```

Not sure if this will help, but a use case like this can highlight the issue without looking at the GitHub issue.
Otherwise, the code looks good to me.

PS: I tried the steps from 1 to 5 and they produce the expected result.

tuf/api/metadata.py
```diff
@@ -539,6 +539,25 @@ def to_securesystemslib_key(self) -> Dict[str, Any]:
...
+            key_dict["scheme"],
+            key_dict["keyval"],
+        )
+        return cls(
+            key_dict["keyid"],
+            key_meta["keytype"],
+            key_meta["scheme"],
+            key_meta["keyval"],
+            key_dict,
+        )
```
jku: this is interesting -- I was half expecting we'd have to make assumptions about keyval contents in this function but you found an sslib function that mostly does what you need... I think this looks reasonable except on the last line you add the whole key_dict (including the secret) as "unrecognised fields", oops. No need for unrecognised_fields here.

You _could_ modify the test so it looks for the secret in the serialized format of the key to be sure... but that sounds a bit much. Checking to_dict() output manually (to see that the above issue is fixed) should be fine.

tests/test_api.py
```diff
@@ -440,6 +444,17 @@ def test_key_class(self):
...
+        # Test if from_securesystemslib_key removes the private key from keyval
+        # of a securesystemslib key dictionary.
+        sslib_key = generate_ed25519_key()
+        key = Key.from_securesystemslib_key(sslib_key)
+
+        private_key = False
+        for keyval_key in key.keyval.keys():
+            private_key |= keyval_key == 'private'
+        
+        self.assertFalse(private_key)
```
jku:
```suggestion
        self.assertFalse('private' in key.keyval.keys())
```
or do i miss something?

jku (changes requested):

tuf/api/metadata.py
avelichka: Thaks, @jku! Addressed.

tests/test_api.py
avelichka: Thaks, @jku! Addressed.

jku (approved):
Thanks, LGTM. I think this makes sense so all of our example code and repository code can easily create new keys and use them without added code just to handle this issue.

The thing that makes me wonder about this is adding more and more functionality into Metadata API that the client does not need. We'll need to be careful not to overdo this (or figure out some RepositoryMixin classes or something ...)

joshuagl:
> The thing that makes me wonder about this is adding more and more functionality into Metadata API that the client does not need. We'll need to be careful not to overdo this (or figure out some RepositoryMixin classes or something ...)
> 

💯 Agree with this train of thought, metadata.py is very large.

PR #1521, opened 2021-08-10
Add an option to create TargetFile from data/file

### LINKED ISSUE #1447: Metadata API: Add TargetFile.from_data()
This is a repository tooling use case but also useful for testing (#1444).

There should be a way to create TargetFile from a file object or bytes, so something like:

    TargetFile.from_data(data: Union[bytes, BinaryIO])
    # if needed could also have a helper:
    TargetFile.from_file(filepath: str)

Here the function would look at the data and calculate length and hashes for me, and would be useful when I want to update my targets object:

    # in tests
    targets.signed.targets["targetname"] = TargetFile.from_data(b"inline test content")

    # in production repository tools when new target is being added:
    targets.signed.targets["targetname"] = TargetFile.from_data(target_fileobject)

This is already clearly functionality that client will never need... but I think it's fine to add into metadata.py: let's just keep an eye on this and avoid adding too much repository-only functionality in files that will be shared by both

avelichka:
This is a repository tooling use case but is also helpful when testing.
It could be useful when we need to update the targets object.

that is provided, the better.

Fixes #1447

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1219,6 +1219,78 @@ class TargetFile(BaseFile):
             **self.unrecognized_fields,
         }
 
+    @classmethod
+    def from_file(
+        cls,
+        target_file_path: str,
+        local_path: str,
+        hash_algorithms: Optional[List[str]] = None,
+    ) -> "TargetFile":
+        """Creates TargetFile object from a file.
+        Arguments:
+            target_file_path: The TargetFile path.
+            local_path: The local path to the file to create TargetFile from.
+            hash_algorithms: An optional list of hash algorithms to create
+                the hashes with. If not specified the securesystemslib default
+                hash algorithm is used.
+        Raises:
+            FileNotFoundError: The file doesn't exist.
+            UnsupportedAlgorithmError: The hash algorithms list
+                contains an unsupported algorithm.
+        """
+        with open(local_path, "rb") as file:
+            return cls.from_data(target_file_path, file, hash_algorithms)
+
+    @classmethod
+    def from_data(
+        cls,
+        target_file_path: str,
+        data: Union[bytes, IO[bytes]],
+        hash_algorithms: Optional[List[str]] = None,
+    ) -> "TargetFile":
+        """Creates TargetFile object from bytes.
+        Arguments:
+            target_file_path: The TargetFile path.
+            data: The data to create TargetFile from.
+            hash_algorithms: An optional list of hash algorithms to create
+                the hashes with. If not specified the securesystemslib default
+                hash algorithm is used.
+        Raises:
+            UnsupportedAlgorithmError: The hash algorithms list
+                contains an unsupported algorithm.
+        """
+        if isinstance(data, bytes):
+            length = len(data)
+        else:
+            data.seek(0, io.SEEK_END)
+            length = data.tell()
+
+        hashes = {}
+
+        if hash_algorithms is None:
+            hash_algorithms = [sslib_hash.DEFAULT_HASH_ALGORITHM]
+
+        for algorithm in hash_algorithms:
+            try:
+                if isinstance(data, bytes):
+                    digest_object = sslib_hash.digest(algorithm)
+                    digest_object.update(data)
+                else:
+                    digest_object = sslib_hash.digest_fileobject(
+                        data, algorithm
+                    )
+            except (
+                sslib_exceptions.UnsupportedAlgorithmError,
+                sslib_exceptions.FormatError,
+            ) as e:
+                raise exceptions.UnsupportedAlgorithmError(
+                    f"Unsupported algorithm '{algorithm}'"
+                ) from e
+
+            hashes[algorithm] = digest_object.hexdigest()
+
+        return cls(length, hashes, target_file_path)
+
     def verify_length_and_hashes(self, data: Union[bytes, IO[bytes]]) -> None:
         """Verifies that length and hashes of "data" match expected values.
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -47,6 +47,7 @@ from securesystemslib.interface import (
     import_ed25519_privatekey_from_file
 )
 
+from securesystemslib import hash as sslib_hash
 from securesystemslib.signer import (
     SSlibSigner,
     Signature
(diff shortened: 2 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -1123,6 +1124,41 @@ def to_dict(self) -> Dict[str, Any]:
             **self.unrecognized_fields,
         }
 
+    @classmethod
+    def from_file(cls, filepath: str) -> "TargetFile":
+        """Creates TargetFile object from a file."""
+        # Ensure the file path is valid/exists.
+        if not os.path.exists(filepath):
+            raise exceptions.NotFoundError(repr(filepath) + " does not exist.")
```
jku: We should use existing system exceptions when possible (old code uses a lot of custom exceptions, but I don't want this to happen in Metadata API if we can avoid it). Luckily open() already raises FileNotFoundError in this case so we don't need to have this code (but we may want to document the FileNotFoundError).

tuf/api/metadata.py
```diff
@@ -1123,6 +1124,41 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        with open(filepath, "rb") as file:
+            data = file.read()
+
+        return cls.from_data(data)
```
jku: we need to think about the case of very large files here (probably not uncommon for target files), and not read to memory: your from_data() looks complete enough that I _think_ you can just do 
```suggestion
        with open(filepath, "rb") as file:
            return cls.from_data(file)
```
this should end up in the file object branch which uses digest_fileobject() which does not read the whole file at once.

jku (changes requested):
I had completely glossed over the hash algorithm selection when writing the issue description... Seeing it implemented, it now seems to me that Metadata API cannot make that decision -- implementations will want to make their own choices --  and so hard coding sslib default is wrong. I guess this needs to be part of the API so:

```python
def from_bytes(cls, data: Union[bytes, BinaryIO], hash_algorithms: List[str]) -> "TargetFile"
def from_file(cls, filepath: str, hash_algorithms: List[str]) -> "TargetFile"
```

should there be a default value for hash_algorithms? Maybe ... but it's also something we can add afterwards without breaking api. If there is a default we possibly could use the sslib default instead of hard coding our own. In any case you'll need to document what happens on unsupported hash algorithm use

tuf/api/metadata.py
avelichka: Thanks, @jku! Addressed.

tuf/api/metadata.py
avelichka: Thanks, @jku! Addressed.

tuf/api/metadata.py
```diff
@@ -1123,6 +1123,73 @@ def to_dict(self) -> Dict[str, Any]:
...
+        Raises:
+            FileNotFoundError: The file doesn't exist.
+        """
+        with open(filepath, "rb") as file:
+            return cls.from_data(file, hash_algorithms)
+
+    @classmethod
+    def from_data(
+        cls, data: Union[bytes, BinaryIO], hash_algorithms: List[str]
+    ) -> "TargetFile":
```
jku: If you want to support a default hash algorithms value (as I said it isn't strictly necessary), it should work like Python default values elsewhere:

`hash_algorithms: Optional[List[str]] = None`

(and then the test call for that becomes `TargetFile.from_data(data)`)

jku (changes requested):
The implementation becomes surprisingly complex once all the details are handled -- but that's why I wanted this method to exist: so repository code doesn't have to figure this out...

I don't like using an empty list to mean "fill this list with something" but otherwise this LGTM.

tuf/api/metadata.py
avelichka: Thanks, @jku. Addressed.

tuf/api/metadata.py
```diff
@@ -1219,6 +1219,84 @@ def to_dict(self) -> Dict[str, Any]:
...
+        cls, filepath: str, hash_algorithms: Optional[List[str]] = None
+    ) -> "TargetFile":
+        """Creates TargetFile object from a file.
+        Attributes:
+            filepath: The path to the file to create the TaretFile object from.
+            hash_algorithms: An optional list of hash algorithms to create
+                the hashes with. If not specified the securesystemslib default
+                hash algorithm is used.
+        Raises:
+            FileNotFoundError: The file doesn't exist.
```
jku: * Arguments, not Attributes
* also raises UnsupportedAlgorithmError

jku (changes requested):
Hmm, sorry for leaving this hanging: for some reason Github is still showing this as "changes requested" (and not "review required")... I will have to learn this thing better.

* is there something that prevents calling `from_data()` in `from_file()` to avoid the repetition?
* It looks like the newly added path constructor argument does have to be an argument in your methods as well: "" is not a useful default and even in the from_file() case, local file path is not necessarily the same as target path (which is part of the target url that client will use). So something like  `TargetFile.from_file(local_path, path, "sha-256")` and `TargetFile.from_data(data, path, "sha-256")` -- not ideal but the two paths are different things and required.

tuf/api/metadata.py
avelichka: @jku, thanks for the feedback and the patience! I've addressed your comments.

tuf/api/metadata.py
```diff
@@ -1219,6 +1219,74 @@ def to_dict(self) -> Dict[str, Any]:
             **self.unrecognized_fields,
         }
 
+    @classmethod
+    def from_file(
+        cls, filepath: str, hash_algorithms: Optional[List[str]] = None
+    ) -> "TargetFile":
```
jku: This method still needs the target path argument that from_data() already has:
* `filepath` in this methods args is the local path to the content we should hash (note that in from_data() the same name is used for the target path)
* additionally we need a target path (which is a URL path-part decided by caller) that ends up as `TargetFile.path`. Currently code uses the local path but that's not necessarily correct.

The naming should be consistent so that both methods use the same argument name for target path. Maybe this methods arguments should be `local_path` and `target_path` while from_data() has `data` and `target_path`?

The order of args should be consistent as well: so target_path is in same position for both methods.

jku (changes requested):
Thanks. `from_file()` still needs a target path argument (left a more detailed comment in code), otherwise this looks good to me.

Confusion with target paths is such an everyday thing that we should try fix the root causes... If you have any improvement ideas for e.g. TargetFile docstring I'm all ears. Also if the explanation in the comment is unclear let's chat.

tuf/api/metadata.py
avelichka: I've misunderstood the path argument. Thanks for clarifying that for me, @jku! I've changed the code accordingly.

jku (approved):
LGTM, thanks.

The tests could use a different value for  local_path and target_file_path for clarity but it's not wrong so let's go with this.

Issue #1523, opened 2021-08-18
ngclient: review snapshot hashes/length check


jku:
Update process if timestamp meta contains hashes for snapshot:
* local root is loaded, root is updated until final
* local timestamp is loaded
* timestamp is updated (_meta (version, hashes, length) for snapshot changes here_)
* local snapshot is loaded  (_hashes do not match meta in timestamp, preventing loading_)
* remote snapshot is loaded -- **rollback checks are not done because local snapshot was not loaded**

The same issues do not exist for targets as they have no rollback checks.

Similar issues for version numbers and expiry are being fixed by delaying the checks until we know we have the "final" snapshot. We don't want to do the exact same thing here as the hashes are meant to prevent even _parsing_ data we don't trust... This seems to be a case where "trusted local metadata" and "untrusted data from network" is the deciding factor. TrustedMetadataSet does not have this information currently so possibly we need to add an argument to update_snapshot():

    def update_snapshot(self, data: bytes, trusted: Optional[bool] = False)

where trusted data would not be hash/length checked, but untrusted data would. Looks a bit ugly since none of the other update functions need this but it would seem to work in all cases:
* when loading local snapshot, we know the data has at some point been trusted (signatures have been checked): it does not need to match hashes now
* if there is no local snapshot and we're updating from remote, the remote data must match meta hashes in timestamp
* if there is a local snapshot and we're updating from remote, the remote data must match meta hashes in timestamp

MVrachev:
I would like to take this issue and will assign myself.
If somebody else is working on it, please let me know.

MVrachev:
I will wait with the implementation of a pr resolving this issue until there is a prototype for #1444 which will make creating a test
for this use case easier.
@jku is already working on it as he shared in his comment under this issue: https://github.com/theupdateframework/python-tuf/issues/1444#issuecomment-911843113

PR #1524, opened 2021-08-18
ngclient rollback improvements

### LINKED ISSUE #1498: ngclient: improve rollback protection
I believe ngclient implements the spec as its written... but fails to follow the spirit of the spec in one aspect: as the spec does not explicitly say that expired snapshot and timestamp can be used to do the rollback checks of new snapshot and timestamp (while it does explicitly allow this for root), ngclient does not use expired snapshot and timestamp and instead considers those expired versions invalid. This means a rollback attack (giving the client an older version of the metadata file) is easier as long as the local metadata versions have had time to expire.

Spec (client workflow) could be improved:
* specify when local cached metadata should be used to maximize the effectiveness of the rollback checks -- currently spec talks about "trusted timestamp" and "trusted snapshot" but never explains how they become trusted or when they're loaded
* specify when using expired metadata is allowed
  * root: when used to do rollback checks for newer root _(this is already in spec)_
  * timestamp: when used to do rollback checks for newer timestamp
  * snapshot: when used to do rollback checks for newer snapshot
* specify when using a snapshot version that does not match the timestamp meta version is allowed: when used to do rollback checks for newer snapshot

I believe it makes sense to work on the ngclient improvements at the same time. I think this should do it:
* for all of [root, timestamp, snapshot], do all validity checks in `TrustedMetadataSet.update_*()` as now, except do not check expiry: this allows updating TrustedMetadataSet first with a local (possibly expired) file and then updating with a new downloaded one
* instead check the expiry when the _next_ top-level type is updated (this is safe because of the strong checks we do: TrustedMetadataSet guarantees that top-level updates can only be done in strict order):
  * when timestamp is updated, first check root expiry
  * when snapshot is updated, first check timestamp expiry
  * when targets is updated, first check snapshot expiry and meta version
* deprecate `root_update_finished()` as the expiry check is now done consistently as described above

jku:
Fixes #1498

**Improve rollback protection in ngclient**

Spec does not explicitly say so but the intent is that "local" timestamp and snapshot metadata can be trusted for rollback checks of new versions of the same metadata even if the local metadata is expired. This is similar to root (but in that case the intent is explicitly explained in the spec).

Likewise intent is that "local" snapshot can be trusted for rollback checks of new versions of the snapshot even if local snapshot does not match the meta version in current timestamp.

Implement the above by moving the expiry and meta version checks to the "next" metadata update: check timestamp expiry just before snapshot is updated, and check snapshot expiry and meta version just before targets is updated. This means that when e.g. a second timestamp (that was downloaded from remote) is loaded, the current (first) timestamp is used for rollback checks even if it is expired -- but when snapshot is loaded the expiry of the current, final timestamp is checked.

Some additional details -- note especially the two divergences from the spec letter (but IMO not intent):
 * the expiry/meta checks are now done every time the "next" metadata updates: e.g. snapshot expiry and meta version are checked every time a targets metadata is updated -- it would be fine to do it only once (when initial "targets" is first updated) but this is simple to implement and the cost is likely minimal
 * metadata does get written to disk before we know if it fulfills the "complete" requirements for expiry and meta version: this is not what the spec says but seems correct to me as the data is "trusted" (signed by threshold of correct keys) and spec documents similar process for root. Even if the final metadata really is expired and the update fails, if the remote later starts serving non-expired metadata, the client will succeed in a later update (and will correctly use the earlier expired metadata for rollback checks!)
 * As Martin points out in comments, the "next" metadata may in some cases be downloaded before the final checks are done (so a snapshot file may be downloaded, but not parsed, before final timestamp expiry is checked): this does not have security implications but is not the order spec demands
 * This does not solve the issue with snapshot hashes/length in timestamp meta: it seems to require TrustedMetadataSet API change so I left it in separate issue: #1523

### DIFF
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -178,6 +178,10 @@ class TrustedMetadataSet(abc.Mapping):
     def update_timestamp(self, data: bytes):
         """Verifies and loads 'data' as new timestamp metadata.
 
+        Note that an expired intermediate timestamp is considered valid so it
+        can be used for rollback checks on newer, final timestamp. Expiry is
+        only checked for the final timestamp in update_snapshot().
+
         Args:
             data: unverified new timestamp metadata as bytes
 
@@ -227,14 +231,18 @@ class TrustedMetadataSet(abc.Mapping):
                     self.timestamp.signed.meta["snapshot.json"].version,
                 )
 
-        if new_timestamp.signed.is_expired(self.reference_time):
-            raise exceptions.ExpiredMetadataError("New timestamp is expired")
+        # expiry not checked to allow old timestamp to be used for rollback
+        # protection of new timestamp: expiry is checked in update_snapshot()
 
         self._trusted_set["timestamp"] = new_timestamp
         logger.debug("Updated timestamp")
 
     def update_snapshot(self, data: bytes):
         """Verifies and loads 'data' as new snapshot metadata.
+
+        Note that an expired intermediate snapshot is considered valid so it
+        can be used for rollback checks on newer, final snapshot. Expiry is
+        only checked for the final snapshot in update_delegated_targets().
 
         Args:
             data: unverified new snapshot metadata as bytes
@@ -249,6 +257,11 @@ class TrustedMetadataSet(abc.Mapping):
         if self.targets is not None:
             raise RuntimeError("Cannot update snapshot after targets")
         logger.debug("Updating snapshot")
+
+        # Local timestamp was allowed to be expired to allow for rollback
+        # checks on new timestamp but now timestamp must not be expired
+        if self.timestamp.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("timestamp.json is expired")
 
         meta = self.timestamp.signed.meta["snapshot.json"]
 
@@ -301,8 +314,8 @@ class TrustedMetadataSet(abc.Mapping):
                         f"{new_fileinfo.version}, got {fileinfo.version}."
                     )
 
-        if new_snapshot.signed.is_expired(self.reference_time):
-            raise exceptions.ExpiredMetadataError("New snapshot is expired")
+        # expiry not checked to allow old snapshot to be used for rollback
+        # protection of new snapshot: expiry is checked in update_targets()
 
         self._trusted_set["snapshot"] = new_snapshot
         logger.debug("Updated snapshot")
@@ -335,6 +348,11 @@ class TrustedMetadataSet(abc.Mapping):
         """
         if self.snapshot is None:
             raise RuntimeError("Cannot load targets before snapshot")
+
+        # Local snapshot was allowed to be expired to allow for rollback
+        # checks on new snapshot but now snapshot must not be expired
+        if self.snapshot.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("snapshot.json is expired")
 
         delegator: Optional[Metadata] = self.get(delegator_name)
         if delegator is None:
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -240,9 +240,11 @@ class TrustedMetadataSet(abc.Mapping):
     def update_snapshot(self, data: bytes):
         """Verifies and loads 'data' as new snapshot metadata.
 
-        Note that an expired intermediate snapshot is considered valid so it
-        can be used for rollback checks on newer, final snapshot. Expiry is
-        only checked for the final snapshot in update_delegated_targets().
+        Note that intermediate snapshot is considered valid even if it is
+        expired or the version does not match the timestamp meta version. This
+        means the intermediate snapshot can be used for rollback checks on
+        newer, final snapshot. Expiry and meta version are only checked for
+        the final snapshot in update_delegated_targets().
 
         Args:
             data: unverified new snapshot metadata as bytes
@@ -285,18 +287,10 @@ class TrustedMetadataSet(abc.Mapping):
 
         self.root.verify_delegate("snapshot", new_snapshot)
 
-        if (
-            new_snapshot.signed.version
-            != self.timestamp.signed.meta["snapshot.json"].version
-        ):
-            raise exceptions.BadVersionNumberError(
-                f"Expected snapshot version "
-                f"{self.timestamp.signed.meta['snapshot.json'].version}, "
-                f"got {new_snapshot.signed.version}"
-            )
+        # version not checked against meta version to allow old snapshot to be
+        # used in rollback protection: it is checked when targets is updated
 
-        # If an existing trusted snapshot is updated,
-        # check for a rollback attack
+        # If an existing trusted snapshot is updated, check for rollback attack
         if self.snapshot is not None:
             for filename, fileinfo in self.snapshot.signed.meta.items():
                 new_fileinfo = new_snapshot.signed.meta.get(filename)
@@ -315,10 +309,24 @@ class TrustedMetadataSet(abc.Mapping):
                     )
 
         # expiry not checked to allow old snapshot to be used for rollback
-        # protection of new snapshot: expiry is checked in update_targets()
+        # protection of new snapshot: it is checked when targets is updated
 
         self._trusted_set["snapshot"] = new_snapshot
         logger.debug("Updated snapshot")
+
+    def _check_final_snapshot(self):
+        if self.snapshot.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("snapshot.json is expired")
+
+        if (
+            self.snapshot.signed.version
+            != self.timestamp.signed.meta["snapshot.json"].version
+        ):
+            raise exceptions.BadVersionNumberError(
+                f"Expected snapshot version "
+                f"{self.timestamp.signed.meta['snapshot.json'].version}, "
+                f"got {self.snapshot.signed.version}"
+            )
 
     def update_targets(self, data: bytes):
         """Verifies and loads 'data' as new top-level targets metadata.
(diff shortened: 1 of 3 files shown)


jku:
I'm pretty happy with the result. One disadvantage is that a call to `update_snapshot()` can lead to "Timestamp is expired" error: that is a bit surprising but it is documented and I think beats the alternative (requiring the caller to explicitly decide when root/timestamp/snapshot is "final" and should be checked for expiry etc).

I may have gone a bit overboard with the code comments but this seems like a tricky situation so comments seemed useful: if someone manages to shorten them to one-liners, I'll be happy to edit...

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -213,11 +185,15 @@ def update_timestamp(self, data: bytes):
...
         """
-        if not self._root_update_finished:
-            raise RuntimeError("Cannot update timestamp before root")
         if self.snapshot is not None:
             raise RuntimeError("Cannot update timestamp after snapshot")
 
+        # client workflow 5.3.10: Make sure final root is not expired.
+        if self.root.signed.is_expired(self.reference_time):
+            raise exceptions.ExpiredMetadataError("Final root.json is expired")
+        # No need to check for 5.3.11 (fast forward attack recovery):
```
MVrachev, line 198: This made me think: did we do that check before?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -250,6 +258,11 @@ def update_snapshot(self, data: bytes):
             raise RuntimeError("Cannot update snapshot after targets")
         logger.debug("Updating snapshot")
```
MVrachev, line 262:
```suggestion

        # client workflow 5.4.4: Make sure the final timestamp is not expired.
```
Maybe it will be even clearer if you add the step number of the client workflow?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -336,6 +349,11 @@ def update_delegated_targets(
         if self.snapshot is None:
             raise RuntimeError("Cannot load targets before snapshot")
```
MVrachev, line 359:
```suggestion

        # client workflow 5.5.4 and 5.5.6
```
Maybe it will be even clearer if we add the steps numbers of the client workflow?

MVrachev (commented):
> Spec does not explicitly say so but the intent is that "local" timestamp and snapshot metadata can be trusted for rollback checks of new versions of the same metadata even if the local metadata is expired.

The above sounds so logical and I am wondering how we haven't thought about it before.

I have a couple of questions and I will number them for easier cross-reference.

> Likewise intent is that "local" snapshot can be trusted for rollback checks of new versions of the snapshot even if local snapshot does not match the meta version in the current timestamp.

That is true because timestamp could have been updated and as a consequence its timestamp.meta["snapshot"].version could be bumped right?

By checking only the last timestamp/snapshot for expiration isn't there a risk of an attacker sending an intermediate expired version of timestamp/snapshot?

What I liked in the initial implementation was that it closely mapped to the client workflow defined in the specification and now some steps are swapped: for example step 5.5.6 will happen between steps 5.6.1 and 5.6.2.
I imagine this could be confusing (even though you added good explanatory comments).
This is something that should be addressed in the specification.
From what I understood, @joshuagl will file an issue about this?

I am even wondering won't it be better if we fix the problem in the specification first and then work on this pr?
Then you will be able to reference the specific step like client workflow 5.6.6. and quote the spec as an argument why the expiration checks are moved.

tuf/ngclient/_internal/trusted_metadata_set.py
jku: current ngclient code has a similar comment, and does not do the check. the legacy updater does do this check because it may have untrusted metadata in memory

tuf/ngclient/_internal/trusted_metadata_set.py
jku: It would but unfortunately it also requires keeping these numbers updated with spec changes... so I'm more inclined to remove the step numbers I did use in the PR then I am to add more of them.

jku:
Sorry, I accidentally edited your comment while trying to answer it: I tried to revert but github does not make it easy.

> > Likewise intent is that "local" snapshot can be trusted for rollback checks of new versions of the snapshot even if local snapshot does not match the meta version in the current timestamp.

> That is true because timestamp could have been updated and as a consequence its timestamp.meta["snapshot"].version could be bumped right?

yes, in fact that's required: if you update snapshot the only way to make that snapshot part of the repo for the clients is to publish a new timestamp with updated meta.

> By checking only the last timestamp/snapshot for expiration isn't there a risk of an attacker sending an intermediate expired version of timestamp/snapshot?

for both of these only two versions are potentially loaded:

* local version may be loaded (but a local version is not required to exist)
* the remote version may be loaded (spec includes always downloading both timestamp and snapshot but especially for snapshot that seems like a specification bug: local version may be the exact version we need, there's no need to load in that case)

One of these is the final version that ends up getting expiry checked -- and in any case even if a buggy client somehow loaded more expired snapshots during the process that would be fine: no security issues that I can see.

> What I liked in the initial implementation was that it closely mapped to the client workflow defined in the specification and now some steps are swapped: for example step 5.5.6 will happen between steps 5.6.1 and 5.6.2.

This is a good point! the expiry (and meta version) checks currently jump over two spec items:
* writing local metadata to disk (this I covered in my PR cover letter)
* downloading "next" metadata (5.6.1 in your example)

I believe the first jump is correct and this seems to be what legacy updater does as well. The second item is an implementation detail and could be avoided by having snapshot_is_final() type API... but I dislike that because it requires more work from the caller.

> I am even wondering won't it be better if we fix the problem in the specification first and then work on this pr?

I don't really want to wait for the spec:

* ngclient is currently lacking a feature that legacy updater has (the rollback protection) and I consider that a bug
* the spec changes are likely to be nontrivial: I think it requires defining how local metadata is loaded as well as this tricky interaction between expiry / meta version and rollback checks
* we keep finding new things (or at least I do!) by implementing this: I believe the spec tasks are getting better defined because this PR is progressing

jku:
> > Spec does not explicitly say so but the intent is that "local" timestamp and snapshot metadata can be trusted for rollback checks of new versions of the same metadata even if the local metadata is expired.
> 
> The above sounds so logical and I am wondering how we haven't thought about it before.

This is a combo of two things:
* spec is almost completely silent on when and how to load local metadata (it just expects there to be "trusted metadata" already at some point in the workflow without explaining where it came from)
*  it's  hard to figure out which of the rollback protections are actually in use in legacy updater (I'm unsure about the expiry cases for local metadata): in any case they load all local metadata in the beginning, and only unload if keys are rotated -- this let's the implementation just forget about issues like "is our currently loaded snapshot actually the snapshot that our new timestamp considers correct", it just assumes this potential problem will be fixed sometime later in the process.

MVrachev:
> > By checking only the last timestamp/snapshot for expiration isn't there a risk of an attacker sending an intermediate expired version of timestamp/snapshot?
> 
> for both of these only two versions are potentially loaded:
> 
> * local version may be loaded (but a local version is not required to exist)
> * the remote version may be loaded (spec includes always downloading both timestamp and snapshot but especially for snapshot that seems like a specification bug: local version may be the exact version we need, there's no need to load in that case)
> 
> One of these is the final version that ends up getting expiry checked -- and in any case even if a buggy client somehow loaded more expired snapshots during the process that would be fine: no security issues that I can see.

I think I understand what you mean.
Even if a user is fed by an attacker with expired `timestamp/snapshot` data when calling `update_timestamp/update_snapshot` that won't be a security concern, because the next time when `update_snapshot` (for timestamp)  or `update_targets` (for snapshot) is called an expiry check will be done.

> 
> > I am even wondering won't it be better if we fix the problem in the specification first and then work on this pr?
> 
> I don't really want to wait for the spec:
> 
> * ngclient is currently lacking a feature that legacy updater has (the rollback protection) and I consider that a bug

Do the legacy updater checks this? From your next comment:

> it's hard to figure out which of the rollback protections are actually in use in legacy updater (I'm unsure about the expiry cases for local metadata): in any case they load all local metadata in the beginning, and only unload if keys are rotated -- this let's the implementation just forget about issues like "is our currently loaded snapshot actually the snapshot that our new timestamp considers correct", it just assumes this potential problem will be fixed sometime later in the process.

seems like the old updater doesn't take care of this case.

> * the spec changes are likely to be nontrivial: I think it requires defining how local metadata is loaded as well as this tricky interaction between expiry / meta version and rollback checks
> * we keep finding new things (or at least I do!) by implementing this: I believe the spec tasks are getting better defined because this PR is progressing

I agree that this pr helps to define what should be changed in the spec.
I am just worried that this moves steps from the update process which is basically the heart of TUF and I don't want us to be in a situation when we are not following the spec update process for a while.
Let's at least make sure we have an issue and discussion going at the spec repo and we are moving in that direction, can we?

jku:
> seems like the old updater doesn't take care of this case.

I did not say this. I said it's hard to figure out if it does the checks in all cases (because the checks are done in places that are rather unobvious at least to me -- it's difficult to be sure that all possible scenarios are actually covered... but they may well be).

The legacy updater definitely has rollback protections in place that ngclient does not have, this is not in question.

jku:
> I am just worried that this moves steps from the update process which is basically the heart of TUF and I don't want us to be in a situation when we are not following the spec update process for a while.
Let's at least make sure we have an issue and discussion going at the spec repo and we are moving in that direction, can we?

This should cover it maybe https://github.com/theupdateframework/specification/issues/179

I think fixing the spec might be complex enough (difficult to get right, and also difficult to negotiate WRT existing implementation decisions) that I don't like to just wait for the spec to be updated... we should do our best effort to implement what we think is the spec intent.

MVrachev:
> > I am just worried that this moves steps from the update process which is basically the heart of TUF and I don't want us to be in a situation when we are not following the spec update process for a while.
> > Let's at least make sure we have an issue and discussion going at the spec repo and we are moving in that direction, can we?
> 
> This should cover it maybe [theupdateframework/specification#179](https://github.com/theupdateframework/specification/issues/179)

Yes, it does.

> 
> I think fixing the spec might be complex enough (difficult to get right, and also difficult to negotiate WRT existing implementation decisions) that I don't like to just wait for the spec to be updated... we should do our best effort to implement what we think is the spec intent.

Now that at least we have documented what you have found in the spec repo and hopefully discussion can start I agree that it
makes sense to do what we believe it's right from a security standpoint and document it well (as I think you did).

joshuagl:
> * metadata does get written to disk before we know if it fulfills the "complete" requirements for expiry and meta version: this is not what the spec says but seems correct to me as the data is "trusted" (signed by threshold of correct keys) and spec documents similar process for root. Even if the final metadata really is expired and the update fails, if the remote later starts serving non-expired metadata, the client will succeed in a later update (and will correctly use the earlier expired metadata for rollback checks!)

Agree, this seems OK. The well-defined process for root is because we explicitly download intermediate versions we expect to be expired, whereas for the other top-level files we do not. There have been some preliminary efforts to define how to load existing downloaded metadata for rollback protection here https://github.com/theupdateframework/taps/pull/128, but I think the size of the spec change is a blocker. I really want to find time to work on https://github.com/theupdateframework/specification/issues/121

joshuagl (approved):
LGTM. The removal of `root_update_finished()` makes the API cleaner.

@sechkova what do you think?

sechkova (approved):
I'm ok with these changes too.

jku:
Thank you, let's see how this works out!

PR #1531, opened 2021-08-22
Fixes #1526

### LINKED ISSUE #1526: ngclient: persist metadata in a safer way

            f.write(data)

* write content to tempfile

suvadityamuk:
that is provided, the better.

Fixes #1526 

Updated _persist_metadata() according to instructions in #1526 

requirements**:

- [ Checked] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [Checked ] Tests have been added for the bug fix or new feature: Passes all local tests
- [Not applicable ] Docs have been added for the bug fix or new feature

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -292,8 +292,10 @@ class Updater:
             return f.read()
 
     def _persist_metadata(self, rolename: str, data: bytes):
-        with open(os.path.join(self._dir, f"{rolename}.json"), "wb") as f:
-            f.write(data)
+        original_filepath = os.path.join(self._dir, f"{rolename}.json")
+        temp_file = open(f'{rolename}_temp.json', 'wb+')
+        temp_file.write(data)
+        sslib_util.persist_temp_file(temp_file, original_filepath, should_close=True)
 
     def _load_root(self) -> None:
         """Load remote root metadata.
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -297,6 +297,7 @@ class Updater:
         temp_file.write(data)
         sslib_util.persist_temp_file(temp_file, original_filepath, should_close=True)
 
+
     def _load_root(self) -> None:
         """Load remote root metadata.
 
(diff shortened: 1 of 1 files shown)


suvadityamuk:
I somehow can't get this DCO thing completed. I have given the commit message accordingly but still it does not recognise or approve it

jku:
Thanks!

That DCO complaint is a bit strange... I think maybe it's getting confused by the first commit which you maybe made with github web ui unlike the other commits. There's some minor comments so we could see if DCO is still an issue after those:
* Could you squash the commits to a single commit? Now there are two commits with just whitespace changes
* The commit title could be more informative (something like "ngclient: persist metadata using a temp file")

All the nitpicks aside... I'm now wondering if my original advice is correct. I read` securesystemslib.persist_temp_file()` and it does not look like an atomic operation as I expected :( I'll have a closer look and get back to you: _no need to improve this PR before that_.

suvadityamuk:
Should I make a fresh PR? To see if this DCO problem gets solved then?
This is what Actions is telling me is the problem
image

jku:
> Should I make a fresh PR? To see if this DCO problem gets solved then?
> This is what Actions is telling me is the problem
> image

That looks like what I suspected: the committer info in that one commit (unlike the other two) looks like it was maybe created in github UI? possibly you later added the signed-off-by to that commit with a (seemingly) different email address and the DCO checker is now confused.

Creating a new PR won't help but using just a single new commit with all the changes in it should work...  but don't spend a lot of your time polishing this yet: as I mentioned my original advice may have been wrong -- We're trying to make sure the final write is safe and atomic and I thought `persist_temp_file()` is but that might not actually be the case

suvadityamuk:
Okay, understood. I'll stand by till then

jku:
I've added my new opinion on how to solve this in #1526 (sorry for misleading in the original issue description). My earlier comments on commit structure also still stand.

suvadityamuk:
Understood. I'll make the changes accordingly. Could you assign this task?

PR #1532, opened 2021-08-23
Address Pylint new unspecified-encoding warning

MVrachev:

A new warning appeared from `pylint` when calling "tox -e lint" on the
"develop" branch with id "unspecified-encoding".
I read about the warning and it makes sense, so maybe we shouldn't ignore it?
I chose `UTF8` as a parameter because it made sense to me at least.

Read more about the warning here:
https://github.com/PyCQA/pylint/issues/3826

requirements**:

### DIFF
--- a/tuf/developer_tool.py
+++ b/tuf/developer_tool.py
@@ -749,7 +749,7 @@ def _save_project_configuration(metadata_directory, targets_directory,
     project_config['public_keys'][key] = key_metadata
 
   # Save the actual file.
-  with open(project_filename, 'wt') as fp:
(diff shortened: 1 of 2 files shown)


sechkova (commented):
The recommended solution by [PEP 597](https://www.python.org/dev/peps/pep-0597/#prepare-to-change-the-default-encoding-to-utf-8),  encoding='locale' is not available before python 3.10 so I guess utf-8 "makes sense".

jku (approved):
Thanks. In other code (like Metadata API) we're explicitly encoding data to utf-8 (bytes) then using binary mode to write -- but this seems like a fine fix for these cases.

PR #1533, opened 2021-08-23
Metadata API: improve metadata module documentation

### LINKED ISSUE #1504: improve "generic" Metadata module documentation
We should improve the Metadata module documentation just a bit so that it explains what the API is. 
I don’t think it needs a lot. But something like:
* Defines an interface to single metadata role/file
* does not have concepts like “repository” or “trusted collection of metadata”
* Aims to provide safe de/serialization, **and** a safe API for accessing/modifying the data
* Does not implement client update workflows or repository-side logic, but offers a solid base for components that do

MVrachev:
Fixes #1504 

Clarify that metadata API is a low-level API and as such it doesn't use
concepts like "repository" or "trusted collection of metadata" and
don't implement the repository logic or client updater workflow.

Additionally, explain that each of the top-level metadata classes
represents exactly one metadata file and explain the purpose of the
helper classes.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -3,9 +3,28 @@
 
 """TUF role metadata model.
 
-This module provides container classes for TUF role metadata, including methods
-to read and write from and to file, perform TUF-compliant metadata updates, and
-create and verify signatures.
+This module contains low-level API through container classes for TUF role
+metadata. The API aims to provide:
+
+* Safe de/serialization of metadata to and from files.
+* Access to and modification of signed metadata content.
+* Signing metadata and verifying signatures.
+
+Each of the top level metadata roles is an instance of the Metadata[T] class
+where the "signed" portion of each of the roles (or the "T") is an instance
+of one of the classes Root, Timestamp, Snapshot or Targets.
+For example, Metadata[Root] represents the TUF root role and that in practice
+means that this is a Metadata object with a signed attribute of type Root.
+
+Additionally, there are helper classes providing abstractions over the complex
+metadata fields inside the four top level classes - Root, Timestamp, Snapshot
+and Targets.
+
+Note: the metadata module provides a low-level API and as such it doesn't use
+concepts like "repository" or "trusted collection of metadata".
+In this file there is no implementation of the repository-side logic or client
+update workflows, but instead it provides solid base for other components to do
+so.
 
 The metadata model supports any custom serialization format, defaulting to JSON
 as wireline format and Canonical JSON for reproducible signature creation and


MVrachev:
Please note that `tox -e lint` fails because of a new `pylint` warning. This issue is fixed in #1532

MVrachev:
Pr #1532 was merged and the lint failing was resolved.

tuf/api/metadata.py
```diff
@@ -5,7 +5,21 @@
 
 This module provides container classes for TUF role metadata, including methods
 to read and write from and to file, perform TUF-compliant metadata updates, and
-create and verify signatures.
+create and verify signatures. The intent of the API is to provide safe
+serialization and deserializaiton and a safe API for accessing and modifying
+the metadata.
```
jku: This feels like it duplicates parts of the original sentence. I think your text about _"safe API for accessing and modifying the metadata"_ is better than original though... maybe we could mostly just use the original sentence but replace _"TUF-compliant metadata updates"_ (which seems a little vague and even incorrect) with your wording?

I guess another alternative is to actually present features as a list (but I'll leave decision to you):
* Safe de/serialization of metadata to and from files
* Access to and modification of signed metadata content
* Signing metadata and verifying signatures

tuf/api/metadata.py
```diff
@@ -5,7 +5,21 @@
...
+
+Each of the top level metadata roles is represented by a separate class and each
+of those classes denotes exactly one metadata file.
```
jku: I really like documenting this part (as it's the weird part of Metadata API and requires an explanation I think) but this feels a bit vague for someone seeing Metadata API for the first time and inaccurate: Root does not map to a root metadata file, it maps to the signed portion of the metadata file... Metadata is the equivalent of a metadata file.

I'd like to instead see actual names of classes (with links directly to the relevant doc), explanation of Root,Targets,etc classes vs Metadata class and how to work with them. Maybe you could do it with an example of how `Metadata[Root]` represents the TUF root role and that in practice it means a Metadata object with a signed attribute of type Root.

tuf/api/metadata.py
```diff
@@ -5,7 +5,21 @@
...
+Additionally, there are helper classes providing abstraction over the complex
+metadata fields inside the metadata files and are used inside the four top
+level classes.
```
jku: This sounds good, but "four top level classes" could maybe be spelled out so it's clear what it means.

jku (commented):
Left some specific comments. 
Other comments:
* I wonder if we could/should re-order the classes in the metadata module documentation so that  Metadata and the signed classes are first?
* alternatively should we have a "Table of contents"-like list in this text that lists the main classes more prominently than the helpers?

... basically I'd love it if we could communicate to the reader that "these 5 classes are the ones you should look at, others not so much". I'm fine with leaving this for future if you don't feel like solving this here.

jku:
Also please rebase so we get the automatic readthedocs build

tuf/api/metadata.py
```diff
@@ -3,9 +3,31 @@
...
+This module contains low-level API through container classes for TUF role
+metadata. The API aims to provide:
+- Safe de/serialization of metadata to and from files
+- Access to and modification of signed metadata content
+- Signing metadata and verifying signatures
+
+Each of the top level metadata roles is an instance of the Metadata[T] class
+where the "signed" portion of each of the roles (or the "T") is an instance
+of one of the classes:
+- Root (See https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.metadata.html#tuf.api.metadata.Root)
```
MVrachev: I am not sure how do we want to add those links: links that could be copied from a source code editor and pasted in the browser or formatted links according to the readthedocs documentation: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-sections-in-the-same-document

MVrachev: I played a lot about how I can beautify the content following the readthedocs format.
@jku what do you think?

MVrachev:
> * I wonder if we could/should re-order the classes in the metadata module documentation so that  Metadata and the signed classes are first?

I tried it, but it seems that on readthedocs the classes are documented in alphabetical order instead of the order of definition.

tuf/api/metadata.py
jku: Yeah... no really good options here. What you did is otherwise great (the fact that it works in source is nice), but we also plan to publish docs for released versions and you can't just link to latest from there.

I think the viable options are to not link at all or to use the anchor references (that won't work in source code). Neither is a great solution.

jku:
> > * I wonder if we could/should re-order the classes in the metadata module documentation so that  Metadata and the signed classes are first?
> 
> I tried it, but it seems that on readthedocs the classes are documented in alphabetical order instead of the order of definition.

There are configuration options for conf.py and automodule to configure this.
Alternatively we could use many autoclass directives instead of one automodule directive and control the order in the RST file

tuf/api/metadata.py
MVrachev: I think if we move the top-level metadata classes before the helper classes and do not add links could be fine.

MVrachev:
> There are configuration options for conf.py and automodule to configure this.
> Alternatively we could use many autoclass directives instead of one automodule directive and control the order in the RST file

Which one do you think is better? I am trying to find the configuration option for Sphinx conf.py about how to change the default behavior of how the documentation about classes is ordered, but I was wondering do you think autoclass directives are better?

Also, @jku  can you give an opinion about the text changes besides the links?

jku:
> Which one do you think is better? I am trying to find the configuration option for Sphinx conf.py about how to change the default behavior of how the documentation about classes is ordered, but I was wondering do you think autoclass directives are better?

I don't know which one is better. Using configuration means we hard-code less about the classes in our documentation files which is good... but using handcrafted autoclass directives gives us a lot more control over the way docs look and might be useful now that Metadata API is unlikely to change that much (in the sense that classes would keep appearing and disappearing).

If you have trouble finding the docs: Everything related to generating docs from docstrings happens in the autodoc extension (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html), not sphinx core.  Knowing the available configuration there will be helpful

I think we should maybe consider the re-ordering out of scope here and just get the text changes in, I think they looked ok.

MVrachev:
> I think we should maybe consider the re-ordering out of scope here and just get the text changes in, I think they looked ok.

I agree and think the focus of this pr should be the additions and modifications of the introduction text.
Other tweaks and experiments can be done when addressing an issue https://github.com/theupdateframework/python-tuf/issues/1543 in a separate pr.
Removed the commit which moved the metadata classes and the links as they are ugly if you read them from a source code editor.

jku (approved):

PR #1535, opened 2021-08-24
ngclient: persist metadata using atomic write to solve #1526

### LINKED ISSUE #1526: ngclient: persist metadata in a safer way

suvadityamuk:
that is provided, the better.

Fixes #1526 

Added changes as mentioned by @jku in #1526 for supporting atomic operation for persisting metadata.

requirements**:

- [Checked ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [Checked ] Tests have been added for the bug fix or new feature (all tests passing)
- [ Not applicable] Docs have been added for the bug fix or new feature

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -292,8 +292,10 @@ class Updater:
             return f.read()
 
     def _persist_metadata(self, rolename: str, data: bytes):
-        with open(os.path.join(self._dir, f"{rolename}.json"), "wb") as f:
-            f.write(data)
+        original_filepath = os.path.join(self._dir, f"{rolename}.json")
+        temp_file = open(f'{rolename}_temp.json', 'wb+')
+        temp_file.write(data)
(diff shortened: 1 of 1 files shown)


jku (changes requested):
I'd like to see single commit (instead of four) with these changes, with the committer info matching your signed-off-by. One way to achieve that could be something like:
```
git reset HEAD^4 # removes the four commits you currently have but preserves changes
git add tuf/ngclient/updater.py
git commit -s # create new commit with nice commit message and signed-off-by
```
but there's some additional code comments:
* no need to open original_filepath at all as far as I can see (blog post reads from that file but we don't)
* replace should happen outside the with-contextmanager (the issue is explained in the linked blog post and I believe still exists even though we only open one of the files)
* a docstring could explain something like """Writes to temporary file and moves it to ensure content change is atomic""" -- otherwise this might look like unnecessary magic

PR #1536, opened 2021-08-24
Fixed #1526 using tempfile creation and atomic move

### LINKED ISSUE #1526: ngclient: persist metadata in a safer way

suvadityamuk:

that is provided, the better.

Fixes #1526

Uses `tempfile.NamedTemporaryFile` to first write to a temp file which is then moved on top of the original file. Passes all tests on local system.

requirements**:

- [ Checked ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ Checked ] Tests have been added for the bug fix or new feature
- [ Not Applicable ] Docs have been added for the bug fix or new feature


suvadityamuk:
I don't understand where it is failing, could you please let me know what is wrong here?

PR #1537, opened 2021-08-24
Enforce role name uniqueness and add Targets key helpers

### LINKED ISSUE #1426: Metadata API: enforce role name uniqueness in delegations

### LINKED ISSUE #1469: Metadata API: Missing key helper functions in Targets/Delegations
Root class has functionality to add and remove keys for delegated metadata (`add_key()/remove_key()`) but the other delegator Targets does not. It should provide same/similar functionality.

Some options:
* Add `Targets.add_key(self, role: str, key: Key)` and `Targets.remove_key(self, role: str, keyid: str)`
* Handle the same in Delegations object

I think the former option makes sense so that the Targets API matches Root as well as possible (and Delegations remains mostly a Targets implementation detail).

MVrachev:
Fixes #1426, #1469

This pr contains three types of changes:

**Enforce Delegation role name uniqueness**

The spec does not say anything about role name uniqueness in a
delegations object, but I believe we cannot safely allow multiple roles
with the same role name in the roles array of a delegations object.
If we did then the roles could have different keyids, and then we would
end up in a situation where metadata may be both a valid delegation
and an invalid delegation at the same time, depending on how the role
gets chosen and that does not seem like the intention of the design.
There is an issue open in the specification with number 167 about
that issue.

Regardless of the Metadata API, I think we should enforce role name
uniqueness.
I chose to change the data structure containing roles to
OrderedDict, where keys are role names and values, are DelegatedRole
instances, because role names are the unique identifier of a role and
their order is important to the way they are traversed afterward.

**Add missing key helpers in Delegations**

Root class has the functionality to add and remove keys for delegated
metadata (add_key()/remove_key()) but the other delegator Targets does
not.
It should provide the same/similar functionality.

**Some cleaning**

1. Moves a couple of Delegations serialization tests inside `test_metadata_serialization.py`
2. Removed some unused imports.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -29,7 +29,6 @@ from tuf.api.metadata import (
     Key,
     MetaFile,
     TargetFile,
-    Delegations,
     DelegatedRole,
 )
 
@@ -536,6 +535,82 @@ class TestMetadata(unittest.TestCase):
         self.assertEqual(
             targets.signed.targets[filename].to_dict(), fileinfo.to_dict()
         )
+
+    def test_targets_key_api(self):
+        targets_path = os.path.join(
+                self.repo_dir, 'metadata', 'targets.json')
+        targets: Targets = Metadata[Targets].from_file(targets_path).signed
+
+        # Add a new delegated role "role2" in targets
+        delegated_role = DelegatedRole.from_dict({
+                "keyids": [],
+                "name": "role2",
+                "paths": ["fn3", "fn4"],
+                "terminating": False,
+                "threshold": 1
+        })
+        targets.delegations.roles["role2"] = delegated_role
+
+        key_dict = {
+            "keytype": "ed25519",
+            "keyval": {
+                "public": "edcd0a32a07dce33f7c7873aaffbff36d20ea30787574ead335eefd337e4dacd"
+            },
+            "scheme": "ed25519"
+        }
+        key = Key.from_dict("id2", key_dict)
+
+        # Assert that delegated role "role1" does not contain the new key
+        self.assertNotIn(key.keyid, targets.delegations.roles["role1"].keyids)
+        targets.add_key("role1", key)
+
+        # Assert that the new key is added to the delegated role "role1"
+        self.assertIn(key.keyid, targets.delegations.roles["role1"].keyids)
+
+        # Confirm that the newly added key does not break the obj serialization
+        targets.to_dict()
+
+        # Try adding the same key again and assert its ignored.
+        past_keyid = targets.delegations.roles["role1"].keyids.copy()
+        targets.add_key("role1", key)
+        self.assertEqual(past_keyid, targets.delegations.roles["role1"].keyids)
+
+        # Try adding a key to a delegated role that doesn't exists
+        with self.assertRaises(ValueError):
+            targets.add_key("nosuchrole", key)
+
+        # Add the same key to "role2" as well
+        targets.add_key("role2", key)
+
+        # Remove the key from "role1" role ("role2" still uses it)
+        targets.remove_key("role1", key.keyid)
+
+        # Assert that delegated role "role1" doesn't contain the key.
+        self.assertNotIn(key.keyid, targets.delegations.roles["role1"].keyids)
+        self.assertIn(key.keyid, targets.delegations.roles["role2"].keyids)
+
+        # Remove the key from "role2" as well
+        targets.remove_key("role2", key.keyid)
+        self.assertNotIn(key.keyid, targets.delegations.roles["role2"].keyids)
+
+        # Try remove key not used by "role1"
+        with self.assertRaises(ValueError):
+            targets.remove_key("role1", key.keyid)
+
+        # Try removing a key from delegated role that doesn't exists
+        with self.assertRaises(ValueError):
+            targets.remove_key("nosuchrole", key.keyid)
+
+        # Remove delegations as a whole
+        targets.delegations = None
+        # Test that calling add_key and remove_key throws an error
+        # and that delegations is still None after each of the api calls
+        with self.assertRaises(ValueError):
+            targets.add_key("role1", key)
+        self.assertTrue(targets.delegations is None)
+        with self.assertRaises(ValueError):
+            targets.remove_key("role1", key.keyid)
+        self.assertTrue(targets.delegations is None)
 
 
     def  test_length_and_hash_validation(self):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -466,6 +466,10 @@ class TestMetadata(unittest.TestCase):
         # Add the same key to targets role as well
         root.signed.add_key('targets', key_metadata)
 
+        # Add the same key to a nonexistent role.
+        with self.assertRaises(ValueError):
+            root.signed.add_key("nosuchrole", key_metadata)
+
         # Remove the key from root role (targets role still uses it)
         root.signed.remove_key('root', keyid)
         self.assertNotIn(keyid, root.signed.roles['root'].keyids)
@@ -476,8 +480,10 @@ class TestMetadata(unittest.TestCase):
         self.assertNotIn(keyid, root.signed.roles['targets'].keyids)
         self.assertNotIn(keyid, root.signed.keys)
 
-        with self.assertRaises(KeyError):
+        with self.assertRaises(ValueError):
             root.signed.remove_key('root', 'nosuchkey')
+        with self.assertRaises(ValueError):
+            root.signed.remove_key('nosuchrole', keyid)
 
     def test_is_target_in_pathpattern(self):
         supported_use_cases = [
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -512,23 +512,6 @@ class TestMetadata(unittest.TestCase):
             )
 
 
-    def test_delegation_class(self):
-        # empty keys and roles
-        delegations_dict = {"keys":{}, "roles":[]}
-        delegations = Delegations.from_dict(delegations_dict.copy())
-        self.assertEqual(delegations_dict, delegations.to_dict())
-
-        # Test some basic missing or broken input
-        invalid_delegations_dicts = [
-            {},
-            {"keys":None, "roles":None},
-            {"keys":{"foo":0}, "roles":[]},
-            {"keys":{}, "roles":["foo"]},
-        ]
-        for d in invalid_delegations_dicts:
-            with self.assertRaises((KeyError, AttributeError)):
-                Delegations.from_dict(d)
-
     def test_metadata_targets(self):
         targets_path = os.path.join(
                 self.repo_dir, 'metadata', 'targets.json')
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -12,13 +12,11 @@ import logging
 import unittest
 import copy
 
-from typing import Dict, Callable, Optional, Mapping, Any
-from datetime import datetime
+from typing import Dict, Callable
 
 from tests import utils
 
 from tuf.api.metadata import (
-    Signed,
     Root,
     Snapshot,
     Timestamp,
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -305,6 +305,11 @@ class TestSerialization(unittest.TestCase):
 
 
     invalid_delegations: DataSet = {
+        "empty delegations": '{}',
+        "bad keys": '{"keys": "foo", \
+            "roles": [{"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": false, "threshold": 3}]}',
+        "bad roles": '{"keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": ["foo"]}',
         "duplicate role names": '{"keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
             "roles": [ \
                 {"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": false, "threshold": 3}, \
@@ -316,7 +321,7 @@ class TestSerialization(unittest.TestCase):
     @run_sub_tests_with_dataset(invalid_delegations)
     def test_invalid_delegation_serialization(self, test_case_data: str):
         case_dict = json.loads(test_case_data)
-        with self.assertRaises(ValueError):
+        with self.assertRaises((ValueError, KeyError, AttributeError)):
             Delegations.from_dict(copy.deepcopy(case_dict))
 
 
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -304,6 +304,22 @@ class TestSerialization(unittest.TestCase):
             DelegatedRole.from_dict(copy.copy(case_dict))
 
 
+    invalid_delegations: DataSet = {
+        "duplicate role names": '{"keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": false, "threshold": 3}, \
+                {"keyids": ["keyid2"], "name": "a", "paths": ["fn3"], "terminating": false, "threshold": 2}    \
+                ] \
+            }',
+    }
+
+    @run_sub_tests_with_dataset(invalid_delegations)
+    def test_invalid_delegation_serialization(self, test_case_data: str):
+        case_dict = json.loads(test_case_data)
+        with self.assertRaises(ValueError):
+            Delegations.from_dict(copy.deepcopy(case_dict))
+
+
     valid_delegations: DataSet = {
         "all":
             '{"keys": { \
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -358,10 +358,7 @@ class Metadata(Generic[T]):
                 raise ValueError(f"No delegation found for {delegated_role}")
 
             keys = self.signed.delegations.keys
-            roles = self.signed.delegations.roles
-            # Assume role names are unique in delegations.roles: #1426
-            # Find first role in roles with matching name (or None if no match)
-            role = next((r for r in roles if r.name == delegated_role), None)
+            role = self.signed.delegations.roles.get(delegated_role)
         else:
             raise TypeError("Call is valid only on delegator metadata")
 
@@ -1146,16 +1143,17 @@ class Delegations:
 
     Attributes:
         keys: Dictionary of keyids to Keys. Defines the keys used in 'roles'.
-        roles: List of DelegatedRoles that define which keys are required to
-            sign the metadata for a specific role. The roles order also
-            defines the order that role delegations are considered in.
+        roles: Ordered dictionary of role names to DelegatedRoles instances. It
+            defines which keys are required to sign the metadata for a specific
+            role. The roles order also defines the order that role delegations
+            are considered in.
         unrecognized_fields: Dictionary of all unrecognized fields.
     """
 
     def __init__(
         self,
         keys: Dict[str, Key],
-        roles: List[DelegatedRole],
+        roles: "OrderedDict[str, DelegatedRole]",
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
         self.keys = keys
@@ -1170,17 +1168,19 @@ class Delegations:
         for keyid, key_dict in keys.items():
             keys_res[keyid] = Key.from_dict(keyid, key_dict)
         roles = delegations_dict.pop("roles")
-        roles_res = []
+        roles_res: "OrderedDict[str, DelegatedRole]" = OrderedDict()
         for role_dict in roles:
             new_role = DelegatedRole.from_dict(role_dict)
-            roles_res.append(new_role)
+            if new_role.name in roles_res:
+                raise ValueError(f"Duplicate role {new_role.name}")
+            roles_res[new_role.name] = new_role
         # All fields left in the delegations_dict are unrecognized.
         return cls(keys_res, roles_res, delegations_dict)
 
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self."""
         keys = {keyid: key.to_dict() for keyid, key in self.keys.items()}
-        roles = [role_obj.to_dict() for role_obj in self.roles]
+        roles = [role_obj.to_dict() for role_obj in self.roles.values()]
         return {
             "keys": keys,
             "roles": roles,
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1385,3 +1385,34 @@ class Targets(Signed):
     def update(self, fileinfo: TargetFile) -> None:
         """Assigns passed target file info to meta dict."""
         self.targets[fileinfo.path] = fileinfo
+
+    def add_key(self, role: str, key: Key) -> None:
+        """Adds new signing key for delegated role 'role'.
+
+        Raises:
+            ValueError: If there are no delegated roles or if 'role' is not
+                delegated by this Target.
+        """
+        if self.delegations is None or role not in self.delegations.roles:
+            raise ValueError(f"Delegated role {role} doesn't exist")
+        self.delegations.roles[role].keyids.add(key.keyid)
+        self.delegations.keys[key.keyid] = key
+
+    def remove_key(self, role: str, keyid: str) -> None:
+        """Removes key from delegated role 'role' and updates the delegations
+        key store.
+
+        Raises:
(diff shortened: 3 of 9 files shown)


tuf/api/metadata.py
```diff
@@ -343,7 +343,9 @@ def verify_delegate(
             roles = self.signed.delegations.roles
             # Assume role names are unique in delegations.roles: #1426
             # Find first role in roles with matching name (or None if no match)
-            role = next((r for r in roles if r.name == delegated_role), None)
+            role = next(
+                (r for r in roles.values() if r.name == delegated_role), None
+            )
```
jku: I would imagine you would now have a slightly easier way of looking up delegated roles by name? :grin:

please update the comment as well

MVrachev: Yeaa... you are right. Updated the pr.

tuf/api/metadata.py
```diff
@@ -1090,16 +1090,19 @@ class Delegations:
...
         unrecognized_fields: Dictionary of all unrecognized fields.
     """
 
+    # TODO: Change "roles" annotation to OrderedDict when we drop python 3.6.
+    # See https://docs.python.org/3/library/typing.html#typing.OrderedDict
     def __init__(
         self,
         keys: Dict[str, Key],
-        roles: List[DelegatedRole],
+        roles: Dict[str, DelegatedRole],
```
jku: Isn't this the exact opposite of what we want: We currently support 3.6 which does _not_ guarantee that Dict is ordered. So we need to use OrderedDict at least as long as we support 3.6

I don't think a TODO is needed in any case

jku:
> It seems logical to me that those functions will be part of the
> "Delegations" class instead of "Targets" as those helper functions are
> related to delegated roles.

This is true. The argument against is that 
* root and targets are both delegators with same key functionality... but now you'd access the functionality differently in those two
* users of the library already have to learn about Metadata and Targets. If we push API top-level functionality into these helpers it means users have to also learn all the little helper classes. In some cases this is unavoidable (e.g. Key) but Delegations feels like an implementation detail of Targets that maybe no-one (except us) really needs to know about?

tuf/api/metadata.py
MVrachev: The point is we don't have `OrderDict` as a typing annotation before python version 3.7.2 and if we run this code
with python 3.6 this will lead to an error as it won't import `OrderDict` from the `typing` module.
That's why I added to `TODO` item as well.

MVrachev:
> > It seems logical to me that those functions will be part of the
> > "Delegations" class instead of "Targets" as those helper functions are
> > related to delegated roles.
> 
> This is true. The argument against is that
> 
> * root and targets are both delegators with same key functionality... but now you'd access the functionality differently in those two
> * users of the library already have to learn about Metadata and Targets. If we push API top-level functionality into these helpers it means users have to also learn all the little helper classes. In some cases this is unavoidable (e.g. Key) but Delegations feels like an implementation detail of Targets that maybe no-one (except us) really needs to know about?

Hmm yes, you are right about that. We don't want to make our users know each of the helper classes.
Then, if the key helper functions are used in Targets, then at least shouldn't we use more clarifying names for them?
If you call `Targets.add_key()` then where are we adding this key? Yes, there is function documentation, but wonder if
names like `add_delegated_key()` and `remove_delegated_key()` (which are both a lot longer :( ) will be better?

tuf/api/metadata.py
jku: _**EDIT**: didn't realize you did use an actual OrderedDict but annotated as Dict. below comment is misinformed_

python 3.6 Dict is **not guaranteed to be ordered**. We can't use it. If you can't figure out how to use OrderedDict then we  won't do this yet: using a non-ordered dictionary when we absolutely need a ordered one is not an option.

But I would ask you to research this a bit more. [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) has been part of collections for a long time, I'd like to believe you can use it on 3.6. There may indeed be an issue with _annotations_ on older python but if indeed that is the case, then we try to solve the annotation issue separately.

jku: oh you did! sorry :1st_place_medal: 

Hmm, I wonder which is better: annotating as `OrderedDict` (not generic) or as `Dict[str, DelegatedRole]`... I guess both will work

jku: Another idea, what happens if you use `"OrderedDict[str, DelegatedRole]"` (so quoted typing)

MVrachev: Interesting. When quoting the whole like you showed `"OrderedDict[str, DelegatedRole]"` I don't have to import `OrderDict`, because we will use [Forward reference](https://www.python.org/dev/peps/pep-0484/#forward-references) and all types of `NamedErrors` will be suppressed.
This way the code works both on my python3.6 vevn and python3.8.

What is interesting is that when using a type hint inside a function as I did on line 1120, I don't have to quote the `OrderDict` type hint. It's accepted.

_**EDIT**_ Even though the code runs fine if I don't quote `OrderDict` type hint at line 1120, still `pylint` gives me a warning. Will add a quote there too.

tests/test_metadata_serialization.py
```diff
@@ -12,13 +12,11 @@
 import unittest
 import copy
 
-from typing import Dict, Callable, Optional, Mapping, Any
-from datetime import datetime
+from typing import Dict, Callable
```
sechkova, line 15: Since you've started doing this, can you remove also the unused Union from here 😁: https://github.com/theupdateframework/tuf/blob/7731738590f291041c06b3c525e248d245cb1f33/tests/test_trusted_metadata_set.py#L2

MVrachev: Did that and additionally removed a couple of other imports from other places.

MVrachev:
Inspired by #1553 commit "tests: Improve add_key/remove_key API tests" I added additional commit testing
the same thing as this commit does.
Also, I had to rebase.

tuf/api/metadata.py
```diff
@@ -1153,23 +1154,51 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
...
+        roles_res: "OrderedDict[str, DelegatedRole]" = OrderedDict()
         for role_dict in roles:
             new_role = DelegatedRole.from_dict(role_dict)
-            roles_res.append(new_role)
+            role_name = new_role.name
+            if roles_res.get(role_name) is not None:
+                raise ValueError(
+                    f"Delegated role names should be unique! "
+                    f"Got a role name '{role_name}' which is duplicate!"
+                )
```
jku: I'm confident you could squeeze the error into a single line

Creating temporary variables like role_name is not a bad practice when the variable makes things more readable or prevents multiple dictionary lookups... I'm not sure there is a reason in this case?

jku (commented):
> Then, if the key helper functions are used in Targets, then at least shouldn't we use more clarifying names for them?
> If you call `Targets.add_key()` then where are we adding this key? Yes, there is function documentation, but wonder if
> names like `add_delegated_key()` and `remove_delegated_key()` (which are both a lot longer :( ) will be better?

I don't think Targets.add_key() is a bad name. If the reader is confused about delegators and delegated roles we likely aren't going to fix that with function names: they could misread your suggestion the same way _"oh Targets.add_delegated_key()! this is the function I use to add the key that delegates from root to targets"_

MVrachev:
> > Then, if the key helper functions are used in Targets, then at least shouldn't we use more clarifying names for them?
> > If you call `Targets.add_key()` then where are we adding this key? Yes, there is function documentation, but wonder if
> > names like `add_delegated_key()` and `remove_delegated_key()` (which are both a lot longer :( ) will be better?
> 
> I don't think Targets.add_key() is a bad name. If the reader is confused about delegators and delegated roles we likely aren't going to fix that with function names: they could misread your suggestion the same way _"oh Targets.add_delegated_key()! this is the function I use to add the key that delegates from root to targets"_

I can't say I am a fan of `Targets.add_key()` as an API name, but it seems there is no better option. I thought this name could be misleading for Targets compared to Root because in Root there is a top-level `keyids` dictionary and IMO this makes it clearer.
Will move the `add_key` API call to Targets.

tuf/api/metadata.py
MVrachev: Addressed both the temporary variable and made the comment one line.

MVrachev:
I rebased on top of develop and had to resolve a couple of conflicts.
Additional changes I made are:
- renamed the pr
- moved the `add_key` and `remove_key` functions to `Targets`
- added a check if `delegations` is `None` reminded by `mypy` that that is a possible case when calling the key helpers
- created a new test function `test_targets_key_api` because the use cases became a lot

I was wondering what should we do if somebody calls `targets.add_key()` or `targets.remove_key()` and `delegations` is `None`.
It seemed logical to me that we can just ignore such calls.
What do the other think?

tests/test_api.py
```diff
@@ -507,23 +507,6 @@ def test_is_target_in_pathpattern(self):
...
-    def test_delegation_class(self):
-        # empty keys and roles
-        delegations_dict = {"keys":{}, "roles":[]}
-        delegations = Delegations.from_dict(delegations_dict.copy())
-        self.assertEqual(delegations_dict, delegations.to_dict())
-
-        # Test some basic missing or broken input
-        invalid_delegations_dicts = [
-            {},
-            {"keys":None, "roles":None},
```
sechkova, line 524: Don't we care about this case anymore? It is a bit different than `"empty delegations": '{}'`

tuf/api/metadata.py
```diff
@@ -1301,3 +1301,30 @@ def to_dict(self) -> Dict[str, Any]:
...
         self.targets[fileinfo.path] = fileinfo
+
+    def add_key(self, role: str, key: Key) -> None:
+        """Adds new signing key for delegated role 'role'.
+
+        Raises:
+            KeyError: If there is no role 'role' in the Delegations object.
+        """
+        if self.delegations is None:
+            return
```
sechkova: We shouldn't  silently ignore failures. I think an exception should be raised here. `exceptions.py `defines UnknownRoleError (as well as RoleAlreadyExistsError for the uniqueness check but a ValueError should be ok too).

tuf/api/metadata.py
```diff
@@ -1301,3 +1301,30 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+    def remove_key(self, role: str, keyid: str) -> None:
+        """Removes key from delegated role 'role' and updates the delegations
+        key store.
+
+        Raises:
+            KeyError: If there is no role 'role' in the Delegations object.
+        """
+        if self.delegations is None:
+            return
```
sechkova: Same thing as add_key.

tests/test_api.py
MVrachev: Testing that delegations are `{"keys":None, "roles":None}` can't be done with the current code and I don't see a reason to test for it, because in JSON there is no such thing as `None` and we are calling `case_dict = json.loads(test_case_data)` before making the actual tests which throws the `json.decoder.JSONDecodeError` exception.

MVrachev:
> We shouldn't silently ignore failures. I think an exception should be raised here. exceptions.py defines UnknownRoleError (as well as RoleAlreadyExistsError for the uniqueness check but a ValueError should be ok too).

Now as you said it makes sense to warn the user that nothing happened when calling `add_key` or `remove_key`. 
Added a `UnknownRole` exception there.
Additionally, I swapped the `ValueError` exception with `RoleAlreadyExistsError` which seems more concrete for that case.

sechkova:
> Additionally, I swapped the `ValueError` exception with `RoleAlreadyExistsError` which seems more concrete for that case.

I'm afraid someone may have a different opinion on this one :D @jku

tuf/api/metadata.py
```diff
@@ -1301,3 +1301,32 @@ def to_dict(self) -> Dict[str, Any]:
...
+    def add_key(self, role: str, key: Key) -> None:
+        """Adds new signing key for delegated role 'role'.
+
+        Raises:
+            UnknownRoleError: If there are no 'delegations' in the Targets obj.
+            KeyError: If there is no role 'role' in the Delegations object.
+        """
+        if self.delegations is None:
+            raise exceptions.UnknownRoleError("Delegations doesn't exists!")
+        self.delegations.roles[role].keyids.add(key.keyid)
```
jku: * now we get two different errors when a nonexisting role is used (UnknownRoleError if delegations does not exist, and KeyError if it does). This is silly: no user of the API wants that.
* Also (@sechkova knows me) we should only use custom errors when they provide actual value -- when callers might want to catch specifically that error and the base library errors would be too vague. The error message can always be used to inform the developer what they are doing wrong

There's basically two slightly overlapping classes of exceptions:

1. the ones we expect the caller to handle: these should be carefully designed and documented (and here custom exceptions _may_ provide added value)
2. the ones that represent caller programming error: these especially should be standard base library exceptions.

this case is mostly in group 2: adding a key for a non-existing role is unlikely to be handled in calling code. The "correct" error would be ValueError I guess but the two nonexisting role cases listed above should use the same error -- in that context I don't think KeyError is a bad choice (since we get that for free when delegations does exist).

jku: also, the error here (from callers perspective) isn't _"Delegations doesn't exists!"_: that's an implementation detail. The error is that caller tried to add a key to a role that does not exist.

tuf/api/metadata.py
```diff
@@ -1301,3 +1301,32 @@ def to_dict(self) -> Dict[str, Any]:
...
+        self.delegations.keys[key.keyid] = key
+
+    def remove_key(self, role: str, keyid: str) -> None:
+        """Removes key from delegated role 'role' and updates the delegations
+        key store.
+
+        Raises:
+            UnknownRoleError: If there are no 'delegations' in the Targets obj.
+            KeyError: If there is no role 'role' in the Delegations object.
+        """
```
jku: I don't think we need to talk about the Delegations object here at all: it's 99% an implementation detail.

There's two errors that caller might make: 
* role is not delegated by this Targets
* key is not actually used by role

Both maybe worth a mention in the docstring

jku (changes requested):

jku:
I think in the future I'd suggest not combining issues like these two: If this had been two separate PRs it would have meant a bit more work for you (to rebase the second PR and to figure out when to put the second PR to review) but it would have meant reviewers need less context, can do faster reviews and the first PR would land faster.

tuf/api/metadata.py
```diff
@@ -1153,17 +1154,21 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
...
         roles = delegations_dict.pop("roles")
-        roles_res = []
+        roles_res: "OrderedDict[str, DelegatedRole]" = OrderedDict()
         for role_dict in roles:
             new_role = DelegatedRole.from_dict(role_dict)
-            roles_res.append(new_role)
+            if roles_res.get(new_role.name) is not None:
+                raise exceptions.RoleAlreadyExistsError(
+                    f"Delegated role names must be unique! Got {new_role.name}"
+                )
```
jku: This is code that gets called from serialization (which wraps it in a catch all try-except) so using custom exceptions seems not useful.

repository code that wants to add roles is now going to use  `targets.delegation.roles[role.name] = role` so the exception won't be useful in that case either, as far as I can see...

We might want to provide some ways to easily order the dict at some point... but even then I don't see when a custom error would be useful

MVrachev:
> I think in the future I'd suggest not combining issues like these two: If this had been two separate PRs it would have meant a bit more work for you (to rebase the second PR and to figure out when to put the second PR to review) but it would have meant reviewers need less context, can do faster reviews and the first PR would land faster.

Yes, I guess you are right. I thought I can combine fixing those two issues and that you will have the context given that 
you suggested that here https://github.com/theupdateframework/python-tuf/issues/1469#issuecomment-871271393 and both of the issues are submitted by you.

tuf/api/metadata.py
MVrachev: Okay, will change the exception to `ValueError`  and will shorten the message.

tuf/api/metadata.py
MVrachev: Maybe the best solution is just to use `KeyError` as it's the simplest one and we are already throwing it as you said.

tuf/api/metadata.py
MVrachev: If the role is not delegated by this Target we can throw ValueError as the given "role" is of wrong value.
Will update the docstring.

If the key is not used by the given role, then there is nothing to be removed. I don't think we need to add any documentation for that.

jku: If the key is not used by the given role, we do raise something. ValueError probably?

MVrachev: Yes, I am mistaken. We do throw a `KeyError` when trying to delete on line 1332.
I will update the docstring and change the exception to ValueError.

tuf/api/metadata.py
```diff
@@ -1299,3 +1299,34 @@ def to_dict(self) -> Dict[str, Any]:
     def update(self, fileinfo: TargetFile) -> None:
         """Assigns passed target file info to meta dict."""
         self.targets[fileinfo.path] = fileinfo
+
+    def add_key(self, role: str, key: Key) -> None:
+        """Adds new signing key for delegated role 'role'.
+
+        Raises:
+            KeyError: If there are no delegations, delegate role 'role' or key
+                with id `key.keyid`.
```
sechkova: > "or key with id `key.keyid`."

  Where would this error come from? Aren't we adding the key with key.keyid?

tuf/api/metadata.py
```diff
@@ -1299,3 +1299,34 @@ def to_dict(self) -> Dict[str, Any]:
...
+            raise KeyError("Delegations doesn't exists!")
+        self.delegations.roles[role].keyids.add(key.keyid)
+        self.delegations.keys[key.keyid] = key
+
+    def remove_key(self, role: str, keyid: str) -> None:
+        """Removes key from delegated role 'role' and updates the delegations
+        key store.
+
+        Raises:
+            ValueError: If there are no delegated roles or if 'role' is not
```
sechkova, line 1417: How did we end up with two equivalent functions throwing `ValueError` and `KeyError` for the same thing. I am not sure about this.

tuf/api/metadata.py
MVrachev: Oops, mistake. Removing this.

tuf/api/metadata.py
MVrachev: Now as you highlighted it, it's not great we throw two different types of exceptions for `Root.remove_key()` - `KeyError` and `Targets.remove_key()` - `ValueError`.
IMO the correct exception here should be `ValueError` as the problems are in the given **values** -`role` doesn't exist or `keyid` doesn't point to a key used by that role.
@jku and @sechkova if you both agree I can change the `Targets.remove_key()` exception to be `ValueError` as well.

tuf/api/metadata.py
```diff
@@ -340,9 +340,9 @@ def verify_delegate(
 
             keys = self.signed.delegations.keys
             roles = self.signed.delegations.roles
-            # Assume role names are unique in delegations.roles: #1426
-            # Find first role in roles with matching name (or None if no match)
-            role = next((r for r in roles if r.name == delegated_role), None)
+            # Role names are unique in delegations.roles
+            # Find role in roles with matching name (or None if no match).
+            role = roles.get(delegated_role)
```
jku: nit: I think these comments are not needed anymore: they now just explain how a dict works

Is the temporary variable `roles` needed -- it looks like it might be readable without?

tuf/api/metadata.py
```diff
@@ -1153,17 +1154,19 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
         for keyid, key_dict in keys.items():
             keys_res[keyid] = Key.from_dict(keyid, key_dict)
         roles = delegations_dict.pop("roles")
-        roles_res = []
+        roles_res: "OrderedDict[str, DelegatedRole]" = OrderedDict()
         for role_dict in roles:
             new_role = DelegatedRole.from_dict(role_dict)
-            roles_res.append(new_role)
+            if roles_res.get(new_role.name) is not None:
```
jku: this works (and is probably used elsewhere) but I guess the pythonic way is
`if new_role.name in roles_res:`

tuf/api/metadata.py
jku: (EDIT: this is a response to your  comment, see context in the discussion)

I suppose you mean _change Targets.add_key() exception_: sounds good to me. ValueError is no doubt the correct error (but it mostly matters for aesthetics: this is a programmer error not an error that will be handled)

btw: you don't need to say _"If there are no delegated roles"_ in the docstring: that is an implementation detail the caller does not need to know: for them the error is _"'role' is not delegated by this Targets"_

tuf/api/metadata.py
```diff
@@ -1296,3 +1299,33 @@ def to_dict(self) -> Dict[str, Any]:
...
+    def remove_key(self, role: str, keyid: str) -> None:
+        """Removes key from delegated role 'role' and updates the delegations
+        key store.
+
+        Raises:
+            ValueError: If there are no delegated roles or if 'role' is not
+                delegated by this Target or if key is not used by 'role'.
+        """
+        if self.delegations is None or self.delegations.roles.get(role) is None:
+            raise ValueError(f"Delegated role {role} doesn't exists!")
```
jku: very minor nit: exclamations are not needed in errors, s/exists/exist/

jku (commented):
Looks good, left some style comments: probably none of them absolutely necessary.

tuf/api/metadata.py
MVrachev: Removed the comments.
Yea, I think we can remove `roles`. It looks fine without it.

tuf/api/metadata.py
MVrachev: I made an API change to `Root.add_key()` and `Root.remove_key()` so that they throw `ValueError` instead of `KeyError` on wrong given arguments.
The reason for this change is that `ValueError` in my opinion is the right exception that should be thrown when the value is not correct and I think it's important to be consistent when calling `delegator.add_key` or delegator.remove_key` no matter if the delegator is `Root` or `Targets`.

tuf/api/metadata.py
```diff
@@ -1296,3 +1296,33 @@ def to_dict(self) -> Dict[str, Any]:
...
         self.targets[fileinfo.path] = fileinfo
+
+    def add_key(self, role: str, key: Key) -> None:
+        """Adds new signing key for delegated role 'role'.
+
+        Raises:
+            ValueError: If there are no delegations or delegate role 'role'.
+        """
+        if self.delegations is None or role not in self.delegations.roles:
+            raise ValueError("Delegations doesn't exist")
```
jku: This is not the correct error message.

Also I still think the docstrings should not talk about implementation details (whether delegations is none or not) here or in remove_key(): they should talk about what the callers mistake was -- but this is a nitpick.

tuf/api/metadata.py
```diff
@@ -741,16 +741,24 @@ def to_dict(self) -> Dict[str, Any]:
...
 
     def remove_key(self, role: str, keyid: str) -> None:
         """Removes key from 'role' and updates the key store.
 
         Raises:
-            KeyError: If 'role' does not include the key
+            ValueError: If 'role' does not include the key.
         """
+        if keyid not in self.roles[role].keyids:
+            raise ValueError(f"Key with id {keyid} is not used by {role}")
```
jku: If you change the API, please make it match the new one you added for Targets. nonexisting role still raises a keyerror here and ValueError in Targets

jku (changes requested):
Looks ok, still have two comments on the errors though.

I'm fine with the API change (KeyError -> ValueError) even if I usually would be against changing API with no practical benefit: the errors in question are IMO clear programming mistakes so we don't expect there to exist code that might break if we change the errors, but more consistent errors should be useful to developers making the programming mistake.

jku:
If you can rebase as well that would be great (the workflow jobs changed name and branch protection will now think you haven't passed the tests)

tuf/api/metadata.py
MVrachev: Yes, you are correct.
I didn't see that case before.

jku (approved):
I think this is a correct change, thanks.

Commit messages do not make clear that the Delegations API has a breaking change, we should get better at this as we can't expect the release notes writer to always make the connection... Anyway LGTM: we can mention API change in the merge commit maybe?

MVrachev:
> I think this is a correct change, thanks.
> 
> Commit messages do not make clear that the Delegations API has a breaking change, we should get better at this as we can't expect the release notes writer to always make the connection... Anyway LGTM: we can mention API change in the merge commit maybe?

Now it should be clearer and I am not sure what will be the value in adding a note about the API changes anymore.

PR #1539, opened 2021-08-24
Fixes #1526 to atomic write metadata operation

### LINKED ISSUE #1526: ngclient: persist metadata in a safer way

suvadityamuk:

that is provided, the better.

Fixes #1526 

This is an atomic-write operation where a temp file is first written to and atomically re-placed on top of the old original file.

requirements**:


tuf/ngclient/updater.py
```diff
@@ -292,8 +293,15 @@ def _load_local_metadata(self, rolename: str) -> bytes:
...
 
     def _persist_metadata(self, rolename: str, data: bytes):
-        with open(os.path.join(self._dir, f"{rolename}.json"), "wb") as f:
-            f.write(data)
+        """To ensure there is no chance for loss of data during writing, the data is first written to a
+        temp file followed by an atomic move operation to the correct file location to make sure that 
+        if data writing process is halted or interrupted, the original data is not lost.
+        """
+        
+        original_filename = open(f"{rolename}.json", 'a+')
```
trishankatdatadog: I'm not sure we support f-strings yet. Maybe best to stick to our current convention... which is String.format()? @jku

tuf/ngclient/updater.py
```diff
@@ -292,8 +293,15 @@ def _load_local_metadata(self, rolename: str) -> bytes:
...
+        file_out = tempfile.NamedTemporaryFile(dir=os.path.dirname(original_filename), delete=False)
```
trishankatdatadog: Hmm, don't we want to clean up after temp files after we are done moving?

suvadityamuk: If there is any problem or the original file is somehow corrupted during any operation, the temp file will act as backup. That's the idea behind keeping the temp file intact, as I can see.

tuf/ngclient/updater.py
suvadityamuk: The original code that I pulled my first fork out of used `f-strings` in place, that is why I thought this would be the correct way to go about it. 
When I looked at the tests and where they failed, it seems my mistake is the fact that I am passing an `iowrapper` to the `NamedTemporaryFile` function where it expects a `str`, so I think that change could possibly solve the thing. 
But I really do not understand why is it that these errors do not come up when I run `tox` or the `aggregate_tests.py` functions on my system.

jku:
You can use the same PR for multiple revisions of the PR (force-pushing a new version of the same branch is fine by me): this would be better so we don't lose discussion in the previous PRs

some remaining issues:
 * there's still multiple commits in the branch for a single change
 * I don't understand why original_file_touch is opened?
 * you removed "with NamedTemporaryFile" in this version: that means the file descriptor is open when we call os.replace(): this will lead to failures on Windows

I expect something like this would work
```
    with NamedTemporaryFile(dir=self._dir, delete=False) as tmpfile:
        tmpfile.write(data)
    os.replace(tmpfile.name, file_name)
```

the gains I'm hoping to get are:
 * NamedTemporaryFile selects a safe filename, tmpfile is on same filesystem as file_name
 * "with" closes the tmpfile file descriptor
 * os.replace should work even on windows because there are now no open file descriptors
 * after replace tmpfile does not exist anymore
 * most importantly the actual file can only be either original content or complete new content after os.replace
 
 so the only issue I see is that if tmpfile.write or os.replace fails, the tempfile probably still exists: but i'm fine with that side effect

jku:
The _Run TUF tests and linter / build (3.x, ubuntu-latest, with-sslib-master, true)_ failure is not your fault btw, it's an unrelated issue

suvadityamuk:
That error was scary, thanks for letting me know. The others, I have made the change. I did not know all too well about the Windows side-effects so thank you once again for that. 
I had originally thought that there was some case where the file was not created at all and hence it was unable to detect or overwrite it or something along those lines.  Seems like I was pretty off there as well.

suvadityamuk:
Hey, do check and let me know how I can better this?

jku (changes requested):
remaining issues:
* there's still multiple commits in the branch for a single change: I'd rather see all squashed into one commit. If you are having trouble with this you may want read up on  "interactive rebase" and "squashing": https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History (and as I mentioned, force pushing a modified branch is fine by me)
* the final filename/path has been changed from `os.path.join(self._dir, f"{rolename}.json")` to `f"{rolename}.json"`: this is incorrect

the tests really _should_ have failed on that second issue, but it looks like we don't test for this currently (the ngclient updater tests are a work in progress)... I did not understand what was happening myself so had to write a quick test to figure it out: it's in https://github.com/jku/tuf/commit/1250890ff7ba25fd0bbb39910c67d9fa0c53385e : feel free to use that change (or if that feels difficult, I can combine your PR and this commit once the PR is otherwise good)

suvadityamuk:
@jku Is this fine?

MVrachev:
Thanks @suvadityamuk that you continue the pr refinement!
Could you please squash the two commits that are left into one?
When you do, that please rewrite the commit message of the resulted commit , so that it doesn't contain merge of the messages together (as it's the case for commit https://github.com/theupdateframework/tuf/pull/1539/commits/eccdf2b97108e40307f8578fb62016364fd1a37d0).

Additionally, please note that you will have to rebase on top of the `develop` branch and fix the conflicting cases.

suvadityamuk:
Hi. I think I have rebased the current commits on top of the `develop` branch, but I am really unable to figure out how do I squash the previous commits because they have somehow just cleared off and I cannot find them in the `git rebase -i HEAD`. Is there any chance you could excuse this or something of that sort?

PR #1549, opened 2021-08-29
Fixes #1526 to make the persist_metadata function an atomic operation

### LINKED ISSUE #1526: ngclient: persist metadata in a safer way

suvadityamuk:

that is provided, the better.

Fixes #1526 

Makes use of the `tempfile.NamedTemporaryFile` to make the `_persist_metadata` function more resistant to interruptions.

requirements**:

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -62,6 +62,7 @@ Example::
 
 import logging
 import os
+import tempfile
 from typing import List, Optional, Set, Tuple
 from urllib import parse
 
@@ -290,8 +291,17 @@ class Updater:
             return f.read()
 
     def _persist_metadata(self, rolename: str, data: bytes):
-        with open(os.path.join(self._dir, f"{rolename}.json"), "wb") as f:
-            f.write(data)
+        """Acts as an atomic write operation to make sure
+        that if the process of writing is halted, at least the
+        original data is left intact.
+        """
+
+        original_filename = os.path.join(self._dir, f"{rolename}.json")
+        with tempfile.NamedTemporaryFile(
+            dir=self._dir, delete=False
+        ) as temp_file:
+            temp_file.write(data)
+        os.replace(temp_file.name, original_filename)
 
     def _load_root(self) -> None:
         """Load remote root metadata.


suvadityamuk:
@jku I am sorry but I made quite a few missteps while trying to do the rebasing and the entire history looked really weird to repair/change. Hence, I closed the previous PR and made this fresh one, written on top of the latest code of the original `develop` branch. This has all the changes made into it, tested successfully via Tox, has linter-tests passing and clears all aggregate tests.
Once again, sorry for the PR mess here, but this seemed like the best thing to do given the circumstances.

tuf/ngclient/updater.py
```diff
@@ -290,8 +291,17 @@ def _load_local_metadata(self, rolename: str) -> bytes:
...
+        that if the process of writing is halted, at least the
+        original data is left intact.
+        """
+
+        original_filename = os.path.join(self._dir, f"{rolename}.json")
+        with tempfile.NamedTemporaryFile(
+            dir=self._dir, delete=False
+        ) as temp_file:
+            temp_file.write(data)
+        os.replace(temp_file.name, original_filename)
```
trishankatdatadog, line 301: I'm still a little concerned about not proactively deleting temporary files. One can imagine a very long-lived process clogging up disk space.

suvadityamuk: Well then, should I flip the `delete=False` statement? @jku Please comment as well.

jku: So just to document what happens here: 
* if `write` and `os.replace` both succeed (as we expect), there will be no temporary files left on disk
* we cannot set `delete=True` because you can't do both that and os.replace -- that does not make sense
* Not being able to write metadata to disk is, at least currently, a fatal error: Updater will not continue

I'm not objecting to handling write and os.replace errors as well but I think doing so makes the function significantly more complex and hard to test, something like this:

```python
try:
    # all current code  (tempfile, write, os.replace) here
except OSError as e:
    # before raising, delete tempfile if it exists
    if tempfile:
        try:
            os.remove(tempfile.name)
        except FileNotFoundError:
            pass
    raise e
```

Testing this kind of code (in a way that actually checks that I got it correct) can be a bit of a pain.... Now that I think about it, making `original_filename` read-only would test at least one case, maybe that would be enough (_EDIT: of course it would not: replacing files does not require write access to original file_). Hmmm.

jku:
> Once again, sorry for the PR mess here

No problem, thanks for your persistence. PRs being messy is totally fine: what matters in the end is the branch that gets merged. If you look through the PRs in this repo you will find PRs with dozens or even hundreds of comments and loads of pushed versions that were broken or wrong: this is just how the sausage gets made.

The current branch implements what I asked for (thanks)... but let's consider Trishanks reasonable worry about temp files in error situations. I'll think a bit and post a suggestion.

suvadityamuk:
Understood, @jku

jku:
My stance:
* Trishanks concern is legitimate: I can't come up with a practical example but I can imagine a frequently updating client that has a bug or configuration issue that leads to every refresh failing to write, creating new temp files...
* Current situation is arguably worse: if a root.json write fails and leaves the file empty, that bricks the client for good.

I suggest we merge this and file a new issue for "ngclient: make sure temp files are deleted in error situations".

Unfortunately while we discussed I merged another change that has a slight conflict with this (the method annotation changed). @suvadityamuk could you rebase on top of current develop branch once more and resolve the conflict? Do ask for advice if this is unclear.

Thanks

suvadityamuk:
I could, actually. Could you tell me how I could go about doing that? Thank you!

jku:
The git book is not a horrible resource: https://git-scm.com/book/en/v2 I suggest looking at rebasing and [basic merge conflicts](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging#_basic_merge_conflicts)

if you don't know how to get started, this is what I'd do (assuming "origin" is the name for this repository remote):
```bash
git fetch origin
git rebase origin/develop
# you'll get a "Merge conflict": you'll have to edit updater.py _persist_metadata() so that it only has your 
# version with the added "-> None" in the method signature. When this is done, test:
tox -e lint
tox -e py39
# if tests pass, continue rebase
git add tuf/ngclient/updater.py
git rebase --continue
# rebase should succeed, your branch should look good and pass tests: push it with "--force" to replace the previous version
```

suvadityamuk:
Okay, will follow the steps and do it. Thanks.

trishankatdatadog:
Thanks for helping out a first-time contributor, @jku, very kind of you.

@suvadityamuk another way to resolve conflicts is to use the web UI, it should give you a nice GUI.

image

suvadityamuk:
Ahh, I did notice that and wanted to use it, but wasn't too sure if that would end up causing irreparable problems, haha. And yes, Thanks a ton to @jku @MVrachev and @trishankatdatadog for helping me out. I really look forward to contributing a lot here, this being only the beginning.

trishankatdatadog:
> Ahh, I did notice that and wanted to use it, but wasn't too sure if that would end up causing irreparable problems, haha.

It probably would except for simple conflicts. Best to stick to manual rebase.

> I really look forward to contributing a lot here, this being only the beginning.

Looking fwd!

suvadityamuk:
Do let me know if this is fine. I have done the Web UI resolve because that was pretty much the simplest way

trishankatdatadog:
> Do let me know if this is fine. I have done the Web UI resolve because that was pretty much the simplest way

It's not the cleanest history, but we can always squash and merge as a single commit.

suvadityamuk:
Since I resolved the conflict with the Web UI, it seems I cannot see that specific commit using git bash on my system for some reason. Is there a chance you can merge this itself? I'll make sure this won't happen from the next time.

suvadityamuk:
Hi, any updates?

jku:
> Is there a chance you can merge this itself?

Thanks let's do that. As seen in the test results the rebase in this PR wasn't correct: the method signature in the result was still
`def _persist_metadata(self, rolename: str, data: bytes)`
when it was supposed have changed to 
`def _persist_metadata(self, rolename: str, data: bytes) -> None`

I took the original commit, rebased (fixing the signature issue) and added a test commit: https://github.com/theupdateframework/python-tuf/pull/1574. Please review if you'd like to double check that it looks good.

Closing this PR, let's continue in #1574

PR #1553, opened 2021-08-31
tests: Add some missing coverage

jku:
Add full test coverage for `Root.remove_key()`  and `DelegatedRole.is_delegated_path()`

After this metadata.py coverage only has a few missing validation branches (that should probably be handled by test_metadata_serialization.py)

The review comment I would give myself on this is: 

> Could we avoid the use of  `for testcase in list_of_test_cases` that is used many times in test_api.py? It makes deciphering failing test results hard since you only get `AssertionError: True is not false` without any context (no idea which case failed). 

... we could. 
* Serialization tests use the sub-test-decorator approach that would work here as well. It does add a bit of complexity -- but on the other hand allows separating test data and test code quite nicely and shows test results with very good context.
* alternatively we could manually add better messages in our `self.assert_*()` calls

Still, I've opted to just do what the other tests do for now.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -431,16 +431,14 @@ class TestMetadata(unittest.TestCase):
         self.assertFalse('private' in key.keyval.keys())
 
 
-    def test_metadata_root(self):
+    def test_root_add_key_and_remove_key(self):
         root_path = os.path.join(
                 self.repo_dir, 'metadata', 'root.json')
         root = Metadata[Root].from_file(root_path)
 
-        # Add a second key to root role
+        # Create a new key
         root_key2 =  import_ed25519_publickey_from_file(
                     os.path.join(self.keystore_dir, 'root_key2.pub'))
-
-
         keyid = root_key2['keyid']
         key_metadata = Key(keyid, root_key2['keytype'], root_key2['scheme'],
             root_key2['keyval'])
@@ -465,11 +463,17 @@ class TestMetadata(unittest.TestCase):
         root.signed.add_key('root', key_metadata)
         self.assertEqual(pre_add_keyid, root.signed.roles['root'].keyids)
 
-        # Remove the key
-        root.signed.remove_key('root', keyid)
+        # Add the same key to targets role as well
+        root.signed.add_key('targets', key_metadata)
 
-        # Assert that root does not contain the new key anymore
+        # Remove the key from root role (targets role still uses it)
+        root.signed.remove_key('root', keyid)
         self.assertNotIn(keyid, root.signed.roles['root'].keyids)
+        self.assertIn(keyid, root.signed.keys)
+
+        # Remove the key from targets as well
+        root.signed.remove_key('targets', keyid)
+        self.assertNotIn(keyid, root.signed.roles['targets'].keyids)
         self.assertNotIn(keyid, root.signed.keys)
 
         with self.assertRaises(KeyError):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -623,6 +623,28 @@ class TestMetadata(unittest.TestCase):
             self.assertRaises(exceptions.LengthOrHashMismatchError,
                 file1_targetfile.verify_length_and_hashes, file1)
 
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -465,11 +463,17 @@ def test_metadata_root(self):
...
-        # Remove the key
-        root.signed.remove_key('root', keyid)
+        # Add the same key to targets as well
+        root.signed.add_key('targets', key_metadata)
 
-        # Assert that root does not contain the new key anymore
+        # Remove the key from root
+        root.signed.remove_key('root', keyid)
         self.assertNotIn(keyid, root.signed.roles['root'].keyids)
+        self.assertIn(keyid, root.signed.keys)
```
MVrachev, line 472: Maybe add one clarifying comment before the `self.assertIn(...)` like:
`Make sure the key is still in keys dict as it's still used by targets`
Just to make it 100 % clear what you are checking here.

tests/test_api.py
```diff
@@ -465,11 +463,17 @@ def test_metadata_root(self):
...
+
+        # Remove the key from targets as well
+        root.signed.remove_key('targets', keyid)
+        self.assertNotIn(keyid, root.signed.roles['targets'].keyids)
```
MVrachev, line 476: And optionally one comment before `self.assertNotIn(...)` here as well:
`key is not used by any role and should have been deleted from keys`

MVrachev (changes requested):
Overall good suggestions.
Both of the commits add tests that can help you understand how our API should operate
which is important both from a testing standpoint, but also for educational purposes.
Your self-review raised a good topic for discussion:

> Could we avoid the use of for testcase in list_of_test_cases that is used many times in test_api.py? It makes deciphering failing test results hard since you only get AssertionError: True is not false without any context (no idea which case failed).

I would say for our serialization tests where our use cases used a more complicated input, then it made perfect sense to use a decorator.
I am wondering if it would simplify the code if you use decorator here.
You will have to split these tests into two test functions with two datasets which I think could be worth it if we find applicable cases for the decorator in other tests in `test_api.py`.
If we decide that we won't use a decorator then yes, I am 100 % for improving the messages in our `self.assert_*()` calls.

tests/test_api.py
avelichka: @MVrachev, why do we need that clarification? What might confuse folks in that assert?

MVrachev: Because he doesn't state what is the testing in this use case.
In his commit message, he explains that, but without that context is not immediately clear.
You are removing a key and still expect it to be in `root.signed.keys` why?

avelichka: IMHO the code is quite self-explanatory  - the key is added for two roles ('root' and 'targets') and is then removed from 'root'. The two asserts check the expected behavior which is to have the key removed from 'root' but not from the key store since 'targets' is still using it.

jku:
> If we decide that we won't use a decorator then yes, I am 100 % for improving the messages in our `self.assert_*()` calls.

I should have mentioned that the reason I did not do it is that comments and messages become outdated over time. I would like the test code (and other code) to not need explanations because the explanations do over time turn into misinformation.

jku:
last change: Tweaked the comment for removing one key usage but not the other

tests/test_api.py
jku: this one I did not do yet, I think the comment is pretty clear: key has now been removed from both roles. If the purpose of remove_key() is not clear then let's document that in the API reference not in the tests.

avelichka (approved):

tests/test_api.py
MVrachev: Agreed that this one is better and it's okay without a change. That's why I said `optionally` probably should have said `consider` as our favorite word :D

MVrachev (approved):
Looks good to me now!

sechkova (approved):
I agree with doing a small amount of effort now to improve coverage. 
`test_api` in general can be further improved later on.

PR #1564, opened 2021-09-03
Add ngclient to coverage report

### LINKED ISSUE #1499: ngclient Updater: write tests to handling coverage report
This is the minimal amount of testing we should have to enable coverage for the ngclient updater and the first step of resolving #1462 for the Updater.
When resolving the missing lines reported by `coverage` we will decide if we want to handle all of them
or there is a sense of leaving some of them.

### LINKED ISSUE #1552: ngclient coverage: SlowRetrievalError testing

Add test for fetcher/requests_fetcher covering the case of urllib3.ReadTimeoutError/tuf.SlowRetrievalError.
This can be achieved by using `slow_retrieval_server.py` instead of `simple_servier.py` for serving the test file.

One gotcha is that by default `RequestFetcher.socket_timeout` is set to 4s and the same value is used both for socket connect and read timeout.  It means that the value would have to be modified to speed up the test case.
For example:

```python
    # Read timeout
    def test_read_timeout(self):
        self.fetcher.socket_timeout = (4, 0.1) # connect timeout: 4s, read timeout: 0.1s
        # Launch a new "slow retrieval" server sending one byte each 40s
        slow_server_process_handler = utils.TestServerProcess(log=logger, server='slow_retrieval_server.py')
        ...
```

sechkova:
Fixes #1499, #1552 

Add ngclient tests for: requests_fetcher.py, trusted_metadata_set.py and updater.py.

Add ngclient to the total coverage report after we've reached 97%.

requirements**:

### DIFF
--- a/tests/test_trusted_metadata_set.py
+++ b/tests/test_trusted_metadata_set.py
@@ -161,21 +161,24 @@ class TestTrustedMetadataSet(unittest.TestCase):
             self.metadata["role1"], "role1", "targets"
         )
 
+    def test_root_with_invalid_json(self):
+        # Test loading initial root and root update
+        for test_func in [TrustedMetadataSet, self.trusted_set.update_root]:
+            # root is not json
+            with self.assertRaises(exceptions.RepositoryError):
+                test_func(b"")
 
-    def test_update_with_invalid_json(self):
-        # root.json not a json file at all
-        with self.assertRaises(exceptions.RepositoryError):
-            TrustedMetadataSet(b"")
-        # root.json is invalid
-        root = Metadata.from_bytes(self.metadata["root"])
-        root.signed.version += 1
-        with self.assertRaises(exceptions.RepositoryError):
-            TrustedMetadataSet(root.to_bytes())
+            # root is invalid
+            root = Metadata.from_bytes(self.metadata["root"])
+            root.signed.version += 1
+            with self.assertRaises(exceptions.UnsignedMetadataError):
+                test_func(root.to_bytes())
 
-        # update_root called with the wrong metadata type
-        with self.assertRaises(exceptions.RepositoryError):
-            self.trusted_set.update_root(self.metadata["snapshot"])
+            # metadata is of wrong type
+            with self.assertRaises(exceptions.RepositoryError):
+                test_func(self.metadata["snapshot"])
 
+    def test_top_level_md_with_invalid_json(self):
         top_level_md = [
             (self.metadata["timestamp"], self.trusted_set.update_timestamp),
             (self.metadata["snapshot"], self.trusted_set.update_snapshot),
@@ -186,9 +189,10 @@ class TestTrustedMetadataSet(unittest.TestCase):
             # metadata is not json
             with self.assertRaises(exceptions.RepositoryError):
                 update_func(b"")
+
             # metadata is invalid
             md.signed.version += 1
-            with self.assertRaises(exceptions.RepositoryError):
+            with self.assertRaises(exceptions.UnsignedMetadataError):
                 update_func(md.to_bytes())
 
             # metadata is of wrong type
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -16,7 +16,7 @@ import tuf.unittest_toolbox as unittest_toolbox
 
 from tests import utils
 from tuf.api.metadata import Metadata
-from tuf import ngclient
+from tuf import exceptions, ngclient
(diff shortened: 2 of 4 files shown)


tests/test_trusted_metadata_set.py
```diff
@@ -161,21 +161,24 @@ def test_out_of_order_ops(self):
...
-        # root.json is invalid
-        root = Metadata.from_bytes(self.metadata["root"])
-        root.signed.version += 1
-        with self.assertRaises(exceptions.RepositoryError):
-            TrustedMetadataSet(root.to_bytes())
+            # root is invalid
+            root = Metadata.from_bytes(self.metadata["root"])
+            root.signed.version += 1
+            with self.assertRaises(exceptions.RepositoryError):
+                test_func(root.to_bytes())
```
MVrachev: IMO it's not immediately clear why root is valid here, because the exception itself is too general.
Maybe change the comment on line 171 to something like: 
`root cannot be verified, because signature has changed`

tests/test_updater_ng.py
```diff
@@ -24,255 +24,315 @@
...
+
+        # Needed because in some tests simple_server.py cannot be found.
+        # The reason is that the current working directory
+        # has been changed when executing a subprocess.
+        cls.SIMPLE_SERVER_PATH = os.path.join(os.getcwd(), "simple_server.py")
+
+        # Launch a SimpleHTTPServer (serves files in the current directory).
+        # Test cases will request metadata and target files that have been
+        # pre-generated in 'tuf/tests/repository_data', which will be served
+        # by the SimpleHTTPServer launched here.  The test cases of 'test_updater.py'
```
MVrachev:
```suggestion
        # by the SimpleHTTPServer launched here. The test cases of 'test_updater._ng.py'
```

tests/test_updater_ng.py
```diff
@@ -228,6 +233,46 @@ def test_refresh_with_only_local_root(self):
...
+        targetinfo.length = 44
+        self.repository_updater.download_target(targetinfo, self.destination_directory)
+
+      with self.assertRaises(exceptions.RepositoryError):
+        targetinfo.length = length
+        targetinfo.hashes = {"sha256": "abcd"}
+        self.repository_updater.download_target(targetinfo, self.destination_directory)
+
+  def test_updating_root(self):
+      self._modify_repository_root(lambda root: None, bump_version=True)
```
MVrachev: The modification function `lambda root: None` won't do any changes on root right?
Wanted to make sure we only wanted to test here bumping the version and resigning.

MVrachev (changes requested):
I think there are good additions in this pr and the new tests are understandable.
I have only a couple of small suggestions.

I have only one question/suggestion: I see you applied black on `test_updater_ng.py`.
Have you also enabled linting checks on `test_updater_ng.py` when calling `tox -e lint`?
Also, there are other test modules where we haven't applied black and lining - `test_apy.py`, `test_metadata_serialization.py` and `test_trusted_metadata_set.py` (I am not sure for that one).

tests/test_trusted_metadata_set.py
jku: updating the exception to be more specific is an option too

sechkova:
> I have only one question/suggestion: I see you applied black on `test_updater_ng.py`.
> Have you also enabled linting checks on `test_updater_ng.py` when calling `tox -e lint`?
> Also, there are other test modules where we haven't applied black and lining - `test_apy.py`, `test_metadata_serialization.py` and `test_trusted_metadata_set.py` (I am not sure for that one).

I applied black mostly because the file was still using the old code style with 2 spaces but it also needed some beautifying. The rest of the test files are much better formatted. I think the common agreement is that we won't apply the code checkers that strictly on the tests (yet), while running black while editing them probably won't hurt if it seems needed.

tests/test_trusted_metadata_set.py
sechkova: Updated the exception to UnsignedMetadataError: https://github.com/theupdateframework/python-tuf/pull/1564/commits/61e8f40346e0fbde3a4f4e80ed8410342f0a06f6

tests/test_updater_ng.py
sechkova: Yes, I needed a modification function doing nothing.

sechkova:
I modified some comments as suggested in the reviews, you can have another look.

jku (approved):
LGTM -- I am trying rewrite the whole testing scheme in #1569, but this PR gets us to some level of code coverage right now and mine does not.

PR #1565, opened 2021-09-06
ngclient: Fix intermediate metadata loading with rollback checks

### LINKED ISSUE #1563: ngclient: recent rollback checks are buggy
I recently added the rollback protection checks in ngclient and am now improving ngclient testing situation. With hindsight the order is wrong: I've found an issue with the rollback checks
* TrustedMetadataSet now considers intermediate timestamp and snapshot valid even if it is expired or version does not match meta version: it only checks final validity when the next metadata type is loaded
* Updater currently decides if it needs to download a new metadata version based whether the local metadata was loaded correctly by TrustedMetadataSet

as a result updater thinks it does not need to download a new version of e.g. snapshot  but then targets update fails because the final snapshot is not the correct version...

This should not be too hard to fix, but I'm trying to think if we can somehow combine this with #1523: it feels like a similar issue

jku:
The rollback checks themselves work, but they create a situation
where Updater does not realize that it needs to download e.g. a new
snapshot because the local snapshot is valid as _intermediate_ snapshot
(that can be used for rollback protection but nothing else), but is not
valid as final snapshot.

The PR changes very little but documentation is improved in several places:
* defines `_check_final_timestamp()` for consistency with `_check_final_snapshot()`
* `update_timestamp()` and `update_snapshot()` will still load "intermediate" metadata as before, but will now raise in the end of function if the metadata is not valid final metadata
* this is what Updater is expecting so there are no changes in Updater

The reason this bug was introduced (by me) was lacking tests. The tests are still missing: I found the issue while working on the tests in #1444. It might make sense to merge this commit when the tests are merged but I wanted to have separate review for this as it's clearly an implementation bug.

### DIFF
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -178,16 +178,20 @@ class TrustedMetadataSet(abc.Mapping):
     def update_timestamp(self, data: bytes) -> None:
         """Verifies and loads 'data' as new timestamp metadata.
 
-        Note that an expired intermediate timestamp is considered valid so it
-        can be used for rollback checks on newer, final timestamp. Expiry is
-        only checked for the final timestamp in update_snapshot().
+        Note that an intermediate timestamp is allowed to be expired:
+        TrustedMetadataSet will throw an ExpiredMetadataError in this case
+        but the intermediate timestamp will be loaded. This way a newer
+        timestamp can still be loaded (and the intermediate timestamp will
+        be used for rollback protection). Expired timestamp will prevent
+        loading snapshot metadata.
 
         Args:
             data: unverified new timestamp metadata as bytes
 
         Raises:
-            RepositoryError: Metadata failed to load or verify. The actual
-                error type and content will contain more details.
+            RepositoryError: Metadata failed to load or verify as final
+                timestamp. The actual error type and content will contain
+                more details.
         """
         if self.snapshot is not None:
             raise RuntimeError("Cannot update timestamp after snapshot")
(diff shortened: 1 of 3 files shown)


jku:
The other way of solving this would have been to return something from the `update_*()` methods to indicate whether metadata is valid final metadata or not. I chose to not do that because this way keeps Updater implementation simple:  a single try-except around the local metadata loading is enough to decide if loading remote metadata is needed

joshuagl (approved):
Tidy little change, LGTM.

sechkova (approved):
I approve the set of code changes and I agree that this approach is better than using the update_* methods return values. 

I think we should in addition make some effort to explain the new terminology that we are adopting: intermediate timestamp, snapshot etc. are not explicitly mentioned in the specification and are not exactly the same as intermediate root versions. The same metadata is referred to as "trusted" in 5.3.11 of the spec,  the current client uses the notion of previous/current while seen from Updater this is the loaded local metadata.
Not to be misunderstood, I'm ok with intermediate but we need a definition somewhere.

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -13,11 +13,18 @@
...
-The rules for top-level metadata are
- * Metadata is updatable only if metadata it depends on is loaded
- * Metadata is not updatable if any metadata depending on it has been loaded
- * Metadata must be updated in order:
-   root -> timestamp -> snapshot -> targets -> (delegated targets)
+The rules that TrustedMetadataSet follows for top-level metadata are
+ * Metadata must be loaded in order:
+   root -> timestamp -> snapshot -> targets -> (delegated targets).
+ * Metadata can be loaded even if it is expired (or in the snapshot case if the
+   meta info does not match): this is called "intermediate metadata".
```
joshuagl, line 20: Nit: it might be worth mentioning rollback protection explicitly here?

joshuagl (approved):

jku:
Closing since this merged in #1569

PR #1569, opened 2021-09-07
test ngclient with simulated repo

### LINKED ISSUE #1444: Improve testing using generated metadata

jku:
I'm not sure if this is quite ready but I do think I need other people to comment on this.

* Add a RepositorySimulator as a test utility
* Add some tests that demonstrate how to use RepositorySimulator
* Add "--dump" option to the tests to show how to debug (or just how to create repository test data if that's needed)
* Include the fix from #1565: let's review that one there even if we merge all together

So I've not tried to implement any real test coverage yet: I'm making a PR now since there's a lot of activity around client tests and maybe this will help others if it turns out reasonable. At the very least the folks creating tests should have opinions :) I do think that if this looks like a good solution we should move most updater tests to this testing style and have extensive coverage here -- and only leave existing tests as smoke test for the whole stack with a network connection etc. 

The note worthy aspects of this are
* `test_updater_with_simulator.py`  is tiny for what it does: 125 lines includes one test with 5 different repository versions and another with 4 different repository versions. I think our current tests _never_ have that many versions anywhere
* it should also be quite fast (although I've not benchmarked this) as network connections are not created and the only file access is by ngclient itself
* `python3 test_updater_with_simulator.py -vvv` will be fairly informative: you can see all the metadata versions that get loaded
* `python3 test_updater_with_simulator.py --dump` is quite good for debugging: it stores a repository version every time the client refreshes. These versions can be easily diffed: `diff /tmp/tmp614pepb4/test_keys_and_signatures/2 /tmp/tmp614pepb4/test_keys_and_signatures/3`
* Writing the tests is not as simple as I'd hope (with e.g. `publish_root()` complication) but maybe still understandable
* I'm still not quite sure _what exactly_ we should check when a refresh succeeds... but this is not unique to my approach here

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -15,16 +15,18 @@ network connections or even file access happens as RepositorySimulator serves
 everything from memory.
 """
 
-import logging
 from collections import OrderedDict
 from datetime import datetime, timedelta
-from tuf.api.serialization.json import JSONSerializer
+import logging
+import os
+import tempfile
 from securesystemslib.keys import generate_ed25519_key
 from securesystemslib.signer import SSlibSigner
-from tuf.exceptions import FetcherHTTPError
 from typing import Dict, Iterator, List, Optional, Tuple
 from urllib import parse
 
+from tuf.api.serialization.json import JSONSerializer
+from tuf.exceptions import FetcherHTTPError
 from tuf.api.metadata import(
     Key,
     Metadata,
@@ -56,6 +58,9 @@ class RepositorySimulator(FetcherInterface):
 
         # signers are used on-demand at fetch time to sign metadata
         self.signers: Dict[str, List[SSlibSigner]] = {}
+
+        self.dump_dir = None
+        self.dump_version = 0
 
         self._initialize()
 
@@ -179,3 +184,29 @@ class RepositorySimulator(FetcherInterface):
 
         self.snapshot.version += 1
         self.update_timestamp()
+
+    def write(self):
+        """Dump current repository metadata to self.dump_dir
+
+        This is a debugging tool: dumping repository state before running
+        Updater refresh may be useful while debugging a test.
+        """
+        if self.dump_dir is None:
+            self.dump_dir = tempfile.mkdtemp()
+            print(f"Repository Simulator dumps in {self.dump_dir}")
+
+        self.dump_version += 1
+        dir = os.path.join(self.dump_dir, str(self.dump_version))
+        os.makedirs(dir)
+
+        for ver in range(1, len(self.signed_roots) + 1):
+            with open(os.path.join(dir, f"{ver}.root.json"), "wb") as f:
+                f.write(self._fetch_metadata("root", ver))
+
+        for role in ["timestamp", "snapshot", "targets"]:
+            with open(os.path.join(dir, f"{role}.json"), "wb") as f:
+                f.write(self._fetch_metadata(role))
+
+        for role in self.md_delegates.keys():
+            with open(os.path.join(dir, f"{role}.json"), "wb") as f:
+                f.write(self._fetch_metadata(role))
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -18,6 +18,7 @@ everything from memory.
 import logging
 from collections import OrderedDict
 from datetime import datetime, timedelta
+from tuf.api.serialization.json import JSONSerializer
 from securesystemslib.keys import generate_ed25519_key
 from securesystemslib.signer import SSlibSigner
 from tuf.exceptions import FetcherHTTPError
@@ -43,21 +44,24 @@ SPEC_VER = ".".join(SPECIFICATION_VERSION)
 
 class RepositorySimulator(FetcherInterface):
     def __init__(self):
-        # all root versions are stored
-        self.md_roots: Dict[int, Metadata[Root]] = {}
+        self.md_root: Metadata[Root] = None
         self.md_timestamp: Metadata[Timestamp] = None
         self.md_snapshot: Metadata[Snapshot] = None
         self.md_targets: Metadata[Targets] = None
-        # all targets in one dict
         self.md_delegates: Dict[str, Metadata[Targets]] = {}
 
+        # other metadata is signed on-demand (when fetched) but roots must be
+        # explicitly published with publish_root() which maintains this list
+        self.signed_roots: List[bytes] = []
+
+        # signers are used on-demand at fetch time to sign metadata
         self.signers: Dict[str, List[SSlibSigner]] = {}
 
         self._initialize()
 
     @property
     def root(self) -> Root:
-        raise NotImplementedError
+        return self.md_root.signed
 
     @property
     def timestamp(self) -> Timestamp:
@@ -75,14 +79,9 @@ class RepositorySimulator(FetcherInterface):
         for role, md in self.md_delegates.items():
             yield role, md.signed
 
-    def _create_key(self, role:str) -> Key:
+    def create_key(self) -> Tuple[Key, SSlibSigner]:
         sslib_key = generate_ed25519_key()
-        if role not in self.signers:
-            self.signers[role] = []
-        self.signers[role].append(SSlibSigner(sslib_key))
-
-        key = Key.from_securesystemslib_key(sslib_key)
-        return key
+        return Key.from_securesystemslib_key(sslib_key), SSlibSigner(sslib_key)
 
     def _initialize(self):
         """Setup a minimal valid repository"""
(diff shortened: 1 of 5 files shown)


sechkova (approved):
I support the idea, I'd like to give it a try.
It is hard to review the existing features of the `RepositorySimulator` without clear requirements but a couple of things grabbed my attention:
- Since we have already `publish_root()`, I think an explicit invocation of a public signing method of the simulator would be better than signing on fetch. This gives you the flexibility to choose creating valid and correctly signed metadata or invalid one, depending on the needs of the test case.
- Deriving from FetcherInterface looks simple and works nice so far. Another option would be a separate "SimulatorFetcher" implemenatation that is initialized with a `RepositorySimulator` instance. This will mimic better the real world usage where Updater would use the fetcher and not the repo simulator directly. But I'm ok with the current design too until we know better the test cases.
- Dumping is cool idea, it feels like it can be further developed to become a dumping utility class for more flexible choice of when and what to dump.

sechkova:
One side question, any ideas how this "client-testing" repository simulator can evolve into repository code and/or a repository testing suit?

jku:
> * Since we have already `publish_root()`, I think an explicit invocation of a public signing method of the simulator would be better than signing on fetch. This gives you the flexibility to choose creating valid and correctly signed metadata or invalid one, depending on the needs of the test case.

You do have a point... I've tried it but finally ended with the implicit sign because:
* incorrect signatures (or root updates) are likely a small minority of tests
* large majority of tests would have to add many  sign() calls if signing is explicit -- and I really like the fact that many repo changes now are literally one or two line lines
* incorrect signatures can still be done: I included test_keys_and_signatures() to show that -- it's a bit complex because you have to make a mental connection between the signers and the implicit signing, but it is flexible: I think you can do whatever test you need with it

All that said, I'm not married to implicit signing:  if you think the explicit version is definitely clearer, I can change it

> any ideas how this "client-testing" repository simulator can evolve into repository code and/or a repository testing suit?

No clear ones, but I figure between this work, repository-editor-for-tuf and Warehouse usage we would have examples where we can start looking at patterns and see what could be in a Metadata API or separate library.

repository testing is tricky... as a first step it would be nice to have a "repository workflow" in the spec that defines in detail how some common thing should be done (new timestamp, new snapshot, common key rotations, adding new targets, adding new delegation, key rotation with ffwd attack recovery, etc).

jku:
@sechkova approved with some valid comments... I think I'm going to make an executive decision and merge it with this approval: this is a testing utility and it's not supposed to be a complete design or "final api" in any way, so let's get other people to try this out, modify/improve it -- we can even remove it without big issues it if it turns out to not work in reality.

The one change I'll do before merge is rebase on #1565 to get the docstring fixes for it.

PR #1574, opened 2021-09-08
Ngclient: persist metadata safely

### LINKED ISSUE #1526: ngclient: persist metadata in a safer way

jku:
Fixes #1526

Persist new metadata in an atomic manner. This contains:
* @Suvadityamuk's commit from #1549, I've modified the commit by fixing the method signature as required to successfully rebase the change on current develop
* my commit that adds some tests

Copying a more thorough explanation of why this is done as it is from https://github.com/theupdateframework/python-tuf/issues/1526#issuecomment-903674931:
* use NamedTemporaryFile to generate a safe temp file name (to avoid overwriting an existing file)
* use the dir argument to make sure the temp file is in same filesystem as the final file (otherwise replace can't be atomic)
* use delete=False argument so NamedTemporaryFile doesn't try to delete the temp file (because we want to move it and can't move if it's been deleted)

This change has one side-effect: if the write fails mid-way or the replace fails, the temp file is not cleaned: #1575 .

### DIFF
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -11,6 +11,7 @@ import shutil
 import tempfile
 import logging
 import sys
+from typing import List
 import unittest
 import tuf.unittest_toolbox as unittest_toolbox
 
@@ -172,6 +173,12 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
             )
         )
 
+    def _assert_files(self, roles: List[str]):
+        """Assert that local metadata files exist for 'roles'"""
+        expected_files = [f"{role}.json" for role in roles]
+        client_files = sorted(os.listdir(self.client_directory))
+        self.assertEqual(client_files, expected_files)
+
     def test_refresh_on_consistent_targets(self):
         # Generate a new root version where consistent_snapshot is set to true
         def consistent_snapshot_modifier(root):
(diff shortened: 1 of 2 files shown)


suvadityamuk:
Looks good to me, thank you for handling the rebasing @jku !

suvadityamuk (approved):

trishankatdatadog:
Thanks, @jku!

@suvadityamuk, please do learn how to rebase correctly going fwd 🙂

tests/test_updater_ng.py
```diff
@@ -172,6 +173,12 @@ def _modify_repository_root(
             )
         )
 
+    def _assert_files(self, roles: List[str]):
+        """Assert that local metadata files exist for 'roles'"""
+        expected_files = [f"{role}.json" for role in roles]
+        client_files = sorted(os.listdir(self.client_directory))
+        self.assertEqual(client_files, expected_files)
```
trishankatdatadog, line 180: Would this work even if `expected_files` is not sorted?

suvadityamuk:
> Thanks, @jku!
> 
> @suvadityamuk, please do learn how to rebase correctly going fwd slightly_smiling_face

Will do!

tests/test_updater_ng.py
joshuagl: Quite right. Fortunately every caller has expected roles in sorted order. Throwing a `sorted()` might save a future test editor some woes.

joshuagl (approved):
Thanks for the contribution @suvadityamuk and the patient shepherding@jku !

PR #1583, opened 2021-09-17
GitHub CI workflow: tweak names

jku:
Currently the github UI dropdown for checks looks useless since every check is named "Run TUF tests and...". Tweak the workflow and job names to hopefully fit the actual step name in the UI.

Also start using python 3.x for lint since pylint now supports it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Run TUF tests and linter
+name: CI
 
 on:
   push:
@@ -8,7 +8,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  tests:
+    name: Tests
     strategy:
       fail-fast: false
       # Run regular TUF tests on each OS/Python combination, plus special tests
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
👍 thanks for fixing the Python version too

jku:
> Required statuses must pass before merging

Hmm, I think the job id may be used in some branch protection rules... I should maybe try not changing the id and just change the name so we don't have to edit the rules

trishankatdatadog (approved):

jku:
I've tried with with the old job id, that doesn't help: the required statuses aren't marked as passed. I think someone with permissions to do so has to play with the branch protection settings: possibly disable the required status checks, merge this PR, then enable them with the new job id/name after merging

(also the coveralls failure seems to be a botched maintenance break: they are 100% down)

joshuagl:
Merged and required status checks have been updated to all of the "Tests" apart from with-sslib-master

trishankatdatadog:
> ## Pull Request Test Coverage Report for [Build 1246248090](https://coveralls.io/builds/42956759)

Why on Earth are we getting an update for this now?

PR #1584, opened 2021-09-17
Fix pylint warnings

MVrachev:
that is provided, the better.

New pylint warnings appeared (for the first time here: https://github.com/theupdateframework/python-tuf/pull/1446/checks?check_run_id=3630842220) with code:
`W0602: Using global for X but no assignment is done (global-variable-not-assigned)`
where X is a global variable.
In the lines pointed by the warnings, "global" was used inside a
function scope in order to enable assignment of the global variable to
a new value in the function.
The reason for these warnings was that pylint noticed we are using
"global" for the variable X inside a function scope without actually
assigning it to a new value.
Overall that was true for the reported cases and I removed the use of
"global", but there were only two cases where pylint reported a
false-positive. Then we were using "global", but pylint didn't
understand there were assignments.
In those cases, I disabled the warnings.

requirements**:

### DIFF
--- a/tuf/keydb.py
+++ b/tuf/keydb.py
@@ -429,10 +429,8 @@ def clear_keydb(repository_name='default', clear_all=False):
   sslib_formats.NAME_SCHEMA.check_match(repository_name)
   sslib_formats.BOOLEAN_SCHEMA.check_match(clear_all)
 
-  global _keydb_dict
-
   if clear_all:
-    _keydb_dict = {}
+    _keydb_dict.clear()
     _keydb_dict['default'] = {}
 
   if repository_name not in _keydb_dict:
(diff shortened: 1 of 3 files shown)


jku (changes requested):
I looked at  a few cases and pylint is correct: those globals are not needed.

`global` is only required if you want to _assign_ to the global variable. mutating the value of that variable does not require global.

jku:
> https://www.programiz.com/python-programming/global-keyword

these third party documentation sites have really variable quality... The whole point of `global` is that _assignment_ to global variables can only work with global, yet this is not mentioned at all. 

I try to find references for this sort of things on python.org (the search engines really fail hard with python searches, always offering this sort of pages before python.org content)

MVrachev:
> I looked at a few cases and pylint is correct: those globals are not needed.
> 
> `global` is only required if you want to _assign_ to the global variable. mutating the value of that variable does not require global.

Updated the pr and removed all but two occurrences of `global` which are false-positives.

> these third party documentation sites have really variable quality... The whole point of global is that assignment to global variables can only work with global, yet this is not mentioned at all.

Yes, you are right. Removed the link from the commit message and pr description.

>I try to find references for this sort of things on python.org (the search engines really fail hard with python searches, always offering this sort of pages before python.org content)

I can't find where on python org globals are mentioned and that's why I won't give a link.

jku (approved):
> -55 lines, fixes CI

Nice

PR #1585, opened 2021-09-20
CI: Do not require coveralls-fin to succeed

jku:
We already do not require individual build uploads to succeed: let's
also not require the final step to succeed.
    
The immediate context for this is that coveralls has been down for
three days now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,7 @@ jobs:
     container: python:3-slim
     steps:
       - name: Finalize publishing on coveralls.io
+        continue-on-error: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |


joshuagl (approved):
> The immediate context for this is that coveralls has been down for
three days now.

Seems reasonable to me. Whether coverage results are published feels like much less of an issue than whether coverage drops below the desired level. As it stands, our required coverage level is high enough and the coverage synthetic enough that I'm not sure awe lose anything by not being able to see historical trends?

Any way, LGTM.

jku:
> Seems reasonable to me. Whether coverage results are published feels like much less of an issue than whether coverage drops below the desired level. As it stands, our required coverage level is high enough and the coverage synthetic enough that I'm not sure awe lose anything by not being able to see historical trends?

We'll still see the results if coveralls works... admittedly now the only thing that tells us if it's not working is the _lack_ of a coverage comment in PRs. Seems to be best we can do though

Issue #1576, opened 2021-09-10
ngclient: consistent targets handled differently than legacy updater


jku:
The definition of consistent targets in the spec is quite bad (as we're working with a URL but the spec insists on talking about filenames) so I'm not really sure what it wants:

Spec on consistent targets in repository:
> consistent target files should be written to non-volatile storage as digest.filename.ext

Spec on consistent targets when client builds the download URL:
> the filename is of the form HASH.FILENAME.EXT

### Issue

* legacy updater has decided this means a target path `a/b` will translate to a download url path `a/{HASH}.b`
* ngclient translates target path `a/b` to a download url path `{HASH}.a/b`

So there's a couple of hings here:
* first we probably should not just change how we do this translation  _<--- this issue is about fixing this_
* second, this translation means our target paths have a undocumented requirement that URLs in general do not: the URL must contain a filename part -- or this translation needs to define what happens with a missing filename part

CC @MVrachev

joshuagl:
> The definition of consistent targets in the spec is quite bad (as we're working with a URL but the spec insists on talking about filenames) so I'm not really sure what it wants:
> 
> Spec on consistent targets in repository:
> 
> > consistent target files should be written to non-volatile storage as digest.filename.ext
> 
> Spec on consistent targets when client builds the download URL:
> 
> > the filename is of the form HASH.FILENAME.EXT

Could you file an issue against the specification?

> ### Issue
> * legacy updater has decided this means a target path `a/b` will translate to a download url path `a/{HASH}.b`
> * ngclient translates target path `a/b` to a download url path `{HASH}.a/b`

Please let's fix that to do translation the same as legacy client.

> * second, this translation means our target paths have a undocumented requirement that URLs in general do not: the URL must contain a filename part -- or this translation needs to define what happens with a missing filename part

What I think you're saying here is that https://example.com/foo is a valid URL and therefore seemingly a valid [`TARGETPATH`](https://theupdateframework.github.io/specification/latest/#targetpath), but that does not include a filename part?

The specification has a lot of file-system assumptions baked in and we could be clearer here, all we specifically ask for in `TARGETPATH` definition is "path to a file". Could you file this as an issue against the specification too?

jku:
> What I think you're saying here is that https://example.com/foo is a valid URL and therefore seemingly a valid [`TARGETPATH`](https://theupdateframework.github.io/specification/latest/#targetpath), but that does not include a filename part?

[...]

PR #1587, opened 2021-09-21
tests: Add target support to RepositorySimulator

jku:
* Add very simple targets support into simulator
* Add documentation for the simulator
* Add an example targets test

This might need to be tweaked and/or extended as we add tests but the
implementation should give a good indication of how to extend it.

As an example, non-consistent targets (or metadata) are not yet supported, but
making fetch() check for the consistent_snapshot state and respond
accordingly should be easy.

---

This should be useful to properly test consistent targets (#1576 ) and then implement good test coverage for updater in general.

CC @MVrachev, marking you a reviewer as I know you might want to use it. If you have any improvement ideas let me know. A potential one is a way for simulator to verify targets client has downloaded (a helper, so tests could say _"assert that my target <file path> is correct"_), but I figured we'd first write some tests and see where we duplicate code...

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -13,20 +13,51 @@ RepositorySimulator implements FetcherInterface so Updaters in tests can use it
 as a way to "download" new metadata from remote: in practice no downloading,
 network connections or even file access happens as RepositorySimulator serves
 everything from memory.
+
+Metadata and targets "hosted" by the simulator are made available in URL paths
+"/metadata/..." and "/targets/..." respectively.
+
+Example::
+
+    # constructor creates repository with top-level metadata
+    sim = RepositorySimulator()
+
+    # metadata can be modified directly: it is immediately available to clients
+    sim.snapshot.version += 1
+
+    # As an exception, new root versions require explicit publishing
+    sim.root.version += 1
+    sim.publish_root()
+
+    # there are helper functions
+    sim.add_target("targets", b"content", "targetpath")
+    sim.targets.version += 1
+    sim.update_snapshot()
+
+    # Use the simulated repository from an Updater:
+    updater = Updater(
+        dir,
+        "https://example.com/metadata/",
+        "https://example.com/targets/",
+        sim
+    )
+    updater.refresh()
 """
 
 from collections import OrderedDict
+from dataclasses import dataclass
 from datetime import datetime, timedelta
 import logging
 import os
 import tempfile
+from securesystemslib.hash import digest
 from securesystemslib.keys import generate_ed25519_key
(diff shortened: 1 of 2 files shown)


MVrachev (approved):
I tried it locally and it looks good to me.
For now, I don't have any concrete suggestions.

tests/test_updater_with_simulator.py
```diff
@@ -71,6 +79,33 @@ def test_refresh(self):
...
+        updater = self._run_refresh()
+        self.assertIsNone(updater.get_one_valid_targetinfo("file"))
+
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", b"content", "file")
+        self.sim.update_snapshot()
+
+        # target now exists, is not in cache yet
+        updater = self._run_refresh()
+        info = updater.get_one_valid_targetinfo("file")
```
MVrachev: `info` sounds a little ambiguous.
Maybe `file_info`?
So when we add more targets they will be `file2`, `file3` and when we get their info with `get_one_valid_targetinfo()` their corresponding variables will be `file2_info`, `file3_info`, etc.

tests/repository_simulator.py
```diff
@@ -136,10 +176,30 @@ def fetch(self, url: str) -> Iterator[bytes]:
...
+            dir_parts, sep , prefixed_filename = path.rpartition("/")
+            prefix, _, filename = prefixed_filename.partition(".")
+            target_path = f"{dir_parts}{sep}{filename}"
+
+            yield self._fetch_target(target_path, prefix)
         else:
             raise FetcherHTTPError(f"Unknown path '{spliturl.path}'", 404)
 
+    def _fetch_target(self, target_path: str, hash: Optional[str]) -> bytes:
+        """Return data for 'target_path', checking 'hash' if it is given"""
```
sechkova: Can you please mention that hash will be requested if consistent snapshot -> consistent targets' naming is used? Took me some time to understand why the hash check is done here. Although when you see metadata and version you can get the idea ... so maybe it's my problem only :)

tests/test_updater_with_simulator.py
jku: yeah... I'm trying to avoid long lines in test code but you may be right: the issue in these lines is `get_one_valid_targetinfo` and `self.assertEquals` not the variable name

tests/repository_simulator.py
jku: added some docstrings and a NotImplementedError for `not consistent_snapshot` to make this clearer

jku:
Modified based on feedback, rebased on develop

sechkova (approved):

MVrachev (approved):

Issue #1543, opened 2021-08-26
Improve ngclient/metadata docs 


sechkova:
**Description of issue or feature request**:

The documentation that will hopefully be hosted on readthedocs after the work on #1517 is done needs some further improvement:

- [ ] Metadata API page is big, could be broken down for easier navigation
- [ ] Class constructors have a lot of arguments which does not look good on the rendered page
- [ ] A section for internal client details (TrustedMetadataSet, RequestsFetcher) could be useful
- [ ] Expand the description of tuf.ngclient.fetcher (see https://github.com/theupdateframework/tuf/pull/1517#discussion_r694700337)

sechkova:
Some proposals from @jku:

>  Metadata API page is big, could be broken down for easier navigation

We could document the main classes (Metadata, all signed derivatives) page-per-class and then have one page for the "helper classes
>   Class constructors have a lot of arguments which does not look good on the rendered page

In general sphinx wants us to document our constructors in the class docstring and not in `__init__` so we can work on that. Also this is one options for making the signature more readable.
```
autodoc_typehints = "description"
autodoc_typehints_description_target = "documented"
```

jku:
https://theupdateframework.readthedocs.io/en/stable/ is nothing yet: needs to be configured somehow I suppose?

EDIT: this also makes sense considering no release includes the config yet... they claim it's automatic:

> We also create a stable version, if your project has any tagged releases. stable will be automatically kept up to date to point at your highest version.

so... let's get a release out

sechkova:
> so... let's get a release out

Yes, exactly, I just tried manually building stable and the version it checked out is:
`HEAD is now at 60875f91 Merge pull request #1284 from joshuagl/joshuagl/release-v0.17.0`

jku:
Quick review of metadata doc:
* As mentioned before, should make it clear that Metadata, Root, Targets, Snapshot and Timestamp are important, other stuff is helpers: An alternative to splitting to many pages  may be to write a TOC in the beginning that makes those 5 stand out?
* there's a lot of repetition of "tuf.api.metadata." in the signatures that is not required. `add_module_names = False` and `python_use_unqualified_type_names = True` fix most of the cases
* to make the docs less complex we should try to hide classes that are implementation details: both Signed and Basefile should not be in public docs, we should try to use `:inherited-members:` in the relevant classes to make the API docs for the derived classes complete instead (basefile might not be need at all since it has no public API)
[...]

PR #1590, opened 2021-09-23
API doc improvements

jku:
Many separate changes to published docs and the annotations: I think this makes sense as one big change (as they often depend on each other in the sense of "we could improve this but then we'd have to fix that new issue that pops up") but I may be able to remove specific changes from the bunch if they're not wanted.

List of changes:
* constructor args are now documented (instead of attributes)
* this is better because sphinx can show annotated types of constructor args and the layout is nicer
* The main classes now have their own pages: I think this is easier to read, at least if we add some higher level architecture doc that explains their relations (in another PR)
* serialization abstraction is now actually documented, but to_dict()/from_dict() are no longer included in docs: they are still part of the API, just not shown to Metadata API users
 * Some type info has been removed from docstrings: this is already visible in annotations and the published documentation
 * Metadata module doc rewritten to fit better into the context it's published in
 * Removed mention of custom serializers: this is not a functioning feature (but the serialization API is now documented)
 * No longer annotating `__init__()` return value: this seems to not be required now and looks better in docs

Please review result in https://theupdateframework--1590.org.readthedocs.build/en/1590/api/tuf.api.html


jku:
Link to RTD: https://theupdateframework--1590.org.readthedocs.build/en/1590/api/tuf.api.html

I've not tried to update the text on that page yet: if you take a look, please look at the child pages content

jku:
Future work: Improve the actual language in docstrings to better serve the _user of the API_: As an example 

> **unrecognized fields**: Dictionary of all unrecognized fields.

is bad: we should explain A) what the argument is for, and B) that Metadata API does not manage the content

... but I'm trying to keep this PR to a reasonable size, let's just collect these problems and file a new issue for them

jku:
This is linked to #1543 but I won't autoclose as the content in the issue needs to reviewed and remaining issues possibly filed as separate ones.

docs/api/tuf.api.metadata.rst
```diff
@@ -4,4 +4,3 @@ Metadata
 .. automodule:: tuf.api.metadata
    :members:
    :undoc-members:
-   :show-inheritance:
```
sechkova, line 7: nit: Remind me, why didn't we want to see the inheritance?

tuf/api/metadata.py
```diff
@@ -104,7 +104,10 @@ class Metadata(Generic[T]):
     "[Root]" is not validated at runtime (as pure annotations are not available
     then).
 
-    Attributes:
+    *All parameters named below are not just constructor arguments but also
+    instance attributes.*
+
+    Args:
```
sechkova, line 100: nit: I'm not the person who is going to insist on blindly following the guidelines but should we explain this "deviation" to potential new contributors somewhere? (The style guideline says public attributes of classes should be documented)

tuf/api/metadata.py
```diff
@@ -104,9 +94,12 @@ class Metadata(Generic[T]):
...
-    Attributes:
-        signed: A subclass of Signed, which has the actual metadata payload,
-            i.e. one of Targets, Snapshot, Timestamp or Root.
+    *All parameters named below are not just constructor arguments but also
+    instance attributes.*
+
+    Args:
+        signed: The actual metadata payload, i.e. one of Targets, Snapshot,
+            Timestamp or Root.
         signatures: An ordered dictionary of keyids to Signature objects, each
```
sechkova, line 103: nit: I couldn't figure out why `OrderedDictionary` is not displayed as a type here but it is for `roles` in `class Delegations`

sechkova (approved):
LGTM, some nits and candidates for follow up issues:

1) Some of the functions with missing parameters description:
- add_key, remove_key, update, is_delegated_path,  from_securesystemslib_key, serialization
 https://theupdateframework--1590.org.readthedocs.build/en/1590/api/tuf.api.metadata.root.html#tuf.api.metadata.Root.add_key

2) There is a pylint disable in UpdaterConfig which is rendered in the docs too:
 https://theupdateframework--1590.org.readthedocs.build/en/1590/api/tuf.ngclient.config.html

tuf/api/metadata.py
jku: huh... I thought autodoc wasn't able to handle the type-as-string `"OrderedDict"` but you are right that the same thing works in Delegations. How weird

docs/api/tuf.api.metadata.rst
jku: For almost all classes it's in my opinion not useful: either they have a `Bases: Object` line that's just noise or they identify a parent object (like Signed or BaseFile) that we don't document publically because it's an implementation detail that library users don't need. The only case where we use derived classes as part of the API is Serialization and there I left it visible.

tuf/api/metadata.py
jku: It's certainly not orthodox... The explanation you quoted is the best I could come up with but I'll definitely take suggestions.

I'm against duplicating the docs so both arguments and attributes are documented as long as the arguments and attributes are the same ... if they end up diverging the situation is different.

tuf/api/metadata.py
jku: the types do appear on my local build -- I think this might be a missing feature in the version of python/sphinx/autodoc that RTD uses: at least Generics support had been improved in recent releases.

jku:
I removed the ngclient pylint disable in a new commit and filed a new issue for missing argument docs: https://github.com/theupdateframework/python-tuf/issues/1599

File another for improving the language of the docstrings https://github.com/theupdateframework/python-tuf/issues/1600

On the args vs attributes documentation: I'm against duplicating the docs so both arguments and attributes are documented as long as the arguments and attributes are the same ... if they end up diverging the situation is different. Documenting arguments but not attributes is clearly a bit weird but it's the best compromise I've been able to come up with: let's change that when/if we think of something better.

Metadata constructor arguments are currently not type annotated in RTD (but they are in my local build): could be related to RTD using a python that's a few versions behind -- maybe the generics annotation data is not as easily available there :shrug:

tuf/api/metadata.py
```diff
@@ -522,21 +521,20 @@ def bump_version(self) -> None:
 class Key:
     """A container class representing the public portion of a Key.
 
-    Please note that "Key" instances are not semanticly validated during
-    initialization: this only happens at signature verification time.
+    Supported key content (type, scheme and keyval) is defined in
+    Securesystemslib.
```
MVrachev, line 515:
```suggestion
    Supported key content (type, scheme and keyval) is defined in
    and validated in Securesystemslib.
```
As you are replacing a sentence about validation it's probably good to add a word for that?

docs/api/tuf.api.metadata.supporting.rst
```diff
@@ -0,0 +1,27 @@
...
+
+.. autoclass:: tuf.api.metadata.Delegations
+
+.. autoclass:: tuf.api.metadata.Key
+
+.. autoclass:: tuf.api.metadata.MetaFile
+
+.. autoclass:: tuf.api.metadata.Role
+
+.. autoclass:: tuf.api.metadata.TargetFile
```
MVrachev, line 27: In my opinion, the order of definition could be better.
Starting with the Root helper classes (as the first of the four metadata classes defined in metadata.py) - `Key` and `Role`, then going with Timestamp and Snapshot helper class - `MetaFile` and finish with Targets helper classes `Delegations` and `DelegatedRole`.

MVrachev (commented):
Even though I just saw the pr is merged, I will leave some small comments.
Besides them, I like the changes.

Issue #1579, opened 2021-09-14
Design test strategy


joshuagl:
**Description of issue or feature request**:

The following list captures some areas of interest for testing and some example test cases for each area. For each _area_ it would be useful to:
1. better flesh out what we are trying to test, what is not in scope and generate a(n initial) list of tests to implement
2. research and propose an appropriate test paradigm for the area which prioritises readability and maintainability of the tests
3. given 1 and 2, design appropriate test suite

Test areas:
- Expected error cases
  - Specification defined error cases (need to extract from the specification)
  - The known attacks on package managers (define test scenarios for each attack)
  - Python-tuf implementation error cases (review new code and identify error sequences which are testable)
- Specific sequences the client should be able to handle – different types of updates sequences that may exist in a repository and which the updater should be able to handle
  - Key rotations
  - Recovery from fast-forward attack
  - Client that has been offline for N expirations of root metadata
  - Air-gapped client outside of timestamp expiration window
  - Some of the previously reported security issues?
  - etc
- Configuration permutations
  - Ensure we test all the relevant sequences of events with and without consistent snapshots
  - Identify permutations of [default config](https://github.com/theupdateframework/python-tuf/blob/develop/tuf/ngclient/config.py) to test (where they are not already adequately tested by the unit tests) 
- Table testing for a set of defined tests with different parameters, i.e. 
  - target paths (input to get_one_valid_targetinfo())
  - root rotations
  - different complexity delegation trees – both the client and repository side of the delegation tree operations
    - idea: describe delegations in Graphviz dot format?

joshuagl:
It probably makes sense to split each test area into a separate issue where the scope, actual tests and proposed test architecture can be defined.

jku:
Thanks for writing this down. Edited the list to add target paths to table testing section

jku:
I'm looking at legacy updater tests currently: I'll try to document the actual tests they contain (at least the ones that still apply) and take notes on what useful approaches they might use.

sechkova:
After some iterations over the potential test areas and topics, we've narrowed down several bigger ones that we want to start the testing with and keep refining and improving further. 
Since tracking of all the tests we want to implement is hard in a GitHub issue, there is a [WIP Client test scenarios](https://docs.google.com/spreadsheets/d/1HQI-FdRN-XqEZa-6WTuEkcx4RQAdPRDsgGADJahzdPo/edit?usp=sharing) spread sheet that we mean to constantly update to mark progress.

[...]

Issue #1571, opened 2021-09-08
ngclient: downloading targets fails if  targetpath includes subdirectories


jku:
ngclient has so far copied the legacy client decision of downloading files to the same directory structure as the repository uses. It has however removed the code that creates the directory structure.

So downloading any targets with paths like `"path/to/file"` will fail because the local directories do not exist.

jku:
I think this is actually a good time to get rid of this idea of using target paths as local path with directories (as it has several issues, not just that we don't create the subdirectories)... I'll take this

jku:
We might want to fix this with #1412
https://github.com/theupdateframework/python-tuf/issues/1412#issuecomment-917966743

jku:
I have a branch but this is in practice blocked by #1587 : The legacy tests do not include any targets with sub directories (!) and I can't be bothered to add them, will wait for #1587.

I also suppose this one makes #1576 trickier to test (as the downloads just won't work).

MVrachev:
> I think this is actually a good time to get rid of this idea of using target paths as local path with directories (as it has several issues, not just that we don't create the subdirectories)... I'll take this

@jku I am sorry I created a commit before reading this issue and noticing you are assigned to it.
I created one commit possibly solving that issue: https://github.com/theupdateframework/python-tuf/pull/1591/commits/6f108c9a1f7822b863b0118d8159e108b850bca8, but the question remains do we want 
to further support targetpaths as local paths.

What other do you mean besides this one if we consider targetpaths as local paths?

jku:
> @jku I am sorry I created a commit before reading this issue and noticing you are assigned to it.
> I created one commit possibly solving that issue: [6f108c9](https://github.com/theupdateframework/python-tuf/commit/6f108c9a1f7822b863b0118d8159e108b850bca8), 

That's fine, I'm sure you needed to do something to test -- and we are annoyingly touching the same bits of code. We'll see how bad the merging/rebasing gets...

> but the question remains do we want to further support targetpaths as local paths.

I don't want to create local directories based on a URL -- a specific TUF implementation might know that is a reasonable thing to do but I don't think the library can know that. I'm suggesting we use the whole target path (url encoded) as a filename, and then also offer a way for the client application to decide the filepath themselves (so if the client knows using target path as path is safe, it can create the directories).

PR #1591, opened 2021-09-24
Fix handling consistent targets same as legacy updater

### LINKED ISSUE #1576: ngclient: consistent targets handled differently than legacy updater

MVrachev:
**Note:** This pr is dependent on pr #1592.

Fixes #1576

The definition of consistent targets in the spec is ambiguous:
"consistent target files should be written to non-volatile storage
as digest.filename.ext"
Additionally, the specification describes consistent targets when the
client builds the download URL as follows:
"The filename is of the form HASH.FILENAME.EXT".
The issue is about how we interpreted those quotes.
The legacy updater has decided this means a target path `a/b` will translate to a download url path `a/{HASH}.b`.
The ngclient however translates the target path `a/b` to a download URL path `{HASH}.a/b`.

We decided we want to follow the same approach taken from the legacy
updater and thus change how we construct the consistent targets.
Additionally, we want to make sure we test for cases when the TARGETPATH
is an empty string or points to a directory.

requirements**:

### DIFF
--- a/tests/test_updater_with_simulator.py
+++ b/tests/test_updater_with_simulator.py
@@ -107,6 +107,9 @@ class TestUpdater(unittest.TestCase):
             [file_info]
         )
 
+        # Assert consistent_snapshot is True and downloaded targets have prefix.
+        self.assertTrue(self.sim.root.consistent_snapshot)
+        self.assertTrue(updater.config.prefix_targets_with_hash)
         # download target, assert it is in cache and content is correct
         local_path = updater.download_target(file_info, self.targets_dir)
         self.assertEqual(
--- a/tests/test_updater_with_simulator.py
+++ b/tests/test_updater_with_simulator.py
@@ -82,7 +82,8 @@ class TestUpdater(unittest.TestCase):
     def test_targets(self):
         targets = {
             "targetpath": b"content",
-            "åäö": b"more content"
+            "åäö": b"more content",
+            "dir/targetpath": b"dir target content"
         }
 
         # Add targets to repository
@@ -110,8 +111,13 @@ class TestUpdater(unittest.TestCase):
             with open(local_path, "rb") as f:
                 self.assertEqual(f.read(), content)
 
-        # TODO: run the same download tests for target paths like "dir/file2")
-        # This currently fails because issue #1576
+            if "/" in targetpath:
+                # assert local_path != targetpath because of the URL encoding
+                # make target_path absolute as local_path
+                target_path = os.path.join(self.targets_dir, targetpath)
+                self.assertNotEqual(target_path, local_path)
+
+
 
     def test_keys_and_signatures(self):
         """Example of the two trickiest test areas: keys and root updates"""
--- a/tests/test_updater_with_simulator.py
+++ b/tests/test_updater_with_simulator.py
@@ -10,7 +10,7 @@ import logging
 import os
 import sys
 import tempfile
-from typing import Optional
+from typing import Optional, Tuple
 from tuf.exceptions import UnsignedMetadataError
 import unittest
 
@@ -18,6 +18,7 @@ from tuf.ngclient import Updater
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
+
 
 class TestUpdater(unittest.TestCase):
     # set dump_dir to trigger repository state dumps
@@ -79,43 +80,45 @@ class TestUpdater(unittest.TestCase):
 
         self._run_refresh()
 
-    def test_targets(self):
-        targets = {
-            "targetpath": b"content",
-            "åäö": b"more content",
-            "dir/targetpath": b"dir target content"
-        }
+    targets: utils.DataSet = {
+        "standard case": ("targetpath", b"content", "targetpath"),
+        "non-asci case": ("åäö", b"more content", "%C3%A5%C3%A4%C3%B6"),
+        "subdirectory case": ("a/b/c/targetpath", b"dir target content", "a%2Fb%2Fc%2Ftargetpath"),
+    }
+
+    @utils.run_sub_tests_with_dataset(targets)
+    def test_targets(self, test_case_data: Tuple[str, bytes, str]):
+        targetpath, content, encoded_path = test_case_data
+        # target does not exist yet
+        updater = self._run_refresh()
+        self.assertIsNone(updater.get_one_valid_targetinfo(targetpath))
 
         # Add targets to repository
         self.sim.targets.version += 1
-        for targetpath, content in targets.items():
-            self.sim.add_target("targets", content, targetpath)
+        self.sim.add_target("targets", content, targetpath)
         self.sim.update_snapshot()
 
         updater = self._run_refresh()
-        for targetpath, content in targets.items():
-            # target now exists, is not in cache yet
-            file_info = updater.get_one_valid_targetinfo(targetpath)
-            self.assertIsNotNone(file_info)
-            self.assertEqual(
-                updater.updated_targets([file_info], self.targets_dir),
-                [file_info]
-            )
+        # target now exists, is not in cache yet
+        file_info = updater.get_one_valid_targetinfo(targetpath)
+        self.assertIsNotNone(file_info)
+        self.assertEqual(
+            updater.updated_targets([file_info], self.targets_dir),
+            [file_info]
+        )
 
-            # download target, assert it is in cache and content is correct
-            local_path = updater.download_target(file_info, self.targets_dir)
-            self.assertEqual(
-                updater.updated_targets([file_info], self.targets_dir), []
-            )
-            self.assertTrue(local_path.startswith(self.targets_dir))
-            with open(local_path, "rb") as f:
-                self.assertEqual(f.read(), content)
+        # download target, assert it is in cache and content is correct
+        local_path = updater.download_target(file_info, self.targets_dir)
+        self.assertEqual(
+            updater.updated_targets([file_info], self.targets_dir), []
+        )
+        self.assertTrue(local_path.startswith(self.targets_dir))
+        with open(local_path, "rb") as f:
+            self.assertEqual(f.read(), content)
 
-            if "/" in targetpath:
-                # assert local_path != targetpath because of the URL encoding
-                # make target_path absolute as local_path
-                target_path = os.path.join(self.targets_dir, targetpath)
-                self.assertNotEqual(target_path, local_path)
+        # Assert that the targetpath was URL encoded as expected.
+        encoded_absolute_path = os.path.join(self.targets_dir, encoded_path)
+        self.assertEqual(local_path, encoded_absolute_path)
 
 
 
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -20,8 +20,10 @@
   Provide common utilities for TUF tests
 """
 
-import argparse
 from contextlib import contextmanager
+from typing import Dict, Any, Callable
+import unittest
+import argparse
 import errno
 import logging
 import socket
@@ -38,6 +40,22 @@ logger = logging.getLogger(__name__)
 
 # Used when forming URLs on the client side
 TEST_HOST_ADDRESS = '127.0.0.1'
+
+
+# DataSet is only here so type hints can be used.
+DataSet = Dict[str, Any]
+
+# Test runner decorator: Runs the test as a set of N SubTests,
+# (where N is number of items in dataset), feeding the actual test
+# function one test case at a time
+def run_sub_tests_with_dataset(dataset: DataSet):
+    def real_decorator(function: Callable[[unittest.TestCase, Any], None]):
+        def wrapper(test_cls: unittest.TestCase):
+            for case, data in dataset.items():
+                with test_cls.subTest(case=case):
+                    function(test_cls, data)
+        return wrapper
+    return real_decorator
 
 
 class TestServerProcessError(Exception):
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -229,6 +229,7 @@ class Updater:
                 download URL. Default is the value provided in Updater()
 
         Raises:
+            ValueError: Invalid arguments
             TODO: download-related errors
             TODO: file write errors
 
@@ -251,8 +252,9 @@ class Updater:
         consistent_snapshot = self._trusted_set.root.signed.consistent_snapshot
         if consistent_snapshot and self.config.prefix_targets_with_hash:
             hashes = list(targetinfo.hashes.values())
-            target_filepath = f"{hashes[0]}.{target_filepath}"
-        full_url = parse.urljoin(target_base_url, target_filepath)
+            dirname, sep, basename = target_filepath.rpartition("/")
+            target_filepath = f"{dirname}{sep}{hashes[0]}.{basename}"
+        full_url = f"{target_base_url}{target_filepath}"
 
         with self._fetcher.download_file(
             full_url, targetinfo.length
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -12,7 +12,7 @@ import logging
 import unittest
 import copy
 
-from typing import Dict, Callable
+from typing import Dict
 
 from tests import utils
 
@@ -31,28 +31,12 @@ from tuf.api.metadata import (
 
 logger = logging.getLogger(__name__)
 
-# DataSet is only here so type hints can be used:
-# It is a dict of name to test dict
-DataSet = Dict[str, str]
-
-# Test runner decorator: Runs the test as a set of N SubTests,
-# (where N is number of items in dataset), feeding the actual test
-# function one test case at a time
-def run_sub_tests_with_dataset(dataset: DataSet):
-    def real_decorator(function: Callable[["TestSerialization", str], None]):
-        def wrapper(test_cls: "TestSerialization"):
-            for case, data in dataset.items():
-                with test_cls.subTest(case=case):
-                    function(test_cls, data)
-        return wrapper
-    return real_decorator
-
 
 class TestSerialization(unittest.TestCase):
 
     # Snapshot instances with meta = {} are valid, but for a full valid
     # repository it's required that meta has at least one element inside it.
-    invalid_signed: DataSet = {
+    invalid_signed: utils.DataSet = {
         "no _type": '{"spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
         "no spec_version": '{"_type": "signed", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
         "no version": '{"_type": "signed", "spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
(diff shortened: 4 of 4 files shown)


MVrachev:
While working on this one multiple questions appeared:
1. If a targetpath contains directories, then do we want to create the missing directories? 
There is a separate issue about this and I left a comment there: https://github.com/theupdateframework/python-tuf/issues/1571#issuecomment-926595131
2. How can we create better error messages from the asserts in the for cycles in the unit tests? 
Once one of the asserts failed and I wasn't sure which case failed.
Do we want to use decorators as we did in `test_metadata_serialization.py`?
3. Are there other cases I should cover for TARGETPATH inside the for loops?
4. Do we want to validate the destination directory in download_target?
I think we do need to do that. Are the added checks sufficient? 
5. Don't we want to filter the `target_path` input for `get_one_valid_targetinfo` as well?
6. Are there other file write errors that we should document in the `Raises` for `download_target`
or we can delete one of the TODO items in the Raises section?

jku (commented):
I think the errors you raise both look like things we could handle earlier: the target path originally comes from the client so if we're validating, let's validate where it really is input -- get_one_valid_targetinfo.

The commit also uses `os.path` to form a URL. It may be possible that Python has enough workarounds to make that work in practice in Windows but I don't like it. `os.path` is for local file paths

jku:
> While working on this one multiple questions appeared:
> 
> 1. If a targetpath contains directories, then do we want to create the missing directories?
>    There is a separate issue about this and I left a comment there: [ngclient: downloading targets fails if  targetpath includes subdirectories #1571 (comment)](https://github.com/theupdateframework/python-tuf/issues/1571#issuecomment-926595131)

I'll try to have a PR for that ASAP now that the blocker is merged. But feel free to include the directory creating commit until then

> 2. How can we create better error messages from the asserts in the for cycles in the unit tests?
>    Once one of the asserts failed and I wasn't sure which case failed.
>    Do we want to use decorators as we did in `test_metadata_serialization.py`?

Yeah, that's one of the core advantages of proper table testing, we probably do.

> 4. Do we want to validate the destination directory in download_target?
>    I think we do need to do that. Are the added checks sufficient?

I think we maybe don't need to validate that it exists or can be written to -- I think we can just mention OSError in the docstring. But should test that something reasonable really happens on those cases

I already mentioned the target path checks (if we do them) should IMO happen much earlier.

> 5. Don't we want to filter the `target_path` input for `get_one_valid_targetinfo` as well?

:+1: new issue might be appropriate

MVrachev:
> > 5. Don't we want to filter the `target_path` input for `get_one_valid_targetinfo` as well?
> 
> 👍 new issue might be appropriate

Created an issue for that: #1594.

MVrachev:
This pr is ready for review after the merge of #1587 and  #1592.

> > 1. How can we create better error messages from the asserts in the for cycles in the unit tests?
> >    Once one of the asserts failed and I wasn't sure which case failed.
> >    Do we want to use decorators as we did in `test_metadata_serialization.py`?

> Yeah, that's one of the core advantages of proper table testing, we probably do.

I experimented with a new commit where I used the same decorator as `test_metadata_serialization.py`.
IMO it looks good and the error messages are a lot better.

> > 1. Do we want to validate the destination directory in download_target?
> >    I think we do need to do that. Are the added checks sufficient?
> 
> I think we maybe don't need to validate that it exists or can be written to -- I think we can just mention OSError in the docstring. But should test that something reasonable really happens on those cases
> 
> I already mentioned the target path checks (if we do them) should IMO happen much earlier.

I won't focus on validation in this pr.
This should be resolved in #1594.

> The commit also uses os.path to form a URL. It may be possible that Python has enough workarounds to make that work in practice in Windows but I don't like it. os.path is for local file paths

I changed that and instead used a manual file path and URL creation.

@jku can you answer my last question too: 

6. Are there other file write errors that we should document in the Raises for download_target
or we can delete one of the TODO items in the Raises section?

tests/test_updater_with_simulator.py
```diff
@@ -82,7 +82,8 @@ def test_refresh(self):
     def test_targets(self):
         targets = {
             "targetpath": b"content",
-            "åäö": b"more content"
+            "åäö": b"more content",
+            "dir/targetpath": b"dir target content"
```
sechkova: Can you add a test case with more that one separator:
"a/b/targetpath"

tests/test_updater_with_simulator.py
```diff
@@ -110,8 +111,13 @@ def test_targets(self):
             with open(local_path, "rb") as f:
                 self.assertEqual(f.read(), content)
 
-        # TODO: run the same download tests for target paths like "dir/file2")
-        # This currently fails because issue #1576
+            if "/" in targetpath:
+                # assert local_path != targetpath because of the URL encoding
+                # make target_path absolute as local_path
+                target_path = os.path.join(self.targets_dir, targetpath)
+                self.assertNotEqual(target_path, local_path)
```
sechkova: I don't know if this test case makes sense, it is too specific and may not be correct when the api changes further. Can we at least move it to a separate test?

tests/test_updater_with_simulator.py
```diff
@@ -19,6 +19,23 @@
...
 from tests.repository_simulator import RepositorySimulator
 
+# DataSet is only here so type hints can be used:
+# It is a dict of name to test dict
+DataSet = Dict[str, str]
+
+# Test runner decorator: Runs the test as a set of N SubTests,
+# (where N is number of items in dataset), feeding the actual test
+# function one test case at a time
+def run_sub_tests_with_dataset(dataset: DataSet):
```
sechkova: Instead of copying the decorator code twice, it would be better to move it to a dedicated tests/util/helper/etc file and reuse it.

sechkova (changes requested):
The implementation looks good but I have some reservations for the tests that I put in the comments. Also there isn't really a test for the hash prefix implementation anywhere, the tests rely on the assumption that the default workflow uses consistent targets and no errors were raised.

tests/test_updater_with_simulator.py
MVrachev: I will change my commit description to clarify that the decorator is similar, not the same with the one in `test_metadata_serialization.py`.
First, the `DataSet` type here should be `DataSet = Dict[str, Tuple[str, str]]` instead of `DataSet = Dict[str, str]` which I would fix.
Second, the decorator annotations (including the nested two functions) have different annotations than the ones inside `test_metadata_serialization.py`.

One solution in order to reuse the decorator will be to remove the annotations, but I think annotations are crucial in understanding how the decorator works and I think they should exist and be correct.

tests/test_updater_with_simulator.py
MVrachev: I wanted to showcase that the given path is encoded and as a consequence changed if it includes a directory in it.
I decided to use an if statement in order to add as little as possible changes as the code used for the directory case is the same as the other two cases.

Do you still think I should move it to a separate test?

MVrachev:
> Also there isn't really a test for the hash prefix implementation anywhere, the tests rely on the assumption that the default workflow uses consistent targets and no errors were raised.

I will think about how can I enhance the test and leave a comment when I am ready.

tests/test_updater_with_simulator.py
jku: This is one of the few places where `Any` annotation makes sense: the dict values are specific to the test and the decorator does not care what the actual type is.

jku:
> 6\. Are there other file write errors that we should document in the `Raises` for `download_target`
>     or we can delete one of the TODO items in the Raises section?

Sorry I don't understand the question? I don't see any write errors documented currently.

Can file writes fail? Yes. Does our code raise exceptions in those cases? probably: someone needs to read the code and find out

jku:
> Also there isn't really a test for the hash prefix implementation anywhere, the tests rely on the assumption that the default workflow uses consistent targets and no errors were raised.

I think this might be fine: we definitely want the default setup to always be consistent_snapshot (test could assert that this is the case, or just add a comment that it's expected).

But I have been thinking that RepositorySimulator could easily have some tracking on what gets requested/downloaded... so a test could then assert that e.g. a specific url was used. This sounds like a potential improvement but maybe a sidetrack for this issue?

tests/test_updater_with_simulator.py
sechkova: Yes, I believe the test cases should be as short and easy to understand as possible and this deserves a separate test.

tests/test_updater_with_simulator.py
MVrachev: What about the other annotations? I mean the ones we are currently using:
`function: Callable[["TestUpdater", str], None]` and `test_cls: "TestUpdater"`.
Do you think we should use `Any` here too?
As I mentioned above my only concern is if it will be understandable for a new contributor.

jku: There's nothing in those functions that actuall requires a `TestUpdater`, AFAICT. A `unittest.TestCase` should work just fine?

MVrachev: I guess you are right, those classes inherit `unittest.TestCase`, so this can be generalized and still give information with the annotations.

MVrachev:
Updated the pr with the following changes:
- moved the test case for targetpath with directories in a separate function
- moved the decorator in `tests/utils.py` and used it in both `test_metadata_serialization.py` and `test_updater_with_simulator.py`
- added checks that consistent_snapshot` and `prefix_targets_with_hash` are used before downloading a target in the test and a check inside the repo_simulator.

tests/utils.py
```diff
@@ -20,8 +20,10 @@
   Provide common utilities for TUF tests
 """
 
-import argparse
 from contextlib import contextmanager
+from typing import Dict, Any, Callable
+import unittest
```
sechkova, line 25: If the type is quoted ("unittest.TestCase"), this import should not be needed?

tests/repository_simulator.py
```diff
@@ -184,6 +184,8 @@ def fetch(self, url: str) -> Iterator[bytes]:
             path = spliturl.path[len("/targets/") :]
             dir_parts, sep , prefixed_filename = path.rpartition("/")
             prefix, _, filename = prefixed_filename.partition(".")
+            if prefix == "":
+                raise ValueError("File name must have hash prefix")
```
sechkova: Isn't this error duplicating  `NotImplementedError("non-consistent snapshot not supported")`?

tests/test_updater_with_simulator.py
```diff
@@ -115,13 +114,40 @@ def test_targets(self, test_case_data: Tuple[str, str]):
         with open(local_path, "rb") as f:
             self.assertEqual(f.read(), content)
 
-        if "/" in targetpath:
-            # assert local_path != targetpath because of the URL encoding
-            # make target_path absolute as local_path
-            target_path = os.path.join(self.targets_dir, targetpath)
-            self.assertNotEqual(target_path, local_path)
+    def test_targets_with_subdirectories(self):
```
sechkova: We've misunderstood each other, sorry about that. Duplicating the whole piece of code like this makes no sense. 
I imagined a couple of lines downloading a file and comparing the paths.

But let's forget about the separate test. I suggest we do something else. In the targets DataSet, the Tuple can be expanded with one more item representing the encoded targethpath. In the ascii case this will duplicate the targetpath. For example:

```python
    targets: utils.DataSet = {
        "standard case": ("targetpath", b"content", "targetpath"),
        "non-asci case": ("åäö", b"more content", "%C3%A5%C3%A4%C3%B6"),
        "subdirectory case": ("a/b/c/targetpath", b"dir target content", "a%2Fb%2Fc%2Ftargetpath"),
    }
```
You can maybe use a named-tuple if you think this is not readable enough. 
At the end of `test_targets`, instead of asserting that the two paths are not equal, we can assert that the local path  really is the encoded string from the DataSet:
`self.assertEqual(local_path, os.path.join(self.targets_dir, encoded_path))`

What do you think about this option?

sechkova (changes requested):

tests/repository_simulator.py
jku: I think it's not quite a duplicate, but it's not correct either. Our updaters expect to be able to download non-prefixed targets even if repo is consistent_snapshot: they even have a configuration option to do that... _Maybe_ that was a bad idea considering the unfortunate prefixed-filename-design but it is what Warehouse intends to use.

I don't think a repository can support _completely arbitrary_ target names with this prefix-design (if client asks for target `a.b` we have no way of 100% reliably knowing if `a` is a hash or just part of non-prefixed filename, unless there's some external knowledge of how clients work)... but we just have to implement something in RepositorySimulator so we can test both prefixed and non-prefixed downloads.

Long term maybe there should actually be a `consistent_targets` root variable in the spec to solve this

tests/utils.py
MVrachev: It is quoted and it doesn't work without importing `unittest`. 
I think the reason is that quoting is `forward reference`, but it's still a reference that should be resolved: [PEP 0484 forward reference](https://www.python.org/dev/peps/pep-0484/#forward-references).

MVrachev: It is quoted and it doesn't work without importing `unittest`. 
I think the reason is that quoting is `forward reference`, but it's still a reference that should be resolved: [PEP 0484 forward reference](https://www.python.org/dev/peps/pep-0484/#forward-references).

tests/repository_simulator.py
MVrachev: I don't think so, because if the prefix is "" here this means you have enabled consistent snapshot, but not used hash prefix.
Now when I think for this case this seems to be a possible valid situation - that's why we have the separate `prefix_targets_with_hash` variable, but also the repository simulator doesn't have access to it...
Wondering if I should remove that check considering that the repository simulator doesn't care about the value of `prefix_targets_with_hash` as long as consistent snapshot is True.

tests/test_updater_with_simulator.py
MVrachev: I like this approach. Will try it.

tests/repository_simulator.py
MVrachev: > Long term maybe there should actually be a consistent_targets root variable in the spec to solve this

Yes, I like that suggestion. Multiple times I make the mistake to think that `consistent_targets` will be used only if `consistent_snapshot` is True. 
I think it will be beneficial if we split the idea of consistent targets and consistent snapshots.

The question remains: what should we do now? Maybe just remove that check as we don't have enough context to do it well?

jku: yeah the check is definitely not correct

tests/test_updater_with_simulator.py
MVrachev: It worked well I think.
You can have a look now.

tests/repository_simulator.py
MVrachev: Removed the check and the exception.

tests/utils.py
sechkova: Understood. Then probably the quotes can be removed because there will be an unused import which pylint will detect.

tests/test_updater_with_simulator.py
```diff
@@ -79,43 +86,45 @@ def test_refresh(self):
...
-            "dir/targetpath": b"dir target content"
-        }
+    targets: DataSet = {
+        "standard case": ("targetpath", b"content", "targetpath"),
+        "non-asci case": ("åäö", b"more content", "%C3%A5%C3%A4%C3%B6"),
+        "subdirectory case": ("a/b/c/targetpath", b"dir target content", "a%2Fb%2Fc%2Ftargetpath"),
+    }
+
+    @utils.run_sub_tests_with_dataset(targets)
+    def test_targets(self, test_case_data: Tuple[str, str]):
```
sechkova: Since you've defined `class TargetData`, this type annotation is no longer true or at least it should be `Tuple[str, str, str]`. Not sure what would be the opinion of `mypy` actually.

MVrachev: I got no complaints when I was using a `Tuple` and now when I used `TargetData`, but it's better and correct to use `TargetData`.
Thanks for noticing.

tests/utils.py
MVrachev: I didn't receive unused import, because `unittest` was actually used to resolve the forward reference.
Still, you are correct I should remove the quotes because they are unnecessary.

sechkova: Interesting, I got the error when I applied pylint to the test file.

tests/utils.py
```diff
@@ -49,8 +49,8 @@
 # (where N is number of items in dataset), feeding the actual test
 # function one test case at a time
 def run_sub_tests_with_dataset(dataset: DataSet):
-    def real_decorator(function: Callable[["unittest.TestCase", Any], None]):
-        def wrapper(test_cls: "unittest.TestCase"):
+    def real_decorator(function: Callable[[unittest.TestCase, Any], None]):
```
sechkova, line 52: The fix should've been in the previous commit but ok.

sechkova (approved):
Seems ok now.

tuf/ngclient/updater.py
```diff
@@ -229,6 +229,8 @@ def download_target(
                 download URL. Default is the value provided in Updater()
 
         Raises:
+            ValueError: If target_base_url and self._target_base_url are
+                simultaniosly None.
```
jku: typo.

I'd be fine with just saying "ValueError: Invalid arguments" here: the error message will contain the real details and this docstring would likely stay valid if we add other argument validation

tests/test_updater_with_simulator.py
```diff
@@ -79,39 +86,50 @@ def test_refresh(self):
...
 
-    def test_targets(self):
-        targets = {
-            "targetpath": b"content",
-            "åäö": b"more content"
-        }
+    targets: DataSet = {
+        "standard case": ("targetpath", b"content", "targetpath"),
+        "non-asci case": ("åäö", b"more content", "%C3%A5%C3%A4%C3%B6"),
+        "subdirectory case": ("a/b/c/targetpath", b"dir target content", "a%2Fb%2Fc%2Ftargetpath"),
```
jku, line 86: I suppose this works... but you are not actually creating TargetData objects here, just normal tuples

jku (approved):
Looks fine to me. 

Some observations
* The targets testing is still lacking in volume: we should have vastly more variations on this... but let's do that on it's own issue, as result of #1579 
* I think all of the different DataSet types are _probably_ unneeded now -- it was useful in the very crowded decorator implementation but now the different DataSets might confuse more than help :shrug: 

left two more comments but they are nits as well

tests/test_updater_with_simulator.py
```diff
@@ -19,6 +19,13 @@
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
 
+class TargetData(NamedTuple):
+    targetpath: str
+    content: str
```
jku: not str btw, it's bytes

tests/utils.py
MVrachev: 🤦 will fix that.

MVrachev:
> * I think all of the different DataSet types are _probably_ unneeded now -- it was useful in the very crowded decorator implementation but now the different DataSets might confuse more than help 🤷

Yea, maybe you are right, given that as you have pointed out in one of your comments I should create a `TargetData` instance if I want to be correct with the annotations.
Additionally, what is the content of the specific target data becomes clear when I unfold it into specific variables.
I will remove `TargetData` and the specific `DataType` altogether.

sechkova (approved):

Issue #1580, opened 2021-09-15
ngclient: reveal final form (finish API changes)


jku:
There's a bunch of small API changes that I think would still make the API better (more consistent, simpler, safer). The implementations of the individual changes can probably be separate but I think it makes sense to discuss the design as a whole: it'll be a lot easier to discuss when the final form of the API is visible.

* simplify the caching support: I've advocated for just dropping the "cache support" in `updated_targets()` before (since it's not really a specification feature) but... it looks like it can be just simplified
* solve potential issues with target paths as local file paths
* `download_target()` should return the local file path, not force client to guess
* make the API more obvious and consistent: `Updater.updated_targets()` return value is not obvious, and using a list is weird
* `Updater.get_one_valid_targetinfo()`: name is long without adding value
* `Updater.refresh()`: there is no reason to force client to explicitly call this

### Planned API usage

_NOTE: these examples have been updated as the ideas have crystallized: earlier comments may refer to a slightly different example_

```python
    updater = Updater(
        metadata_dir,
        "https://example.com/metadata/",
        targets_cache_dir,
        "https://example.com/targets/",
    )
    # Update metadata for a target, then download target into cache if needed
    info = updater.get_targetinfo("file.txt")
    path = updater.find_cached_target(info)
    if path is None:
        path = updater.download_target(info)

    # Alternatively if client knows the local filename it wants to use
    info = updater.get_targetinfo("file.txt")
    if not updater.find_cached_target(info, "./file.txt"):
        updater.download_target(info, "./file.txt")
```

Things to note there:
* `refresh()` is available but not required: it will be done on first `get_targetinfo()` if not explicitly called before
* the functions for finding a local file and downloading a file work the same way: signatures are almost identical
* both functions return a file path: caller does not have to guess where the file is stored locally
* the local path argument to both functions can be a filename or a directory. If it is a directory, then Updater chooses a safe filename for target path (similar to current functionality). If the client knows the filename it wants, it can just use that as the argument. This allows using the Updater in a caching manner (using the same cache dir over time), or just to download a file without any cache
* no local directories are created by any of the calls: If `download_target()` chooses the filename, it will encode the target path before using it as a filename.
* `find_cached_target()` name is not great but I want it to be a counterpart to download_target (and I believe it's much better than the ambiguous `updated_targets()`)

See #1604 for more details.

### Considered, but not included changes

I considered these as well, but am currently not planning to include them:
[...]

PR #1592, opened 2021-09-27
ngclient: Don't use target path as local path

### LINKED ISSUE #1571: ngclient: downloading targets fails if  targetpath includes subdirectories

jku:
Doing so is not always safe and has various other issues
(like target paths "a/../b" and "b" ending up as the same
local path).

Instead URL-encode the target path to make it a plain filename. This
removes any opportunity for path trickery and removes the need to create
the required sub directories (which we were not doing currently, leading
to failed downloads). URL-encoding encodes much more than we really need
but doing so should not hurt: the important thing is that it encodes
all path separators.

Return the actual filepath as return value. I would like to modify the
arguments so caller could decide the filename if they want to. But I
won't do it now because updated_targets() (the caching mechanism)
relies on filenames being chosen by TUF. The plan is to make it
possible for caller to choose the filename though.

This is clearly a "filesystem API break" for anyone depending on the
actual target file names, and does not make sense if we do not plan to
go forward with other updated_targets()/download_target() changes
listed in #1580.

This is part of bigger plan in #1580
Fixes #1571

---

### DIFF
--- a/tests/test_updater_with_simulator.py
+++ b/tests/test_updater_with_simulator.py
@@ -80,32 +80,37 @@ class TestUpdater(unittest.TestCase):
         self._run_refresh()
 
     def test_targets(self):
-        # target does not exist yet
-        updater = self._run_refresh()
-        self.assertIsNone(updater.get_one_valid_targetinfo("file"))
+        targets = {
+            "targetpath": b"content",
+            "åäö": b"more content"
+        }
 
+        # Add targets to repository
         self.sim.targets.version += 1
-        self.sim.add_target("targets", b"content", "file")
+        for targetpath, content in targets.items():
+            self.sim.add_target("targets", content, targetpath)
         self.sim.update_snapshot()
 
-        # target now exists, is not in cache yet
         updater = self._run_refresh()
-        file_info = updater.get_one_valid_targetinfo("file")
-        self.assertIsNotNone(file_info)
-        self.assertEqual(
-            updater.updated_targets([file_info], self.targets_dir), [file_info]
-        )
+        for targetpath, content in targets.items():
+            # target now exists, is not in cache yet
+            file_info = updater.get_one_valid_targetinfo(targetpath)
+            self.assertIsNotNone(file_info)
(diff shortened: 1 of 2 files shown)


MVrachev (approved):
Looks good to me!
I tried to think if there are problem corner cases, but everything seems fine.

sechkova (approved):
A bit annoying that the tests do not include the case it is supposed to fix (targetpath includes subdirectories) because of #1576 but ok.

jku:
> A bit annoying that the tests do not include the case it is supposed to fix (targetpath includes subdirectories) because of #1576 but ok.

yes and I should have pointed that out in the PR message, sorry about that... we can either merge this with the pinky promise that the real tests with "directories" are going to be added, or alternatively we combine this PR and 1756 and merge the result to develop.

@MVrachev 1756 is yours, any opinions? I'd maybe lean towards merging this as is, and then you rebasing your work -- is that ok?

MVrachev:
> > A bit annoying that the tests do not include the case it is supposed to fix (targetpath includes subdirectories) because of #1576 but ok.
> 
> yes and I should have pointed that out in the PR message, sorry about that... we can either merge this with the pinky promise that the real tests with "directories" are going to be added, or alternatively we combine this PR and 1756 and merge the result to develop.
> 
> @MVrachev 1756 is yours, any opinions? I'd maybe lean towards merging this as is, and then you rebasing your work -- is that ok?

I don't have any strong opinions, but if you prefer merging this pr as it is I am okay.
Maybe this will be faster as #1576 is not just about the tests and there could be a discussion.
If you decide to merge it, then as you have mentioned an update on the commit message will be good to happen.

jku:
ok, let's try this: if it doesn't end up working we can always admit defeat and bring back the old API -- we have not promised ngclient is stable yet

PR #1593, opened 2021-09-27
ngclient: Avoid loading targets metadata twice

### LINKED ISSUE #1578: ngclient: top-level targets is unnecessarily updated twice
ngclient updates top-level targets during `refresh()`. Then on `get_one_valid_targetinfo()` the top-level targets is updated again:
* this is not dangerous and does not break any TrustedMetadataSet constraints
* nothing is downloaded but the local metadata file is re-read and re-parsed

Mostly this is just annoying in the logs because it looks like a bug, with two updates usually one right after the other.

This could be fixed as a special case for top-level targets in the delegation loop but I think a similar issue comes up if other targets are already loaded, but then are needed in `_preorder_depth_first_walk()`. Maybe `_load_targets()` should check if a targets metadata is loaded and avoid re-loading in that case (if a targets is loaded it should be completely final and valid)?

sechkova:
Fixes #1578 

When traversing the delegations tree looking for targets, avoid re-loading already verified targets metadata.

requirements**:

### DIFF
--- a/tests/test_updater_with_simulator.py
+++ b/tests/test_updater_with_simulator.py
@@ -6,11 +6,13 @@
 """Test ngclient Updater using the repository simulator.
 """
 
+import builtins
 import os
 import sys
 import tempfile
 import unittest
 from typing import Optional, Tuple
+from unittest.mock import MagicMock, Mock, patch
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
@@ -229,6 +231,32 @@ class TestUpdater(unittest.TestCase):
         # the snapshot.meta["targets.json"] version by 1 and throws an error.
         with self.assertRaises(BadVersionNumberError):
             self._run_refresh()
+
+    @patch.object(builtins, "open", wraps=builtins.open)
+    def test_not_loading_targets_twice(self, wrapped_open: MagicMock):
+        # Do not load targets roles more than once when traversing
+        # the delegations tree
+
+        # Add new delegated targets, update the snapshot
+        spec_version = ".".join(SPECIFICATION_VERSION)
+        targets = Targets(1, spec_version, self.sim.safe_expiry, {}, None)
+        self.sim.add_delegation("targets", "role1", targets, False, ["*"], None)
+        self.sim.update_snapshot()
+
+        # Run refresh, top-level roles are loaded
+        updater = self._run_refresh()
+        # Clean up calls to open during refresh()
+        wrapped_open.reset_mock()
+
+        # First time looking for "somepath", only 'role1' must be loaded
+        updater.get_targetinfo("somepath")
+        wrapped_open.assert_called_once_with(
+            os.path.join(self.metadata_dir, "role1.json"), "rb"
+        )
+        wrapped_open.reset_mock()
+        # Second call to get_targetinfo, all metadata is already loaded
+        updater.get_targetinfo("somepath")
+        wrapped_open.assert_not_called()
 
 
 if __name__ == "__main__":
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -370,6 +370,11 @@ class Updater:
 
     def _load_targets(self, role: str, parent_role: str) -> Metadata[Targets]:
         """Load local (and if needed remote) metadata for 'role'."""
+
+        # Avoid loading 'role' more than once during "get_targetinfo"
+        if role in self._trusted_set:
+            return self._trusted_set[role]
+
         try:
             data = self._load_local_metadata(role)
             delegated_targets = self._trusted_set.update_delegated_targets(


jku:
Did you consider if we could use RepositorySimulator to test this (to continue avoiding the actual web server tests)?

A similar mocking approach could work or this could even be a feature in RespositorySimulator if you think there might be more demand for it

EDIT: actually the second loading of course loads the _local_ metadata only so the repository side is irrelevant and your mocking should work as is in a RepositorySimulator test

sechkova:
> Did you consider if we could use RepositorySimulator to test this (to continue avoiding the actual web server tests)?

It has to be extended a bit for delegations but why not doing it now.

jku:
We discussed this already but documenting here: An implementation of delegation support for RepositorySimulator exists but not in develop yet

The options for this PR are:
* merge as is with expectation that we update the tests when the delegation support is in
* leave this on back burner until delegation support is in

I think the latter is fine: this is not an urgent change

jku:
okay, current RepositorySimulator should have basic delegation support now

tests/test_updater_ng.py
```diff
@@ -363,6 +364,29 @@ def test_missing_targetinfo(self):
...
+        os.remove(os.path.join(self.client_directory, "role2.json"))
+
+        self.repository_updater.refresh()
+
+        # Create a "spy" for _load_targets()
+        with patch.object(
+            self.repository_updater,
+            "_load_targets",
+            wraps=self.repository_updater._load_targets,
+        ) as wrapped_load_targets:
```
jku: I think this could mock `builtins.open` to actually test for file loads: the test wouldn't then depend so much on knowledge of updater implementation details.

It's not a major thing I guess: If mocking open() is significantly more complex then let's not do it.

sechkova:
Added the test in `test_updater_with_simulator` until a better place comes up (#1641).

sechkova:
Now these changes are not compatible with #1680 ...

tuf/ngclient/updater.py
```diff
@@ -415,9 +415,10 @@ def _preorder_depth_first_walk(
                 logger.debug("Skipping visited current role %s", role_name)
                 continue
 
-            # The metadata for 'role_name' must be downloaded/updated before
-            # its targets, delegations, and child roles can be inspected.
-            self._load_targets(role_name, parent_role)
+            # The metadata for 'role_name' must be loaded
+            # before its targets and delegations can be inspected.
+            if not self._trusted_set.get(role_name):
```
jku: more pythonic IMO:

```suggestion
            if role_name not in self._trusted_set:
```

tests/test_updater_with_simulator.py
```diff
@@ -230,6 +232,32 @@ def test_snapshot_rollback_with_local_snapshot_hash_mismatch(self):
...
+        # Add new delegated targets, update the snapshot
+        spec_version = ".".join(SPECIFICATION_VERSION)
+        targets = Targets(1, spec_version, self.sim.safe_expiry, {}, None)
+        self.sim.add_delegation("targets", "role1", targets, False, ["*"], None)
+        self.sim.update_snapshot()
+
+        # Run refresh, top-level roles are loaded
+        updater = self._run_refresh()
+        # Clean up calls to open during refresh()
+        wrapped_open.reset_mock()
```
jku, line 249: you're adding a test for a specific bug fix, that makes sense... but this obviously leads to _"shouldn't we check that refresh opens correct files the correct number of times as well?"_ Maybe a follow up issue if you don't feel like handling here?

jku (approved):
I like it, mocking `open()` feels like the correct thing to check for and is readable.

LGTM to merge. Left two comments for you to decide.

tests/test_updater_with_simulator.py
sechkova: I was planning to open an issue about testing the loading of metadata. Doing it now ... https://github.com/theupdateframework/python-tuf/issues/1681

tuf/ngclient/updater.py
sechkova: yeah, very pythonic :) it is updated now.

jku:
> Now these changes are not compatible with #1680 ...

oof, I just understood what you mean (originally thought you just referred to a plain merge conflict)... this change looks trivial but with that other PR it means you need to handle the else-case in `_preorder_depth_first_walk()` as well making a it a bit ugly. Let's make sure the approach is reasonable WRT 1680 before merging this.

I think that `_preorder_depth_first_walk()` should not have to handle these different cases (it's complex enough): 
So should _load_targets() handle the "already loaded" case and return the cached one if it is already loaded? I think that makes sense: the top-level metadata is "immutable" so if we have loaded a specific targets, it is the correct one. It'll mean `_load_targets()` operates a little differently than other `_load_*()` but I think that's natural: loading targets is a special case as we do it implicitly during `get_targetinfo()`.

jku (changes requested):
marking "request changes" for the merge conflict plus previous comment

sechkova:
It should be ready now.

jku (approved):
LGTM, thanks

PR #1603, opened 2021-10-04
Snapshot abstraction

mnm678:
This moves generation of snapshot metadata into a new interface so that variations on snapshot generation like snapshot merkle trees in [TAP 16](https://github.com/theupdateframework/taps/blob/master/tap16.md) and snapshot RSA accumulators can be added to the interface and more easily swapped out.

This is a draft pr that is missing a few things:
* impact of changes to snapshot metadata on the fields in timestamp metadata (ie including the merkle tree root hash)
* better utilization of `add_to_snapshot` and `remove_from_snapshot` in the implementation of `ManifestSnapshot` (for example, maintaining `fileinfodict` in `ManifestSnapshot`, rather than re-computing the whole dict every time)
* more tests

I would appreciate any general feedback about the abstraction design, whether this is a good idea, or thoughts about what should/shouldn't be moved to this snapshot object.

cc @znewman01

### DIFF
--- a/tuf/repository_lib.py
+++ b/tuf/repository_lib.py
@@ -47,6 +47,7 @@ from tuf import log
 from tuf import roledb
 from tuf import settings
 from tuf import sig
+from tuf import snapshot
 
 
 # See 'log.py' to learn how logging is handled in TUF.
@@ -90,7 +91,7 @@ def _generate_and_write_metadata(rolename, metadata_filename,
   increment_version_number=True, repository_name='default',
   use_existing_fileinfo=False, use_timestamp_length=True,
   use_timestamp_hashes=True, use_snapshot_length=False,
-  use_snapshot_hashes=False):
+  use_snapshot_hashes=False, snapshot_backend=None):
   """
   Non-public function that can generate and write the metadata for the
   specified 'rolename'.  It also increments the version number of 'rolename' if
@@ -117,7 +118,9 @@ def _generate_and_write_metadata(rolename, metadata_filename,
 
 
   elif rolename == 'snapshot':
-    metadata = generate_snapshot_metadata(metadata_directory,
+    if snapshot_backend == None:
+      snapshot_backend = snapshot.ManifestSnapshot()
+    metadata = snapshot_backend.generate_snapshot_metadata(metadata_directory,
         roleinfo['version'], roleinfo['expires'],
         storage_backend, consistent_snapshot, repository_name,
         use_length=use_snapshot_length, use_hashes=use_snapshot_hashes)
@@ -393,7 +396,7 @@ def _delete_obsolete_metadata(metadata_directory, snapshot_metadata,
     # have a prepended version number even though the repository is now
     # a non-consistent one.
     if metadata_role not in snapshot_metadata['meta']:
-      metadata_role, junk = _strip_version_number(metadata_role,
+      metadata_role, junk = strip_version_number(metadata_role,
           consistent_snapshot)
 
     else:
@@ -442,7 +445,7 @@ def _get_written_metadata(metadata_signable):
 
 
 
-def _strip_version_number(metadata_filename, consistent_snapshot):
+def strip_version_number(metadata_filename, consistent_snapshot):
   """
   Strip from 'metadata_filename' any version number (in the
   expected '{dirname}/<version_number>.rolename.<ext>' format) that
@@ -851,7 +854,7 @@ def get_delegated_roles_metadata_filenames(metadata_directory,
     # Example:  '10.django.json' --> 'django.json'
     consistent = \
       metadata_role.endswith('root.json') or consistent_snapshot == True
-    metadata_name, junk = _strip_version_number(metadata_role,
+    metadata_name, junk = strip_version_number(metadata_role,
       consistent)
 
     if metadata_name.endswith(METADATA_EXTENSION):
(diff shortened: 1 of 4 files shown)


tuf/snapshot.py
```diff
@@ -0,0 +1,258 @@
...
+
+  @abc.abstractmethod
+  def generate_snapshot_metadata(self, metadata_directory, version, expiration_date,
+    storage_backend, consistent_snapshot=False,
+    repository_name='default', use_length=False, use_hashes=False):
+    """
+    <Purpose>
+		 Create the snapshot metadata
+
+    <Arguments>
```
znewman01, line 86: A lot of these seem to me like the belong to the class (and specifically, should correspond to *different* classes in some cases). At least the following:

- consistent_snapshot
- use_length
- use_hashes

tuf/snapshot.py
```diff
@@ -0,0 +1,258 @@
...
+  Defines an interface for abstract snapshot metadata operations
+  to be implemented for a variety of snapshot creation methods,
+  including the classic manifest snapshot metadata and snapshot
+  merkle trees
+  """
+
+  __metaclass__ = abc.ABCMeta
+
+  @abc.abstractmethod
+  def add_to_snapshot(self, rolename : str):
```
znewman01, line 35: naming nit: maybe just `add()` (and similar throughout)?

tuf/snapshot.py
```diff
@@ -0,0 +1,258 @@
...
+# The extension of TUF metadata.
+METADATA_EXTENSION = '.json'
+
+# The metadata filenames of the top-level roles.
+ROOT_FILENAME = 'root' + METADATA_EXTENSION
+TARGETS_FILENAME = 'targets' + METADATA_EXTENSION
+SNAPSHOT_FILENAME = 'snapshot' + METADATA_EXTENSION
+TIMESTAMP_FILENAME = 'timestamp' + METADATA_EXTENSION
+
+class SnapshotInterface():
```
znewman01, line 23: Can we distinguish between:

- the repository state (i.e., all metadata about all current packages)
- the artifact that clients will download

I think that will help make the roles/names of various classes clearer.

tuf/repository_lib.py
```diff
@@ -1541,158 +1544,6 @@ def _get_hashes_and_length_if_needed(use_length, use_hashes, full_file_path,
...
-      version=version,
-      expires=expiration_date,
-      meta=fileinfodict)
-
-
-
-
-
 
 def generate_timestamp_metadata(snapshot_file_path, version, expiration_date,
```
znewman01, line 1548: So, this stays here? That kind of makes sense, but is a little asymmetric.

tuf/snapshot.py
```diff
@@ -0,0 +1,258 @@
...
+  merkle trees
+  """
+
+  __metaclass__ = abc.ABCMeta
+
+  @abc.abstractmethod
+  def add_to_snapshot(self, rolename : str):
+    """
+    <Purpose>
+     Indicate to the snapshot interface that 'rolename'
```
znewman01, line 38: I'm a little confused by this. What does it mean to add a role to a snapshot?

znewman01: IIUC from the PR description, this would allow increased efficiency.

tuf/snapshot.py
```diff
@@ -0,0 +1,258 @@
...
+
+# The metadata filenames of the top-level roles.
+ROOT_FILENAME = 'root' + METADATA_EXTENSION
+TARGETS_FILENAME = 'targets' + METADATA_EXTENSION
+SNAPSHOT_FILENAME = 'snapshot' + METADATA_EXTENSION
+TIMESTAMP_FILENAME = 'timestamp' + METADATA_EXTENSION
+
+class SnapshotInterface():
+  """
+  <Purpose>
```
znewman01, line 25: I'm not sure from the code/naming whether this should have state or it's just a way to swap out implementations.

tuf/snapshot.py
```diff
@@ -0,0 +1,258 @@
...
+
+		<Exceptions>
+			securesystemslib.exceptions.FormatError, if the arguments are improperly
+			formatted.
+
+			securesystemslib.exceptions.Error, if an error occurred trying to generate
+			the snapshot metadata object.
+
+		<Side Effects>
+			The 'root.json' and 'targets.json' files are read.
```
znewman01, line 137: IMO nicer to move the file i/o out of this method.

So the "SnapshotInterface" would be responsible for giving you a serializable snapshot object that you could distribute.

znewman01 (commented):
I like the direction you're going in, I'd just like you to go further :)

Specifically, I'd like to see the verification procedures moved in as well, and in general a move towards a more OOP-y design.

jku:
> I would appreciate any general feedback about the abstraction design, whether this is a good idea, or thoughts about what should/shouldn't be moved to this snapshot object.

I'll try to do that later but I'll leave this feedback for now:

I totally understand you are doing this in the legacy code base because it has loads and loads of repository side code in it and adding one more argument to a function call there is almost certainly the path of least resistance... but I also hope you understand that this makes reviewing much less interesting: It's hard to see the long term value.

I think the project should aim to build repository side components on top of Metadata API -- I am not sure what exactly should be included in those components but a "snapshot update handler"  sounds like a prime candidate and I'd definitely be interested in collaborating on something like that.

lukpueh:
Ping @mnm678, same questions as in https://github.com/theupdateframework/python-tuf/pull/1113#issuecomment-1043028730.

mnm678:
Closing in favor of writing this on top of the new Metadata api.

PR #1604, opened 2021-10-05
Ngclient api polish

jku:
NOTE: This PR including the description has been updated based on initial comments: this means the first comments look a bit strange now...

---
This is part 2 of #1580 (ngclient API tweaks): originally I thought we could just replicate old client API but the more I've used it, the more I've wanted to change it. So here's my proposal.

The PR contains a couple of semi-related changes:
* shorten `get_one_valid_targetinfo()` to `get_targetinfo()`
* Replace the unintuitive `updated_targets()` with `find_cached_target()` which operates almost exactly like `download_target()` but looks at local disk only. 
* Add a `target_dir` constructor argument that both `find_cached_target()` and `download_target()` can use: this lets Updater choose the file name. Both methods also take an optional filepath so clients can control the filename & path.

So the two ways of using the API are:

1. with direct local filepaths (when client knows the local filepath it wants to use)
```python
updater = Updater(...)
info = updater.get_targetinfo("my/target")
if not updater.find_cached_target(info, path)
    updater.download_target(info, path)
```

2. with cache directory (Updater decides filename)
```python
updater = Updater(..., target_dir="cache/dir/")
info = updater.get_targetinfo("my/target")
path = updater.find_cached_target(info)
if path is None:
    path = updater.download_target(info)
# path == "cache/dir/my%2Ftarget"
```

Based on how much better the test code now looks, I think we're approaching the correct design here...

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -24,8 +24,8 @@ High-level description of Updater functionality:
       * :func:`~tuf.ngclient.updater.Updater.get_one_valid_targetinfo()` is
         used to find information about a specific target. This will load new
         targets metadata as needed (from local cache or remote repository).
-      * :func:`~tuf.ngclient.updater.Updater.updated_targets()` can be used to
-        check if target files are already locally cached.
+      * :func:`~tuf.ngclient.updater.Updater.find_cached_target()` can be used
+        to check if a target file is already locally cached.
       * :func:`~tuf.ngclient.updater.Updater.download_target()` downloads a
         target file and ensures it is verified correct by the metadata.
 
@@ -43,27 +43,30 @@ Example::
 
     from tuf.ngclient import Updater
 
-    # Load trusted local root metadata from client metadata cache. Define the
-    # remote repository metadata URL prefix and target URL prefix.
+    # Load trusted local root metadata from client metadata cache. Define
+    # where metadata and targets will be downloaded from.
     updater = Updater(
         repository_dir="~/tufclient/metadata/",
         metadata_base_url="http://localhost:8000/tuf-repo/",
+        target_dir="~/tufclient/downloads/",
         target_base_url="http://localhost:8000/targets/",
     )
 
     # Update top-level metadata from remote
     updater.refresh()
 
-    # Securely download a target:
-    # Update target metadata, then download and verify target
-    targetinfo = updater.get_one_valid_targetinfo("file.txt")
-    updater.download_target(targetinfo, "~/tufclient/downloads/")
+    # Update metadata, then download target if needed
+    info = updater.get_one_valid_targetinfo("file.txt")
+    path = updater.find_cached_target(info)
+    if path is None:
+        path = updater.download_target(info)
+    print(f"Local file {path} contains target {info.path}")
 """
 
 import logging
 import os
 import tempfile
-from typing import List, Optional, Set, Tuple
+from typing import Optional, Set, Tuple
 from urllib import parse
 
 from securesystemslib import util as sslib_util
@@ -84,6 +87,9 @@ class Updater:
         repository_dir: Local metadata directory. Directory must be
             writable and it must contain a trusted root.json file.
         metadata_base_url: Base URL for all remote metadata downloads
+        target_dir: Local targets directory. Directory must be writable. It
+            will be used as the default target download directory by
+            ``find_cached_target()`` and ``download_target()``
         target_base_url: Optional; Default base URL for all remote target
             downloads. Can be individually set in download_target()
         fetcher: Optional; FetcherInterface implementation used to download
@@ -98,12 +104,14 @@ class Updater:
         self,
         repository_dir: str,
         metadata_base_url: str,
+        target_dir: Optional[str] = None,
         target_base_url: Optional[str] = None,
         fetcher: Optional[FetcherInterface] = None,
         config: Optional[UpdaterConfig] = None,
     ):
         self._dir = repository_dir
         self._metadata_base_url = _ensure_trailing_slash(metadata_base_url)
+        self.target_dir = target_dir
         if target_base_url is None:
             self._target_base_url = None
         else:
@@ -145,7 +153,7 @@ class Updater:
         """Returns TargetFile instance with information for 'target_path'.
 
         The return value can be used as an argument to
-        :func:`download_target()` and :func:`updated_targets()`.
+        :func:`download_target()` and :func:`find_cached_target()`.
 
         :func:`refresh()` must be called before calling
         `get_one_valid_targetinfo()`. Subsequent calls to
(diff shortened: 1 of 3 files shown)


joshuagl:
> This is draft as I'm still thinking if I should add a new optional constructor argument `target_dir` to Updater:

I think this would be a good change. It's good to have the API versatility to download targets to different locations, but I think downloading all targets to the same location is a common enough pattern that this optional constructor argument makes sense.

sechkova:
> > This is draft as I'm still thinking if I should add a new optional constructor argument `target_dir` to Updater:
> 
> I think this would be a good change. It's good to have the API versatility to download targets to different locations, but I think downloading all targets to the same location is a common enough pattern that this optional constructor argument makes sense.

It sounds good but does the API need to be that flexible, giving the option to set the `target_dir` at two places? 
I am trying to imagine when the `directory` argument of `download_target` could be needed then:
- A client with more than one cache dir wants to call `download_target` on the same `updater` but with a different cache directory?
- The cache dir is not known yet by the client when constructing the Updater but only later during the `download_target` call?

jku:
> It sounds good but does the API needs to be that flexible, giving the option to set the `target_dir` at two places? I am trying to imagine when the `directory` argument of `download_target` could be needed then:

Good point! If we could drop the directory argument completely from the actual `find_cached_target()/download_target()` calls that would be nice. Then there would be two ways to use the API:

1. With cache dir
```
updater = Updater(..., target_dir="my/cache/dir/")
info = updater.get_targetinfo("my/target")
path = updater.find_cached_target(info)
if path is None:
    path = updater.download_target(info)
```
1. With explicit filepaths:
```
updater = Updater(...)
info = updater.get_targetinfo("my/target")
if not updater.find_cached_target(info, "/my/local/target")
    updater.download_target(info, "/my/local/target")
```
_(I'm just assuming refresh() is not required in these examples)_

> * A client with more than one cache dir wants to call `download_target` on the same `updater` but with a different cache directory?
> * The cache dir is not known yet by the client when constructing the Updater but only later during the `download_target` call?

I think nothing prevents target_dir from being a modifiable public attribute that client could change if need be

sechkova:
> Good point! If we could drop the directory argument completely from the actual `find_cached_target()/download_target()` calls that would be nice. Then there would be two ways to use the API:

And we can escape the mutually exclusive directory<->filepath parameters! I think I'm convinced.

jku:
I have incorporated the suggestions and am quite happy with the results (the PR message is also updated so check the current status there): The test code looks a lot more readable so we may be on the right track here. The only thing I don't particularly like is the constructor with 4 or even five arguments but it seems a like a good compromise

I believe the PR is ready for real review.

This will definitely conflict with #1591 : I would expect that one to merge first so I'll clean up afterwards

tuf/ngclient/updater.py
```diff
@@ -44,26 +44,29 @@
     from tuf.ngclient import Updater
 
     # Load trusted local root metadata from client metadata cache. Define the
```
```suggestion
    # Load trusted local root metadata from client metadata cache. Define
```

sechkova (approved):
I have no objections here.

tuf/ngclient/updater.py
```diff
@@ -173,69 +181,83 @@ def get_one_valid_targetinfo(
...
-        destination_directory: str,
+        filepath: Optional[str] = None,
         target_base_url: Optional[str] = None,
     ) -> str:
-        """Downloads the target file specified by 'targetinfo'.
+        """Downloads the target file specified by ``targetinfo``.
+
+        If the ``target_dir`` argument was given to constructor, ``filepath``
+        argument is not required here: in this case a filename will be
+        generated based on the target path.
```
MVrachev: Same comment here  as the one for `find_chaced_target()`.

tuf/ngclient/updater.py
```diff
@@ -173,69 +181,83 @@ def get_one_valid_targetinfo(
...
+        filepath: Optional[str] = None,
+    ) -> Optional[str]:
+        """Checks whether a local file is an up to date target
 
-        After retrieving the target information for the targets that should be
-        updated, updated_targets() can be called to determine which targets
-        have changed compared to locally stored versions.
+        If the ``target_dir`` argument was given to constructor, ``filepath``
+        argument is not required here: in this case a filename will be
+        generated based on the target path like in ``download_target()``
```
MVrachev: What will happen if target_dir is set and filepath is given?
From the code, I see that filepath is with higher priority and will be used.
Do we need to document that?

MVrachev (commented):
Overall, I like the changes.
I have only two small comments.

tuf/ngclient/updater.py
jku: makes sense to document what the arguments default value does at least... I think that (and the text you quote) should cover all the cases?

jku:
Updated docstrings and comments based on feedback (only "ngclient: Simplify caching" (d717775) has changes)

MVrachev (approved):
LGTM!

sechkova:
> This will definitely conflict with #1591 : I would expect that one to merge first so I'll clean up afterwards

There it is, #1591 is merged, could you fix the conflicts in this one?

jku:
CI failure handled in #1628

jku:
rebased to current develop to get tests passing again

tuf/ngclient/updater.py
```diff
@@ -84,6 +87,9 @@ class Updater:
         repository_dir: Local metadata directory. Directory must be
             writable and it must contain a trusted root.json file.
         metadata_base_url: Base URL for all remote metadata downloads
+        target_dir: Local targets directory. Directory must be writable. It
+            will be used as the default target download directory by
+            ``find_cached_target()`` and ``download_target()``
         target_base_url: Optional; Default base URL for all remote target
             downloads. Can be individually set in download_target()
```
lukpueh, line 94: Not really an issue of this PR (so far), but might make  sense in the course of an api polish:

The pairs `repository_dir` - `metadata_base_url`, and  `target_dir` - `target_base_url` both describe the local and remote location respectively for the given file type. Maybe we could use names that show this duality more directly? Here are some naming suggestions

1. Rename `repository_dir` to `metadata_dir`. IMO `repository_dir` sounds more like something that contains metadata, targets and keys (at least that was the case in the old implementation), whereas here it only contains metadata.
2. Use `remote` and `local` prefixes, e.g. `local_metadata_dir` and `remote_metadata_base_url` or, to make the latter shorter, `remote_metadata_dir`?

What do others think?

tuf/ngclient/updater.py
lukpueh: I'd remove the *"like in ``download_target()``"* part. The reader of this docstring might not (and does not need to) know about `download_target` when reading this.

lukpueh: Also, I think the text could be misinterpreted that a previously set `target_dir` has precedence over a passed `filepath`, or that they are somehow used together.

lukpueh: Maybe a slightly more detailed `filepath` arg description would eliminate the need for this paragraph here? See below for a suggestion.

tuf/ngclient/updater.py
```diff
@@ -173,59 +178,63 @@ def get_one_valid_targetinfo(
...
+        generated based on the target path like in ``download_target()``
 
-        for target in targets:
-            # URL encode to get local filename like download_target() does
-            filename = parse.quote(target.path, "")
-            local_path = os.path.join(destination_directory, filename)
+        Args:
+            targetinfo: TargetFile from ``get_targetinfo()``.
+            filepath: Local path to file. By default a filename is generated
+                and file is looked for in ``target_dir`` (see note above).
```
lukpueh: What about this? Feels a bit less ambiguous.

```suggestion
            filepath: Local path to file. If not passed, filepath is generated using 
            `self.target_dir`, plus the url-quoted `path` from the passed `targetinfo`.
```

tuf/ngclient/updater.py
```diff
@@ -173,59 +178,63 @@ def get_one_valid_targetinfo(
...
+            None if file is not found or it is not up to date.
+        """
+        if filepath is not None:
+            pass
+        elif self.target_dir is not None:
+            # Use URL encoded target path as name, like download_target()
+            filename = parse.quote(targetinfo.path, "")
+            filepath = os.path.join(self.target_dir, filename)
+        else:
+            raise ValueError("filepath or target_dir must be set")
```
lukpueh: Is it worth moving above if/elif/else to a helper à la:
```python
def _get_local_target_filepath(filepath: Optional[str] = None) -> str:
```
to use here and in `download_target`?

lukpueh (commented):
Amazing work, @jku et al.! I'm just catching up on the new client and I like it lot. Here are a few comments from reading through `ngclient/updater.py` line by line. 

(Haven't looked at the tests yet)

tuf/ngclient/updater.py
jku: yeah the intent is clearer: I don't think I'll include the actual filename algorithm here (as it's internal to updater and the final path is returned) but otherwise this makes sense.

tuf/ngclient/updater.py
jku: -> #1638

tuf/ngclient/updater.py
jku: yes, thanks, that looks better

jku:
* Rebased to get some test fixes from develop: there are no changes in the existing commits
* added a new commit with the suggested refactor (added `_generate_target_file_path()`) and docstring tweaks

tuf/ngclient/updater.py
jku: I've tried to incorporate this idea and removed the paragraphs in the method docstrings

lukpueh (commented):
Thanks for addressing my comments, @jku, they LGTM! Let me know if you also want me to review the tests. Otherwise, and given @sechkova's and @MVrachev's approvals, I think this is good to go.

PR #1605, opened 2021-10-05
Introduce the idea of trusted/untrusted snapshot

### LINKED ISSUE #1523: ngclient: review snapshot hashes/length check

MVrachev:
Fixes #1523 

If you do the following steps with a repository that contains hashes in meta dicts:
1. call Updater.refresh() and load, verify and cache all metadata files
2. modify timestamp snapshot meta information:
(One or more of hashes or length for snapshot changes here)
3. call Updater.refresh() again
4. root and timestamp will be updated to their latest versions
5. local snapshot will be loaded, but hashes/length will be different
then the ones in timestamp.snapshot_meta and that will prevent loading
6. remote snapshot is loaded and verification starts

then when executing step 6 the rollback checks will not be done because
the old snapshot was not loaded on step 5.

In order to resolve this issue, we are introducing the idea of trusted and
untrusted snapshot.
Trusted snapshot is the locally available cached version. This version has
been verified at least once meaning hashes and length were already checked
against timestamp.snapshot_meta hashes and length.
That's why we can allow loading a trusted snapshot version even if there is a
mismatch between the current timestamp.snapshot_meta hashes/length and
hashes/length inside the trusted snapshot.
Untrusted snapshot is the one downloaded from the web. It hasn't been verified
before and that's why we mandate that timestamp.snapshot_meta hashes and length
should match the hashes and legth calculated on this untrusted version of
snapshot.

As the TrustedMetadataSet doesn't have information on which snapshot is trusted or
not, so possibly the best solution is to add a new argument "trusted"
to update_snapshot.
Even though this is ugly as the rest of the update functions doesn't
have such an argument, it seems the best solution as it seems to work
in all cases:
- when loading a local snapshot, we know the data has at some point been
trusted (signatures have been checked): it doesn't need to match hashes
now
- if there is no local snapshot and we're updating from remote, the
remote data must match meta hashes in timestamp
- if there is a local snapshot and we're updating from remote, the remote
data must match meta hashes in timestamp

Lastly, I want to point out that  hash checks for metadata files are not
essential to TUF security guarantees: they are just an additional layer of
security that allows us to avoid even parsing json that could be malicious -
we already know the malicious metadata would be stopped at metadata
verification after the parsing.

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -124,7 +124,8 @@ class RepositorySimulator(FetcherInterface):
     def targets(self) -> Targets:
         return self.md_targets.signed
 
-    def delegates(self) -> Iterator[Tuple[str, Targets]]:
+    def all_targets(self) -> Iterator[Tuple[str, Targets]]:
+        yield "targets", self.md_targets.signed
         for role, md in self.md_delegates.items():
             yield role, md.signed
 
@@ -266,8 +267,7 @@ class RepositorySimulator(FetcherInterface):
         self.timestamp.version += 1
 
     def update_snapshot(self):
-        self.snapshot.meta["targets.json"].version = self.targets.version
-        for role, delegate in self.delegates():
+        for role, delegate in self.all_targets():
             self.snapshot.meta[f"{role}.json"].version = delegate.version
 
             if self.compute_metafile_hashes_length:
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -50,14 +50,14 @@ from datetime import datetime, timedelta
 import logging
 import os
 import tempfile
-from securesystemslib.hash import digest
+import securesystemslib.hash as sslib_hash
 from securesystemslib.keys import generate_ed25519_key
 from securesystemslib.signer import SSlibSigner
 from typing import Dict, Iterator, List, Optional, Tuple
 from urllib import parse
 
 from tuf.api.serialization.json import JSONSerializer
-from tuf.exceptions import FetcherHTTPError, RepositoryError
+from tuf.exceptions import FetcherHTTPError
 from tuf.api.metadata import (
     Key,
     Metadata,
@@ -99,6 +99,9 @@ class RepositorySimulator(FetcherInterface):
 
         # target downloads are served from this dict
         self.target_files: Dict[str, RepositoryTarget] = {}
+
+        # Whether to compute hashes and legth for meta in snapshot/timestamp
+        self.compute_metafile_hashes_length = False
 
         self.dump_dir = None
         self.dump_version = 0
@@ -243,8 +246,22 @@ class RepositorySimulator(FetcherInterface):
             )
             return md.to_bytes(JSONSerializer())
 
+    def _compute_hashes_and_length(
+        self, role: str
+    ) -> Tuple[Dict[str, str], int]:
+        data = self._fetch_metadata(role)
+        digest_object = sslib_hash.digest(sslib_hash.DEFAULT_HASH_ALGORITHM)
+        digest_object.update(data)
+        hashes = {sslib_hash.DEFAULT_HASH_ALGORITHM:  digest_object.hexdigest()}
+        return hashes, len(data)
+
     def update_timestamp(self):
         self.timestamp.snapshot_meta.version = self.snapshot.version
+
+        if self.compute_metafile_hashes_length:
+            hashes, length = self._compute_hashes_and_length("snapshot")
+            self.timestamp.snapshot_meta.hashes = hashes
+            self.timestamp.snapshot_meta.length = length
 
         self.timestamp.version += 1
 
@@ -252,6 +269,11 @@ class RepositorySimulator(FetcherInterface):
         self.snapshot.meta["targets.json"].version = self.targets.version
         for role, delegate in self.delegates():
             self.snapshot.meta[f"{role}.json"].version = delegate.version
+
+            if self.compute_metafile_hashes_length:
+                hashes, length = self._compute_hashes_and_length(role)
+                self.snapshot.meta[f"{role}.json"].hashes = hashes
+                self.snapshot.meta[f"{role}.json"].length = length
 
         self.snapshot.version += 1
         self.update_timestamp()
--- a/tests/test_updater_with_simulator.py
+++ b/tests/test_updater_with_simulator.py
@@ -169,6 +169,14 @@ class TestUpdater(unittest.TestCase):
         # when the local snapshot is loaded even when there is a hash mismatch
         # with timestamp.snapshot_meta.
 
+        # By raising this flag on timestamp update the simulator would:
+        # 1) compute the hash of the new modified version of snapshot
+        # 2) assign the hash to timestamp.snapshot_meta
+        # The purpose is to create a hash mismatch between timestamp.meta and
+        # the local snapshot, but to have hash match between timestamp.meta and
+        # the next snapshot version.
+        self.sim.compute_metafile_hashes_length = True
+
         # Initialize all metadata and assign targets version higher than 1.
         self.sim.targets.version = 2
         self.sim.update_snapshot()
@@ -177,16 +185,6 @@ class TestUpdater(unittest.TestCase):
         # The new targets should have a lower version than the local trusted one.
         self.sim.targets.version = 1
         self.sim.update_snapshot()
-
-        # Calculate the hash of the new modified version of snapshot.
-        # The purpose is to create a hash mismatch between timestamp.meta and
-        # the local snapshot, but to have hash match between timestamp.meta and
-        # the next snapshot version.
-        digest_object = sslib_hash.digest("sha256")
-        digest_object.update(self.sim._fetch_metadata("snapshot"))
-        new_snapshot_hash = digest_object.hexdigest()
-        self.sim.timestamp.snapshot_meta.hashes = {"sha256": new_snapshot_hash}
-        self.sim.update_timestamp()
 
         # During the snapshot update, the local snapshot will be loaded even if
         # there is a hash mismatch with timestamp.snapshot_meta, because it will
--- a/tests/test_updater_with_simulator.py
+++ b/tests/test_updater_with_simulator.py
@@ -6,18 +6,18 @@
 """Test ngclient Updater using the repository simulator
 """
 
-import logging
 import os
 import sys
 import tempfile
 from typing import Optional, Tuple
-from tuf.exceptions import UnsignedMetadataError
+from tuf.exceptions import UnsignedMetadataError, BadVersionNumberError
 import unittest
 
 from tuf.ngclient import Updater
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
+from securesystemslib import hash as sslib_hash
 
 
 class TestUpdater(unittest.TestCase):
@@ -163,6 +163,39 @@ class TestUpdater(unittest.TestCase):
         self.sim.update_snapshot()
 
         self._run_refresh()
+
+    def test_snapshot_rollback_with_local_snapshot_hash_mismatch(self):
+        # Test triggering snapshot rollback check on a newly downloaded snapshot
+        # when the local snapshot is loaded even when there is a hash mismatch
+        # with timestamp.snapshot_meta.
+
+        # Initialize all metadata and assign targets version higher than 1.
+        self.sim.targets.version = 2
+        self.sim.update_snapshot()
+        self._run_refresh()
+
+        # The new targets should have a lower version than the local trusted one.
+        self.sim.targets.version = 1
+        self.sim.update_snapshot()
+
+        # Calculate the hash of the new modified version of snapshot.
+        # The purpose is to create a hash mismatch between timestamp.meta and
+        # the local snapshot, but to have hash match between timestamp.meta and
+        # the next snapshot version.
+        digest_object = sslib_hash.digest("sha256")
+        digest_object.update(self.sim._fetch_metadata("snapshot"))
+        new_snapshot_hash = digest_object.hexdigest()
+        self.sim.timestamp.snapshot_meta.hashes = {"sha256": new_snapshot_hash}
+        self.sim.update_timestamp()
+
+        # During the snapshot update, the local snapshot will be loaded even if
+        # there is a hash mismatch with timestamp.snapshot_meta, because it will
+        # be considered as trusted.
+        # Should fail as a new version of snapshot will be fetched which lowers
+        # the snapshot.meta["targets.json"] version by 1 and throws an error.
+        with self.assertRaises(BadVersionNumberError):
+            self._run_refresh()
+
 
 if __name__ == "__main__":
     if "--dump" in sys.argv:
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -258,7 +258,9 @@ class TrustedMetadataSet(abc.Mapping):
         if self.timestamp.signed.is_expired(self.reference_time):
             raise exceptions.ExpiredMetadataError("timestamp.json is expired")
 
-    def update_snapshot(self, data: bytes) -> None:
+    def update_snapshot(
+        self, data: bytes, trusted: Optional[bool] = False
+    ) -> None:
         """Verifies and loads 'data' as new snapshot metadata.
 
         Note that an intermediate snapshot is allowed to be expired and version
@@ -271,6 +273,11 @@ class TrustedMetadataSet(abc.Mapping):
 
         Args:
             data: unverified new snapshot metadata as bytes
+            trusted: whether data has at some point been verified by
+                TrustedMetadataSet as a valid snapshot. Purpose of trusted is
+                to allow loading of locally stored snapshot as intermediate
+                snapshot even if hashes in current timestamp meta no longer
+                match data. Default is False.
 
         Raises:
             RepositoryError: data failed to load or verify as final snapshot.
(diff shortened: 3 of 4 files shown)


tests/test_updater_with_simulator.py
```diff
@@ -152,6 +152,33 @@ def test_keys_and_signatures(self):
...
+        self.sim.targets.version = 1
+        self.sim.update_snapshot()
+
+        # Calculate the hash of the new modified version of snapshot.
+        digest_object = sslib_hash.digest("sha256")
+        digest_object.update(self.sim._fetch_metadata("snapshot"))
+        new_snapshot_hash = digest_object.hexdigest()
+        # Assign the new hash to meta in timestamp, so that loading the local
+        # snapshot will fail, but loading the next snapshot version will succeed.
+        self.sim.timestamp.snapshot_meta.hashes = {"sha256": new_snapshot_hash}
```
jku: looking at this with fresh eyes, this seems out of place here: our tests should not be this complex. I don't know where the meta file hashing itself should eventually happen but the test should not have to do any of this... The repo (in this case RepositorySimulator) should instead be able to create `meta` dicts that contains length and hashes.

I think the test should just have something like  `self.sim.metafile_hashes = True` in the beginning of the test and then `self.sim.update_snapshot()` should just work without any manual modification. 

This means `RepositorySimulator.update_snapshot()` (and `update_timestamp()`) would include hashes/length in the meta dicts if the `metafile_hashes` attribute is set.

Can you try if this approach is sensible?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -271,6 +273,8 @@ def update_snapshot(self, data: bytes) -> None:
 
         Args:
             data: unverified new snapshot metadata as bytes
+            trusted: whether to consider the given data as trusted. Default is
+                False.
```
jku: please try rephrasing this: current text tells the reader nothing on top of the variable name. Try answering the questions "what does trusted mean?" and/or "when should I set it to True?" -- might not be easy given how complex the use case is

tests/test_updater_with_simulator.py
```diff
@@ -152,6 +152,33 @@ def test_keys_and_signatures(self):
 
         self._run_refresh()
 
+    def test_snapshot_rollback_check_with_missing_local_snapshot(self):
+        # Test triggering snapshot rollback check on a newly downloaded snapshot
+        # when the local snapshot is not loaded due to hash mismatch.
```
jku: the local snapshot does get loaded as intermediate snapshot (as in, we know it won't be valid as final snapshot but we want to load it for the rollback checks)

jku (changes requested):
I think the commit/PR title is misleading (ans so are some of the comments): the snapshot rollback checks _require_ local snapshot, and do not work without local snapshot. What you are doing is allowing the loading of local snapshot _as intermediate snapshot_ even if the hashes do not match (because, like meta version, the meta hashes reference the new remote snapshot). We want to load the local snapshot even though we already know it's not valid as final because the local snapshot allows us to do rollback checks on the new remote snapshot.

What might be worth mentioning is that the hash checks for metadata files are _not_ essential to TUF security guarantees: they are just an additional layer of security that allows us to avoid even parsing json that could be malicious (we already know the malicious metadata would be stopped at metadata verification after the parsing).

It's a complex situation and because of the complexity we absolutely must be able to clearly express what happens. If we are not able to do that then we should avoid implementing the metafile hashing altogether...

MVrachev:
@jku  I tried to address all of your comments.
First and foremost I changed the commit, pr and test description. 
I know they are lengthy but didn't found another way to describe the situation.
Can you give me some ideas on how to shorten it? 

Second, I experimented with adding hashes and length calculations in the Repository simulator.
What do you think?

Finally, I updated the `trusted` description as requested, but again it's a little long...

tests/repository_simulator.py
```diff
@@ -243,16 +249,64 @@ def _fetch_metadata(self, role: str, version: Optional[int] = None) -> bytes:
             )
             return md.to_bytes(JSONSerializer())
 
+    def _compute_hashes_and_length(
+        self, data: bytes, hash_algorithms: Optional[List[str]]
+    ):
```
jku: for a test implementation, this feels complex: I don't think we need exception handling or configurable algorithms (I think you can just hard code the sslib default)

You're probably thinking about possibly implementing this in Metadata API, but I think we should keep this implementation as simple as possible.

tests/test_updater_with_simulator.py
```diff
@@ -152,6 +152,38 @@ def test_keys_and_signatures(self):
...
+        self.sim.update_snapshot()
+
+        # By raising this flag on timestamp update the simulator would:
+        # 1) compute the hash of the new modified version of snapshot
+        # 2) assign the hash to timestamp.snapshot_meta
+        # The purpose is to create a hash mismatch between timestamp.meta and
+        # the local snapshot, but to have hash match between timestamp.meta and
+        # the next snapshot version.
+        self.compute_metafile_hashes_length = True
+        self.sim.update_timestamp()
```
jku: if you set the option earlier you don't need to call `update_timestamp()` because `update_snapshot()` has already done that (possibly we should rename update_snapshot if it is confusing).

jku (commented):
Are you sure this actually runs the hash computing code? Can you double check with `--dump` or debug prints or something?

the docstrings are indeed hard, I'll maybe try to suggest something when I'm more caffeinated... otherwise left two code suggestions.

tests/repository_simulator.py
```diff
@@ -243,16 +249,64 @@ def _fetch_metadata(self, role: str, version: Optional[int] = None) -> bytes:
...
+            hashes, length = self._compute_hashes_and_length(
+                self._fetch_metadata("snapshot"), hash_algo
+            )
+            self.timestamp.snapshot_meta.hashes = hashes
+            self.timestamp.snapshot_meta.length = length
+
         self.timestamp.version += 1
 
     def update_snapshot(self):
         self.snapshot.meta["targets.json"].version = self.targets.version
```
jku: meta["targets.json"] hashes and length are never set.

Modifying RepositorySimulator is fine if it makes things easier: maybe `self.delegates()` should actually be `self.all_targets()` (that returns targets+delegatedtargets)

tests/test_updater_with_simulator.py
MVrachev: I forgot `update_snapshot()` calls `update_timestamp()` and because I had been manually changing timestamp before and I had to call `update_timestamp()` manually I forgot I don't have to do it now.

MVrachev: It will be good if we have a more descriptive name, but not sure what it could be...

tests/repository_simulator.py
MVrachev: If we are using it only in `update_snapshot()` (as it's the case now) I think that makes sense.
If we ever need to access all delegates without `targets` we would change that.

MVrachev:
> Are you sure this actually runs the hash computing code? Can you double check with `--dump` or debug prints or something?

Ohh... I didn't run it because I was assigning `self.compute_metafile_hashes_length` to True
instead of `self.sim.compute_metafile_hashes_length` 🤦. 
Great catch!

I think I addressed your comments.

> the docstrings are indeed hard, I'll maybe try to suggest something when I'm more caffeinated... otherwise left two code suggestions.

Will wait for your other review.

tests/repository_simulator.py
```diff
@@ -121,8 +124,10 @@ def snapshot(self) -> Snapshot:
     def targets(self) -> Targets:
         return self.md_targets.signed
 
-    def delegates(self) -> Iterator[Tuple[str, Targets]]:
-        for role, md in self.md_delegates.items():
+    def all_targets(self) -> Iterator[Tuple[str, Targets]]:
+        all_targets = self.md_delegates
+        all_targets["targets"] = self.md_targets
```
jku: this modifies self.md_delegates, that's not right. 

I'm not against changing how the metadata is stored in RepositorySimulator -- maybe merging md_targets into md_delegates does make sense... but not like this.

I would just yield "targets", self.md_targets here

tests/repository_simulator.py
```diff
@@ -243,16 +248,42 @@ def _fetch_metadata(self, role: str, version: Optional[int] = None) -> bytes:
...
             return md.to_bytes(JSONSerializer())
 
+    def _compute_hashes_and_length(self, data: bytes):
+        hashes = {}
+        hash_algorithms = [sslib_hash.DEFAULT_HASH_ALGORITHM]
+
+        for algorithm in hash_algorithms:
+            digest_object = sslib_hash.digest(algorithm)
+            digest_object.update(data)
+            hashes[algorithm] = digest_object.hexdigest()
```
jku: loop and array (and even pre-initializing the dict) seem unnecessary

tests/repository_simulator.py
```diff
@@ -243,16 +248,42 @@ def _fetch_metadata(self, role: str, version: Optional[int] = None) -> bytes:
...
+
+        return hashes, len(data)
+
     def update_timestamp(self):
         self.timestamp.snapshot_meta.version = self.snapshot.version
 
+        if self.compute_metafile_hashes_length:
+            hashes, length = self._compute_hashes_and_length(
+                self._fetch_metadata("snapshot")
+            )
```
jku: this is just a nit but it might look better and would fit into one line if the function argument was `rolename:str` and the function did the fetch.

jku (changes requested):
I would maybe move the `compute_metafile_hashes_length = True` to the start of the test: we should try to make tests that don't modify the repository in the middle of the test in unrealistic ways and here it seems we should be testing against a repository that includes hashes for all meta versions, not just a single snapshot update.

Can you make sure the client testing plan (document or issue) includes this new repository setting as a configuration we should keep in mind when improving tests (so that this doesn't remain the single test using this configuration)

I left some specific comments, the important ones are the new bug in RepositorySimulator and how the hash computation is still more complex than it needs to be

jku:
As a review of my original RepositorySimulator design:

This use case shows a (sort of) disadvantage of signing on demand: since the meta hashes require signatures to be included in the hash, the snapshot  update now requires signing every targets metadata (which in practice means that files will typically be signed twice per refresh(): once to create snapshot, once to serve the actual file). I think signing on demand is still probably a reasonable/correct approach: manually signing sounds like a lot of work we don't want to do in the tests

MVrachev:
> I would maybe move the `compute_metafile_hashes_length = True` to the start of the test: we should try to make tests that don't modify the repository in the middle of the test in unrealistic ways and here it seems we should be testing against a repository that includes hashes for all meta versions, not just a single snapshot update.

Agreed. Changed that.

> Can you make sure the client testing plan (document or issue) includes this new repository setting as a configuration we should keep in mind when improving tests (so that this doesn't remain the single test using this configuration)

Yes, left a comment here: https://github.com/theupdateframework/python-tuf/issues/1579#issuecomment-940801497
 
> I left some specific comments, the important ones are the new bug in RepositorySimulator and how the hash computation is still more complex than it needs to be

I think I addressed all of your comments.
IMO the only questionable item left are the comments. 
Will appreciate a review from you as one who writes documentation in a more succinct way.

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -271,6 +273,10 @@ def update_snapshot(self, data: bytes) -> None:
 
         Args:
             data: unverified new snapshot metadata as bytes
+            trusted: whether to consider the given data as trusted. The data is
+                trusted if it has been verified by this function. When trusted,
+                the data will be loaded even if there is a hash/length mismatch.
+                Default is False.
```
jku:
```suggestion
            trusted: whether data has at some point been verified by
                TrustedMetadataSet as a valid snapshot. Purpose of trusted is
                to allow loading of locally stored snapshot as intermediate
                snapshot even if hashes in current timestamp meta no longer
                match data. Default is False.
```

:shrug: My version is even longer but it does try to explain the purpose of the argument. Up to you.

jku (approved):
I think this looks correct. 

The case really is kind of a corner case of a corner case so very hard to explain clearly. @joshuagl does this look reasonable to you?

MVrachev:
>🤷 My version is even longer but it does try to explain the purpose of the argument. Up to you.

Liked your version a little more.
Added it.
Additionally, I added a return annotation to `_compute_hashes_and_length`.

tests/repository_simulator.py
```diff
@@ -100,6 +100,9 @@ def __init__(self):
         # target downloads are served from this dict
         self.target_files: Dict[str, RepositoryTarget] = {}
 
+        # Whether to compute hashes and legth for meta in snapshot/timestamp
```
joshuagl, line 103:
```suggestion
        # Whether to compute hashes and length for meta in snapshot/timestamp
```

tests/repository_simulator.py
```diff
@@ -243,16 +247,34 @@ def _fetch_metadata(self, role: str, version: Optional[int] = None) -> bytes:
             )
             return md.to_bytes(JSONSerializer())
 
+    def _compute_hashes_and_length(
+        self, role: str
+    ) -> Tuple[Dict[str, str], str]:
+        data = self._fetch_metadata(role)
+        digest_object = sslib_hash.digest(sslib_hash.DEFAULT_HASH_ALGORITHM)
```
joshuagl, line 254: Nit: Most likely outside of the scope of this PR, but would it be prudent to have our own default value for the hash algorithm, rather than relying on sslib? (Both here and in metadata.py) Would we notice if this was changed in sslib? What would the impact of such an unnoticed change be?

tuf/ngclient/updater.py
```diff
@@ -360,7 +360,7 @@ def _load_snapshot(self) -> None:
                 version = snapshot_meta.version
 
             data = self._download_metadata("snapshot", length, version)
-            self._trusted_set.update_snapshot(data)
+            self._trusted_set.update_snapshot(data, trusted=False)
```
joshuagl:
```suggestion
            self._trusted_set.update_snapshot(data)
```

Nit: we defined a default, should we just use it? Is being explicit here a style recommendation?

joshuagl (approved):
This looks reasonable to me, thank you both.

tuf/ngclient/updater.py
MVrachev: Even though we have a default value I prefer if we are explicit about this.
That way it's easier to distinguish when we consider the data as `trusted` and when we do not.
In this case, I see the default value more as a defense mechanism that `trusted` will be set.

tests/repository_simulator.py
MVrachev: I will prefer it if we don't discuss this here as will slow the merging of this pr.
I created an issue for that https://github.com/theupdateframework/python-tuf/issues/1613.

tuf/ngclient/updater.py
joshuagl: Not requiring any change here. But if we want to force ourselves to be explicit about how we call this internal API, why have the default?

MVrachev: What do you think @jku? Should we have a default value for `trusted`?

jku: I don't care very much, but defining a default in internal API and then explicitly avoiding using that default as defense mechanism seems weird. So I'd say use the default or don't define it in the first place... (but it's really not a required change)

tests/repository_simulator.py
```diff
@@ -243,16 +247,34 @@ def _fetch_metadata(self, role: str, version: Optional[int] = None) -> bytes:
             )
             return md.to_bytes(JSONSerializer())
 
+    def _compute_hashes_and_length(
+        self, role: str
+    ) -> Tuple[Dict[str, str], str]:
```
jku: return annotation seems wrong for length

tuf/ngclient/updater.py
MVrachev: I checked and if I decide to change the code to mandate `trusted` it will have too many other unrelated changes, so the tests could pass. It doesn't worth it.
I the argument from this function call as Joshua suggested.

MVrachev:
The changes I made: 
- followed Joshua's advice and didn't pass `trusted` at all when calling it with downloaded data. 
- fixed the return type annotation error pointed out by Jussi.

jku (approved):

MVrachev:
I am puzzled why the windows checks failed.
Do we want to restart the run or merge it as it is?

jku:
> Update - Currently investigating inability to process Windows jobs. Linux and Mac jobs are not impacted.
> Oct 13, 09:44 UTC

let's try again a little later

PR #1610, opened 2021-10-11
Add support for python 3.10

MVrachev:

Python 3.10 is released on October 4-th 2021 and it seems
logical to add support for it as it doesn't require any major effort
from the project.

For reference read:
https://www.python.org/downloads/release/python-3100/

requirements**:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       # Run regular TUF tests on each OS/Python combination, plus special tests
       # (sslib master) and linters on Linux/Python3.x only.
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
         os: [ubuntu-latest, macos-latest, windows-latest]
         toxenv: [py]
         include:
--- a/setup.py
+++ b/setup.py
@@ -101,6 +101,7 @@ setup(
     'Programming Language :: Python :: 3.7',
     'Programming Language :: Python :: 3.8',
     'Programming Language :: Python :: 3.9',
+    'Programming Language :: Python :: 3.10',
     'Programming Language :: Python :: Implementation :: CPython',
     'Topic :: Security',
     'Topic :: Software Development'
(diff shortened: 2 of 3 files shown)


MVrachev:
It seems I have to do more to add support for python3.10 for the GitHub workflow. 
I had problems when I just added the number `3.10` to do `github/workflow/ci.yml` and then I tried by adding quotes `'` as I saw it from another project. Now, I am not sure I understand the cause of the error.

jku:
looks like server output has changed. I think this PR is not useful: someone needs to setup 3.10 locally or in a container and actually try the server script and the test.

MVrachev:
> looks like server output has changed. I think this PR is not useful: someone needs to setup 3.10 locally or in a container and actually try the server script and the test.

I don't understand why the PR is not useful. Can you elaborate more on this?
Will the dependabot notice that there is a new version and take action?

jku:
> I don't understand why the PR is not useful. 

CI is showing that test utilities are somehow broken in 3.10: that part is useful and making the PR was useful. So this needs someone to setup the test environment, debug the issue and fix it.

But PRs shouldn't be just another type of issue/bug stating "this doesn't work"... Open PRs should be, in general, ready for review. Obviously there's a fuzzy area in between but in this case we could have a comment in the issue stating "test utilities are somehow broken in 3.10, needs to be investigated" instead.

MVrachev:
> > I don't understand why the PR is not useful.
> 
> CI is showing that test utilities are somehow broken in 3.10: that part is useful and making the PR was useful. So this needs someone to setup the test environment, debug the issue and fix it.
> 
> But PRs shouldn't be just another type of issue/bug stating "this doesn't work"... Open PRs should be, in general, ready for review. Obviously there's a fuzzy area in between but in this case we could have a comment in the issue stating "test utilities are somehow broken in 3.10, needs to be investigated" instead.

I am planning to investigate why it's failing. That's why it's still open.
I will create an issue about it and close this one.

MVrachev:
Closing due to CI failures https://github.com/theupdateframework/python-tuf/pull/1610/checks?check_run_id=3861875262
which will be discussed and investigated on issue #1614.

PR #1618, opened 2021-10-13
Reuse computation of hashes and length

MVrachev:
I experimented moving the computation of hashes and length into two separate functions,
so it can be reused instead of copying the same code at a couple of
different places.

Do you think that's better or should we leave it the way it was?

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -50,13 +50,13 @@ from datetime import datetime, timedelta
 import logging
 import os
 import tempfile
-import securesystemslib.hash as sslib_hash
 from securesystemslib.keys import generate_ed25519_key
 from securesystemslib.signer import SSlibSigner
 from typing import Dict, Iterator, List, Optional, Tuple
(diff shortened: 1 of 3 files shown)


jku (commented):
It's not a clear cut case...
* it's code reuse but the amount of code does not get smaller. In fact metadata.py grows -- not a great sign
* within Metadata the functionality is clearly part of BaseFile... but I guess it's moved to standalone functions so you can use it in tests as well?
* tests now imports a private function from Metadata, not to test it but just to use it to avoid three lines of code
* is_delegated_path() relies on the default being sha-256 instead of setting the algorithm explicitly

I think the idea might be sound but I'm maybe not fully convinced this implementation is worth the churn.

MVrachev:
I wanted to experiment with that idea as I was not sure if there is sense to do that as well.
Wanted to hear an opinion from somebody else.
I agree there aren't many advantages to doing that, so I will just close it.

Issue #1600, opened 2021-10-01
Metadata API: Improve docstrings language


jku:
Let's fix some docstrings in metadata.py: some of them document internal details or just repeat the name of the attribute/argument in a way that's not useful for reader:
* Assume reader is not going to read the source code 
* If it's not obvious, explain what argument/attribute is for -- why is it exposed in API?

I'd like to get at least this common line fixed:

> **unrecognized fields**: Dictionary of all unrecognized fields.

This does not explain what they are or why they are exposed in the API.

lukpueh:
Some related thoughts with regards to consistency of docs that might be addressed here:
- Don't use different aliases for the same docs section, e.g. "Args" vs. "Arguments"
- Quote (`'` vs. `"` vs. ``` `` ```) or don't quote all proper names, e.g. args, return values, class names, etc.
- Start all args/raises/returns oneliners with definite, indefinite or no article.
- As @jku says, don't state the obvious, more specifically reduce redundancy in variable names, typing information, method summary and args/raises/returns descriptions.

lukpueh:
> Quote (' vs. " vs. ``) or don't quote all proper names, e.g. args, return values, class names, etc.

Just saw that @jku already has a ticket for this. I say we try standardise to double backtick and see if it looks good in source code and on RTD.

ivanayov:
Arguments left to explain after #1815:
```
signer: ``securesystemslib.signer.Signer`` implementation
```
and some still questionable:
```
data: Metadata content.
...
filename: Path to write the file to.
etc.
```
plus clarify `json/dict representation`

jku:
Further improvements are a good idea but I think I'd like to close this issue -- or alternatively remove the blocker label: I think the docs look good enough for a release :shrug: ?

PR #1620, opened 2021-10-14
Add missing method args docs in metadata API

### LINKED ISSUE #1599: metadata API: add missing arguments docs
We're missing method argument docs in the docstrings for 
```
Root.add_key()
Root.remove_key()
Targets.add_key()
Targets.remove_key()
DelegatedRole.is_delegated_path()
Key.from_securesystemslib_key()
```
https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.html

we should have (preferably one-liner) descriptions for each argument

lukpueh:
Fixes #1599 

Affected methods are:

- Root.add_key()
- Root.remove_key()
- Targets.add_key()
- Targets.remove_key()
- DelegatedRole.is_delegated_path()
- Key.from_securesystemslib_key()

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -575,9 +575,11 @@ class Key:
 
     @classmethod
     def from_securesystemslib_key(cls, key_dict: Dict[str, Any]) -> "Key":
-        """
-        Creates a Key object from a securesystemlib key dict representation
+        """Creates a Key object from a securesystemlib key dict representation
         removing the private key from keyval.
+
+        Args:
+            key_dict: A key in securesystemlib dict representation.
         """
         key_meta = sslib_keys.format_keyval_to_metadata(
             key_dict["keytype"],
@@ -762,9 +764,12 @@ class Root(Signed):
         )
         return root_dict
 
-    # Update key for a role.
     def add_key(self, role: str, key: Key) -> None:
         """Adds new signing key for delegated role 'role'.
+
+        Args:
+            role: The name of the role, for which 'key' is added.
+            key: The signing key to be added for 'role'.
 
         Raises:
             ValueError: If 'role' doesn't exist.
@@ -776,6 +781,10 @@ class Root(Signed):
 
     def remove_key(self, role: str, keyid: str) -> None:
         """Removes key from 'role' and updates the key store.
+
+        Args:
+            role: The name of the role, for which a signing key is removed.
+            key: The identifier of the key to be removed for 'role'.
 
         Raises:
             ValueError: If 'role' doesn't exist or if 'role' doesn't include
@@ -1151,6 +1160,10 @@ class DelegatedRole(Role):
         supported as target path separator. Leading separators are not handled
         as special cases (see `TUF specification on targetpath
         <https://theupdateframework.github.io/specification/latest/#targetpath>`_).
+
+        Args:
+            target_filepath: URL path to a target file, relative to a base
+                targets URL.
         """
 
         if self.path_hash_prefixes is not None:
@@ -1440,6 +1453,10 @@ class Targets(Signed):
     def add_key(self, role: str, key: Key) -> None:
         """Adds new signing key for delegated role 'role'.
 
+        Args:
+            role: The name of the role, for which 'key' is added.
+            key: The signing key to be added for 'role'.
+
         Raises:
             ValueError: If there are no delegated roles or if 'role' is not
                 delegated by this Target.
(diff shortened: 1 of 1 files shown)


tuf/api/metadata.py
```diff
@@ -575,9 +575,11 @@ def to_securesystemslib_key(self) -> Dict[str, Any]:
 
     @classmethod
     def from_securesystemslib_key(cls, key_dict: Dict[str, Any]) -> "Key":
-        """
-        Creates a Key object from a securesystemlib key dict representation
+        """Creates a Key object from a securesystemlib key dict representation
         removing the private key from keyval.
+
+        Args:
+            key_dict: A key in securesystemlib dict representation.
```
lukpueh, line 582: I admit that the description I added is not particularly helpful, but I'm unsure what to do. Can we maybe quickly rethink the API design here? 

From what I see, the relevant difference between `from_dict` and `from_securesystemslib_key` is that the former accepts any `keyval` (including private ones ... in the long run we will probably want to choke here, right?), whereas the latter removes `keyval["private"]`,  if present. 

So maybe we can rename `from_securesystemslib_key` to `from_private_key`?

Also, the method becomes a lot easier if we just pop `keyval["private"]` instead of calling the mystic `format_keyval_to_metadata` function, which does other things that we don't really care for (e.g. vet values using `schema`, temporarily assign `keyid_hash_algorithms`, etc.).

@avelichka, would you care to weigh in? I'm happy to re-phrase the docs here or with #1600, once I understand the purpose of this method better. :)

lukpueh:
While skimming the [the API Reference on RTD](https://theupdateframework.readthedocs.io/en/latest/api/api-reference.html) I noticed more missing args oneliners in:

```
{Target,Timestamp,Snapshot}.update
MetadataDeserializer.deserialize
MetadataSerializer.serialize
SignedSerializer.serialize
JSONDeserializer.deserialize
JSONSerializer.serialize
CanonicalJSONSerializer.serialize
```


Our style guide says, that these oneliners may be omitted if obvious, but if we do so, we should probably fix the RTD-rendering, see e.g. incomplete Parameters sections in [tuf.api.serialization](https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.serialization.html)).


Happy to fix here or ticketize and fix in a follow-up.

jku:
On the other missing args, no strong opinions on the documentation but I'll share my thoughts...

I've avoided working on the serialization so far: it is sort of part of the same API but still not as critical as the metadata module itself. Looking at serialization is only necessary if
1. there's a bug so you're looking at code anyway
2. you're creating a new serialization implementation... so you're looking at code anyway

 
`{Targets,Timestamp,Snapshot}.update()` would be good to document, but they are also something I'd like to take a good hard look at: some of them are not great API in my opinion. As an example `Targets.update()`:
* the current implementation is literally a one-liner that modifies a public dict attribute: we could document doing that instead and have a smaller API surface
* it inserts or replaces the TargetFile... but a Targets handles both TargetFiles and delegations: what if I wanted to modify delegations instead? is it obvious that TargetFile needs a function but delegations do not?

tuf/api/metadata.py
jku: Adding the method was my suggestion, and avoiding poking the securesystemslib key_dict internals may have been as well: none of us were probably entirely sure how much of the dict return value is sort of "implicitly API" or if it's a secursystemslib implementation detail -- are we going to be in trouble later if we just start removing things there. I think you've got a much better idea of this so feel free to propose something.

The purpose is that we have needed roughly the following code in multiple places (this is from test RepositorySimulator where we generate new Metadata):

```
    def create_key(self) -> Tuple[Key, SSlibSigner]:
        return Key.from_securesystemslib_key(sslib_key), SSlibSigner(sslib_key)
```
It turned out quite easy to make the mistake of including the private key in the TUF metadata :) and I wanted to make it easier to avoid that...

Other solutions for this are totally fine too.

jku (approved):
LGTM. 

If you'd like the sslib-key-to-tuf-public-key transform to be redesigned, maybe a new issue/pr makes sense?

tuf/api/metadata.py
jku: oh, and if it wasn't clear: some of the places where this is needed are _outside of python-tuf_: so developers using python-tuf need to be able to do this so it felt like a function in Metadata API was a good idea even if `del sslib_key["private"]` probably would have worked fine

lukpueh: Thanks for the context, @jku! I'll try to condense some of this in the docstring with #1600.

lukpueh:
Thanks for the review, @jku!

Re serialization API: I'm fine with leaving the docstrings as they are.  
Re `{Targets,Timestamp,Snapshot}.update()`: I agree with you and I remember that we had a discussion about this in a related PR. Let me dig that up and create an issue.
Re from_securesystemslib_key: I'm fine with leaving it as is and will try to add some of you insights from above to the docs in #1600

Issue #1582, opened 2021-09-15
tests: Start linting tests


jku:
I believe we should enable linting (`tox -e lint`) for new test files (maybe even for some recent additions). Possibly some exceptions are needed but in general I think the tests would be better if they were linted.

Good start for this would be test_updater_with_simulator.py and repository_simulator.py: it's new code and files are small so the changes are easy -- the real job is figuring out how to select what to lint... Possible options:
* separate directory for legacy (unlinted) tests?
* define list of linted files in tox.ini / setup.cfg ? (I don't like this one as it means default for new files is no lint)
* define list of files excluded from lint in tox.ini & setup.cfg ?
* some other strategy?

joshuagl:
Splitting tests across multiple directories _may_ make sense, but I think the right combination of simple and not prone to error is a list of files excluded from lint? That way, new test files are linted and there's no ambiguity about where to place new tests.

Also, the no-lint list gets shorter as we replace/remove existing tests.

jku:
I notice I did not explicitly state this: we currently have lint files/directories listed in multiple places in tox.ini and then in setup.cfg for static type checking with mypy

MVrachev:
I think I will experiment by renaming all old test files with `<name>_old.py` and experiment with exclusion rules.

sechkova:
> I think I will experiment by renaming all old test files with `<name>_old.py` and experiment with exclusion rules.

Can we use a prefix instead of a suffix so the files appear sorted in a way?

MVrachev:
> > I think I will experiment by renaming all old test files with `<name>_old.py` and experiment with exclusion rules.
> 
> Can we use a prefix instead of a suffix so the files appear sorted in a way?

I thought about adding a suffix instead of a prefix as the filenames will be easier to read.
Although as they are test files for the old legacy code maybe we want to see them grouped and that will make the other files more visible.
Do the others have an opinion?

jku:
I expect prefixes would make more sense, if files need to be mass-renamed

MVrachev:
Another reasonable solution seems to move the old files in a separate folder, but if we move them won't we lose the git history?

MVrachev:
I read the GIT FAQ on this matter https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F and found out that both renames and file relocations will cause the git history to be lost.
That's why I researched for another option that won't change the titles or move the test files.
[...]

PR #1623, opened 2021-10-18
Apply black on tests

MVrachev:
Related to: #1582 

Currently, we are using 4 tools: `black`, `pylint`, `isort` and `mypy`.
We want to run all 4 of them on our tests for the new codebase to make
it more readable and maintainable, but we don't want to run it on the
old test files as they are following the old style guidelines.

In order to achieve that we want to use an exclusion list instead of
an inclusion list as the new code base will be developed in the future
and new test files will appear. On another hand, we don't expect any
more additional test files against the old code, so this list is static.

I decided to hardcode the names we want to exclude because that way we
won't have to remove the Git history, as opposed to renaming or moving
the old test files.
Even though the list is big around 30 files, I think this solution is
fine as this list will only contain files testing the old code meaning
the list will have static content.

I will apply each of the linters in a separate pr.


MVrachev:
After researching the only solution I found to exclude specific tests was to introduce a new config file `pyproject.toml` and use a long multiline regular expression. The `"""` are needed to allow splitting the regular expression into separate lines.
I know it's not beautiful, but at least it does what we want without changing the Git history.

MVrachev:
Replaced by #1624.

PR #1624, opened 2021-10-19
Start linting tests with black, isort and mypy and exclude old test files

MVrachev:
Addresses #1582 

There will be another pr for mypy.

Currently, we are using 4 tools: `black`, `pylint`, `isort` and `mypy`.
We want to run all 4 of them on our tests for the new codebase to make
it more readable and maintainable, but we don't want to run it on the
old test files as they are following the old style guidelines.

In order to achieve that we want to use an exclusion list instead of
an inclusion list as the new code base will be developed in the future
and new test files will appear. On another hand, we don't expect any
more additional test files against the old code, so this list is static.

I decided to hardcode the names we want to exclude because that way we
won't have to remove the Git history, as opposed to renaming or moving
the old test files.
Even though the list is big, around 30 files, I think this solution is
fine as this list will only contain files testing the old code meaning
the list will have static content.

requirements**:


MVrachev:
For context, I am planning to add this list of excluded files in `setup.cfg`  for mypy.
Unfortunately, `mypy` doesn't have good options to exclude multiple files beautifully and that's why I had to use
a regex like this:
```
exclude = (tests\/aggregate_tests.py|tests\/fast_server_exit.py|tests\/simple_https_server.py|tests\/simple_server.py|tests\/slow_retrieval_server.py|tests\/utils.py|test_utils.py|test_repository_lib.py|test_arbitrary_package
_attack.py|test_developer_tool.py|test_download.py|test_endless_data_attack.py|test_extraneous_dependencies_attack.py|test_formats.py|test_indefinite_freeze_attack.py|test_key_revocation_integration.py|test_keydb.py|test_log.p
y|test_mirrors.py|test_mix_and_match_attack.py|test_multiple_repositories_integration.py|test_repository_tool.py|test_replay_attack.py|test_roledb.py|test_root_versioning_integration.py|test_sig.py|test_slow_retrieval_attack.p
y|test_tutorial.py|test_unittest_toolbox.py|test_updater.py|test_updater_root_rotation_integration.py|tests\/repository_data\/generate_project_data.py|tests\/repository_data\/generate.py)
```

which is ugly, but it's not complicated to understand and it works.

tests/repository_simulator.py
```diff
@@ -174,6 +184,7 @@ def publish_root(self):
         logger.debug("Published root v%d", self.root.version)
 
     def fetch(self, url: str) -> Iterator[bytes]:
+        """Fetches data from the given url and yeilds an Iterator"""
```
jku, line 192: returns an Iterator (or yields bytes).

jku (commented):
I think doing all the linters at once is the correct choice (so we get to see the whole configuration at once) but instead of including changes in 8 test files maybe one representative file would be a good start -- the patch becomes quite big?

Mostly it looks good though. There seem to be quite a few pylint disables in the code: I can understand some of them (e.g. test data is longer than line length) but not really others. I saw at least these that seemed like they should be fixed:
* invalid-name: let's fix the variable names or lets change the pylint config to accept "_" as dummy variable: adding individual disables in the code is just spam).
* consider-iterating-dictionary
* redefined-builtin
* no-self-use

With regards to how we choose what to lint: four different almost identical lists in three different files is not great... maybe it's not the end of he world but let's keep an open mind on the possibility of moving files as well.

jku:
@lukpueh I forget, have we discussed dummy-variables-rgx in pylintrc at some point? I rather like using `_` (in places like `prefix, _, name = filename.rpartition(".")`)  and somehow I'm always surprised when it does not work in python-tuf (you'd think I'd remember by now but no...). 

It's not a big deal though: if this is a conscious decision then let's just rename the variables -- there's just one or two anyway

lukpueh:
@jku, I don't think we have. Does pylint choke on `_`? Briefly looking at `pylintrc` and `tuf/api/pylintrc` it shouldn't. If it does, let's revise the patterns. IMHO `_` is rather idiomatic (it's also a pylint default)


FYI, it seems like in the past `junk` was preferred over `_` (see  old [style guide](https://github.com/secure-systems-lab/code-style-guidelines/blob/532c8347402822b0b0a70d4cf93e6690795ebf13/README.md#ignored-values) and e.g. ceb50dd6918755edb4aaff5b4111bd0f4ff7f892). But our revised style guide does not have this preference anymore.

MVrachev:
> I think doing all the linters at once is the correct choice (so we get to see the whole configuration at once) but instead of including changes in 8 test files maybe one representative file would be a good start -- the patch becomes quite big?

I can do that, but also some of the cases and disabled errors of the tests won't be seen that way.

> Mostly it looks good though. There seem to be quite a few pylint disables in the code: I can understand some of them (e.g. test data is longer than line length) but not really others. 

I tried to make sensible decisions for that, but each of the disabled errors should be examined individually. 

> With regards to how we choose what to lint: four different almost identical lists in three different files is not great... maybe it's not the end of he world but let's keep an open mind on the possibility of moving files as well.

Yea, I know it's not beautiful. The problem is each of the test files has it's own format for excluding multiple files.
I wanted to experiment with using `pyproject.yml` for `pylint` and `mypy` as well, but they already have their own configuration files.
I was thinking of a way to define the exclusion list/lists in one file and somehow import it into the configuration files.
Haven't really researched for that. Do you think that would be better @jku?
Another option is to research if we can move `pylint` or `mypy` configuration options inside `pyproject.yml` and use this one alone?

I will do some research on those items later today.

jku:
> I was thinking of a way to define the exclusion list/lists in one file and somehow import it into the configuration files.

Looking at how different the formats are, this will probably get difficult and ugly very quickly.

> Another option is to research if we can move pylint or mypy configuration options inside pyproject.yml and use this one alone?

that would be better than multiple files (still not great but a lot better)

jku:
> Does pylint choke on `_`? Briefly looking at pylintrc and tuf/api/pylintrc it shouldn't.

Hmm, you have a point, I'm not sure what the issue is exactly.

I do remember stumbling on this before and based on the patch Martin did now as well:  @MVrachev if you can figure out which rule forced you to use `disable=invalid-name` that would be good. It may be that we need to include "_" in `good-names`: comparing with upstream pylintrc might be helpful

MVrachev:
@jku I have good news: I was able to move both `pylint` and `mypy` configurations to `pyproject.toml`.
For mypy: I was successful in adding an exclusion list in an almost identical manner as I was going to add it in `setup.cfg`:
```
exclude = "(tests/aggregate_tests.py|tests/fast_server_exit.py|tests/simple_https_server.py|tests/simple_server.py|tests/slow_retrieval_server.py|tests/utils.py|test_utils.py|test_repository_lib.py|test_arbitrary_package_attack.py|test_developer_tool.py|test_download.py|test_endless_data_attack.py|test_extraneous_dependencies_attack.py|test_formats.py|test_indefinite_freeze_attack.py|test_key_revocation_integration.py|test_keydb.py|test_log.py|test_mirrors.py|test_mix_and_match_attack.py|test_multiple_repositories_integration.py|test_repository_tool.py|test_replay_attack.py|test_roledb.py|test_root_versioning_integration.py|test_sig.py|test_slow_retrieval_attack.py|test_tutorial.py|test_unittest_toolbox.py|test_updater.py|test_updater_root_rotation_integration.py|tests/repository_data/generate_project_data.py|tests/repository_data/generate.py)"
```
I didn't add it here as we had discussed that `mypy` changes will be added in a later pr.

After reading different issues It seems as after [PEP 518](https://www.python.org/dev/peps/pep-0518/#tool-table) many projects opt-in using `pyproject.toml` as a place for multi-tool configuration.

So, @jku what do you think?

PS: I had to force-push so many times as I was not able to reproduce a `mypy` linting error from the CI.

MVrachev:
It was not easy, but I think I resolved all of the conflicts.
You can have a look into the specific changes now @jku.

MVrachev:
I noticed the CI run 19 tests here, but for #1636 it runs 20.
Any idea why that happens?

jku:
> You can have a look into the specific changes now @jku.

If you mean e.g. that I could go through all of the `pylint: disable` lines... I would prefer you do it: please try to fix issues instead of disabling lints if you can. If something is not a clear cut case, leave a review comment. After this I can review what's left -- so we make sure we agree that they are appropriate disables.

Actual issues in the code that you find could be fixed with separate PRs (that would get handled before this PR) or as part of this one -- the former approach will make them faster to merge. In fact even the code fixes already in this PR could easily be made in separate smaller PRs: we'd fix the issues first, and then enabling the new config would be easier to review as it would only include the config changes plus the required `pylint: disable` lines?

MVrachev:
> > You can have a look into the specific changes now @jku.
> 
> If you mean e.g. that I could go through all of the `pylint: disable` lines... I would prefer you do it: please try to fix issues instead of disabling lints if you can. If something is not a clear cut case, leave a review comment. After this I can review what's left -- so we make sure we agree that they are appropriate disables.
> 

I think I have used disables only for the places I find appropriate.
In most of the cases, I have followed the linters recommendations and applied them.
Still, I will have another look into them.

> Actual issues in the code that you find could be fixed with separate PRs (that would get handled before this PR) or as part of this one -- the former approach will make them faster to merge. In fact, even the code fixes already in this PR could easily be made in separate smaller PRs: we'd fix the issues first, and then enabling the new config would be easier to review as it would only include the config changes plus the required `pylint: disable` lines?

I will do that on Monday and create new prs about the code changes that make sense from the linters and leave the disables and config changes in separate pr.
Still, it would have been better if I knew that's what you preferred all along and not create those two prs.

MVrachev:
This pr will be replaced with a couple of other prs.
As Jussi has suggested, I will first introduce the automatic linting changes, then pylint and mypy changes and disables, and in the end, I will add a pr with the configuration options.

Closing this pr.

PR #1628, opened 2021-10-20
Add python3.10 and fix test failures on python3.10

### LINKED ISSUE #1614: Test utils fail on every OS when using Python version 3.10
I tried adding support for Python version 3.10 in pr #1610, but there were CI problems with our test utilities when using every platform we are testing against - Windows, Ubuntu and MacOS: 
https://github.com/theupdateframework/python-tuf/pull/1610/checks?check_run_id=3861875262

**Current behavior**:
CI fails when using our test utilities with Python version 3.10.

**Expected behavior**:
Resolve the above issue and add support for Python version 3.10.

MVrachev:
Fixes #1614

Python 3.10 is released on October 4-th 2021 and it seems
logical to add support for it as it doesn't require any major effort
from the project.

For reference read:
https://www.python.org/downloads/release/python-3100/

I tried adding support for Python3.10 in https://github.com/theupdateframework/python-tuf/pull/1610.
Then, we had CI errors due to test failures: https://github.com/theupdateframework/python-tuf/pull/1610/checks?check_run_id=3861875325
The problem comes from the fact that we start a subprocess
executing simple_https_server.py, but then we fail to communicate the
message we expect from the server process to the main process actually
running the test. We expect our custom message to be the first line
printed from the server process, but instead, a deprecation warning is
printed first about the usage of ssl.wrap_socket(). Our custom message
is printed second.
As of Python 3.7 this function has been deprecated:
https://docs.python.org/3/library/ssl.html#ssl.wrap_socket and for
whatever the reason we didn't get a warning when using it before.

My fix does what is suggested in the warning and replaces the usage of
ssl.wrap_socket() by instantiating a ssl.SSLContext object and then
calling SSLContext.wrap_socket().
This removes the warning.

requirements**:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       # Run regular TUF tests on each OS/Python combination, plus special tests
       # (sslib master) and linters on Linux/Python3.x only.
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-latest, macos-latest, windows-latest]
         toxenv: [py]
         include:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       # Run regular TUF tests on each OS/Python combination, plus special tests
       # (sslib master) and linters on Linux/Python3.x only.
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
         os: [ubuntu-latest, macos-latest, windows-latest]
         toxenv: [py]
         include:
--- a/tests/simple_https_server.py
+++ b/tests/simple_https_server.py
@@ -22,8 +22,8 @@
   to verify that https downloads are permitted.
 
 <Reference>
-  ssl.wrap_socket:
-    https://docs.python.org/2/library/ssl.html#functions-constants-and-exceptions
+  ssl.SSLContext.wrap_socket:
+    https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket
 
   SimpleHTTPServer:
     http://docs.python.org/library/simplehttpserver.html#module-SimpleHTTPServer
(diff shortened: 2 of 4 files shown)


MVrachev:
I test locally and found that when you have Python 3.10 installed locally, then both 310 and 3.10 work when defining the new Python version in the `tox.ini` envlist and using it later with `tox -e`.
I think I prefer `3.10`.

tests/simple_https_server.py
```diff
@@ -44,8 +44,9 @@
 httpd = http.server.HTTPServer(('localhost', 0),
    http.server.SimpleHTTPRequestHandler)
 
-httpd.socket = ssl.wrap_socket(
-    httpd.socket, keyfile=keyfile, certfile=certfile, server_side=True)
+context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+context.load_cert_chain(certfile, keyfile)
+httpd.socket = context.wrap_socket(httpd.socket)
```
jku: Does this wrap_socket() not need `server_side=True`?

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = lint,py{36,37,38,39}
+envlist = lint,py{36,37,38,39,3.10}
```
jku: This does create an environment called `py3.10`, but the magic that sets up the correct interpreter doesn't understand the number (or just recognises the "3") so you just get a new environment with default python. `tox -e <envname>` should fail with InterpreterNotFound on a system that does not have that python installed.

jku (changes requested):

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       # Run regular TUF tests on each OS/Python combination, plus special tests
       # (sslib master) and linters on Linux/Python3.x only.
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```
jku, line 18: Format with untyped values bites developer, part 9000 :sob: 

I wonder how many people will independently fix this exact issue in the next year

tox.ini
MVrachev: Ohh, okay. Fixing it.

tests/simple_https_server.py
MVrachev: Strangely it works without it too but will add it as it's logical.

MVrachev:
@jku addressed your comments.

jku (approved):
Thank you

PR #1630, opened 2021-10-22
Metadata API: validate root role names

### LINKED ISSUE #1516: Metadata API validation: validate root roles dictionary content
Spec says this about root roles:

> A role for each of "root", "snapshot", "timestamp", and "targets" MUST be specified in the key list. The role of "mirror" is OPTIONAL. 

The reference to "key list" seems like a mistake (maybe referring to  an earlier design?) but the intent seems clear: the roles dictionary is allowed to contain either the roles for `["root", "snapshot", "timestamp", "targets"]` or the roles for `["root", "snapshot", "timestamp", "targets", "mirror"]` (and I suppose we are not going to implement mirror for now)). Other roles are not valid and not having the first four roles is not valid.

It makes sense to validate this in Root construction/deserialization. 

Aside/thought: _maybe_ it also makes sense to validate in serialization but I think documenting that roles dict is not intended to be modified goes a long way.

MVrachev:
Fixes #1516

Validate that root role names are 4 and that they are exactly
"root", "snapshot", "targets" and "timestamp" as described in
the spec:
https://theupdateframework.github.io/specification/latest/#root-role

Additionally, fix the valid_roots dataset, so each of the cases contains
the top metadata role names inside the roles dictionary.

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -152,10 +152,12 @@ class RepositorySimulator(FetcherInterface):
         timestamp = Timestamp(1, SPEC_VER, self.safe_expiry, snapshot_meta)
         self.md_timestamp = Metadata(timestamp, OrderedDict())
 
-        root = Root(1, SPEC_VER, self.safe_expiry, {}, {}, True)
-        for role in ["root", "timestamp", "snapshot", "targets"]:
+        top_level_md = ["root", "timestamp", "snapshot", "targets"]
+        roles = {role_name: Role([], 1) for role_name in top_level_md}
+        root = Root(1, SPEC_VER, self.safe_expiry, {}, roles, True)
+
+        for role in top_level_md:
             key, signer = self.create_key()
-            root.roles[role] = Role([], 1)
             root.add_key(role, key)
             # store the private key
             if role not in self.signers:
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -56,6 +56,7 @@ from securesystemslib.signer import SSlibSigner
 from typing import Dict, Iterator, List, Optional, Tuple
 from urllib import parse
 
+from tuf.api.metadata import TOP_LEVEL_ROLE_NAMES
 from tuf.api.serialization.json import JSONSerializer
 from tuf.exceptions import FetcherHTTPError
 from tuf.api.metadata import (
@@ -152,11 +153,10 @@ class RepositorySimulator(FetcherInterface):
         timestamp = Timestamp(1, SPEC_VER, self.safe_expiry, snapshot_meta)
         self.md_timestamp = Metadata(timestamp, OrderedDict())
 
-        top_level_md = ["root", "timestamp", "snapshot", "targets"]
-        roles = {role_name: Role([], 1) for role_name in top_level_md}
+        roles = {role_name: Role([], 1) for role_name in TOP_LEVEL_ROLE_NAMES}
         root = Root(1, SPEC_VER, self.safe_expiry, {}, roles, True)
 
-        for role in top_level_md:
+        for role in TOP_LEVEL_ROLE_NAMES:
             key, signer = self.create_key()
             root.add_key(role, key)
             # store the private key
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -142,23 +142,38 @@ class TestSerialization(unittest.TestCase):
                 "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
                 "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
             "roles": { \
+                "root": {"keyids": ["keyid1"], "threshold": 1}, \
+                "timestamp": {"keyids": ["keyid2"], "threshold": 1}, \
                 "targets": {"keyids": ["keyid1"], "threshold": 1}, \
                 "snapshot": {"keyids": ["keyid2"], "threshold": 1}} \
             }',
         "no consistent_snapshot": '{ "_type": "root", "spec_version": "1.0.0", "version": 1, \
             "expires": "2030-01-01T00:00:00Z", \
             "keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"} }}, \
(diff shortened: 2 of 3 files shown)


MVrachev:
The CI failures are fixed in #1631.

tuf/api/metadata.py
```diff
@@ -726,6 +726,10 @@ def __init__(
         super().__init__(version, spec_version, expires, unrecognized_fields)
         self.consistent_snapshot = consistent_snapshot
         self.keys = keys
+        mandatory_roles = ["root", "timestamp", "snapshot", "timestamp"]
```
lukpueh: Did you mean?
```suggestion
        mandatory_roles = ["root", "timestamp", "snapshot", "targets"]
```
IMO, we should define something like a `TOP_LEVEL_ROLE_NAMES` global for this list somewhere.

tuf/api/metadata.py
```diff
@@ -726,6 +726,10 @@ def __init__(
...
+        if not (len(roles) == 4 and all(r in roles for r in mandatory_roles)):
```
lukpueh: I admit I had to look up the docs for `all`. Wouldn't it be easier (and more performant) to do something like:
```suggestion
        if  set(roles) != TOP_LEVEL_ROLE_NAMES :
```

Note `TOP_LEVEL_ROLE_NAMES` could be a global `set` that contains all top-level role names and `roles` is a dictionary so we can rule out duplicates (don't have to check len).

tests/repository_simulator.py
```diff
@@ -152,10 +153,11 @@ def _initialize(self):
         timestamp = Timestamp(1, SPEC_VER, self.safe_expiry, snapshot_meta)
         self.md_timestamp = Metadata(timestamp, OrderedDict())
 
-        root = Root(1, SPEC_VER, self.safe_expiry, {}, {}, True)
-        for role in ["root", "timestamp", "snapshot", "targets"]:
+        roles = {role_name: Role([], 1) for role_name in TOP_LEVEL_ROLE_NAMES}
+        root = Root(1, SPEC_VER, self.safe_expiry, {}, roles, True)
+
+        for role in TOP_LEVEL_ROLE_NAMES:
```
lukpueh, line 159: Note that `set`s are unordered. If the original order, i.e. `["root", "timestamp", "snapshot", "targets"]`, needs to be guaranteed, you can't use `TOP_LEVEL_ROLE_NAMES` here.

tuf/api/metadata.py
```diff
@@ -728,6 +728,9 @@ def __init__(
         super().__init__(version, spec_version, expires, unrecognized_fields)
         self.consistent_snapshot = consistent_snapshot
         self.keys = keys
+        if set(roles) != TOP_LEVEL_ROLE_NAMES:
+            raise ValueError("Role names must be the top-level metadata roles")
+
```
lukpueh, line 734: It might be helpful to include the received, i.e. `roles` and expected, i.e. `TOP_LEVEL_ROLE_NAMES` value in the error message. I can't remember, do we have a general recommendation for error messages in that regard?

lukpueh (commented):
Nice patch, @MVrachev, and thanks for considering my comments from earlier! I have two more comments/questions. Other than that this lgtm.

tuf/api/metadata.py
jku: I'd avoid any complexity. We're preparing for one of two options:
* deserialization: error message won't be useful
* construction of new Root: this is code that we hopefully write once in a repository component and that's it

complex error message would have very limited value in both cases

tuf/api/metadata.py
```diff
@@ -54,7 +54,7 @@
 from securesystemslib.storage import FilesystemBackend, StorageBackendInterface
 from securesystemslib.util import persist_temp_file
 
-from tuf import exceptions
+from tuf import TOP_LEVEL_ROLE_NAMES, exceptions
```
jku: Is there a reason to put this variable outside of metadata.py?

tests/test_metadata_serialization.py
```diff
@@ -142,23 +142,38 @@ def test_role_serialization(self, test_case_data: str):
...
-        "empty keys and roles": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
+        "empty keys": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
             "expires": "2030-01-01T00:00:00Z", "consistent_snapshot": false, \
             "keys": {}, \
-            "roles": {} \
+            "roles": { \
+                "root": {"keyids": ["keyid3"], "threshold": 1}, \
+                "timestamp": {"keyids": ["keyid4"], "threshold": 1}, \
+                "targets": {"keyids": ["keyid1"], "threshold": 1}, \
+                "snapshot": {"keyids": ["keyid2"], "threshold": 1}} \
```
jku: This is fine I guess but you could also leave the keyids array empty so you don't need to use keyids that don't actually have keys defined -- it's ok to have empty keyids with a threshold 1 as the metadata does not need to be useful, just syntactically valid.

This applies to all other data as well: they all refer to non-existing keyids which is a bit unexpected. I would use existing keys (it's fine to use the same keyid for multiple roles) or empty keyids

tests/repository_simulator.py
MVrachev: Good note. I don't think the order matters here, am I right @jku?

Also, I decided to use a set, instead of a list for `TOP_LEVEL_ROLE_NAMES` because of this check:
`if set(roles) != TOP_LEVEL_ROLE_NAMES:` in tuf/api/metadata.py in Root.__init__() will result as `false` even if roles have the same content as TOP_LEVEL_ROLE_NAMES. The reason is that one is a list and the other is a set.

tuf/api/metadata.py
MVrachev: Hmm, I thought that a variable like that could be used anywhere in TUF, but as the only two use cases are tuf/api/metadata.py and repository simulator, then I will just move it.
If other use cases appear we can always move it back.

tests/repository_simulator.py
lukpueh: I think using a set makes sense. If it turns out that there are many use cases that require the same particular order of the top-level role names, we can still add another global that is a list, or make this one an ordered set.

lukpueh (approved):
LGTM!

tests/repository_simulator.py
jku: WFM

jku (approved):
LGTM

PR #1635, opened 2021-10-26
tests: Add ngclient root key rotation tests

### LINKED ISSUE #1607: tests: Updater key rotation tests
This is part of #1579, we want to test the Updater in various key/signature change scenarios in a setup where:
* tests are readable / understandable
* a new test scenario adds as little code as possible
* adding new scenarios should be easy and the added test runtime should be negligible

I have been looking at existing tests (_test_key_revocation_integration.py,  test_root_versioning_integration.py and test_updater_root_rotation_integration.py_) and I believe they all revolve around a few variables that we should be able to turn into a table testing setup where no additional code is needed for most new tests. The most interesting and complex case is root rotations (because root delegates not just to other roles but to "future root role") so I'll start there.

All root key/signature change cases can be defined by listing N sequential root states, each with
* key list
* threshold
* signature list
* expected result

In practice a table like this should do it:

```python
@dataclass
class RootVersion:
    keys: list[int]
    threshold: int
    signatures: list[int]
    result: Optional[Exception] = None

root_cases = {
    "key rotation: 1-of-1 scenario" : [
        RootVersion([1], 1, [1]),
        RootVersion([2], 1, [2,1]),
        RootVersion([2], 1, [2]),
    ],
    "failing threshold decrease: 2-of-2 becomes 1-of-2" : [
        RootVersion([1,2], 2, [1,2]),
        RootVersion([1,2], 1, [2], UnsignedMetadataError),
    ]
}
```
The "key rotation" test case written out would be
* root v1 contains root key1 with threshold of 1, signed by key1
* root v2 contains root key2 with threshold of 1, signed by key1 and key2
* root v2 contains root key2 with threshold of 1, signed by key2

There is no exception listed so Updater is expected to load all the versions correctly

The "failing" test case written out would be
* root v1 contains root key1 and key2 with threshold of 2, signed by key1 and key2
* root v2 contains root key1 and key2 with threshold of 1, signed by key2

Updater should fail to update v2 as it's not signed by both keys as v1 requires

This is not super readable (using numbers to represent keys is admittedly a bit confusing) but similar 3-5 lines can describe quite tricky situations like threshold increases while rotating keys etc

jku:
Add 15 key rotation cases to ngclient tests.

Maybe fixes #1607? This only tests root keys but I think this is the tricky part: maybe there should be a small test for other keys as well...

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -98,7 +98,8 @@ class RepositorySimulator(FetcherInterface):
         self.signed_roots: List[bytes] = []
 
         # signers are used on-demand at fetch time to sign metadata
-        self.signers: Dict[str, List[SSlibSigner]] = {}
+        # keys are roles, values are dicts of {keyid: signer}
+        self.signers: Dict[str, Dict[str, SSlibSigner]] = {}
 
         # target downloads are served from this dict
         self.target_files: Dict[str, RepositoryTarget] = {}
@@ -139,6 +140,11 @@ class RepositorySimulator(FetcherInterface):
         sslib_key = generate_ed25519_key()
         return Key.from_securesystemslib_key(sslib_key), SSlibSigner(sslib_key)
 
+    def add_signer(self, role:str, signer: SSlibSigner):
+        if role not in self.signers:
+            self.signers[role] = {}
+        self.signers[role][signer.key_dict["keyid"]] = signer
+
     def _initialize(self):
         """Setup a minimal valid repository"""
 
@@ -159,18 +165,16 @@ class RepositorySimulator(FetcherInterface):
         for role in TOP_LEVEL_ROLE_NAMES:
             key, signer = self.create_key()
             root.add_key(role, key)
-            # store the private key
-            if role not in self.signers:
-                self.signers[role] = []
-            self.signers[role].append(signer)
+            self.add_signer(role, signer)
+
         self.md_root = Metadata(root, OrderedDict())
         self.publish_root()
 
     def publish_root(self):
         """Sign and store a new serialized version of root"""
         self.md_root.signatures.clear()
-        for signer in self.signers["root"]:
-            self.md_root.sign(signer)
+        for signer in self.signers["root"].values():
+            self.md_root.sign(signer, append=True)
 
         self.signed_roots.append(self.md_root.to_bytes(JSONSerializer()))
         logger.debug("Published root v%d", self.root.version)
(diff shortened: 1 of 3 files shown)


tests/repository_simulator.py
```diff
@@ -134,10 +135,16 @@ def all_targets(self) -> Iterator[Tuple[str, Targets]]:
...
-    def create_key(self) -> Tuple[Key, SSlibSigner]:
+    @staticmethod
+    def create_key() -> Tuple[Key, SSlibSigner]:
         sslib_key = generate_ed25519_key()
         return Key.from_securesystemslib_key(sslib_key), SSlibSigner(sslib_key)
 
+    def add_signer(self, role:str, signer: SSlibSigner):
+            if role not in self.signers:
+                self.signers[role] = {}
+            self.signers[role][signer.key_dict["keyid"]] = signer
```
jku: oops, indent issue here

jku:
I should also fix some linter issues if we're going to enable the linter on tests...

tests/test_updater_key_rotations.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
+
+# Copyright 2021, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Test ngclient Updater using the repository simulator
```
sechkova: nit: the docstring needs to be updated.

tests/test_updater_key_rotations.py
```diff
@@ -0,0 +1,203 @@
...
+            RootVersion([0,1,2,3,4], 3, [0,2,4]),
+            RootVersion([0,1,2,3,4], 3, [0,4,1]),
+            RootVersion([0,1,2,3,4], 3, [0,1,3]),
+            RootVersion([0,1,2,3,4], 3, [0,1,3]),
+        ],
+        "3-of-5, one key rotated" : [
+            RootVersion([0,1,2,3,4], 3, [0,2,4]),
+            RootVersion([0,1,3,4,5], 3, [0,4,1]),
+        ],
+        "3-of-5, one key rotate fails: not signed with 3 keys from last root" : [
```
sechkova: nit: "from new root"

sechkova (commented):
It looks ok to me, I didn't hit the approve button so that you fix the indent and the commits creator :)

sechkova:
ah and yeah, I think we need to add smaller tests for other keys as well, not necessarily with this PR.

sechkova (approved):

jku:
The last push is just rebase on recent RepositorySimulator changes.

> I should also fix some linter issues if we're going to enable the linter on tests...

Not promising it's 100% there (hard to test without merging with martins PR), but most of the issues should be handled: I'll merge this with the current approval.  @MVrachev if there's something exceptionally annoying when you try rebasing, let me know

Issue #1606, opened 2021-10-06
ngclient testing: client workflow sequences


sechkova:
**Description of issue or feature request**:

_If needed, narrow down the scope of this issue and spawn separate ones for top-level metadata update, delegations metadata update, Fetching targets._

**edit:  This issue deals with top-level-metadata updates. Delegated roles and targets will be covered #1641 #1642  #1643 .**

Following the high-level steps described in #1579 implement tests for the various client workflow sequences:

- Tests scenarios from specification (Can we mark these to be our conformance tests?)
- Specific sequences the client should be able to handle go here
- Extracted useful scenarios from legacy tests
- Specific python-tuf sequences (for ex. loading and validating local metadata)

Decide what we want to test sometimes and what should be always included:

- errors / no errors
- what metadata is downloaded?
- what metadata is read from disk?
- what metadata is on disk afterwards?
- are the currently loaded metadata versions correct?
- (some of the same bullet points repeated for targets)
- ? more ?

The tests should be `RepositorySimulator` based.

MVrachev:
> Decide what we want to test sometimes and what should be always included:
errors / no errors
what metadata is downloaded?
what metadata is read from disk?
what metadata is on disk afterwards?
are the currently loaded metadata versions correct?
(some of the same bullet points repeated for targets)
? more ?

When we decide on that we can use it as a start for capturing an ADR resolving #1129?

jku:
> what metadata is downloaded?
> what metadata is read from disk?

I think these are worth testing (because it's so easy to regress into loading something twice or something), but I think making all tests do this would be ugly... I would suggest a specific set of tests that e.g. runs a few  refresh(), get_targetinfo() and download_target() calls with a reasonably complex repository setup, and ensures only the expected downloads and file reads  are done.

> what metadata is on disk afterwards?
> are the currently loaded metadata versions correct?

These should be checked on more tests than just a specific one... at least the root key rotation tests should ensure that local trusted metadata is the expected file.

sechkova:
After experimenting in #1636 with top-level-roles update:

> - Tests scenarios from specification (Can we mark these to be our conformance tests?)

These are probably the most straight-forward test to implement, they tests step by step the detailed client workflow (6565365). In case of a specific need to generate many variations of a test, like in the case of keys rotation, this can be expanded in a seperate test class: #1607 

> - Specific sequences the client should be able to handle go here

These are still to be defined

> - Extracted useful scenarios from legacy tests

To be added

[...]

PR #1636, opened 2021-10-26
ngclient: top-level-roles update tests

sechkova:
Addresses: #1606 

Tries to address the problem of sticking meaningful tests for ngclient/updater
- Adds tests for top-level metadata update following the Detailed client workflow
- Adds one experimental test case for loading metadata. 

Trying to answer some pending questions:
- I like keeping test cases simple and short even though a sequence of refresh -> modify repo -> new refresh can be tempting for combining in one test. This is sort of a personal preference, I'm glad to hear other opinions.
- For this subset of test I think viewing updater as a black box and checking whether files were written on disk is enough. Checking the contents of files seems redundant. We serve exactly the same metadata we've generated from the simulator. Checking whether we've downloaded the correct file version could be enough.
- Testing whether metadata was loaded from cache requires inspecting the internals of Updater or alternatively "spying" the repository's fetch calls.
- Testing if expired metadata from the local cache is loaded and used to verify the new version  requires peeking as deep as 
`Updater._trusted_metadata_set["timestamp] != None`  I don't find it fir for these tests, it seems like TrustedMetadataSet tests should accommodate it. 

Missing tests cases from the spec:
- Fast forward attack recovery
- Consistent snapshot off tests

requirements**:

### DIFF
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -85,6 +85,11 @@ class TestRefresh(unittest.TestCase):
         with open(os.path.join(self.metadata_dir, f"{role}.json"), "rb") as f:
             self.assertEqual(f.read(), expected_content)
 
+    def _assert_version_equals(self, role: str, expected_version: int) -> None:
+        """Assert that local metadata version is the expected"""
+        md = Metadata.from_file(os.path.join(self.metadata_dir, f"{role}.json"))
+        self.assertEqual(md.signed.version, expected_version)
+
     def test_first_time_refresh(self) -> None:
         # Metadata dir contains only the mandatory initial root.json
         self._assert_files_exist(["root"])
@@ -164,18 +169,17 @@ class TestRefresh(unittest.TestCase):
             self.sim.root.version += 1
             self.sim.publish_root()
 
-        root_path = os.path.join(self.metadata_dir, "root.json")
-        md_root = Metadata.from_file(root_path)
+        md_root = Metadata.from_file(
+            os.path.join(self.metadata_dir, "root.json")
+        )
         initial_root_version = md_root.signed.version
 
         updater.refresh()
 
         # Assert that root version was increased with no more
         # than 'max_root_rotations'
-        md_root = Metadata.from_file(root_path)
-        self.assertEqual(
-            md_root.signed.version,
-            initial_root_version + updater.config.max_root_rotations,
+        self._assert_version_equals(
+            "root", initial_root_version + updater.config.max_root_rotations
         )
 
     def test_intermediate_root_incorrectly_signed(self) -> None:
@@ -286,10 +290,7 @@ class TestRefresh(unittest.TestCase):
         with self.assertRaises(ReplayedMetadataError):
             self._run_refresh()
 
-        md_timestamp = Metadata.from_file(
-            os.path.join(self.metadata_dir, "timestamp.json")
-        )
-        self.assertEqual(md_timestamp.signed.version, 2)
+        self._assert_version_equals("timestamp", 2)
 
     def test_new_timestamp_snapshot_rollback(self) -> None:
         # Check for a rollback attack.
@@ -304,10 +305,7 @@ class TestRefresh(unittest.TestCase):
         with self.assertRaises(ReplayedMetadataError):
             self._run_refresh()
 
-        md_timestamp = Metadata.from_file(
-            os.path.join(self.metadata_dir, "timestamp.json")
-        )
-        self.assertEqual(md_timestamp.signed.version, 2)
+        self._assert_version_equals("timestamp", 2)
 
     def test_new_timestamp_expired(self) -> None:
         # Check for a freeze attack
@@ -338,15 +336,8 @@ class TestRefresh(unittest.TestCase):
         with self.assertRaises(RepositoryError):
             self._run_refresh()
 
-        md_timestamp = Metadata.from_file(
-            os.path.join(self.metadata_dir, "timestamp.json")
-        )
-        self.assertEqual(md_timestamp.signed.version, 3)
-
-        md_snapshot = Metadata.from_file(
-            os.path.join(self.metadata_dir, "snapshot.json")
-        )
-        self.assertEqual(md_snapshot.signed.version, 1)
+        self._assert_version_equals("timestamp", 3)
+        self._assert_version_equals("snapshot", 1)
 
     def test_new_snapshot_unsigned(self) -> None:
         # Check for an arbitrary software attack
@@ -382,10 +373,7 @@ class TestRefresh(unittest.TestCase):
         with self.assertRaises(ReplayedMetadataError):
             self._run_refresh()
 
-        md_snapshot = Metadata.from_file(
-            os.path.join(self.metadata_dir, "snapshot.json")
-        )
-        self.assertEqual(md_snapshot.signed.version, 2)
+        self._assert_version_equals("snapshot", 2)
 
     def test_new_snapshot_expired(self) -> None:
         # Check for a freeze attack
@@ -417,15 +405,8 @@ class TestRefresh(unittest.TestCase):
         with self.assertRaises(RepositoryError):
             self._run_refresh()
 
-        md_snapshot = Metadata.from_file(
-            os.path.join(self.metadata_dir, "snapshot.json")
-        )
-        self.assertEqual(md_snapshot.signed.version, 3)
-
-        md_targets = Metadata.from_file(
-            os.path.join(self.metadata_dir, "targets.json")
-        )
-        self.assertEqual(md_targets.signed.version, 1)
+        self._assert_version_equals("snapshot", 3)
+        self._assert_version_equals("targets", 1)
 
     def test_new_targets_unsigned(self) -> None:
         # Check for an arbitrary software attack
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -347,19 +347,15 @@ class TestRefresh(unittest.TestCase):
 
         self._assert_files_exist(["root", "timestamp"])
 
-    # TODO: RepositorySimulator works always with consistent snapshot
-    # enabled which forces the client to look for the snapshot version
-    # written in timestamp (which leads to "Unknown snapshot version").
-    # This fails the test for a snapshot version mismatch.
+    def test_new_snapshot_version_mismatch(self):
+        # Check against timestamp role’s snapshot version
 
-    # def test_new_snapshot_version_mismatch(self):
-    #     # Check against timestamp role’s snapshot version
+        # Increase snapshot version without updating timestamp
+        self.sim.snapshot.version += 1
+        with self.assertRaises(BadVersionNumberError):
+            self._run_refresh()
 
-    #     # Increase snapshot version without updating
-    #     # timestamp's snapshot version
-    #     self.sim.snapshot.version += 1
-    #     with self.assertRaises(BadVersionNumberError):
-    #         self._run_refresh()
+        self._assert_files_exist(["root", "timestamp"])
 
     def test_new_snapshot_version_rollback(self) -> None:
         # Check for a rollback attack
@@ -416,16 +412,15 @@ class TestRefresh(unittest.TestCase):
 
         self._assert_files_exist(["root", "timestamp", "snapshot"])
 
-    # TODO: RepositorySimulator works always with consistent snapshot
-    # enabled which forces the client to look for the targets version
-    # written in snapshot (which leads to "Unknown targets version").
-    # This fails the test for a targets version mismatch.
+    def test_new_targets_version_mismatch(self):
+        # Check against snapshot role’s targets version
 
-    # def test_new_targets_version_mismatch(self):
-    #     # Check against snapshot role’s targets version
-    #     self.sim.targets.version += 1
-    #     with self.assertRaises(BadVersionNumberError):
-    #         self._run_refresh()
+        # Increase targets version without updating snapshot
+        self.sim.targets.version += 1
+        with self.assertRaises(BadVersionNumberError):
+            self._run_refresh()
+
+        self._assert_files_exist(["root", "timestamp", "snapshot"])
 
     def test_new_targets_expired(self) -> None:
         # Check for a freeze attack.
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -8,13 +8,13 @@
 import os
 import sys
 import tempfile
-from typing import List, Optional
 import unittest
 from datetime import datetime, timedelta
+from typing import Iterable, Optional
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
-from tuf.api.metadata import Metadata, TOP_LEVEL_ROLE_NAMES
+from tuf.api.metadata import TOP_LEVEL_ROLE_NAMES, Metadata
 from tuf.exceptions import (
     BadVersionNumberError,
     ExpiredMetadataError,
@@ -24,13 +24,14 @@ from tuf.exceptions import (
 )
 from tuf.ngclient import Updater
 
+
 class TestRefresh(unittest.TestCase):
     """Test update of top-level metadata following
     'Detailed client workflow' in the specification."""
 
     past_datetime = datetime.utcnow().replace(microsecond=0) - timedelta(days=5)
 
-    def setUp(self):
+    def setUp(self) -> None:
         self.temp_dir = tempfile.TemporaryDirectory()
         self.metadata_dir = os.path.join(self.temp_dir.name, "metadata")
         self.targets_dir = os.path.join(self.temp_dir.name, "targets")
@@ -45,7 +46,7 @@ class TestRefresh(unittest.TestCase):
             )
             f.write(root)
 
-    def tearDown(self):
+    def tearDown(self) -> None:
         self.temp_dir.cleanup()
 
     def _run_refresh(self) -> Updater:
@@ -70,13 +71,15 @@ class TestRefresh(unittest.TestCase):
             self.sim,
         )
 
-    def _assert_files_exist(self, roles: List[str]) -> None:
+    def _assert_files_exist(self, roles: Iterable[str]) -> None:
         """Assert that local metadata files exist for 'roles'"""
         expected_files = sorted([f"{role}.json" for role in roles])
         local_metadata_files = sorted(os.listdir(self.metadata_dir))
         self.assertListEqual(local_metadata_files, expected_files)
 
-    def _assert_content_equals(self, role: str, version: Optional[int]=None) -> None:
+    def _assert_content_equals(
+        self, role: str, version: Optional[int] = None
+    ) -> None:
         """Assert that local file content is the expected"""
         expected_content = self.sim._fetch_metadata(role, version)
         with open(os.path.join(self.metadata_dir, f"{role}.json"), "rb") as f:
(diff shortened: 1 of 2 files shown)


tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+        self.sim.root.expires = datetime.utcnow().replace(
+            microsecond=0
+        ) - timedelta(days=5)
+        self.sim.root.version += 1
+        self.sim.publish_root()
+
+        # Final root v4 is up to date
+        self.sim.root.expires = datetime.utcnow().replace(
+            microsecond=0
+        ) + timedelta(days=5)
```
jku: I added RepositorySimulator.safe_expiry for this purpose -- it might be in the wrong place but maybe worth using... you could define a similar one for the expired case to avoid the ugly date math in multiple tests

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+
+        self._run_refresh()
+
+        # Top-level metadata can be found in metadata dir
+        metadata_files_after_refresh = os.listdir(self.metadata_dir)
+        metadata_files_after_refresh.sort()
+        self.assertListEqual(
+            metadata_files_after_refresh,
+            ["root.json", "snapshot.json", "targets.json", "timestamp.json"],
+        )
```
jku: these checks are currently done in this one test only but I wonder if it's worth making this file list check more ergonomic and a one-liner (see `_assert_files()` in test_updater_ng.py, although yours is better in that it uses assertListEqual) and then do this check in more tests -- I'd really like to be sure our local metadata cache contains what we expect in all situations, especially the failure situations...

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+                args, kwargs = call
+                self.assertNotIn("snapshot", args)
+                self.assertNotIn("targets", args)
+
+            args, kwargs = wrapped_download_metadata.call_args_list[0]
+            self.assertIn("root", args)
+            args, kwargs = wrapped_download_metadata.call_args_list[1]
+            self.assertIn("timestamp", args)
+
+    def test_trusted_root_os_error(self):
```
jku: maybe
```suggestion
    def test_trusted_root_missing(self):
```

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+    def test_trusted_root_expired(self):
+        # Local trusted root is expired
+        root_path = os.path.join(self.metadata_dir, "root.json")
+        md_root = Metadata.from_file(root_path)
+        md_root.signed.expires = datetime.utcnow().replace(
+            microsecond=0
+        ) - timedelta(days=5)
+        for signer in self.sim.signers["root"]:
+            md_root.sign(signer)
+        md_root.to_file(root_path)
```
jku: I dislike the pattern where local metadata is modified (unless that's the only way to test something, like  `test_trusted_root_unsigned()`). I think in this case you could instead
* use simulator to create an expired root, 
* refresh (expecting failure)
* check that current local root is now the new expired root
* initialize a new updater: this should succeed since local root is allowed to be expired

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+            self.assertEqual(wrapped_download_metadata.call_count, 2)
+            for call in wrapped_download_metadata.call_args_list:
+                args, kwargs = call
+                self.assertNotIn("snapshot", args)
+                self.assertNotIn("targets", args)
+
+            args, kwargs = wrapped_download_metadata.call_args_list[0]
+            self.assertIn("root", args)
+            args, kwargs = wrapped_download_metadata.call_args_list[1]
+            self.assertIn("timestamp", args)
```
jku: not a huge fan of hooking into updater internal methods...  I think implementing download counters in RepositorySimulator would be more reliable and readable.

I'd also like to see this done for more than one single call to refresh() -- it seems like this only tests a single special case when metadata has not been updated in remote: why not verify that the first refresh downloads what you expect as well?

We could leave this test out of this PR, and file a new issue for download/file open tests. If you'd like to include this already, I'd at least ask you to have another look at the `call_args_list` parsing: it feels vastly more complex than it should be -- I think you should be able to verify the exact argument for two calls instead of doing four separate `assert*In()` calls.

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+
+        with self.assertRaises(UnsignedMetadataError):
+            self._run_refresh()
+
+    def test_max_root_rotations(self):
+        # Root must stop looking for new versions after Y number of
+        # intermediate files were downloaded.
+
+        # Create some big number of root files in the repository
+        highest_repo_root_version = UpdaterConfig.max_root_rotations + 10
```
jku: haha I love this -- why not create 42 different root versions if it means you don't have to modify updater config from default!

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+            initial_root_version + UpdaterConfig.max_root_rotations,
+        )
+
+    def test_intermediate_root_incorrectly_signed(self):
+        # Check for an arbitrary software attack
+
+        # Intermediate root v3 is unsigned
+        self.sim.root.version += 1
+        root_signers = self.sim.signers["root"]
+        self.sim.signers["root"].clear()
```
jku: I think this does not do what you expect: root_signers is also empty after this

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+        # Check for a freeze attack.
+        self.sim.snapshot.expires = datetime.utcnow().replace(
+            microsecond=0
+        ) - timedelta(days=5)
+        self.sim.update_snapshot()
+
+        with self.assertRaises(ExpiredMetadataError):
+            self._run_refresh()
+
+    def test_new_targets_hash_mismatch(self):
```
jku: this test is the only one I don't quite understand... maybe we can catch up on this in chat

jku (changes requested):
I really like how much cleaner most of this is than either the legacy tests and even some of the comparable trustedmetadataset tests: e.g. `test_new_snapshot_version_rollback()` is directly readable -- something I can't say of any rollback tests elsewhere.

The two main issues I have are:
* overall niggles about download count test
* most tests only check raised errors: file existence on disk is not checked, versions or content of the files is not checked. I'd like to see at least explanations of why these are not needed -- especially for the error cases I think we really should be checking that local files are what we expect

Both of these could be solved as followup issues: the tests are certainly already useful.

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+        with open(os.path.join(self.metadata_dir, "root.json"), "bw") as f:
+            root = self.sim.download_bytes(
+                "https://example.com/metadata/1.root.json", 100000
+            )
+            f.write(root)
+
+    def tearDown(self):
+        self.temp_dir.cleanup()
+
+    def _run_refresh(self) -> Updater:
```
MVrachev, line 52: Can you add doc to this function? That was a pylint warning I had to address for all functions besides the ones with names such as `test_*`, `setUp*` and `tearDown*`.
For this function in `test_updater_with_simulator.py` I had added `"""Creates a new updater and runs refresh."""`. You decide if you like it or not.

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,396 @@
...
+        updater = Updater(
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            "https://example.com/targets/",
+            self.sim,
+        )
+        updater.refresh()
+        return updater
+
+    def _init_updater(self) -> Updater:
```
MVrachev, line 64: Can you add doc to this function? That was a pylint warning I had to address for all functions besides the ones with names such as `test_*`, `setUp*` and `tearDown*`.

jku:
> Testing whether metadata was loaded from cache requires inspecting the internals of Updater or alternatively "spying" the repository's fetch calls.

I think we agree on the possible checks but I'll just reiterate so there's no confusion:
1. checking that expected metadata is read from cache 
   * can be done by hooking into updater internals...
   * but I _think_ wrapping `open()` would work just as well
1. checking that the expected files are downloaded from remote
   * can be done by hooking into updater internals...
   * but adding some download counters into RepositorySimulator could be cleaner and more maintainable
1. checking if cached metadata is used to verify new metadata downloaded from remote 
   * trickier but I suppose we can rely on negative tests here: tests that are expected to fail on that verification
1. checking that correct metadata is in cache after successful or failed refresh() 
   * this seems essential to me
   * in practice could mean just checking the version number but comparing the bytes might be as easy
1. checking refresh() return value

There's no need to do all of these for all tests : 1 and 2 could be checked by just some tests, 3 might not need explicit checks but I think 4 is something we should check almost as extensively as 5

tests/test_updater_top_level_update.py
sechkova: I added some convenience methods and a lot of additional checks with 5394c471046d8250aa8a7608b777c6ef23a41752.

tests/test_updater_top_level_update.py
sechkova: oh well ... in order to test the default config :)) 
I updated the test with randomly chosen smaller values.

tests/test_updater_top_level_update.py
sechkova: oh :facepalm:

tests/test_updater_top_level_update.py
sechkova: as we already discussed, I've assumed incorrectly what `compute_metafile_hashes_length` does and in the same time its implementation is buggy. I opened  #1651 and updated the tests.

tests/test_updater_top_level_update.py
sechkova: I removed the entire commit. I'll open a follow up issue about testing whether cached metadata is loaded and verified as expected.

I did the weird parsing since the full call list was something like call("root", 123242, 2) but I agree it is ugly.

sechkova:
After iterating over the tests again, I added much more checks related to expected contents on disk after refresh. 
I will open follow up issues about expected cached/downloaded metadata, etc.

tests/test_updater_top_level_update.py
```diff
@@ -67,10 +68,21 @@ def _init_updater(self) -> Updater:
...
 
+    def _assert_files_exist(self, roles: List[str]) -> None:
+        """Assert that local metadata files exist for 'roles'"""
+        expected_files = sorted([f"{role}.json" for role in roles])
+        local_metadata_files = sorted(os.listdir(self.metadata_dir))
+        self.assertListEqual(local_metadata_files, expected_files)
+
+    def _assert_content_equals(self, role: str, version: Optional[int]=None) -> None:
+        """Assert that local file content is the expected"""
+        expected_content = self.sim._fetch_metadata(role, version)
```
sechkova, line 84: I've abused using the internal `_fetch_metadata`. Should we make it public? Much more convenient when you want to get the repository content in bytes than download_bytes(url, size ...).

jku: Agreed we should somehow make that ergonomic... Your proposal sounds ok to me. 

The only alternative I can think of is adding a  `RepositorySimulator.assert_metadata(directory: str, roles: List[str])` which asserts that `directory` contains metadata matching current repository state for `roles` -- possibly there's a good default value for `roles` as well.

You probably would not need _assert_files_exist() anymore as the new function could do that at the same time.

But I guess then you'd need to handle the failing cases somehow differently... Your call, could leave this as enhancement.

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,462 @@
...
+        self.temp_dir.cleanup()
+
+    def _run_refresh(self) -> Updater:
+        """Create a new Updater instance and refresh"""
+        updater = Updater(
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            "https://example.com/targets/",
+            fetcher=self.sim,
+        )
```
jku: I should have made more noise about the recent constructor API change (sorry): please add the target dir argument to this call

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,462 @@
...
+        # saved as a local root.
+        updater = self._init_updater()
+        with self.assertRaises(ExpiredMetadataError):
+            updater.refresh()
+
+        self._assert_files_exist(["root"])
+        self._assert_content_equals("root", 2)
+
+        # The expiration of the trusted root metadata file does not lead
+        # to failure and Updater in successfully initialized.
```
jku: something wrong with sentence end here but maybe this could be shorter as well:

```suggestion
        # Local root metadata can be loaded even if expired
```

in general I really appreciate it if we manage to keep comments in tests to single lines -- makes it easier to read

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,462 @@
...
+        # Create some number of roots greater than 'max_root_rotations'
+        while self.sim.root.version < updater.config.max_root_rotations + 3:
+            self.sim.root.version += 1
+            self.sim.publish_root()
+
+        root_path = os.path.join(self.metadata_dir, "root.json")
+        md_root = Metadata.from_file(root_path)
+        initial_root_version = md_root.signed.version
+
+        updater.refresh()
```
jku, line 177: is there really no error here? I would expect an error. can you check legacy updater and possibly file an issue for ngclient?

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,462 @@
...
+        # Modify snapshot contents without updating
+        # timestamp's snapshot hash
+        self.sim.snapshot.expires += timedelta(days=1)
+        self.sim.snapshot.version += 1  # snapshot v2
+        self.sim.timestamp.snapshot_meta.version = self.sim.snapshot.version
+        self.sim.timestamp.version += 1  # timestamp v3
+
+        # Hash mismatch error
+        with self.assertRaises(RepositoryError):
+            self._run_refresh()
```
jku, line 337: I assume the real error is something more specific?

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,462 @@
...
+        self.sim.signers["snapshot"].clear()
+        with self.assertRaises(UnsignedMetadataError):
+            self._run_refresh()
+
+        self._assert_files_exist(["root", "timestamp"])
+
+    # TODO: RepositorySimulator works always with consistent snapshot
+    # enabled which forces the client to look for the snapshot version
+    # written in timestamp (which leads to "Unknown snapshot version").
+    # This fails the test for a snapshot version mismatch.
```
jku: An option is to change the simulator so it does always return metadata, even if the requested version does not match the actual version. I don't think it would break anything... Feel free to just file an issue for this TODO though

jku (commented):
Looks quite nice to me. Left a few smaller comments.

I think the assert methods could still be improved... but maybe as a follow up? Just to document what it looks like to me, it seems that there are three separate assert types:
* _assert_files_exist
* _assert_content_equals
* _assert_version_equals (this function does not exist but same four lines are repeated a lot). This is mostly used in error cases where client version should not match repository version

The issue I have with this is that it's hard to easily see if a specific test is using the correct asserts or not... Anyway i can see it's not trivial to solve so we can leave as follow up task

tests/test_updater_top_level_update.py
sechkova: Surprisingly (for me too), this is the correct behaviour. The spec treats this case the same way as if the next root version was not found. The legacy code does the same thing.

>  If this file is not available, or we have downloaded more than Y number of root metadata files (because the exact number is as yet unknown), then go to step 5.3.10.

tests/test_updater_top_level_update.py
sechkova: For reasons that I cannot remember now, `LengthOrHashMismatchError` does not inherit from `RepositoryError` but the latter is raised from it:
https://github.com/theupdateframework/python-tuf/blob/8ae944ccb256ba0aa13a62f691fd79d4e16c11ad/tuf/ngclient/_internal/trusted_metadata_set.py#L303

As a result `LengthOrHashMismatchError` is not raised here. Should we reconsider this usage?

tests/test_updater_top_level_update.py
sechkova: Did it in e9e5965. This means that consistent_snapshot with non-root metadata means nothing for the simulator, though.

sechkova:
I did some improvement of the asserts as you suggested and opened #1669 to figure out a smarter way for asserting the local metadata state.

tests/test_updater_top_level_update.py
jku: that seems like a very weird choice... thanks for checking

tests/test_updater_top_level_update.py
jku: Thanks. Hmm, maybe we should reconsider that, but this PR seems fine

tests/test_updater_top_level_update.py
jku: well, it still supports consistent_snapshot... but only the latest/current consistent snapshot: trying to download using previous snapshot versions will fail in interesting ways (because the metadata will be incorrect). I think this is fine?

tests/test_updater_top_level_update.py
```diff
@@ -0,0 +1,435 @@
...
+
+        self._assert_files_exist(["root", "timestamp", "snapshot"])
+
+    def test_new_targets_version_mismatch(self):
+        # Check against snapshot role’s targets version
+
+        # Increase targets version without updating snapshot
+        self.sim.targets.version += 1
+        with self.assertRaises(BadVersionNumberError):
+            self._run_refresh()
```
jku, line 421: no _assert_files_exist() here?

jku (approved):
Yeah looks nice: the assertion changes really improve readability, I can see what's being tested.

No need to block merging for this... but are you able to test how much runtime these tests add? It is a big chunk of tests that creates and verifies _a lot_ of metadata... I think there's a lot of optimization we could do in RepositorySimulator and how we use it but I'm not interested in doing that unless it's a real bottleneck

tests/test_updater_top_level_update.py
sechkova: Oh yeah, I amended the last commit, thanks.

sechkova:
> No need to block merging for this... but are you able to test how much runtime these tests add? It is a big chunk of tests that creates and verifies _a lot_ of metadata... I think there's a lot of optimization we could do in RepositorySimulator and how we use it but I'm not interested in doing that unless it's a real bottleneck

Well testing time is relative but RepositorySimulator seems to be performing good enough for our criteria. 
Some super quick results:

```
> time python test_updater_top_level_update.py
........................
Ran 24 tests in 0.091s

OK

real	0m0.271s
user	0m0.217s
sys	0m0.051s
```
In comparison with test_updater_ng.py and test_api.py which use the pre-generated data on disk:
```
> time python test_updater_ng.py 
.........
Ran 9 tests in 0.248s

OK

real	0m0.432s
user	0m0.350s
sys	0m0.076s

> time python test_api.py 
..................
Ran 18 tests in 0.208s

OK

real	0m0.336s
user	0m0.302s
sys	0m0.029s
```

and key rotation tests:
```
> time python test_updater_key_rotations.py
.
Ran 1 test in 0.100s

OK

real	0m0.272s
user	0m0.231s
sys	0m0.040s
```

In total the whole test suite now takes ~25 seconds to complete in my environment.

PR #1649, opened 2021-10-28
Clarify key rotations test cases by using keywords

MVrachev:

As I mentioned in my comment here https://github.com/theupdateframework/python-tuf/issues/1607#issuecomment-953160724 while I was reviewing #1635 it was a little hard for me to remember the placement of the arguments for RootVersion considering they are 15 root rotations tested and the dataset becomes a larger piece of code.

That's why I suggest using keywords whenever possible to do that without splitting the `RootVersion` instance creation into two lines.

requirements**:

### DIFF
--- a/tests/test_updater_key_rotations.py
+++ b/tests/test_updater_key_rotations.py
@@ -85,6 +85,7 @@ class TestUpdaterKeyRotations(unittest.TestCase):
         )
         updater.refresh()
 
+    # fmt: off
     root_rotation_cases = {
         "1-of-1 key rotation": [
             RootVersion(keys=[1], threshold=1, sigs=[1]),
@@ -154,6 +155,7 @@ class TestUpdaterKeyRotations(unittest.TestCase):
             RootVersion(keys=[1, 2], threshold=2, sigs=[1, 2]),
         ],
     }
+    # fmt: on
 
     @run_sub_tests_with_dataset(root_rotation_cases)
     def test_root_rotation(self, root_versions: List[RootVersion]) -> None:
--- a/tests/test_updater_key_rotations.py
+++ b/tests/test_updater_key_rotations.py
@@ -26,8 +26,8 @@ from tuf.ngclient import Updater
 class RootVersion:
     keys: List[int]
     threshold: int
-    signatures: List[int]
-    result: Optional[Type[Exception]] = None
+    sigs: List[int]
+    res: Optional[Type[Exception]] = None
 
 
 class TestUpdaterKeyRotations(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tests/test_updater_key_rotations.py
```diff
@@ -87,71 +87,71 @@ def _run_refresh(self) -> None:
...
-            RootVersion([2], 1, [3, 2, 4]),
+            RootVersion(keys=[1], threshold=1, sigs=[3, 1, 4]),
+            RootVersion(keys=[2], threshold=1, sigs=[3, 2, 1, 4]),
+            RootVersion(keys=[2], threshold=1, sigs=[3, 2, 4]),
         ],
         "1-of-1 key rotation fail: not signed with old key": [
-            RootVersion([1], 1, [1]),
-            RootVersion([2], 1, [2, 3, 4], UnsignedMetadataError),
+            RootVersion(keys=[1], threshold=1, sigs=[1]),
+            RootVersion([2], 1, sigs=[2, 3, 4], res=UnsignedMetadataError),
```
jku: this variance is IMO making it harder to read than it was -- now it's really hard to see what changed between these two versions whereas in the original that's trivial. 

I think adding the names makes sense, but only if it doesn't impact the ability to notice changes.

jku: maybe changing res to a boolean would help (unlikely that we'll need other errors)... or maybe we can allow longer lines for this part of the code somehow

MVrachev: I think it's best if we add keywords for all cases and disable long lines warning for those where we have problems.
That way the code will be consistent. 
The long cases are little hard to follow now, but with added keywords they will be much better.

jku: a potential issue with that plan is that we currently enforce line length with 3 (!) different tools: two of those would probably require a rule disable here. Tools are:
* black: this makes sense, it actually formats the code
* isort: also makes sense: it modifies imports so needs to know line length (isort is also not a problem in this case since it only cares about imports)
* pylint: this may be unnecessary? why does pylint also need to check the length?

jku: > * pylint: this may be unnecessary? why does pylint also need to check the length?

It is unnecessary, we can disable the whole "format" checker. Even pylints own pylintrc does that  https://github.com/PyCQA/pylint/blob/main/pylintrc#L80 :joy:

MVrachev: pylint format option is disabled in #1659.

Will update the pr after it's merged.

MVrachev:
#1659  was merged and I updated the pr to use keywords in all test cases.
@jku do you think I should rename `res` back to `results` and `sigs` back to `signatures`?

jku (approved):
LGTM. Is black going to accept these without complaints when you enable black?

MVrachev:
Ohh... I forgot to run black on that change.
It's complaining...
I disabled  `black` with `# fmt: off` for the test cases and then enabled it again with `# fmt: on` after them.

Issue #1641, opened 2021-10-27
ngtests: Add tests for delegated roles metadata update


sechkova:
**Description of issue or feature request**:

Add tests for updating delegated roles metadata following [5.6.7](https://theupdateframework.github.io/specification/v1.0.25/index.html#update-targets) from the specification. A preliminary list of expected test cases:

- Successful update finds target's metadata
- Successful update does not find target's metadata
- Not loading targets metadata twice when looking for a target file
- TargetInfo is what was expected
- Visited max number of roles
- Terminating delegations

The tests should:
- Use `RepositorySimulator` for serving test files in-memory.
- "black-box" Updater, don't look at its internals, check if files are downloaded as expected (or not downloaded as expected).

jku:
Leaving a note here: once #1654 merges, let's mostly use the implicit refresh in this tests: it makes the code one line shorter and makes sure everything gets tested (as the explicit refresh is tested plenty in `test_updater_top_level_update.py` already)

PR #1654, opened 2021-11-02
ngclient: Implicitly call refresh()

### LINKED ISSUE #1650: ngclient: Make refresh() optional
A core design change in ngclient (compared to legacy) is that the updater only loads metadata in "dependency order": this means reloading e.g. root metadata after timestamp metadata is loaded is just not possible. This means that calling Updater.refresh() multiple times is not possible: if user wants to "load  a new repository snapshot" they should create a new Updater*.

This leads to `refresh()` call becoming rather unnecessary: we can just run that automatically the first time `get_targetinfo()` is called. I don't want to remove `refresh()` from the API completely as it does do multiple http requests and the targets response could be large, so some clients might want to do `refresh()` explicitly. I do want to make calling `refresh()` optional.

This is the last part of #1580.

_*) we could make multiple refreshes work: it would just mean almost the same thing as creating a new Updater. The only thing that would be optimized is not reading root.json from disk_

jku:
Refresh can still be done explicitly (so client has more control over
when data is downloaded) but there's no reason we can't do it
automatically as well: Call `refresh()` from `get_targetinfo()` if
top-level targets is not loaded.

Update API documentation accordingly.

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -10,24 +10,23 @@ Updater provides an API to query available targets and to download them in a
 secure manner: All downloaded files are verified by signed metadata.
 
 High-level description of Updater functionality:
-  * Initializing an :class:`~tuf.ngclient.updater.Updater` loads and validates
-    the trusted local root metadata: This root metadata is used as the source
-    of trust for all other metadata.
-  * Calling :func:`~tuf.ngclient.updater.Updater.refresh()` will update root
-    metadata and load all other top-level metadata as described in the
-    specification, using both locally cached metadata and metadata downloaded
-    from the remote repository.
-  * When metadata is up-to-date, targets can be dowloaded. The repository
-    snapshot is consistent so multiple targets can be downloaded without
-    fear of repository content changing. For each target:
+  * Initializing an ``Updater`` loads and validates the trusted local root
+    metadata: This root metadata is used as the source of trust for all other
+    metadata.
+  * ``refresh()`` can optionally be called to update and load all top-level
+    metadata as described in the specification, using both locally cached
+    metadata and metadata downloaded from the remote repository. If refresh is
+    not done explicitly, it will happen automatically during the first target
+    info lookup.
+  * Updater can be used to download targets. For each target:
 
-      * :func:`~tuf.ngclient.updater.Updater.get_targetinfo()` is
-        used to find information about a specific target. This will load new
-        targets metadata as needed (from local cache or remote repository).
-      * :func:`~tuf.ngclient.updater.Updater.find_cached_target()` can be used
-        to check if a target file is already locally cached.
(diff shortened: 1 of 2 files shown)


jku:
Marking ready for review: #1636 is unlikely to move forward this week so I added a minimal test case that should allow merging if this looks good otherwise, and I will file an issue to modify more tests a bit later (as #1636 is going to add many relevant tests: modifying those will make sense).

tuf/ngclient/updater.py
```diff
@@ -184,6 +181,9 @@ def get_targetinfo(self, target_path: str) -> Optional[TargetFile]:
         Returns:
             A TargetFile instance or None.
         """
+
+        if self._trusted_set.targets is None:
+            self.refresh()
```
jku, line 186: here `get_targetinfo()` decides that not having "targets" means it should run refresh itself -- I think this is appropriate as Updater really has only two states it can be in (assuming it has not raised errors):
* trusted set contains root but nothing else
* trusted set contains all top-level metadata (and potentially other metadata)

tuf/ngclient/updater.py
```diff
@@ -130,11 +126,16 @@ def refresh(self) -> None:
         specified order (root -> timestamp -> snapshot -> targets) implementing
         all the checks required in the TUF client workflow.
 
-        The metadata for delegated roles are not refreshed by this method as
-        that happens on demand during get_targetinfo().
+        A ``refresh()`` can be done only once during the lifetime of an Updater.
```
MVrachev, line 129: It seems that I have forgotten, but why can we call refresh only once during the lifetime of an Updater?

MVrachev (commented):
I like the changes you proposed and they make sense.
I have only one clarifying question which is shown above.

tuf/ngclient/updater.py
jku: the issue (#1650) explains it -- with current design allowing a second refresh would be essentially the same as creating a new updater

jku:
nothing has happened here (apart from Martins review, thanks) and #1636 has now been merged.

I guess it makes sense for me to add the proper tests into this PR now that the testing infrastructure is there...

jku:
> I guess it makes sense for me to add the proper tests into this PR now that the testing infrastructure is there...

I did not think this entirely through: Teodoras merged tests include top-level metadata update tests... but the `get_targetinfo()` tests are still coming, so there are currently no RepositorySimulator-based tests where "implicit refresh()" is a natural fit.

So I change my mind: **this is ready for review** with just one test, let's add more when (or after) the delegated targets testing is added, I think in #1641

sechkova (approved):
Thanks for the change, it comes just on time. I noticed while writing tests that we did not add any user-friendly error message when `get_targetinfo` is called without `refresh` and we end up with the assert error from `assert self._trusted_set.snapshot is not None  # nosec` which is not great.

PR #1658, opened 2021-11-02
New implementation tests: apply black and isort automatic linting fixes

### LINKED ISSUE #1656: New implementation tests: apply black and isort automatic linting fixes

The first step to do that would be to apply `black` and `isort` on the new test files testing the new code.

Synchronize with https://github.com/theupdateframework/python-tuf/issues/1582

MVrachev:
Fixes #1656

Currently, we don't do any listing for our tests.
I believe we should enable linting (tox -e lint) for new test files.

The first step to do that would be to apply the automatic fixes proposed
by `black` and isort on the new test files testing the new code.
The next step would be to propose fixes for `pylint` and `mypy` and the last
the part will be to enable the linters themselves. 

Additionally, I saw one opportunity to simplify a little test_api.py and that's
what the last commit is all about.

Related to issue #1582.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -91,21 +91,17 @@ class TestMetadata(unittest.TestCase):
             # Load JSON-formatted metdata of each supported type from file
             # and from out-of-band read JSON string
             path = os.path.join(self.repo_dir, "metadata", metadata + ".json")
-            metadata_obj = Metadata.from_file(path)
+            md_obj = Metadata.from_file(path)
             with open(path, "rb") as f:
-                metadata_obj2 = Metadata.from_bytes(f.read())
+                md_obj2 = Metadata.from_bytes(f.read())
 
             # Assert that both methods instantiate the right inner class for
             # each metadata type and ...
-            self.assertTrue(isinstance(metadata_obj.signed, inner_metadata_cls))
-            self.assertTrue(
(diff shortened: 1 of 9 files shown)


tests/test_api.py
```diff
@@ -643,87 +613,108 @@ def  test_length_and_hash_validation(self):
             # test exceptions
             expected_length = snapshot_metafile.length
             snapshot_metafile.length = 2345
-            self.assertRaises(exceptions.LengthOrHashMismatchError,
-                snapshot_metafile.verify_length_and_hashes, data)
+            self.assertRaises(
+                exceptions.LengthOrHashMismatchError,
+                snapshot_metafile.verify_length_and_hashes,
+                data,
+            )
```
jku, line 620: It think these are more evidence of how we should ban `self.assertRaises()` unless it's used in a `with self.assertRaises():`.

however, I'm fine with this PR being a mostly "machine-generated" change so that's just a side note.

jku (approved):
Thanks, I support this: it's a massive patch but I'd rather have that pain now and then be able to enforce the coding style later (once we get the linters running in CI).

I've resisted the temptation to make improvement suggestions: this is (mostly) a machine generated patch and makes sense to keep it like that. LGTM.

tests/test_api.py
MVrachev: I created an issue about that #1660.

PR #1659, opened 2021-11-02
pylintc for new code: disable format checker

MVrachev:

By default pylint does format checks:
https://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=format#format-checker

The problem is we also use black and isort who have format checkers as
well. This makes pylint format checks obsolete.

Also, it's possible that you would want to disable a warning and you
can end up in the situation where you will have to disable it for
two tools altogether.

### DIFF
--- a/tuf/api/pylintrc
+++ b/tuf/api/pylintrc
@@ -14,6 +14,7 @@
 disable=fixme,
         too-few-public-methods,
         too-many-arguments,
+        format,
 
 [BASIC]
 good-names=i,j,k,v,e,f,fn,fp,_type
--- a/tuf/api/pylintrc
+++ b/tuf/api/pylintrc
@@ -35,15 +35,6 @@ no-docstring-rgx=(__.*__|main|test.*|.*test|.*Test)$
 variable-rgx=^[a-z][a-z0-9_]*$
 docstring-min-length=10
 
-[FORMAT]
-ignore-long-lines=(?x)(
-  ^\s*(\#\ )?<?https?://\S+>?$|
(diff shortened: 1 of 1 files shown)


MVrachev:
As Jussi has pointed out, even the pylint project had disabled `format` for pylint:
https://github.com/PyCQA/pylint/blob/main/pylintrc#L80 😆.

MVrachev:
I am finding more and more checks for pylint that overlap with what black 
I found one more check that overlaps with black work [inconsistent-quotes](https://pycodequ.al/docs/pylint-messages/w1405-inconsistent-quotes.html)
and another one that overlaps with what isort does [wrong-import-order](https://vald-phoenix.github.io/pylint-errors/plerr/errors/imports/C0411.html)

@jku  do you think we should disable those checks?

jku:
If the checks are problematic, then sure why not. But let's not disable things if they're working fine

jku (approved):
I agree with this:
* black seems to do a good job of format enforcement
* if we need to add local disables (like we'd like to in some test code), having two checkers for the same thing is problematic

MVrachev:
> If the checks are problematic, then sure why not. But let's not disable things if they're working fine

Locally I noticed a couple of warnings from pylint when I wanted to prepare a pr for pylint.
I will wait to have #1658 merged and then propose a pr for pylint.
If those warnings are still visible then I will disable them there.

jku:
Now that I think about it... you could remove the format checker configuration section from pylintrc as well: if it does not get used, it should not be there

jku (approved):

PR #1663, opened 2021-11-04
github: explicitly set workflow permissions

### LINKED ISSUE #1662: github actions: set workflow permissions to read-only

We could set the CI workflow permissions to read-only in the workflow:
* **pull_request runs are already always read-only** so they are not  affected: this only affects the runs on merge commits
* this is not about creating a real security barrier but more of a devops hygiene thing: having write access to everything in the repository and github project is not great, so it would be better if **a commit that needs write access then also needs to visibly change the permissions in the workflow file**.

**Current behavior**:
Default GITHUB_TOKEN Permissions on merge runs:
```
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
```

**Expected behavior**:
These permissions are likely enough for our CI workflow:
```
  Contents: read
  Metadata: read
```

jku:
* current workflow only needs to read git content
* if the workflow in the future does need write access (or even other read access than the git repo), it's good to
  see permissions explicitly changing

For context: "pull_request" runs never have any write access anyway, so this
significantly changes only the "push" runs that happen when branches are
merged to develop.

Fixes #1662

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   tests:
     name: Tests


jku:
For anyone wondering: you can see the permissions by going to the [job logs](https://github.com/theupdateframework/python-tuf/runs/4103353736), opening "Set up job", then opening "GITHUB_TOKEN Permissions"

lukpueh (approved):
Cross-checked with relevant [GH docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions). LGTM. Thanks, Jussi!

Issue #1657, opened 2021-11-02
New implementation test files: resolve linting issues discovered by pylint and mypy


MVrachev:
I believe we should enable linting (tox -e lint) for new test files testing the new implementation.

The first step to do that would be to apply black and isort on the new test files testing the new code
which is documented in #1656.

The next step is to resolve the lining issues discovered by pylint and mypy.

TODO:
- [ ] address warnings by `pylint`
- [ ] address warnings by `mypy`

Synchronize with https://github.com/theupdateframework/python-tuf/issues/1656 and https://github.com/theupdateframework/python-tuf/issues/1582

MVrachev:
I will assign myself as have commits that I can use for pr resolving this issue.

PR #1664, opened 2021-11-05
New implementation tests: address pylint warnings

MVrachev:
Related to #1657:

The first step to do that would be to apply the automatic fixes proposed
by `black` and `isort` on the new test files testing the new code.
This happened in #1658.

The next step would be to propose fixes for `pylint` and `mypy` and the last
the part will be to enable the linters themselves.

In this pr, I include fixes for the warnings by `pylint` only as those changes are **NOT**
automatically by the tool and are more than enough for a pr.

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -73,7 +73,7 @@ from tuf.api.metadata import (
     Timestamp,
 )
 from tuf.api.serialization.json import JSONSerializer
-from tuf.exceptions import FetcherHTTPError, RepositoryError
+from tuf.exceptions import FetcherHTTPError
 from tuf.ngclient.fetcher import FetcherInterface
 
 logger = logging.getLogger(__name__)
@@ -83,13 +83,16 @@ SPEC_VER = ".".join(SPECIFICATION_VERSION)
 
 @dataclass
 class RepositoryTarget:
-    """Contains actual target data and the related target metadata"""
+    """Contains actual target data and the related target metadata."""
 
     data: bytes
     target_file: TargetFile
 
 
 class RepositorySimulator(FetcherInterface):
+    """Simulates a repository that can be used for testing."""
+
+    # pylint: disable=too-many-instance-attributes
     def __init__(self):
         self.md_root: Metadata[Root] = None
         self.md_timestamp: Metadata[Timestamp] = None
@@ -136,6 +139,7 @@ class RepositorySimulator(FetcherInterface):
         return self.md_targets.signed
 
     def all_targets(self) -> Iterator[Tuple[str, Targets]]:
+        """Yield role name and signed portion of targets one by one."""
         yield "targets", self.md_targets.signed
         for role, md in self.md_delegates.items():
             yield role, md.signed
@@ -151,7 +155,7 @@ class RepositorySimulator(FetcherInterface):
         self.signers[role][signer.key_dict["keyid"]] = signer
 
     def _initialize(self):
-        """Setup a minimal valid repository"""
+        """Setup a minimal valid repository."""
 
         targets = Targets(1, SPEC_VER, self.safe_expiry, {}, None)
         self.md_targets = Metadata(targets, OrderedDict())
@@ -176,7 +180,7 @@ class RepositorySimulator(FetcherInterface):
         self.publish_root()
 
     def publish_root(self):
-        """Sign and store a new serialized version of root"""
+        """Sign and store a new serialized version of root."""
         self.md_root.signatures.clear()
         for signer in self.signers["root"].values():
             self.md_root.sign(signer, append=True)
@@ -185,6 +189,9 @@ class RepositorySimulator(FetcherInterface):
         logger.debug("Published root v%d", self.root.version)
 
     def fetch(self, url: str) -> Iterator[bytes]:
+        """Fetches data from the given url and returns an Iterator (or yields
+        bytes).
+        """
         if not self.root.consistent_snapshot:
             raise NotImplementedError("non-consistent snapshot not supported")
         path = parse.urlparse(url).path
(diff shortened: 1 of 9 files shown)


tests/repository_simulator.py
```diff
@@ -136,6 +139,7 @@ def targets(self) -> Targets:
         return self.md_targets.signed
 
     def all_targets(self) -> Iterator[Tuple[str, Targets]]:
+        """Yield role name and signed portion of targets one by one"""
```
kairoaraujo: I am not sure if relevant comment.  
To be consistent with another docstring you had added, finish with a period.
If it makes sense, it would be interesting to apply the other changes that are part of this PR.

tests/repository_simulator.py
```diff
@@ -185,6 +189,9 @@ def publish_root(self):
         logger.debug("Published root v%d", self.root.version)
 
     def fetch(self, url: str) -> Iterator[bytes]:
+        """Fetches data from the given url and
```
kairoaraujo: See below for a suggestion to add more content within 80 columns line. (optional, I guess)

```python
        """Fetches data from the given url and returns an Iterator (or yields
        bytes).
        """
```

tests/repository_simulator.py
```diff
@@ -234,35 +246,35 @@ def _fetch_metadata(
             # return a version previously serialized in publish_root()
             if version is None or version > len(self.signed_roots):
                 raise FetcherHTTPError(f"Unknown root version {version}", 404)
```
kairoaraujo, line 248: Shall we log something before raising the ``FetcherHTTPError``? 
If it makes sense would be good to apply also to lines 263, 265

tests/repository_simulator.py
```diff
@@ -274,6 +286,7 @@ def _compute_hashes_and_length(
         return hashes, len(data)
 
     def update_timestamp(self):
+        """Update timestamp and assign snapshot ver to snapshot_meta ver"""
```
kairoaraujo: Add the snapshot_meta between simple quotes. I would suggest being explicit and using version instead ver.

tests/test_api.py
```diff
@@ -44,6 +44,8 @@
 
 
 class TestMetadata(unittest.TestCase):
+    """Tests for public API of all classes in tuf/api/metadata.py"""
```
kairoaraujo: Add the tuf/api/metadata.py in between single quotes.

tests/test_api.py
```diff
@@ -155,8 +157,8 @@ def test_to_from_bytes(self):
             self.assertEqual(metadata_obj_2.to_bytes(), obj_bytes)
 
     def test_sign_verify(self):
-        root_path = os.path.join(self.repo_dir, "metadata", "root.json")
-        root = Metadata[Root].from_file(root_path).signed
+        path = os.path.join(self.repo_dir, "metadata", "root.json")
+        root = Metadata[Root].from_file(path).signed
```
kairoaraujo: Above you changed from ``dir`` to ``dest_dir``, I see it as more explicit.
At these lines, I see the ``root_path`` as ``path``, any specific reason for this?

tests/test_updater_ng.py
```diff
@@ -17,20 +17,22 @@
...
 from tests import utils
-from tuf import exceptions, ngclient
+from tuf import exceptions, ngclient, unittest_toolbox
 from tuf.api.metadata import Metadata, TargetFile
 
 logger = logging.getLogger(__name__)
 
 
 class TestUpdater(unittest_toolbox.Modified_TestCase):
+    """Test the Updater class from tuf.ngclient.updater.py"""
```
kairoaraujo: I'm not sure that the '.py' is needed. Add between simple quotes (?).

tests/test_api.py
jku: dir was likely changed because it's a builtin, this change is likely to keep lines below 80 chars

tests/repository_simulator.py
jku: I don't think this is needed here: the repository (or the FetcherInterface implementation to be exact) is saying it does not have this metadata, that's it (we also have zero other details to log)

tests/repository_simulator.py
MVrachev: I used "ver" instead of "version" in order to keep it a one-line docstring as this is a short function, but if it's unclear will change it.

tests/test_api.py
MVrachev: `root_path` was changed to `path`, because there was a pylint warning we were using too many variables in a single function and with this change, we can reuse `path`.
About `dir` Jussi is correct.

tests/test_updater_ng.py
MVrachev: Yea, well it's doesn't really matter, I can change it, but I am not sure if there is some benefit in doing so.
I think I have used it in other places as well.

tests/repository_simulator.py
kairoaraujo: It is clear, but IMHO, I would pick up the exact reference in the documentation between a one-line docstring and an explicit reference. 
I still see my comment as optional as it is test documentation.

tests/test_updater_ng.py
kairoaraujo: Indeed. My comment was more about having a single standard for consistency.

tests/repository_simulator.py
MVrachev: I will add `.` at the end of the docstrings.
Do you have a better suggestion for this docstring? 
The function is simple and I am not sure what else can I write here.

tests/repository_simulator.py
MVrachev: I tried splitting it where it made sense logically, but that's a personal taste.
Will do it as you suggested.

MVrachev:
I addressed comments made by @kairoaraujo:
- added `.` at the end of the docstrings
- used one consistent format when mentioning a specific python file
- small docstring improvements

tests/repository_simulator.py
kairoaraujo: I don't have a better suggestion for this docstring. LGTM

[General comment, not sure if the right place]
I think what we are doing in the ``ngclient`` for docstring is the best option. In a sense, it can become good documentation in the end (as part of readthedocs/sphinx).
I understand that this is a test part, but it can help a lot of new contributors. Any thoughts @jku ?

tests/test_api.py
jku: re-using variables is usually a code smell but this looks OK to me. The best solution might be a refactor of the function (to maybe multiple tests) but I appreciate not doing large changes under "address pylint warnings": anyway LGTM

jku (approved):
Thanks Martin, Kairo. LGTM.

jku:
I was a bout to merge ... but first merged another change that created a conflict. Sorry about that, could you resolve please?

MVrachev:
Rebased on top of develop as there were conflicts after the merge of #1636.

MVrachev:
I proposed a pr fixing pylint warning about too many local variables in `test_sign_verify()` by splitting it: #1671.
Maybe we can merge #1671 and then I will update this one without reusing `path`?

MVrachev:
#1671 is merged and I have rebased and removed the change which handled too many local variables (renaming `root_path` to `path`).

tests/test_api.py
MVrachev: This is no longer of a concern after the merge of #1671 and I removed this change.

PR #1666, opened 2021-11-08
RepositorySimulator: add non-consistent snapshot support

### LINKED ISSUE #1573: tests: consistent_snapshot tests in test_updater_ng.py are broken
`test_refresh_on_consistent_targets()` does not actually try to refresh any metadata apart from root and timestamp (and it couldn't because the files don't exist in the repository). The test only passes because client already has all the metadata in the metadata cache.

I'm not sure it makes sense to fix this before #1444 but I'm filing so we don't make this mistake again.

### LINKED ISSUE #1637: RepositorySimulator: implement non-consistent snapshot
Repository simulator currently does not implement non-consistent snapshot mode.

**Current behavior**:
NotImplementedError("non-consistent snapshot not supported")

**Expected behavior**:
Able to serve data with consistent snapshot turned off.

### LINKED ISSUE #1667: ngtests: add tests toggling consistent snapshot
Currently different configurations of `consistent_snapshot`, `prefix_targets_with_hash`  are tested in `test_updater_ng.py` which uses pre-generated test files.

After adding support for non-consistent snapshot #1637, analyze what would be meaningful consistent snapshot tests that preferably don't duplicate the tests code for each config and implement them using `RepositorySimulator`.

sechkova:
Extend filename partitioning to support serving non-versioned
metadata and non-prefixed target files.

Fixes #1637, fixes #1667, fixes #1573

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -111,8 +111,11 @@ class RepositorySimulator(FetcherInterface):
         # target downloads are served from this dict
         self.target_files: Dict[str, RepositoryTarget] = {}
 
-        # Whether to compute hashes and legth for meta in snapshot/timestamp
+        # Whether to compute hashes and length for meta in snapshot/timestamp
         self.compute_metafile_hashes_length = False
+
+        # Enable hash-prefixed target file names
+        self.prefix_targets_with_hash = True
 
         self.dump_dir = None
         self.dump_version = 0
@@ -192,24 +195,32 @@ class RepositorySimulator(FetcherInterface):
         """Fetches data from the given url and returns an Iterator (or yields
         bytes).
         """
-        if not self.root.consistent_snapshot:
-            raise NotImplementedError("non-consistent snapshot not supported")
         path = parse.urlparse(url).path
         if path.startswith("/metadata/") and path.endswith(".json"):
+            # figure out rolename and version
             ver_and_name = path[len("/metadata/") :][: -len(".json")]
-            # only consistent_snapshot supported ATM: timestamp is special case
-            if ver_and_name == "timestamp":
-                version = None
-                role = "timestamp"
-            else:
-                version, _, role = ver_and_name.partition(".")
+            version, _, role = ver_and_name.partition(".")
+            # root is always version-prefixed while timestamp is always NOT
+            if role == "root" or (
+                self.root.consistent_snapshot and ver_and_name != "timestamp"
+            ):
                 version = int(version)
+            else:
+                # the file is not version-prefixed
+                role = ver_and_name
+                version = None
+
             yield self._fetch_metadata(role, version)
         elif path.startswith("/targets/"):
             # figure out target path and hash prefix
             target_path = path[len("/targets/") :]
             dir_parts, sep, prefixed_filename = target_path.rpartition("/")
-            prefix, _, filename = prefixed_filename.partition(".")
+            # extract the hash prefix, if any
+            if self.root.consistent_snapshot and self.prefix_targets_with_hash:
+                prefix, _, filename = prefixed_filename.partition(".")
+            else:
+                filename = prefixed_filename
+                prefix = None
             target_path = f"{dir_parts}{sep}{filename}"
 
             yield self._fetch_target(target_path, prefix)
@@ -257,7 +268,7 @@ class RepositorySimulator(FetcherInterface):
         elif role == "targets":
             md = self.md_targets
         else:
-            md = self.md_delegates[role]
+            md = self.md_delegates.get(role)
 
         if md is None:
             raise FetcherHTTPError(f"Unknown role {role}", 404)
--- a/tests/test_updater_consistent_snapshot.py
+++ b/tests/test_updater_consistent_snapshot.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python
+
+# Copyright 2021, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Test ngclient Updater toggling consistent snapshot"""
+
+import os
+import sys
+import tempfile
+import unittest
+from typing import Any, Dict, Iterable, List, Optional
+from unittest.mock import call, patch
+
+from tests import utils
+from tests.repository_simulator import RepositorySimulator
+from tuf.api.metadata import (
+    SPECIFICATION_VERSION,
+    TOP_LEVEL_ROLE_NAMES,
+    Targets,
+)
+from tuf.ngclient import Updater
+
+
+class TestConsistentSnapshot(unittest.TestCase):
+    """Test different combinations of 'consistent_snapshot' and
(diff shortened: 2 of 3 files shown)


sechkova:
I didn't add tests in this PR, I opened this one instead: #1667

tests/repository_simulator.py
```diff
@@ -185,24 +200,19 @@ def publish_root(self):
...
         path = parse.urlparse(url).path
         if path.startswith("/metadata/") and path.endswith(".json"):
             ver_and_name = path[len("/metadata/") :][: -len(".json")]
-            # only consistent_snapshot supported ATM: timestamp is special case
-            if ver_and_name == "timestamp":
-                version = None
-                role = "timestamp"
-            else:
-                version, _, role = ver_and_name.partition(".")
+            role, version = self._partition_filename(ver_and_name)
```
jku: hmm, it seems to do the right thing, but your function is called "partition filename" and then you need to feed it something that is explicitly just a portion of a filename ...

jku (commented):
Did you somehow test this privately?

The core difficulty with this issue is the corner we painted ourselves in by telling warehouse they can be consistent_snapshot **and** offer non-consistent targets for download at the same time: so we can't 100% tell if a targetname is hash prefixed or not when a client requests it ...

This code seems to assume that _if target name contains a "." then it contains a prefix_. I realize there probably aren't perfect solutions here but this seems non-optimal: it should be possible to request non-prefixed targets with dots in their names

sechkova:
> This code seems to assume that _if target name contains a "." then it contains a prefix_. I realize there probably aren't perfect solutions here but this seems non-optimal: it should be possible to request non-prefixed targets with dots in their names

Oh yes, you are right. I guess I will rework it in parallel with tests to cover all the weird cases.

jku (changes requested):
marking "request changes" so it's visible in issue list

sechkova:
I reworked the implementation and now `RepositorySimulator` makes the decision whether the client requests versioned/prefixed files or not based on: `self.root.consistent_snapshot` and `self.prefix_targets_with_hash`.  While the specification describes how repository and client sync on consistent_snapshot, I presumed that they both need to know in advance whether they will work with hash-prefixed targets and I added the same `prefix_targets_with_hash` option to the repository. 

I also added one new test file for consistent snapshot. I couldn't decide if the class `TestConsistentSnapshot` needs to be in a separate test file or integrated in another. Let me know your opinion on this as well as the general idea/readability of the tests.
Since we want to test the client and not that much the `RepositorySimulator` and it is not possible to verify that internally Updater builds the correct prefix+name, I "wrapped" `RepositorySimulator._fetch_*` to check the calls made by Updater.

tests/repository_simulator.py
```diff
@@ -192,24 +195,32 @@ def fetch(self, url: str) -> Iterator[bytes]:
...
         if path.startswith("/metadata/") and path.endswith(".json"):
+            # figure out rolename and version
             ver_and_name = path[len("/metadata/") :][: -len(".json")]
-            # only consistent_snapshot supported ATM: timestamp is special case
-            if ver_and_name == "timestamp":
-                version = None
-                role = "timestamp"
-            else:
+            # root is always version-prefixed while timestamp is always NOT
+            if ver_and_name.endswith("root") or (
```
sechkova: This will fail if a delegated role name ends with "root" (and consistent_snapshot is not enabled). But otherwise it is hard to generalize and I guess we don't want to support that weird cases in RepositorySimulator?

jku: yeah no need to spend a lot of time polishing a test utility to perfection... 

that said you could move `version, _, role = ver_and_name.partition(".")` to before this line so you could compare `role == "root"` (but would keep rest of the code as is -- I admit the else clause is unintuitive if this is done)?

up to you

tests/test_updater_consistent_snapshot.py
```diff
@@ -0,0 +1,236 @@
...
+        prefix_targets_with_hash: bool = test_case_data["prefix_targets"]
+        hash_algo: Optional[str] = test_case_data["hash_algo"]
+        targetpaths = ["file", "file.txt", "..file.ext", "f.le"]
+
+        sim = self._init_repo(consistent_snapshot, prefix_targets_with_hash)
+        # Add targets to repository
+        for targetpath in targetpaths:
+            sim.targets.version += 1
+            sim.add_target("targets", b"content", targetpath)
+            sim.update_snapshot()
```
jku: It's not important but there should be no need to update snapshot more than once

tests/test_updater_consistent_snapshot.py
```diff
@@ -0,0 +1,236 @@
+#!/usr/bin/env python
+
+# Copyright 2021, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Test ngclient Updater consistent snapshot"""
```
jku: this could be better (we're testing consistent snapshot in all client tests, hopefully). This file seems to test 
* that the client works with all combinations of consistent_snapshot and prefix_targets_with_hash
* that expected URLs are being fetched with all of these combinations

jku (commented):
I think your reasoning about prefix_targets_with_hash option makes sense.

I also think counting the fetches at repository side is fine and actually better than poking into updater internals. 

The patching in the tests is not pretty but it is much improved in this version. I think the asserts you use do not check whether _other_ fetches were also made though, which I think we want (so we know it doesn't download things twice or something)... I think we should be comparing our expected calls list directly to the actually made calls list. WDYT?

...which leads me to: Making the call counters a RepositorySimulator feature might make the tests more readable and probably wouldn't be a lot of code: `sim.call_stats.metadata` could contain basically the same list of tuples that your testdata contains -- but this could also be a followup idea if you'd rather get these tests in ASAP.

tests/repository_simulator.py
sechkova: I tried to do it in a readable way ... at least now it is correct for the cases I can think of.

sechkova:
> ...which leads me to: Making the call counters a RepositorySimulator feature might make the tests more readable and probably wouldn't be a lot of code: sim.call_stats.metadata could contain basically the same list of tuples that your testdata contains -- but this could also be a followup idea if you'd rather get these tests in ASAP.

I think I will open a separate issue. It may double some of the work but it would be easier at least for me to do it as a follow up.
There it is: #1682

sechkova:
Just noting that I'd like to squash the last two commits before merging :)

jku (approved):
:+1:

sechkova:
Squashed the last two commits, should be ready now.

sechkova:
I now saw this issue:  #1573
I think `test_refresh_on_consistent_targets()` can be replaced by the new test file (`test_updater_consistent_snapshot.py`) and we can close it. I can do it in this PR if you agree?

jku:
Sounds good to me!

jku:
Thanks

Issue #1558, opened 2021-09-01
Metadata API: prevent Delegation role names to be one of top level metadata roles


MVrachev:
**Description of issue or feature request**:

**Current behavior**:
Delegation role names are not restricted in any way in the spec, but they are targets metadata role names.
That could lead to a problem if delegation role names are one of `root`, `timestamp`, `snapshot` or `targets`.

**Expected behavior**:
Make sure that delegation role names aren't one of the top-level metadata roles.

joshuagl:
I wonder if this kind of logic and role name sanitisation more generally belongs in metadata API? Perhaps it should be part of the (as yet undesigned) repository interface?

jku:
same check does make sense in a client: we would not want to accidentally overwrite our local root.json because a targets decides to delegate to a "root" role version 9000.

I don't think there are practical risks of that -- the client would have to succeed in downloading and verifying the file _as a targets metadata_ before it would be written to local disk but this still seems like a reasonable check for a client to make

jku:
the legacy client also prevents empty string as rolename -- this seems like  a reasonable take for ngclient too

joshuagl:
@MVrachev is this issue covered by https://github.com/theupdateframework/python-tuf/pull/1630?

MVrachev:
No, as the validation itself happens in the delegator class who have the full list of roles or in the case of #1630 that's `Root`.
Here the validation should happen in `Targets` the same way it's done in #1630.

MVrachev:
When solving this issue please add tests in `test_metadata_serialization.py` the same way it's done in #1630 specifically see `incorect_roots`.

kairoaraujo:
I am working on this issue.

Issue #1158, opened 2020-10-02
client/updater design: parallel downloads


jku:
Updater should _in some way_ support parallel downloads.

This came up because I found an obscure corner in pip that does index file downloads in parallel in pip (`pip list --outdated`): if I plug my TUF-support in there, TUF Updater gets called from multiple threads and very bad things happen.

Some notes on this:
* for the pip use case, letting pip handle the actual downloads would remove 50% of the need for this  -- but bin metadata downloads would still be sequential (and `Updater.refresh()` would still be thread-unsafe)
* Still, parallel downloads (for both targets and bin metadata) is probably a wanted feature in other user-facing apps as well: it is faster and is something package managers already often do
* this feature does not mean that TUF API absolutely needs to be threadsafe: many use cases would be covered by small API change: `Updater.get_valid_targetinfos(list)`  and `Updater.download_targets(list)` could still be a single-threaded API but they could be implemented in a way that uses multiple threads and multiple connections to the server. I have no idea how easy this would be though.
* The nicest solution might be a low-level [Sans-IO](https://sans-io.readthedocs.io/) style API (a state machine with events) or at least something that is asynchronous to the users. This  API is then wrapped by a simple I/O layer that could look much like it now does. This would mean a large change.

jku:
Issue #1213 _updater: abstract out the network IO_ might provide a reasonable path to parallelized target downloads: Updater would not need multi-threaded code or to even be thread-safe: custom Fetchers could implement the parallelization and the Updater API just needs to handle a list of targets to update instead of a single target.

PR #1668, opened 2021-11-09
Metadata API: make Root roles a Mapping

### LINKED ISSUE #1356: New metadata API: Validate dictionary keys as well as values
By resolving issue #1139 we are going to add classes for all complex fields inside the top-level metadata classes e.g. `meta`/ `targets` (in Timestamp, Snapshot, Targets), `delegations` (in Targets), `keys`/`roles` (in not yet existent Delegation).

The problem is, that many of those complex fields are represented by a dictionary and we are adding classes for the values of
the dictionary in order to easily validate them later, but we should make sure to validate the dictionary keys as well.

**Current behavior**:
We are not validating dictionary keys yet.

**Expected behavior**:
We should make sure to validate:

- [ ] [KEYID](https://theupdateframework.github.io/specification/latest/#root-keyid) in Root and Delegations
- [ ] [ROLE](https://theupdateframework.github.io/specification/latest/#root-role) in Root | #1516
- [ ] [METAPATH ?](https://theupdateframework.github.io/specification/latest/#snapshot-metapath) in Snapshot and Timestamp
- [ ] [TARGETPATH ?](https://theupdateframework.github.io/specification/latest/#targetpath) in Targets

MVrachev:
Fixes #1356 

This pr includes a couple of smaller changes:
- make root roles `Mapping`  to indicate that users should not add or remove
values from the dictionary during the lifetime of the Root object

**_EDIT: Those two were included in the initial version of the pr._**
**Then together with @jku we agreed we don't want to do that.
See my comment https://github.com/theupdateframework/python-tuf/pull/1668#issuecomment-971755506**
- validate that TARGETPATHs are not empty strings
- validate that METAPATHs are not empty strings or any of the top-level metadata roles

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -722,7 +722,7 @@ class Root(Signed):
         spec_version: str,
         expires: datetime,
         keys: Dict[str, Key],
-        roles: Dict[str, Role],
+        roles: Mapping[str, Role],
         consistent_snapshot: Optional[bool] = None,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):


MVrachev:
@jku your suggestion:
> accept roles=None in the constructor (and create the default set in that case). This is not strictly validation but lowers the need for validation since then no-one ever needs to manually create the roles dict.

I tried implementing it, but then I realized that If I am given `keys` dictionary containing 4 keys I am not sure how to map keys to roles.

MVrachev:
METPATHs and TARGETPATHs are not validated that are strings as per specification they can't be anything else but a string.

I am still wondering if we should create a sanity check that you can't use `targets` as a TARGETPATH...

tuf/api/metadata.py
```diff
@@ -1026,6 +1026,10 @@ def __init__(
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
         super().__init__(version, spec_version, expires, unrecognized_fields)
+        if any(meta_path == "" for meta_path in meta):
+            raise ValueError("All meta paths must be non-empty strings")
+        if any(meta_path in TOP_LEVEL_ROLE_NAMES for meta_path in meta):
+            raise ValueError("Meta paths must not be top-level metadata roles")
```
jku: this came up in the issue discussion already: meta keys are filenames not rolenames (not necessarily real filenames, but imaginary `{rolename}.json` filenames)

tuf/api/metadata.py
```diff
@@ -1026,6 +1026,10 @@ def __init__(
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
         super().__init__(version, spec_version, expires, unrecognized_fields)
+        if any(meta_path == "" for meta_path in meta):
+            raise ValueError("All meta paths must be non-empty strings")
```
jku: you originally suggested checking that "meta_path is a non-empty string": now you check that "metapath is not an empty string". These are subtly different as the former includes type checking and the implemented check is so specific that I don't think it's worth two lines of code.

I'm still not sure how much type checking we actually should do... maybe we should validate the dictionary key types as that doesn't happen "implicitly" like it tends to happen for other data that is wrongly typed. If we don't do that we could deserialize (and then later serialize) data that is not actually  valid metadata

jku:
> @jku your suggestion:
> 
> > accept roles=None in the constructor (and create the default set in that case). This is not strictly validation but lowers the need for validation since then no-one ever needs to manually create the roles dict.
> 
> I tried implementing it, but then I realized that If I am given `keys` dictionary containing 4 keys I am not sure how to map keys to roles.

sorry, I don't understand the connection here. Why do you need to map keys to roles in a constructor?

jku (commented):
> METPATHs and TARGETPATHs are not validated that are strings as per specification they can't be anything else but a string.

We're trying to validate that data we got is according to spec, and should not assume that things are strings because spec says so... that said, I too am unsure about type checks: I've been negative about them but am beginning to think we might want to do type checks for dictionary keys as those never get checked by other code (as most other types do). More on that in code comment.


> I am still wondering if we should create a sanity check that you can't use `targets` as a TARGETPATH...

Sorry, I don't understand this.

MVrachev:
> We're trying to validate that data we got is according to spec, and should not assume that things are strings because spec says so... that said, I too am unsure about type check

> > METPATHs and TARGETPATHs are not validated that are strings as per specification they can't be anything else but a string.
> 
> We're trying to validate that data we got is according to spec, and should not assume that things are strings because spec says so... 

I was unclear with my statement. I meant the JSON format spec which doesn't allow non-string keys for dictionaries
and that's why I am not sure if there is a point in validating that TARGETPATHs and METAPATHs are strings.

tuf/api/metadata.py
MVrachev: You are right.
This begs the question do we want to have a `TOP_LEVEL_FILE_NAMES` set as well?

jku:
> I was unclear with my statement. I meant the JSON format spec which doesn't allow non-string keys for dictionaries and that's why I am not sure if there is a point in validating that TARGETPATHs and METAPATHs are strings.

Oh this is a really good point! we're supposedly format-independent so maybe shouldn't trust this ... but it's still a convincing argument

tuf/api/metadata.py
jku: Maybe? It is going to make our "format-independence" even more clearly unrealistic... but we already have ".json" embedded in metadata.py in several places so I don't think this is any worse.

Just or reference, actual format-independence would mean
* probably fixing the spec: meta keys should not be filenames, that makes no sense
* file storage should be tied to serialization (so that when a filename is needed for e.g. saving a file, the serialization implementation gets to decide the final filename: `Serializer.get_filename(rolename: str)` or something)

MVrachev:
I think you are correct that if we want to be able to easily switch to a new format it's better if we do the type checks.
@jku should I go forward and add a commit about that?

jku:
> I think you are correct that if we want to be able to easily switch to a new format it's better if we do the type checks. @jku should I go forward and add a commit about that?

Supporting a new format is not going to be easy whatever we do (with the way the spec is written)... so I'd rather avoid investing any lines of code in that right now.

Maybe just fix the bug with meta keys and make sure the check that prevents using toplevel filenames (except targets) is as simple as possible: Could build a private list/set of _invalid_snapshot_meta_keys at metadata.py import time (maybe a snapshot private class variable) and then use that list in snapshot init, for example.

Speaking of simplicity I've been thinking about this and I think checking empty string for meta keys (or even checking whether the string ends in ".json") might not be that useful? It does not guarantee the key is a real filename or even that it's a safe filename... In any case meta should only be used like this `metainfo = meta[f"{role}.json"]` in a client so things like empty string keys are not a problem. IMO checking for empty string meta key is not useful.

For target paths the empty string check could make sense, especially if this is mentioned in the spec... I can't find that reference though, is this really in the spec?

MVrachev:
> Maybe just fix the bug with meta keys 

I haven't spent time on this issue for a while, but which bug do you mean?
I don't find where we discussed a bug?
Do you mean this one https://github.com/theupdateframework/python-tuf/pull/1668#discussion_r745680003

> and make sure the check that prevents using toplevel filenames (except targets) is as simple as possible: Could build a private list/set of _invalid_snapshot_meta_keys at metadata.py import time (maybe a snapshot private class variable) and then use that list in snapshot init, for example.

Will try it, good idea.

> Speaking of simplicity I've been thinking about this and I think checking empty string for meta keys (or even checking whether the string ends in ".json") might not be that useful? It does not guarantee the key is a real filename or even that it's a safe filename... In any case meta should only be used like this `metainfo = meta[f"{role}.json"]` in a client so things like empty string keys are not a problem. IMO checking for empty string meta key is not useful.

I agree. At some point, we would want to open what's inside meta keys and if there is an empty string there will be an OSError.

> For target paths the empty string check could make sense, especially if this is mentioned in the spec... I can't find that reference though, is this really in the spec?

I don't find it explicitly mentioned either.
Why do you think it will make sense to check for an empty string compared to the case with meta?

jku:
> I don't find where we discussed a bug?

Current PR is, I believe, preventing meta keys from being toplevel rolenames. That is not relevant: meta keys being toplevel role _filenames_ could be a problem.

That said, now that I look at this and #1558 together I'm wondering if this PR is worth it overall: meta keys are never meant to be iterated over, client always looks up data from meta with a key constructed with `f"{role}.json"` where the role comes from DelegatedRole. Maybe we should concentrate on validating DelegatedRole name: that seems to have most value as it will in practice guarantee valid meta keys as well.

I think I'm fine with closing this (or just doing the Dict->Mapping change) in favor of #1158. What do you think?

MVrachev:
After a discussion with @jku we decided that:
- we don't want to do validation for `METAPATH`s. We know that they are strings (JSON keys specification) and that's enough. That is the case because snapshot stores information about targets and delegated targets in its meta. The METAPATHs themselves are not used to open a file anywhere but are used to uniquely distinguish the different target and delegated target files. What actually is used is version, length, and hashes from snapshot meta.

- DelegationRole names are important to be validated as a client only processes target files by:
1. going through the delegated roles, finding a role it needs
2. looking up version etc for _that_ role in meta
3. loading the file for _that_ version and role

meaning DelegatedRole names are actually used to open files

- TARGETPATHs are strings (JSON keys specification) and it seems it works to use `""` as a valid TARGETPATH. The client will end up constructing a download URL that is basically just `updater._target_base_url` : weird but we don't see a real problem with that either. Basically, we agreed that we shouldn't absolutely need to special-case handling for this weird case out of many possible weird cases

Because of those reasons, I will drop the other commits which validate TARGETPATH and METAPATH and will leave only the first one.

jku (approved):
LGTM, thanks for going through the steps so we could figure out what we actually wanted here

PR #1672, opened 2021-11-12
Create constants for top-level rolenames

### LINKED ISSUE #1648: metadata API: Constants for top-level rolenames

Top-level role names - "root", "timestamp", "snapshot" and "targets" - are hardcoded throughout the code base. Good coding practice suggests to create constants for these and use them instead.

See
```bash
grep -rE  [\"\']\(root\|targets\|snapshot\|timestamp\)[\"\'] tuf/api/ tuf/ngclient/
```

**Current behavior**:
- Top-level role names are hardcoded

**Expected behavior**:
- Constants for top-level role names are defined in `tuf/api/metadata.py` and
- Constants are used in TUF 1.0-bound code, that is at least in `tuf/api/*` and `tuf/ngclient/*` and probably any new test modules.

ivanayov:
Fixes #1648

This is a change in the metadata API to remove hardcoded role names
and use constants instead.

- [x] Tests have been added for the bug fix or new feature - updated only
- [ ] Docs have been added for the bug fix or new feature - n/a

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -139,7 +139,7 @@ class RepositorySimulator(FetcherInterface):
 
     def all_targets(self) -> Iterator[Tuple[str, Targets]]:
         """Yield role name and signed portion of targets one by one."""
-        yield "targets", self.md_targets.signed
+        yield Targets.type, self.md_targets.signed
         for role, md in self.md_delegates.items():
             yield role, md.signed
 
@@ -181,7 +181,7 @@ class RepositorySimulator(FetcherInterface):
     def publish_root(self) -> None:
         """Sign and store a new serialized version of root."""
         self.md_root.signatures.clear()
-        for signer in self.signers["root"].values():
+        for signer in self.signers[Root.type].values():
             self.md_root.sign(signer, append=True)
 
         self.signed_roots.append(self.md_root.to_bytes(JSONSerializer()))
@@ -197,8 +197,8 @@ class RepositorySimulator(FetcherInterface):
             ver_and_name = path[len("/metadata/") :][: -len(".json")]
             version_str, _, role = ver_and_name.partition(".")
             # root is always version-prefixed while timestamp is always NOT
-            if role == "root" or (
-                self.root.consistent_snapshot and ver_and_name != "timestamp"
+            if role == Root.type or (
+                self.root.consistent_snapshot and ver_and_name != Timestamp.type
             ):
                 version: Optional[int] = int(version_str)
             else:
@@ -248,7 +248,7 @@ class RepositorySimulator(FetcherInterface):
 
         If version is None, non-versioned metadata is being requested.
         """
-        if role == "root":
+        if role == Root.type:
             # return a version previously serialized in publish_root()
             if version is None or version > len(self.signed_roots):
                 raise FetcherHTTPError(f"Unknown root version {version}", 404)
@@ -257,11 +257,11 @@ class RepositorySimulator(FetcherInterface):
 
         # sign and serialize the requested metadata
         md: Optional[Metadata]
-        if role == "timestamp":
+        if role == Timestamp.type:
             md = self.md_timestamp
-        elif role == "snapshot":
+        elif role == Snapshot.type:
             md = self.md_snapshot
-        elif role == "targets":
+        elif role == Targets.type:
             md = self.md_targets
         else:
             md = self.md_delegates.get(role)
@@ -297,7 +297,7 @@ class RepositorySimulator(FetcherInterface):
         self.timestamp.snapshot_meta.version = self.snapshot.version
 
         if self.compute_metafile_hashes_length:
-            hashes, length = self._compute_hashes_and_length("snapshot")
+            hashes, length = self._compute_hashes_and_length(Snapshot.type)
             self.timestamp.snapshot_meta.hashes = hashes
             self.timestamp.snapshot_meta.length = length
 
@@ -320,7 +320,7 @@ class RepositorySimulator(FetcherInterface):
 
     def add_target(self, role: str, data: bytes, path: str) -> None:
         """Create a target from data and add it to the target_files."""
-        if role == "targets":
+        if role == Targets.type:
             targets = self.targets
         else:
             targets = self.md_delegates[role].signed
@@ -339,7 +339,7 @@ class RepositorySimulator(FetcherInterface):
         hash_prefixes: Optional[List[str]],
     ) -> None:
         """Add delegated target role to the repository."""
-        if delegator_name == "targets":
+        if delegator_name == Targets.type:
             delegator = self.targets
         else:
             delegator = self.md_delegates[delegator_name].signed
@@ -375,9 +375,9 @@ class RepositorySimulator(FetcherInterface):
 
         for ver in range(1, len(self.signed_roots) + 1):
             with open(os.path.join(dest_dir, f"{ver}.root.json"), "wb") as f:
-                f.write(self._fetch_metadata("root", ver))
+                f.write(self._fetch_metadata(Root.type, ver))
 
-        for role in ["timestamp", "snapshot", "targets"]:
+        for role in [Timestamp.type, Snapshot.type, Targets.type]:
             with open(os.path.join(dest_dir, f"{role}.json"), "wb") as f:
                 f.write(self._fetch_metadata(role))
 
--- a/tests/test_trusted_metadata_set.py
+++ b/tests/test_trusted_metadata_set.py
@@ -58,10 +58,10 @@ class TestTrustedMetadataSet(unittest.TestCase):
         )
         cls.metadata = {}
         for md in [
-            "root",
-            "timestamp",
-            "snapshot",
-            "targets",
+            Root.type,
+            Timestamp.type,
+            Snapshot.type,
+            Targets.type,
             "role1",
             "role2",
         ]:
@@ -71,10 +71,10 @@ class TestTrustedMetadataSet(unittest.TestCase):
         keystore_dir = os.path.join(os.getcwd(), "repository_data", "keystore")
         cls.keystore = {}
         root_key_dict = import_rsa_privatekey_from_file(
-            os.path.join(keystore_dir, "root" + "_key"), password="password"
+            os.path.join(keystore_dir, Root.type + "_key"), password="password"
         )
-        cls.keystore["root"] = SSlibSigner(root_key_dict)
-        for role in ["delegation", "snapshot", "targets", "timestamp"]:
+        cls.keystore[Root.type] = SSlibSigner(root_key_dict)
+        for role in ["delegation", Snapshot.type, Targets.type, Timestamp.type]:
             key_dict = import_ed25519_privatekey_from_file(
                 os.path.join(keystore_dir, role + "_key"), password="password"
             )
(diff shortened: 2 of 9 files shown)


tuf/api/metadata.py
```diff
@@ -61,14 +61,20 @@
     SignedSerializer,
 )
 
+class Rolename:
+    ROOT = "root"
+    TIMESTAMP = "timestamp"
+    SNAPSHOT = "snapshot"
+    TARGETS = "targets"
```
MVrachev: Generally, I would prefer if we use 4 separate constants like that:
```
ROOT = "root"
TIMESTAMP = "timestamp"
SNAPSHOT = "snapshot"
TARGETS = "targets"
```
instead of creating a separate class containing the constants.
The reason is that using directly the constants is significantly shorter also we don't expect to change the number of role names anytime soon and I am not sure what advantage do we get by using a class and then the constants nested inside.

MVrachev (changes requested):
Thanks for your contribution @ivanayov!
I have one preference about the way you defined your constants.
Still, I would want to hear an opinion from someone from the maintainers.
What do you think @sechkova?
Do you agree with my comment?

tuf/api/metadata.py
ivanayov: Thanks for the review @MVrachev !

There are 2 reasons I've put those in a class:
1. If a new role is being added in future
2. All "root", "timestamp", "snapshot", "targets" are role names in TUF's context, but excluding the specific use they are very general names and being explicitly referenced with a role name creates better readability.

On the other side I can agree that `Rolename` is used for nothing else than a constant "reference" and this is a bit overkill in terms of resources. 
So if 1. or 2. are not considered important, I can remove the class.

sechkova:
I agree with both of you ... 
On one hand, Root, Targets etc. classes already exist in metadata.py and now if we add similar constants for role names, we'll be importing things like 
```
from tuf.api.metadata import Targets, TARGETS
```
On the other hand, names do grow in size.

I was thinking if it there is a convenient data structure than would:
- encapsulate rolenames
- act as iterator allowing us to call list(Rolenames); for role in Rolenames: and by doing this replaces TOP_LEVEL_ROLE_NAMES too
- have a short enough name :shrug: 

But this is only a suggestion and if implementing this feels like an overkill for python-tuf, I'm ok with simple constants.

MVrachev:
@lukpueh do you have an opinion on the matter?
I am asking you because we had discussed such issues before with you.

lukpueh:
FWIW, [in warehouse they use an enum subclass](https://github.com/pypa/warehouse/blob/0524db370d1c2b5fd2d16433e0410d77a0b0c362/warehouse/tuf/constants.py#L16-L23 ) to do this, maybe that's the data structure @sechkova was thinking of, it seems to at least allow iterating etc.? :)

Another way of encapsulation would be to make them class attributes of the corresponding metadata classes. 

I'm really fine either way.

ivanayov:
I think enum covers all the cases, as @lukpueh also mentioned.

ivanayov:
I pushed an update with enum.

tests/repository_simulator.py
```diff
@@ -368,9 +369,9 @@ def write(self):
 
         for ver in range(1, len(self.signed_roots) + 1):
             with open(os.path.join(dest_dir, f"{ver}.root.json"), "wb") as f:
-                f.write(self._fetch_metadata("root", ver))
+                f.write(self._fetch_metadata(RoleNames.ROOT.value, ver))
 
-        for role in ["timestamp", "snapshot", "targets"]:
+        for role in [RoleNames.TIMESTAMP.value, RoleNames.SNAPSHOT.value, RoleNames.TARGETS.value]:
```
MVrachev: You can make it shorter:
```suggestion
        for role in RoleNames:
```

tuf/api/metadata.py
```diff
@@ -61,14 +62,25 @@
...
+
+class RoleNames(Enum):
+    ROOT: str = "root"
+    SNAPSHOT: str = "snapshot"
+    TARGETS: str = "targets"
+    TIMESTAMP: str = "timestamp"
+
+    @staticmethod
+    def all() -> set:
+        return set(map(lambda r: r.value, RoleNames))
```
```suggestion
class RoleNames(str, Enum):
    ROOT: str = "root"
    SNAPSHOT: str = "snapshot"
    TARGETS: str = "targets"
    TIMESTAMP: str = "timestamp"

    @staticmethod
    def all() -> set:
        return set(map(lambda r: r.value, RoleNames))
```

If you inherit the string class as well you will be able to do comparisons like:
```
if RoleNames.TARGETS == role
```
instead of doing:
```
if RoleNames.TARGETS.value == role
```
Other such examples from python docs:
https://docs.python.org/3/library/enum.html#others

MVrachev (changes requested):
Honestly, I am still uncertain about the usage of an enum here compared to using separate constants, and given that they will be uppercase I don't think they will be mistaken with something else.

Still, if we decide to use enums I have two suggestions:
1. when defining RoleNames make sure you inherit `str`, this will shorten the comparisons as I explained in my comment 
2. when iterating all of the roles make sure to do: `for role in RoleNames` instead of mentioning each of the roles explicitly.
I left a comment about that too.

**_EDIT_**: I think you can directly just inherit `StrEnum` with the same success as inheriting `Enum` and `str` separatly. 
For more info: https://docs.python.org/3/library/enum.html#others

tuf/api/metadata.py
MVrachev: Additionally, when do we use `RoleNames.all()`? Why do we need it?

ivanayov: It refers to @sechkova's comment for removing `TOP_LEVEL_ROLE_NAMES`. Enums in python don't have any not-breaking-the-rules option to access all their values in a list/set/etc.

ivanayov: I'm fine to go with simple constants if that's the agreement, even though in perspective it doesn't feel like the best option. For example, one would need to always use 
```
from tuf.api.metadata import  Root, ROOT, Targets, TARGETS, Snapshot, SNAPSHOT, Timestamp, TIMESTAMP
```
instead of
```
from tuf.api.metadata import  Root, Targets, Snapshot, Timestamp, RoleNames
```

and if for some reason in future let's say timestamp for example is used in another context as well within TUF, using only TIMESTAMP constant might be confusing.

I changed to enum to address @sechkova's proposal, but if the agreement is to use constants, then we can keep `TOP_LEVEL_ROLE_NAMES `.

jku (commented):
I don't have opinions on the actual data structure but I have a couple on the usage:
* a few legacy test files are modified: The split between legacy and metadata API is confusing so not your fault, but let's not start mixing the two implementations. Basically if the file did not import `tuf.api.metadata` already, it's probably legacy and we don't want to modify the file. so test_repository_tool, test_sig, test_updater should not be touched
* It's great that we try to avoid typos by avoiding hard-coded strings... but it should not come at at the expense of readability: I think going from `"root"` to `RoleNames.ROOT.value` is not a clear cut win as the latter is more than 3 times as long (!) and looks so complex -- I could not guess from the name that I'm looking at a simple role name.

tuf/api/metadata.py
jku: > For example, one would need to always use
> 
> from tuf.api.metadata import  Root, ROOT, Targets, TARGETS, Snapshot, SNAPSHOT, Timestamp, TIMESTAMP

This is why I've never done this*... it seems very hard to provide this as API that is not a repetitive mess but still offers the role names in a way that allows elegant code.

 *) I'm very happy someone is trying though: there could be a neat solution we haven't seen yet

tuf/api/metadata.py
jku: > 1. If a new role is being added in future

If a new role is added, python-tuf likely cannot cope with that without major code changes. So I don't think this structure needs to anticipate that either

jku:
It seems like the major reason to use an enum or a class is to avoid top level variable name pollution (so to avoid `from tuf.api.metadata import Root, ROOT, Timestamp, TIMESTAMP, ....`), correct?

What if instead of a new class we used the classes we are already importing? Root and others already have a class variable `_signed_type` which is already exposed as an instance property `type`... Maybe we could just simplify that so the class variable is called `type` (and is public): then we could just refer to `Root.type` to get the string without instantiating a Root?

(for reference: the class variable _signed_type is probably currently private to make it very clear it should not be modified, but as python does not have "class properties" that means we can't keep using a property in this case(?). I think solving the problem described in this issue is more useful than keeping the type private)

lukpueh:
> What if instead of a new class we used the classes we are already importing? Root and others already have a class variable `_signed_type` which is already exposed as an instance property `type`... Maybe we could just simplify that so the class variable is called `type` (and is public): then we could just refer to `Root.type` to get the string without instantiating a Root?

This is what I meant above with ...

> Another way of encapsulation would be to make them class attributes of the corresponding metadata classes.

One downside would be that we'd have to define a global `TOP_LEVEL_ROLE_NAMES` that uses these class variables after the class definitions, i.e. at the bottom of the metadata module. But I guess that's okay.

jku:
> One downside would be that we'd have to define a global `TOP_LEVEL_ROLE_NAMES` that uses these class variables after the class definitions, i.e. at the bottom of the metadata module.

I'm also not going to lose sleep over a single instance of duplicating the strings within the API implementation if that makes the API definition look neater: what's important is removing the magic strings from the code that imports `tuf.api.metadata`

ivanayov:
I've put the constants within the role classes. Agree that it's more clear, simple and convenient.

As those are both required private (we don't want to be able to change them, whatever "be able" means in python for private properties :D ), but at the same time we need them for static reference within Updater, and we don't want to break the file format which uses `_type`, and don't want to do a breaking API change for such a simple reason with changing `type` to `TYPE`, we came to and agreement with @jku that changing `_signed_type` to `type` and keeping the `_type` property would be the best possible option.
Yep, it's still not very nice to reference the public variable from the private property, but that's the least painful and best to use change, unless someone comes with a genius idea that we cannot see yet.

tests/test_trusted_metadata_set.py
```diff
@@ -15,6 +15,7 @@
 from tests import utils
 from tuf import exceptions
 from tuf.api.metadata import (
+    T,
```
jku: is this needed?

tests/test_updater_top_level_update.py
```diff
@@ -35,7 +35,7 @@ class TestRefresh(unittest.TestCase):
     def setUp(self) -> None:
         self.temp_dir = tempfile.TemporaryDirectory()
         self.metadata_dir = os.path.join(self.temp_dir.name, "metadata")
-        self.targets_dir = os.path.join(self.temp_dir.name, "targets")
+        self.targets_dir = os.path.join(self.temp_dir.name, Targets.type)
```
jku: This is an unrelated "targets" and should not use Targets.type -- in fact it could be renamed "downloads" to be clear about it (it's the directory that client downloads the actual target files to)

tuf/api/metadata.py
```diff
@@ -61,14 +61,19 @@
     SignedSerializer,
 )
 
+_ROOT: str = "root"
+_SNAPSHOT: str = "snapshot"
+_TARGETS: str = "targets"
+_TIMESTAMP: str = "timestamp"
```
jku: I can see why you added these, re-ordering the code wouldn't be enough to avoid them...

I might have just hard coded the strings in the three places inside this file (to make the code in those places easier to read) but I guess this works, it's just one more indirection.

That said, these variables should not need type annotations, right? I'd like to not include annotations unless they are necessary

jku (approved):
LGTM, left a couple of comments but I think none were blockers: please take a look and adapt or don't (leave a comment if you don't), either way works

jku:
> it's still not very nice to reference the public variable from the private property, but that's the least painful and best to use change, unless someone comes with a genius idea that we cannot see yet.

I think the genius idea is to just remove `_type` property: it's from the time when the API matching the file format was considered a bit more important than it is now. IMO the `_type` property is not worth the confusion it creates . But we can handle that in later PRs.

lukpueh:
> just remove _type property

Agreed!

>  it's from the time when the API matching the file format was considered a bit more important than it is now

I  actually suspect that in this case the (Python) implementation influenced the naming decision in the spec, to avoid shadowing the `type` global on json-to-dict deserialisation.

Regardless of this specific case, I still think that it very important that the spec file format is recognisable in the **reference** implementation.

ivanayov:
Thanks @jku. Just addressed your comments.

PR #1674, opened 2021-11-12
Tests: address new pylint warnings

MVrachev:
After the recent changes, there are a couple of new pylint warnings that
appeared.
They are caused by the new test file that was added
test_updater_top_level_update.py and the limit of public functions was
reached in the TestMetadata class in test_api.py
The warnings should be addressed before enabling all of the linters
on the tests files.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -43,6 +43,7 @@ from tuf.api.serialization.json import CanonicalJSONSerializer, JSONSerializer
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


jku:
It's not a problem to add _a few_ fixes like this in the lint enabling PR: that's sort of expected since the code evolves all the time...

jku (approved):
these look fine to me (but feel free to include in lint enabling PR as well).

If we keep adding too many pylint disables, we might want to actually define a separate pylint profile for tests... but for now this seems fine

PR #1675, opened 2021-11-15
TUF Python Client Example/Tutorial

kairoaraujo:
It is a simple example of TUF ngclient implementation.

This example contains a README.rst that is a tutorial/how-to-use
this simple client using static test data from TUF repository.

The code aims to be straightforward implementation, using basic
concepts from Python and Command Line Interface.

This is part of theupdateframework#1518

that is provided, the better.

requirements**:


kairoaraujo:
I am still tagging it as WIP to improve in general.
The code uses basic concepts about Python. I avoided OO models and implemented the minimum number of functions to be more sequential and easier to understand. Is it a good approach?

The ``client_example`` is a subfolder in ``examples``, and the ``README.rst`` can be visible as the tutorial for it, any better suggestion?

jku (commented):
nice, thanks. Some comments (some of these could be left for future work):
* I think argparse is probably a good choice: it's verbose but part of the base library and will allow multiple subcommands without making a mess
* init should be implicit, I think (as in it should always happen)? it's the correct example for real clients and as long as the printed output is valid it should never be a problem either (also if init is implicit, then it can probably be combined with tuf_updater)
* I would not use "[INFO]" and "[ERROR]" like this: to me it implies you are using logging and that I can decrease/increase verbosity but I can't, it's just print()... maybe you should be using logging (and set default level to INFO or something)?
* linting: we want this code linted (with the same config that ngclient uses)
* docs: I think in the end the documentation should get included in the published docs (probably under a new top level "examples" title)... this does not have to happen right now but I think we should delay reviewing and polishing the doc until we see the published form. My gut feeling is that we should _not_ include details like how to clone the repo or where to find the example code etc: these are things that change over time and then someone would need to update them. We have a "instructions for contributors" document, let's link to that if we really have to but let's not repeat ourselves.

examples/client_example/client_example.py
```diff
@@ -0,0 +1,145 @@
...
+
+    client_args = argparse.ArgumentParser(
+        description="TUF Python Client Example"
+    )
+
+    # Global arguments
+    client_args.add_argument(
+        "--init",
+        default=False,
+        help="Force register a new Engine.",
```
lukpueh: ?

lukpueh (commented):
Cool stuff! I suggest we start running pylint/black/isort over code in the new examples directory too. Can you add the directory to the `lint` directive in `tox.ini`?

Also, I think the printed info text could be slightly improved (I say this as a non-native English speaker). :)

examples/client_example/README.rst
```diff
@@ -0,0 +1,92 @@
...
+#####################
+
+Introduction
+============
+
+Python Client Example, using ``python-tuf``.
+
+For information about installing ``python-tuf``, please refer to the
+`Installation documentation <https://theupdateframework.readthedocs.io/en/latest/INSTALLATION.html>`_.
+
```
lukpueh: Any reason to not use Markdown?

examples/client_example/client_example.py
kairoaraujo: @lukpueh, Friday evening, I did use a piece of my code for ``argparse``.

examples/client_example/README.rst
kairoaraujo: No reason to use Markdown, actually I did start with Markdown, but I thought it could be used with the RTD/Sphinx.

examples/client_example/client_example.py
kairoaraujo: I have to say that English is the hard part for me here. 😅

examples/client_example/README.rst
lukpueh: RTD can do Markdown too.

lukpueh:
> I have to say that English is the hard part for me here. 😅

Maybe we can ask our technical writer and chief of words @jhdalek55 for help with that.

kairoaraujo:
> nice, thanks. Some comments (some of these could be left for future work):
> 
> * I think argparse is probably a good choice: it's verbose but part of the base library and will allow multiple subcommands without making a mess 

Yes, there are more fancy libraries, but I think argparse will do the work without extra dependencies.

> * init should be implicit, I think (as in it should always happen)? it's the correct example for real clients and as long as the printed output is valid it should never be a problem either (also if init is implicit, then it can probably be combined with tuf_updater)

Yes, indeed. I was in doubt about which direction to choose.

> * I would not use "[INFO]" and "[ERROR]" like this: to me it implies you are using logging and that I can decrease/increase verbosity but I can't, it's just print()... maybe you should be using logging (and set default level to INFO or something)?

Actually, I started with logging, but I tried to make it less complex in the sense most of the basic tutorials try to use "print".
I'm open to both.

> * linting: we want this code linted (with the same config that ngclient uses) 

👍 

> * docs: I think in the end the documentation should get included in the published docs (probably under a new top level "examples" title)... this does not have to happen right now but I think we should delay reviewing and polishing the doc until we see the published form. My gut feeling is that we should _not_ include details like how to clone the repo or where to find the example code etc: these are things that change over time and then someone would need to update them. We have a "instructions for contributors" document, let's link to that if we really have to but let's not repeat ourselves.

I understand that we do not want to repeat ourselves, making it more maintainable, and I agree.
I think the challenge here is to give to who land to the example tutorial a non-segregated experience. I mean, something that you can follow in a single experience and not switch across different docs.

examples/client_example/README.rst
kairoaraujo: Should we agree to move to Markdown?

jhdalek55:
@jhdalek55 ...I'm happy to work on any English language issues. Just let me know what needs to be reviewed and when. I generally prefer to hold off on reviews until the technical content is more or less confirmed.

jku:
> Actually, I started with logging, but I tried to make it less complex in the sense most of the basic tutorials try to use "print".
I'm open to both.

Same, either way works for me.

On the question of "is this good for a tutorial/example?", I think the goals should be
* example that works out of the box
* example code that shows how to write good code that uses python-tuf 
* tutorial material

_in this order_: my opinion is that good example code is more important than a tutorial that literally anyone could follow.

kairoaraujo:
The last commit addresses more general aspects such as lint, README format, etc.
I will work on some designs improvement mentioned by @jku , so still WIP. 👨🏽‍💻

kairoaraujo:
Some questions:
- Should we remove the README.md and make the approach used in the https://github.com/theupdateframework/python-tuf/pull/1685?
- Should we add to the example code more TUF exceptions handling

lukpueh:
> * Should we remove the README.md and make the approach used in the [Docs: Add repository tutorial based on metadata API  (WIP) #1685](https://github.com/theupdateframework/python-tuf/pull/1685)?

Let me finalize my example in that PR and gather some more feedback. I'm still unsure, if a markdown wouldn't be more appropriate, in terms of making the accompanying text more readable. IMO the question boils down to readability vs. ease of testing. (see https://github.com/theupdateframework/python-tuf/issues/808#issuecomment-442155081 and https://github.com/in-toto/demo/pull/7 for related discussion about markdown doc testing)

jku:
> * Should we remove the README.md and make the approach used in the [Docs: Add repository tutorial based on metadata API #1685](https://github.com/theupdateframework/python-tuf/pull/1685)?

I think your code and 1685 are different for two reasons: 
1. this code can be, or can become, a "real" example application. The metadata API tutorial is a stopgap solution until we have a repository library
2. the client side is "well defined": all clients will operate in much the same way so it is possible for us to define the "canonical use pattern". Repository code on the other hand might be wildly different (think CLI apps vs Warehouse)

So I don't think there's a need for these two solutions to look similar.

> * Should we add to the example code more TUF exceptions handling

For the client app my thinking has been that it should need to handle RepositoryError only (although I see you had to handle some requests issues as well -- we can consider those after this merges): RepositoryError means the update cannot continue because of an issue with TUF metadata: there could be many specific reasons but the client code cannot possibly deal with any of them.

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+        description="TUF Python Client Example"
+    )
+
+    # Global arguments
+    client_args.add_argument(
+        "--init",
+        default=False,
+        help="Initializes the Client structure",
+        action="store_true",
+    )
```
jku: Do we really need an explicit init?

Couldn't the app always initialize if the expected location does not contain root.json? As long as the logging/output is clear about what it is doing it should be very clear to user: "Found trusted root in {dir}" vs "Bootstrapping trusted root in {dir}"

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+    download_parser.add_argument(
+        "target",
+        metavar="TARGET",
+        help="Target file",
+    )
+
+    command_args = vars(client_args.parse_args())
+    sub_commands_args = command_args.get("sub_commands")
+
+    verbose = command_args.get("verbose")
```
jku: argparse is not a great API but I still like using the namespace object more than downgrading it to a dict:

```suggestion
    args = client_args.parse_args()

    if args.verbose <= 1:
       loglevel = logging.ERROR

    ...

    if args.sub_commands == "download":
        download(args.target)
```
(also sub_command**s** should probably be sub_command?)

there could be a complication I don't understand though, so just a suggestion

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+
+    try:
+        updater = tuf_updater()
+
+    # if the metadata is not available, it will initialize the TUF client
+    except FileNotFoundError:
+        print("Client infrastructure not found")
+        print("Initializing the Client Infrastructure")
+        init()
```
jku: Ah so there is implicit init here as well -- I think it would look neater to always call init() but make init() cope with the fact that initialization may already be done. The `download()` (or main if you call init from there) does not need to worry or print anything about initialization, it's all handled in `init()`

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+        updater = tuf_updater()
+
+    # handle specific TUF error (root.json corrupted in the client metadata)
+    # check out ``tuf.exceptions`` for more information
+    except RepositoryError as e:
+        print(e)
+        sys.exit(1)
```
jku: No exits in middle of app code IMO: you can just return

Also, you should expect errors in other tuf calls as well -- maybe you could just handle this once (even outside download() maybe?)

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+
+    updater.refresh()
+
+    info = updater.get_targetinfo(target)
+    print(f"Target {target} information fetched")
```
jku: You don't know if it was fetched or already in cache. Maybe this print is not needed?

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if info is None:
+        print(f"Target {target} not found")
+        sys.exit(1)
+
+    path = updater.find_cached_target(info)
+    print(f"Cached target {target} verified")
```
jku: It may have been verified, failed to verify, or not even found... Maybe not a useful message?

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+    if path:
+        print(f"Target is already available in {DOWNLOAD_DIR}/{info.path}")
```
jku: this is not always correct  -- info.path is the url path which may or may not match the filename on disk. `path` contains the filesystem path

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+        sys.exit(0)
+
+    path = updater.download_target(info)
+
+    print(f"Target is available in {DOWNLOAD_DIR}/{info.path}")
```
jku: This should make it clear that we actually _downloaded_ the target

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+METADATA_DIR = f"{HOME_DIR}/.local/share/tuf_metadata_example"  # metadata dir
+CLIENT_EXAMPLE_DIR = os.path.dirname(os.path.abspath(__file__))  # example dir
+
+
+def init():
+    """
+    The function that initializes the TUF client infrastructure.
+
+    It creates the metadata directory and downloads the ``root.json``.
+    """
```
jku: please review the printing in this function: I think it prints too much but more importantly the prints are incorrect (it always claims to have done a thing even if it did not do the thing at all). 

the docstring is incorrect: root.json is not downloaded, just copied from source dir (which is the correct choice in my opinion: we are giving a good example of shipping the root.json with the app)

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+            f"{CLIENT_EXAMPLE_DIR}/1.root.json", f"{METADATA_DIR}/root.json"
+        )
+        print("Bootstrap initial root metadata")
+
+
+def tuf_updater():
+    """
+    The function that creates the TUF updater using ``tuf.ngclient``.
+    """
+    url = "http://127.0.0.1:8000"
```
jku: This is a tricky one... how do we know it's 8000? what if that's in use already?

We can leave solving this to future PRs, but the magic string should not be hidden in this helper function: you have constants at the top of file already, maybe put this there as well for now?

If you define BASE_URL at the top of file, then this function is probably unneeded as it's a single call.

examples/client_example/client_example.py
```diff
@@ -0,0 +1,171 @@
...
+import sys
+from pathlib import Path
+
+from tuf.exceptions import RepositoryError
+from tuf.ngclient import Updater
+
+# define directory constants
+HOME_DIR = Path.home()  # user home dir
+DOWNLOAD_DIR = "./downloads"  # download dir
+METADATA_DIR = f"{HOME_DIR}/.local/share/tuf_metadata_example"  # metadata dir
```
jku: the directory name should be the app name so maybe 
`METADATA_DIR = f"{HOME_DIR}/.local/share/python-tuf-client-example"
`?

Also HOME_DIR seems to be defined just for this -- maybe the variable is not needed?

jku (changes requested):
wow, ended up with a lot of comments -- if this is still WIP don't mind me

jku:
This should probably be a future PR: we want the example tested by CI (should file a new issue for that)

For inspiration, I have some (quite bad) tests for a CLI app in https://github.com/vmware-labs/repository-editor-for-tuf/blob/main/tests/test_cli.py -- I think just a very simple test is fine: run a few commands, check if output contains correct keywords, check that files exist where they should

examples/client_example/client_example.py
kairoaraujo: It was only a habit of using dictionaries for it. 🙂
As we discussed is more a choice, but I would like to use the namespaces as well.

examples/client_example/client_example.py
kairoaraujo: I decided to return a boolean, but not sure if I should always return None for now.

examples/client_example/client_example.py
kairoaraujo: I will review what I did, but insights are welcome.

examples/client_example/client_example.py
kairoaraujo: Please, confirm if it sounds ok.

kairoaraujo:
I am still looking if ``download()`` could be more "elegant." More about the exception handling.

examples/client_example/README.md
```diff
@@ -0,0 +1,57 @@
+# Python Client Example
```
jku: I think "Python" is not the primary keyword we are interested in: either `Python-tuf Client Example` or `TUF Client Example` maybe? Same below where appropriate.

examples/client_example/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Repository
+==========
+
+This example demonstrates how to use the ``python-tuf`` to build a client
+application.
+
+The repository will use static files.
+The static files are available in the ``python-tuf`` source code repository in
+``tests/repository_data/repository``.
```
jku: The start is a bit weird for a section called "Repository"... Maybe something like 

```suggestion
The example client expects to find a TUF repository running on localhost. We can use the static metadata files in ``tests/repository_data/repository`` to set one up.
```

examples/client_example/README.md
```diff
@@ -0,0 +1,57 @@
...
+   $ python3 -m http.server -d tests/repository_data/repository
+   Serving HTTP on :: port 8000 (http://[::]:8000/) ...
+```
+
+Client Example
+==============
+
+The [Client Example source code](./client_example.py>) is available entirely
+in this source code repository.
+
```
jku: Is this a useful paragraph?

examples/client_example/README.md
```diff
@@ -0,0 +1,57 @@
...
+1. Download the ``file1.txt``
+
+   ```console
+   $ ./client_example.py download file1.txt
+   Download directory [./downloads] was created
+   Metadata folder [<metadata dir>] was created
+   Added trusted root in /Users/kdearaujo/.local/share/python-tuf-client-example
+   Found trusted root in <metadata dir>
+   Target downloaded and available in ./downloads/file1.txt
+   ```
```
jku: I wonder if we should just not include the output? I expect we'll tweak and improve the script, and then the README will be out of date, and anyone reading will be confused. 

Thoughts?

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+    It creates the metadata directory and adds a trusted ``root.json``.
+    """
+
+    if not os.path.isdir(DOWNLOAD_DIR):
+        os.mkdir(DOWNLOAD_DIR)
+        print(f"Download directory [{DOWNLOAD_DIR}] was created")
+
+    if not os.path.isdir(METADATA_DIR):
+        os.makedirs(METADATA_DIR)
+        print(f"Metadata folder [{METADATA_DIR}] was created")
```
jku: I think both of these prints are spammy (not needed): in 99.9% of the cases user is not interested, and will anyway get the message about trusted root later on.

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+        shutil.copy(
+            f"{CLIENT_EXAMPLE_DIR}/1.root.json", f"{METADATA_DIR}/root.json"
+        )
+        print(f"Added trusted root in {METADATA_DIR}")
+
+    else:
+        print(f"Found trusted root in {METADATA_DIR}")
+
+
+def download(target):
```
jku: type annotation

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+
+        info = updater.get_targetinfo(target)
+
+        if info is None:
+            print(f"Target {target} not found")
+            return True
+
+        path = updater.find_cached_target(info)
+        if path:
+            print(f"Target is available in {DOWNLOAD_DIR}/{info.path}")
```
jku: `info.path` is not a filesystem path. `path` is the filesystem path

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+            return True
+
+        path = updater.download_target(info)
+        print(f"Target downloaded and available in {DOWNLOAD_DIR}/{info.path}")
```
jku: `info.path` is not a filesystem path. `path` is the filesystem path

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+        print(str(e))
+        return False
+
+    return True
+
+
+if __name__ == "__main__":
+
+    # initialize the Python Client Example infrastructure
+    init()
```
jku: nitpick but does this now happen before e.g. "--help" is handled by argparse? Maybe init() should only be called from download (and the possible other top-level commands) so that --help or syntax errors would not create files and dirs?

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+        print(f"Target downloaded and available in {DOWNLOAD_DIR}/{info.path}")
+
+    except (FileNotFoundError, RepositoryError) as e:
+        print(str(e))
+        return False
+
+    return True
+
+
+if __name__ == "__main__":
```
jku: nitpick: I guess the neatest way to handle this is to just call a `main()`  from here? this way any variables in the function are not made global

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+    download_parser.add_argument(
+        "target",
+        metavar="TARGET",
+        help="Target file",
+    )
+
+    command_args = client_args.parse_args()
+
+    if command_args.verbose <= 1:
+        loglevel = logging.ERROR
```
jku: default is 0 so doesn't this mean a single "-v" does nothing?

jku (changes requested):
I think the overall structure is good: client_args construction could be in its own function but so far it's so simple that I don't mind the current way.

Can you enable mypy for this? (currently still happens in setup.cfg I think)

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+METADATA_DIR = f"{Path.home()}/.local/share/python-tuf-client-example"
+CLIENT_EXAMPLE_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+def init():
+    """
+    The function that initializes the TUF client infrastructure.
+
+    It creates the metadata directory and adds a trusted ``root.json``.
+    """
```
jku: This almost sounds like you could fit these five lines into a single line?

`"""Initialize local trusted metadata and create a directory for downloads"""`

or something like it

examples/client_example/client_example.py
```diff
@@ -0,0 +1,140 @@
...
+        print(f"Added trusted root in {METADATA_DIR}")
+
+    else:
+        print(f"Found trusted root in {METADATA_DIR}")
+
+
+def download(target):
+    """
+    The function that downloads the target file using the TUF ``nglcient``
+    Updater.
```
jku: Fixed a typo, made this more in line with the style guide:
```suggestion
    """Download the target file using ``ngclient`` Updater.
```

examples/client_example/README.md
kairoaraujo: Visual in the source code landing page in Github is good. But I agree that it is not maintainable.

examples/client_example/client_example.py
kairoaraujo: I thought it would be interesting to show the structure created on the client side.
As we show errors such as ``FileNotFound``, the user understands what is created in the local system.

examples/client_example/client_example.py
kairoaraujo: I also was dubious about it.
- Any new functionality (top-level commands) call init() inside
- Call init from main() as this is basic infrastructure for the client

I choose the second one, should I change?

jku: it's a nitpick, so no need to change in this PR: just wanted to point that calling "--help" probably should not create files

examples/client_example/README.md
jku: I think the PR makes my point as the output listed here does not seem in sync with the code :)

examples/client_example/client_example.py
kairoaraujo: I wasn't clear in my comment.
I addressed the --help or syntax errors.

jku (approved):
I think this looks good: we'll definitely want to keep improving but this looks like a fine first step. The commit history could be squashed as it's a little random but since it's new code I don't really mind either way.

Anyone else want to have opinions here (@joshuagl, @sechkova)? 

Potential future work:
* Consider if we want the README published on RTD, modify it as required if we do
* Add a simple test for the example
* support other server locations than hard-coded localhost:8000
* support boostrapping root.json for other servers (like the repository generated by the repository example) somehow
* document the application decisions better (as an example _"why is is local metadata stored in $HOME/.local/share/"_, _"why not download first root.json over http?"_). I didn't do this in the review as I think it's a tricky balance: too much detail means no-one will read it...
* start thinking about how to use examples to show that TUF actually works, and prevents attacks. This might be more related to repository examples but ...

jku:
I didn't actually run this before mysefl but now I did: I seem to get debug level output if i don't provide a verbosity level? This clearly seems like a bug?

jku:
Merging this. 

Please file a new issue for adding a test for this example.

Maybe a new issue for the RTD documention inclusion as well? I'm not 100% sure we should include the example there but maybe someone should take a stab and see if it works well and whether it needs any changes.

PR #1677, opened 2021-11-15
Advertize type annotations with py.typed

### LINKED ISSUE #1633: advertize type annotations
Using mypy in code that imports python-tuf does not work yet. We probably need to ship a `py.typed` file (but this need to be tested: I am not sure how this works)

https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

jku:
Existence of py.typed in the top-level package tells users of the
packages that all packages are annotated.

This should fix mypy errors like this in downstream projects:
> error: Skipping analyzing "tuf.api.metadata":
> found module but no type hints or library stubs

Fixes #1633

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,6 +45,9 @@ install_requires =
 [options.packages.find]
 exclude = tests
 
+[options.package_data]
+tuf = py.typed
+
 [check-manifest]
 ignore =
   .fossa.yml


jku:
The one issue here might be upstream projects that
* use annotations
* use legacy python-tuf API 

I'm not sure if any such projects like this exist but for them py.typed would be misleading and they will have to add some mypy config to skip tuf packages... Not optimal but maybe something we can live with until we remove legacy code?

joshuagl (approved):

PR #1680, opened 2021-11-16
Update ngclient to return loaded metadata

### LINKED ISSUE #1507: ngclient: return loaded metadata from Updater._load_targets
During the traversal of the delegations tree, each targets metadata file is first loaded/downloaded and then "visited".
It would be convenient if `Updater._load_targets` directly returns the loaded metadata to avoid lookup on the next line.

**Current behavior:**
```python
            self._load_targets(role_name, parent_role)
            role_metadata: Targets = self._trusted_set[role_name].signed
```

**Expected behavior:**
```python
role_metadata = self._load_targets(role_name, parent_role)
```

ivanayov:
This changes `TrustedMetadataSet` to return new trusted Metadata
on successful calls of the `update_<role>` functions and also
changes `Updater._load_targets` to return loaded metadata as well

Fixes #1507 

- [ ] Docs have been added for the bug fix or new feature - n/a

### DIFF
--- a/tests/test_trusted_metadata_set.py
+++ b/tests/test_trusted_metadata_set.py
@@ -129,6 +129,22 @@ class TestTrustedMetadataSet(unittest.TestCase):
 
         self.assertTrue(count, 6)
 
+    def test_update_metadata_output(self):
+        timestamp = self.trusted_set.update_timestamp(self.metadata["timestamp"])
+        snapshot = self.trusted_set.update_snapshot(self.metadata["snapshot"])
+        targets = self.trusted_set.update_targets(self.metadata["targets"])
+        delegeted_targets_1 = self.trusted_set.update_delegated_targets(
+            self.metadata["role1"], "role1", "targets"
+        )
+        delegeted_targets_2 = self.trusted_set.update_delegated_targets(
+            self.metadata["role2"], "role2", "role1"
+        )
+        self.assertIsInstance(timestamp.signed, Timestamp)
+        self.assertIsInstance(snapshot.signed, Snapshot)
+        self.assertIsInstance(targets.signed, Targets)
+        self.assertIsInstance(delegeted_targets_1.signed, Targets)
+        self.assertIsInstance(delegeted_targets_2.signed, Targets)
+
     def test_out_of_order_ops(self):
         # Update snapshot before timestamp
         with self.assertRaises(RuntimeError):
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -141,7 +141,7 @@ class TrustedMetadataSet(abc.Mapping):
         return self._trusted_set.get("targets")
 
     # Methods for updating metadata
-    def update_root(self, data: bytes) -> None:
+    def update_root(self, data: bytes) -> Metadata[Root]:
         """Verifies and loads 'data' as new root metadata.
 
         Note that an expired intermediate root is considered valid: expiry is
@@ -153,6 +153,9 @@ class TrustedMetadataSet(abc.Mapping):
         Raises:
             RepositoryError: Metadata failed to load or verify. The actual
                 error type and content will contain more details.
+
+        Returns:
+            Deserialized and verified root Metadata object
         """
         if self.timestamp is not None:
             raise RuntimeError("Cannot update root after timestamp")
@@ -182,7 +185,9 @@ class TrustedMetadataSet(abc.Mapping):
         self._trusted_set["root"] = new_root
         logger.info("Updated root v%d", new_root.signed.version)
 
-    def update_timestamp(self, data: bytes) -> None:
+        return new_root
+
+    def update_timestamp(self, data: bytes) -> Metadata[Timestamp]:
         """Verifies and loads 'data' as new timestamp metadata.
 
         Note that an intermediate timestamp is allowed to be expired:
@@ -199,6 +204,9 @@ class TrustedMetadataSet(abc.Mapping):
             RepositoryError: Metadata failed to load or verify as final
                 timestamp. The actual error type and content will contain
                 more details.
+
+        Returns:
+            Deserialized and verified timestamp Metadata object
         """
         if self.snapshot is not None:
             raise RuntimeError("Cannot update timestamp after snapshot")
@@ -251,6 +259,8 @@ class TrustedMetadataSet(abc.Mapping):
         # timestamp is loaded: raise if it is not valid _final_ timestamp
         self._check_final_timestamp()
 
+        return new_timestamp
+
     def _check_final_timestamp(self) -> None:
         """Raise if timestamp is expired"""
 
@@ -260,7 +270,7 @@ class TrustedMetadataSet(abc.Mapping):
 
     def update_snapshot(
         self, data: bytes, trusted: Optional[bool] = False
-    ) -> None:
+    ) -> Metadata[Snapshot]:
         """Verifies and loads 'data' as new snapshot metadata.
 
         Note that an intermediate snapshot is allowed to be expired and version
@@ -282,6 +292,9 @@ class TrustedMetadataSet(abc.Mapping):
         Raises:
             RepositoryError: data failed to load or verify as final snapshot.
                 The actual error type and content will contain more details.
+
+        Returns:
+            Deserialized and verified snapshot Metadata object
         """
 
         if self.timestamp is None:
@@ -347,6 +360,8 @@ class TrustedMetadataSet(abc.Mapping):
         # snapshot is loaded, but we raise if it's not valid _final_ snapshot
         self._check_final_snapshot()
 
+        return new_snapshot
+
     def _check_final_snapshot(self) -> None:
         """Raise if snapshot is expired or meta version does not match"""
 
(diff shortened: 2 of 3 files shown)


tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -182,7 +182,9 @@ def update_root(self, data: bytes) -> None:
         self._trusted_set["root"] = new_root
         logger.info("Updated root v%d", new_root.signed.version)
 
-    def update_timestamp(self, data: bytes) -> None:
+        return new_root
+
+    def update_timestamp(self, data: bytes) -> Metadata:
```
MVrachev:
``` suggestion
    def update_timestamp(self, data: bytes) -> Metadata[Timestamp]:
```
This will give hint to our IDEs that the return type is not just Metadata, but Timestamp.
See an example:
https://github.com/theupdateframework/python-tuf/blob/ec285a448c5fa2ef35f765bd297b147d92e408b1/tuf/ngclient/_internal/trusted_metadata_set.py#L162

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -260,7 +264,7 @@ def _check_final_timestamp(self) -> None:
 
     def update_snapshot(
         self, data: bytes, trusted: Optional[bool] = False
-    ) -> None:
+    ) -> Metadata:
```
MVrachev: Same comment here about the type annotation.

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -375,7 +381,7 @@ def update_targets(self, data: bytes) -> None:
 
     def update_delegated_targets(
         self, data: bytes, role_name: str, delegator_name: str
-    ) -> None:
+    ) -> Metadata:
```
MVrachev: Same comment here for the type annotation.

MVrachev (commented):
I left some small suggestions.
Will review it tomorrow more precisely.

tuf/ngclient/updater.py
```diff
@@ -417,10 +429,9 @@ def _preorder_depth_first_walk(
 
             # The metadata for 'role_name' must be downloaded/updated before
             # its targets, delegations, and child roles can be inspected.
-            self._load_targets(role_name, parent_role)
+            role_metadata = self._load_targets(role_name, parent_role)
```
jku: Maybe something like this:
```suggestion
            targets = self._load_targets(role_name, parent_role).signed
```
and then using `targets` in the rest of the code instead of `role_metadata.signed` would make the code lines a little shorter -- I realize it leads to using "targets.targets", but that's more an issue with the API...

jku (commented):
Looks pretty good... in addition to the annotation comment Martin made (that applies everywhere where we know the contained type):
* it looks like we'll only use the return values from update_delegated_targets() but I agree it still makes sense to make all `TrustedMetadataSet.update_*()` methods return values in the same way: it keeps the TrustedMetadataSet API consistent
* `TrustedMetadataSet.update_targets()` does not return a value yet, it should
* I wonder about `Updater._load_*()` functions though -- they are internal to Updater... maybe it does not make sense to modify a function unless we actually want to use the return value. So this would mean we'd only modify _load_targets(), and not the other methods. I'm fine with other opinions in this one though

ivanayov:
Thanks for the reviews!

The reason I left `_load_root` unchanged is because it needs to collect metadata from all root versions and I'm not sure if we need such collection and if we actually want to create it. The only reason for creating one is just to make all `_load_*` functions unified, which I don't find strong enough argument, but I'd be happy to hear feedback if it really makes sense.

ivanayov:
We agreed with @jku on returning metadata from all `update_*` functions to keep the API consistent and return it only from `_load_delegated_targets` within the `Updater`, as the output from others is never used.

tuf/ngclient/updater.py
```diff
@@ -368,12 +368,15 @@ def _load_snapshot(self) -> None:
             self._trusted_set.update_snapshot(data)
             self._persist_metadata("snapshot", data)
 
-    def _load_targets(self, role: str, parent_role: str) -> None:
+    def _load_targets(self, role: str, parent_role: str) -> Metadata:
```
MVrachev: The return type annotation should be the same as the one for `trusted_set.update_delegated_targets` or `Metadata[Targets]`.

tests/test_trusted_metadata_set.py
```diff
@@ -129,6 +129,21 @@ def test_update(self):
...
+        delegeted_targets_1 = self.trusted_set.update_delegated_targets(
+            self.metadata["role1"], "role1", "targets"
+        )
+        delegeted_targets_2 = self.trusted_set.update_delegated_targets(
+            self.metadata["role2"], "role2", "role1"
+        )
+        self.assertIsInstance(timestamp, Metadata)
+        self.assertIsInstance(snapshot, Metadata)
+        self.assertIsInstance(delegeted_targets_1, Metadata)
+        self.assertIsInstance(delegeted_targets_2, Metadata)
```
MVrachev:
```suggestion
        self.assertIsInstance(timestamp, Timestam)
        self.assertIsInstance(snapshot, Snapshot)
        self.assertIsInstance(delegeted_targets_1, Targets)
        elf.assertIsInstance(delegeted_targets_2, Targets)
```
I think it makes sense to be more precise about the types in these asserts.

MVrachev:
I realized that probably there is no sense in annotating the return type as `Metadata[Root], `Metadata[Timestamp]` or other `Metadata[T]` when we actually return `Root`, `Timestamp` from `update_root`, `update_timestamp`, etc.
Do you agree with me @jku?

jku:
> I realized that probably there is no sense in annotating the return type as `Metadata[Root], `Metadata[Timestamp]`or other`Metadata[T]`when we actually return`Root`, `Timestamp`from`update_root`, `update_timestamp`, etc. Do you agree with me @jku?

We should return Metadata[T] and it seems we do, I don't see a problem.

If we returned Root from `update_root()` mypy should complain as well.

tests/test_trusted_metadata_set.py
jku: I don't think this suggestion is correct. You'd have to check `timestamp.signed` type if you want to compare to `Timestamp`

EDIT: this is in response to Martins review

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -153,6 +153,9 @@ def update_root(self, data: bytes) -> None:
         Raises:
             RepositoryError: Metadata failed to load or verify. The actual
                 error type and content will contain more details.
+
+        Returns:
+            Metadata[Root]: deserialized root metadata
```
jku: Let's not duplicate the type annotation. Maybe:

```suggestion
            Deserialized and verified root Metadata object
```

same comment on other docstrings

jku (commented):
Thanks, looks nice. Quite a bit of work to avoid one single dict lookup, but I think the end result is a better TrustedMetadataSet API (and might be useful in the repository work later).

I just had one nit about the return value docstring

MVrachev:
> I realized that probably there is no sense in annotating the return type as `Metadata[Root], `Metadata[Timestamp]`or other`Metadata[T]`when we actually return`Root`, `Timestamp`from`update_root`, `update_timestamp`, etc. Do you agree with me @jku?

Apologize for my review here. I made a wrong assumption.
You should annotate the return type as `Metadata[Root]`, instead of `Root` as in `update_root` you are returning `new_root` which is initialized as `new_root = Metadata[Root].from_bytes(data)`.
It's the same situation for the other cases I discuss here as well.

tests/test_trusted_metadata_set.py
MVrachev: You are correct, then I will fix my suggestion to:
```suggestion
        self.assertIsInstance(timestamp.signed, Timestam)
        self.assertIsInstance(snapshot.signed, Snapshot)
        self.assertIsInstance(delegeted_targets_1,.signed Targets)
        elf.assertIsInstance(delegeted_targets_2.signed, Targets)
```
which would be better to check the `signed` type instead of the generic `Metadata` type.

Answer to Jussi's comment.

MVrachev (approved):
Thanks for addressing my comments!
LGTM!

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -370,12 +385,16 @@ def update_targets(self, data: bytes) -> None:
         Raises:
             RepositoryError: Metadata failed to load or verify. The actual
                 error type and content will contain more details.
+
+        Returns:
+            Deserialized and verified targets Metadata object
         """
-        self.update_delegated_targets(data, "targets", "root")
+        targets = self.update_delegated_targets(data, "targets", "root")
+        return targets
```
jku: trivial nitpick: unnecessary variable

jku (approved):
This looks good to me. Left a really minor comment but no need to update just for that

tuf/ngclient/_internal/trusted_metadata_set.py
ivanayov: I wondered here and decided to add it for better readability, but I'm fine with removing it.

PR #1683, opened 2021-11-18
Update ngclient to skip visited nodes on delegation tree traversal

### LINKED ISSUE #1528: ngclient: skipping visited roles
spec says (for each role in delegation tree):

> If this role has been visited before, then skip this role

ngclient does:

```python
if (role_name, parent_role) in visited_role_names:
    continue
```
So it only skips if role has been visited _from same parent_. If there is a good reason for allowing the same role multiple times in the delegation tree (against the spec), it should be documented somehow. Alternatively we should do what the spec says and make visited_role_names just a set of role names (and not a set of tuple)...

Pinging @sechkova for comment: there's probably a  reason ngclient does this (considering legacy client does not do this), happen to remember what it is?

ivanayov:
This change edits the ngclient `Updater` to traverse the delegation
tree on nodes, instead of edges in order to skip already visited
nodes.

For more detailed clarification, please review
theupdateframework/specification#177

Fixes #1528

Signed-off-by: Ivana Atanasova <iyovcheva@iyovcheva-a02.vmware.com>

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -62,7 +62,7 @@ Example::
 import logging
 import os
 import tempfile
-from typing import Optional, Set, Tuple
+from typing import Optional, Set
 from urllib import parse
 
 from securesystemslib import util as sslib_util
@@ -401,7 +401,7 @@ class Updater:
         # List of delegations to be interrogated. A (role, parent role) pair
         # is needed to load and verify the delegated targets metadata.
         delegations_to_visit = [("targets", "root")]
-        visited_role_names: Set[Tuple[str, str]] = set()
+        visited_role_names: Set[str] = set()
         number_of_delegations = self.config.max_delegations
 
         # Preorder depth-first traversal of the graph of target delegations.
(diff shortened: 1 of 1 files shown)


ivanayov:
Testing the delegation tree traversal requires changes in the implementation that are not related to this change and as a major update of the tests is already in the process of writing, I'm pushing this change relying only on current tests and logic assumptions.

tuf/ngclient/updater.py
```diff
@@ -401,17 +401,19 @@ def _preorder_depth_first_walk(
...
+        visited_role_names: Set[str] = set()
         number_of_delegations = self.config.max_delegations
 
         # Preorder depth-first traversal of the graph of target delegations.
         while number_of_delegations > 0 and len(delegations_to_visit) > 0:
 
             # Pop the role name from the top of the stack.
             role_name, parent_role = delegations_to_visit.pop(-1)
+            if parent_role not in visited_role_names:
+                visited_role_names.add(parent_role)
```
sechkova: I don't think this is needed. We always start from "targets" and do a pre-order depth-first search, if the traversal code is correct, the parent is always visited before the child.

sechkova (changes requested):
Unfortunately creating cyclic delegation graphs is not trivial yet both with old and new test code. I've been working on it #1641 ... so I'd say this is ok for now except for my comment below.

sechkova (approved):

PR #1685, opened 2021-11-18
Docs: Add repository tutorial based on metadata API

### LINKED ISSUE #1673: Docs: Add repository tutorial based on metadata API

For the lack of dedicated repository tooling - *`tuf/repository_tool.py` will be deprecated; alternatives are still being designed* - I suggest to create a simple document, that demonstrates how to create and maintain TUF repos, using only the metadata API.

These scenarios are described in the [advanced tutorial](https://github.com/theupdateframework/python-tuf/blob/develop/docs/TUTORIAL.md) and in [this usage scenario draft document](https://docs.google.com/document/d/1qBNR-_HuS312H9YjFdIbll8MW8Abe9ORYzx43z-bpEk). Plus the [repo editor](https://github.com/vmware-labs/repository-editor-for-tuf) demonstrates how to use the metadata API for this purpose (its generic UI aspect may be ignored).

**Current behavior**:
- No example code that shows how to create and maintain TUF repositories, using the metadata API exclusively.

**Expected behavior**:
- Example code that shows how to create and maintain TUF repositories, using the metadata API exclusively.

lukpueh:
*[EDIT 11/23/2021: remove draft status and update PR description]*
*[EDIT 11/25/2021: remove reviewer question about test, the PR includes one]*

Fixes #1673

**Description**

Add an excessively commented python script to demonstrate repository creation/operation using only the low-level metadata API.

**Context**
As 'repository_tool' and 'repository_lib' are being deprecated, repository metadata must to be created and maintained *manually* using the low-level Metadata API. The example code in these files shall serve as temporary replacement until a new repository tool is available.

**Contents**
 - creation of top-level metadata
 - target file handling
 - consistent snapshots
 - key management
 - top-level delegation and signing thresholds
 - target delegation
 - in-band and out-of-band metadata signing
 - writing and reading metadata files
 - root key rotation

**Questions for reviewers**
 * Is this readable enough? Should this rather be a markdown document?
 * Is it too broad? Should we move some usage scenarios to different files?
   E.g. hashed bin delegation will be in a separate file (see follow-up PR)
 * How can we demonstrate validity? With a TrustedMetadata example?
 * How can we demonstrate usage? With a client example follow-up?

**Please verify and check that the pull request fulfills the following requirements:**

### DIFF
--- a/examples/repo_example/basic_repo.py
+++ b/examples/repo_example/basic_repo.py
@@ -1,0 +1,401 @@
+"""
+A TUF repository example using the low-level TUF Metadata API.
+
+
+As 'repository_tool' and 'repository_lib' are being deprecated, repository
+metadata must be created and maintained *manually* using the low-level
+Metadata API. The example code in this file demonstrates how to
+implement similar functionality to that of the legacy 'repository_tool'
+and 'repository_lib' until a new repository implementation is available.
+
+
+Contents:
+ * creation of top-level metadata
+ * target file handling
+ * consistent snapshots
+ * key management
+ * top-level delegation and signing thresholds
+ * target delegation
+ * in-band and out-of-band metadata signing
+ * writing and reading metadata files
+ * root key rotation
+
+NOTE: Metadata files will be written to a 'tmp*'-directory in CWD.
+
+"""
+import os
+import tempfile
+from collections import OrderedDict
+from datetime import datetime, timedelta
+from pathlib import Path
+
+from securesystemslib.keys import generate_ed25519_key
+from securesystemslib.signer import SSlibSigner
+
+from tuf.api.metadata import (
+    DelegatedRole,
+    Delegations,
+    Key,
+    Metadata,
+    MetaFile,
+    Role,
+    Root,
+    Snapshot,
+    TargetFile,
+    Targets,
+    Timestamp,
+)
+from tuf.api.serialization.json import JSONSerializer
+
+
+def _in(days):
+    """Adds 'days' to now and returns datetime object w/o microseconds."""
+    return datetime.utcnow().replace(microsecond=0) + timedelta(days=days)
+
+
+# Create top-level metadata
+# =========================
+# Every TUF repository has at least four roles, i.e. the top-level roles
+# 'targets', 'snapshot', 'timestamp' and 'root'. Below we will discuss their
+# purpose, show how to create the corresponding metadata, and how to use them
+# to provide integrity, consistency and freshness for the files TUF aims to
+# protect, i.e. target files.
+
+# Common fields
+# -------------
+# All roles have the same metadata container format, for which the metadata API
+# provides a generic 'Metadata' class. This class has two fields, one for
+# cryptographic signatures, i.e. 'signatures', and one for the payload over
+# which signatures are generated, i.e. 'signed'. The payload must be an
+# instance of either 'Targets', 'Snapshot', 'Timestamp' or 'Root' class. Common
+# fields in all of these 'Signed' classes are:
+#
+# spec_version -- The supported TUF specification version number.
+# version -- The metadata version number.
+# expires -- The metadata expiry date.
+#
+# The 'version', which is incremented on each metadata change, is used to
+# reference metadata from within other metadata, and thus allows for repository
+# consistency in addition to protecting against rollback attacks.
+#
+# The date the metadata 'expires' protects against freeze attacks and allows
+# for implicit key revocation. Choosing an appropriate expiration interval
+# depends on the volatility of a role and how easy it is to re-sign them.
+# Highly volatile roles (timestamp, snapshot, targets), usually have shorter
+# expiration intervals, whereas roles that change less and might use offline
+# keys (root, delegating targets) may have longer expiration intervals.
+
+SPEC_VERSION = "1.0.19"
+
+# Define containers for role objects and cryptographic keys created below. This
+# allows us to sign and write metadata in a batch more easily.
+roles = {}
+keys = {}
+
+
+# Targets (integrity)
+# -------------------
+# The targets role guarantees integrity for the files that TUF aims to protect,
+# i.e. target files. It does so by listing the relevant target files, along
+# with their hash and length.
+roles["targets"] = Metadata[Targets](
+    signed=Targets(
+        version=1, spec_version=SPEC_VERSION, expires=_in(7), targets={}
+    ),
+    signatures=OrderedDict(),
+)
+
+# For the purpose of this example we use the top-level targets role to protect
+# the integrity of this very example script. The metadata entry contains the
+# hash and length of this file at the local path. In addition, it specifies the
+# 'target path', which a client uses to locate the target file relative to a
+# configured mirror base URL.
+#
+#      |----base URL---||-------target path-------|
+# e.g. tuf-examples.org/repo_example/basic_repo.py
+
+local_path = Path(__file__).resolve()
+target_path = f"{local_path.parts[-2]}/{local_path.parts[-1]}"
+
+target_file_info = TargetFile.from_file(target_path, local_path)
+roles["targets"].signed.targets[target_path] = target_file_info
+
+# Snapshot (consistency)
+# ----------------------
+# The snapshot role guarantees consistency of the entire repository. It does so
+# by listing all available targets metadata files at their latest version. This
+# becomes relevant, when there are multiple targets metadata files in a
+# repository and we want to protect the client against mix-and-match attacks.
+roles["snapshot"] = Metadata[Snapshot](
+    Snapshot(
+        version=1,
+        spec_version=SPEC_VERSION,
+        expires=_in(7),
+        meta={"targets.json": MetaFile(version=1)},
+    ),
+    OrderedDict(),
+)
+
+# Timestamp (freshness)
+# ---------------------
+# The timestamp role guarantees freshness of the repository metadata. It does
+# so by listing the latest snapshot (which in turn lists all the latest
+# targets) metadata. A short expiration interval requires the repository to
+# regularly issue new timestamp metadata and thus protects the client against
+# freeze attacks.
+#
+# Note that snapshot and timestamp use the same generic wireline metadata
+# format. But given that timestamp metadata always has only one entry in its
+# 'meta' field, i.e. for the latest snapshot file, the timestamp object
+# provides the shortcut 'snapshot_meta'.
+roles["timestamp"] = Metadata[Timestamp](
+    Timestamp(
+        version=1,
+        spec_version=SPEC_VERSION,
+        expires=_in(1),
+        snapshot_meta=MetaFile(version=1),
+    ),
+    OrderedDict(),
+)
+
+# Root (root of trust)
+# --------------------
+# The root role serves as root of trust for all top-level roles, including
+# itself. It does so by mapping cryptographic keys to roles, i.e. the keys that
+# are authorized to sign any top-level role metadata, and signing thresholds,
+# i.e. how many authorized keys are required for a given role (see 'roles'
+# field). This is called top-level delegation.
+#
+# In addition, root provides all public keys to verify these signatures (see
+# 'keys' field), and a configuration parameter that describes whether a
+# repository uses consistent snapshots (see section 'Persist metadata' below
+# for more details).
+#
+# For this example, we generate one 'ed25519' key pair for each top-level role
+# using python-tuf's in-house crypto library.
+# See https://github.com/secure-systems-lab/securesystemslib for more details
+# about key handling, and don't forget to password-encrypt your private keys!
+for name in ["targets", "snapshot", "timestamp", "root"]:
+    keys[name] = generate_ed25519_key()
+
+# Create root metadata object
+roles["root"] = Metadata[Root](
+    signed=Root(
+        version=1,
+        spec_version=SPEC_VERSION,
+        expires=_in(365),
+        keys={
+            key["keyid"]: Key.from_securesystemslib_key(key)
+            for key in keys.values()
+        },
+        roles={
+            role: Role([key["keyid"]], threshold=1)
+            for role, key in keys.items()
+        },
+        consistent_snapshot=True,
+    ),
+    signatures=OrderedDict(),
+)
+
+# NOTE: We only need the public part to populate root, so it is possible to use
+# out-of-band mechanisms to generate key pairs and only expose the public part
+# to whoever maintains the root role. As a matter of fact, the very purpose of
+# signature thresholds is to avoid having private keys all in one place.
+
+# Signature thresholds
+# --------------------
+# Given the importance of the root role, it is highly recommended to require a
+# threshold of multiple keys to sign root metadata. For this example we
+# generate another root key (you can pretend it's out-of-band) and increase the
+# required signature threshold.
+another_root_key = generate_ed25519_key()
+roles["root"].signed.add_key(
+    "root", Key.from_securesystemslib_key(another_root_key)
+)
+roles["root"].signed.roles["root"].threshold = 2
+
+
+# Sign top-level metadata (in-band)
+# =================================
+# In this example we have access to all top-level signing keys, so we can use
+# them to create and add a signature for each role metadata.
+for name in ["targets", "snapshot", "timestamp", "root"]:
+    key = keys[roles[name].signed.type]
+    signer = SSlibSigner(key)
+    roles[name].sign(signer)
+
+
+# Persist metadata (consistent snapshot)
+# ======================================
+# It is time to publish the first set of metadata for a client to safely
+# download the target file that we have registered for this example repository.
+#
+# For the purpose of this example we will follow the consistent snapshot naming
+# convention for all metadata. This means that each metadata file, must be
+# prefixed with its version number, except for timestamp. The naming convention
+# also affects the target files, but we don't cover this in the example. See
+# the TUF specification for more details:
+# https://theupdateframework.github.io/specification/latest/#writing-consistent-snapshots
+#
+# Also note that the TUF specification does not mandate a wireline format. In
+# this demo we use a non-compact JSON format and store all metadata in
+# temporary directory at CWD for review.
+PRETTY = JSONSerializer(compact=False)
+TMP_DIR = tempfile.mkdtemp(dir=os.getcwd())
+
+for name in ["root", "targets", "snapshot"]:
+    filename = f"{roles[name].signed.version}.{roles[name].signed.type}.json"
+    path = os.path.join(TMP_DIR, filename)
+    roles[name].to_file(path, serializer=PRETTY)
+
+roles["timestamp"].to_file(
+    os.path.join(TMP_DIR, "timestamp.json"), serializer=PRETTY
+)
+
+
+# Threshold signing (out-of-band)
+# ===============================
+# As mentioned above, using signature thresholds usually entails that not all
+# signing keys for a given role are in the same place. Let's briefly pretend
+# this is the case for the second root key we registered above, and we are now
+# on that key owner's computer. All the owner has to do is read the metadata
+# file, sign it, and write it back to the same file, and this can be repeated
+# until the threshold is satisfied.
+root_path = os.path.join(TMP_DIR, "1.root.json")
+roles["root"].from_file(root_path)
+roles["root"].sign(SSlibSigner(another_root_key), append=True)
+roles["root"].to_file(root_path, serializer=PRETTY)
+
+
+# Targets Delegation
+# ==================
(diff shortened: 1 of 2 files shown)


examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,403 @@
...
+# Every TUF repository has at least four top-level roles, i.e. 'targets',
+# 'snapshot', 'timestamp' and 'root'. Below we will discuss their purpose, show
+# how to create the corresponding metadata, and how to use them to provide
+# integrity, consistency and freshness for the files TUF aims to protect, i.e.
+# target files.
+
+
+# Common fields
+# -------------
+# All roles have the same metadata container format. On the top-most level,
```
sechkova: Maybe we can somehow point to the Metadata class here?

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,403 @@
...
+)
+
+# For the purpose of this example we use the top-level targets role to protect
+# the integrity of this very example script.
+# We use the local absolute path to compute the file info, but add a relative
+# path to the targets metadata. The latter is the relative path to the mirror
+# from which a client would download the target file.
+target_local_path = os.path.abspath(__file__)
+target_file_path = os.path.basename(__file__)
+target_file_info = TargetFile.from_file(target_file_path, target_local_path)
```
sechkova: I have no good suggestion for improvement but these paths seem confusing. I admit I had to read it twice even though I was supposed to know what is going on :grin: Maybe an example path and/vs URL could help.

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,403 @@
...
+    ),
+    OrderedDict(),
+)
+
+# Timestamp (freshness)
+# ---------------------
+# The timestamp role guarantees freshness of the repository metadata. It does
+# so by listing the latest snapshot (which in turn lists all the latest
+# targets) metadata. Choosing a short expiration interval, requires clients to
+# often re-download timestamp and thus immediately detect if new target files
```
sechkova: Doesn't the client always try to download timestamp?

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,403 @@
...
+roles["root"].signed.version += 1
+
+# Clear old signatures
+roles["root"].signatures.clear()
+
+# Sign with threshold of old keys and write
+# Just assume we do out-of-band signing for keys we don't have
+for key in [keys["root"], another_root_key] + new_root_keys:
+    roles["root"].sign(SSlibSigner(key), append=True)
+
```
sechkova, line 397: I think the part "signing with a threshold of old keys AND a threshold of the new keys" needs an emphasis.

sechkova (commented):
Thanks for the refreshment exercise :) I left some comments on the text itself.
I don't think the example is too broad, having everything at one place seems like a good idea. Maybe in time, more detailed examples can be added for some trickier cases (complex delegations, key rotations ...).

examples/repo_example/basic_repo.py
lukpueh: I can see that how it confusing. I'll try to better explain/exemplify those two paths.

examples/repo_example/basic_repo.py
lukpueh: ➕  I absolutely agree.

examples/repo_example/basic_repo.py
lukpueh: Ha! So much for "fully internalized spec". Thanks for catching this. I'll reword.

lukpueh:
Thanks for your review, @sechkova! I just finalised the example texts and squashed all commits into one. Your comments should be addressed in another commit on top of that for re-review convenience, i.e. 2fc5945.

I'd appreciate another look. Also see the updated PR description, especially the "Questions for reviewers".

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
+"""
+A TUF repository example using the low-level TUF Metadata API.
+
+
+As 'repository_tool' and 'repository_lib' are being deprecated, repository
+metadata must to be created and maintained *manually* using the low-level
```
sechkova: nit:
```suggestion
metadata must be created and maintained *manually* using the low-level
```

examples/repo_example/basic_repo.py
```diff
@@ -103,14 +105,19 @@ def _in(days):
...
-target_file_path = os.path.basename(__file__)
-target_file_info = TargetFile.from_file(target_file_path, target_local_path)
-roles["targets"].signed.targets[target_file_path] = target_file_info
+# the integrity of this very example script. The metadata entry contains the
+# hash and length of this file at the local path. In addition, it specifies the
+# 'target path', which a client uses to locate the target file relative to a
+# configured mirror base URL.
+#
+#      |----base URL---||-------target path-------|
+# e.g. tuf-examples.org/repo_example/basic_repo.py
```
sechkova, line 115: I like it!

lukpueh:
I just added very basic testing that runs the example script using `exec` and checks if the metadata files were created as expected. It currently only checks for file existence, but it wouldn't be hard to add more thorough inspection of the created files.

Note that the test class is intended to be re-used for a similar example script that I have in the pipe.

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# The snapshot role guarantees consistency of the entire repository. It does so
+# by listing all available targets metadata files at their latest version. This
+# becomes relevant, when there are multiple targets metadata files in a
+# repository and we want to protect the client against mix-and-match attacks.
+roles["snapshot"] = Metadata[Snapshot](
+    Snapshot(
+        version=1,
+        spec_version=SPEC_VERSION,
+        expires=_in(31),
+        meta={"targets.json": MetaFile(1)},
```
sechkova:
```suggestion
        meta={"targets.json": MetaFile(version=1)},
```
It is a nit but since it is an example ...

examples/repo_example/basic_repo.py
joshuagl: Typo:
```suggestion
metadata must be created and maintained *manually* using the low-level
```

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
+"""
+A TUF repository example using the low-level TUF Metadata API.
+
+
+As 'repository_tool' and 'repository_lib' are being deprecated, repository
+metadata must to be created and maintained *manually* using the low-level
+Metadata API. The example code in this file shall serve as temporary
+replacement until a new repository tool is available.
```
joshuagl: WDYT to:
```suggestion
Metadata API. The example code in this file demonstrates how to
implement similar functionality to that of the legacy 'repository_tool'
and 'repository_lib' until a new repository implementation is available.
```

Indicates it's an example (and that a user should think before copy/paste verbatim) and doesn't promise to provide a new repository *tool*.

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# instance of either 'Targets', 'Snapshot', 'Timestamp' or 'Root' class. Common
+# fields in all of these 'Signed' classes are:
+#
+# spec_version -- The supported TUF specification version number.
+# version -- The metadata version number.
+# expires -- The metadata expiry date.
+#
+# The 'version', bumped on each metadata change, is used to reference metadata
+# from within other metadata, and thus allows for repository consistency in
+# addition to protecting against rollback attacks.
```
joshuagl: Is bumped obvious enough a term? If not, perhaps:
```suggestion
# The 'version', which is incremented on each metadata change, is used to
# reference metadata from within other metadata, and thus allows for repository
# consistency in addition to protecting against rollback attacks.
```

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+#
+# The expiry date, protects against freeze attacks and allows for implicit key
```
joshuagl: consistency:
```suggestion
# The 'expiry' date protects against freeze attacks and allows for implicit key
```

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# revocation. Choosing an appropriate expiration interval depends on the
+# volatility of a role and how easy it is to re-sign them. Highly volatile
+# roles (timestamp, snapshot, delegated targets) usually have shorter
+# expiration intervals, whereas roles that change less and might use offline
+# keys (root, top-level targets) can expire less frequently.
```
joshuagl: Nit: use a similar phrase?
```suggestion
# keys (root, top-level targets) may have longer expiration intervals.
```

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# 'target path', which a client uses to locate the target file relative to a
+# configured mirror base URL.
+#
+#      |----base URL---||-------target path-------|
+# e.g. tuf-examples.org/repo_example/basic_repo.py
+
+local_path = Path(__file__)
+target_path = f"{local_path.parts[-2]}/{local_path.parts[-1]}"
+
+target_file_info = TargetFile.from_file(target_path, local_path.resolve())
```
joshuagl: Nit: might be cleaner to resolve() earlier, i.e. ln 116. It looks a bit funny doing operations on `local_path` to create `target_path` and only later resolving the full path. I understand why this works, but had to read twice to make sure it did.

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# ----------------------
+# The snapshot role guarantees consistency of the entire repository. It does so
+# by listing all available targets metadata files at their latest version. This
+# becomes relevant, when there are multiple targets metadata files in a
+# repository and we want to protect the client against mix-and-match attacks.
+roles["snapshot"] = Metadata[Snapshot](
+    Snapshot(
+        version=1,
+        spec_version=SPEC_VERSION,
+        expires=_in(31),
```
joshuagl: Observation: the comments earlier suggest that snapshot is usually a shorter expiration interval than targets, but the inverse is true for this code.

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+        },
+        consistent_snapshot=True,
+    ),
+    signatures=OrderedDict(),
+)
+
+# NOTE: We only need the public part to populate root, so it is possible to use
+# out-of-band mechanisms to generate key pairs and only expose the public part
+# to whoever maintains the root role. As a matter of fact, the very purpose of
+# signature thresholds is to avoid that private keys are in one place.
```
joshuagl:
```suggestion
# signature thresholds is to avoid having private keys all in one place.
```

sechkova (approved):
Would be nice if more people provide feedback but you have my approval :grin: 
I also filled in the questionnaire.

>     Is this readable enough? Should this rather be a markdown document?
No strong opinion here, I'm ok with this one
>     Is it too broad? Should we move some usage scenarios to different files?
>     E.g. hashed bin delegation will be in a separate file (see follow-up PR)
Nope, I like it,
>     How can we demonstrate validity? With a TrustedMetadata example?
Seems like the options are either a full Updater with a server or using  TrustedMetadata by providing it with the metadata files as bytes. The second option sounds more interesting to me :) 
>     How can we demonstrate usage? With a client example follow-up?
I have nothing better to propose.

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+roles["root"].to_file(root_path, serializer=PRETTY)
+
+
+# Targets Delegation
+# ==================
+# Similar to how the root role delegates responsibilities about integrity,
+# consistency and freshness to the corresponding top-level roles, a targets
+# role may further delegate its responsibility for target files (or a subset
+# thereof) to other targets roles. This allows to build a granular trust
+# hierarchy, and further reduces the impact of a single role compromise.
```
joshuagl:
```suggestion
# thereof) to other targets roles. This allows creation of a granular trust
# hierarchy, and further reduces the impact of a single role compromise.
```

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# ---------
+# Akin to top-level delegation, the delegator expresses its trust in the
+# delegatee by authorizing a threshold of cryptographic keys to provide
+# signatures for the delegatee metadata. It also provides the corresponding
+# public key store.
+# The delegation info defined by the delegator further requires the provision
+# of a unique delegatee name and constraints about the target files the
+# delegatee is responsible for, e.g. a list of path patterns. For details about
+# all configuration parameters see
+# https://theupdateframework.github.io/specification/latest/#delegations for
```
joshuagl: typo:
```suggestion
# https://theupdateframework.github.io/specification/latest/#delegations
```

examples/repo_example/basic_repo.py
joshuagl: Nit: aids readability IMHO
```suggestion
        meta={"targets.json": MetaFile(version=1)},
```

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# Note that snapshot and timestamp use the same generic wireline metadata
+# format. But given that timestamp metadata always has only one entry in its
+# 'meta' field, i.e. for the latest snapshot file, the timestamp object
+# provides the shortcut 'snapshot_meta'.
+roles["timestamp"] = Metadata[Timestamp](
+    Timestamp(
+        version=1,
+        spec_version=SPEC_VERSION,
+        expires=_in(1),
+        snapshot_meta=MetaFile(1),
```
joshuagl: Nit: same here
```suggestion
        snapshot_meta=MetaFile(version=1),
```

tests/test_examples.py
```diff
@@ -0,0 +1,84 @@
...
+import tempfile
+import unittest
+from pathlib import Path
+
+
+class TestRepoExamples(unittest.TestCase):
+    """Unit test class for repo example scripts.
+
+    Provides a '_run_example_script' method to run (exec) a script located in
+    the 'repo example' directory.
```
joshuagl:
```suggestion
    the 'repo_example' directory.
```

tests/test_examples.py
```diff
@@ -0,0 +1,84 @@
...
+
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Locate and cache repo example dir."""
```
joshuagl:
```suggestion
        """Locate and cache repo_example dir."""
```

tests/test_examples.py
```diff
@@ -0,0 +1,84 @@
...
+        base = Path(__file__).resolve().parents[1]
+        cls.repo_examples_dir = base / "examples" / "repo_example"
```
joshuagl, line 27: Should we use `os.path.join()` or do we not want to run this on Windows?

tests/test_examples.py
```diff
@@ -0,0 +1,84 @@
...
+        os.chdir(self.base_test_dir)
+
+    def tearDown(self):
+        """Change back to original dir and remove test dir, which may contain
+        dirs/files the test created at test-time CWD."""
+        os.chdir(self.original_cwd)
+        shutil.rmtree(self.base_test_dir)
+
+    def _run_script_and_assert_files(self, script_name, filenames_created):
+        """Run script in 'repo example' dir and assert that it created the
```
joshuagl:
```suggestion
        """Run script in 'repo_example' dir and assert that it created the
```

joshuagl (approved):
Thanks for this @lukpueh, beautiful work – very clear example and a testament to the Metadata API. 🎉 

> Is this readable enough? Should this rather be a markdown document?

To me this is very readable, appreciate the detail and care here.

> Is it too broad? Should we move some usage scenarios to different files?

LGTM as is. Covers the common cases, I think.

> How can we demonstrate validity? With a TrustedMetadata example?

Implicit in the below? And future work once the proposed repository API validation piece is implemented? i.e. I think this is a solid submission as-is.

> How can we demonstrate usage? With a client example follow-up?

Could be a simple follow-on that runs a http server over the data generated here and uses ngclient to fetch the python file?

examples/repo_example/basic_repo.py
```diff
@@ -0,0 +1,397 @@
...
+                    threshold=1,
+                    terminating=True,
+                    paths=["*.py"],
+                ),
+            )
+        ]
+    ),
+)
+
+# Remove target file info from top-level targets (delegatee is now repsonible)
```
jku:
```suggestion
# Remove target file info from top-level targets (delegatee is now responsible)
```

tests/test_examples.py
```diff
@@ -0,0 +1,84 @@
...
+        self.base_test_dir = os.path.realpath(tempfile.mkdtemp())
+        os.chdir(self.base_test_dir)
+
+    def tearDown(self):
+        """Change back to original dir and remove test dir, which may contain
+        dirs/files the test created at test-time CWD."""
+        os.chdir(self.original_cwd)
+        shutil.rmtree(self.base_test_dir)
+
+    def _run_script_and_assert_files(self, script_name, filenames_created):
```
jku, line 42: martin is trying to enable mypy (along with the linters) in new test files so argument type annotations would be nice -- but handling these in a follow-up issue is fine IMO.

jku (approved):
LGTM, I think this is fine as the stop gap: it's demonstrating the basic elements that a repository needs to handle, without pretending to actually be a repository implementation.

My reaction to the "is it too broad" question is that we shouldn't put a lot of effort into expanding or splitting it (although the hashed bin example should be useful), if we can push the repository library and its example code (or a repository tool) forward instead...

The repository validation question is more interesting... but even there maybe the answer isn't to expand this tutorial code . 

On the usage question: this could be a follow-up issue that is solved when initial versions of both this and the client example are present. Possible solution:
* document how to serve the generated directory over http
* add a "bootstrap from this root.json" option to the client so it can be used with the generated repository

examples/repo_example/basic_repo.py
lukpueh: The field name is 'expires'. Should we write *"The 'expires' date... "*? Does consistency beat grammar? :D 
Regardless, I'll remove the stray comma.

examples/repo_example/basic_repo.py
lukpueh: Agreed. Will update!

examples/repo_example/basic_repo.py
joshuagl: 🤦 of course.

I think it is better to bind the comments to the variables, maybe:
"The date the metadata 'expires'..." ?

examples/repo_example/basic_repo.py
lukpueh: Very good observation! My (admittedly not obvious) thinking here was that the top-level targets in the example is used *like* delegated targets, i.e. to list target files, and thus might change often, and thus might expire more quickly than a top-level target file that only delegates to other targets.

Do you think it is worth including this reasoning in the example, or does it go beyond its scope? I could try to update the comments and maybe also change the expiration interval below, when I make top-level targets a pure delegator.

tests/test_examples.py
lukpueh: This actually works on Windows too, because `pathlib` overloads the `/` operator to make it act like `join`.

https://docs.python.org/3/library/pathlib.html#operators

tests/test_examples.py
lukpueh: If you really don't mind, I'll leave this to a follow-up.

examples/repo_example/basic_repo.py
lukpueh: @joshuagl, do you think 83764c6 is enough?

lukpueh:
Thanks for the thorough reviews, @sechkova, @joshuagl and @jku! I addressed your comments in  4327214 and  83764c6 (the latter in response to https://github.com/theupdateframework/python-tuf/pull/1685#discussion_r756967300).

May I ask for another quick round of LGTMs? (NOTE: the commits are labeled WIP, because I intend to squash them before merging.)

jku (approved):
:+1:

tests/test_examples.py
joshuagl: Oh, wow. TIL. I cannot decide whether I like this magic or not 😆

examples/repo_example/basic_repo.py
joshuagl: Seems more than sufficient, thank you!

joshuagl (approved):
LGTM, thanks!

MVrachev (commented):
I just read it and liked the comments and comparing that to `tuf/repository_tool.py` is such an improvement and simplification.
Great work!

lukpueh:
Thanks, everyone!

PR #1686, opened 2021-11-18
Remove URL normalisation

### LINKED ISSUE #1483: ngclient: remove URL normalization

Document that URL is expected to be is a percent encoded URL throughout the client code and remove URL string normalization _in lower level methods_ like the one in: 
https://github.com/theupdateframework/tuf/blob/745a8f7680c69ea290d48bc5a2066dd4ad9781d3/tuf/ngclient/_internal/download.py#L88

ivanayov:
As a target path is a URL path it's not correct to consider it as
interchangeable with a filepath within every operation system. The
unquote is also removed as the ngclient cannot assume correctly
which encoding is intended and which not

Fixes #1483

### DIFF
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -112,7 +112,7 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
             + utils.TEST_HOST_ADDRESS
             + ":"
             + str(self.server_process_handler.port)
-            + repository_basepath
+            + repository_basepath.replace("\\","/")
         )
 
         self.metadata_url = f"{url_prefix}/metadata/"
--- a/tuf/ngclient/fetcher.py
+++ b/tuf/ngclient/fetcher.py
@@ -10,7 +10,6 @@ import logging
 import tempfile
 from contextlib import contextmanager
 from typing import IO, Iterator
-from urllib import parse
 
 from tuf import exceptions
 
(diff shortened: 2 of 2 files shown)


jku:
tuf/download.py is legacy code base, should not be touched.

The other failures are all in test_updater_ng: my first guess would be the test setup is broken: maybe the test setup uses local file paths as URL parts.

jku (approved):
I really dislike hard coding OS specific path separators even in test code... but it looks like not doing that would mean we'd have to make the test infrastructure simpler, which is likely a bigger task and requires more infrastructure context.

So... maybe we take this as a win (removing hard coded OS specific path separators from actual library is still a big improvement, even if we just move them to tests), and move on.

@sechkova mind having a look, any opinions?

tests/test_fetcher_ng.py
```diff
@@ -54,7 +54,7 @@ def setUp(self):
             self.file_contents = target_fileobj.read()
             self.file_length = len(self.file_contents)
 
-        self.rel_target_filepath = os.path.basename(target_filepath)
+        self.rel_target_filepath = os.path.basename(target_filepath).replace("\\","/")
```
sechkova: I think `os.path.basename `always returns the "basename" and does not contain separators. I also didn't see any errors in the CI run related to `test_fetcher_ng.py`. Is this replace here "just in case" or I am missing something?

sechkova:
I agree with you, I'm ok with changing the tests like this for now. I only have one comment about `test_fetcher_ng`.

tests/test_fetcher_ng.py
jku: good catch, looks like the correct analysis to me: replace not necessary here

ivanayov: The test was failing and this is why I added replace here, but I'll try once again without to double verify.

jku (changes requested):
marking "request changes" for @sechkovas basename finding: leave a comment if it turns out to be actually necessary

ivanayov:
@jku @sechkova you've been right. Probably `test_fetcher_ng` used to fail due to some other change that's been removed now. I discarded the changes there.

jku (approved):

jku:
Thanks!

PR #1687, opened 2021-11-19
Tests on the new implementation: address mypy warnings

### LINKED ISSUE #1657: New implementation test files: resolve linting issues discovered by pylint and mypy

MVrachev:
Fixes #1657 
Related to #1582.

This pr includes:
- configure `mypy` to show error codes
- fixed type annotations in test_metadata_serialization
- two additional invalid serialization tests
- manual fixes for a lot of `mypy` warnings.

When there were `mypy` warnings that we are calling non-annotated function
in annotated context, I decided to add annotations instead of ignoring
those warnings.
That's how I end up adding annotations in the whole tests/utils.py
module.

**NOTE:** Those changes are not done by automatic tool, but instead are manual.

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -93,11 +93,7 @@ class RepositorySimulator(FetcherInterface):
     """Simulates a repository that can be used for testing."""
 
     # pylint: disable=too-many-instance-attributes
-    def __init__(self):
-        self.md_root: Metadata[Root] = None
-        self.md_timestamp: Metadata[Timestamp] = None
-        self.md_snapshot: Metadata[Snapshot] = None
-        self.md_targets: Metadata[Targets] = None
+    def __init__(self) -> None:
         self.md_delegates: Dict[str, Metadata[Targets]] = {}
 
         # other metadata is signed on-demand (when fetched) but roots must be
@@ -117,7 +113,7 @@ class RepositorySimulator(FetcherInterface):
         # Enable hash-prefixed target file names
         self.prefix_targets_with_hash = True
 
-        self.dump_dir = None
+        self.dump_dir: Optional[str] = None
         self.dump_version = 0
 
         now = datetime.utcnow()
@@ -152,12 +148,12 @@ class RepositorySimulator(FetcherInterface):
         sslib_key = generate_ed25519_key()
         return Key.from_securesystemslib_key(sslib_key), SSlibSigner(sslib_key)
 
-    def add_signer(self, role: str, signer: SSlibSigner):
+    def add_signer(self, role: str, signer: SSlibSigner) -> None:
         if role not in self.signers:
             self.signers[role] = {}
         self.signers[role][signer.key_dict["keyid"]] = signer
 
-    def _initialize(self):
+    def _initialize(self) -> None:
         """Setup a minimal valid repository."""
 
         targets = Targets(1, SPEC_VER, self.safe_expiry, {}, None)
@@ -182,7 +178,7 @@ class RepositorySimulator(FetcherInterface):
         self.md_root = Metadata(root, OrderedDict())
         self.publish_root()
 
-    def publish_root(self):
+    def publish_root(self) -> None:
         """Sign and store a new serialized version of root."""
         self.md_root.signatures.clear()
         for signer in self.signers["root"].values():
@@ -199,12 +195,12 @@ class RepositorySimulator(FetcherInterface):
         if path.startswith("/metadata/") and path.endswith(".json"):
             # figure out rolename and version
             ver_and_name = path[len("/metadata/") :][: -len(".json")]
-            version, _, role = ver_and_name.partition(".")
+            version_str, _, role = ver_and_name.partition(".")
             # root is always version-prefixed while timestamp is always NOT
             if role == "root" or (
                 self.root.consistent_snapshot and ver_and_name != "timestamp"
             ):
-                version = int(version)
+                version: Optional[int] = int(version_str)
             else:
                 # the file is not version-prefixed
                 role = ver_and_name
@@ -216,11 +212,10 @@ class RepositorySimulator(FetcherInterface):
             target_path = path[len("/targets/") :]
             dir_parts, sep, prefixed_filename = target_path.rpartition("/")
             # extract the hash prefix, if any
+            prefix: Optional[str] = None
+            filename = prefixed_filename
             if self.root.consistent_snapshot and self.prefix_targets_with_hash:
                 prefix, _, filename = prefixed_filename.partition(".")
-            else:
-                filename = prefixed_filename
-                prefix = None
             target_path = f"{dir_parts}{sep}{filename}"
 
             yield self._fetch_target(target_path, prefix)
@@ -261,8 +256,9 @@ class RepositorySimulator(FetcherInterface):
             return self.signed_roots[version - 1]
 
         # sign and serialize the requested metadata
+        md: Optional[Metadata]
         if role == "timestamp":
-            md: Metadata = self.md_timestamp
+            md = self.md_timestamp
         elif role == "snapshot":
             md = self.md_snapshot
         elif role == "targets":
@@ -294,7 +290,7 @@ class RepositorySimulator(FetcherInterface):
         hashes = {sslib_hash.DEFAULT_HASH_ALGORITHM: digest_object.hexdigest()}
         return hashes, len(data)
 
-    def update_timestamp(self):
+    def update_timestamp(self) -> None:
         """Update timestamp and assign snapshot version to snapshot_meta
         version.
         """
@@ -307,7 +303,7 @@ class RepositorySimulator(FetcherInterface):
 
         self.timestamp.version += 1
 
-    def update_snapshot(self):
+    def update_snapshot(self) -> None:
         """Update snapshot, assign targets versions and update timestamp."""
         for role, delegate in self.all_targets():
             hashes = None
@@ -322,7 +318,7 @@ class RepositorySimulator(FetcherInterface):
         self.snapshot.version += 1
         self.update_timestamp()
 
-    def add_target(self, role: str, data: bytes, path: str):
+    def add_target(self, role: str, data: bytes, path: str) -> None:
         """Create a target from data and add it to the target_files."""
         if role == "targets":
             targets = self.targets
@@ -341,7 +337,7 @@ class RepositorySimulator(FetcherInterface):
         terminating: bool,
         paths: Optional[List[str]],
         hash_prefixes: Optional[List[str]],
-    ):
+    ) -> None:
         """Add delegated target role to the repository."""
         if delegator_name == "targets":
             delegator = self.targets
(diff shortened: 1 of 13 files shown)


jku (commented):
Quick comments:
* Can you leave comments explaining the ignores -- they don't seem entirely obvious at first glance?
* Are requests and dateutil really not annotated anywhere ? I wonder how this has not been an issue so far but now requires ignores...
* test_fetcher (and as follow up tuf/requests_fetcher.py) should not need changing as they are part of legacy code

jku:
> * Can you leave comments explaining the ignores -- they don't seem entirely obvious at first glance?

I mean PR comments, just for review

tests/test_api.py
```diff
@@ -183,7 +191,7 @@ def test_sign_verify(self):
         # Test verifying with explicitly set serializer
         targets_key.verify_signature(md_obj, CanonicalJSONSerializer())
         with self.assertRaises(exceptions.UnsignedMetadataError):
-            targets_key.verify_signature(md_obj, JSONSerializer())
+            targets_key.verify_signature(md_obj, JSONSerializer())  # type: ignore[arg-type]
```
MVrachev, line 193: Here we receive this warning:
```
tests/test_api.py:194: error: Argument 2 to "verify_signature" of "Key" has incompatible
type "JSONSerializer"; expected "Optional[SignedSerializer]"  [arg-type]
```

I disabled it because the test wanted to test setting an explicit set serializer and which will raise
an `UnsignedMetadataError` and that's what happens.
Please note that `JSONSerializer` inherits the `MetadataSerializer` and not the `SignedSerializer` (as for example the `CanonicalJSONSerializer` does).
If we replace the `JSONSerializer` with the `CanonicalJSONSerializer` the test no longer passes
as we do that same thing on line 192.

tests/test_api.py
```diff
@@ -648,7 +657,7 @@ def test_length_and_hash_validation(self):
 
             # Test wrong algorithm format (sslib.FormatError)
             snapshot_metafile.hashes = {
-                256: "8f88e2ba48b412c3843e9bb26e1b6f8fc9e98aceb0fbaa97ba37b4c98717d7ab"
+                256: "8f88e2ba48b412c3843e9bb26e1b6f8fc9e98aceb0fbaa97ba37b4c98717d7ab"  # type: ignore[dict-item]
```
MVrachev, line 659: Here we receive the following mypy warning:
```
tests/test_api.py:660: error: Dict entry 0 has incompatible type "int": "str";
expected "str": "str"  [dict-item]
```
This error hints we have annotated the `snapshot_metafile.hashes` as a `Dict[str, str]`, but we
are using it as `Dict[int, str]`.
This is expected behavior here as we purposely test an invalid hash algorithm.

tuf/requests_fetcher.py
```diff
@@ -97,7 +98,7 @@ def chunks():
           # Before beginning the round, sleep (if set) for a short amount of
           # time so that the CPU is not hogged in the while loop.
           if settings.SLEEP_BEFORE_ROUND:
-            time.sleep(settings.SLEEP_BEFORE_ROUND)
+            time.sleep(settings.SLEEP_BEFORE_ROUND)  # type: ignore[unreachable]
```
MVrachev: Here I received the error:
```
tuf/requests_fetcher.py:101: error: Statement is unreachable  [unreachable]
```
mypy warns that it's unreachable, because `settings.SLEEP_BEFORE_ROUND` is `None`, but in real
application that could be changed and that's why this warning is unnecessary.

tests/test_api.py
jku: isn't this a sign of a bug somewhere then (presumably in the annotations since everything seems to work)?

MVrachev:
> Quick comments:
> 
> * Can you leave comments explaining the ignores -- they don't seem entirely obvious at first glance?

I left comments about each of them.

> * Are requests and dateutil really not annotated anywhere ? I wonder how this has not been an issue so far but now requires ignores...

Hmm, I don't remember why I had to add them to the list in `setup.cfg`. Now when I remove them there aren't any new warnings.
Will remove them.

> * test_fetcher (and as follow up tuf/requests_fetcher.py) should not need changing as they are part of legacy code

I decided to add type annotations for them as I received warnings that I am calling unannotated code in annotated context.
I don't see it as a problem as those changes are small.

tests/test_api.py
jku: oh... you mean the test explicitly wants to use a serializer argument that is of a completely different class... but that just happens to have a method of same name `serialize()`.

That's a weird test but not a bug then

jku:
> I decided to add type annotations for them as I received warnings that I am calling unannotated code in annotated context.

I would like to avoid touching any more legacy code at least in this PR than is necessary. 

However, I am interested in those warnings -- where do they come from? Why do we end up importing unannotated code that we did not expect to import?

tuf/requests_fetcher.py
jku: if we wanted to annotate the old code base, the fix for this would be to annotate SLEEP_BEFORE_ROUND correctly... 

but I would rather not do that (and instead remove the tuf/requests_fetcher.py changes from this PR)

MVrachev:
I removed all annotations added to `tuf/request_fetcher.py` and the mypy warnings I received are:
```
tests/test_fetcher.py:61: error: Call to untyped function "RequestsFetcher" in typed context  [no-untyped-call]
tests/test_fetcher.py:75: error: Call to untyped function "fetch" in typed context  [no-untyped-call]
tests/test_fetcher.py:84: error: Call to untyped function "fetch" in typed context  [no-untyped-call]
tests/test_fetcher.py:92: error: Call to untyped function "fetch" in typed context  [no-untyped-call]
tests/test_fetcher.py:101: error: Call to untyped function "fetch" in typed context  [no-untyped-call]
tests/test_fetcher.py:117: error: Call to untyped function "fetch" in typed context  [no-untyped-call]
```
The problem is that we are testing the request_fetcher inside a test for the newer implementation.
Am I correct to consider `test_fetcher.py` as a test file for the new code?

MVrachev:
I updated the pr with:
- removing `ignore_missing_imports` from `setup.cfg` for `requests` and `dateutil`
- removed changes for `tuf/request_fetcher.py` and `tests/test_fetcher.py` as they contain code for the legacy code

tests/repository_simulator.py
```diff
@@ -115,9 +111,9 @@ def __init__(self):
         self.compute_metafile_hashes_length = False
 
         # Enable hash-prefixed target file names
-        self.prefix_targets_with_hash = True
+        self.prefix_targets_with_hash: bool = True
```
jku: nitpick: this should not need annotation? or am I missing something?

tests/test_api.py
```diff
@@ -47,8 +49,14 @@
 class TestMetadata(unittest.TestCase):
     """Tests for public API of all classes in 'tuf/api/metadata.py'."""
 
+    temporary_directory: ClassVar[str]
+    server_process_handler: ClassVar[utils.TestServerProcess]
```
jku: this seems unused?

tests/test_fetcher_ng.py
```diff
@@ -136,44 +139,44 @@ def test_response_read_timeout(self, mock_session_get):
...
             self.url, self.file_length + 4
         ) as temp_file:
             temp_file.seek(0, io.SEEK_END)
             self.assertEqual(self.file_length, temp_file.tell())
 
     # Download a file bigger than expected
-    def test_download_file_length_mismatch(self):
+    def test_download_file_length_mismatch(self) -> Iterator[Any]:
         with self.assertRaises(exceptions.DownloadLengthMismatchError):
             yield self.fetcher.download_file(self.url, self.file_length - 4)
```
jku: this has to be a bug: why would a test ever yield?

(not an important detail though: we expect a raise anyway)

tests/test_updater_with_simulator.py
```diff
@@ -115,7 +115,7 @@ def test_targets(self, test_case_data: Tuple[str, bytes, str]):
         updater = self._run_refresh()
         # target now exists, is not in cache yet
         info = updater.get_targetinfo(targetpath)
-        self.assertIsNotNone(info)
+        assert(isinstance(info, TargetFile))
```
jku: nitpick: I believe these type of asserts could still just check for none (`assert info is not None`) just in a form that mypy recognises.

tests/utils.py
```diff
@@ -48,9 +48,9 @@
 # Test runner decorator: Runs the test as a set of N SubTests,
 # (where N is number of items in dataset), feeding the actual test
 # function one test case at a time
-def run_sub_tests_with_dataset(dataset: DataSet):
-    def real_decorator(function: Callable[[unittest.TestCase, Any], None]):
-        def wrapper(test_cls: unittest.TestCase):
+def run_sub_tests_with_dataset(dataset: DataSet) -> Callable[[Callable], Callable]:
+    def real_decorator(function: Callable[[unittest.TestCase, Any], None]) -> Callable[[unittest.TestCase], None]:
```
jku: this line might fail black once you enable it...

tests/utils.py
```diff
@@ -213,12 +219,12 @@ def _start_process(self, extra_cmd_args, popen_cwd):
...
 
 
-  def _start_redirect_thread(self):
+  def _start_redirect_thread(self) -> None:
     """Starts a thread responsible to redirect stdout/stderr to the Queue."""
 
     # Run log_queue_worker() in a thread.
     # The thread will exit when the child process dies.
-    self._log_queue = queue.Queue()
+    self._log_queue: queue.Queue = queue.Queue()
```
jku: unlikely to need annotation?

jku (approved):
Wow, the lack of implicit None return value really makes this a big patch...

Good work: I left some comments but I don't think any are really required fixes. Take a look at the comments but I'm fine with merging as is.

tests/test_fetcher_ng.py
sechkova: It is actually a way to force download_file to execute and raise the error since it is a context manager and returns Iterator[IO]. The other option would be probably:
```
with self.fetcher.download_file(self.url, self.file_length - 4) as f:
    pass
```

tests/test_updater_with_simulator.py
sechkova: and why not `self.assertIsInstance` ?

tests/repository_simulator.py
MVrachev: Yes, you are right. Removing it.

tests/test_api.py
MVrachev: `temporary_directory` is used inside `tearDownClass`, but about `server_process_handler` you are correct.
Will remove it.

tests/utils.py
MVrachev: I received a warning before about that, but I can't reproduce it now.
So, yea removing it.

tests/test_updater_with_simulator.py
MVrachev: I tried  `self.assertIsInstance` and `mypy` doesn't recognize as it doesn't run our code, but just statically checks it. However, what worked was what @jku suggested using `assert info is not None`.
Will change all places where I use `assert` with this approach.

tests/test_fetcher_ng.py
MVrachev: I think it's more clear what happens the way it's now.
I added a comment using the explanation @sechkova gave.
Do you think it's better now?

PR #1688, opened 2021-11-22
Delegations: paths & path_hash_prefixes validation

### LINKED ISSUE #1461: Metadata Attribute validation: DelegationRole attributes
The targets class is implemented as a combination of two classes - `TargetFile` and `Delegations` plus the common signed attributes.

All Signed attributes are validated in various prs:
- [https://github.com/theupdateframework/tuf/pull/1430](https://github.com/theupdateframework/tuf/pull/1430) 
- [https://github.com/theupdateframework/tuf/issues/1418](https://github.com/theupdateframework/tuf/issues/1418)
- [https://github.com/theupdateframework/tuf/issues/1420](https://github.com/theupdateframework/tuf/issues/1420)

We already have validation for `TargetFile` attributes added thanks to @sechkova in [https://github.com/theupdateframework/tuf/pull/1451](https://github.com/theupdateframework/tuf/pull/1451).

For `Delegations` class. we have `keys` which are mappings of a string to a `Key` instance and we have already added validation for the `Key` instances here: [https://github.com/theupdateframework/tuf/pull/1449](https://github.com/theupdateframework/tuf/pull/1449).

Actually, the only piece class left for validation is the one responsible for `roles` in `Delegations` or `DelegatedRole`.

From the `DelegatedRole` attributes we are already validating `threshold` and `keyids` inside the parent class `Role` which we added here [https://github.com/theupdateframework/tuf/pull/1450](https://github.com/theupdateframework/tuf/pull/1450) and validation for `keyid` is added with the Key class validation  [https://github.com/theupdateframework/tuf/pull/1449](https://github.com/theupdateframework/tuf/pull/1449).

**So, we have to decide how we are going to validate `name`, `paths` and `path_hash_prefixes` attributes from `DelegatedRole`.**

MVrachev:
Fixes #1461

Add sanity types checks on Targets delegation `paths` and
`path_hash_prefixes` making sure that they are strings.

I tried using a lambda function in tuf/api/metadata.py to validate
`paths` and `path_hash_prefixes` types in order to make the checks in single lines,
but the lambda didn't have type annotations and `tox -e lint` failed with mypy warnings.

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -310,6 +310,10 @@ class TestSerialization(unittest.TestCase):
         "missing hash prefixes and paths": '{"name": "a", "keyids": ["keyid"], "threshold": 1, "terminating": false}',
         "both hash prefixes and paths": '{"name": "a", "keyids": ["keyid"], "threshold": 1, "terminating": false, \
             "paths": ["fn1", "fn2"], "path_hash_prefixes": ["h1", "h2"]}',
+        "invalid path type": '{"keyids": ["keyid"], "name": "a", "paths": [1,2,3], \
+            "terminating": false, "threshold": 1}',
+        "invalid path_hash_prefixes type": '{"keyids": ["keyid"], "name": "a", "path_hash_prefixes": [1,2,3], \
+            "terminating": false, "threshold": 1}',
     }
 
     @utils.run_sub_tests_with_dataset(invalid_delegated_roles)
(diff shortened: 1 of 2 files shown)


jku (approved):
I've been a bit negative about type checks in a many places but I think I agree with this... Without this check we'll never really notice that the data is invalid according to spec, and would even serialize the invalid metadata "succesfully"

PR #1689, opened 2021-11-22
Test delegation graphs

### LINKED ISSUE #1643: ngtests: test complex delegation graphs

Add tests dedicated to forming complex delegation graphs:
- Cyclic, diamond, big graphs
- Delegated hashed bins (RepoSimulator doesn’t support them yet)

Explore formats for describing delegations (idea: Graphviz dot format)

sechkova:
Fixes #1643 

I added the first tests that came to my mind for traversing the delegations tree/graph. I'd like to receive some feedback on these before continuing further.

For each test/subtest:
- create delegations graphs with different complexity 
- init a new repository with the respective delegations
- init a new Updater
- look for non-existing target path to force visiting all delegated roles
- verify that Updater tried to fetch the expected roles in the expected order
- verify that Updater stored the expected metadata files

requirements**:

### DIFF
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -49,7 +49,11 @@ class TestDelegationsGraphs(unittest.TestCase):
     """Test creating delegations graphs with different complexity
     and successfully updating the delegated roles metadata"""
 
+    # set dump_dir to trigger repository state dumps
+    dump_dir: Optional[str] = None
+
     def setUp(self) -> None:
+        self.subtest_count = 0
         self.temp_dir = tempfile.TemporaryDirectory()
         self.metadata_dir = os.path.join(self.temp_dir.name, "metadata")
         self.targets_dir = os.path.join(self.temp_dir.name, "targets")
@@ -58,6 +62,26 @@ class TestDelegationsGraphs(unittest.TestCase):
 
     def tearDown(self) -> None:
         self.temp_dir.cleanup()
+
+    def setup_subtest(
+        self, delegations: List[TestDelegation]
+    ) -> RepositorySimulator:
+        sim = self._init_repo(delegations)
+
+        self.subtest_count += 1
+        if self.dump_dir is not None:
+            # create subtest dumpdir
+            name = f"{self.id().split('.')[-1]}-{self.subtest_count}"
+            sim.dump_dir = os.path.join(self.dump_dir, name)
+            os.mkdir(sim.dump_dir)
+            # dump the repo simulator metadata
+            sim.write()
+
+        return sim
+
+    def teardown_subtest(self) -> None:
+        # clean up after each subtest
+        utils.cleanup_dir(self.metadata_dir)
 
     def _init_updater(self, sim: RepositorySimulator) -> Updater:
         """Create a new Updater instance"""
@@ -213,7 +237,7 @@ class TestDelegationsGraphs(unittest.TestCase):
             exp_files = [*TOP_LEVEL_ROLE_NAMES, *test_data.visited_order]
             exp_calls = [call(role, 1) for role in test_data.visited_order]
 
-            sim = self._init_repo(test_data.delegations)
+            sim = self.setup_subtest(test_data.delegations)
             updater = self._init_updater(sim)
             # Call explicitly refresh to simplify the expected_calls list
             updater.refresh()
@@ -233,11 +257,14 @@ class TestDelegationsGraphs(unittest.TestCase):
                 self.assertListEqual(wrapped_fetch.call_args_list, exp_calls)
                 self._assert_files_exist(exp_files)
         finally:
-            # clean up after each subtest
-            utils.cleanup_dir(self.metadata_dir)
+            self.teardown_subtest()
 
 
 if __name__ == "__main__":
+    if "--dump" in sys.argv:
+        TestDelegationsGraphs.dump_dir = tempfile.mkdtemp()
+        print(f"Repository Simulator dumps in {TestDelegationsGraphs.dump_dir}")
+        sys.argv.remove("--dump")
 
     utils.configure_test_logging(sys.argv)
     unittest.main()
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -12,7 +12,6 @@ import tempfile
 import unittest
 from dataclasses import astuple, dataclass, field
 from typing import Iterable, List, Optional
-from unittest.mock import call, patch
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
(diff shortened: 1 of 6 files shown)


tests/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,218 @@
...
+@dataclass
+class Delegation:
+    def __init__(
+        self,
+        delegator: str,
+        rolename: str,
+        keyids: Optional[List[str]] = None,
+        threshold: int = 1,
+        terminating: bool = False,
+        paths: Optional[List[str]] = ["*"],
```
jku: I think the default should be none, and then you can special case that 
```
if paths is None and path_hash_prefixes is None:
   paths = ["*"]
```
otherwise you get bugs when someone actually adds path_hash_prefixes tests

tests/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,218 @@
...
+    SPECIFICATION_VERSION,
+    TOP_LEVEL_ROLE_NAMES,
+    DelegatedRole,
+    Targets,
+)
+from tuf.ngclient import Updater
+
+
+@dataclass
+class Delegation:
```
jku: it's probably not a dataclass if it has a constructor.

the name should maybe change... this is a bit confusing wrt Metadata API

tests/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,218 @@
...
+            "visited roles": ["A", "C", "D", "B"],
+        },
+        "3-level tree terminating": {
+            "delegations": [
+                Delegation("targets", "A"),
+                Delegation("targets", "B"),
+                Delegation("A", "C", terminating=True),
+                Delegation("C", "D"),
+            ],
+            "visited roles": ["A", "C", "D"],
```
jku: these could have a comment pointing out the interesting bit (if it does not fit into the subtest name) -- like I think this is testing that _B is not visited because C terminates_ but I think spelling it out while you still remember could be good...

Alternatively maybe the comment could be the test name?

tests/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,218 @@
...
+        """Assert that local metadata files exist for 'roles'"""
+        expected_files = sorted([f"{role}.json" for role in roles])
+        local_metadata_files = sorted(os.listdir(self.metadata_dir))
+        self.assertListEqual(local_metadata_files, expected_files)
+
+    graphs: utils.DataSet = {
+        "single delegation": {
+            "delegations": [Delegation("targets", "A")],
+            "visited roles": ["A"],
+        },
```
jku: The only tweak I could think of to the test data definition is to define a dataclass for the subtest data as well... 
```python
"single_delegation": DelegationTest(
   delegations = [Delegation("targets", "A")],
   expect_visited = ["A"]
)
```
but it is just as verbose so not a very useful suggestion. It does get  rid of using strings when attributes can work, and would make the types visible in the test code: so you could just use the automatically annotated `data.delegations` instead having to do `delegations: List[Delegation] = data["delegations"]`

jku (commented):
Looks pretty cool, probably a good way to go. It's definitely an improvement over anything that currently exists.

The test data definition is quite chunky, 5-10 lines per subtest, but on the other hand it is readable (at least if test names are better or a comment is added)... and I also can't think of better alternatives: your method allows testing cyclic graphs etc which a dict would have trouble doing. 

I would like to be able to run "--dump" to see the actual client side metadata to verify what it looks like. It's not 100% trivial and might look a bit goofy with subtests but can you try to enable that? There's an example in test_updater_key_rotations -- the trick with subtests is that you need to use a new dump directory for each subtest and the directory names are then not great.

All of the comments are really just suggestions: this looks good to me.

jku:
> * look for non-existing target path to force visiting all delegated roles

I'm sure there's loads of tests to be written but at the very least we will want to add tests that actually find the target ... that might require modifying the test data structure as well but starting with this and then adding tests in other PRs sounds good to me

tests/test_updater_delegation_graphs.py
jku: actually the name isn't _that_ bad: I thought it overrode the Metadata API class but that is Delegation**s**... an alternative non-confusing name would be better but the current one works fine.

sechkova:
The main improvements are:
- the data classes TestDelegation and DelegationsTestCase
- tried to improve the sub test names and comments, not sure how much better it is now
- added --dump option to all new updater tests

I did modifications at several places so at the end I thought a whole re-review would be better than new commits with the changes.

tests/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,246 @@
...
+        spec_version = ".".join(SPECIFICATION_VERSION)
+
+        for d in delegations:
+            if d.rolename in sim.md_delegates:
+                targets = sim.md_delegates[d.rolename].signed
+            else:
+                targets = Targets(1, spec_version, sim.safe_expiry, {}, None)
+
+            # unpack 'd' but skip "delegator"
+            role = DelegatedRole(*astuple(d)[1:])
```
jku, line 110: oh wow. mypy is going to have an aneurysm I bet :)

tests/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,246 @@
...
+            sim = self._init_repo(test_data.delegations)
+            updater = self._init_updater(sim)
+            # Call explicitly refresh to simplify the expected_calls list
+            updater.refresh()
+            # Check that metadata dir contains only top-level roles
+            self._assert_files_exist(TOP_LEVEL_ROLE_NAMES)
+
+            with patch.object(
+                sim, "_fetch_metadata", wraps=sim._fetch_metadata
+            ) as wrapped_fetch:
```
jku: Once the fetcher statistics PR is in we should modify this to use them ... but it's not in quite yet

jku (approved):
I think these are cool and something this project has never had. With merge conflict fixed LGTM.

Your  sub test cleanup commit implies we should have a proper solution for that? The subtest model seems to work really well for our use cases but the manual cleanup is really not optimal...

These tests only cover the "no search results case" but I suppose the same setup (maybe even same test data?) can be used for the positive case in follow up PRs?

sechkova:
> Your sub test cleanup commit implies we should have a proper solution for that? The subtest model seems to work really well for our use cases but the manual cleanup is really not optimal...

I agree, I will open an issue for improving it, I haven't figured out a clever solution for now. 

> These tests only cover the "no search results case" but I suppose the same setup (maybe even same test data?) can be used for the positive case in follow up PRs?

The follow up PR: #1711

tests/test_updater_delegation_graphs.py
sechkova: Maybe it was paralysed by the shock because it stayed silent when I tried 😁

jku:
> I agree, I will open an issue for improving it, I haven't figured out a clever solution for now.

I guess it would have to integrate with the subtest runner somehow: maybe we could make `class SubTestCase(unittest.TestCase)` that contains the subtest decorator as class method (if that is even possible) and then the decorator could handle calling `cls.setup_subtest()` and `cls.teardown_subtest()`?

I guess the adult way of solving this is to use some test library that does this for us :shrug: but where's the fun in that?

sechkova:
The conflicts are resolved, I added one extra [commit](
ff8e5a3 ) for using `ReposiotrySimulator.fetch_tracker`.

sechkova:
> > I agree, I will open an issue for improving it, I haven't figured out a clever solution for now.
> 
> I guess it would have to integrate with the subtest runner somehow: maybe we could make `class SubTestCase(unittest.TestCase)` that contains the subtest decorator as class method (if that is even possible) and then the decorator could handle calling `cls.setup_subtest()` and `cls.teardown_subtest()`?
> 
> I guess the adult way of solving this is to use some test library that does this for us shrug but where's the fun in that?

Here is the issue: #1719

MVrachev:
Could you please rebase on top of develop?
In #1710 we started linting our tests and it will be good to lint the changes made in this pr.

sechkova:
Several "rebases" happened after the last review. Should be ok.

tests/test_updater_top_level_update.py
```diff
@@ -53,7 +56,16 @@ def setUp(self) -> None:
...
 
+        if self.dump_dir is not None:
+            # create test specific dump directory
+            name = self.id().split(".")[-1]
+            self.sim.dump_dir = os.path.join(self.dump_dir, name)
+            os.mkdir(self.sim.dump_dir)
+
     def tearDown(self) -> None:
+        if self.dump_dir is not None:
+            self.sim.write()
```
jku: Minor nit: dumping here and not when client refreshes means that you only get the end state -- some of the tests do have multiple repository states (like `test_new_timestamp_snapshot_rollback`) and those are not visible like this.

tests/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@dataclass
+class TestDelegation:
+    delegator: str
+    rolename: str
+    keyids: List[str] = field(default_factory=list)
+    threshold: int = 1
+    terminating: bool = False
+    paths: List[str] = field(default_factory=lambda: ["*"])
```
jku, line 34: nit: when you want to actually test path_hash_prefixes this needs to be optional... but we can fix that when it happens

jku (approved):
I'm good to go with this as a start. Left some minor comments, please have a look and decide for yourself.

We do want more delegation tests in future. Some obvious ones:
* delegation paths that actually match the targetpath (current test never has any hits)
* delegation paths with directories (and wildcards in directories)
* path_hash_prefixes

tests/test_updater_top_level_update.py
sechkova: Yes, that's right, I moved it to `_run_refresh` and `_init_updater`: 1ba93015c33f4141f5b097e8ac601c4522823cad

sechkova:
> We do want more delegation tests in future. Some obvious ones:
> 
>     * delegation paths that actually match the targetpath (current test never has any hits)
This is done in the follow up PR #1711 

> 
>     * delegation paths with directories (and wildcards in directories)
Yeah, this can be added I guess here  #1711 
>     * path_hash_prefixes
Already an issue about it: https://github.com/theupdateframework/python-tuf/issues/1639#issuecomment-988959898

PR #1690, opened 2021-11-22
Prevents delegate role name as top-level role name

### LINKED ISSUE #1558: Metadata API: prevent Delegation role names to be one of top level metadata roles

kairoaraujo:
This commit adds the validation in the ``metadata.Delegations``
to prevent that one of the delegate role names given is a top-level
role name.

A ``ValueError`` will be raised if one of the roles names in the
the list given to as delegated contains the role name as one of the
top-level roles.

The _mirrors role_ is not covered in this PR.

that is provided, the better.

Fixes #1558 

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -330,6 +330,11 @@ class TestSerialization(unittest.TestCase):
                 {"keyids": ["keyid2"], "name": "a", "paths": ["fn3"], "terminating": false, "threshold": 2}    \
                 ] \
             }',
+        "using empty string role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid1"], "name": "", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
+            }',
         "using root as delegate role name": '{"keys": { \
                 "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
             "roles": [ \
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -330,6 +330,33 @@ class TestSerialization(unittest.TestCase):
                 {"keyids": ["keyid2"], "name": "a", "paths": ["fn3"], "terminating": false, "threshold": 2}    \
                 ] \
             }',
+        "using root as delegate role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid1"], "name": "root", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
+            }',
+        "using snapshot as delegate role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid1"], "name": "snapshot", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
+            }',
+        "using targets as delegate role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid1"], "name": "targets", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
+            }',
+        "using timestamp as delegate role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid1"], "name": "timestamp", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
+            }',
+        "using valid and top-level role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
+                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid1"], "name": "b", "terminating": true, "paths": ["fn1"], "threshold": 3}, \
+                {"keyids": ["keyid2"], "name": "root", "terminating": true, "paths": ["fn2"], "threshold": 4} ] \
+            }',
     }
 
     @utils.run_sub_tests_with_dataset(invalid_delegations)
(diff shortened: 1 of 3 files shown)


tests/test_metadata_serialization.py
```diff
@@ -360,6 +360,43 @@ def test_delegation_serialization(self, test_case_data: str):
         delegation = Delegations.from_dict(copy.deepcopy(case_dict))
         self.assertDictEqual(case_dict, delegation.to_dict())
 
+    valid_delegations: utils.DataSet = {
+        "using root as delegate role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid"], "name": "root", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
```
MVrachev:
```suggestion
                {"keyids": ["keyid1"], "name": "root", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
```

Please note that the `keyids` used in `Targets.Delegations.DelegationRole.keys` should then be used inside `Targets.Delegations.DelegationRole.roles`.
See an example in our hardcoded targets file [here](https://github.com/theupdateframework/python-tuf/blob/develop/tests/repository_data/repository/metadata/targets.json).

Please make that change in the other test cases as well.

tests/test_metadata_serialization.py
```diff
@@ -360,6 +360,43 @@ def test_delegation_serialization(self, test_case_data: str):
         delegation = Delegations.from_dict(copy.deepcopy(case_dict))
         self.assertDictEqual(case_dict, delegation.to_dict())
 
+    valid_delegations: utils.DataSet = {
```
MVrachev: Can you please use the dataset defined at line 321 https://github.com/theupdateframework/python-tuf/blob/acb201d6cc685c2560f1000faf47e75db73df7fb/tests/test_metadata_serialization.py#L321
and the test function defined in 336 https://github.com/theupdateframework/python-tuf/blob/acb201d6cc685c2560f1000faf47e75db73df7fb/tests/test_metadata_serialization.py#L336
instead of defining a `valid_delegations` and a new test function?

We aim to group all valid tests related to a specific class in one dataset and test function.
Also, the name `valid_delegations` is not correct.
Those are actually cases of `invalid_delegations` as when you call `Delegations.from_dict(copy.deepcopy(case_dict))` it causes `ValueError` as you catch it inside your the test function.

tests/test_metadata_serialization.py
```diff
@@ -360,6 +360,43 @@ def test_delegation_serialization(self, test_case_data: str):
...
+            }',
+        "using targets as delegate role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid"], "name": "targets", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
+            }',
+        "using timestamp as delegate role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid"], "name": "root", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
```
MVrachev:
```suggestion
                {"keyids": ["keyid"], "name": "timestamp", "terminating": true, "paths": ["fn1"], "threshold": 3}] \
```

tests/test_metadata_serialization.py
```diff
@@ -360,6 +360,43 @@ def test_delegation_serialization(self, test_case_data: str):
...
+            }',
+        "using valid and top-level role name": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
+                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid"], "name": "b", "terminating": true, "paths": ["fn1"], "threshold": 3}, \
```
MVrachev:
```suggestion
                {"keyids": ["keyid1"], "name": "b", "terminating": true, "paths": ["fn1"], "threshold": 3}, \
```

tuf/api/metadata.py
```diff
@@ -1213,6 +1213,12 @@ def __init__(
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
         self.keys = keys
+        if [role for role in set(roles) if role in TOP_LEVEL_ROLE_NAMES]:
+            raise ValueError(
+                "A delegated role can not use top-level role names ("
+                f"{', '.join(TOP_LEVEL_ROLE_NAMES)})"
+            )
```
MVrachev: Can you use a shorter exception message?
That's something I have been guilty of on numerous occasions and @jku asked me to refine my messages. :D 
Maybe something like
``` suggestion
            raise ValueError("Delegated roles cannot use top-level role names")
```
An example of a similar message for Root roles where they should be top-level metadata names:
https://github.com/theupdateframework/python-tuf/blob/acb201d6cc685c2560f1000faf47e75db73df7fb/tuf/api/metadata.py#L733

MVrachev (changes requested):
Thank you @kairoaraujo for your pr and effort!
There are a couple of things to fix, but it's a great start.

Please note that at the current stage `mirrors` won't be implemented in the new implementation. For more reference read the second point from issue #1317.

tests/test_metadata_serialization.py
kairoaraujo: I see I messed up it.

tuf/api/metadata.py
kairoaraujo: Sounds good. What about only showing the role conflicting, such as:
https://github.com/theupdateframework/python-tuf/blob/acb201d6cc685c2560f1000faf47e75db73df7fb/tuf/api/metadata.py#L1231

tests/test_metadata_serialization.py
kairoaraujo: I have a concern about how to make sure that the ``ValueError`` is the correct unit we want.
I mean, we test the ``ValueError`` raised, but we don't check if the message comes from the part we want to cover.
Unless we don't want to go deep in this level, it is ok.

MVrachev: I think this is a more general question that probably is worth adressing.
I think for sure we want to have all invalid tests for `Delegations` in one single place, but I understand what you mean.
Maybe we can check the error message and type inside the with?
What do you think @jku?

tuf/api/metadata.py
MVrachev: I don't think it's absolutely correct to say that this role is a duplicate.
Yes, we are using the role name for our top-level roles, but it can be understood as if there is another role with the same name inside `Targets.Delegations` which is untrue.
Here the focus is that we don't want delegation role names to be one of the top-level metadata roles as this could lead to unexpected behaviors.

tests/test_metadata_serialization.py
kairoaraujo: Yes, indeed. If agreed in checking the type and error message, we can address it in another issue.

sechkova:
The [issue discussion](https://github.com/theupdateframework/python-tuf/issues/1558#issuecomment-924846015) mentions also preventing delegated roles from being empty strings. Do you mind adding this check as part of this PR? Sorry if it's been done already somewhere and I've missed that.

tuf/api/metadata.py
```diff
@@ -1213,8 +1213,12 @@ def __init__(
...
     ):
         self.keys = keys
-        if [role for role in set(roles) if role in TOP_LEVEL_ROLE_NAMES]:
-            raise ValueError("Delegated roles cannot use top-level role names")
+
+        for role in set(roles):
+            if not role or role in TOP_LEVEL_ROLE_NAMES:
+                raise ValueError(
+                    "Delegated roles cannot be empty or use top-level role names"
+                )
```
MVrachev: You still can write the check as a one-line like this:
```suggestion
        if any(not role or role in TOP_LEVEL_ROLE_NAMES for role in set(roles)):
            raise ValueError(
                "Delegated roles cannot be empty or use top-level role names"
            )
```
and I personally prefer it.
@sechkova what do you think?

MVrachev:
@sechkova can you approve running the CI workflows as GitHub says that:
`First-time contributors need a maintainer to approve running workflows`.

tuf/api/metadata.py
kairoaraujo: Yes, I had a small talk with @jku, and we came out about making it more readable. I'm open about any option 🙂

sechkova: Thanks for adding the check, I agree that separate lines are more readable.

sechkova (commented):
I don't know if @MVrachev has any more comments, looks ok to me.

MVrachev:
Thanks for asking.
I don't have any other comments.
LGTM!

sechkova (approved):

tuf/api/metadata.py
jku: set() does not seem necessary?

jku (commented):
LGTM, but there does seem to be an unnecessary set construction in there

jku:
Thanks!

PR #1691, opened 2021-11-23
Tests: add tests for non-root key rotations

### LINKED ISSUE #1647: tests: ngclient non-root key rotations
We probably should have a few key rotation tests for non-root metadata as well. There aren't as many different scenarios as with root since "key rotation" just means verifying a delegate with threshold in this case (and we already test `verify_delegate()` itself pretty well) but let's add something to test_updater_key_rotations.py, for all non-root keys in root metadata and for a delegated targets key in a targets metadata as well.

See also #1607

MVrachev:
Fixes #1647

The amount of non-root key rotations is lower than the one for root as
the repository stores all root versions in order to be able to do
rollback check compared to timestamp, snapshot, and targets where it
stores only one version.

All of the root roles are needed in those tests again as root
delegates the keys for each of the other three top-level metadata roles.

Additionally, this pr contains some small fixes in the test_updater_key_rotations.py module.

requirements**:

### DIFF
--- a/tests/test_updater_key_rotations.py
+++ b/tests/test_updater_key_rotations.py
@@ -10,14 +10,14 @@ import sys
 import tempfile
 import unittest
 from dataclasses import dataclass
-from typing import List, Optional, Type
+from typing import Dict, List, Optional, Type
 
 from securesystemslib.signer import SSlibSigner
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
 from tests.utils import run_sub_tests_with_dataset
-from tuf.api.metadata import Key, Root
+from tuf.api.metadata import Key, Metadata, Root
 from tuf.exceptions import UnsignedMetadataError
 from tuf.ngclient import Updater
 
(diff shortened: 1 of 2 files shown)


MVrachev:
I have rebased on top of develop.
@jku can you have a look?

tests/test_updater_key_rotations.py
```diff
@@ -10,20 +10,20 @@
...
 from tests.utils import run_sub_tests_with_dataset
-from tuf.api.metadata import Key
+from tuf.api.metadata import Key, Metadata
 from tuf.exceptions import UnsignedMetadataError
 from tuf.ngclient import Updater
 
 
 @dataclass
-class RootVersion:
+class MetadataVersion:
```
jku: The lines are already long, this makes them even longer. Maybe `MdVersion` or just `Version`?

tests/test_updater_key_rotations.py
```diff
@@ -36,23 +36,25 @@ class TestUpdaterKeyRotations(unittest.TestCase):
...
 
-    def setUp(self) -> None:
-        self.sim: RepositorySimulator
-        self.metadata_dir: str
-        self.subtest_count = 0
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.sim: RepositorySimulator
+        cls.metadata_dir: str
+        cls.subtest_count = 0
```
jku: subtest_count doesn't look like class variable, It seems more like a test specific variable to me

Now `subtest_count` is admittedly a hack that's used to create directory names when dumping repository state (because the subtest name does not seem to be available during the test) so I will also happily take a solution that removes this variable altogether... 

One option I considered but did not bother testing was to use `str(md_version)`  (where md_version is the subtests input) as the directory name instead of subtest_count  -- that should be a legal directory name on linux at least? If you want to check if that works, I'm fine with trying that

tests/test_updater_key_rotations.py
```diff
@@ -188,20 +192,96 @@ def test_root_rotation(self, root_versions: List[RootVersion]) -> None:
...
+
+        Assert that refresh() result is expected and that local metadata on disk
+        is the expected one after all roots have been loaded from remote using
+        the standard client update workflow.
+        """
+        self.setup_subtest()
+        roles = ["timestamp", "snapshot", "targets"]
+        for role in roles:
+
+            # clear root keys, signers
```
jku:
```suggestion
            # clear role keys, signers
```

tests/test_updater_key_rotations.py
```diff
@@ -188,20 +192,96 @@ def test_root_rotation(self, root_versions: List[RootVersion]) -> None:
...
+                # Call fetch_metadata to sign metadata with new keys
+                expected_local_md: Metadata = self.sim._fetch_metadata(role)
+                # assert local metadata role is on disk as expected
+                md_path = os.path.join(self.metadata_dir, f"{role}.json")
+                with open(md_path, "rb") as f:
+                    data = f.read()
+                    self.assertEqual(data, expected_local_md)
+                    # md.version = 1 as only 1 version has been published
+                    md = Metadata.from_bytes(data)
+                    self.assertEqual(md.signed.version, 1)
```
jku: I wonder if these checks are useful, or if checking that correct files exist is enough?

jku (commented):
I wonder if the `for role in roles` loop really needs to contain everything it does? Publishing one new root version (that includes the changes for all three metadata) and refreshing once should be enough?

modifying keys, threshold and sigs does need to be in a loop, but nothing else is _really_ required. If the resulting file contents are checked, that needs to happen in a loop as well but I wonder if just checking that the files are there might be enough (assuming subtest_setup removed the files in between)?

tests/test_updater_key_rotations.py
MVrachev: I was wondering the same while I was writing.
The part that is in question for me is the version check as I am not sure if there is a case when the version can end up something different than 1.
The other checks seem okay for me - checking that the local metadata file exists and it's with the content we expect.
Will remove the version check and then we can discuss it again.

tests/test_updater_key_rotations.py
MVrachev: I have another workaround.
See https://github.com/theupdateframework/python-tuf/pull/1691/commits/51f6baf57df988e711e0a3db9e99a8f197c68521

tests/test_updater_key_rotations.py
MVrachev: Removed the version check. What do you think about the rest of the checks?

MVrachev:
@jku I addressed your comments and the pr is ready for another review.

MVrachev:
Rebased on top of latest develop changes because of a conflict.

tests/test_updater_key_rotations.py
jku: that is very clever

jku (approved):
The test case name trick for dumps is very clever :clap: (I hope others find the --dump feature, it really makes it easier to see what's going on)

The loop in the test makes it a little difficult to figure out what is actually being tested but it does seem to work correctly. I remember we discussed possibly implementing three separate functions for timestamp/snapshot/targets (that would use the same data and maybe same helper function): that would probably be easier to understand ... but maybe it's not worth all that work: this version works just fine.

LGTM, thanks Martin

Issue #1459, opened 2021-06-21
Create constructor(s) for creating Metadata from scratch


jku:
I think we should have method(s) that make creating new valid metadata from scratch easier as the current constructor has really been designed for the deserialization use case
* They do not necessarily need to be in metadata.py (client has no need for them)
* They should use reasonable defaults for everything possible, but take the arguments that are needed to create valid metadata

These would be useful for testing but also will make any repository tools and tutorials better.

Back of envelope proposal with a single method:

`Metadata.new_with_defaults(signed_type: Type, expires: datetime) -> Metadata`

This would do roughly
* for all, set signed.version = 1
* for all, set signed.spec_version = ".".join(SPECIFICATION_VERSION)
* for all, set  signatures = OrderedDict()
* for root, set up all of the roles (but metadata might not be valid until keys are added)
* for timestamp, create meta with snapshot Metafile with version = 1?
* for snapshot, create empty meta dict: target MetaFiles must be added afterwards
* for targets, create no delegations and empty targets

This might require a bit of experimentation to find a reasonable solution

jku:
To make this a bit more concrete:  I have another PR that creates a root from scratch that looks like this:

```python
root = Root(1, SPEC_VER, expiry, {}, {}, True)
for role in ["root", "timestamp", "snapshot", "targets"]:
    key, signer = self.create_key()
    root.roles[role] = Role([], 1)
    root.add_key(role, key)
    # store role:signer elsewhere
self.md_root = Metadata(root, OrderedDict())
```

Not pretty: it calls three constructors with at least one useless argument. It could look like 

```python
self.md_root = Metadata.new_root(expiry)
for role in self.md_root.signed.roles.keys():
    key, signer = self.create_key()
    self.md_root.signed.add_key(role, key)
    # store role:signer elsewhere
```

better? other ideas?

jku:
I have an alternative suggestion that would not add new API:

```python
md_root = Metadata(Root(expires=expiry_date))
```
* support default `None` for constructor arguments where it makes sense: I think everything except expiry has a reasonable default (the meta dictionaries are debatable)
* default values would **not**  be used by deserialization code
* using default value means the constructor should choose a good initial value for new metadata

I think I like this. The only disadvantage is that it makes the constructor signatures even longer, but I can live with that.


I think creating a minimal initial repo would look like
```python
targets = Metadata(Targets(expires=expiry_date))
snapshot = Metadata(Snapshot(expires=expiry_date))
[...]

PR #1693, opened 2021-11-24
ADR: Add New repository library design

### LINKED ISSUE #1679: Write ADR for repository library design
I'm trying to provide the following in the coming days:

* PR with Architectural Decision Record on repository library design based on Metadata API. I see three main options:
  * No repository library, rely on Metadata API to be easy enough for apps to solve the issues themselves
  * provide a repository_tool-like API
  * provide a _minimal repository abstraction_, show how to implement repositories with it
* A design document for Minimal Repository Abstraction proposal to back that up

This is part of #1136. My current writing is in https://docs.google.com/document/d/1ZsXWMP_JgsI6RLhyiDn1Fr6rSCMFVoYSBwBakMzspjs but it's _really_ not a proper design document yet

jku:
Document the decision to build a repository library on top of Metadata
API.

This ADR has "attachments": 
* [Design document for minimal repository abstraction](https://docs.google.com/document/d/1YY83J4ihztsi1Qv0dJ22EcqND8dT80AGTduwgh0trpY)
* [Prototype implementation of minimal repository abstraction](https://github.com/vmware-labs/repository-editor-for-tuf/) (this is a whole application: only small parts are part of the prototype: please see design doc)

Comments on the ADR, design doc and prototype are welcome here.

Fixes #1679


docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,128 @@
...
+## Decision Drivers
+
+* There is a consensus on removing the legacy code from python-tuf due to
+  maintainability issues
+* Metadata API makes modifying metadata far easier than legacy code base: this
+  makes significantly different designs possible
+* Not providing a "repository library" (and leaving implementers on their own)
+  may be a short term solution because of the previous point, but it does seem
+  like the project would benefit from some shared repository code and shared
+  repository design
```
joshuagl: I think we can be a little more assertive with this point.
```suggestion
* Not providing a "repository library" (and leaving implementers on their own)
  may be a short term solution because of the previous point, but in order to
  make adoption easier and to help adopters create safe implementations the
  project would benefit from some shared repository code and a shared
  repository design
```

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,128 @@
...
+
+
+## Pros and Cons of the Options
+
+### No repository packages
+
+Metadata API makes editing the repository content vastly simpler. There are
+already repository implementations built with it (RepositorySimulator in
+python-tuf tests is an in-memory implementation, while
+repository-editor-for-tuf is an external CLI tool) so clearly a repository
```
joshuagl: Might be useful to provide permalinks to these two examples

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,128 @@
...
+library is not an absolute requirement.
+
+Not providing repository packages in python-tuf does mean that external
+projects could experiment and create implementations without adding to the
+maintenance burden of python-tuf. This would be the easiest way to iterate many
+different designs and hopefully find good ones in the end.
+
+That said, there are some tricky parts of repository maintenance (e.g.
+initialization, snapshot update, hashed bin management) that would benefit from
+having a canonical implementation. Likewise, a well designed library could make
```
joshuagl: Two reasons for the canonical implementation: i) easier for python-tuf users, ii) gives other implementers something to reference.
```suggestion
having a canonical implementation, both for easier adoption of python-tuf and as
a reference for other implementations. Likewise, a well designed library could make
```

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,128 @@
...
+However, repository_tool (and parts of repo.py) are not a great API. It is
+likely that a similar API suffers from some of the same issues: it might end up
+being a substantial amount of code that is only a good fit for one application.
+
+### Minimal repository abstraction
+
+python-tuf could define a tiny repository API that
+* provides carefully selected core functionality (like core snapshot update)
+  but...
+* does not implement all repository actions itself, instead i makes it easy
```
joshuagl:
```suggestion
* does not implement all repository actions itself, instead it makes it easy
```

joshuagl (commented):
This looks good to me, thanks for the great work on this Jussi!

I made a few suggestions to make some of the recommendations more explicit. The design doc looks good, I think one thing that might make it easier to understand is a diagram.

The design has 3 layers:
1. an API with interfaces provided by python-tuf
2. implementations of the interfaces provided by the application or python-tuf
3. the application-specific repository code

A diagram showing the three layers and who owns which pieces could help folks to understand.

Also, can we translate the design doc to markdown and include it in the repo before we merge? That feels more durable than linking to a Google doc.

Aside: the design feels a lot like the delegation pattern, it reminds me a lot of how in Objective-C/NextStep derived frameworks allow you to plug in a delegate object that implements defined protocols (rather than having to subclass complex objects).

jku:
> Also, can we translate the design doc to markdown and include it in the repo before we merge? That feels more durable than linking to a Google doc.

I think I'll do this (google docs are indeed a bit too ephemeral)... The reason I didn't and am still hesitant is that the document will become outdated. But maybe if I date it and specify that it is the attachment to the design ADR it's still valuable.

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,128 @@
...
+  maintainability issues
+* Metadata API makes modifying metadata far easier than legacy code base: this
+  makes significantly different designs possible
+* Not providing a "repository library" (and leaving implementers on their own)
+  may be a short term solution because of the previous point, but it does seem
+  like the project would benefit from some shared repository code and shared
+  repository design
+* Maintainability of new library code must be a top concern
+* Allowing a wide range of repository implementations (from CLI tools to
+  minimal in-memory implementations to large scale applications like Warehouse)
```
kairoaraujo: Would it be good having a link reference for Warehouse?

kairoaraujo (commented):
My comments are small (irrelevant) details :)

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,128 @@
...
+* There is a _very_ large amount of code to maintain: repo.py,
+  repository_tool.py and repository_lib.py alone are almost 7000 lines of code.
+* The "library like" parts of the implementation do not form a good coherent
+  API: methods routinely have a large number of arguments, code still depends
+  on globals in a major way and application (repo.py) still implements a lot of
+  "repository code" itself
+* The "library like" parts of the implementation make decisions that look like
+  application decisions. As an example, repository_tool loads _every_ metadata
+  file in the repository: this is fine for CLI that operates on a small
+  repository but is unlikely to be a good choice for PyPI.
```
kairoaraujo: Would it be good having a link reference for PyPI?

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,128 @@
...
+  same abstract API
+
+The approach does have some downsides:
+* it's not a drop in replacement for repository_tool or repo.py
+* A prototype has been implemented (see Links below) but the concept is still
+  unproven
+
+## Links
+[Design document for minimal repository abstraction](https://docs.google.com/document/d/1YY83J4ihztsi1Qv0dJ22EcqND8dT80AGTduwgh0trpY)
+[Prototype implementation of minimal repository abstraction](https://github.com/vmware-labs/repository-editor-for-tuf/)
```
kairoaraujo: Both links show in the same line. Maybe use bullets?

docs/adr/0010-repository-library-design.md
jku: hmm good point... maybe I'll just avoid naming any large scale applications to keep it from bloating

jku:
Comments have been addressed. The design doc is now in git, and two diagrams have been added*.

*) the issue with diagrams is that they are images so uneditable without my google slides original of course... so we may need to re-invent this for actual API docs when that time comes. Maybe ascii art would work as well?

jku:
@lukpueh @sechkova comments?

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,136 @@
...
+The concept is still unproven but validating the design should be straight
+forward: decision could be re-evaluated in a few months if not in weeks.
+
+
+## Pros and Cons of the Options
+
+### No repository packages
+
+Metadata API makes editing the repository content vastly simpler. There are
+already repository implementations built with it[^1] so clearly a repository
```
lukpueh, line 67: Hooray for GitHub Markdown footnotes! 🎉

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,136 @@
...
+* Maintainability of new library code must be a top concern
+* Allowing a wide range of repository implementations (from CLI tools to
+  minimal in-memory implementations to large scale application servers)
+  would be good: unfortunately these can have wildly differing requirements
+
+
+## Considered Options
+
+1. No repository packages
+2. repository_tool -like API
```
lukpueh, line 43:
```suggestion
2. repository_tool-like API
```
?

docs/adr/0010-repository-library-design.md
```diff
@@ -0,0 +1,136 @@
...
+different designs and hopefully find good ones in the end.
+
+That said, there are some tricky parts of repository maintenance (e.g.
+initialization, snapshot update, hashed bin management) that would benefit from
+having a canonical implementation, both for easier adoption of python-tuf and
+as a reference for other implementations. Likewise, a well designed library
+could make some repeated actions (e.g. version bumps, expiry updates, signing)
+much easier to manage.
+
+### repository_tool -like API
```
lukpueh, line 82:
```suggestion
### repository_tool-like API
```
?

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
+# Python-tuf repository API proposal: _minimal repository abstraction_
+
+This is an attachment to ADR 10: _Repository library design built on top of
+Metadata API_, and documents the design proposal in Dec 2020.
```
lukpueh:
```suggestion
Metadata API_, and documents the design proposal in Dec 2021.
```
?

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+  where things are implemented: The repository library does not have to
+  implement every little detail as we can safely let specific implementations
+  handle things, see goal #3
+* This variance means we can start by implementing a minimal design: as
+  experience from implementations is collected, we can then move implementation
+  details into the library (goals #2, #3)
+
+## Design
+
+![Design: Application and library components](repository-library-design-ownership.jpg)
```
lukpueh, line 36: I was a bit confused by the word "generic" in "Generic App Code". Isn't this actually the code *specific* to an application? Maybe it's me, but I briefly thought that it means that it is something provided by python-tuf (generic and thus eligible for all applications).

Would "App management code" be a good choice of words?

(Same goes for the second diagram)

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+    for all code levels to use. It also implements some core edit actions like
+    snapshot update
+  * A small amount of related functionality is also provided (private key
+    management API, maybe repository validation)
+  * is a very small library: possibly a few hundred lines of code
+* Concrete Repository implementation (typically part of application code,
+  implements interfaces provided by the repository API in python-tuf)
+  * Contains the “application level” decisions that the Repository abstraction
+    requires to operate: examples of application decisions include
+    * _when should “targets” metadata next expire when it is edited?_
```
lukpueh: nit: Case consistency:
```suggestion
    * _When should “targets” metadata next expire when it is edited?_
```

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+    * _What is the current “targets” metadata version? Where do we load it 
+      from?_
+    * _Where to store current “targets” after editing? Should the previous
+      version be deleted from storage?_
+* Actual application
+  * Uses the Repository API to do the repository actions it needs to do
+
+For context here’s a trivial example showing what “ergonomic editing” means --
+this key-adding code could be in the application or in the python-tuf library:
```
lukpueh: > could be in the application or in the python-tuf library

Isn't this a decision that we should make and give the rationale for in this design document?

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+* Hides the complexity of loading and persisting metadata, and updating expiry
+  and versions from the editing code (by putting it in the repository
+  implementation – which may still be provided by the application)
+* Still allows completely arbitrary edits on the metadata in question: now the
+  library does not need to anticipate what application wants to do and on the
+  other hand library can still provide e.g. snapshot functionality without
+  knowing about the application decisions mentioned in previous point.
+
+Other designs do not seem to manage both of these.
+
```
lukpueh, line 83: By reading the rendered markdown only, I wasn't sure whether you are describing two alternative designs in these diagrams, or different perspectives on the same design. This was reinforced by the preceding sentence "Other designs...".

Maybe adding a title, a la "How components are used" would do the trick.

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+![Design: How components are used](repository-library-design-usage.jpg)
+
+The core idea here is that because editing is ergonomic enough, when new
+functionality (like “developer uploads new targets”) is added, _it can be added
+at any level_: the application might add a `handle_new_target_files()` method
+that adds a bunch of targets into the metadata, but one of the previous layers
+could offer that as a helper function as well: code in both cases would look
+similar as it would use the common editing interface.
```
lukpueh, line 93: Shouldn't the design document make the decision on what layer such functionality should be implemented? Or at least explain based on what factors the decision should be made?

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+## Design
+
+![Design: Application and library components](repository-library-design-ownership.jpg)
+
+The design expects a fully functional repository application to contain code at
+three levels:
+* Repository library (abstract classes that are part of python-tuf)
+  * The Repository abstract class provides an ergonomic metadata editing API
+    for all code levels to use. It also implements some core edit actions like
+    snapshot update
```
lukpueh: I think it is important to be clear about what python-tuf actually implements and for what it provides abstract interfaces, e.g.

```suggestion
  * The Repository abstract class provides an ergonomic abstract metadata editing API
    for all code levels to use. It also provides concrete implementations for some core edit actions like
    snapshot update
```

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+    targets.add_key(“role1”, key)
+```
+
+This code loads current targets metadata for editing, adds the key to a role,
+and handles version and expiry bumps before persisting the new targets version.
+The reason for the context manager style is that it manages two things
+simultaneously:
+* Hides the complexity of loading and persisting metadata, and updating expiry
+  and versions from the editing code (by putting it in the repository
+  implementation – which may still be provided by the application)
```
lukpueh: If I understand your design correctly, this  **must** (not may) be provided by the application, does it?

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+|Application code | [cli.py (command line app)](https://github.com/vmware-labs/repository-editor-for-tuf/blob/main/tufrepo/cli.py), [keys_impl.py (keyring implementation)](https://github.com/vmware-labs/repository-editor-for-tuf/blob/main/tufrepo/keys_impl.py) |
+| Repository validation | [verifier.py (very rough, not intended for python-tuf)](https://github.com/vmware-labs/repository-editor-for-tuf/blob/main/tufrepo/verifier.py)
+| Improved Metadata editing | [helpers.py](https://github.com/vmware-labs/repository-editor-for-tuf/blob/main/tufrepo/helpers.py) 
+
+
+### Repository API
+
+Repository itself is a minimal abstract class: The value of this class is in
+defining the abstract method signatures (most importantly `_load`, `_save()`,
+`edit()`) that enable ergonomic metadata editing. The Repository class in this
```
lukpueh, line 140: Can `_load` and `_save` be seen as pre- and post-`edit` hooks? If that is correct, what is the benefit of prescribing hooks, if the thing that's being hooked into needs to be implemented by the users anyways? (apologies for not remembering/re-looking-up your implementation)

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+metadata is unlikely to do so.
+
+_Prototype status: A very rough implementation exists in verifier.py : this is
+unlikely to be very useful_
+
+### Improved metadata editing
+
+Currently the identified improvement areas are:
+* Metadata initialization: this could potentially be improved by adding
+  default argument values to Metadata API constructors
```
lukpueh, line 219: I see potential problems for our current deserialization implementation when providing default arguments, but we can discuss this separately.

rdimitrov:
I have a few minor formatting suggestions, but other than that everything's in line 👍 

* In `docs/adr/0010-repository-library-design.md`, it might be more explicit if code-related namings (such as repository_tool, _repo.py_ , etc) are also formatted in the form of `<text>`. 
* There are a few file names and examples mentioned (like `GitRepository`). I see there's a table with references at one place, but it'd be also nice if the text is an actual link so when clicked it redirects to that source file/repo.
* The diagrams are not so complicated, so as you pointed using something like https://asciiflow.com/ might make it easier to edit

lukpueh (commented):
Thanks for providing these documents, @jku! They should really help us move forward. The entire ADR-document and the "Design principles" in the Design document look perfectly fine to me.

Regarding the "Design"-section in the design document I must admit that I had to go through it several times to better understand. The "Details"-section definitely helps, but it was also not 100% obvious to me how your ideas/decision from the POC map to the overall design, especially in regards to what can/can't or should/shouldn't implemented on which level and why.

I made a few small suggestions that might help to clarify, but overall I'm still missing the preferably exhaustive list of specific functionality a minimal repository abstraction includes/leaves to application, and the underlying rationale.

docs/repository-library-design.md
jku: yeah, "generic" was meant in the sense that _"this code does not need to be part of the class implementing Repository interface"_...

docs/repository-library-design.md
jku: I'm proposing we start with the absolute minimum in the library: possibly just _snapshot_ and _sign_ implemented in the library itself. So the application is left with it's own design choices for the rest  (should `MyDerivedRepository` offer helper functions or should the application code just use the edit() contextmanager itself -- the proof of concept goes with the latter choice)... 

Nothing prevents us from moving some functionality into the library if that seems to make sense for multiple users but I don't see a reason to do that without seeing some implementations.

docs/repository-library-design.md
jku: I think that we should include almost nothing like this in python-tuf as a starting point: I don't think we know e.g. how applications want to add keys (_multiple keys at once? are they always going to do  some other edits at the same time?_), and shouldn't try to provide solutions before we do... I'm just saying that if this turns out to be something repositories do a lot, we could later offer it as a helper function

docs/repository-library-design.md
jku: yes that would be better phrasing. I've gone back and forth between mentioning that we could ship implementations in python-tuf source code: but as I think those would be examples and not library code, your phrasing is better

docs/repository-library-design.md
jku: yes they mostly can. `_load()` and `_save()` could be left as implementation details of the `edit()` implementation... but at least currently I've implemented `sign()` as 
```python
        md = self._load(role)
        self._save(role, md, clear_sigs=False)
```
in the abstract class -- I can't call `edit()` as it could modify the version, expiry, etc...  I think this is the main reason _load/_save are defined.

You may be right that we should figure out some alternative solution for sign(), and the _load/_save should be implementation details of `edit()` implementation

docs/repository-library-design.md
```diff
@@ -0,0 +1,194 @@
...
+  be interested in validating (after some edits) that the repository is correct
+  according to client workflow and that it contains the expected changes. The
+  Repository API should provide some validation, but we should recognise that
+  validation may be implementation specific.
+* Improved metadata editing: There are a small number of improvements that
+  could be made to metadata editing. These do not necessarily need to be part
+  of the repository API: they could be part of Metadata API as well
+
+It would make sense for python-tuf to ship with at least one concrete
+Repository implementation: possibly a repo.py look alike. This implementation
```
mnm678, line 116: It might be nice to have this example concrete implementation support a CLI for demos and such.

docs/repository-library-design.md
lukpueh: Thanks for the explanation, it makes a lot of sense! Would you mind trying to
1. emphasise on the process-like nature of this design proposal for python-tuf (*start with almost nothing, then aggregate common functionality from applications as they are implemented*) possibly in the beginning of the "Design" section, and 
2. rephrase this paragraph to show your preference in this example (*key-adding code for now left to application as we don't know common usage pattern... multiple keys at once?, other edits at the same time?, etc.)*

?

docs/repository-library-design.md
lukpueh: Again, with your explanation here the document makes more sense. (cf. my [reply above](https://github.com/theupdateframework/python-tuf/pull/1693/files/bcab2e96b03abd28aa180c7ad51b90e000fbc1dd#r760910439))

docs/repository-library-design.md
lukpueh: Would it be possible to explain the relationship/dependencies between abstract interfaces `_load/_save/edit` and concrete implementations `sign/snapshot` a little bit more in this paragraph?
IIUC, in order to use the concrete implementation, the application is required to first implement the abstract interfaces, right? I think that's worth mentioning, to better understand the whole "what's provided by python-tuf and what isn't"-thing. :)

docs/repository-library-design.md
jku: Yes. I didn't want to bloat these docs but I see a few potential implementations:
* some sort of CLI tool, to be able to generate and modify metadata for demo and testing purposes
* an example repository server (a "mini-warehouse"): a focused and opinionated repository example that could maybe include a custom web api for uploading new targets and a tiny developer tool to do the upload. It should be possible to make this into an example that runs out-of-the-box

Both of these could live inside or outside the repo... I'm not against the first one at all but I kind of feel the second one is even more important. The focus on CLI tools in TUF community (where every project seems to be a CLI tool) is not just a good thing IMO: it creates this impression that production repos can be shell scripts calling CLI tools**... which I don't think is such a great idea.

** _yes, I realize I've written exactly this myself: don't do as I do, do as I say :grimacing:_

docs/repository-library-design.md
jku: let's do that in #1459 (see https://github.com/theupdateframework/python-tuf/issues/1459#issuecomment-953729281)

Possibly some `from_dict()` implementations would have to be extra careful not to pass deserialized None/null values to a constructor.

docs/repository-library-design.md
MVrachev: Agree with the first suggestion.
To emphasize the first point will give more insight into the process of the design document creation which I think will be a valuable context.

docs/repository-library-design.md
jku: This makes sense, I'll add a separate section for this

jku:
I believe I've addressed the comments so far.

docs/repository-library-design.md
```diff
@@ -104,6 +116,28 @@ It would make sense for python-tuf to ship with at least one concrete
 Repository implementation: possibly a repo.py look alike. This implementation
 should not be part of the library but an example.
 
+### What a concrete Repository implementation must implement
+
+The current prototype API requires five methods that take care of access to
+metadata storage, expiry updates, version updates and signing.
+
+* **keyring()**: A property that returns the private key Mapping that should be
```
lukpueh:
```suggestion
* **keyring()**: A property that returns the private key mapping that should be
```
?

docs/repository-library-design.md
```diff
@@ -104,6 +116,28 @@ It would make sense for python-tuf to ship with at least one concrete
...
+  used for signing
+
+* **_load()**: Loads metadata from storage or cache. Is used by edit() and
+sign()
+
+* **_save()**: signs and persits metadata in cache/storage. Is used by edit()
```
lukpueh:
```suggestion
* **_save()**: Signs and persists metadata in cache/storage. Is used by edit()
```

docs/repository-library-design.md
```diff
@@ -104,6 +116,28 @@ It would make sense for python-tuf to ship with at least one concrete
...
+and sign()
+
+* **edit()**: The ContextManager that enables ergonomic metadata metadata
```
lukpueh:
```suggestion
* **edit()**: The ContextManager that enables ergonomic metadata
```

docs/repository-library-design.md
```diff
@@ -104,6 +116,28 @@ It would make sense for python-tuf to ship with at least one concrete
...
+### What a concrete Repository implementation must implement
+
+The current prototype API requires five methods that take care of access to
+metadata storage, expiry updates, version updates and signing.
+
+* **keyring()**: A property that returns the private key Mapping that should be
+  used for signing
+
+* **_load()**: Loads metadata from storage or cache. Is used by edit() and
+sign()
```
lukpueh: You reader only learns about `sign()` in the next section in the context of your POC implementation. Maybe this is confusing here in the more generic part of the document? Any reason to not move this entire list down to the Details section?

lukpueh (changes requested):
Thanks for addressing the comments, @jku! I think the added information is very helpful. I found some typos and have one general request.

docs/repository-library-design.md
jku: Hmm, this is all true, will change as suggested

jku:
Thanks: all suggestions make a lot of sense. I've moved the "abstract methods that must be implemented" part down to the Repository details section.

lukpueh (approved):
Thanks for bearing with me and my many comments, @jku!

PR #1695, opened 2021-11-24
Fix inconsistent behaviour of compute_metafile_hashes_length

### LINKED ISSUE #1651: RepositorySimulator: inconsistent behaviour of compute_metafile_hashes_length

`RepositorySimulator.compute_metafile_hashes_length` is documented as a boolean deciding " Whether to compute hashes and length for meta in snapshot/timestamp".

- When calculating snapshot metafile's hashes in `update_timestamp` and `compute_metafile_hashes_length` is switched to `False`, the old hashes and length  still remain.
- When calculating targets metafile's hashes in `update_snapshot` and `compute_metafile_hashes_length` is switched to `False`, the old hashes and length are cleaned up and the new values are set to None.

**Expected behavior**:
The correct behaviour should be the one in `update_snapshot` (clean up the hashes and length) and  it needs to be replicated in `update_timestamp`.

ivanayov:
Previously when `compute_metafile_hashes_length` was set to `False`
`update_timestamp` did not set the hash and length values to `None`
as expected. This change fixes that, so they are not `None` when
`compute_metafile_hashes_length=True` and `None` when
`compute_metafile_hashes_length=False`

Fixes #1651

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -294,12 +294,15 @@ class RepositorySimulator(FetcherInterface):
         """Update timestamp and assign snapshot version to snapshot_meta
         version.
         """
-        self.timestamp.snapshot_meta.version = self.snapshot.version
 
+        hashes = None
+        length = None
         if self.compute_metafile_hashes_length:
             hashes, length = self._compute_hashes_and_length(Snapshot.type)
-            self.timestamp.snapshot_meta.hashes = hashes
-            self.timestamp.snapshot_meta.length = length
+
+        self.timestamp.snapshot_meta = MetaFile(
+            self.snapshot.version, length, hashes
+        )
 
         self.timestamp.version += 1
 
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -434,6 +434,20 @@ class TestRefresh(unittest.TestCase):
 
         self._assert_files_exist([Root.type, Timestamp.type, Snapshot.type])
 
+    def test_compute_metafile_hashes_length(self) -> None:
+        self.sim.compute_metafile_hashes_length = True
+        self.sim.update_snapshot()
+        self._run_refresh()
+        self._assert_version_equals("timestamp", 2)
+        self._assert_version_equals("snapshot", 2)
+
+        self.sim.compute_metafile_hashes_length = False
+        self.sim.update_snapshot()
+        self._run_refresh()
+
+        self._assert_version_equals("timestamp", 3)
+        self._assert_version_equals("snapshot", 3)
+
 
 if __name__ == "__main__":
 


tests/repository_simulator.py
```diff
@@ -300,10 +300,12 @@ def update_timestamp(self):
...
         self.timestamp.snapshot_meta.version = self.snapshot.version
 
+        hashes = None
+        length = None
         if self.compute_metafile_hashes_length:
             hashes, length = self._compute_hashes_and_length("snapshot")
-            self.timestamp.snapshot_meta.hashes = hashes
-            self.timestamp.snapshot_meta.length = length
+        self.timestamp.snapshot_meta.hashes = hashes
+        self.timestamp.snapshot_meta.length = length
```
sechkova, line 306: This implementation is correct and  it does fix the bug but looking at `update_snapshot`, what do you think about replacing these three lines
```
        self.timestamp.snapshot_meta.version = self.snapshot.version
        self.timestamp.snapshot_meta.hashes = hashes
        self.timestamp.snapshot_meta.length = length
```
by directly creating a new `MetaFile` object instead?

tests/test_updater_top_level_update.py
```diff
@@ -434,6 +434,27 @@ def test_new_targets_expired(self) -> None:
 
         self._assert_files_exist(["root", "timestamp", "snapshot"])
 
+    def test_compute_metafile_hashes_length(self) -> None:
```
sechkova, line 437: This test is actually testing `RepositorySimulator`, something that we have not done up to now. I don't want to say a test is not needed but @jku any ideas where to place it?

jku: yeah should not add simulator tests in updater test files at least.

I think you could make this into an updater test that just happens to test the simulator functionality as well? I suppose running `_run_refresh()` after each update would do it right? that should fail with current code but should work with the PR, am I reading this correctly? Then you wouldn't even need the asserts for meta content -- those would be implied...

Also, I think the testing update_snapshot() in the two modes is probably enough -- that will call update_timestamp() for you anyway.

ivanayov: @jku `_run_refresh()` is already run after all updates, but the tests don't fail because of that. 
`update_snapshot` is not enough, as it sets the hashes and length to `None` and they are not updated in `update_timestamp` after as `compute_metafile_hashes_length` is `False`, so cannot reproduce the problem with it only.
May be I should think of moving the test to let's say `TestSimulator`?

jku: > _run_refresh() is already run after all updates, ...

Maybe we are not talking about the same thing: I'm looking at the test you are adding and it never runs the updater at all

> but the tests don't fail because of that.

I'm assuming that if you add `_run_refresh()` calls into your test (after each update), you can make the tests fail if you run without the fix.

If that does not happen, then I need to re-evaluate this as I must have misunderstood the problem.

jku: oh, `update_timestamp()` alone likely does not trigger the updater failure: if you just update the timestamp then the hashes/length that are mistakenly left in snapshot_meta happen to be correct (because snapshot hashes/length have not changed).

So you have to run `update_snapshot()` so that both snapshot and timestamp get a new version (and now the snapshot_meta might contain old hashes/length if the bug is not fixed).

I was going to suggest running `python3 test_updater_top_level_update.py --dump TestRefresh.test_compute_metafile_hashes_length ` after you add some refresh calls to see the actual repository metadata during the test... but apparently this test file does not support `--dump` yet :|

ivanayov: @jku with this update in the test `RepositoryError` is raised without the fix and no error is raised with the fix. When I remove the fix, if I assert the `RepositoryError` before refresh, then `assertIsNone` fails, which means the hashes value is not correct, as we already know.
Unfortunately the library doesn't support `assertNotRaises`, so I cannot do a clean verification and once the bug is reproduced again an error would be thrown, rather than a test failure. But it works fine this way and is kind of verification though.

sechkova: I think we can at least check if the snapshot and timestamp versions are the expected similar to the other tests. Not sure if it should be done after each refresh or only once at the end is enough.
```
self._assert_version_equals("timestamp", 3)
self._assert_version_equals("snapshot", 3)
```

tests/test_updater_top_level_update.py
```diff
@@ -434,6 +434,18 @@ def test_new_targets_expired(self) -> None:
...
+        self.sim.compute_metafile_hashes_length = True
+        self.sim.update_snapshot()
+        self._run_refresh()
+
+        self.sim.compute_metafile_hashes_length = False
+        self.sim.update_snapshot()
+        self._run_refresh()
+
+        self.assertIsNone(self.sim.timestamp.snapshot_meta.hashes)
+        self.assertIsNone(self.sim.timestamp.snapshot_meta.length)
```
sechkova: If we assert about `sim.timestamp.snapshot_meta`, should we do it for `sim.snapshot.meta["targets.json"]` too?
Or if we consider it a regression test for this particular bug should we add 

```
self.assertIsNotNone(self.sim.timestamp.snapshot_meta.hashes)
self.assertIsNotNone(self.sim.timestamp.snapshot_meta.length)
```
after we switch  self.sim.compute_metafile_hashes_length to True?

sechkova (changes requested):

tests/test_updater_top_level_update.py
jku: I don't think we necessarily need to assert anything about simulator -- I get that this was fixing a bug in the simulator but the simulator is still just a test utility and we know that this updater test happens to regression test that bug as well.

Sure, we could write a test suite for simulator itself but I don't think think it's necessary

tests/test_updater_top_level_update.py
jku: > Unfortunately the library doesn't support assertNotRaises

"assertNotRaises" is kind of on by default: raising inside a test is a valid way of failing the test

tests/test_updater_top_level_update.py
sechkova: @ivanayov I guess I agree with Jussi. Can you remove the two asserts above and I think we'll be all set.

ivanayov: @sechkova just addressed your comment

sechkova (approved):

jku (approved):
Thanks, LGTM

PR #1626, opened 2021-10-19
Modernise packaging configuration

joshuagl:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

Fixes #1161 

A grab-bag of modernisation in our packaging

**Description of the changes being introduced by the pull request**:
* Move from invoking setup.py to build dists to instead using the `build` tool (adds `pyproject.toml`)
* Move setuptools metadata from the dynamic setup.py file to setup.cfg (removes `setup.py` completely)
* Add build and release dependencies to requirements-dev (build, wheel and twine packages)
* Include documentation link in project_urls
* update docs/RELEASE.md for packaging changes

**Note**: our sdist is effectively everything except the very few CI related files (.github/*, .fossa.yml, .readthedocs.yaml) – we could prune this back to be only the source we need to produce the wheel, but that is probably later work after we've cleaned up the documentation we keep in the repo (see #808)

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #1699, opened 2021-11-30
Move linters configurations in pyproject.toml

### LINKED ISSUE #1684: Configure black and isort only once in pyproject.toml

#1626 added a `pyproject.toml`, which can be used to configure code formatters black and isort in one place only. Let's put the config parameters (line length, which files to include, etc) there.

See corresponding docs
- https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file
- https://pycqa.github.io/isort/docs/configuration/config_files.html

**Current behavior**:
Repeated black and isort config in:
- [Contributor Docs](https://github.com/theupdateframework/python-tuf/blob/develop/docs/CONTRIBUTORS.rst#auto-formatting)
- [tox (CI)](https://github.com/theupdateframework/python-tuf/blob/bb15ecfb994290e723dd69fc728a2c7bd9201811/tox.ini#L44-L45)
- [Local .pre-commit hooks](https://github.com/theupdateframework/python-tuf/blob/be0cef067cecd566d60af96c15dc2e48bc4467af/.pre-commit-config.yaml#L11-L23) (NOTE: This is not part of the repo, but I have one locally, also see [contributor doc recommendation](https://github.com/theupdateframework/python-tuf/blame/develop/docs/CONTRIBUTORS.rst#L123-L125))
  
  

**Expected behavior**:
Single source of truth for black and isort config in `pyproject.toml`

MVrachev:
Fixes #1684

In pr #1626 a `pyproject.toml` was added, which can be used to configure code formatters all of our
code formatters in one place only. It's good if we use one file as a source for truth for all linter
configurations.

I was able to move all linter configurations for `pylint` and `mypy` inside `pyproject.toml`,
but I faced difficulties with `black` and `isort`.
I explained what problems I faced moving some of the arguments in those comments:
- `black` target folders: https://github.com/theupdateframework/python-tuf/issues/1684#issuecomment-981859253
- `isort` target folders: https://github.com/theupdateframework/python-tuf/issues/1684#issuecomment-982655761
- `isort` options `--check` and `--dff`: https://github.com/theupdateframework/python-tuf/issues/1684#issuecomment-982695656

I decided to call `black` and `isort` without specifying the config file as they automatically find `pyproject.toml`
as it's at the root of the project.
It's not the same situation for `pylint`.

requirements**:


MVrachev:
@lukpueh will appreciate a look from you as this is a TODO item added by you 9 months ago :D

pyproject.toml
```diff
@@ -1,3 +1,90 @@
+# Build-system section
 [build-system]
 requires = ["setuptools>=40.8.0", "wheel"]
 build-backend = "setuptools.build_meta"
+
+# Black section
+[tool.black]
+line-length=80
+check="True"
+diff="True"
```
lukpueh: IMO the default for black should be to actually re-format. At least, that's what I would expect when copying the command from the Contributors docs, or using it in a precommit hook. So I suggest to  remove the `check` and `diff` options here and supply them in tox.ini only, as you do for `isort`.

pyproject.toml
```diff
@@ -1,3 +1,90 @@
...
+disallow_untyped_defs = "True"
+disallow_untyped_calls = "True"
+show_error_codes = "True"
+files = [
+  "tuf/api/",
+  "tuf/ngclient",
+  "tuf/exceptions.py"
+]
+
+[[tool.mypy.overrides]]
```
lukpueh, line 85: Why double brackets?

pyproject.toml
```diff
@@ -1,3 +1,90 @@
...
+method-rgx="""(?x)^(?:(?P<exempt>_[a-z0-9_]+__|runTest|setUp|tearDown|setUpTestCase|tearDownTestCase|setupSelf|tearDownClass|setUpClass|(test|assert)_*[A-Z0-9][a-zA-Z0-9_]*|next)|(?P<camel_case>_{0,2}[A-Z][a-zA-Z0-9_]*)|(?P<snake_case>_{0,2}[a-z][a-z0-9_]*))$"""
+argument-rgx="""^[a-z][a-z0-9_]*$"""
+attr-rgx="""^_{0,2}[a-z][a-z0-9_]*$"""
+class-attribute-rgx="""^(_?[A-Z][A-Z0-9_]*|__[a-z0-9_]+__|_?[a-z][a-z0-9_]*)$"""
+class-rgx="^_?[A-Z][a-zA-Z0-9]*$"
+const-rgx="^(_?[A-Z][A-Z0-9_]*|__[a-z0-9_]+__|_?[a-z][a-z0-9_]*)$"
+inlinevar-rgx="^[a-z][a-z0-9_]*$"
+module-rgx="^(_?[a-z][a-z0-9_]*|__init__)$"
+no-docstring-rgx="(__.*__|main|test.*|.*test|.*Test)$"
+variable-rgx="^[a-z][a-z0-9_]*$"
```
lukpueh, line 56: Why are some regexes in triple-quotes and others not? `"""rgx"""` vs `"rgx"`?

tox.ini
```diff
@@ -40,10 +40,9 @@ commands =
 changedir = {toxinidir}
 commands =
     # Use different configs for new (tuf/api/*) and legacy code
-    # TODO: configure black and isort args in pyproject.toml (see #1161)
-    black --check --diff --line-length 80 tuf/api tuf/ngclient
-    isort --check --diff --line-length 80 --profile black -p tuf tuf/api tuf/ngclient
-    pylint -j 0 tuf/api tuf/ngclient --rcfile=tuf/api/pylintrc
+    black tuf/api tuf/ngclient
+    isort --check --diff tuf/api tuf/ngclient
+    pylint -j 0 tuf/api tuf/ngclient --rcfile=pyproject.toml
```
lukpueh, line 45: Let's also format and lint the scripts in the newly added `examples` directory. (see #1685)

docs/CONTRIBUTORS.rst
```diff
@@ -113,9 +113,8 @@ CI/CD will check that new TUF code is formatted with `black
 Auto-formatting can be done on the command line:
 ::
 
-    $ # TODO: configure black and isort args in pyproject.toml (see #1161)
-    $ black --line-length 80 tuf/api
-    $ isort --line-length 80 --profile black -p tuf tuf/api
+    $ black tuf/api tuf/ngclient
+    $ isort tuf/api tuf/ngclient
```
lukpueh: Let's also format and lint the scripts in the newly added `examples` directory. (see #1685)

lukpueh (changes requested):
Thanks for working on this @MVrachev. I very briefly took a stab at it before writing #1684, but got too annoyed with the configuration issues you described. So, I all the more appreciate that you followed through! 👍

pyproject.toml
MVrachev: Yes, it makes sense.
Will remove them from `pyproject.toml` and move them back in `tox.ini`.

docs/CONTRIBUTORS.rst
jku: could just do this to avoid the maintenance here:

```suggestion
    $ black <filename>
    $ isort <filename>
```

lukpueh: The whole `black`/`isort` file-based case handling should hopefully soon be irrelevant, when we remove the legacy code.

pyproject.toml
MVrachev: That was suggested by the `mypy` pyproject.toml configuration file documentation [here](https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file).
I tried running it without one of the bracket pairs and got the error: 

```pyproject.toml: tool.mypy.overrides sections must be an array. Please make sure you are using double brackets like so: [[tool.mypy.overrides]]```

File: pyproject.toml (hunk 3 above)
Comment (MVrachev) [reply pyproject.toml:56]: A good catch. I copied the regex content from `tuf/api/pylintrc` and made the mistake of not reading it.
Fixing it.

Comment (lukpueh) [reply pyproject.toml:56]: But in `tuf/api/pylintrc` there are no quotes at all in these lines. (I actually compared the lines in both files while reviewing and the only difference are the quotes.)

File: tox.ini (hunk 4 above)
Comment (MVrachev) [reply tox.ini:45]: Will add a separate pr for that as I noticed there are a couple of mypy warnings that must be addressed as well.

File: pyproject.toml (hunk 3 above)
Comment (MVrachev) [reply pyproject.toml:56]: Aaa, I think I wanted to split them multiline but decided against it.
No matter what was the reason it's fixed now.

File: tox.ini (hunk 4 above)
Comment (lukpueh) [reply tox.ini:45]: Oh right, we already have a ticket for that in #1697

Comment (MVrachev) [reply tox.ini:45]: After we agree that this pr is good, I will submit a pr for #1697.

**Review by lukpueh — APPROVED** [review]

**Comment by MVrachev** [comment]
I had to force-push because I linked the `black` configuration documentation in the `build` section instead of the `black` section.
No other changes were made.

File: pyproject.toml [hunk 6]
```diff
@@ -1,3 +1,90 @@
...
+warn_unreachable = "True"
+strict_equality = "True"
+disallow_untyped_defs = "True"
+disallow_untyped_calls = "True"
+show_error_codes = "True"
+files = [
+  "tuf/api/",
+  "tuf/ngclient",
+  "tuf/exceptions.py"
+]
```
jku, line 83: these make sense here so you can just run `mypy`... but I wonder if maintenance would be easier if we gave these on the command line in tox.ini like all other tools -- then we could define a `lint_dirs` variable in there once and use it for all tools (mypy would need a exceptions.py added until we move the file to api/ but that's fine)

jku (approved):
LGTM, with one suggestion that could be future work

pyproject.toml
lukpueh: Generally a good idea! But let's hold off this future work until we've removed legacy code, shall we? Maybe we won't need a `lint_dirs` then.

PR #1700, opened 2021-12-01
doc:  basic hash bin delegation repo example + test

### LINKED ISSUE #1673: Docs: Add repository tutorial based on metadata API

lukpueh:
Fixes #1673 (together with #1685)

As 'repository_tool' and 'repository_lib' are being deprecated, hash bin delegation interfaces are no longer available in this    implementation. The example code in this file demonstrates how to  easily implement those interfaces, and how to use them together with the TUF metadata API, to perform hash bin delegation.    

Note, the hash bin delegation logic in this example is largely  copied from 'repository_{lib, tool}', but modernized and simplified  for this purpose.

**Notes to reviewers**

- I sneaked in a tiny unrelated docfix (954c159e1ece42a3b917d3a71f7d5ce3171140e7). If someone is appalled by this breach of pull request protocol, I'm *happy* to move it to a separate one. :P
- I would appreciate simplification-suggestions regarding my probably sometimes long-winded phrasing in the explanatory texts/comments of hashed_bin_delegation.py. (@jhdalek55 to the rescue!)
- Besides purely technical linguistic help, I'd appreciate an assessment whether the example needs more/less/other explanations about TUF/Metadata API/Hash bin delegation, etc..

Thanks!!

requirements**:

### DIFF
--- a/examples/repo_example/hashed_bin_delegation.py
+++ b/examples/repo_example/hashed_bin_delegation.py
@@ -49,38 +49,38 @@ keys: Dict[str, Dict[str, Any]] = {}
 
 # Hash bin delegation
 # ===================
-# Hash bin delegation allows to automatically distribute a large number of
-# target files over a fixed number of targets metadata, in order to reduce the
-# size of the specific targets metadata a client needs to download for a given
-# target file, and thus reduce the metadata network overhead caused by TUF.
+# Hash bin delegation allows to distribute a large number of target files over
+# multiple delegated targets metadata. The consequence is smaller metadata
+# files and thus a lower network overhead for repository-client communication.
 #
-# It is achieved by uniformly delegating the responsibility for target files
-# based on the leading digits of the hash of their file path, where each
-# delegated targets role is responsible for an incremental range of target path
-# hash prefixes.
+# The assignment of target files to targets metadata is done automatically,
+# based on the hash of the target file name. More precisely, only a prefix of
+# the target file name hash is needed to assign it to the correct hash bin.
 #
-# The only number that needs to be configured is the number of bins, everything
-# else is derived using the mathematical operations outlined below. An
-# appropriate number of bins depends on the expected number of target files in
-# a repository. For the purpose of this example we choose...
-
-# The fixed number of bins determines the length of any considered hash prefix,
-# how many prefixes exist in total, and how many prefixes fall into each bin.
-NUMBER_OF_BINS = 32  # power of 2 for even distribution of hash prefixes
-
-# The available digits in the hexadecimal representation of the number of bins
-# (minus one, counting starts at zero) determines the length of any hash prefix,
-# i.e. how many left digits need to be considered to assign the hash to a bin.
-PREFIX_LEN = len(f"{NUMBER_OF_BINS - 1:x}")  # 2
-
-# The total number of distinct hash prefixes is determined by the highest
-# hexadecimal number that can be represented with the given number of digits.
-NUMBER_OF_PREFIXES = 16 ** PREFIX_LEN  # 256
-
-# If the number of bins is a power of two, all hash prefixes can be evenly
-# distributed over all bins and thus allow us to compute the size of any bin.
-BIN_SIZE = NUMBER_OF_PREFIXES // NUMBER_OF_BINS  # 8
-
+# The number of bins is the only number that needs to be configured. Everything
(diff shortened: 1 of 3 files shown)


jku (approved):
LGTM, I don't have any specific comments on this.

In the future this definitely looks like something that would make sense to package with the repository API somehow:
* manage a "bin container" Targets automatically so repository API user can just "add a new target to bin container" (and in reality target is added to the correct bin)
* figure out how to handle NUMBER_OF_BINS changes (at least provide advice on when and how to handle growth in targets numbers) and other maintenance things.

... but like a lot of repository functionality it looks like something I'd really rather see experimented in an implementation before we add it to any library code

lukpueh:
> figure out how to handle NUMBER_OF_BINS changes

I had the same thought, while revising (and trying to re-understand) the binning logic. In reality a user does not care about the exact NUMBER_OF_BINS, but rather about an "appropriate" distribution of target files over bins. So it probably makes more sense to provide an interface, where the user specifies the expected number of target files (or corresponding classes of target numbers).

examples/repo_example/hashed_bin_delegation.py
```diff
@@ -0,0 +1,234 @@
...
+# a repository. For the purpose of this example we choose...
+
+# The fixed number of bins determines the length of any considered hash prefix,
+# how many prefixes exist in total, and how many prefixes fall into each bin.
+NUMBER_OF_BINS = 32  # power of 2 for even distribution of hash prefixes
+
+# The available digits in the hexadecimal representation of the number of bins
+# (minus one, counting starts at zero) determines the length of any hash prefix,
+# i.e. how many left digits need to be considered to assign the hash to a bin.
+PREFIX_LEN = len(f"{NUMBER_OF_BINS - 1:x}")  # 2
```
MVrachev: Maybe add a simple word or comment on what `1:x` does.
I had to find it online and because I hadn't used it, it wasn't trivial to me.
In that context, it could be useful to explain somewhere why we use hexadecimal strings.

lukpueh: Thanks for the comment, @MVrachev! I just pushed a commit trying to make this clearer. Let me know what you think.

examples/repo_example/hashed_bin_delegation.py
```diff
@@ -0,0 +1,233 @@
...
+#
+# The right number of bins depends on the expected number of target files in a
+# repository. For the purpose of this example we choose ...
+NUMBER_OF_BINS = 32  # ..., which determines the length of any hash prefix
+# considered for bin assignment (PREFIX_LEN), how many hash prefixes are
+# covered by all bins (NUMBER_OF_PREFIXES), and how many prefixes are covered
+# by each individual bin (BIN_SIZE):
+#
+# The prefix length is the number of digits in the hexadecimal representation
+# (see ':x') of the number of bins minus one (counting starts at zero), i.e. ...
```
MVrachev:
```suggestion
# (see "x" integer representation from https://docs.python.org/3/library/string.html#module-string)
of the number of bins minus one (counting starts at zero), i.e. ...
```

MVrachev (approved):
Only one small suggestion.
Other than that as a reader who has never read anything specific about hash bin delegations
it seems understandable.
LGTM!

examples/repo_example/hashed_bin_delegation.py
lukpueh: Thanks, Martin! I included a slightly modified version of this suggestion. Merging when CI/CD returns...

PR #1701, opened 2021-12-01
Start linting the examples folder

### LINKED ISSUE #1697: examples: run mypy on repo example
repository example is not completely annotated. mypy config (currently in setup.cfg) does not cover the example: it should cover the whole `examples/` directory. 

It looks like this means some minor annotation fixes in the example (mostly to deal with `roles` dictionary items).

MVrachev:
Fixes #1697

The examples folder currently contains a repository example and it's
good if we start linting its content and as a result add type
annotations.

I run all of the linters on the example folder instead of only using `mypy`
as I thought that will be better for the code included there.

requirements**:

### DIFF
--- a/examples/repo_example/basic_repo.py
+++ b/examples/repo_example/basic_repo.py
@@ -28,6 +28,7 @@ import tempfile
 from collections import OrderedDict
 from datetime import datetime, timedelta
 from pathlib import Path
+from typing import Any, Dict
 
 from securesystemslib.keys import generate_ed25519_key
 from securesystemslib.signer import SSlibSigner
@@ -48,7 +49,7 @@ from tuf.api.metadata import (
 from tuf.api.serialization.json import JSONSerializer
 
 
-def _in(days):
+def _in(days: float) -> datetime:
     """Adds 'days' to now and returns datetime object w/o microseconds."""
     return datetime.utcnow().replace(microsecond=0) + timedelta(days=days)
 
@@ -89,8 +90,8 @@ SPEC_VERSION = "1.0.19"
 
 # Define containers for role objects and cryptographic keys created below. This
 # allows us to sign and write metadata in a batch more easily.
-roles = {}
-keys = {}
+roles: Dict[str, Metadata] = {}
+keys: Dict[str, Dict[str, Any]] = {}
 
 
 # Targets (integrity)
@@ -117,7 +118,7 @@ roles["targets"] = Metadata[Targets](
 local_path = Path(__file__).resolve()
 target_path = f"{local_path.parts[-2]}/{local_path.parts[-1]}"
 
-target_file_info = TargetFile.from_file(target_path, local_path)
+target_file_info = TargetFile.from_file(target_path, str(local_path))
 roles["targets"].signed.targets[target_path] = target_file_info
 
 # Snapshot (consistency)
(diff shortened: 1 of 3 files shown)


examples/repo_example/basic_repo.py
```diff
@@ -332,7 +333,7 @@ def _in(days):
 del roles["targets"].signed.targets[target_path]
 
 # Increase expiry (delegators should be less volatile)
-roles["targets"].expires = _in(365)
+roles["targets"].signed.expires = _in(365)
```
lukpueh, line 336: 👀 Whoopsi, thanks for catching this! Did mypy detect it?

examples/repo_example/basic_repo.py
```diff
@@ -117,7 +118,7 @@ def _in(days):
 local_path = Path(__file__).resolve()
 target_path = f"{local_path.parts[-2]}/{local_path.parts[-1]}"
 
-target_file_info = TargetFile.from_file(target_path, local_path)
+target_file_info = TargetFile.from_file(target_path, str(local_path))
```
lukpueh, line 121: Was this also flagged by mypy?

(Also it is odd, because I remember already casting this to `str` at some point because `TargetFile.from_file` raised an error with local_path being a `Path` object. But I can't find it in #1685. I wonder were it went... Anyways, good that you/mypy fixed it.)

pyproject.toml
```diff
@@ -79,7 +79,8 @@ show_error_codes = "True"
 files = [
   "tuf/api/",
   "tuf/ngclient",
-  "tuf/exceptions.py"
+  "tuf/exceptions.py",
+  "examples/",
 ]
```
jku: left a comment about this in the other bug: https://github.com/theupdateframework/python-tuf/pull/1699#discussion_r760934677 (we could remove these and just define the "linter dirs" in one place in tox.ini to make maintenance easier)

examples/repo_example/basic_repo.py
```diff
@@ -89,8 +90,8 @@ def _in(days):
 
 # Define containers for role objects and cryptographic keys created below. This
 # allows us to sign and write metadata in a batch more easily.
-roles = {}
-keys = {}
+roles: Dict[str, Metadata] = {}
+keys: Dict[str, Dict[str, Any]] = {}
```
lukpueh, line 94: Quick question, what is the criterion for a variable name to get type annotated?

lukpueh (commented):
Thanks for the PR, @MVrachev! From a quick glance it looks good to me! I'll take another look when it no longer has the draft status, which I think you can remove now that #1699 has been merged.

examples/repo_example/basic_repo.py
MVrachev: Yes, it did.

examples/repo_example/basic_repo.py
jku: it's needed if mypy can't figure it out from the context (and mypy is not very advanced in this). So in this case `roles = { "root": root_md }` would have been fine but an empty container needs typing

MVrachev: I think it does need annotation when you use containers and personally, I think it's better to define annotations in these cases. It makes it so much easier to understand how the container should be used or is used later.

examples/repo_example/basic_repo.py
MVrachev: Yes, `mypy` noticed that.
Yet another argument as to why we do need linting quickly as possible on new code.

MVrachev:
I rebased on top of develop and created a commit adding a `lint_dirs` variable as proposed by Jussi.
@lukpueh is right that this list will be simplified after we drop the old code base, but still, it will be useful to have
a `lint_dirs` variable even if the list is simply `./`

This pr is ready for review.

lukpueh (approved):
Works like a charm!

PR #1702, opened 2021-12-01
Sanitize repo simulator role names when writing

MVrachev:

Sanitize the role names in the repo simulator when dumping in a
directory the same way the ngclient does.
That's necessary because when testing fishy role names from `test_updater_with_simulator` leads to an
error:
`PermissionError: [Errno 13] Permission denied: '/.json'`

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -397,5 +397,6 @@ class RepositorySimulator(FetcherInterface):
                 f.write(self.fetch_metadata(role))
 
         for role in self.md_delegates:
-            with open(os.path.join(dest_dir, f"{role}.json"), "wb") as f:
+            quoted_role = parse.quote(role, "")
+            with open(os.path.join(dest_dir, f"{quoted_role}.json"), "wb") as f:
                 f.write(self.fetch_metadata(role))


jku:
alternatively could sanitize the delegated rolename when writing the file like client does (`quoted_role = parse.quote(role, "")`)

sechkova:
Well it just happened that I did the same thing yesterday: dc908eee52dcf2b66dcd38eb54c972a2b3e17ec2
(without the part in test_updater_with_simulator.py)

MVrachev:
> Well it just happened that I did the same thing yesterday: dc908ee
(without the part in test_updater_with_simulator.py)

@sechkova I am sorry it happened that way, I haven't read #1689 and didn't notice that. 
I will drop this commit and leave only the commit related to the error.

sechkova:
Bad timing on both sides 🤷 Can you have a look at my commit before dropping yours? Maybe I've missed something.

MVrachev:
> Bad timing on both sides 🤷 Can you have a look at my commit before dropping yours? Maybe I've missed something.

I just had a look and yours seems to cover what I did here, so don't worry.

MVrachev:
I updated the code by using what @jku suggested.
Additionally, the commit and pr descriptions were changed together with the pr title.

jku (changes requested):
This commit does not do what the message says.

* It url encodes role names that do not need sanitizing (but does not url encode the delegated role names that could need it)
* Then it skips dumping metadata for roles that actually would need sanitizing if they were used as filenames?

MVrachev:
> This commit does not do what the message says.
> 
> * It url encodes role names that do not need sanitizing (but does not url encode the delegated role names that could need it)
> * Then it skips dumping metadata for roles that actually would need sanitizing if they were used as filenames?

I changed the code to dump the delegated roles and removed the code where I skipped dumping.

jku (approved):

PR #1703, opened 2021-12-02
Test loading of cached metadata in ngclient

### LINKED ISSUE #1681: ngtests: test loading of cached metadata
Taken from @jku's [comment](https://github.com/theupdateframework/python-tuf/pull/1636#issuecomment-953669058) that summarizes it very well,  tests are needed which:

- check that expected metadata is read from cache the expected number of times
     - wrapping open() should work fine here (see tests in #1593)
-  check if cached metadata is used to verify new metadata downloaded from remote
      - trickier but I suppose we can rely on negative tests here: tests that are expected to fail on that verification

ivanayov:
After making a successful update of valid metadata which stores it
in cache and performing a second update with a new updater while
the metadata is already stored in cache, this test verifies that
timestamp, snaphot and targets are loaded from cache and not 
downloaded

Addresses #1681

Signed-off-by: Ivana Atanasova <iyovcheva@vmware.com>

### DIFF
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -12,7 +12,7 @@ import tempfile
 import unittest
 from datetime import datetime, timedelta
 from typing import Iterable, Optional
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, call, patch
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
@@ -635,6 +635,50 @@ class TestRefresh(unittest.TestCase):
         # the snapshot.meta["targets.json"] version by 1 and throws an error.
         with self.assertRaises(BadVersionNumberError):
             self._run_refresh()
+
+    @patch.object(builtins, "open", wraps=builtins.open)
+    def test_load_metadata_from_cache(self, wrapped_open: MagicMock) -> None:
+
+        # Add new delegated targets
+        spec_version = ".".join(SPECIFICATION_VERSION)
+        targets = Targets(1, spec_version, self.sim.safe_expiry, {}, None)
+        role = DelegatedRole("role1", [], 1, False, ["*"], None)
+        self.sim.add_delegation("targets", role, targets)
+        self.sim.update_snapshot()
+
+        # Make a successful update of valid metadata which stores it in cache
+        updater = self._run_refresh()
+        updater.get_targetinfo("non_existent_target")
+
+        # Clean up calls to open during refresh()
+        wrapped_open.reset_mock()
+        # Clean up fetch tracker metadata
+        self.sim.fetch_tracker.metadata.clear()
+
+        # Create a new updater and perform a second update while
+        # the metadata is already stored in cache (metadata dir)
+        updater = Updater(
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            self.targets_dir,
+            "https://example.com/targets/",
+            self.sim,
+        )
+        updater.get_targetinfo("non_existent_target")
+
+        # Test that metadata is loaded from cache and not downloaded
+        wrapped_open.assert_has_calls(
+            [
+                call(os.path.join(self.metadata_dir, "root.json"), "rb"),
+                call(os.path.join(self.metadata_dir, "timestamp.json"), "rb"),
+                call(os.path.join(self.metadata_dir, "snapshot.json"), "rb"),
+                call(os.path.join(self.metadata_dir, "targets.json"), "rb"),
+                call(os.path.join(self.metadata_dir, "role1.json"), "rb"),
+            ]
+        )
+
+        expected_calls = [("root", 2), ("timestamp", None)]
+        self.assertListEqual(self.sim.fetch_tracker.metadata, expected_calls)
 
 
 if __name__ == "__main__":


kairoaraujo:
Hi @ivanayov, I think my PR #1704  will help you to remove the mock part. You will be able to test it directly against the Repository Simulator.

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +257,42 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+                "https://example.com/metadata/",
+                self.targets_dir,
+                "https://example.com/targets/",
+                self.sim,
+            )
+            updater.get_targetinfo(self.targets_dir)
+
+            # Test that snaphot and targets are loaded from cache and not downloaded
+            # and that timestamp is loaded from cache and used to verify the new timestamp
+            wrapped_fetch.assert_has_calls([call("root", 2), call("timestamp", None)])
```
jku: testing this has some merit... but I bet we could add this test to `test_top_level_roles_update()`and `test_delegated_roles_update()`:  existing tests that already test whether the expected files are being fetched.

However, I would like to _actually verify what we load from disk_. The PR currently does not do that. So my (possibly naive) ideas was to patch the builtin open() and see what gets loaded from disk during `Updater()`, `refresh()` and `get_targetinfo()`

ivanayov:
@jku I addressed your comment, if something is not very correct I'll edit it. (It doesn't cover if the loaded timestamp is used to verify the new timestamp, but AFAIR it's already covered and we wanted to check only what's loaded from disk.)

Thanks @kairoaraujo . With the latest change I don't check `fetch` anymore, but will keep in mind for future tests.

MVrachev:
Could you please rebase on top of develop?
In #1710 we started linting our tests and it will be good to lint the changes made in this pr.

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +257,44 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+        self.sim.add_target("targets", b'some content', self.targets_dir)
+        self.sim.update_snapshot()
+
+        # Make a successful update of valid metadata which stores it in cache
+        updater = self._run_refresh()
+
+        # Clean up calls to fetch during refresh()
+        wrapped_open.reset_mock()
+
+        updater.get_targetinfo(self.targets_dir)
```
jku: I think this call does not need to be here (and the argument is weird)

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +257,44 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+        # Create a new updater and perform a second update while
+        # the metadata is already stored in cache (metadata dir)
+        updater = Updater(
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            self.targets_dir,
+            "https://example.com/targets/",
+            self.sim,
+        )
+        updater.get_targetinfo(self.targets_dir)
```
jku: Calling `get_targetinfo()` feels correct but the argument is confusing: maybe just use `"non_existent_target"` or something like that as target_path -- the value does not matter but using `targets_dir` is misleading.

jku (changes requested):
Yes, this looks ok. Left some comments on get_targetinfo() calls: please fix those.

Since you are already modifying I'll leave another nitpick: can you double check that comments and commit messages do not claim to do something that isn't done: at least one comment says _"Test that timestamp, snaphot and targets are ... not downloaded"_ which is not true. The test only checks that local files are opened in the expected way (and that's fine, this is useful and would have caught a past issue we had).

There is a potential follow-up for 
* adding a delegated target to repository (se the test can check that delegated target is also correctly loaded from local disk)
* separately testing constructor, refresh() and get_targetinfo()

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +257,44 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
         updater.get_targetinfo("somepath")
         wrapped_open.assert_not_called()
 
+    @patch.object(builtins, "open", wraps=builtins.open)
+    def test_update_metadata_to_cache(self, wrapped_open: MagicMock) -> None:
```
sechkova: `test_load_metadata_from_cache` or something like it would match better what this test is trying to achieve.

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +257,44 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+        updater = self._run_refresh()
+
+        # Add targets to repository
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", b'some content', self.targets_dir)
+        self.sim.update_snapshot()
```
sechkova: Adding target files is not needed at all for this test. Target files, at least in this case, are unrelated to metadata files. Adding a delegated role is what would be useful so that delegated metadata is tested too.

tests/test_updater_with_simulator.py
sechkova: +1. This call only makes sense if a delegated role was added to the repository.

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +257,44 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+
+        # Test that timestamp, snaphot and targets
+        # are loaded from cache and not downloaded
+        wrapped_open.assert_has_calls([
+            call(os.path.join(self.metadata_dir, "root.json"), "rb"),
+            call(os.path.join(self.metadata_dir, "timestamp.json"), "rb"),
+            call(os.path.join(self.metadata_dir, "snapshot.json"), "rb"),
+            call(os.path.join(self.metadata_dir, "targets.json"), "rb"),
+        ])
+        wrapped_open.reset_mock()
```
sechkova: The test is executed only once, this reset causes no harm but is also useless.

sechkova (changes requested):
I added some comments to the code but then realized, this test does not prove much. Updater always tries to load metadata from cache. There will always be calls to open even if there are no cached files. If this is what we want to demonstrate, ok. Then we are testing that "Updater always tries to read the cache before downloading metadata" and a single call to "refresh" is enough. The whole test would be:
```
updater = self._run_refresh()
wrapped_open.assert_has_calls([
            call(os.path.join(self.metadata_dir, "root.json"), "rb"),
            call(os.path.join(self.metadata_dir, "timestamp.json"), "rb"),
            call(os.path.join(self.metadata_dir, "snapshot.json"), "rb"),
            call(os.path.join(self.metadata_dir, "targets.json"), "rb"),
        ])
```

Probably a way to show that it opened and used the cached metadata is to show that when cached metadata is loaded from the file system successfully, there are no calls to RepositorySimulator.fetch_metadata for snapshot and targets. This should be easier now, after the addition of RepositorySimulator.fetch_counters. "Usage" of local (intermediate) root is described in the spec and should have be covered in test_top_level_update.py. Timestamp is a special case that can be tested in another set of tests covering invalid metadata.

The test is not really consistent. It randomly calls `_run_refresh` and `get_targetinfo`.
In case it is not clear, the options are either:
1. Test only top-level roles metadata 
 - create a new updater, call updater.refresh to get the metadata locally
 - reset the mock
 - create a new updater, call updater.refresh a second time to get the metadata loaded from the local cache dir
 - check the calls to open, check the calls to RepositorySimulator.fetch_tracker.metadata

Then as Jussi suggests, add delegations as follow up.

2. Test both top-level metadata and a delegated role
- add a delegated role to simulator, there is an example in https://github.com/theupdateframework/python-tuf/blob/bdf1cbbbb56839cd99b965fedae2661ed873c4a2/tests/test_updater_with_simulator.py#L148
- create a new updater, call updater.get_targetinfo("non_existent_target") to get the metadata locally, delegated role included
- reset the mock
 - create a new updater, call updater.get_targetinfo("non_existent_target")  a second time to get the metadata loaded from the local cache dir
 - check the the calls to open, check the calls to RepositorySimulator.fetch_tracker.metadata

tests/test_updater_with_simulator.py
```diff
@@ -254,6 +254,44 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+
+        # Add new delegated targets
+        spec_version = ".".join(SPECIFICATION_VERSION)
+        targets = Targets(1, spec_version, self.sim.safe_expiry, {}, None)
+        self.sim.add_delegation("targets", "role1", targets, False, ["*"], None)
+        self.sim.update_snapshot()
+
+        # Make a successful update of valid metadata which stores it in cache
+        updater = self._run_refresh()
+        updater.get_targetinfo("non_existent_target")
```
sechkova: If you look at get_targetinfo, you'll see it calls refresh, so no need for a double refresh here, only creating a new updater instance + updater.get_targetinfo.

sechkova (changes requested):
The test makes sense to me now, thanks for updating it. Some minor comments still but it looks ok to me.

tests/test_updater_with_simulator.py
jku: on the other hand, creating a new updater is multiple lines of code because there's no helper function for that currently in this test file...

ivanayov: I did it with the idea that `_run_refresh()` creates new updater and the update functions are not repeated in `get_targetinfo()` which is practically the same as creating new updater and calling only `get_targetinfo()`. But I'm fine with changing it the other way.

jku (commented):
LGTM: @sechkova all yours to merge when you're happy

tests/test_updater_with_simulator.py
ivanayov: Changed without `_run_refresh()` for better readability.

ivanayov:
Latest force-push is with addressing @sechkova 's [comment](https://github.com/theupdateframework/python-tuf/pull/1703#discussion_r768397735), nothing else has changed.

sechkova (approved):

sechkova:
@jku  GH seems to believe you're still requesting changes and I cannot merge without your approval.

jku:
Sorry Ivana, I just made your life a bit harder by merging Teodoras test move.

Could you rebase?

(test looks really nice btw)

ivanayov:
@jku it's fine, I migrated it to `tests/test_updater_top_level_update.py`

jku (approved):

PR #1704, opened 2021-12-02
Issue#1682/repository simulator fetch tracker

### LINKED ISSUE #1682: RepositorySimulator: add trackers/counters for downloaded files
There is a need in some test cases to check whether Updater requested the correct files from the repository.
For a concrete use case see #1666.

One option used in #1666 is to use `unittest.mock` and "wrap" `RepositorySimulator._fetch_*` methods which does the job but does make the tests less readable.
Therefore as an alternative,  proposed by @jku https://github.com/theupdateframework/python-tuf/pull/1666#pullrequestreview-807253797,  checking that the expected files are downloaded from remote can be done by adding some download counters into `RepositorySimulator` which would  be cleaner and more maintainable

kairoaraujo:
This commit implements a feature in Repository Simulator to
track the fetch calls to the metadata and targets. This feature was
mentioned in PR #1666 that generated issue #1682.
This commit adds ``RepositorySimulator.fetch_tracker``. It also changes
the ``tests/test_updater_consistent_snapshot.py`` to use the
``fetch_tracker`` instead of using mock.

It implements a ``dataclass`` that stores the calls to fetch metadata
(``_fetch_metadata``) in ``fetch_tracker.metadata`` and targets
(``_fetch_targets``) in ``fetch_tracker.targets``.

The fetch calls for metadata, and targets are stored as lists.

that is provided, the better.

Fixes  #1682

requirements**:


tests/repository_simulator.py
```diff
@@ -229,6 +239,8 @@ def _fetch_target(
 
         If hash is None, then consistent_snapshot is not used.
         """
+        self.fetch_tracker.targets.append((target_path, target_hash))
```
kairoaraujo, line 242: We append to the fetch_tracker.targets all fetch requests. Should we append-only successful ones?

tests/repository_simulator.py
```diff
@@ -248,6 +260,8 @@ def _fetch_metadata(
 
         If version is None, non-versioned metadata is being requested.
         """
+        self.fetch_tracker.metadata.append((role, version))
```
kairoaraujo, line 263: We append to the fetch_tracker.metadata all fetch requests. Should we append-only successful ones?

kairoaraujo (commented):
I am storing it as a ``tuple('role name', 'version')`` for metadata and ``tuple('target path', 'target hash')``. I am still in doubt about readability; maybe we should move it to a dictionary, then we don't need to rely on the code comments to understand it at the first code read.

tests/repository_simulator.py
```diff
@@ -81,6 +81,14 @@
 SPEC_VER = ".".join(SPECIFICATION_VERSION)
 
 
+@dataclass
+class FetchCounter:
+    """Fetcher counter for metadata and targets."""
+
+    metadata: list = field(default_factory=list)
+    targets: list = field(default_factory=list)
```
sechkova:
```suggestion
    metadata: List[Tuple[str, Optional[int]]] = field(default_factory=list)
    targets: List[Tuple[str, Optional[str]]] = field(default_factory=list)
```

tests/test_updater_consistent_snapshot.py
```diff
@@ -197,36 +199,29 @@ def test_download_targets(self, test_case_data: Dict[str, Any]):
...
-                updater.download_target(info)
-                expected_prefix = (
-                    None if not hash_algo else info.hashes[hash_algo]
-                )
-                # files are fetched with the expected hash prefix (or None)
-                wrapped_fetch_target.assert_called_once_with(
-                    info.path, expected_prefix
-                )
-                # target files are always persisted without hash prefix
-                self._assert_targets_files_exist([info.path])
```
sechkova, line 228: I think we want to keep this check?

tests/repository_simulator.py
sechkova: I know my suggestion is ugly but if we want to be strict about types ... I guess it's fine until it appears only once.

tests/repository_simulator.py
sechkova: We want to track the all requests coming from the client and I think this is the correct place.

tests/repository_simulator.py
```diff
@@ -116,6 +124,8 @@ def __init__(self) -> None:
         self.dump_dir: Optional[str] = None
         self.dump_version = 0
 
+        self.fetch_tracker: FetchCounter = FetchCounter()
```
sechkova:
```suggestion
        self.fetch_tracker = FetchCounter()
```
I think `mypy` and the code editors are smart enough to infer the type in this case.

tests/test_updater_consistent_snapshot.py
```diff
@@ -197,36 +199,29 @@ def test_download_targets(self, test_case_data: Dict[str, Any]):
         consistent_snapshot: bool = test_case_data["consistent_snapshot"]
         prefix_targets_with_hash: bool = test_case_data["prefix_targets"]
         hash_algo: Optional[str] = test_case_data["hash_algo"]
-        targetpaths = ["file", "file.txt", "..file.ext", "f.le"]
+        targetpaths: list = test_case_data["targetpaths"]
```
sechkova:
```suggestion
        targetpaths: List[str] = test_case_data["targetpaths"]
```

sechkova (changes requested):
Thanks, the trackers look good, I think this is what we needed.
My comments are minor, related to types but also I think one check is missing. 

I was wondering if defining a dataclass (or two dataclasses for metadata and targets) instead of using tuples may simplify the expected calls lists. We can set default values and "typing" will be simpler but on the other hand instead of (root, 1) we'll have to write SomeCall(root, 1) ... so I am not sure about this. This is a way to say that I am looking for a second opinion, not requesting changes :grin:  We can keep it like this for now, maybe when more use cases come this choice we'll be more obvious.

kairoaraujo:
I closed this PR to avoid too many unrelated commits to fix my mess in merging the _develop_ on top of _kairoaraujo:issue#1682/repository-simulator-fetch-tracker_ branch.
The above-linked PR will be taking over with all relevant information.

PR #1705, opened 2021-12-03
Implemented fetch_tracker to RepositorySimulator

### LINKED ISSUE #1682: RepositorySimulator: add trackers/counters for downloaded files

kairoaraujo:
This commit implements a feature in Repository Simulator to
track the fetch calls to the metadata and targets. This feature was
mentioned in PR #1666 that generated issue #1682.
This commit adds RepositorySimulator.fetch_tracker. It also changes
the tests/test_updater_consistent_snapshot.py to use the
fetch_tracker instead of using mock.

It implements a dataclass that stores the calls to fetch metadata
(_fetch_metadata) in fetch_tracker.metadata and targets
(_fetch_targets) in fetch_tracker.targets.

The fetch calls for metadata, and targets are stored as lists.

that is provided, the better.

Fixes #1682

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -48,7 +48,7 @@ import logging
 import os
 import tempfile
 from collections import OrderedDict
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from typing import Dict, Iterator, List, Optional, Tuple
 from urllib import parse
@@ -79,6 +79,14 @@ from tuf.ngclient.fetcher import FetcherInterface
 logger = logging.getLogger(__name__)
 
 SPEC_VER = ".".join(SPECIFICATION_VERSION)
+
+
+@dataclass
+class FetchTracker:
+    """Fetcher counter for metadata and targets."""
+
+    metadata: List[Tuple[str, Optional[int]]] = field(default_factory=list)
+    targets: List[Tuple[str, Optional[str]]] = field(default_factory=list)
 
 
 @dataclass
@@ -115,6 +123,8 @@ class RepositorySimulator(FetcherInterface):
 
         self.dump_dir: Optional[str] = None
         self.dump_version = 0
+
+        self.fetch_tracker = FetchTracker()
 
         now = datetime.utcnow()
         self.safe_expiry = now.replace(microsecond=0) + timedelta(days=30)
@@ -229,6 +239,8 @@ class RepositorySimulator(FetcherInterface):
 
         If hash is None, then consistent_snapshot is not used.
         """
+        self.fetch_tracker.targets.append((target_path, target_hash))
+
         repo_target = self.target_files.get(target_path)
         if repo_target is None:
             raise FetcherHTTPError(f"No target {target_path}", 404)
@@ -248,6 +260,8 @@ class RepositorySimulator(FetcherInterface):
 
         If version is None, non-versioned metadata is being requested.
         """
+        self.fetch_tracker.metadata.append((role, version))
+
         if role == Root.type:
             # return a version previously serialized in publish_root()
             if version is None or version > len(self.signed_roots):
--- a/tests/test_updater_consistent_snapshot.py
+++ b/tests/test_updater_consistent_snapshot.py
@@ -10,7 +10,6 @@ import sys
 import tempfile
 import unittest
 from typing import Any, Dict, Iterable, List, Optional
-from unittest.mock import call, patch
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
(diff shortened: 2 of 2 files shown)


kairoaraujo:
I decided to submit a new PR after my big mess on PR #1704, and I have to say sorry.
I addressed all comments done by @sechkova 

As mentioned by @sechkova in the PR #1704

> Thanks, the trackers look good, I think this is what we needed.
> My comments are minor, related to types but also I think one check is missing.

> I was wondering if defining a dataclass (or two dataclasses for metadata and targets) instead of using tuples may simplify the expected calls lists. We can set default values and "typing" will be simpler but on the other hand instead of (root, 1) we'll have to write SomeCall(root, 1) ... so I am not sure about this. This is a way to say that I am looking for a second opinion, not requesting changes 😁 We can keep it like this for now, maybe when more use cases come this choice we'll be more obvious.

I also thought to have different dataclasses for metadata and targets in the initial design.

tests/repository_simulator.py
```diff
@@ -139,7 +149,7 @@ def targets(self) -> Targets:
 
     def all_targets(self) -> Iterator[Tuple[str, Targets]]:
         """Yield role name and signed portion of targets one by one."""
-        yield Targets.type, self.md_targets.signed
+        yield "targets", self.md_targets.signed
```
ivanayov: What is the reason behind reverting back to strings instead of using the constants?

tests/repository_simulator.py
```diff
@@ -257,11 +271,11 @@ def _fetch_metadata(
 
         # sign and serialize the requested metadata
         md: Optional[Metadata]
-        if role == Timestamp.type:
+        if role == "timestamp":
```
ivanayov: Same here.

tests/test_updater_consistent_snapshot.py
```diff
@@ -147,7 +146,7 @@ def test_delegated_roles_update(self, test_case_data: Dict[str, Any]):
         consistent_snapshot: bool = test_case_data["consistent_snapshot"]
         expected_version: Optional[int] = test_case_data["expected_version"]
         rolenames = ["role1", "..", "."]
-        expected_calls = [call(role, expected_version) for role in rolenames]
+        expected_calls = [(role, expected_version) for role in rolenames]
```
ivanayov, line 148: I like this simplification.

ivanayov:
Looks good to me, just one question

tests/repository_simulator.py
kairoaraujo: It is related to my mess-up. I fixed it.

sechkova (approved):
Looks ok to me.  Do you mind squashing the two commits before merging?

tests/test_updater_consistent_snapshot.py
```diff
@@ -197,36 +199,32 @@ def test_download_targets(self, test_case_data: Dict[str, Any]):
         consistent_snapshot: bool = test_case_data["consistent_snapshot"]
         prefix_targets_with_hash: bool = test_case_data["prefix_targets"]
         hash_algo: Optional[str] = test_case_data["hash_algo"]
-        targetpaths = ["file", "file.txt", "..file.ext", "f.le"]
+        targetpaths: List[str] = test_case_data["targetpaths"]
```
jku, line 200: is there an advantage to this change?

tests/test_updater_consistent_snapshot.py
```diff
@@ -197,36 +199,32 @@ def test_download_targets(self, test_case_data: Dict[str, Any]):
...
-                    None if not hash_algo else info.hashes[hash_algo]
-                )
-                # files are fetched with the expected hash prefix (or None)
-                wrapped_fetch_target.assert_called_once_with(
-                    info.path, expected_prefix
-                )
-                # target files are always persisted without hash prefix
-                self._assert_targets_files_exist([info.path])
-                wrapped_fetch_target.reset_mock()
+        expected_result = []
```
jku: This is a nitpick but... I think the variable name is quite vague. The variable could probably also be avoided altogether by doing the assert (and clearing the fetch list) inside the loop: then the asserts for local files and fetches would happen at same time and same place.

jku (approved):
Looks ok to merge to me but I did leave two comments. I can merge as is but maybe have a look and let me know what you think?

tests/test_updater_consistent_snapshot.py
kairoaraujo: If combined with your comment below, I would say readability (?) adding the `targetpaths` to the test dataset.

jku:
Oh sorry, one more thing (trying to make life easier for @MVrachev): this PR probably has unused imports that will start failing with #1710 ? Can you check the unittest.mock imports?

Feel free to squash the PR to a single commit if you do end up making a change.

kairoaraujo:
Thank you @jku, I will pay more attention.

jku:
I think we'll go with this version :) thanks

PR #1707, opened 2021-12-03
Test expired metadata from cache

### LINKED ISSUE #1681: ngtests: test loading of cached metadata

ivanayov:
This tests that an expired timestamp/snapshot/targets when loaded
from cache is not stored as final but is used to verify the new
timestamp

Addresses #1681

### DIFF
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -12,7 +12,7 @@ import tempfile
 import unittest
 from datetime import datetime, timedelta
 from typing import Iterable, Optional
-from unittest.mock import MagicMock, call, patch
+from unittest.mock import MagicMock, Mock, call, patch
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
@@ -661,6 +661,32 @@ class TestRefresh(unittest.TestCase):
 
         expected_calls = [("root", 2), ("timestamp", None)]
         self.assertListEqual(self.sim.fetch_tracker.metadata, expected_calls)
+
+    def test_expired_metadata(self) -> None:
+        # Test that expired local timestamp/snapshot can be used for updating
+        # from remote
+
+        # Make a successful update of valid metadata which stores it in cache
+        self._run_refresh()
+
+        # Simulate expired local metadata by mocking system time one second ahead
+        mock_time = Mock()
+        mock_time.return_value = (
+            int(self.sim.timestamp.expires.strftime("%Y%m%d%H%M%S")) + 1
+        )
+        with patch("time.time", mock_time):
+            self.sim.targets.version += 1
+            self.sim.update_snapshot()
+            # Create a new updater and perform a second update while
+            # the metadata is already stored in cache (metadata dir)
+            self._run_refresh()
+
+        # Assert that the final version of timestamp/snapshot is version 2
+        # which means a successful refresh is performed
+        # with expired local metadata
+        for role in ["timestamp", "snapshot", "targets"]:
+            md = Metadata.from_file(f"{self.metadata_dir}/{role}.json")
+            self.assertEqual(md.signed.version, 2)
 
 
 if __name__ == "__main__":
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -306,6 +306,46 @@ class TestRefresh(unittest.TestCase):
 
         self._assert_files_exist([Root.type])
 
+    def test_expired_timestamp_version_rollback(self) -> None:
+        self._run_refresh()
+
+        mock_time = Mock()
+        mock_time.return_value = (
+            int(self.sim.timestamp.expires.strftime("%Y%m%d%H%M%S")) + 1
+        )
+        with patch("time.time", mock_time):
+            # Check for a rollback attack
+            self.sim.timestamp.version = 2
+            self._run_refresh()
+
+            self.sim.timestamp.version = 1
+            with self.assertRaises(ReplayedMetadataError):
+                self._run_refresh()
+
+            self._assert_version_equals(Timestamp.type, 2)
+
+    def test_expired_timestamp_snapshot_rollback(self) -> None:
+        self._run_refresh()
+
+        mock_time = Mock()
+        mock_time.return_value = (
+            int(self.sim.timestamp.expires.strftime("%Y%m%d%H%M%S")) + 1
+        )
+        with patch("time.time", mock_time):
+            # Check for a rollback attack.
+            self.sim.snapshot.version = 2
+            self.sim.update_timestamp()  # timestamp v2
+            self._run_refresh()
+
+            # Snapshot meta version is smaller than previous
+            self.sim.timestamp.snapshot_meta.version = 1
+            self.sim.timestamp.version += 1  # timestamp v3
+
+            with self.assertRaises(ReplayedMetadataError):
+                self._run_refresh()
+
+            self._assert_version_equals(Timestamp.type, 2)
+
     def test_new_timestamp_version_rollback(self) -> None:
         # Check for a rollback attack
         self.sim.timestamp.version = 2
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -310,43 +310,78 @@ class TestRefresh(unittest.TestCase):
 
     @patch.object(datetime, "datetime", wraps=datetime.datetime)
     def test_expired_timestamp_version_rollback(self, mock_time: Mock) -> None:
+        """Verifies that local timestamp is used in rollback checks even if it is expired.
+
+        The timestamp updates and rollback checks are performed
+        with the following timing:
+         - Timestamp v1 expiry set to day 7
+         - First updater refresh performed on day 0
+         - Repository publishes timestamp v2 on day 0
+         - Timestamp v2 expiry set to day 21
+         - Second updater refresh performed on day 18:
+           assert that rollback check uses expired timestamp v1"""
+
+        now = datetime.datetime.utcnow()
+        self.sim.timestamp.expires = now + datetime.timedelta(days=7)
+
+        self.sim.timestamp.version = 2
+
+        # Make a successful update of valid metadata which stores it in cache
         self._run_refresh()
 
+        self.sim.timestamp.expires = now + datetime.timedelta(days=21)
+
+        self.sim.timestamp.version = 1
+
         mock_time.utcnow.return_value = (
-            datetime.datetime.utcnow() + datetime.timedelta(seconds=1)
+            datetime.datetime.utcnow() + datetime.timedelta(days=18)
         )
         with patch("datetime.datetime", mock_time):
-            # Check for a rollback attack
-            self.sim.timestamp.version = 2
-            self._run_refresh()
-
-            self.sim.timestamp.version = 1
+            # Check that a rollback protection is performed even if
+            # local timestamp has expired
             with self.assertRaises(BadVersionNumberError):
                 self._run_refresh()
 
-            self._assert_version_equals(Timestamp.type, 2)
+        self._assert_version_equals(Timestamp.type, 2)
 
     @patch.object(datetime, "datetime", wraps=datetime.datetime)
     def test_expired_timestamp_snapshot_rollback(self, mock_time: Mock) -> None:
+        """Verifies that rollback protection is done even if local timestamp has expired.
+
+        The snapshot updates and rollback protection checks are performed
+        with the following timing:
+         - Timestamp v1 expiry set to day 7
+         - Repository bumps snapshot to v3 on day 0
+         - First updater refresh performed on day 0
+         - Timestamp v2 expiry set to day 21
+         - Second updater refresh performed on day 18:
+           assert that rollback protection is done with expired timestamp v1"""
+
+        now = datetime.datetime.utcnow()
+        self.sim.timestamp.expires = now + datetime.timedelta(days=7)
+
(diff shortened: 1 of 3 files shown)


ivanayov:
I'm leaving this with the non-matching description on purpose, as the expected behaviour in
the issue description is to load expired snapshot/targets, which is not the case. Not loading them
is actually correct from what I understand from the spec - you shouldn't be able to use outdated timestamp for verification, thus you cannot verify snapshot and targets as well. Please, correct me if I misunderstand. 
So we could ether reword the requirement, of if there's a bug we can fix it.

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +259,50 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+        with self.assertRaises(ExpiredMetadataError):
+            updater.get_targetinfo(self.targets_dir)
+
+        # Test that an expired timestamp/snapshot/targets
+        # when loaded from cache is not stored as final
+        # but is used to verify the new timestamp
+        wrapped_open.assert_has_calls([
+            call(os.path.join(self.metadata_dir, "root.json"), "rb"),
+            call(os.path.join(self.metadata_dir, "timestamp.json"), "rb"),
+        ])
```
MVrachev: When reading the changes made in this pr I am comparing them to @sechkova comment here: https://github.com/theupdateframework/python-tuf/issues/1681#issuecomment-977851021.
It's not clear to me after you finish step 2 which steps do you cover with those asserts.
@ivanayov can you give more details on that?

ivanayov: @MVrachev This PR is for step 4 only. The others are covered in #1703 .

ivanayov: Step 5 is already covered in the tests, so we discussed with @jku that it's not necessary to repeat it here.

MVrachev: I will look at it today a little later.

MVrachev: Please keep in mind that by using `Fixes` in your pr description you are going to close the issue you are referencing with this pr.
That's okay if you handle all of the steps here, but as it seems you don't test for all of the steps here, but you also rely on #1703 in order to do that.
Now the first pr that merges between this one and #1703 will close the issue even though none of the two prs fixes the issue entirely by itself.

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +259,50 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+            "https://example.com/targets/",
+            self.sim,
+        )
+
+        with self.assertRaises(ExpiredMetadataError):
+            updater.get_targetinfo(self.targets_dir)
+
+        # Test that an expired timestamp/snapshot/targets
+        # when loaded from cache is not stored as final
+        # but is used to verify the new timestamp
```
MVrachev: Here you mentioned that you are checking that an expired `timestamp/snapshot/targets, when loaded from cache, is not stored as final but is used to verify the new timestamp`, but you are only checking that there are
calls for `root.json` and `timestamp.json`.
I thought from your comment and step 3 that you will have to check for `snapshot` and `targets` as well.

ivanayov: @MVrachev I've described this in https://github.com/theupdateframework/python-tuf/pull/1707#issuecomment-985324978 . Any feedback on that direction is welcome.

tests/test_updater_with_simulator.py
ivanayov: Thanks, I've edited the description.

tests/test_updater_with_simulator.py
MVrachev: I am sorry that I didn't read your comment above. I have read this pr description before the comment was posted and didn't check if there are new comments...

tests/test_updater_with_simulator.py
```diff
@@ -257,6 +259,50 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+        updater = Updater(
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            self.targets_dir,
+            "https://example.com/targets/",
+            self.sim,
+        )
+
+        with self.assertRaises(ExpiredMetadataError):
+            updater.get_targetinfo(self.targets_dir)
```
MVrachev: Why did you call `updater.get_targetinfo(self.targets_dir)`?
Why didn't you use directly `updater.refresh()`?
Also, the argument to `updater.get_targetinfo()` should be a path to a specific target, not the targets directory.

MVrachev: I asked Teodora about this here https://github.com/theupdateframework/python-tuf/issues/1681#issuecomment-987330543,

MVrachev:
I have some ideas about this pr and will have a conversation with Ivana soon.

jku:
> the expected behaviour in
the issue description is to load expired snapshot/targets, which is not the case. Not loading them
is actually correct from what I understand from the spec 

This may have been clarified already but I'l say it here:

We do load expired metadata with one big caveat: the expired metadata can only be used to validate a newer version of the _same type of metadata_. In practice:
* local, possibly expired, metadata is loaded and used to validate newer downloaded metadata. "validate" here means signature verification for root and various version checks for timestamp, snapshot and targets.
* even the downloaded versions may be expired (in practice only root is, because we download all root versions up to the current, hopefully non-expired version)
* the python-tuf implementation  will fail the update once the process moves to the next type of metadata and the previous metadata is still expired (e.g. if final root version is expired, refresh() will fail when it tries to validate timestamp)

ivanayov:
`tox` is giving really strange proposals here:
```
-        past_datetime = datetime.utcnow().replace(microsecond=0) - timedelta(days=5)
+        past_datetime = datetime.utcnow().replace(microsecond=0) - timedelta(
+            days=5
+        )
```
Do we really expect that or it's misconfiguration?

jku:
the core issue here is that _verifying that local files are opened does not mean they are actually used to verify new downloaded metadata_ (this is made clear even in the test itself: it opens `targets.json` but we know that's not used for anything). This does not mean the test is worthless but at least some descriptions in the PR and commit are not describing it correctly. I would like to see a clear explanation of what the test does so that 
 A) we can be sure we don't already have a test that does the same checks
 B) future readers won't be deceived by inaccurate descriptions 
I get that this may be the hardest part of this PR: let's talk if you don't have a good idea already

Just to be clear: I think this is a useful test that probably does not exist yet... but test_updater_top_level_update.py has some clearly related tests like `test_new_timestamp_expired()`: maybe it should have a companion `test_local_timestamp_expired()`? I don't have strong opinions here but maybe @sechkova has ideas about where to put tests like this -- and whether other tests exist already?

Also (and this is a small detail in comparison), I've been trying to advocate not meddling with the client metadata cache manually: in this case I think it would be better to not do that and instead mock current time (pretend that time now > expiry). Search for mock_time in tests for examples of this.

tests/test_updater_with_simulator.py
```diff
@@ -254,6 +255,32 @@ def test_not_loading_targets_twice(self, wrapped_open: MagicMock) -> None:
...
+
+        # Make a successful update of valid metadata which stores it in cache
+        self._run_refresh()
+
+        # Simulate expired local metadata by mocking system time one second ahead
+        mock_time = Mock()
+        mock_time.return_value = (
+            int(self.sim.timestamp.expires.strftime("%Y%m%d%H%M%S")) + 1
+        )
+        with patch("time.time", mock_time):
```
sechkova: I am not sure I understand the exact intentions but the tests pass because patch("time.time", mock_time) does not work as you expected and there isn't any expired metadata. The tests would work the exact same way if you delete the lines doing the "mocking". 
Updater (TrustedMetadataSet) uses datetime.utcnow to calculate the reference time.

sechkova:
> Just to be clear: I think this is a useful test that probably does not exist yet... but test_updater_top_level_update.py has some clearly related tests like `test_new_timestamp_expired()`: maybe it should have a companion `test_local_timestamp_expired()`? I don't have strong opinions here but maybe @sechkova has ideas about where to put tests like this -- and whether other tests exist already?

I admit I find it hard to follow the intentions here, I am sorry if I've put a misleading issue description.
If I understand correctly, `test_trusted_root_expired` from test_updater_top_level_update.py is doing something very similar and can be used as an example.

ivanayov:
@sechkova the idea was to simulate an expired metadata in cache without manually modifying the cache files. I might have misunderstood this mock - the intention is to simulate change in system time, so that what's stored in cache is considered expired, without need of directly touching the local metadata.

jku:
> I admit I find it hard to follow the intentions here, I am sorry if I've put a misleading issue description. If I understand correctly, `test_trusted_root_expired` from test_updater_top_level_update.py is doing something very similar and can be used as an example.

It is pretty similar, except that I think we don't actually persist downloaded non-root metadata if it's expired, even though we probably could (rado pointed this out to me -- I had not noticed). This difference isn't that important for actual working repos but for testing it means we need to mock system time (or utcnow() as you point out) to "make the local metadata expired".

sechkova:
> > I admit I find it hard to follow the intentions here, I am sorry if I've put a misleading issue description. If I understand correctly, `test_trusted_root_expired` from test_updater_top_level_update.py is doing something very similar and can be used as an example.
> 
> It is pretty similar, except that I think we don't actually persist downloaded non-root metadata if it's expired, even though we probably could (rado pointed this out to me -- I had not noticed). This difference isn't that important for actual working repos but for testing it means we need to mock system time (or utcnow() as you point out) to "make the local metadata expired".

Thanks, now I think I got all pieces together. So I agree, `test_updater_top_level_update.py` could host the additional tests for loading expired snapshot and timestamp and since expired metadata and timestamp are not persisted, mocking seems to be the way. 
I think expired local "targets" should be treated the same way as an invalid (unsigned) local metadata and may be part of another set of tests?

jku:
I'm marking this a draft since the issue pointed out by Teodora (mocking the wrong thing) is still there: I might have other comments but I've been postponing review until the basics are there. I'm available for a chat if anything is unclear.

ivanayov:
@jku I changed to mocking `datetime` as Teodora advised, so it's now available for review.

jku (changes requested):
marking request changes so it's visible on PR list: we talked about this over zoom, and found an issue with the timeline of expiries and the mocked refresh: ivana promised to fix

ivanayov:
I've pushed an update.

tests/test_updater_top_level_update.py
```diff
@@ -306,6 +308,69 @@ def test_new_timestamp_unsigned(self) -> None:
 
         self._assert_files_exist([Root.type])
 
+    @patch.object(datetime, "datetime", wraps=datetime.datetime)
+    def test_expired_timestamp_version_rollback(self, mock_time: Mock) -> None:
+
+        now = datetime.datetime.utcnow()
+        self.sim.timestamp.expires = now.replace(
+            microsecond=0
+        ) + datetime.timedelta(days=7)
```
jku: I think I suggested this (since it's what repositorysimulator does) but in fact the `expires` setter nowadays handles the microseconds, so I think this works for all cases where the setter is used:

```suggestion
        self.sim.timestamp.expires = now + datetime.timedelta(days=7)
```

tests/test_updater_top_level_update.py
```diff
@@ -306,6 +308,69 @@ def test_new_timestamp_unsigned(self) -> None:
...
+        self._run_refresh()
+
+        self.sim.timestamp.version = 1
+
+        self.sim.timestamp.expires = now.replace(
+            microsecond=0
+        ) + datetime.timedelta(days=21)
+
+        mock_time.utcnow.return_value = (
+            datetime.datetime.utcnow() + datetime.timedelta(days=18)
```
jku, line 337:
```suggestion
            now + datetime.timedelta(days=18)
```
maybe easier to make the connection to the expiry dates if we use the same construct.

This applies to the other tests as well

jku (changes requested):
Nice, all test code seems correct and clear.

* Left some minor tweak suggestions in code
* These tests should have a docstring to explain what they are testing: they are some of the trickiest tests we have. Something like "Local timestamp must be used in rollback checks even if it is expired" -- maybe the same text that's already before the actual assertRaises

Additionally _maybe_ the dosctring should also include a timeline, example of one test:
```
  * day 0: first updater refresh
  * day 0 (a bit later): repository publishes timestamp v2
  * day 7: timestamp v1 expiry
  * day 18: second updater refresh: assert that rollback check uses expired v1 here
  * day 21: timestamp v2 expiry

```
not sure about that last thing though: it's a lot of lines... but may be helpful to understand what's going on? I'll leave decision to you

ivanayov:
Thank you @jku , I added docstrings to all the test methods for clarification.

jku (approved):
my comment about reusing the `datetime.datetime.utcnow()` result  is marked resolved but all functions still call utcnow() twice... 

That is a trivial complaint so let's go with this version, I think it's a good set of very tricky tests. Thanks.

ivanayov:
@jku ah, I resolved by inertia with the microseconds updates... will note to address in following changes. Thanks

PR #1708, opened 2021-12-03
Lint additional four files in the tests folder

### LINKED ISSUE #1582: tests: Start linting tests

MVrachev:
Related to #1582 

When preparing the pr which was going to resolve #1582 I had to rename the test files
related to the old codebase. That way I was able to exclude the old test files when linting
the `tests/` folder.
While I did that I realized I have to decide whether we want to lint and consequently rename
`utils.py`, `test_utils.py`, `simple_server.py`, and `aggregate_tests.py` after we remove the files testing the old code.
I decided that it's worth linting them for the following reasons:
- `utils.py` provides utilities that are used by our new code such as the `TestServerProcess` class
which handles server handling, logging, and the `run_sub_tests_with_dataset` decorator which
we used in multiple files.
- `test_utils.py` will be needed even after we remove the old test files as we want to have tests
that verify that `TestServerProcess` from `utils.py` is working as expected
- `simple_server.py` is the only helper for server startup used in the new code base and
more precisely to start the server in `test_updater_ng.py`
- `aggregate_tests.py` will be needed to aggregate the test files

That's why I decided to run the linters upon those files.
I tried to summarize the manual changes I had to make in the commit messages
as they weren't many.
Most of the changes are automatic and are result from `black` and `isort`.

requirements**:

### DIFF
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -25,17 +25,21 @@
     http://docs.python.org/library/simplehttpserver.html#module-SimpleHTTPServer
 """
 
-import sys
-import random
 import socketserver
+import sys
 from http.server import SimpleHTTPRequestHandler
+from typing import Type, Union
 
 
 class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
-  """A SimpleHTTPRequestHandler that does not write incoming requests to
-  stderr. """
-  def log_request(self, code='-', size='-'):
-    pass
+    """A SimpleHTTPRequestHandler that does not write incoming requests to
+    stderr."""
+
+    def log_request(
+        self, code: Union[int, str] = "-", size: Union[int, str] = "-"
+    ) -> None:
+        pass
+
 
 # NOTE: On Windows/Python2 tests that use this simple_server.py in a
 # subprocesses hang after a certain amount of requests (~68), if a PIPE is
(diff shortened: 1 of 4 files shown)


tests/simple_server.py
```diff
@@ -25,40 +25,42 @@
...
+handler: Type[
+    Union[SimpleHTTPRequestHandler, QuietHTTPRequestHandler]
+] = QuietHTTPRequestHandler
 
-if use_quiet_http_request_handler:
-  handler = QuietHTTPRequestHandler
-else:
-  handler = SimpleHTTPRequestHandler
+if len(sys.argv) > 2:
+    handler = SimpleHTTPRequestHandler
```
lukpueh: For the purpose of this script the type annotation really seems like an overkill that impairs readability. Would it be problematic if we just did:
```python
if len(sys.argv) > 2 and sys.argv[2]:
  handler = QuietHTTPRequestHandler
else:
  handler = SimpleHTTPRequestHandler
```

tests/utils.py
```diff
@@ -86,253 +91,265 @@ def ignore_deprecation_warnings(module: str) -> Iterator[None]:
...
+                # The process has exited.
+                raise TestServerProcessError(
+                    self.server
+                    + " exited unexpectedly "
+                    + "with code "
+                    + str(self.__server_process.poll())
+                    + "!"
+                )
+
+            if line.startswith(expected_msg):
```
lukpueh, line 291: Out of curiosity, did `black` make this `elif` an `if`? Or did pylint flag it?

lukpueh (approved):
Thanks, Martin! LGTM.

tests/simple_server.py
MVrachev: That way I receive the following two errors by `mypy`:
```
tests/simple_server.py:61: error: Cannot assign multiple types to name "handler"  without an explicit "Type[...]" annotation  [misc]

tests/simple_server.py:61: error: Incompatible types in assignment (expression has type "Type[SimpleHTTPRequestHandler]", variable has type "Type[QuietHTTPRequestHandler]")  [assignment]
```

I wanted to simplify it somehow but didn't come up with an idea of how to do it without a `mypy` error...
That's how I end up with this ugly line...

tests/utils.py
MVrachev: It was `pylint` who flag it with:
`R1720: Unnecessary "elif" after "raise" (no-else-raise)`

MVrachev:
I added a new commit linting `aggregate_tests.py` as we expect to use it even after we drop the old test files.

tests/utils.py
lukpueh: 👍  Oh right, I remember that one.

tests/simple_server.py
lukpueh: I guess you could "declare" `handler` first. Looks pretty unpythonic to me, but a bit nicer then the long line above. 
```python
handler: Type[Union[SimpleHTTPRequestHandler, QuietHTTPRequestHandler]]

if len(sys.argv) > 2 and sys.argv[2]:
  handler = QuietHTTPRequestHandler
else:
  handler = SimpleHTTPRequestHandler
```

But I'm fine with leaving it as is.

MVrachev: You are right that's better even if it's not the most pythonic way it's still more understandable.
Updated the code with that suggestion.

PR #1710, opened 2021-12-06
Start linting the test

### LINKED ISSUE #1582: tests: Start linting tests

MVrachev:
Fixes #1582 

Start linting the files that test the new code. In order to do that the old test files
had to be excluded. They were are renamed to `test_*_old.py`, so we can add globs/regex to exclude
the files testing the old code.
These options will be removed after we remove those files.

Additional `black`, `pylint` and `mypy` errors were found and had to be addressed
before we can start linting.

requirements**:


tests/test_indefinite_freeze_attack_old.py
```diff
@@ -36,7 +36,6 @@
   metadata without the client being aware.
 """
 
-import datetime
```
lukpueh, line 39: Why this change in an "_old.py" file?

tests/test_updater_consistent_snapshot.py
```diff
@@ -117,6 +121,7 @@ def test_top_level_roles_update(self, test_case_data: Dict[str, Any]):
         sim = self._init_repo(consistent_snapshot)
         updater = self._init_updater(sim)
 
+        # pylint: disable=protected-access
```
lukpueh: Maybe we want to disable this globally for all tests?

lukpueh (approved):
Thanks for the PR, @MVrachev! Looks like a straight-forward set of changes. I left some minor suggestions, but will approve regardless.

I see you adopted the filename changes (add "_old" infix) in some doc headers but not all. I suggest to do this consistently and I'm fine with not updating the doc headers, given that these files are bound to be removed.

tests/test_indefinite_freeze_attack_old.py
MVrachev: I just saw it was unused and decided to remove it, but I can return it if it looks strange.

lukpueh: 🤷  Probably not worth the effort. :) Let's leave it there.

lukpueh:
Just merged #1708. Please rebase and mark ready for review!

tests/test_updater_consistent_snapshot.py
MVrachev: I don't think that is possible as of today.
See issue https://github.com/PyCQA/pylint/issues/618 and their project where they keep track of the progress on that issue https://github.com/PyCQA/pylint/projects/1

MVrachev: I don't think that is possible as of today.
See issue https://github.com/PyCQA/pylint/issues/618 and their project where they keep track of the progress on that issue https://github.com/PyCQA/pylint/projects/1

lukpueh: You could call pylint separately for tests and for everything else in tox.ini, as we [do for in-toto](https://github.com/in-toto/in-toto/blob/cb562165b2f62d1fabc80b0b5fdcf212df8fd3d6/tox.ini#L29-L32).

MVrachev:
> Thanks for the PR, @MVrachev! Looks like a straight-forward set of changes. I left some minor suggestions, but will approve regardless.
> 
> I see you adopted the filename changes (add "_old" infix) in some doc headers but not all. I suggest to do this consistently and I'm fine with not updating the doc headers, given that these files are bound to be removed.

I rebased on top of develop and I think changed all the relevant doc headers and places where the file names of
the old test files were mentioned.
The pr is ready for review!

tests/test_updater_consistent_snapshot.py
MVrachev: The only downside is that we will no longer have one source for the folders we lint or in other words we won't be able to use `lint_dirs`: https://github.com/theupdateframework/python-tuf/blob/aebd8b0fafdd52e5d6765f4dd628df0492179c1a/tox.ini#L41 for pylint.
@jku any thoughts here as you suggested `lint_dirs`?

MVrachev:
I had to rebase to change the test name used as an example in `docs/CONTRIBUTIONS.rst`.

tests/test_examples.py
```diff
@@ -7,11 +7,14 @@
...
 import os
 import shutil
+import sys
 import tempfile
 import unittest
 from pathlib import Path
 from typing import ClassVar, List
 
+from tests import utils
+
```
MVrachev, line 18: I linted with `isort` this file and that was what suggested.
I thought the list should be like that:
```
import glob
import os
import shutil
import sys
import tempfile
import unittest

from pathlib import Path
from typing import ClassVar, List
from tests import utils
```
but isort fixes that to the current state.

tests/test_examples.py
```diff
@@ -7,11 +7,14 @@
...
```
MVrachev: I linted with `isort` this file and that was what suggested.
I thought the list should be like that:
```
import glob
import os
import shutil
import sys
import tempfile
import unittest

from pathlib import Path
from typing import ClassVar, List
from tests import utils
```
but isort fixes that to the current state.

pyproject.toml
```diff
@@ -88,6 +88,7 @@ exclude=".*_old.py"
 [[tool.mypy.overrides]]
 module = [
   "securesystemslib.*",
-  "urllib3.*"
+  "urllib3.*",
+  "dateutil.relativedelta",
```
jku:
```
tests/test_api.py:19: error: Library stubs not installed for "dateutil.relativedelta" (or incompatible with Python 3.10)  [import]
tests/test_api.py:19: note: Hint: "python3 -m pip install types-python-dateutil"
```

I would try adding types-python-dateutil into requirements-test.txt instead

tests/test_updater_consistent_snapshot.py
jku: globally disabling the lint is ok ... but at this point might be overkill: As an example, there's already a PR that removes these protected accesses (because there is a better design for these cases): I think we won't have more than a couple in the whole test suite next week.

(this is a response to earlier comment)

tests/test_api.py
```diff
@@ -136,7 +136,12 @@ def test_compact_json(self) -> None:
         )
 
     def test_read_write_read_compare(self) -> None:
-        for metadata in [Root.type, Snapshot.type, Timestamp.type, Targets.type]:
+        for metadata in [
+            Root.type,
+            Snapshot.type,
+            Timestamp.type,
+            Targets.type,
+        ]:
```
jku: it's a nitpick but in cases like this and the one below a bit of tweaking could save ten lines of code:

This probably fits and is a better name:
`        for role in [Root.type, Snapshot.type, Timestamp.type, Targets.type]:`
or :
`        for rolename in TOP_LEVEL_ROLE_NAMES:`

Maybe this one is not worth modifying the PR for but this sort of thing adds up... and makes the tests harder to read

jku (changes requested):
Moving almost obsolete files is a bit annoying but I can see why you want to flick the switch on this one :) style guidelines won't be observed without hard failures... LGTM in general

We should keep statically checking dateutil, otherwise my comments are just discussion.

tests/test_api.py
MVrachev: `for role in [Root.type, Snapshot.type, Timestamp.type, Targets.type]:` is more than 80 characters, but
`for rolename in TOP_LEVEL_ROLE_NAMES:` will do the work.
I thought we have deleted `TOP_LEVEL_ROLE_NAMES` when we introduce the constatnts and it's a nice surprise we didn't.

pyproject.toml
MVrachev: Great, I totally forgot I can do this.

MVrachev:
I have rebased on top of develop.
It was painful to do so as there were multiple changes and there were new warnings from all of the linters meaning I had to update their commits.
Additionally, I addressed all of Jussi's comments and had to add a couple of new commits (the last two) that handle specific problems.
It would be great if we merge this soon as with each new change it's annoying to do that work again.
@jku and/or @lukpueh can you have a new look?

tests/test_updater_old.py
```diff
@@ -21,7 +21,7 @@
   See LICENSE-MIT OR LICENSE for licensing information.
 
 <Purpose>
-  'test_updater.py' provides a collection of methods that test the public /
+  'test_updater.py_old' provides a collection of methods that test the public /
```
lukpueh, line 24: Is this the coding equivalent of the french Verlan? :P Given the short remaining life time of this doc, I say we leave this as is.

lukpueh (approved):
Many thanks for doing this tedious work, Martin! I'll go forward and merge, because it looks like you addressed Jussi's requested change.

PR #1711, opened 2021-12-07
Extend delegations tests 

### LINKED ISSUE #1641: ngtests: Add tests for delegated roles metadata update

sechkova:
~_WIP since it is built on top of #1689_~

Fixes #1641

Extend test_delegation_graphs with tests for:
- max number of delegations
- invalid metadata
- searching for target info

requirements**:

### DIFF
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -3,15 +3,15 @@
 # Copyright 2021, New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-"""Test updating delegated targets roles with various
-delegation hierarchies"""
+"""Test updating delegated targets roles and searching for
+target files with various delegation graphs"""
 
 import os
 import sys
 import tempfile
 import unittest
 from dataclasses import astuple, dataclass, field
-from typing import Iterable, List, Optional
+from typing import Iterable, List, Optional, Tuple
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
@@ -36,17 +36,24 @@ class TestDelegation:
 
 
 @dataclass
+class TestTarget:
+    rolename: str
+    content: bytes
+    targetpath: str
+
+
+@dataclass
 class DelegationsTestCase:
-    """Describes a delegations graph as a list of delegations
-    and the expected order of traversal as 'visited_order'."""
+    """A delegations graph as lists of delegations and target files
+    and the expected order of traversal as a list of role names."""
 
     delegations: List[TestDelegation]
-    visited_order: List[str]
+    target_files: List[TestTarget] = field(default_factory=list)
+    visited_order: List[str] = field(default_factory=list)
 
 
-class TestDelegationsGraphs(unittest.TestCase):
-    """Test creating delegations graphs with different complexity
-    and successfully updating the delegated roles metadata"""
+class TestDelegations(unittest.TestCase):
+    """Base class for delegation tests"""
 
     # set dump_dir to trigger repository state dumps
     dump_dir: Optional[str] = None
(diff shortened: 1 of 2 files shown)


tests/test_updater_delegation_graphs.py
```diff
@@ -251,16 +277,119 @@ def test_graph_traversal(self, test_data: DelegationsTestCase) -> None:
...
+
+    invalid_metadata: utils.DataSet = {
+        "unsigned delegated role": DelegationsTestCase(
+            delegations=[
+                TestDelegation("targets", "A"),
+                TestDelegation("targets", "B"),
+                TestDelegation("A", "C"),
+            ],
+            # The traversal stops after visiting an invalid role
+            visited_order=["A"],
```
jku: could even rename "A" as "invalid" or something to make it clearer what the invalid role is?

tests/test_updater_delegation_graphs.py
```diff
@@ -251,16 +277,119 @@ def test_graph_traversal(self, test_data: DelegationsTestCase) -> None:
...
+
+class TestTargetFileSearch(TestDelegations):
+    r"""
+    Create a single repository with the following delegations:
+
+              targets
+         *.doc / \ r/*/*
+              A   B
+            r/x/* / \ r/y/*.zip
+                 C   D
```
jku: a diagram is not a bad idea but
* missing "*.md"
* shortening releases to r is not trivial to figure out

tests/test_updater_delegation_graphs.py
```diff
@@ -251,16 +277,119 @@ def test_graph_traversal(self, test_data: DelegationsTestCase) -> None:
...
+    def setUp(self) -> None:
+        super().setUp()
+        self._init_repo(self.delegations_tree)
+
+    targets: utils.DataSet = {
+        "targetfile": ("targetfile", []),
+        "README.md": ("README.md", ["A"]),
+        "releases/x/*": ("releases/x/x_v1", ["B", "C"]),
+        "releases/y/*.zip": ("releases/y/y_v1.zip", ["B", "D"]),
+    }
```
jku: Would it make sense to include boolean for whether the target is found or not? then you could include some negative tests here as well, using the same delegations_tree. Or do you want to keep them separate?

As an example 
* Add `TestTarget("C", b"x release in C", "releases/x/x_v2")`
* add test case `"target exists, not delegated": ("releases/x/x_v2", [ ... ], False)`

As another example
* Add `TestTarget("C", b"duplicate x release in C", "releases/x/x_v1")`
* existing test case "releases/x/*" now tests that it finds the correct target (and not the one added above). The case name could be improved to explain this

tests/test_updater_delegation_graphs.py
```diff
@@ -251,16 +277,119 @@ def test_graph_traversal(self, test_data: DelegationsTestCase) -> None:
...
+            TestDelegation("targets", "A", paths=["*.doc", "*.md"]),
+            TestDelegation("targets", "B", paths=["releases/*/*"]),
+            TestDelegation("B", "C", paths=["releases/x/*"]),
+            TestDelegation("B", "D", paths=["releases/y/*.zip"]),
+        ],
+        target_files=[
+            TestTarget("targets", b"content", "targetfile"),
+            TestTarget("A", b"content", "README.md"),
+            TestTarget("C", b"content", "releases/x/x_v1"),
+            TestTarget("D", b"content", "releases/y/y_v1.zip"),
```
jku: having the same content is not great if you ever want to check you got the correct file from the correct delegation

tests/test_updater_delegation_graphs.py
```diff
@@ -251,16 +277,119 @@ def test_graph_traversal(self, test_data: DelegationsTestCase) -> None:
...
+            exp_calls = [(role, 1) for role in visited_order]
+            updater = self._init_updater()
+            # Call explicitly refresh to simplify the expected_calls list
+            updater.refresh()
+            self.sim.fetch_tracker.metadata.clear()
+            target = updater.get_targetinfo(targetpath)
+            # Confirm that the expected TargetFile is found
+            assert target is not None
+            exp_target = self.sim.target_files[targetpath].target_file
+            self.assertDictEqual(target.to_dict(), exp_target.to_dict())
```
jku: since content is the same for every test target (see above), this would return true even if we had two different roles serving the same  targetpath (and we got the wrong target because of a bug)

tuf/ngclient/updater.py
```diff
@@ -473,7 +472,10 @@ def _preorder_depth_first_walk(
                 child_roles_to_visit.reverse()
                 delegations_to_visit.extend(child_roles_to_visit)
 
-        if number_of_delegations == 0 and len(delegations_to_visit) > 0:
+        if (
+            len(visited_role_names) > self.config.max_delegations
+            and len(delegations_to_visit) > 0
+        ):
```
```suggestion
        if len(delegations_to_visit) > 0:
```
I like the simplification you did, and I think this now seems clear... if we have delegations to visit at this point, we must have aborted because max delegations or do I misunderstand?

jku (commented):
Looks pretty good to me. I'm not sure if my comments are really change requests -- maybe have a look and give your opinion?

tests/test_updater_delegation_graphs.py
sechkova: True, I'll fix it but now after your comment I'm not sure how long a diagram can remain up to date with the code.

tests/test_updater_delegation_graphs.py
sechkova: So this is pretty much a bug in the test, I added a different content for each test file.

tuf/ngclient/updater.py
sechkova: This is correct, I've missed it.

tests/test_updater_delegation_graphs.py
sechkova: I think extending it for negative cases makes sense but I didn't understand the second example. We add the target files in a dictionary in the simulator so the duplicate will replace the first target?

sechkova:
Thanks for the suggestions, I find them very reasonable, I did several changes to include them. 
There is this one example not clear to me, I've mentioned it in the comment.

tests/test_updater_delegation_graphs.py
jku: oh right, if the target path is the same then the url is the same... you are correct my example does not make sense.

tuf/ngclient/updater.py
```diff
@@ -423,10 +423,12 @@ def _preorder_depth_first_walk(
         # is needed to load and verify the delegated targets metadata.
         delegations_to_visit = [(Targets.type, Root.type)]
         visited_role_names: Set[str] = set()
-        number_of_delegations = self.config.max_delegations
 
         # Preorder depth-first traversal of the graph of target delegations.
-        while number_of_delegations > 0 and len(delegations_to_visit) > 0:
+        while (
+            len(visited_role_names) <= self.config.max_delegations
```
jku, line 429: now that I look at this again... doesn't this handle one more delegation than it used to?

The new interpretation could be the correct one (with the assumption that only delegated targets are counted so "targets" itself is not a delegation), just wanted to make sure we're not making a mistake. The test you added for max delegations feels correct to me so I guess I'm ok with this interpretation.

jku (approved):
Yeah, I like it. TestTargetFileSearch is quite complex and requires reader to spend some time to figure it out but I think that's just because this is the most complex thing the client does.

Left a comment about the delegation loop change, please have look. I think your version is fine though.

tuf/ngclient/updater.py
sechkova: Yeah, in the previous implementation "targets" was counted in the delegations number which I noticed only after adding the test. So I changed a bit the implementation using "<=" here to compensate for targets. I think I mentioned it in the commit message but I can try adding a comment?

jku: ah you did, I just did not understand the connection...

Issue #1696, opened 2021-11-26
Metadata API: Provide a way to validate object when serializing to dictionary


MVrachev:
**Description of issue or feature request**:

_This issue is discussed during a couple of conversations with @jku._

We can say we are almost done with Metadata API validation during the initialization of `Signed` objects as summarized here: https://github.com/theupdateframework/python-tuf/issues/1140#issuecomment-971588922.

What we **didn't** focus on is validation when serializing the `Signed` objects to dictionaries through `Signed.to_dict()`.
This could be useful for users of Metadata API that changed the signed portion 

I imagine some of the requirements for such validation are:
1. It's optional. As pointed out by @jku in our conversation the client using Metadata API could provide its own validation on the metadata objects. Additionally, if this kind of validation is mandatory it could slow the work of a big python-tuf client such as Warehouse working on thousands of targets.
2.  It doesn't duplicate validation code. It will be best if we find a way to use the same exact code as the mandatory validation during the initialization without duplicating it.

One thing to consider is doing we want to add a little more complex validation, additional to the one done during initialization.
For example, for Root, it probably could be useful to validate that each of the set of `keyid`s defined in the `keys` dictionary is all used inside a particular role in `roles`.

**Current behavior**:
No validation is done when calling `to_dict()` from any of the `Signed` child classes.

**Expected behavior**:
Provide a way to validate an object before calling `to_dict()` or when calling `to_dict()` with an explicit option set.

PS: I didn't include the signatures part from the Metadata objects in this discussion as their validation is a little more complex as you need access to the delegator.

MVrachev:
I decided to create one prototype of a possible solution: https://github.com/MVrachev/tuf/tree/serialization-validation.

jku:
I agree with the basic premise: we'll want to re-evaluate whether our validation is appropriate for the serialization as well (while we've been concentrating on the client case, the focus has been on validating deserialized data).

some quick comments:
* the idea of validating when creating a dict is not necessarily the correct one:` to_dict()` is a helper that json serialization happens to use. There's no guarantee that doing so makes sense for some possible other serialization format. We might want to consider validating during `Metadata.to_bytes()` if possible
* a potential way to 100% prevent incorrect serialization (instead of just checking e.g. specific data types) is during `Metadata.to_bytes()` to serialize Metadata to bytes, then deserialize back to Metadata, and raise if the two metadata instances are not equal.
  * This might be too heavy (cpu/memory use) to be a default choice but could be an option
[...]

PR #1712, opened 2021-12-07
Remove microseconds from metadata API Signed.expires

### LINKED ISSUE #1678: metadata api: Signed.expires string representation should not contain microseconds

 In `signed._common_fields_to_dict` we convert the value of the datetime object in `Signed.expires` to a date-time string.

From the [TUF spec](https://theupdateframework.github.io/specification/latest/#date-time):
> [...] The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". [...] An example date-time string is "1985-10-21T01:21:00Z".

If the original datetime object contained microseconds, the date-time string representation of `Signed.expires` will too.

**Reproducer:**
```python
from datetime import datetime
from tuf.api.metadata import Timestamp, MetaFile
Timestamp(1, "1.0.19", datetime(2021, 11, 16, 12, 0, 0, 123456), MetaFile(1)).to_dict()["expires"]
```

**Current output**:

```
'2021-11-16T12:00:00.123456Z'
```

**Expected behavior**:

```
'2021-11-16T12:00:00Z'
```

**Proposed fix 1**:
```diff
diff --git a/tuf/api/metadata.py b/tuf/api/metadata.py
index 25f14fe7..cdbd8595 100644
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -480,7 +480,7 @@ class Signed(metaclass=abc.ABCMeta):
             "_type": self._type,
             "version": self.version,
             "spec_version": self.spec_version,
-            "expires": self.expires.isoformat() + "Z",
+            "expires": self.expires.replace(microsecond=0).isoformat() + "Z",
             **self.unrecognized_fields,
         }
```

**Proposed fix 2**:
```diff
diff --git a/tuf/api/metadata.py b/tuf/api/metadata.py
index 25f14fe7..20365cfe 100644
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -480,7 +480,7 @@ class Signed(metaclass=abc.ABCMeta):
             "_type": self._type,
             "version": self.version,
             "spec_version": self.spec_version,
-            "expires": self.expires.isoformat() + "Z",
+            "expires": datetime.strftime(self.expires, "%Y-%m-%dT%H:%M:%SZ"),
             **self.unrecognized_fields,
         }
```

ivanayov:
This change removes microseconds from expiry in order to fit TUF
specification

Fixes #1678

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -200,6 +200,9 @@ class TestSerialization(unittest.TestCase):
                 "snapshot": {"keyids": ["keyid2"], "threshold": 1}, \
                 "foo": {"keyids": ["keyid2"], "threshold": 1}} \
             }',
+        "invalid expiry with microsecond    s": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
+            "expires": "2030-01-01T12:00:00.123456Z", "consistent_snapshot": false, \
+            "keys": {}, "roles": {"root": {}, "timestamp": {}, "targets": {}, "snapshot": {}}}',
     }
 
     @utils.run_sub_tests_with_dataset(invalid_roots)
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -408,6 +408,14 @@ class Signed(metaclass=abc.ABCMeta):
     def _type(self) -> str:
         return self.type
 
+    @property
+    def expires(self) -> datetime:
+        return self._expires
+
+    @expires.setter
+    def expires(self, value: datetime) -> None:
+        self._expires = value.replace(microsecond=0)
+
     # NOTE: Signed is a stupid name, because this might not be signed yet, but
     # we keep it to match spec terminology (I often refer to this as "payload",
     # or "inner metadata")
@@ -469,6 +477,7 @@ class Signed(metaclass=abc.ABCMeta):
         # what the constructor expects and what we store. The inverse operation
         # is implemented in '_common_fields_to_dict'.
         expires = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+
         return version, spec_version, expires
 
     def _common_fields_to_dict(self) -> Dict[str, Any]:


jku:
We'll want think this through for all the different cases: deserialization, users calling the api , serialization... it's a little detail but could make life difficult in the future
* I'd want to see docstrings for public API at least
* is it correct that we accept non-spec-compliant dates on deserialize?
* is it correct that we change the date to compliant on deserialize?

ivanayov:
@jku 
> is it correct that we accept non-spec-compliant dates on deserialize?

It think that spec-compliant should mean spec-compliant with no exception, so if metadata is not in the correct form, we should throw an error.

> is it correct that we change the date to compliant on deserialize?

I think it's fine, when it comes to preventing a bug, coming not from the metadata, but from the implementation (and in this case - python specific)   as we guarantee that it won't break the deserialisation. But we should guarantee that the metadata is correct at the same time. 
May be I'd need to add a check, which is now missing.

lukpueh:
I agree with @jku that we need to be careful here. This is what I think about the different cases:

* deserialization:
  The deserializer shouldn't accept non-spec compliant metadata, regardless of its importance. If we start modifying the semantics of inputs during deserialization, we'll likely end up with determinism problems (goal: `data == serialize(deserialize(data)`)

* Object API:
  Ideally, we don't accept non-sepc compliant inputs here either. But in reality this is hard to enforce, because properties can be set in different ways. So I think it is easier and okay to only enforce spec compliance on the de/serialization boundaries.

* Serialization:
  Under above assumptions, i.e. that we don't accept non-spec compliant metadata in deserialization, and that some properties that were modified during object life-time contain additional (unneeded) information, certain transformations should be fine, e.g. removing unnecessary microseconds. We need to find a balance between expected behaviour and usability.

jku:
I think I lean in the same direction (EDIT: I wrote this before seeing lukas' last message :) )
* spec defines the date format exactly: we should only accept that in deserialization?
* constructing a new object should probably accept microseconds and just remove them
* deserialization and constructing new object currently use same code path (in the constructor)

so possibly 
* from_dict() should be strict, and should fail if format is not exactly as defined in spec
* but the constructor should use set_expiry() as your PR does: so constructing new object does not fail on microseconds ...

I'm not 100% sure of this yet, but that's my thinking...

lukpueh:
I agree with the overall direction...

> from_dict() should be strict, and should fail if format is not exactly as defined in spec

This makes sense, and we might need to revise existing data validation in `from_dict/__init__`-methods, because I saw that some is performed in one and some in the other.

> but the constructor should use set_expiry() as your PR does: so constructing new object does not fail on microseconds

I guess we can do that, but we need to to be careful that there's no unexpected behavior, above all upon serialization, if the expires field is modified outside of the constructor

jku:
> but we need to to be careful that there's no unexpected behavior, above all upon serialization, if the expires field is modified outside of the constructor

Yes, and should consider buggy code as well (obviously we'd never write bugs but just theoretically speaking): we should ensure we rather fail than persist metadata that is invalid. #1696 has some discussion on this

ivanayov:
In the latest update I stick to:
* `__ini__()` uses `set_expiry()`, so that we guarantee microseconds are not added on runtime
* a format verification for `"expires"` in `_common_fields_from_dict()` 
* no format changes (which I pushed earlier) in `_common_fields_to_dict()`

tuf/api/metadata.py
```diff
@@ -467,7 +470,16 @@ def _common_fields_from_dict(
         # Convert 'expires' TUF metadata string to a datetime object, which is
         # what the constructor expects and what we store. The inverse operation
         # is implemented in '_common_fields_to_dict'.
-        expires = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+        try:
+            expires = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+        except ValueError:
+            raise exceptions.InvalidConfigurationError(
+                f"""
+            Role {_type} should have "expires" metadata of the format
```
kairoaraujo: There is a white trailing space at the end of this line.

kairoaraujo (commented):
To test your code (lazily), I used the ``tests/metadata_api_tests/test_metadata_serialization.py`` and I added an invalid expire date entry to this data set https://github.com/theupdateframework/python-tuf/blob/d3b877b38ae0e73ff808003eb4321c6c5749ad1d/tests/test_metadata_serialization.py#L169

```python
        "invalid expire date": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
            "expires": "2021-11-16T12:00:00.123456Z", "consistent_snapshot": false, \
            "keys": { \
                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
            "roles": { \
                "root": {"keyids": ["keyid1"], "threshold": 1}, \
                "timestamp": {"keyids": ["keyid2"], "threshold": 1}, \
                "targets": {"keyids": ["keyid1"], "threshold": 1}, \
                "snapshot": {"keyids": ["keyid2"], "threshold": 1}} \
            }',
```

It gave me the expected raise:
```python
  File "/Users/kdearaujo/devel/TUF/python-tuf/tuf/api/metadata.py", line 476, in _common_fields_from_dict
    raise exceptions.InvalidConfigurationError(
tuf.exceptions.InvalidConfigurationError: 
            Role root should have "expires" metadata of the format 
            "yyyy-mm-ddThh:mm:ssZ" but got 2021-11-16T12:00:00.123456Z instead
```

But I am not sure if the right to a place to add the test or even we want to test it there. I noticed that we don't test these custom exceptions in ``tests_metadata_serialization.py``

Maybe some maintainer could comment :)

tuf/api/metadata.py
ivanayov: @kairoaraujo Yep, it's on purpose, as this is printed on a single line.

ivanayov: Regarding the tests - thank you for the notice! I was also thinking of adding one, but was waiting for feedback first to see if we agree on this approach.
`test_metadata_serialization.py` looks a good place to me. May be we should sync with @MVrachev as he's also working on validation.

kairoaraujo: My suggestion:

```python
        except ValueError:
            raise exceptions.InvalidConfigurationError(
                f"Role {_type} should have 'expires' metadata of the format "
                + f"'yyyy-mm-ddThh:mm:ssZ' but got {expires_str} instead"
            ) from None
```

lukpueh: +1 fo Kairo's suggestion. I'd generally not add newlines inside error/log messages unless they help readability of the message (not of the code).

One tiny addition, the `+` is not required for string concatenation inside braces.

lukpueh: Two more comments about the error:
- `InvalidConfigurationError` does not seem like the right exception type. I'd probably just re-raise the `ValueError` for now (see #1131)
- nit: I think the correct ISO8601 spelling would use upper case for the date part, i.e. "YYYY-MM-DDThh:mm:ssZ". But it seems that the [TUF spec](https://theupdateframework.github.io/specification/latest/#date-time) uses all upper case, i.e.  "YYYY-MM-DDTHH:MM:SSZ". I suggest you use either one of those, but I won't insist. :)

lukpueh:
Thanks for working on this @ivanayov. I think it's a fair approach for the first two use cases discussed above. We still need to figure out what to do for the 3rd use case, that is do we allow serialisation of invalid metadata. But maybe we don't have to do this here. Would you mind adding a test for what you have so far?

ivanayov:
@lukpueh @kairoaraujo Just pushed an update with latest comments.

tuf/api/metadata.py
```diff
@@ -468,7 +471,14 @@ def _common_fields_from_dict(
...
         # what the constructor expects and what we store. The inverse operation
         # is implemented in '_common_fields_to_dict'.
-        expires = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+        try:
+            expires = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+        except ValueError:
+            raise ValueError(
+                f"Role {_type} should have 'expires' metadata of the format "
+                f"'YYYY-MM-DDThh:mm:ssZ' but got {expires_str} instead"
+            ) from None
```
jku: I can see you've discussed this already (and I agree with using just a plain ValueError)... but what's wrong with the original ValueError thrown by strptime? I don't see an advantage to adding this code.

tests/test_api.py
```diff
@@ -748,6 +748,19 @@ def test_is_delegated_role(self) -> None:
             self.assertFalse(role.is_delegated_path("a/non-matching path"))
             self.assertTrue(role.is_delegated_path("a/path"))
 
+    def test_invalid_expiration_date(self) -> None:
```
jku: I don't think this test is necessary -- the serialization test seems good enough to me (or at least we should be able to explain why we don't add hundreds of other tests like this for invalid json)

tests/test_metadata_serialization.py
```diff
@@ -200,6 +200,17 @@ def test_root_serialization(self, test_case_data: str) -> None:
                 "snapshot": {"keyids": ["keyid2"], "threshold": 1}, \
                 "foo": {"keyids": ["keyid2"], "threshold": 1}} \
             }',
+        "invalid expire date": '{"_type": "root", "spec_version": "1.0.0", "version": 1, \
+            "expires": "2030-01-01T12:00:00.123456Z", "consistent_snapshot": false, \
+            "keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
+                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
```
jku: some non-important nitpicks here:
* name doesn't say why it's invalid, why not say "invalid expiry with microsecs" ?
* keys could be an empty dict (and the roles keyids lists could be empty) to make this test data smaller

tuf/api/metadata.py
```diff
@@ -429,13 +429,16 @@ def __init__(
...
         self.spec_version = spec_version
-        self.expires = expires
+        self.set_expiry(expires)
 
         if version <= 0:
             raise ValueError(f"version must be > 0, got {version}")
         self.version = version
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
 
+    def set_expiry(self, expires: datetime) -> None:
```
jku: public method without docstring

There are also other methods dealing with expiry already in this class. This method should be grouped with them I think

jku (changes requested):
I left some comments, I think the core of this PR is the new API (`set_expiry()`) which allows both construction of new metadata and modifications to existing expiry to be done safely: I think this is the correct direction but new API needs to to be polished (include API documentation and so on).

The big question is, do we need to solve the related #1727 at the same time (at least on paper)?

jku:
> We still need to figure out what to do for the 3rd use case, that is do we allow serialisation of invalid metadata

Martin opened #1696 about this relevant question in general: maybe we can continue there (the issue is big enough to maybe have a chat as well, to discuss some possible ways to deal with it)

jku:
oh a new idea: since one of the problems here is that our expiry is a datetime but most datetimes are not valid expiries (because microseconds), how about we hide the variable and make `expiry` a property (where the setter is Ivanas function) -- that would effectively prevent a user from accidentally setting `signed.expiry` to an invalid value: something like `signed.expiry = utcnow() + relativedelta(day=3)` would now work correctly because it would use the property setter.

tuf/api/metadata.py
lukpueh: I had the same thought, but I guess there's merit in a contextualised error message... of course, there's also merit in 7 lines of code less. :)

ivanayov: I see your point (as I was thinking the same), but in the end thought that general ValueError doesn't give information on what's expected. We can always say that one may just check the spec, so I'm fine with removing the change if you both vote for that.

jku (approved):
This feels like the correct solution, thank you.
* API user just provides a datetime, we make sure it's valid (removing microseconds) 
* the Metadata object always matches the serialized output (since the microseconds are never stored in the object)
* type "enforcement" is about as good as it gets with python (static checks will catch wrong types and we use the datetime API in the setter so even without static checks things will break on type mismatches) 

Looks good to me, leaving open for @lukpueh to see as well

lukpueh (approved):
Seconding Jussi's assessment. Thanks for the clean fix, @ivanayov!

PR #1721, opened 2021-12-10
Start linting tuf/exceptions.py with all linters

MVrachev:

We are already using tuf/exceptions.py inside
- `tuf/api/metadata.py`
- `tuf/ngclient/__internal__/trusted_metadata_set.py`
- `tuf/ngclient/fetcher.py`
- `tuf/ngclient/updater.py`
- and in our tests using the new code base

This means that we are most likely going to use that
module after we remove the old code.

It makes sense to lint it with the other three
linters besides `mypy` (we already lint tuf/exceptions.py with `mypy`).

requirements**:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -38,7 +38,7 @@ commands =
 
 [testenv:lint]
 changedir = {toxinidir}
-lint_dirs = tuf/api tuf/ngclient examples tests
+lint_dirs = tuf/api tuf/ngclient examples tests tuf/exceptions.py
 commands =
     # Use different configs for new (tuf/api/*) and legacy code
     black --check --diff {[testenv:lint]lint_dirs}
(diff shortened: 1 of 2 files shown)


kairoaraujo (commented):
We have some commented code as 'Directly instance-repoducing'. Any reason to keep it?

jku:
Alternatively 
* copy exceptions.py to wherever it should live in the future (tuf/api/exceptions.py maybe?)
* remove all of the unused exceptions and cruft from that copy
* make the code actually follow the style guide (and not just pass the linter)
* apply all the fixes you did here
* start importing the new copy in api & ngclient
 
I'm not saying this needs to all happen in one PR, but if you start with making the copy and removing old code, then there's far less code to fix/review and you don't need to deal with pylint disables: the way the PR handles disables almost certainly disabled warnings from all of the old code base, not just exceptions.py

MVrachev:
> Alternatively
> 
> * copy exceptions.py to wherever it should live in the future (tuf/api/exceptions.py maybe?)
> * remove all of the unused exceptions and cruft from that copy
> * make the code actually follow the style guide (and not just pass the linter)
> * apply all the fixes you did here
> * start importing the new copy in api & ngclient
> 
> I'm not saying this needs to all happen in one PR, but if you start with making the copy and removing old code, then there's far less code to fix/review and you don't need to deal with pylint disables: the way the PR handles disables almost certainly disabled warnings from all of the old code base, not just exceptions.py

Yes, I can create a new module in `tuf/api/exceptions.py` and copy the relevant exceptions and follow all the steps.
Will do that in the coming days, but it will be as a side item as I have additional goals for the sprint.

MVrachev:
This pr is superseded by #1725 where I actually add a new `exceptions.py` file as @jku suggested.

Issue #1722, opened 2021-12-13
Metadata API: Remove dependency on dateutil


jku:
PR #1710 added a new package dependency (dateutil) to Metadata API: This was likely not intended, and was not caught by the tests because dateutil is a test dependency. 

I think we should fix this before releasing 0.20 (#1694).

jku:
I think bump_expiration() as a method is not super useful.

Current test code looks like this:
```
        delta = relativedelta(years=1)
        timestamp.signed.bump_expiration(delta)

```
Same thing without using the helper function:
```
        timestamp.signed.expires += relativedelta(years=1)
```
this is in my opinion at least as readable and works fine without the API depending on dateutil.

jku:
This breaks RTD documentation:

> WARNING: autodoc: failed to import module 'metadata' from module 'tuf.api'; the following exception was raised:
> No module named 'dateutil'

MVrachev:
I agree we can remove `Signed.bump_expiration()`.
If we have an agreement I can propose a pr for this one today, so we can move forward with #1723.
What do you think @lukpueh?

jku:
we don't need to remove the function to fix the immediate issue.

Let's fix the dependency issue first, then spend the time we need to get the API (and/or docs) right before 1.0 release: as an example I only now realized the test (and the function) are also bad examples: I don't think anyone ever wants to bump to `old_expiry_date+delta`, real users want to bump to `current_date+delta`... which leads to a potential issue with microseconds.

I suggest we 
* modify the tests to use something like my suggestion (even if it's a bit bad example as mentioned)
* remove mentions of dateutil from Metadata API
* file an issue to improve Metadata expiry API

MVrachev:
An issue where we can discuss if we want `Signed.bump_expiry()` https://github.com/theupdateframework/python-tuf/issues/1727

Issue #1709, opened 2021-12-06
doc: misc enhancements to contributors


lukpueh:
**Description of issue or feature request**:
Here are a few suggestions to enhance our [Instructions for Contributors](https://theupdateframework.readthedocs.io/en/latest/CONTRIBUTORS.html) on RTD.

**Current behavior**:
- No mention of Secure Systems Lab development guidelines
- Too extensive (and prominent) section about DCO
- Too extensive (and outdated/opinionated?) virtualenv description
- Very extensive description of test logging/verbosity config 

**Expected behavior**:
- Link to SSLab [development guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/dev-workflow.md)
- Shorten DCO section (and optionally move down in doc) (see e.g. [in-toto Instructions for Contributors](https://github.com/in-toto/in-toto#instructions-for-contributors) for inspiration)
- Just recommend usage of and link to existing virtualenv documentation
- Shorten description of test logging/verbosity config

jku:
> * Shorten description of test logging/verbosity config

#1724 is going to make `tox` run a reasonable approximation of CI tests so we should suggest that as the way to run tests.

Then after that we can include a _short_ mention on how to run specific tests and the options (`-v` for logs, `--dump` for repository dumps when available)

PR #1724, opened 2021-12-13
turn doc build warnings into errors

kairoaraujo:
This commit adds to the docs tox session the flag ``-W``, which
turns the warnings into errors.

The CI will fail once it gets errors.

Linked to #1722

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,6 +9,7 @@ version: 2
 sphinx:
    builder: html
    configuration: docs/conf.py
+   fail_on_warning: true
 
 # Optionally build your docs in additional formats such as PDF
 formats: []
--- a/docs/CONTRIBUTORS.rst
+++ b/docs/CONTRIBUTORS.rst
@@ -184,10 +184,9 @@ a *venv*), and then install ``securesystemslib`` in editable mode too (in the sa
     $ python3 -m pip install -r requirements-dev.txt
 
 
-With `tox <https://testrun.org/tox/>`_ the test suite can be executed in a
-separate *venv* for each supported Python version. While the supported
-Python versions must already be available, ``tox`` will install ``tuf`` and its
-dependencies anew in each environment.
+With `tox <https://testrun.org/tox/>`_, the test suite can be executed in a
+separate *venv*. While a supported Python version must already be available,
+``tox`` will install ``tuf`` and dependencies.
 ::
 
     $ tox
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = lint,py{36,37,38,39,310}
+envlist = lint,docs,py
 skipsdist = true
 
 [testenv]
--- a/tox.ini
+++ b/tox.ini
@@ -59,4 +59,4 @@ deps =
 
 changedir = {toxinidir}
 commands =
-    sphinx-build -b html docs docs/build/html
+    sphinx-build -b html docs docs/build/html -W


sechkova:
While I think this is a good change and we should keep it, we have the docs environment in tox only for local builds. In the CI we use the integrated webhooks from readthedocs.
So in addition to your change, we need to update the `.readthedocs.yaml` file with the `fail_on_warning` option: https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx

sechkova (changes requested):
While I think this is a good change and we should keep it, we have the docs environment in tox only for local builds. In the CI we use the integrated webhooks from readthedocs.
So in addition to your change, we need to update the .readthedocs.yaml file with the fail_on_warning option: https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx

kairoaraujo:
Thank you @sechkova, I see now the failure in the checks.

sechkova (approved):
It makes sense to me.

jku:
Well CI shows it failing as intended...

I'm not sure about this though: reproducing tests locally is pretty important (sending PRs only to get red lights is demoralizing) and our testing setup is already a bit complicated in that regard: `tox -e lint,py` is probably currently the easiest way to test 99% of what CI tests. With this commit it becomes `tox -e lint,docs,py` which is a bit much...

Could we set default `envlist = lint,docs,py` so developers can just call `tox` and it does approximately the right thing? or does someone actually have multiple pythons installed they want to test?

The CI already shouldn't need the envlist as I think it it explicitly sets toxenv to `py` (so uses the current system python)

kairoaraujo:
> Could we set default `envlist = lint,docs,py` so developers can just call `tox` and it does approximately the right thing? or does someone actually have multiple pythons installed they want to test?
> 
> The CI already shouldn't need the envlist as I think it it explicitly sets toxenv to `py` (so uses the current system python)

+1 to set default `envlist = lint,docs,py` and make it straightforward for contributors running `tox`.
I particularly have multiple Python versions installed. Still, I think those with multiple Python versions should explicitly use the ``tox -e py39,py310`` (my case) instead of the default.

jku:
Yeah that sounds good to me. 

 build fix is in (#1726), if you rebase you should also get a green build now...

lukpueh:
> or does someone actually have multiple pythons installed they want to test?

I do, yes. But I'm also fine in changing the way I invoke tox to do it. 

Given this change we should also update the note about tox in the [contributors docs](https://theupdateframework.readthedocs.io/en/latest/CONTRIBUTORS.html) (see bottom). But we can add that to https://github.com/theupdateframework/python-tuf/issues/1709. @kairoaraujo, would you mind adding a note there?

kairoaraujo:
> Given this change we should also update the note about tox in the [contributors docs](https://theupdateframework.readthedocs.io/en/latest/CONTRIBUTORS.html) (see bottom). But we can add that to #1709. @kairoaraujo, would you mind adding a note there?

I don't mind adding a note there, but can someone help me with some decisions?

Current doc:
```
With `tox <https://testrun.org/tox/>`_ the test suite can be executed in a
separate *venv* for each supported Python version. While the supported
Python versions must already be available, ``tox`` will install ``tuf`` and its
dependencies anew in each environment.

    $ tox
```
 
My initial suggestion is:

```
With `tox <https://testrun.org/tox/>`_ the test suite can be executed in a
separate *venv* for the default installed Python version. While the supported
Python version must already be available, ``tox`` will install ``tuf`` and its
dependencies anew.

    $ tox
```

About the part that mentions tox will install tuf's "dependencies anew in each environment.".
That is true for the first run, but it will not automatically recreate that after it. It will reuse it.

Some suggestions:
 - add a mention if you are changing any of the requirements to use ``tox -r`` to recreate the *venv*
 - OR add the  example as ``tox -r``
 - OR add ``recreate = true`` to tox.ini have recreation for every run [IMHO -1]

Should I also add a piece of information about using an alternative installed Python version?

Suggestion
```
Running the test with tox and alternative installed Python version is possible using  ``-e py<version>``.

   $ tox -e py310
```

jku:
As general advice: We should aim to make the document (quite a bit) shorter, not longer.

I can try to come up with actual suggestions later but maybe that helps already?

kairoaraujo:
I added a small and simple suggestion.

jku (approved):
LGTM, thanks

Issue #1646, opened 2021-10-27
metadata API: avoid raising securesystemslib exceptions


jku:
I think Metadata API should not require its users to import securesystemlib exceptions to be able to call Metadata API methods, it just seems weird -- as a user of TUF I am not necessarily directly using securesystemslib at all, yet I need to handle exceptions from it

**Current behavior**:

sign(), from_file() and to_file() are documented to raise these exceptions

**Expected behavior**:

I should only need to import modules from tuf to be able to handle tuf exceptions. This could mean either 
* re-exporting the securesystemslib exceptions in tuf metadata or exceptions module or
* handling the securesystemslib exceptions and raising something in python-tuf: this is what I think `verify_delegate()` does

MVrachev:
I agree with the objective of the issue. 
I think each of the cases should be solved separately as a securesystemslib exception can describe exactly what we want to tell the user in an eventual problem and other times the securesystemslib exception will be an implementational detail.

jku:
After some discussion last week and looking at this again: I think at least sign() should operate like the verify functions: handle the sslib exceptions and raise just a single exception (UnsignedMetadataError seems appropriate). I've been trying to imagine a scenario where a repository would like to handle the individual errors... and I cannot think of any, they don't seem useful.

PR #1725, opened 2021-12-13
Add new exceptions file for exceptions in the new code

MVrachev:
Addresses @jku comment here: https://github.com/theupdateframework/python-tuf/pull/1721#issuecomment-992311731

Add tuf/api/exceptions.py for exceptions in the new code.
I copied the exceptions from tuf/exceptions.py with a few important
decisions:
1. I only added the exceptions that are used in the new code. 
2. I removed the general "Error" class as we can directly inherit
Exceptions. 
3. I tried grouping the exceptions by relevance. 
4. I removed the second argument `signable` in `UnsignedMetadataError` as it was only
kept for backward compatibility and is not used. (See https://github.com/theupdateframework/python-tuf/commit/929b4b2a6b5c499ab741feb1233fdd7f4036b759 where it was added in `tuf/exceptions.py`)
5. I tried following the new code style guidelines and linted the file
with our linters.
6. Made `FetcherHTTPError` a `DownloadError` 
7. Removed `URLParsingError` and replaced it with a `DownloadError`
8. Removed `ReplayedMetadataError`  and replaced it with `BadVersionNumberError`

Additionally, I stopped linting tuf/exceptions.py with `mypy` as we are going to use
`tuf/api/exceptions.py` for exceptions in the new code.

requirements**:

### DIFF
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -8,13 +8,6 @@ there is a good reason not to, and provide that reason in those cases.
 """
 
 
-#### General errors ####
-
-
-class UnsupportedAlgorithmError(Exception):
-    """An error while trying to identify a user-specified algorithm."""
-
-
 class LengthOrHashMismatchError(Exception):
     """An error while checking the length and hash values of an object."""
 
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -19,20 +19,6 @@ class LengthOrHashMismatchError(Exception):
     """An error while checking the length and hash values of an object."""
 
 
-class FetcherHTTPError(Exception):
-    """
-    Returned by FetcherInterface implementations for HTTP errors.
-
-    Args:
-        message: The HTTP error messsage
-        status_code: The HTTP status code
-    """
-
-    def __init__(self, message: str, status_code: int):
-        super().__init__(message)
-        self.status_code = status_code
-
-
 class URLParsingError(Exception):
     """If we are unable to parse a URL -- for example, if a hostname element
     cannot be isoalted."""
@@ -97,3 +83,17 @@ class DownloadLengthMismatchError(DownloadError):
 
 class SlowRetrievalError(DownloadError):
     """Indicate that downloading a file took an unreasonably long time."""
+
+
+class FetcherHTTPError(DownloadError):
+    """
+    Returned by FetcherInterface implementations for HTTP errors.
+
+    Args:
+        message: The HTTP error messsage
+        status_code: The HTTP status code
+    """
+
+    def __init__(self, message: str, status_code: int):
+        super().__init__(message)
+        self.status_code = status_code
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -19,11 +19,6 @@ class LengthOrHashMismatchError(Exception):
     """An error while checking the length and hash values of an object."""
 
 
-class URLParsingError(Exception):
-    """If we are unable to parse a URL -- for example, if a hostname element
-    cannot be isoalted."""
-
-
 #### Repository errors ####
 
 
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -1,0 +1,155 @@
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""
+Define TUF exceptions used inside the new modern implementation.
+The names chosen for TUF Exception classes should end in 'Error' except where
+there is a good reason not to, and provide that reason in those cases.
+"""
+
+from typing import Optional
+
+#### General errors ####
+
+
+class UnsupportedAlgorithmError(Exception):
+    """An error while trying to identify a user-specified algorithm."""
+
+
+class LengthOrHashMismatchError(Exception):
+    """An error while checking the length and hash values of an object."""
+
+
+class FetcherHTTPError(Exception):
+    """
+    Returned by FetcherInterface implementations for HTTP errors.
+
+    Args:
+        message: The HTTP error messsage
+        status_code: The HTTP status code
+    """
+
+    def __init__(self, message: str, status_code: int):
+        super().__init__(message)
+        self.status_code = status_code
+
+
+class URLParsingError(Exception):
+    """If we are unable to parse a URL -- for example, if a hostname element
+    cannot be isoalted."""
+
+
+class RepositoryError(Exception):
+    """An error with a repository's state, such as a missing file."""
+
+
+#### Repository errors ####
+
+
+class UnsignedMetadataError(RepositoryError):
+    """An error about metadata object with insufficient threshold of signatures.
+
+    Args:
+        message: The error message
+    """
+
+    def __init__(self, message: str) -> None:
+        super().__init__()
+        self.exception_message = message
+
+    def __str__(self) -> str:
+        return self.exception_message
+
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + " : " + str(self)
+
+
+class BadVersionNumberError(RepositoryError):
+    """An error for metadata that contains an invalid version number."""
+
+
+class ExpiredMetadataError(RepositoryError):
+    """Indicate that a TUF Metadata file has expired."""
+
+
+class ReplayedMetadataError(RepositoryError):
+    """Indicate that some metadata has been replayed to the client.
+
+    Args:
+        metadata_role: Name of the role that has been replayed
+        downloaded_version: The replayed downloaded version of the metadata
+        current_version: The current locally available version.
+    """
+
+    def __init__(
+        self, metadata_role: str, downloaded_version: int, current_version: int
+    ) -> None:
+        super().__init__()
+
+        self.metadata_role = metadata_role
+        self.downloaded_version = downloaded_version
+        self.current_version = current_version
+
+    def __str__(self) -> str:
+        return (
+            "Downloaded "
+            + repr(self.metadata_role)
+            + " is older ("
+            + repr(self.downloaded_version)
+            + ") than the version currently "
+            "installed (" + repr(self.current_version) + ")."
+        )
+
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + " : " + str(self)
+
+
+#### Download Errors ####
+
+
+class DownloadError(Exception):
+    """An error occurred while attempting to download a file."""
+
+
+class DownloadLengthMismatchError(DownloadError):
+    """Indicate that a mismatch of lengths was seen while downloading a file."""
+
+    def __init__(self, expected_length: int, observed_length: int) -> None:
+        super().__init__()
+
+        self.expected_length = expected_length  # bytes
+        self.observed_length = observed_length  # bytes
+
+    def __str__(self) -> str:
+        return (
+            "Observed length ("
+            + repr(self.observed_length)
+            + ") < expected length ("
+            + repr(self.expected_length)
+            + ")."
+        )
+
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + " : " + str(self)
+
+
+class SlowRetrievalError(DownloadError):
+    """ "Indicate that downloading a file took an unreasonably long time."""
+
+    def __init__(self, average_download_speed: Optional[int] = None) -> None:
+        super().__init__()
+        self.__average_download_speed = average_download_speed  # bytes/second
+
+    def __str__(self) -> str:
+        msg = "Download was too slow."
+        if self.__average_download_speed is not None:
+            msg = (
+                "Download was too slow. Average speed: "
+                + repr(self.__average_download_speed)
+                + " bytes per second."
+            )
+
+        return msg
+
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + " : " + str(self)
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -8,10 +8,6 @@ there is a good reason not to, and provide that reason in those cases.
 """
 
 
-class LengthOrHashMismatchError(Exception):
-    """An error while checking the length and hash values of an object."""
-
-
 #### Repository errors ####
 
 
@@ -30,6 +26,10 @@ class BadVersionNumberError(RepositoryError):
 
 class ExpiredMetadataError(RepositoryError):
     """Indicate that a TUF Metadata file has expired."""
+
+
+class LengthOrHashMismatchError(RepositoryError):
+    """An error while checking the length and hash values of an object."""
 
 
 #### Download Errors ####
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -12,7 +12,9 @@ there is a good reason not to, and provide that reason in those cases.
 
 
 class RepositoryError(Exception):
-    """An error with a repository's state, such as a missing file."""
+    """An error with a repository's state, such as a missing file.
+    It covers all exceptions that come from the repository side when
+    looking from the perspective of users of metadata API or ngclient."""
 
 
 class UnsignedMetadataError(RepositoryError):
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -39,32 +39,6 @@ class ExpiredMetadataError(RepositoryError):
     """Indicate that a TUF Metadata file has expired."""
 
 
-class ReplayedMetadataError(RepositoryError):
-    """Indicate that some metadata has been replayed to the client.
-
-    Args:
-        metadata_role: Name of the role that has been replayed
-        downloaded_version: The replayed downloaded version of the metadata
-        current_version: The current locally available version.
-    """
-
-    def __init__(
-        self, metadata_role: str, downloaded_version: int, current_version: int
-    ) -> None:
-        super().__init__()
-
-        self.metadata_role = metadata_role
-        self.downloaded_version = downloaded_version
-        self.current_version = current_version
-
-    def __str__(self) -> str:
-        return (
-            f"Downloaded {self.metadata_role} is older ("
-            f"{self.downloaded_version}) than the version currently installed"
-            f"({self.current_version})"
-        )
-
-
 #### Download Errors ####
 
 
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -7,7 +7,6 @@ The names chosen for TUF Exception classes should end in 'Error' except where
 there is a good reason not to, and provide that reason in those cases.
 """
 
-from typing import Optional
 
 #### General errors ####
 
@@ -39,29 +38,16 @@ class URLParsingError(Exception):
     cannot be isoalted."""
 
 
+#### Repository errors ####
+
+
 class RepositoryError(Exception):
     """An error with a repository's state, such as a missing file."""
 
 
-#### Repository errors ####
-
-
 class UnsignedMetadataError(RepositoryError):
-    """An error about metadata object with insufficient threshold of signatures.
-
-    Args:
-        message: The error message
-    """
-
-    def __init__(self, message: str) -> None:
-        super().__init__()
-        self.exception_message = message
-
-    def __str__(self) -> str:
-        return self.exception_message
-
-    def __repr__(self) -> str:
-        return self.__class__.__name__ + " : " + str(self)
+    """An error about metadata object with insufficient threshold of
+    signatures."""
 
 
 class BadVersionNumberError(RepositoryError):
@@ -92,16 +78,10 @@ class ReplayedMetadataError(RepositoryError):
 
     def __str__(self) -> str:
         return (
-            "Downloaded "
-            + repr(self.metadata_role)
-            + " is older ("
-            + repr(self.downloaded_version)
-            + ") than the version currently "
-            "installed (" + repr(self.current_version) + ")."
+            f"Downloaded {self.metadata_role} is older ("
+            f"{self.downloaded_version}) than the version currently installed"
+            f"({self.current_version})"
         )
-
-    def __repr__(self) -> str:
-        return self.__class__.__name__ + " : " + str(self)
 
 
 #### Download Errors ####
@@ -114,42 +94,6 @@ class DownloadError(Exception):
 class DownloadLengthMismatchError(DownloadError):
     """Indicate that a mismatch of lengths was seen while downloading a file."""
 
-    def __init__(self, expected_length: int, observed_length: int) -> None:
-        super().__init__()
-
-        self.expected_length = expected_length  # bytes
-        self.observed_length = observed_length  # bytes
-
-    def __str__(self) -> str:
-        return (
-            "Observed length ("
-            + repr(self.observed_length)
-            + ") < expected length ("
-            + repr(self.expected_length)
-            + ")."
-        )
-
-    def __repr__(self) -> str:
-        return self.__class__.__name__ + " : " + str(self)
-
 
 class SlowRetrievalError(DownloadError):
-    """ "Indicate that downloading a file took an unreasonably long time."""
-
-    def __init__(self, average_download_speed: Optional[int] = None) -> None:
-        super().__init__()
-        self.__average_download_speed = average_download_speed  # bytes/second
-
-    def __str__(self) -> str:
-        msg = "Download was too slow."
-        if self.__average_download_speed is not None:
-            msg = (
-                "Download was too slow. Average speed: "
-                + repr(self.__average_download_speed)
-                + " bytes per second."
-            )
-
-        return msg
-
-    def __repr__(self) -> str:
-        return self.__class__.__name__ + " : " + str(self)
+    """Indicate that downloading a file took an unreasonably long time."""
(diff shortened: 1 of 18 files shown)


MVrachev:
The only question that is left is do we think that `tuf/api/exceptions.py` is the right place for that file as it will be used from `tuf/ngclient` as well?
One possibility is renaming the old exceptions file to `exceptions.old` for example and moving the new exception file there so its name will be `tuf/exceptions.py`.
What do you think @jku?

tuf/api/exceptions.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class URLParsingError(Exception):
+    """If we are unable to parse a URL -- for example, if a hostname element
+    cannot be isoalted."""
+
+
+class RepositoryError(Exception):
+    """An error with a repository's state, such as a missing file."""
+
```
kairoaraujo: Should this`` RepositoryError`` be part of the below session ``### Repository errors ###``?

MVrachev:
I argument descriptions in the `DownloadLengthMismatchError` and `SlowRetrievalError` docstrings.

tuf/api/exceptions.py
```diff
@@ -0,0 +1,164 @@
...
+            + repr(self.expected_length)
+            + ")."
+        )
+
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + " : " + str(self)
+
+
+class SlowRetrievalError(DownloadError):
+    """ "Indicate that downloading a file took an unreasonably long time.
```
kairoaraujo:
```suggestion
    """Indicate that downloading a file took an unreasonably long time.
```

tuf/ngclient/fetcher.py
```diff
@@ -35,9 +35,9 @@ def fetch(self, url: str) -> Iterator[bytes]:
             url: A URL string that represents a file location.
 
         Raises:
-            tuf.exceptions.SlowRetrievalError: A timeout occurs while receiving
```
kairoaraujo, line 38: (this is more a generic comment)
I noticed that we refer to the full exception path; in other docstrings  (including tuf/api/metadata), we use only the exception name. Not sure which should be the standard.

jku:
> The only question that is left is do we think that tuf/api/exceptions.py is the right place for that file as it will be used from tuf/ngclient as well?
One possibility is renaming the old exceptions file to exceptions.old for example and moving the new exception file there so its name will be tuf/exceptions.py.

Let's not touch files that are part of the legacy code base.

I think the location of exceptions.py WRT ngclient is not an issue: ngclient already "depends" on tuf.api so exceptions being there is not a problem. There is an issue though (and you may also be thinking about this one): there are exceptions that are clearly client specific (at least FetcherHTTPError) and having those defined in non-client module is not great... but it's also not the worst thing in the world.

tuf/api/exceptions.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __init__(self, message: str) -> None:
+        super().__init__()
+        self.exception_message = message
+
+    def __str__(self) -> str:
+        return self.exception_message
+
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + " : " + str(self)
```
jku: All these functions seem completely unnecessary to me -- Exception should handle this just fine

tuf/api/exceptions.py
```diff
@@ -0,0 +1,164 @@
...
+        self.current_version = current_version
+
+    def __str__(self) -> str:
+        return (
+            "Downloaded "
+            + repr(self.metadata_role)
+            + " is older ("
+            + repr(self.downloaded_version)
+            + ") than the version currently "
+            "installed (" + repr(self.current_version) + ")."
```
jku: f-strings I suppose? this applies to all string formatting in the file (although I think this may be the only exception that needs any implementation so would also be the only one that would need strings fixed)

tuf/api/exceptions.py
```diff
@@ -0,0 +1,164 @@
...
+    Args:
+        expected_length: The expected length of the downloaded file
+        observed_length: The real length observed of the downloaded file
+    """
+
+    def __init__(self, expected_length: int, observed_length: int) -> None:
+        super().__init__()
+
+        self.expected_length = expected_length  # bytes
+        self.observed_length = observed_length  # bytes
```
jku: I'd like to argue the arguments are not useful -- First, observed_length is not a real length: it's just how far into the file we managed to read before noticing that _observed > expected_. Second, not even our tests use these values.

I think we could just remove all the functions in this class and just a pass a proper error message in fetcher.py?

tuf/api/exceptions.py
```diff
@@ -0,0 +1,164 @@
...
+class SlowRetrievalError(DownloadError):
+    """ "Indicate that downloading a file took an unreasonably long time.
+
+    Args:
+        average_download_speed: The average download speed in bytes per second
+    """
+
+    def __init__(self, average_download_speed: Optional[int] = None) -> None:
+        super().__init__()
+        self.__average_download_speed = average_download_speed  # bytes/second
```
jku: The argument is unused AFAICT (calling code never sets it): let's remove all the functions in this class and use a message instead

jku (commented):
I still want to have a quick look at all the errors (are they really something we want to keep) but I left some initial comments about removing a lot of the code

tuf/ngclient/fetcher.py
MVrachev: My opinion is that we shouldn't use the full exception path as it can change (as it happened now).
I think is best if we use the module name as I did in the mentioned example `exceptions.SlowRetrievalError`.
@jku what do you think?

MVrachev:
I addressed all comments by @jku and @kairoaraujo by:
- removing the `__str__` and `__repr__` functions for a couple of exceptions
- updating the `DownloadLengthMismatchError` exception message
- I couldn't update the `SlowRetrievalError` exception message with information about average download speed as we don't have that information

jku:
Some thinking on the exceptions we have / should have : 
* FetcherHTTPError is currently not derived from DownloadError: it should be as we let it propagate from Updater. I think we want to document  refresh, get_targetinfo and download_target as `raises DownloadError`
* Same for URLParsingError -- except no-one will ever want to handle URLParsingError I think: Maybe we replace it with a generic `DownloadError("Failed to parse URL")`
* UnsupportedAlgorithmError seems like it's not really a  TUF problem... but could still be useful info for repository maintainers -- maybe we could re-export securesystemslib UnsupportedAlgorithmError instead (#1646) and then let the error pass through?

tuf/ngclient/fetcher.py
jku: I personally think including the "module path" at all is not very useful in docstrings: it just makes things harder to read. But if we do include them, then they should be complete

MVrachev:
> Some thinking on the exceptions we have / should have :
> 
> * FetcherHTTPError is currently not derived from DownloadError: it should be as we let it propagate from Updater. I think we want to document  refresh, get_targetinfo and download_target as `raises DownloadError`
> * Same for URLParsingError -- except no-one will ever want to handle URLParsingError I think: Maybe we replace it with a generic `DownloadError("Failed to parse URL")`
> * UnsupportedAlgorithmError seems like it's not really a  TUF problem... but could still be useful info for repository maintainers -- maybe we could re-export securesystemslib UnsupportedAlgorithmError instead ([metadata API: avoid raising securesystemslib exceptions #1646](https://github.com/theupdateframework/python-tuf/issues/1646)) and then let the error pass through?

After a conversation with @jku I agreed on both three points mentioned here and a few additional others.
The two points mentioned in his comment are part of the last three commits I added. The third point and a couple
of other enhancements will be in my pr about issue #1646.

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -236,20 +235,17 @@ def update_timestamp(self, data: bytes) -> Metadata[Timestamp]:
...
-            ):
-                raise exceptions.ReplayedMetadataError(
-                    Snapshot.type,
-                    new_timestamp.signed.snapshot_meta.version,
-                    self.timestamp.signed.snapshot_meta.version,
+            snapshot_meta = self.timestamp.signed.snapshot_meta
+            new_snapshot_meta = new_timestamp.signed.snapshot_meta
+            if new_snapshot_meta.version < snapshot_meta.version:
+                raise exceptions.BadVersionNumberError(
+                    f"New snapshot version must be > {snapshot_meta.version}"
```
jku: message does not match the test

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -236,20 +235,17 @@ def update_timestamp(self, data: bytes) -> Metadata[Timestamp]:
         if self.timestamp is not None:
             # Prevent rolling back timestamp version
             if new_timestamp.signed.version < self.timestamp.signed.version:
-                raise exceptions.ReplayedMetadataError(
-                    Timestamp.type,
-                    new_timestamp.signed.version,
-                    self.timestamp.signed.version,
+                raise exceptions.BadVersionNumberError(
+                    f"New timestamp version {new_timestamp.signed.version} must"
+                    f" be above {self.timestamp.signed.version}"
```
jku: message does not match the test

jku (commented):
I believe these are good changes overall, however we should make sure we do all of the exceptions API changes at roughly same time so it's not a constant churn: I am far less interested in the location of the exceptions file than how the exceptions are used in metadata API and ngclient. The remaining specific issues I know of are 
* sslib exceptions in metadata api (#1734)
* verifying use and documentation of the download exceptions in ngclient/fetcher

It _looks_ like we can do those separately (but the focus on this PR should be to verify that any changes are in line with these near future plans).

tuf/ngclient/_internal/trusted_metadata_set.py
MVrachev: What do you mean?
Here we are replacing the `ReplayedMetadataError` which in this case will print
`Downloaded timestamp role is older (1) than the version currently installed (2)`
and the current message will be `New timestamp version 1 must be above 2`.
Should I mention that the new timestamp version is downloaded and that the local version is actually version 2?

tuf/ngclient/_internal/trusted_metadata_set.py
MVrachev: See my comment here https://github.com/theupdateframework/python-tuf/pull/1725/commits/ac1c3b65e1bc73bec75613b8c8c5a9e897530a14#r774067135

MVrachev:
In a conversation with @lukpueh he mentioned that the name `RepositoryError` sounds a little odd when we consider we are throwing those exceptions in the client code as well.
I experimented with the name `MetadataError` instead of `RepositoryError`.
In my head, there were a couple of other names like `MetadataValidationError` or `MetadataVerificationError` which will explain better that all of these errors are related to problems during the validation/verification of a metadata file, but of course, they are longer.
@jku, @lukpueh  what do you think?

jku:
> RepositoryError sounds a little odd when we consider we are throwing those exceptions in the client code as well.
> I experimented with the name MetadataError instead of RepositoryError.

Either one works for me: RepositoryError is not completely self explanatory but I don't see a specific problem with it: Updater uses the error to mean _"this error happens because of metadata we got from repository: The issue cannot be resolved without new metadata from repository"_: the point is that this is **not** a client bug, it's a problem on the repository end. I'd personally avoid an API change just to change the name but I can live with change as well.

The other suggestions are really long.

tuf/ngclient/_internal/trusted_metadata_set.py
jku: I mean the exception message is not accurate based on the if-clause in the previous line

tuf/ngclient/_internal/trusted_metadata_set.py
jku: no, I mean the exception message is not accurate based on the if-clause in the previous line

tuf/api/exceptions.py
```diff
@@ -0,0 +1,68 @@
...
+there is a good reason not to, and provide that reason in those cases.
+"""
+
+
+#### General errors ####
+
+# Reexport securesystemslib exceptions so that users of code that throws them
+# doesn't have to import securesystemslib to handle them.
+# pylint: disable=unused-import
+from securesystemslib.exceptions import UnsupportedAlgorithmError
```
jku: You know what? We can get rid of this without affecting anyones quality of life:

The only real use of UnsupportedAlgorithmError is in `Targetfile.from_data()` and in there it happens if an _argument_ contains a bad algorithm name. I think we can just raise a ValueError there instead without losing anything. Then we would have no more securesystemslib errors in the public Metadata API AFAICT.

tuf/ngclient/_internal/trusted_metadata_set.py
MVrachev: Yes, you are correct. It should be `New snapshot version must be >= {snapshot_meta.version}`

tuf/ngclient/_internal/trusted_metadata_set.py
MVrachev: The same mistake here. Instead of `above`, it should be `>=`.

MVrachev:
I rebased on top of develop and added https://github.com/theupdateframework/python-tuf/pull/1725/commits/a0171dc109cb47d34b56ede56958f453276157e7. Those are the only real changes here.
I think there is a sense in keeping both commits `Remove UnsupportedAlgorithmError and reexport it `  and `Replace UnsupportedAlgorithmError with ValueError` as it shows step by step the changes compared to `tuf/exceptions.py`.

MVrachev:
I followed @jku advice and replaced `UnsupportedAlgorithmError` with `ValueError`.
@jku and @lukpueh you can have a look.

MVrachev:
I added a couple of simplifications in my latest commits including making (De)Serialize errors a RepositoryErros.

On the question of the current exceptions:

**RepositoryErrors**

They allow us to test for those specific use cases without checking the exception messages.

UnsignedMetadataError is useful when we want to denote problems with the keys or signatures. 

It’s used on multiple occasions to communicate different signing problems to the users such as `Key failed to verify delegated role...`, `Delegated role was signed by x/y keys`, `No signature for key X found in the metadata` and others.

Also, I think exceptions related to signatures and keys are useful to have their own exception as they show problems that need a more specific solution - resign, replace a key, etc.

Then we have the TrustedMetadataSet errors. We have three of them BadVersionNumberError, ExpiredMetadataError, and LengthOrHashMismatchError.

They are useful in order to test specific repository errors at the same time the user can just catch 

BadVersionNumberError is needed to test when:

- new metadata version is below the expected one
- the new snapshot version is below the one requested in timestamp meta
- any target version is below the one described in snaphshot meta.

ExpiredMetadataError  is thrown when any of the metadata files has expired

LengthOrHashMismatchError is raised when:

- a metadata length doesn’t match the expected length
- metadata hashes don’t match the expected hashes

LengthOrHashMismatchError is caught twice in the ngclient:

- when a local cached target is found, but cannot be verified and a new version of this target is needed
- when a target is downloaded, but cannot be verified. In this case, it’s caught and reraised as `RepositoryError`, but I think there is a sense in keeping it as in the exception tree the user will be able to see that the problem is actually related to hashes and length verification.

On the question do we need additional exceptions?

Here are the possible new exceptions I see:

- when the metadata types are not the ones we expect - maybe here it makes sense to use a new one
- when information for a target is missing inside a new snapshot version compared to an older snapshot version- not sure we need a new one here
- when updating a delegated target, but snapshot doesn’t contain information for it - not sure we need a new one here

**DownloadErrors**

DownloadLengthMismatchError it’s used as expected and I think it makes sense to keep it for tests.

SlowRetrievalError: I don’t have a strong opinion on this one. On one hand, I see it serves the purpose to unify all timeout exceptions into one on another hand I am asking how does that help us? Do we want to do that?

FetcherHTTPError:

I see that FetcherHTTPError is used almost everywhere for a variety of reasons:

- hash mismatch with the downloaded target
- unknown root version
- unknown path

and others. Then, I see that DownloadError is not used anywhere.
Wondering is there a sense of keeping the general DownloadError and FetcherError at the same time? Can’t we replace the FectherError with DownloadError?

lukpueh:
Thanks for the research and for sharing your detailed insights, @MVrachev!

I mostly agree with you:

- Having two base errors `RepositoryError` for errors related to Metadata and target file verification, and `DownloadError` for communication errors, makes sense.

- The specific `RepositoryError`s make sense, although I don't like most of  their names. They are both long but also inaccurate. Not sure if it's worth renaming (I suggested renames in the past and then took them back) but here are some suggestions:
  - `UnsignedMetadataError` suggests that there is no Signature, which is not what it is exclusively used for. What about a simple `(Metadata)SignatureError`?
  - `BadVersionNumberError ` is probably okay, but I like `(Metadata)VersionError` better
  - `LengthOrHashMismatchError` seems both over-explicit, and also like it's withholding information, because the raiser does know if the length or the hash is the problem ... but maybe that detail is not relevant to the catcher? Would `(Metadata)ContentError` describe both case well? And/or should we split it into `(Metadata)LengthError` and `(Metadata)HashError`?

- Regarding the new `RepositoryError`s you suggest, how do we currently handle them? @ivanayov suggested that there were some tests where she regexed for error messages. That seems like a place to add new exception types. Regardless and without having looked at the code:
  - `(Metadata)TypeError` seems like a useful addition
  - I could see something like a `MissingMetadataError` for the other two scenarios you describe

- Having only one of `DownloadError` and `FetcherHTTPError` seems to make sense from a user perspective. Maybe we need the separation internally, e.g. for internal exception handling or for testing?

- `DownloadLengthMismatchError` again is a name I'm not particular fond of, but not sure if a rename is worth the trouble. Could it just be a `LengthError`?

- `SlowRetrievalError` suggests that an attack is happing, but it could also just be a bad wire. What about a simple `TimeoutError`?

MVrachev:
> * The specific `RepositoryError`s make sense, although I don't like most of  their names. They are both long but also inaccurate. Not sure if it's worth renaming (I suggested renames in the past and then took them back) but here are some suggestions:
>   
>   * `UnsignedMetadataError` suggests that there is no Signature, which is not what it is exclusively used for. What about a simple `(Metadata)SignatureError`?

I agree with you. `UnsignedMetadataError` has been used in a variety of situations. I think `SignatureError` is a good suggestion.

>   * `BadVersionNumberError ` is probably okay, but I like `(Metadata)VersionError` better

Agree with you that the prefix `bad` is unnecessary given that this is an error, so of course, it's bad :D. 
Again I am not sure that the `Metadata` prefix makes sense and `VersionError` sounds good.

>   * `LengthOrHashMismatchError` seems both over-explicit, and also like it's withholding information, because the raiser does know if the length or the hash is the problem ... but maybe that detail is not relevant to the catcher? Would `(Metadata)ContentError` describe both case well? And/or should we split it into `(Metadata)LengthError` and `(Metadata)HashError`?

I think splitting them is better. Those are two different use cases and having two different errors probably could shorten our messages (it needs to be tried).

> * Regarding the new `RepositoryError`s you suggest, how do we currently handle them? @ivanayov suggested that there were some tests where she regexed for error messages. That seems like a place to add new exception types. Regardless and without having looked at the code:

>   * `(Metadata)TypeError` seems like a useful addition

Agree.

>   * I could see something like a `MissingMetadataError` for the other two scenarios you describe

As the cases I described are more about missing information inside a metadata file it sounds to me it will be more appropriate to use `MissingInformationError` for example and in the message we can describe what information is exactly missing.

> * Having only one of `DownloadError` and `FetcherHTTPError` seems to make sense from a user perspective. Maybe we need the separation internally, e.g. for internal exception handling or for testing?

Can you emphasize more on that?

> * `DownloadLengthMismatchError` again is a name I'm not particular fond of, but not sure if a rename is worth the trouble. Could it just be a `LengthError`?

`DownloadLengthMismatchError` is used only once in the new code and we give good error messages as well:

image

so, I think it makes sense to rename it to `LengthError`.

> * `SlowRetrievalError` suggests that an attack is happing, but it could also just be a bad wire. What about a simple `TimeoutError`?

Good suggestion. All of the exceptions we are catching and rethrowing as `SlowRetrievalError` are `TimeoutError`s.

lukpueh:
Okay great, looks like we are pretty much on the same page. Let's have @jku, who's usually more cautious when it comes to renames, also weigh in. :) 


> > * Having only one of `DownloadError` and `FetcherHTTPError` seems to make sense from a user perspective. Maybe we need the separation internally, e.g. for internal exception handling or for testing?
> 
> Can you emphasize more on that?

Without looking at the client implementation, I'd say that `DownloadError` is broader than `FetcherHTTPError`, i.e. the former covers any error related to downloading, whereas the latter is something we raise (and maybe immediately handle) somewhere in the `Fetcher` class. A client user probably shouldn't care about fetcher errors, but our own code maybe needs it?  Does that make sense? I should probably just look at the code and see for myself if that distinction is needed. Or we ask @jku once more. :P

jku:
* I'm not keen on renaming exceptions unless they're really misleading but if you consider it important, go for it.

* Adding new, so far unused, exceptions: I'm not against but can we handle these in separate issues? I think most of those won't be necessary for 1.0 (and will be trivial to add without API break)

* FetcherHTTPError: We need the separate error internally so we can handle 403/404 when needed. Martin makes this error derive from DownloadError in this PR in order to make this easy for client to handle (since we do let other http errors propagate up to caller).

lukpueh:
> * I'm not keen on renaming exceptions unless they're really misleading but if you consider it important, go for it.

Let's defer the decision and take one final look at the issue shortly before the release, when we have resolved everything else (also see https://github.com/theupdateframework/python-tuf/issues/1745#issuecomment-1011192770) @MVrachev, can you ticketize?

> * Adding new, so far unused, exceptions: I'm not against but can we handle these in separate issues? I think most of those won't be necessary for 1.0 (and will be trivial to add without API break)

Agreed.

> * FetcherHTTPError: We need the separate error internally so we can handle 403/404 when needed. Martin makes this error derive from DownloadError in this PR in order to make this easy for client to handle (since we do let other http errors propagate up to caller).

Thanks for the insights.

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -132,9 +132,7 @@ def _get_session(self, url: str) -> requests.Session:
         parsed_url = parse.urlparse(url)
 
         if not parsed_url.scheme or not parsed_url.hostname:
-            raise exceptions.URLParsingError(
-                "Could not get scheme and hostname from URL: " + url
-            )
+            raise exceptions.DownloadError("Failed to parse URL")
```
lukpueh:
```suggestion
            raise exceptions.DownloadError("Failed to parse URL '{url}'")
```
?

lukpueh: Shouldn't this error be documented in the docstring of the calling `fetch` method?

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -175,10 +171,9 @@ def update_root(self, data: bytes) -> Metadata[Root]:
         self.root.verify_delegate(Root.type, new_root)
 
         if new_root.signed.version != self.root.signed.version + 1:
-            raise exceptions.ReplayedMetadataError(
-                Root.type,
-                new_root.signed.version,
-                self.root.signed.version,
+            raise exceptions.BadVersionNumberError(
```
lukpueh, line 174: General question for exceptions raised in methods in this module:
 If none of them are documented in the `Raises` section of the method docstrings (they only mention the broad `RepositoryError`), do we expect anyone to handle them? If we do, we should probably document them. E.g. the `BadVersionNumberError` here, etc.

lukpueh (approved):
I think this is a pretty good first iteration of cleaning up the exception taxonomy and suggest we merge as is. 

But I'd like to pick up the discussion from this PR once more before 1.0. In particular I'd like to review wether:
-  generic errors (`RepositoryError`, `DownloadError`) that we raise should be more specific,
- specific errors (`BadVersionNumberError, `LengthOrHashMismatchError`, etc.) really need to be that specific
-  built-in exceptions are used in an idiomatic fashion
- exception names aren't misleading
- all of above is consistent, i.e. there is no mix of generic and specific or custom and built-in exceptions, where there shouldn't be

tuf/ngclient/_internal/trusted_metadata_set.py
MVrachev: Point 2 of issue https://github.com/theupdateframework/python-tuf/issues/1784 is about that.

MVrachev:
Rebased on top of develop.

> I think this is a pretty good first iteration of cleaning up the exception taxonomy and suggest we merge as is.
> 
> But I'd like to pick up the discussion from this PR once more before 1.0. In particular I'd like to review wether:
> 
> * generic errors (`RepositoryError`, `DownloadError`) that we raise should be more specific,
> * specific errors (`BadVersionNumberError, `LengthOrHashMismatchError`, etc.) really need to be that specific

These two points can be discussed in #1784.

> * built-in exceptions are used in an idiomatic fashion
> * all of above is consistent, i.e. there is no mix of generic and specific or custom and built-in exceptions, where there shouldn't be

Created an issue about it: #1785.

> * exception names aren't misleading

Issue #1786 is about that.

PR #1726, opened 2021-12-14
Metadata API: remove dateutil requirement

### LINKED ISSUE #1722: Metadata API: Remove dependency on dateutil

MVrachev:
Fixes #1722

I added `dateutil` as a possible argument type for
`Signed.bump_expiration()` as we are already testing for this and
implying it should be supported.
The problem is that `dateutil` is not added as a nontest requirement
and after a discussion, we decided we don't want to add it as well.

That's why we decided to remove `dateutil` mentions from the code
and not confuse our users we support it.

We will create a separate issue discussing the validity of
`Metadata.bump_expiration()`.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -16,7 +16,6 @@ import unittest
 from datetime import datetime, timedelta
 from typing import ClassVar, Dict
 
-from dateutil.relativedelta import relativedelta
 from securesystemslib import hash as sslib_hash
 from securesystemslib.interface import (
     import_ed25519_privatekey_from_file,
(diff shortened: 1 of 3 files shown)


tests/test_api.py
```diff
@@ -333,15 +332,6 @@ def test_metadata_timestamp(self) -> None:
...
 
-        # Test whether dateutil.relativedelta works, this provides a much
-        # easier to use interface for callers
-        delta = relativedelta(days=1)
-        timestamp.signed.bump_expiration(delta)
-        self.assertEqual(timestamp.signed.expires, datetime(2031, 1, 3, 0, 0))
-        delta = relativedelta(years=5)
-        timestamp.signed.bump_expiration(delta)
-        self.assertEqual(timestamp.signed.expires, datetime(2036, 1, 3, 0, 0))
-
```
MVrachev, line 344: This test code is removed as if we don't make `timedelta` argument of `Signed` a `Union[timedelta, relativedelta]` then `mypy` complains.

jku (approved):
LGTM.

There are ways to test relativedelta usage without making it explicitly part of the API, and we probably want to do that... but since the expiry API has multiple related issues I think this is fine: let's remove relativedelta code completely for now to get a clean build and look for the correct solution to expiry issue in #1727.

PR #1728, opened 2021-12-14
ngtests: Test fetch target

### LINKED ISSUE #1642: ngtests: add tests for fetching targets
Add tests for updating delegated roles metadata following [5.7](https://theupdateframework.github.io/specification/v1.0.25/index.html#fetch-target) from the specification. A preliminary list of expected test cases:

Fetch target with consistent targets on/off	
    - local target file exists
    - target hashes/length match
Failure to fetch target	
    - Download errors
    - Hashes and length mismatch
Caching

The tests should:

  - Use RepositorySimulator for serving test files in-memory.
  - If possible avoid duplicating test cases when toggling consistent targets
  - "black-box" Updater, don't look at its internals, check if files are downloaded as expected (or not downloaded as expected).

sechkova:
Fixes #1642 

Adds tests for 'Fetch target' from 'Detailed client workflow' and target files storing/loading from cache.

requirements**:

### DIFF
--- a/tests/test_updater_fetch_target.py
+++ b/tests/test_updater_fetch_target.py
@@ -14,6 +14,7 @@ from typing import Optional, Tuple
 
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
+from tuf.exceptions import RepositoryError
 from tuf.ngclient import Updater
 
 
@@ -104,6 +105,63 @@ class TestFetchTarget(unittest.TestCase):
         self.assertEqual(path, updater.download_target(info, path))
         self.assertEqual(path, updater.find_cached_target(info))
         self.assertEqual(path, updater.find_cached_target(info, path))
+
+    def test_invalid_target_download(self) -> None:
+        targetpath = "targetpath"
+        # Add target to repository
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", b"content", targetpath)
+        self.sim.update_snapshot()
+
+        updater = self._init_updater()
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
+
+        # Corrupt the file content to not match the hash
+        self.sim.target_files[targetpath].data = b"conten@"
+        with self.assertRaises(RepositoryError):
+            updater.download_target(info)
+
+        # Corrupt the file content to not match the length
+        self.sim.target_files[targetpath].data = b"cont"
+        with self.assertRaises(RepositoryError):
+            updater.download_target(info)
+
+        # Verify the file is not persisted in cache
+        self.assertIsNone(updater.find_cached_target(info))
+
+    def test_invalid_target_cache(self) -> None:
+        targetpath = "targetpath"
+        content = b"content"
+        # Add target to repository
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", content, targetpath)
+        self.sim.update_snapshot()
+
+        # Download the target
+        updater = self._init_updater()
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
+        path = updater.download_target(info)
+        self.assertEqual(path, updater.find_cached_target(info))
+
+        # Add newer content to the same targetpath
+        content = b"contentv2"
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", content, targetpath)
+        self.sim.update_snapshot()
+
+        # Newer content is detected, old cached version is not used
+        updater = self._init_updater()
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
+        self.assertIsNone(updater.find_cached_target(info))
+
+        # Download target, assert it is in cache and content is the newer
+        path = updater.download_target(info)
+        self.assertEqual(path, updater.find_cached_target(info))
+        with open(path, "rb") as f:
+            self.assertEqual(f.read(), content)
 
 
 if __name__ == "__main__":
--- a/tests/test_updater_fetch_target.py
+++ b/tests/test_updater_fetch_target.py
@@ -45,8 +45,8 @@ class TestFetchTarget(unittest.TestCase):
     def tearDown(self) -> None:
         self.temp_dir.cleanup()
 
-    def _run_refresh(self) -> Updater:
-        """Creates a new updater and runs refresh."""
+    def _init_updater(self) -> Updater:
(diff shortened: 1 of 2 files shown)


tests/test_updater_fetch_target.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Add targets to repository
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", content, targetpath)
+        self.sim.update_snapshot()
+
+        updater = self._init_updater()
+        # target now exists, is not in cache yet
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
```
kairoaraujo, line 105:
```suggestion
        self.assertIsNotNone(info)
```

tests/test_updater_fetch_target.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_invalid_target_download(self) -> None:
+        targetpath = "targetpath"
+        # Add target to repository
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", b"content", targetpath)
+        self.sim.update_snapshot()
+
+        updater = self._init_updater()
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
```
kairoaraujo, line 134:
```suggestion
        self.assertIsNotNone(info)
```

tests/test_updater_fetch_target.py
```diff
@@ -0,0 +1,166 @@
...
+        targetpath = "targetpath"
+        # Add target to repository
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", b"content", targetpath)
+        self.sim.update_snapshot()
+
+        # Download the target
+        updater = self._init_updater()
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
```
kairoaraujo, line 160:
```suggestion
        self.assertIsNotNone(info)
```

tests/test_updater_fetch_target.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Add newer content to the same targetpath
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", b"contentv2", targetpath)
+        self.sim.update_snapshot()
+
+        # Newer content is detected, old cached version is not used
+        updater = self._init_updater()
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
```
kairoaraujo, line 173:
```suggestion
        self.assertIsNotNone(info)
```

kairoaraujo (commented):
I have just a small suggestion.

tests/test_updater_fetch_target.py
```diff
@@ -0,0 +1,166 @@
...
+            # create test specific dump directory
+            name = self.id().split(".")[-1]
+            self.sim.dump_dir = os.path.join(self.dump_dir, name)
+            os.mkdir(self.sim.dump_dir)
+
+    def tearDown(self) -> None:
+        self.temp_dir.cleanup()
+
+    def _init_updater(self) -> Updater:
+        """Creates a new updater and runs refresh."""
```
jku: does not run refresh

tests/test_updater_fetch_target.py
```diff
@@ -0,0 +1,166 @@
...
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            self.targets_dir,
+            "https://example.com/targets/",
+            self.sim,
+        )
+        return updater
+
+    targets: utils.DataSet = {
+        "standard case": ("targetpath", b"content", "targetpath"),
```
jku: I think it would be good to add at least a comment about the contents of this tuple (or make it into a namedtuple or dataclass). Understanding what each "targetpath" means currently requires reading code.

tests/test_updater_fetch_target.py
```diff
@@ -0,0 +1,166 @@
...
+        # Add newer content to the same targetpath
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", b"contentv2", targetpath)
+        self.sim.update_snapshot()
+
+        # Newer content is detected, old cached version is not used
+        updater = self._init_updater()
+        info = updater.get_targetinfo(targetpath)
+        assert info is not None
+        self.assertIsNone(updater.find_cached_target(info))
```
jku: please continue with download_target and verifying that
* find_cached_target() now returns the path
* cached files now has the new content

jku (changes requested):
Looks good. I think the cache-miss case should test that we successfully overwrite the "incorrect" file as well. Other comments are just nits.

It feels like three test cases is not a lot... but I can't think of what else to add so :shrug:

Would using delegated targets give us any useful test coverage? the actual delegation lookups are tested elsewhere so maybe not?

tests/test_updater_fetch_target.py
jku: I think these are for mypy: otherwise you'd need something like 

```
if info is None:
    self.fail("unexpected get_targetinfo() failure")
else:
    # test code using info
```

`assertIsNotNone` is not understood by mypy but `assert` is: then you can avoid the indentation for the rest of the test code

sechkova: What Jussi said :)

tests/test_updater_fetch_target.py
sechkova: I added a data class for target files in another PR before this one and I was planning to use it but it is not yet merged so in other words I got tangled in my own PRs ...
https://github.com/theupdateframework/python-tuf/blob/a0e74c1dbf2ca3b76f32516fda1b71ce01e89db6/tests/test_updater_delegation_graphs.py#L40

sechkova: I added a dataclass `TestTarget` in `test_updater_fetch_target.py` too which is slightly different than the one in `test_updater_delegation_graphs.py` . I realise the same name is not great but I was wondering should I rename one of the classes or we should try making the dataclasses reusable between test files. d1bc201

sechkova:
The cache miss test is improved and I agree defining a dataclass does improve readability I am not sure about the naming though, see my comment above.

jku (approved):
Looks good to me!

kairoaraujo:
LGTM

kairoaraujo (approved):

PR #1731, opened 2021-12-15
ngclient: Rename constructor arg

### LINKED ISSUE #1638: ngclient: review argument names
Now that API seems to stabilizing (and before we have a stable release) let's check that argument names ngclient uses are consistent.

Lukas in #1604:
> The pairs `repository_dir` - `metadata_base_url`, and `target_dir` - `target_base_url` both describe the local and remote location respectively for the given file type. Maybe we could use names that show this duality more directly? Here are some naming suggestions
> 
> 1. Rename `repository_dir` to `metadata_dir`. IMO `repository_dir` sounds more like something that contains metadata, targets and keys (at least that was the case in the old implementation), whereas here it only contains metadata.
> 2. Use remote and local prefixes, e.g. `local_metadata_dir` and `remote_metadata_base_url` or, to make the latter shorter, `remote_metadata_dir`?

At least metadata vs repository seems like an obvious flaw: no need to use two different terms. 

The second one I'm not so sure about: the names are long enough that I already avoid naming the args because of it, and for me "url" vs "dir" does give a hint that one is remote and one is not remote...

jku:
metadata_dir matches metadata_base_url better.

This is an API break for anyone using ngclient with named arguments.

Fixes #1638

---

I've considered the other ideas mentioned in the issue but I think this is the only one that seems clearly worth the change.

### DIFF
--- a/examples/client_example/client_example.py
+++ b/examples/client_example/client_example.py
@@ -52,7 +52,7 @@ def download(target: str) -> bool:
     """
     try:
         updater = Updater(
-            repository_dir=METADATA_DIR,
+            metadata_dir=METADATA_DIR,
             metadata_base_url=f"{BASE_URL}/metadata/",
             target_base_url=f"{BASE_URL}/targets/",
             target_dir=DOWNLOAD_DIR,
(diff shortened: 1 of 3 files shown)


lukpueh (approved):
LGTM! And I agree with you, all the other ideas we discussed in #1638 seem a lot less important.

PR #1734, opened 2021-12-17
Metadata API: Avoid raising securesystemslib exceptions

### LINKED ISSUE #1646: metadata API: avoid raising securesystemslib exceptions

MVrachev:
Fixes #1646 

In this pr, I aim to remove the need of our users to import securesystemslib
in order to catch the exceptions thrown by `tuf/metadata/api.py`. 
I either reexport securesystemslib exceptions or
reraise them by using tuf exceptions.

Reexporting securesystemslib exceptions is required so that users
of code that throws these exceptions doesn't have to import
securesystemslib to handle them.

A couple of additional fixes are added to this pr as well.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -13,6 +13,7 @@ import shutil
 import sys
 import tempfile
 import unittest
+from copy import copy
 from datetime import datetime, timedelta
 from typing import Any, ClassVar, Dict
 
@@ -126,6 +127,16 @@ class TestMetadata(unittest.TestCase):
 
         os.remove(bad_metadata_path)
 
+    def test_md_read_write_file_exceptions(self) -> None:
+        # Test writing to a file with bad filename
+        with self.assertRaises(exceptions.StorageError):
+            Metadata.from_file("bad-metadata.json")
+
+        # Test serializing to a file with bad filename
+        with self.assertRaises(exceptions.StorageError):
+            md = Metadata.from_file(f"{self.repo_dir}/metadata/root.json")
+            md.to_file("")
+
     def test_compact_json(self) -> None:
         path = os.path.join(self.repo_dir, "metadata", "targets.json")
         md_obj = Metadata.from_file(path)
@@ -211,6 +222,17 @@ class TestMetadata(unittest.TestCase):
         timestamp_key.verify_signature(md_obj)
         with self.assertRaises(exceptions.UnsignedMetadataError):
             targets_key.verify_signature(md_obj)
+
+    def test_sign_failures(self) -> None:
+        # Test throwing UnsignedMetadataError because of signing problems
+        # related to bad information in the signer.
+        md = Metadata.from_file(f"{self.repo_dir}/metadata/snapshot.json")
+        key_dict = copy(self.keystore[Snapshot.type])
+        key_dict["keytype"] = "rsa"
+        key_dict["scheme"] = "bad_scheme"
+        sslib_signer = SSlibSigner(key_dict)
+        with self.assertRaises(exceptions.UnsignedMetadataError):
+            md.sign(sslib_signer)
 
     def test_verify_failures(self) -> None:
         root_path = os.path.join(self.repo_dir, "metadata", "root.json")
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -14,7 +14,7 @@ import sys
 import tempfile
 import unittest
 from datetime import datetime, timedelta
-from typing import ClassVar, Dict
+from typing import Any, ClassVar, Dict
 
 from securesystemslib import hash as sslib_hash
 from securesystemslib.interface import (
@@ -51,7 +51,7 @@ class TestMetadata(unittest.TestCase):
     temporary_directory: ClassVar[str]
     repo_dir: ClassVar[str]
     keystore_dir: ClassVar[str]
-    keystore: ClassVar[Dict[str, str]]
+    keystore: ClassVar[Dict[str, Dict[str, Any]]]
 
     @classmethod
     def setUpClass(cls) -> None:
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -304,9 +304,7 @@ class Metadata(Generic[T]):
         Raises:
             tuf.api.serialization.SerializationError:
                 'signed' cannot be serialized.
-            securesystemslib.exceptions.CryptoError, \
-                    securesystemslib.exceptions.UnsupportedAlgorithmError:
-                Signing errors.
+            exceptions.UnsignedMetadataError: Signing errors.
 
         Returns:
             Securesystemslib Signature object that was added into signatures.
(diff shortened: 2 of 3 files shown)


tuf/api/metadata.py
```diff
@@ -1371,10 +1365,7 @@ def from_data(
...
                         data, algorithm
                     )
-            except (
-                sslib_exceptions.UnsupportedAlgorithmError,
-                sslib_exceptions.FormatError,
-            ) as e:
+            except (sslib_exceptions.FormatError,) as e:
                 raise exceptions.UnsupportedAlgorithmError(
                     f"Unsupported algorithm '{algorithm}'"
                 ) from e
```
jku: I mentioned this elsewhere already but after looking at this for a while I've convinced myself that we could just handle both of them and return something generic like UnsignedMetadataError like the verify functions do, no? then I think we we can remove UnsupportedAlgorithmError altogether.

jku: ah no, it is still used by TargetFile.from_data()

jku:
I think we can remove the last SSLIB error as well https://github.com/theupdateframework/python-tuf/pull/1725#discussion_r780236043

opinions?

joshuagl:
> I think we can remove the last SSLIB error as well [#1725 (comment)](https://github.com/theupdateframework/python-tuf/pull/1725#discussion_r780236043)
> 
> opinions?

:+1:
Removing all sslib exceptions from the python-tuf public API seems like a good goal.

MVrachev:
@jku and @lukpueh you can have a look into my list commit where I actually get rid of all securesystemslibs exceptions and there is no longer need for reexporting.

lukpueh:
> @jku and @lukpueh you can have a look into my list commit where I actually get rid of all securesystemslibs exceptions and there is no longer need for reexporting.

Is this ready for review? Or should I wait until we land #1725?

MVrachev:
You can review the last three commits.
Even though https://github.com/theupdateframework/python-tuf/pull/1734/commits/8403a891b8749fde4a7415b440ae7029f7dd6b3d and https://github.com/theupdateframework/python-tuf/pull/1734/commits/9090da2a797509624756e27b4ef6b5681297e3a1 should be squashed.

MVrachev:
@jku @lukpueh this pr is ready for review after the merge of #1725.

tests/test_api.py
```diff
@@ -212,6 +223,21 @@ def test_sign_verify(self) -> None:
         with self.assertRaises(exceptions.UnsignedMetadataError):
             targets_key.verify_signature(md_obj)
 
+    def test_sign_failures(self) -> None:
+        # Make sure there is a case where we throw UnsignedMetadataError.
+        # If one day securesystemslib change their implementation and throw
+        # something different than UnsignedMetadataError,
+        # then make sure the metadata API sign doc will change too.
```
jku: this comment is incorrect (sslib does not raise UnsignedMetadataError) and very long

jku:
I think there's a misunderstanding. My comments on securesystemslib.StorageError vs OSError were made in the context of the client: there a securesystemslib error just makes life harder for no good reason. In Metadata API there was a conscious decision to abstract storage (for the repository use case). Removing StorageError completely makes no sense here: we might want to expose StorageError in our exceptions though so that users don't need to import SSLIB explicitly.

I agree about the other SSLIB errors.

tests/test_api.py
MVrachev: Ohhh... such a bad comment. :D 
Yea, I am definitely updating it.

MVrachev:
@jku the pr is updated.
You are correct we do want to keep the `StorageError` exceptions as we do provide an option to use the `storage_backend` argument and we want to be able to throw exceptions a `StorageError` which generalizes exceptions from different storage interfaces.

jku (approved):
I agree with this. Would appreciate other maintainer opinions (@lukpueh ?) on my suggestion to re-export StorageError: I feel like that's a good strategy but maybe a little unusual?

tuf/api/metadata.py
```diff
@@ -305,9 +304,7 @@ def sign(
         Raises:
             tuf.api.serialization.SerializationError:
                 'signed' cannot be serialized.
```
lukpueh, line 306: This is not raised anymore. Can you move the serialize call out of the try block?

tuf/api/metadata.py
```diff
@@ -320,7 +317,15 @@ def sign(
 
             signed_serializer = CanonicalJSONSerializer()
 
-        signature = signer.sign(signed_serializer.serialize(self.signed))
+        try:
+            signature = signer.sign(signed_serializer.serialize(self.signed))
+        except (
+            sslib_exceptions.CryptoError,
+            sslib_exceptions.UnsupportedAlgorithmError,
```
lukpueh: This `except` block assumes that `signer` is an `SSlibSigner`, but the caller could pass any implementation of the `Signer`  interface. Ideally the `Signer` interface would define what exceptions to expect here, like storage or serialization interfaces do (maybe this is worth a ticket in securesystemslib?).

Should we just except all and re-raise as UnsignedMetadataError?

lukpueh (commented):
Regarding `StorageError`, I think re-exporting makes sense. See inline for more comments...

tuf/api/metadata.py
jku: good catch. For this PR (and 1.0) the last suggestion sounds reasonable to me.

tuf/api/metadata.py
lukpueh: My comment from above is nonsense. What I wanted to say is, if we except all below when calling `sign(serialize...)` and re-raise as `UnsignedMetadataError`, then `SerializationError` won't be raised anymore.

So either we take the call to serialize out of the try block below or we update the docstring

tuf/api/metadata.py
MVrachev: Really great catch.
Totatlly forgot that `signed_serializer` is an object implementing the general `securesystemslib.signer.Signer` interface.

tuf/api/metadata.py
MVrachev: I decided to move the serialization outside the try block so that the serialization error will propagate as I don't think it's good to hide it with `UnsignedMetadataError` exception.

jku: wow I completely missed this. Good solution I think.

tuf/api/metadata.py
MVrachev: > Ideally the Signer interface would define what exceptions to expect here, like storage or serialization interfaces do (maybe this is worth a ticket in securesystemslib?).


The `securesystemslib` issue about this: https://github.com/secure-systems-lab/securesystemslib/issues/386

lukpueh (approved):

PR #1735, opened 2021-12-19
test timestamp fast-forward attack recovery

### LINKED ISSUE #1713: ngclient: test fast-forward recovery
I think we do not currently test successful rollback (aka fast-forward recovery) for timestamp, snapshot and targets:
  * simulate a fast-forward attack: create repository with valid timestamp/snapshot/targets with version=MAX_VERSION (or something large anyway)
  * update client so it sees the large versions 
  * rotate the affected key (timestamp or snapshot) and set metadata version=1
  * update client, assert that new metadata is accepted by ngclient (even though the version number is lower than the local version)

kairoaraujo:
This test simulates the timestamp fast-forward attack recovery.
It simulates that the timestamp keys were compromised, the attacker
generated a new high version of the timestamp.

The repository generates a new key and rollbacks the timestamp
version to the initial version.

that is provided, the better.

Part of #1713 

### DIFF
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -339,6 +339,37 @@ class TestRefresh(unittest.TestCase):
 
         self._assert_files_exist([Root.type])
 
+    def test_new_timestamp_fast_foward_recovery(self) -> None:
+        """Test timestamp fast-forward recovery using key rotation.
+
+        The timestamp recovery is made by the following steps
+         - Remove the timestamp key
+         - Create and add a new key for timestamp
+         - Bump and publish root
+         - Rollback the timestamp version
+        """
+
+        # attacker updates to a higher version
+        self.sim.timestamp.version = 99999
+
+        # client refreshes the metadata and see the new timestamp version
+        self._run_refresh()
+        self._assert_version_equals(Timestamp.type, 99999)
+
+        # repo add new timestamp keys and recovers the timestamp version
+        self.sim.root.roles["timestamp"].keyids.clear()
+        self.sim.signers["timestamp"].clear()
+        key, signer = self.sim.create_key()
+        self.sim.root.add_key("timestamp", key)
+        self.sim.add_signer("timestamp", signer)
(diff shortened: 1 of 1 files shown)


tests/test_updater_top_level_update.py
```diff
@@ -339,6 +339,35 @@ def test_new_timestamp_expired(self) -> None:
 
         self._assert_files_exist([Root.type])
 
+    def test_new_timestamp_fast_foward_recovery(self) -> None:
+        self._init_updater()
+
+        # health initial version of the timestamp
+        self._run_refresh()
+        self._assert_version_equals(Timestamp.type, 1)
```
jku: this part looks unnecessary?

tests/test_updater_top_level_update.py
```diff
@@ -339,6 +339,35 @@ def test_new_timestamp_expired(self) -> None:
...
+
+        # attacker updates to a higher version and add new package
```
jku: I don't understand "add new package"?

jku (commented):
Looks good, thanks!  
* The test seems 100% correct but the first few lines seem unnecessary (for this specific test).
* Could mention the rollback requirement (timestamp key rotation) in a method docstring

Opinions?

tests/test_updater_top_level_update.py
kairoaraujo: I added all the scenarios/steps, such as adding packages as target files before/after. I simplified it, and  I left it by mistake.

jku (approved):
LGTM, the only nitpicks I have are
* the docstring doesn't specify that timestamp rollback requires _timestamp_ key rotation (this is relevant because the snapshot rollback requires rotating both online keys and not everyone knows these differences) ... but the code does make it clear 
* PR is marked as fixing the issue but it only fixes 1/3 (as mentioned snapshot and targets are slightly different cases)

Anyway, I think we can merge this as is, those are really minor. If you do tweak the docstring, feel free to squash to a single commit.

Issue #1727, opened 2021-12-14
Metadata API: rethink the expiry related API


MVrachev:
**Description of issue or feature request**:

In a recent discussion (see discussion in https://github.com/theupdateframework/python-tuf/issues/1722) we reevaluated the need for `Singed.bump_expiry()` as instead of using the API:

```
delta = relativedelta(years=1)
timestamp.signed.bump_expiration(delta)
```
one can just do:
```
timestamp.signed.expires += relativedelta(years=1)
```

As a side benefit this will allow the user to use `dateutil.relativedelta` which is not supported by `Singed.bump_expiry` as of #1726.

jku:
Some additional detail here 
* I think an actual repository would never bump expiry to `previous expiry plus delta`. They'd want to bump expiry to `today plus delta` (if you want reasons why, imagine an original expiry period of 365days: a responsible repository would of course resign much earlier than that, let's say at 121 day mark. Now the next expiry is set 609 days forward. In 121 more days the repository resigns again: now the expiry is 853 days into the future...).
* so a potential reason to include a method is to handle `today plus delta` and then make sure microseconds are not an issue 
* since relativedelta may only be needed for annotations, there is an option of "stringified" annotations (and then importing only if we're statically type checking) but let's only do this if it's really necessary

jku:
https://github.com/theupdateframework/python-tuf/pull/1712 is about handling microseconds with a `set_expiry()`: I think these two are related...

jku:
_(Edited to change example to a property instead of a method)_

One possibility:
* make expiry a property with a setter function that floors microseconds
* remove bump_expiry()
* document a reasonable way to bump expiry:
```python
  # set expiry to seven days from now
  signed.expiry = utcnow() + relativedelta(day=7)
```

I think I'm fine with this.

lukpueh:
> _(Edited to change example to a property instead of a method)_
> 
> One possibility:
> 
> * make expiry a property with a setter function that floors microseconds

Done as per #1712.

> * remove bump_expiry()

Agreed! 

> * document a reasonable way to bump expiry:
> 
> ```python
>   # set expiry to seven days from now
>   signed.expiry = utcnow() + relativedelta(day=7)
> ```

I think that's fair.

lukpueh:
I have a PR in the queue. Will publish tomorrow.

PR #1736, opened 2021-12-20
Metadata API: Remove 3 'update' methods + tests

### LINKED ISSUE #1627: Metadata API: Consider removing 'update()' methods
*[edited on 12/20/2021: Mention ".json" bias]*
*[edited on 11/10/2021: Point to #1230]*

The Metadata API classes for top-level roles Targets, Timestamp and Snapshot each have an `update` method. They have been criticised in several places (https://github.com/theupdateframework/python-tuf/pull/1193#discussion_r514572393 pp, https://github.com/theupdateframework/python-tuf/pull/1223#discussion_r532537746 pp, https://github.com/theupdateframework/python-tuf/pull/1620#issuecomment-947556006).

More specifically the following doubts have been expressed:

- `update` is an ambiguous name, e.g. does it stand for *"__update__ some attributes of object"*  or for *"update object in the course of a repository __update__"*? - https://github.com/theupdateframework/python-tuf/pull/1223#discussion_r532665409

- `update` is unspecific about which attributes are updated - #1230 

- Is `update` a repository action only and should thus be in a repository controller and not on the Metadata API, which is also used by the client? - https://github.com/theupdateframework/python-tuf/issues/1134

- [e.g. `Targets.update()`] is literally a one-liner that modifies a public dict attribute: we could document doing that instead and have a smaller API surface. - https://github.com/theupdateframework/python-tuf/pull/1620#issuecomment-947556006

-  [e.g. `Targets.update()`]   inserts or replaces the TargetFile... but a Targets handles both TargetFiles and delegations: what if I wanted to modify delegations instead? is it obvious that TargetFile needs a function but delegations do not? - https://github.com/theupdateframework/python-tuf/pull/1620#issuecomment-947556006

- [e.g. `Snapshot.update()`] is opinionated in regards to the file extension of the targets metadata to be updated with, i.e. it assumes that it is `.json`. The rest of the metadata API also often defaults to json, but allows a caller to override, which is not the case here.

**Current behavior**:

- purpose of method is unclear
- method name is ambiguous

**Expected behavior**:

- Make a better case for the purpose and names of methods and document it, or replace/remove.

lukpueh:
Fixes #1627

Remove ambiguous, unspecific, opinionated and trivial 'update' methods, which can be replaced by feasible one-liners that assign values directly to the object attribute to be *updated*. (see #1627 for details).

Reasons to have these methods would be increased usability in terms of
- reduced work
- immediate feedback on invalid assignments

However, given above described issues, the reasons against the methods as they are now seem to outweigh the reasons for them. Furthermore, it seems easier to re-add similar methods, which addressed these issues, after the upcoming 1.0.0 release than to remove or modify them.

This patch also removes the corresponding tests as they become irrelevant (there is no need to test object assignment).  In the case of the timestamp test, the removal also includes redundant test logic, which is already tested in `test_metadata_base`.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -32,7 +32,6 @@ from tuf.api.metadata import (
     Delegations,
     Key,
     Metadata,
-    MetaFile,
     Root,
     Snapshot,
     TargetFile,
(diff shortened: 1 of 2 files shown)


jku (approved):
Looks correct to me -- maybe we find out later that we do want some helper functions... but they will almost certainly look different so doing this ASAP seems like the right move. 

the bump_*() methods have a bit of the same issue (they are not easier to use than just accessing the data structure, and bump_expiry does not do what I think repositories would want to do) but this is a move in the right direction.

lukpueh:
> the bump_*() methods have a bit of the same issue (they are not easier to use than just accessing the data structure, and bump_expiry does not do what I think repositories would want to do) but this is a move in the right direction.

Agreed. See for instance the basic repo example. There, for version bump I do the addition assignment myself...

https://github.com/theupdateframework/python-tuf/blob/1f3654fb97ec919085bbe32e74169f9562486077/examples/repo_example/basic_repo.py#L346

...  and for expiration "bump"s, I use a custom oneliner helper, because I needed something similar for the `expires` constructor argument.

https://github.com/theupdateframework/python-tuf/blob/1f3654fb97ec919085bbe32e74169f9562486077/examples/repo_example/basic_repo.py#L52-L54

https://github.com/theupdateframework/python-tuf/blob/1f3654fb97ec919085bbe32e74169f9562486077/examples/repo_example/basic_repo.py#L288-L296

https://github.com/theupdateframework/python-tuf/blob/1f3654fb97ec919085bbe32e74169f9562486077/examples/repo_example/basic_repo.py#L336

lukpueh:
I'll create a ticket for those two bump methods...

lukpueh:
> I'll create a ticket for those two bump methods...

Just saw, there already is one for expiry #1727. And for bump_version I'll just drop a fix right away.

PR #1738, opened 2021-12-21
test snapshot fast-forward attack recovery

kairoaraujo:
This test simulates the snapshot fast-forward attack recovery.
It simulates that the snapshot keys were compromised, the attacker
generated a new high version of the snapshot.

The repository generates new keys for snapshot and timestamp and
rollbacks the snapshot version to the initial version.

Part of #1713 

### DIFF
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -424,6 +424,47 @@ class TestRefresh(unittest.TestCase):
 
         self._assert_version_equals(Snapshot.type, 2)
 
+    def test_new_snapshot_fast_foward_recovery(self) -> None:
+        """Test snapshot fast-forward recovery using key rotation.
+
+        The snapshot recovery requires the snapshot and timestamp key rotation.
+        It is made by the following steps:
+        - Remove the snapshot and timestamp keys
+        - Create and add a new key for snapshot and timestamp
+        - Rollback snapshot version
+        - Bump and publish root
+        - Bump the timestamp
+        """
+
+        # attacker updates to a higher version (bumping timestamp is required)
+        self.sim.snapshot.version = 99999
+        self.sim.update_timestamp()
+
+        # client refreshes the metadata and see the new snapshot version
+        self._run_refresh()
+        self._assert_version_equals(Snapshot.type, 99999)
+
+        # repo add new snapshot and timestamp keys and recovers snapshot version
+        self.sim.root.roles["snapshot"].keyids.clear()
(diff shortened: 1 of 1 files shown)


tests/test_updater_top_level_update.py
```diff
@@ -424,6 +424,47 @@ def test_new_snapshot_version_rollback(self) -> None:
...
+        # attacker updates to a higher version (bumping timestamp is required)
+        self.sim.snapshot.version = 99999
+        self.sim.update_timestamp()
+
+        # client refreshes the metadata and see the new snapshot version
+        self._run_refresh()
+        self._assert_version_equals(Snapshot.type, 99999)
+
+        # repo add new snapshot and timestamp keys and recovers snapshot version
+        self.sim.root.roles["snapshot"].keyids.clear()
```
jku: I didn't even notice this in the previous PR but it's a bit inconsistent that both `Snapshot.type` and `"snapshot"` are used... could just use Snapshot.type everywhere, or does it look worse like that?

tests/test_updater_top_level_update.py
```diff
@@ -424,6 +424,47 @@ def test_new_snapshot_version_rollback(self) -> None:
...
+        self.sim.root.add_key("snapshot", snapshot_key)
+        self.sim.add_signer("snapshot", snapshot_signer)
+        timestamp_key, timestamp_signer = self.sim.create_key()
+        self.sim.root.add_key("timestamp", timestamp_key)
+        self.sim.add_signer("timestamp", timestamp_signer)
+        self.sim.snapshot.version = 1
+        self.sim.root.version += 1
+        self.sim.publish_root()
+
+        self.sim.update_timestamp()
```
jku: This looks correct but for clarity following might be better (root update just contains the key changes: then we rollback snapshot version, signing with the new keys)

```suggestion
        self.sim.root.version += 1
        self.sim.publish_root()

        self.sim.snapshot.version = 1
        self.sim.update_timestamp()
```

jku (commented):
Yep, agreed on the implementation and listed requirements. Left some questions / comments

tests/test_updater_top_level_update.py
kairoaraujo: Any objection about fixing it for the previous PR on this PR?

jku: if it looks fine to you, go ahead

jku (approved):
Looks great, thanks.

PR #1741, opened 2021-12-21
Remove test_updater_with_simulator.py

sechkova:
~_WIP since it is built on top of two other PRs_~

` test_updater_with_simulator.py `was initially added to  to demonstrate the usage of `RepositorySimulator` when testing `Updater`.  Now Updater is tested extensively using simulator in several other test files.

This PR removes the duplicated tests and moves the useful ones to the relevant test file.

requirements**:


sechkova:
`test_not_loading_targets_twice` and `test_snapshot_rollback_with_local_snapshot_hash_mismatch` seemed fit for `test_updater_top_level_update.py` but since more tests related to cached metadata are on its way, I  started thinking if they deserve a separate test file or this one can be renamed to better match its content ...

tests/test_updater_top_level_update.py
```diff
@@ -499,6 +503,33 @@ def test_compute_metafile_hashes_length(self) -> None:
...
+
+        # First time looking for "somepath", only 'role1' must be loaded
+        updater.get_targetinfo("somepath")
+        wrapped_open.assert_called_once_with(
+            os.path.join(self.metadata_dir, "role1.json"), "rb"
+        )
+        wrapped_open.reset_mock()
+        # Second call to get_targetinfo, all metadata is already loaded
+        updater.get_targetinfo("somepath")
+        wrapped_open.assert_not_called()
```
MVrachev, line 531: I am wondering if this is the correct file for this test or it should be `test_updater_delegation_graphs`?
It seems we are actually using and testing delegation roles here.

MVrachev (approved):
I think it looks good with only one question.

tests/test_updater_with_simulator.py
```diff
@@ -1,217 +0,0 @@
...
-
-        updater = self._run_refresh()
-
-        # trigger updater to fetch the delegated metadata, check filenames
-        updater.get_targetinfo("anything")
-        local_metadata = os.listdir(self.metadata_dir)
-        for fname in roles_to_filenames.values():
-            self.assertTrue(fname in local_metadata)
-
-    def test_keys_and_signatures(self) -> None:
```
ivanayov, line 112: There are some tests covering keys and signatures (like `test_new_timestamp_fast_foward_recovery`, `test_root_rotation`, etc) , but I'm not sure they follow the exact specific case covered here. I'm fine with deleting it though, if others find it unnecessary.

jku: I think this is fine: this specific test was just me adding an example of how to modify sigs and keys when using the simulator. I think the simulator documentation and the better tests that now exist cover all that better by now.

tests/test_updater_top_level_update.py
jku: it does define a delegation but the point is just to test file opens (so not specifically delegations or top level updates)... If this is the only issue with the PR, I think I'll merge as is and we can improve later

jku (approved):
Looks good, thanks

PR #1742, opened 2021-12-22
test targets fast-forward attack recovery

kairoaraujo:
This test simulates the targets fast-forward attack recovery.
It simulates that the targets keys were compromised, the attacker
generated a new high version of the targets.

The repository generates new keys for targets and snapshot to
rollback the targets version to the initial version.

that is provided, the better.

Part of #1713 

### DIFF
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -499,6 +499,41 @@ class TestRefresh(unittest.TestCase):
         self._assert_version_equals("timestamp", 3)
         self._assert_version_equals("snapshot", 3)
 
+    def test_new_targets_fast_forward_recovery(self) -> None:
+        """Test targets fast-forward recovery using key rotation.
+
+        The targets recovery is made by issuing new Snapshot keys, by following
+        steps:
+            - Remove the snapshot key
+            - Create and add a new key for snapshot
+            - Bump and publish root
+            - Rollback the target version
+        """
+        # attacker updates to a higher version
+        self.sim.targets.version = 99999
+        self.sim.update_snapshot()
+
+        # client refreshes the metadata and see the new targets version
+        self._run_refresh()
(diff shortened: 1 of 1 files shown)


kairoaraujo:
Two comments here.
- Not 100% confident if my flow is correct.
- I added a target file to be more explicit, giving more context to the test, but it can be removed if irrelevant.

jku:
Test logic looks 100% correct to me. I would personally remove two things from the test to keep it as minimal as possible:
* as you mentioned the target file is not needed for this test
* the targets key rotation seems unnecessary as well: obviously if the targets key had been compromised in a production environment we'd want to rotate it but I think it's not necessary for this test and makes the test more complex

Basically I'm not opposed to adding tests with more steps and complications in them... but I would like the test suite to first contain the most condensed test cases, and only add more complicated ones if needed after that. Opinions on that?

kairoaraujo:
IMHO, we should keep the targets key rotation.
This rotation will also be part of the scenario if the targets key is compromised.

jku:
> This rotation will also be part of the scenario if the targets key is compromised.

Yes this is true ... however 
1. if we're strict about it this is a test for fast forward recovery **not** targets key compromise -- maybe there are other valid reasons to do this than key compromise (a software bug for example, one that bumps the version too many times, or some attack that allows attacker to trigger metadata generation but does not leak keys)
2. I wanted the tests to really show what is actually _required_ to do the version rewinds (this is something apparently no-one has ever written down before us): now reading the test it looks like targets key rotation is required to rewind the targets version, but that's not true

I can merge it like this but personally I'd prefer the smallest possible tests for these cases and then new tests (in other issues) for more complex scenarios if it looks like they are needed.

jku (approved):
Thank you. These are great tests that will make it much easier to write a spec issue about making the key rotation requirements clearer

PR #1749, opened 2021-12-23
Add TargetFile.custom() documentation when building readthedocs

### LINKED ISSUE #1602: metadata API: TargetFile.custom is not visible in documentation
The property is not documented and is not visible in RTD documentation: I think it would be useful to expose as the "supported" way for a client app to read custom metadata for a target

MVrachev:
Fixes #1602

Add the missing `Targetfile.custom()` property to the RTD documentation.
This property is not documented and is not visible in RTD documentation: 
I think it would be useful to expose as the "supported" way for a client app to read custom metadata for a target.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1287,7 +1287,7 @@ class TargetFile(BaseFile):
 
     @property
     def custom(self) -> Any:
-        return self.unrecognized_fields.get("custom", None)
+        return self.unrecognized_fields.get("custom")
 
     @classmethod
     def from_dict(cls, target_dict: Dict[str, Any], path: str) -> "TargetFile":
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1287,6 +1287,8 @@ class TargetFile(BaseFile):
 
     @property
     def custom(self) -> Any:
+        """Can be used to provide implementation specific data related to the
+        target. python-tuf does not use or validate this data."""
         return self.unrecognized_fields.get("custom")
 
     @classmethod


MVrachev:
I managed to add the `TargetFile.custom()` property in the RTD documentation.
Now I want to research how can I add information about the `custom` type: `Dict[str, Any]` by using the `sphinx` format.

MVrachev:
I realized I was mistaken about the `custom` property, it's both documented and type annotated.
The only question remaining is are we sure the `custom` type is `Any` or it's actually `Dict[str, Any]`?
@jku what do you think?

tuf/api/metadata.py
```diff
@@ -1287,7 +1287,8 @@ def __init__(
 
     @property
     def custom(self) -> Any:
-        return self.unrecognized_fields.get("custom", None)
+        """Contains additional information about the TargetFile instance"""
+        return self.unrecognized_fields.get("custom")
```
MVrachev: By adding a docstring to the property RTD noticed `custom` and documented it.

jku:
> The only question remaining is are we sure the `custom` type is `Any` or it's actually `Dict[str, Any]`?

I wouldn't object to either.

Specification (summarized by me):
> **CUSTOM**: An object.

Object is never defined but does seem to mean JSON object -- a dictionary with string keys -- so annotating it as such seems reasonable. There's probably no need for us to validate `custom` in deserialization though.

tuf/api/metadata.py
```diff
@@ -1287,7 +1287,8 @@ def __init__(
 
     @property
     def custom(self) -> Any:
-        return self.unrecognized_fields.get("custom", None)
+        """Contains additional information about the TargetFile instance"""
```
jku: Hmm, let's try to provide something more actionable here for someone just reading the API: now it sounds like python-tuf provides this additional information.

Maybe something like:
```
    """Can be used to provide implementation specific data related to the
    target. python-tuf does not use or validate this data.
    """
```

jku (changes requested):

ivanayov:
@jku A valid json in python works for boolean and number strings as well, so in that case `Dict` might not fit. On the other hand it would be a bit strange to attach such non-informative value to `TargetFile`, unless someone decides to use if for a  hacky implementation in a very specific corner-case. But should we still allow that or just consider it's useless and restrict to `Dict`? With the change it nicer to use, formally it's not `Any` anymore.

MVrachev:
> @jku A valid json in python works for boolean and number strings as well, so in that case Dict might not fit. On the other hand it would be a bit strange to attach such non-informative value to TargetFile, unless someone decides to use if for a hacky implementation in a very specific corner-case. But should we still allow that or just consider it's useless and restrict to Dict? With the change it nicer to use, formally it's not Any anymore.

You are correct that `custom` could be a `List` and `boolean`, `string` and an `int` as well.
We are not sure what it could be in this case even though `Dict` makes the most sense.
I removed the commit that changes the type annotation.

jku (approved):

PR #1750, opened 2022-01-03
docs: Make it clear what is hashed

### LINKED ISSUE #1737: docs: TargetFile should be explicit about what is hashed
looking at the docs: https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.metadata.supporting.html#tuf.api.metadata.TargetFile, it's not clear that `hashes` are hashes of the target file _content_. We should be explicit.

jku:
TUF uses file name hashes elsewhere so make it clear when hashes are for
file _contents_.

Fixes #1737.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -891,7 +891,8 @@ class MetaFile(BaseFile):
     Args:
         version: Version of the metadata file.
         length: Length of the metadata file.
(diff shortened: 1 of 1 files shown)


MVrachev (approved):
LGTM!

lukpueh (approved):

Issue #1753, opened 2022-01-05
tests: test_update_root_new_root_fail_threshold_verification is misleading


jku:
test_trusted_metadata_set.py: `test_update_root_new_root_fail_threshold_verification()`:
* test does not sign the metadata it modifies at all: loading metadata with completely wrong signatures of course fails as expected... but the test really implies this is related to the keyid change 
* it's not clear which threshold verification is being tested (whether the current roots threshold fails or the new roots threshold fails) -- a test for each might be useful

these tests might also be more useful if they increased the threshold by 1 (and re-signed), instead of clearing the keyids. This would allow one signature to be valid but the threshold to fail because there are not enough signatures.

PR #1754, opened 2022-01-05
Metadata API: Make Role.keyids ordered

### LINKED ISSUE #1752: role keyids still not ordered
We tried to fix issues with this in #1478 but apparently failed. Role keyids in data that we deserialize may be unsorted: We must preserve this order as it affects canonical json. Currently we put the keyids in a set which sorts them.

I suppose we have to make Role.keyids  a list again: The only "ordered set" in python seems to be a OrderedDict with keys only -- and that would look weird in an API.

The downside of making Role.keyids a list is that now duplicates could be added to Role.keyids . This should not be a big worry as there is proper API for it: `Root.add_key()`

jku:
keyids are ordered in the data we deserialize: Not preserving that order
breaks canonicalization. Set does not preserve order.

Change Role.keyids type from Set to List. This is strictly speaking
an API change but a minor one: keyids are supposed to be changed
via add_key()/remove_key().

Add tests for this for both Role and DelegatedRole. Shorten a related
exception message.

Fix #1752

---

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -676,15 +676,11 @@ class Role:
         threshold: int,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
-        keyids_set = set(keyids)
-        if len(keyids_set) != len(keyids):
-            raise ValueError(
-                f"keyids should be a list of unique strings,"
-                f" instead got {keyids}"
-            )
+        if len(set(keyids)) != len(keyids):
+            raise ValueError(f"Nonunique keyids: {keyids}")
         if threshold < 1:
             raise ValueError("threshold should be at least 1!")
-        self.keyids = keyids_set
+        self.keyids = keyids
         self.threshold = threshold
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
 
@@ -699,7 +695,7 @@ class Role:
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dictionary representation of self."""
         return {
-            "keyids": sorted(self.keyids),
+            "keyids": self.keyids,
             "threshold": self.threshold,
             **self.unrecognized_fields,
         }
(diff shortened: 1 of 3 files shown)


jku:
This modifies test_update_root_new_root_fail_threshold_verification() which seems broken to me: I left that for #1753

tuf/api/metadata.py
```diff
@@ -676,15 +676,11 @@ def __init__(
...
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
-        keyids_set = set(keyids)
-        if len(keyids_set) != len(keyids):
-            raise ValueError(
-                f"keyids should be a list of unique strings,"
-                f" instead got {keyids}"
-            )
+        if len(set(keyids)) != len(keyids):
+            raise ValueError("Nonunique keyids: {keyids}")
```
```suggestion
            raise ValueError(f"Nonunique keyids: {keyids}")
```

MVrachev (approved):
Looks good to me with only one minor detail that should be fixed.

tuf/api/metadata.py
jku: oof, thanks

jku:
focre-pushed: Only change is a string marked as f-string

tuf/api/metadata.py
```diff
@@ -676,15 +676,11 @@ def __init__(
...
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
-        keyids_set = set(keyids)
-        if len(keyids_set) != len(keyids):
-            raise ValueError(
-                f"keyids should be a list of unique strings,"
-                f" instead got {keyids}"
-            )
+        if len(set(keyids)) != len(keyids):
+            raise ValueError(f"Nonunique keyids: {keyids}")
```
joshuagl, line 680: Small nit, feel free to ignore
```suggestion
            raise ValueError(f"Non-unique keyids: {keyids}")
```

joshuagl (approved):
Brilliant, thanks!

ivanayov:
Looks good to me.

lukpueh (approved):
Thanks, LGTM! @jku, will you `s/nonunique/non-unique`? Let's merge otherwise!

jku:
I will live with nonunique since there's no other changes...

PR #1757, opened 2022-01-05
ngclient updater: clean temp file if write fails

### LINKED ISSUE #1575: ngclient: should clean temp file if write fails
This is assuming #1574 is merged: We should clean temp file in `Updater._persist_metadata()` manually if
* write fails but file was still created (disk could be full?)
* write succeeds but replace fails (can't imagine what could lead to this but probably something can...)

To re-iterate, we don't use Tempfiles automatic cleaning because in the normal case we want to move that file (and we can't both move it and let tempfile clean it).

I think something like this (completely untested code):
```python
try:
    # all current code  (tempfile creation, write, os.replace) here
except OSError as e:
    # delete tempfile if we managed to create one, then let the exception happen
    if tempfile:
        try:
            os.remove(tempfile.name)
        except FileNotFoundError:
            pass
    raise e
```
no idea how to easily test this code though

MVrachev:
Fixes #1575

When calling updater._persist_metadata() there is a possibility that
writing the temporary file to storage can succeed, but moving it with
`os.replace` could fail with OSError.
Make sure we are removing the newly created temporary file in that case.

I tried testing the first case described in the issue, but I had troubles
mocking `write` as it was not `os.write()`.

requirements**:

### DIFF
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -13,6 +13,7 @@ import sys
 import tempfile
 import unittest
 from typing import Callable, ClassVar, List
+from unittest.mock import MagicMock, patch
 
 from securesystemslib.interface import import_rsa_privatekey_from_file
 from securesystemslib.signer import SSlibSigner
@@ -315,6 +316,25 @@ class TestUpdater(unittest_toolbox.Modified_TestCase):
 
         # Get targetinfo for non-existing file
         self.assertIsNone(self.updater.get_targetinfo("file33.txt"))
+
+    @patch.object(os, "replace", wraps=os.replace)
+    @patch.object(os, "remove", wraps=os.remove)
+    def test_persist_metadata_fails(
+        self, wrapped_remove: MagicMock, wrapped_replace: MagicMock
+    ) -> None:
+        # Testing that when write succeeds (the file is created) and replace
+        # fails by throwing OSError, then the file will be deleted.
+        wrapped_replace.side_effect = OSError()
+        with self.assertRaises(OSError):
+            self.updater._persist_metadata("target", b"data")
+
+        wrapped_replace.assert_called_once()
+        wrapped_remove.assert_called_once()
+
+        # Assert that the created tempfile during writing is eventually deleted
+        # or in other words, there is no temporary file left in the folder.
+        for filename in os.listdir(self.updater._dir):
+            self.assertFalse(filename.startswith("tmp"))
 
 
 if __name__ == "__main__":
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -303,15 +303,24 @@ class Updater:
 
     def _persist_metadata(self, rolename: str, data: bytes) -> None:
         """Write metadata to disk atomically to avoid data loss."""
-
-        # encode the rolename to avoid issues with e.g. path separators
-        encoded_name = parse.quote(rolename, "")
-        filename = os.path.join(self._dir, f"{encoded_name}.json")
-        with tempfile.NamedTemporaryFile(
-            dir=self._dir, delete=False
-        ) as temp_file:
-            temp_file.write(data)
-        os.replace(temp_file.name, filename)
+        try:
+            # encode the rolename to avoid issues with e.g. path separators
+            encoded_name = parse.quote(rolename, "")
+            filename = os.path.join(self._dir, f"{encoded_name}.json")
+            with tempfile.NamedTemporaryFile(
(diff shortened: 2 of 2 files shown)


ivanayov:
Looks good to me. Probably adding one more assert that the file doesn't exist?

tests/test_updater_ng.py
```diff
@@ -316,6 +317,19 @@ def test_missing_targetinfo(self) -> None:
         # Get targetinfo for non-existing file
         self.assertIsNone(self.updater.get_targetinfo("file33.txt"))
 
+    @patch.object(os, "replace", wraps=os.replace)
+    @patch.object(os, "remove", wraps=os.remove)
+    def test_persist_metadata_fails(
+        self, wrapped_remove: MagicMock, wrapped_replace: MagicMock,
```
MVrachev: I noticed that `black` doesn't like that I put all of the arguments on one line here.
It requires that I should separate each of them on a new line even though this line is far less than 80 characters.
Honestly, I prefer if I leave it the way instead of unnecessarily using a new line for each argument.
It's possible that that's a new version of black requires that all function arguments that require above 80 characters are placed new line and not allow the above behavior anymore.
The black error is here: https://github.com/theupdateframework/python-tuf/runs/4717489146?check_suite_focus=true

@lukpueh and @jku I hope I can get your opinions on the matter.

MVrachev:
> Looks good to me. Probably adding one more assert that the file doesn't exist?

Yep, maybe you are correct that this way the test will be more complete.
I updated the pr adding a check for that.

tests/test_updater_ng.py
ivanayov: But if we keep it like that, linter will fail for every following PR and not catch errors from the next commands as well.

MVrachev: It's probable that there is a black configuration option for that.
First, I want to hear the other's opinions and then if they agree with me I will research.

jku: black is historically not a very configurable tool so I doubt this config exists -- it's supposed to make choosing a style easy by enforcing one ("There is Only One Style"). This argument splitting behaviour is part of the Black Style (and is documented in https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) and I don't think we should start working around it.

tests/test_updater_ng.py
```diff
@@ -316,6 +318,25 @@ def test_missing_targetinfo(self) -> None:
...
+        with self.assertRaises(OSError):
+            self.updater._persist_metadata("target", b"data")
+
+        wrapped_replace.assert_called_once()
+        wrapped_remove.assert_called_once()
+
+        # Assert that the created tempfile during writing is eventually deleted.
+        encoded_name = parse.quote("target", "")
+        filepath = os.path.join(self.updater._dir, f"{encoded_name}.json")
+        self.assertFalse(os.path.exists(filepath))
```
jku: are you sure this checks what the comment claims?

tests/test_updater_ng.py
MVrachev: I think I found why black was complaining - it was because I placed a comma after the last argument. 🤦 
When I removed it, it was happy.

tests/test_updater_ng.py
MVrachev: Yes, because I am constructing the filepath the same way it's done in `updater._persist_metadata`.

jku: `filepath` seems to be the final filename (the one that never gets created in this test), not the temporary filename. comment claims this checks that tempfile is deleted.

Why not list all the files in the directory and check that this matches your expectations?

MVrachev: I see what you mean. Right now my code checks if the filename is removed, but we want to check if the temporary name exists. The temporary name has the `tmp` prefix and random letters.
I will list the directory content and check for a file with a `tmp` prefix.

MVrachev:
Updated the pr with @jku suggestions.

MVrachev:
The CI failure is fixed in https://github.com/theupdateframework/python-tuf/pull/1762.
I will rebase after the merge of this pr and the failure should be fixed.

jku (approved):
Thank you, looks good. tempfile-not-found-during-exception case is untested but I'm fine with that.

MVrachev:
> Thank you, looks good. tempfile-not-found-during-exception case is untested but I'm fine with that.

I tried adding a new test patching `tempfile.NamedEntities().write()` but after I didn't succeed for a couple of hours I left it.
If in the future we needed I can invest more time in this because surely there should be a way to do it.

PR #1758, opened 2022-01-05
Add tests for Updater input validation

### LINKED ISSUE #1644: ngtests: add input validation tests for Updater API

Test all interesting variations of user input for Updater API. Potentially these tests may already be covered, it is still worth checking.
Table testing (see https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_metadata_serialization.py) seems like the way to go.

ivanayov:
This test covers `targetinfo`, `target_path`, `target_base_url`,
`metadata_dir` and `filepath` input validation of the `Updater`
methods

Fixes #1644

### DIFF
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -314,6 +314,18 @@ class TestUpdater(unittest.TestCase):
         for filename in os.listdir(self.updater._dir):
             self.assertFalse(filename.startswith("tmp"))
 
+    def test_invalid_target_base_url(self) -> None:
+        info = TargetFile(1, {"sha256": ""}, "targetpath")
+        with self.assertRaises(exceptions.DownloadError):
+            self.updater.download_target(info, target_base_url="invalid_url")
+
+    def test_non_existing_target_file(self) -> None:
+        info = TargetFile(1, {"sha256": ""}, "/non_existing_file.txt")
+        # When non-existing target file is given, download fails with
+        # "404 Client Error: File not found for url"
+        with self.assertRaises(exceptions.DownloadHTTPError):
+            self.updater.download_target(info)
+
 
 if __name__ == "__main__":
     utils.configure_test_logging(sys.argv)
--- a/tests/test_updater_validation.py
+++ b/tests/test_updater_validation.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+
+# Copyright 2022, New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Test ngclient Updater validations.
+"""
+
+import os
+import sys
+import tempfile
+import unittest
+
+from tests import utils
+from tests.repository_simulator import RepositorySimulator
+from tuf.ngclient import Updater
+
+
+class TestUpdater(unittest.TestCase):
+    """Test ngclient Updater input validation."""
+
+    def setUp(self) -> None:
+        # pylint: disable-next=consider-using-with
+        self.temp_dir = tempfile.TemporaryDirectory()
+        self.metadata_dir = os.path.join(self.temp_dir.name, "metadata")
+        self.targets_dir = os.path.join(self.temp_dir.name, "targets")
+        os.mkdir(self.metadata_dir)
+        os.mkdir(self.targets_dir)
+
+        # Setup the repository, bootstrap client root.json
+        self.sim = RepositorySimulator()
+        with open(os.path.join(self.metadata_dir, "root.json"), "bw") as f:
+            f.write(self.sim.signed_roots[0])
+
+    def tearDown(self) -> None:
+        self.temp_dir.cleanup()
+
+    def _new_updater(self) -> Updater:
+        return Updater(
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            self.targets_dir,
+            "https://example.com/targets/",
+            fetcher=self.sim,
+        )
+
+    def test_local_target_storage_fail(self) -> None:
+        self.sim.add_target("targets", b"content", "targetpath")
+        self.sim.targets.version += 1
+        self.sim.update_snapshot()
+
+        updater = self._new_updater()
+        target_info = updater.get_targetinfo("targetpath")
+        assert target_info is not None
+        with self.assertRaises(FileNotFoundError):
+            updater.download_target(target_info, filepath="")
+
+    def test_non_existing_metadata_dir(self) -> None:
+        with self.assertRaises(FileNotFoundError):
+            # Initialize Updater with non-existing metadata_dir
+            Updater(
+                "non_existing_metadata_dir",
+                "https://example.com/metadata/",
+                fetcher=self.sim,
+            )
+
+
+if __name__ == "__main__":
+    utils.configure_test_logging(sys.argv)
+    unittest.main()


ivanayov:
IMO linter needs correction for `Optional` types.
It fails with
```
tests/test_updater_validation.py:91: error: Argument 1 to "download_target" of "Updater" has incompatible type "Optional[TargetFile]"; expected "TargetFile"  [arg-type]
```
while the first argument is actually of type `TargetFile`. 

I think it's better to fix the linter, rather than do unnecessary casts.

tests/test_updater_validation.py
```diff
@@ -0,0 +1,107 @@
...
+            "https://example.com/targets/",
+            fetcher=self.sim,
+        )
+
+    def _new_updater_without_targets_setup(self) -> Updater:
+        return Updater(
+            self.metadata_dir,
+            "https://example.com/metadata/",
+            fetcher=self.sim,
+        )
```
MVrachev: I am not convinced there is a need for this helper function.
You use it only once in one test. This code could be in the test itself.

tests/test_updater_validation.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def _new_updater_with_non_existing_metadata_dir(self) -> Updater:
+        return Updater(
+            "non_existing_metadata_dir",
+            "https://example.com/metadata/",
+            fetcher=self.sim,
+        )
+
```
MVrachev: I am not convinced there is a need for this helper function.
You use it only once in one test. This code could be in the test itself.

tests/test_updater_validation.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_empty_target_path_validation(self) -> None:
+        updater = self._new_updater_with_targets_setup()
+        target_file = updater.get_targetinfo("")
+        self.assertIsNone(target_file)
+
+    def test_empty_targetinfo_validation(self) -> None:
+        updater = self._new_updater_without_targets_setup()
+        info = TargetFile(1, {"sha256": ""}, "")
+        with self.assertRaisesRegex(
+            ValueError, "target_dir must be set if filepath is not given"
```
MVrachev: We should discuss if we want error messages when using `assertRaises`.
We haven't used them until now. I realize why you have used it - `ValueError` can be thrown for many reasons and you want to verify that it's the one you expect.
What do you think @jku?

tests/test_updater_validation.py
```diff
@@ -0,0 +1,107 @@
...
+        ):
+            updater.download_target(info)
+
+        updater = self._new_updater_with_targets_setup()
+        with self.assertRaisesRegex(FetcherHTTPError, "No target"):
```
MVrachev: Even if we decide to use `assertRaisesRegex` in order to additionally make sure the exception is the right one I am not sure we need it now.
`FetcherHTTPError` is our custom exception and we don't expect it to come from elsewhere.

tests/test_updater_validation.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def test_empty_filepath_validation(self) -> None:
+        self.sim.add_target("targets", b"content", "targetpath")
+        self.sim.targets.version += 1
+        self.sim.update_snapshot()
+
+        updater = self._new_updater_with_targets_setup()
+        target_info = updater.get_targetinfo("targetpath")
+        with self.assertRaises(securesystemslib.exceptions.StorageError):
+            if target_info is not None:
```
MVrachev: Why do you need to make that check?
When `target_info` is `None` considering you are adding it at the begging of the test?

MVrachev (changes requested):
What I like as a result of this pr is that we can see what exceptions are thrown by different Updater public API functions.
I am planning on fixing the `TODO` items related to `Raises` documentation in the public function docstrings from `tuf/ngclient/updater.py`.

tests/test_updater_validation.py
ivanayov: @MVrachev this one is just for the linter 😄 It fails for non-matching object types - `Optional` vs `TargetFile`. On the other side, I found it as a good correction, as `download_target` doesn't expect None values.

tests/test_updater_validation.py
ivanayov: In this case specifically, I think that without the regex, we don't guarantee that any verification happens on the desired level. If someone changes the message in future, it's not a hard effort to update related tests, considering `assertRaisesRegex` is rarely used.

tests/test_updater_validation.py
ivanayov: I can agree here if most arguments are against using regex. On the other hand `FetcherHTTPError` can be thrown for a variety of reasons, but I'm fine checking the exception only. Let's see what others say.

tests/test_updater_validation.py
ivanayov: Those are just for better readability, because I reuse `_new_updater_with_targets_setup` multiple times and if it's just `_new_updater` it wouldn't be clear why it's not used everywhere, while now the naming is informative and the same approach is used everywhere.

MVrachev: I think because they are functions it adds overhead.
Maybe add a comment instead.

tests/test_updater_validation.py
kairoaraujo: nit: An alternative that I saw in another test is

```python
        updater = self._new_updater_with_targets_setup()
        target_info = updater.get_targetinfo("targetpath")
        assert target_info is not None
        with self.assertRaises(securesystemslib.exceptions.StorageError):
             updater.download_target(target_info, filepath="")
```

tests/test_updater_validation.py
```diff
@@ -0,0 +1,107 @@
...
+            updater.download_target(info)
+
+    def test_empty_filepath_validation(self) -> None:
+        self.sim.add_target("targets", b"content", "targetpath")
+        self.sim.targets.version += 1
+        self.sim.update_snapshot()
+
+        updater = self._new_updater_with_targets_setup()
+        target_info = updater.get_targetinfo("targetpath")
+        with self.assertRaises(securesystemslib.exceptions.StorageError):
```
jku: good find, another non-documented Securesystemslib error.. we can handle this in a separate issue though

tests/test_updater_validation.py
MVrachev: I was going to link to it.
I think it looks better as we already use asserts in our tests.

tests/test_updater_validation.py
```diff
@@ -0,0 +1,103 @@
...
+            "https://example.com/targets/",
+            fetcher=self.sim,
+        )
+
+    def test_empty_target_path_validation(self) -> None:
+        updater = self._new_updater()
+        target_file = updater.get_targetinfo("")
+        self.assertIsNone(target_file)
+
+    def test_empty_targetinfo_validation(self) -> None:
```
jku: the first half of this test (target_dir or filepath must be set) seems the same as `test_updater_ng.py:test_no_target_dir_no_filepath()`

The second half of the test seems to test something completely different: A target file not found on server... this is a good separate test I think but A) should have a better name B) should be done against a real webserver, not simulator (so probably also in test_updater_ng.py)

tests/test_updater_validation.py
```diff
@@ -0,0 +1,103 @@
...
+        with self.assertRaisesRegex(
+            ValueError, "target_dir must be set if filepath is not given"
+        ):
+            updater.download_target(info)
+
+        updater = self._new_updater()
+        with self.assertRaisesRegex(FetcherHTTPError, "No target"):
+            updater.download_target(info)
+
+    def test_empty_filepath_validation(self) -> None:
```
jku: I think the test is useful but maybe the name should be `test_local_target_storage_fail()` or something: we don't validate empty filepath as the test name suggests but testing what happens when local filesystem fails to store the file (which is what the test actually does) is good.

tests/test_updater_validation.py
```diff
@@ -0,0 +1,103 @@
...
+        self.sim.targets.version += 1
+        self.sim.update_snapshot()
+
+        updater = self._new_updater()
+        target_info = updater.get_targetinfo("targetpath")
+        assert target_info is not None
+        with self.assertRaises(securesystemslib.exceptions.StorageError):
+            updater.download_target(target_info, filepath="")
+
+    def test_non_existing_metadata_dir_validation(self) -> None:
```
jku: nitpick: I'd just call it `test_non_existing_metadata_dir` (because we don't validate anything: we just let it fail)

jku (changes requested):
I've left some specific comments in source.

A general comment on testing any download failures or Fetcher errors: They seem like good new ideas but it might be more useful to add those into `test_updater_ng.py` as testing this aspect with Repositorysimulator is going to be quite artificial: there's not even a webserver involved... 

I think that applies to second half of `test_empty_targetinfo_validation` and  `test_invalid_target_base_url`.

tests/test_updater_validation.py
jku: #1761

tests/test_updater_ng.py
```diff
@@ -336,6 +336,16 @@ def test_persist_metadata_fails(
...
 
+    def test_invalid_target_base_url(self) -> None:
+        info = TargetFile(1, {"sha256": ""}, "targetpath")
+        with self.assertRaises(exceptions.URLParsingError):
+            self.updater.download_target(info, target_base_url="invalid_url")
+
+    def test_non_existing_target_file(self) -> None:
+        info = TargetFile(1, {"sha256": ""}, "")
+        with self.assertRaises(exceptions.DownloadLengthMismatchError):
+            self.updater.download_target(info)
```
jku: the error changed here without explanation...

ivanayov: @jku I separated to two different tests to cover both cases with a description of the `DownloadLengthMismatchError`.  The difference comes after I moved to `test_updater_ng` with using a webserver rather than a simulator. In this case dir listing is returned for empty path, so it results to the mismatch error, while with the simulator such listing cannot be returned, as a real fetch doesn't happen actually, so it was failing with `no target` exception. I added one more test for non-existing target file to cover the previous use-case.

ivanayov:
Linter is failing for `examples/repo_example/hashed_bin_delegation.py` which is not related to the current changes. I addressed the requirements to verify locally that `tox` will pass, but didn't commit the change, as it's not related to the PR. This probably came with some linter update, as the file was not changed recently. Can address this in a separate PR.

MVrachev:
@ivanayov you can rebase your pr on develop.
The pr #1814 which fixes the linting problem was merged.

ivanayov:
@MVrachev thank you for addressing the issue! I just pushed an update.

tests/test_updater_ng.py
jku: This is now testing a peculiarity of the webserver we use: it returns a directory listing if you ask for a directory. So the reason for the error is not really the empty filepath but how the webserver reacts in this specific case... I think this test is just reproducing test_length_hash_mismatch() so is not useful.

tests/test_updater_validation.py
```diff
@@ -0,0 +1,76 @@
...
+            "https://example.com/metadata/",
+            self.targets_dir,
+            "https://example.com/targets/",
+            fetcher=self.sim,
+        )
+
+    def test_empty_target_path_validation(self) -> None:
+        updater = self._new_updater()
+        target_file = updater.get_targetinfo("")
+        self.assertIsNone(target_file)
```
jku: empty string is not a special case anywhere so this is the same test as `test_missing_targetinfo()`?

jku (commented):
I think two of the tests do not give us much value and could be removed. 

Other tests look good... Although we're left with just two short local storage tests in the new file: maybe it's not worth it to add 50 lines of setup code for two tests? We could just move those two to `test_updater_ng.py`?

jku (approved):
Approving, these look good.

I left this question that wasn't responded to though:
> ...we're left with just two short local storage tests in the new file: maybe it's not worth it to add 50 lines of setup code for two tests? We could just move those two to test_updater_ng.py?

no need to do it now, just wondering about that

Issue #1634, opened 2021-10-25
ngclient: consistently encode URLs


jku:
Currently we do not explicitly URL-encode rolenames (or target paths) when we use them in URLs. Let's review this code and make a conscious decisions about this.

My original thinking on this was that it's a specification issue: we can't just start encoding e.g. rolenames if the spec does not say we should... but thinking on it further I've started to believe that
* spec should specify how URLS are formed and what must be URL encoded and when
* but we can be more explicit in ngclient before that: it should not break anything that currently works

The latter point is true because requests (the http library) already encodes everything that would be illegal in a URL behind our backs: so emojis, non-ascii characters, most symbols _are already being encoded_, just not explicitly by us. Only some specific symbols that may be valid in URLs are maybe not encoded (examples: /, #, ?, &).

My current thinking is:
1. we should not require rolenames to be encoded: they are just strings
1. we should encode rolenames before using them in URLs. This should not break any reasonable rolename and is going to be safer

I am not as sure about target paths: they are URL-paths and we could just require them to be URL-encoded already (and maybe even validate that they are encoded?). At the very least we can't encode "/" in target paths because it's valid in a path: this leads me to think we should just require the target path to be correctly encoded already? An additional wrinkle for target paths is that the spec requires us to actually modify the path in quite tricky ways for _consistent targets_...

kairoaraujo:
I want to work on this issue.

kairoaraujo:
Should we close this one? 🤔 or at least remove the ``1.0.0 blocker`` tag?

jku:
I think it's done:
* rolenames in URLs are now handled explicitly (practical results are 99.9% same as before but now the encoding is visible and testable)
* target urls do use hashes and targetpath (that both come from remote) but there isn't really anything we can do -- I suppose we could test that they are valid url parts or look for constructs like "../" in them... but I don't see major benefits from doing so.

Closing.

PR #1759, opened 2022-01-06
Explicit encode role names

kairoaraujo:
This commit explicitly encodes role names.
Also, a slight change in the RepositorySimulator will align
with the tests.

This commit partially covers issue #1634

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -259,6 +259,8 @@ class RepositorySimulator(FetcherInterface):
         If version is None, non-versioned metadata is being requested.
         """
         self.fetch_tracker.metadata.append((role, version))
+        # decode role for the metadata
+        role = parse.unquote(role, encoding="utf-8")
 
         if role == Root.type:
             # return a version previously serialized in publish_root()
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -328,9 +328,10 @@ class TestDelegationsGraphs(TestDelegations):
         finally:
             self.teardown_subtest()
 
-    def test_fishy_rolenames(self) -> None:
-        # Test that delegated roles filenames are safely encoded
-        # when persisted to the file system.
+    def test_safely_encoded_rolenames(self) -> None:
+        """Test that delegated roles names are safely encoded in the filenames
+        and URLs.
+        """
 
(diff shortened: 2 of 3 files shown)


tests/test_updater_fetch_target.py
```diff
@@ -84,12 +87,39 @@ def _init_updater(self) -> Updater:
             content=b"dir target content",
             encoded_path="a%2Fb%2Fc%2Ftargetpath",
         ),
+        "delegated role standard case": TestTarget(
+            path="projectA/file_a.txt",
+            content=b"file_a content",
+            encoded_path="projectA%2Ffile_a.txt",
+            rolename="A",
+            delegated_role=DelegatedRole(
+                "A", [], 1, False, ["projectA/**"], None
```
jku: `"projectA/**"`: the double asterisk format (as it's used in many places) is not supported in python-tuf so this looks misleading.

jku (commented):
The actual change is correct IMO, but I have some questions on the tests...

We might need some target downloads tests using delegations but your case does not seem to need target downloads: maybe just add the tests into test_updater_delegation_graphs.py -- I've not  confirmed but I would expect you can just add new test cases without having to modify the test code in any way?

Also could maybe add some explanation to the commit message: that 99% of this is already happening in the real use case as requests silently encodes everything that is not valid in a URL: we're just making this explicit. The only "breaking" change here for Requests-using-code is that "/" in a rolename will now be encoded

Speaking of "/" or "\\" in a rolename: this is a case I'd like to see a test  for... which leads to: we should have some tests for when we really use Requests (and not only simulator). These tests should maybe be in `test_updater_ng.py` as that already runs a webserver but I don't think they absolutely need to be added in this PR.

kairoaraujo:
@jku, I'm still not happy with tests for it. 😔
But I want to move forward with this issue.

jku (approved):
Thanks looks good. I'm approving this but would not mind a better docstring and/or name for the test if you can improve them: it now tests that rolenames are correctly encoded in both filenames and URLs.

kairoaraujo:
> Thanks looks good. I'm approving this but would not mind a better docstring and/or name for the test if you can improve them: it now tests that rolenames are correctly encoded in both filenames and URLs.

I will improve it. Please wait before merging it.

PR #1760, opened 2022-01-07
Add to CI check for specification version.

### LINKED ISSUE #1598: Develop a process for keeping `metadata.SPECIFICATION_VERSION` up-to-date

`tuf.api.metadata.SPECIFICATION_VERSION` is currently 1.0.19, latest TUF spec is 1.0.25. The larger these differences become, the more of a chore it is to ensure we are fully compliant with new specification versions.

Let's develop a process for ensuring we frequently assess new spec versions, make changes to python-tuf where appropriate, and keep `tuf.api.metadata.SPECIFICATION_VERSION` up-to-date.

I think the ideal would be some GitHub Actions automation that monitors the specification releases and files issues to trigger a manual comparison, but in the short-term we could simply add a step to the release checklist?

**Current behavior**:

* Updating `tuf.api.metadata.SPECIFICATION_VERSION` requires remembering to check spec releases and reviewing all changes in bulk
* `tuf.api.metadata.SPECIFICATION_VERSION` is several releases behind the specification

**Expected behavior**:

* Monitoring the spec and making appropriate changes to python-tuf is a standard part of the development process
* `tuf.api.metadata.SPECIFICATION_VERSION` is up-to-date with the specification

kairoaraujo:
This commit adds to the CI an automatic check for the TUF
specification version and compares it with the python-tuf metadata
API version.

If the version does not match and there is not a issue already open,
a new issue is opened.

Closes #1598

### DIFF
--- a/.github/workflows/specification-version.yml
+++ b/.github/workflows/specification-version.yml
@@ -1,0 +1,55 @@
+name: Specification version check
+
+on:
+  schedule:
+    - cron: '0 13 * * *'
+  workflow_dispatch:
+
+jobs:
+  check-spec-version:
+    name: Open an issue when spec version bumps
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - name: Get supported version
+      id: get-version
+      run: |
+        python3 -m pip install -e .
+        script="from tuf.api.metadata import SPECIFICATION_VERSION; \
+                print(f\"v{'.'.join(SPECIFICATION_VERSION)}\")"
+        ver=$(python3 -c "$script")
+        echo "::set-output name=version::$ver"
+    - name: Open issue (if needed)
+      uses: actions/github-script@v5
+      with:
+        script: |
+          const release = await github.rest.repos.getLatestRelease({
+            owner: "theupdateframework",
+            repo: "specification"
+          });
+          const spec_version = release.data.tag_name
+          const supported_version = "${{ steps.get-version.outputs.version }}"
+          if (spec_version != supported_version) {
+            console.log("Specification version does not match...")
+            const repo = context.repo.owner + "/" + context.repo.repo
+            const issues = await github.rest.search.issuesAndPullRequests({
+              q: "specification+has+new+version+in:title+state:open+type:issue:" + repo,
+            })
+            if (issues.data.total_count > 0) {
+              console.log("Issue is already open, not creating.")
+            } else {
+              console.log("Creating a new issue...")
+              await github.rest.issues.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                title: "specification has new version",
+                body: "It seems specification " +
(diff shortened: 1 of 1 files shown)


kairoaraujo:
First, thank you to @jku  for making it so easy to do.

Some additional information:

I run the tests on my forked repo.
The [first run](https://github.com/kairoaraujo/python-tuf/runs/4740564000?check_suite_focus=true) created the issue. 
The [second run]( https://github.com/kairoaraujo/python-tuf/actions/runs/1668195526) does not open as the issue is already opened ( + there is an issue already closed.)

I decided not to add any version information to not make the issue obsolete in case of a new version.

.github/workflows/specification-version.yml
```diff
@@ -0,0 +1,55 @@
...
+            owner: "theupdateframework",
+            repo: "specification"
+          });
+          const spec_version = release.data.tag_name
+          const supported_version = "${{ steps.get-version.outputs.version }}"
+          if (spec_version != supported_version) {
+            console.log("Specification version does not match...")
+            const repo = context.repo.owner + "/" + context.repo.repo
+            const issues = await github.rest.search.issuesAndPullRequests({
+              q: "specification+has+new+version+in:title+state:open+type:issue:" + repo,
```
jku: I never spent the time to test if you can give the repo and owner as separate parameters like in `issues.create()` which would probably be best... However, embedding the repo and owner in the query definitely works but the way it's done here looks like syntax error to me: the query string ends with `"type:issue:theupdateframework/python-tuf"`. Can you check this?

kairoaraujo: I didn't find specific separated parameters in the documentation.
https://docs.github.com/en/rest/reference/search#search-issues-and-pull-requests

But yes, for some reason, I removed the repository parameter. I fixed it now

MVrachev:
The CI failure is fixed in https://github.com/theupdateframework/python-tuf/pull/1762.
If you rebase after the merge of this pr it should be fixed.

.github/workflows/specification-version.yml
```diff
@@ -0,0 +1,55 @@
...
+            owner: "theupdateframework",
+            repo: "specification"
+          });
+          const spec_version = release.data.tag_name
+          const supported_version = "${{ steps.get-version.outputs.version }}"
+          if (spec_version != supported_version) {
+            console.log("Specification version does not match...")
+            const repo = context.repo.owner + "/" + context.repo.repo
+            const issues = await github.rest.search.issuesAndPullRequests({
+              q: "specification+has+new+version+in:title+state:open+type:issue+repo:" + repo,
```
jku, line 44: I'm guessing the "in:title" search might actually require quotes to get an exact search string but I think I'm fine with this: let's have another look if we get false positives in this.

jku (approved):
I think this is good enough, thanks.

I'm a bit on the fence about output (both logging and the issue message): It says what was done but includes nothing that would help debugging if things break (like the actual version strings that were compared)...

So, approving as is but please have a look at the logging/output: if you think it needs work we can do that

PR #1766, opened 2022-01-11
tests: Refactor key rotation in simulator

jku:
Add a method to rotate roles keys into RepositorySimulator (only
top-level roles are supported for now). Rotation is used in four
places already and this refactoring makes the tests easier to
understand.

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -163,6 +163,15 @@ class RepositorySimulator(FetcherInterface):
             self.signers[role] = {}
         self.signers[role][signer.key_dict["keyid"]] = signer
 
+    def rotate_keys(self, role: str) -> None:
+        """remove all keys for role, then add threshold of new keys"""
+        self.root.roles[role].keyids.clear()
+        self.signers[role].clear()
+        for _ in range(0, self.root.roles[role].threshold):
+            key, signer = self.create_key()
+            self.root.add_key(role, key)
+            self.add_signer(role, signer)
+
     def _initialize(self) -> None:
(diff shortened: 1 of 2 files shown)


tests/repository_simulator.py
```diff
@@ -163,6 +163,15 @@ def add_signer(self, role: str, signer: SSlibSigner) -> None:
...
         self.signers[role][signer.key_dict["keyid"]] = signer
 
+    def rotate_keys(self, role: str) -> None:
+        """remove all keys for role, then add threshold of new keys"""
+        self.root.roles[role].keyids.clear()
+        self.signers[role].clear()
+        for _ in range(0, self.root.roles[role].threshold):
+            snapshot_key, snapshot_signer = self.create_key()
+            self.root.add_key(role, snapshot_key)
+            self.add_signer(role, snapshot_signer)
```
kairoaraujo: We could use generic names here for the new keys and signer once it is valid for any role.
```suggestion
            key, signer = self.create_key()
            self.root.add_key(role, key)
            self.add_signer(role, signer)
```

kairoaraujo (commented):
Very nice refactoring.

tests/repository_simulator.py
jku: oh right, thanks

jku:
improved variable names as Kairo suggested, no other changes

kairoaraujo (approved):

MVrachev (approved):
Agree with Kairo.
Great refactoring.

lukpueh (approved):
Nice DRY, @jku!

PR #1769, opened 2022-01-13
doc: remove duplicate outdated docs

lukpueh:
Fixes #-

These documents are also hosted on our website and in a better shape too (kudos to @jhdalek55!). There is no need to maintain outdated duplicates.

This patch removes the documents and updates references as needed.

Note, the full history of these docs can be tracked in the old archived website repo and in the new one (filenames may vary in case and extension):

old: https://github.com/theupdateframework/theupdateframework.github.io
new: https://github.com/theupdateframework/theupdateframework.io/tree/master/content

requirements**:


mnm678 (approved):
Looks good! I agree that we should point people to one set of up-to-date docs

jku (approved):
I agree with this, looks good to me.

I wonder if we want to be more explicit in naming TUF as the specification (as in reference _"TUF specification website"_ instead of _"TUF website"_) to make the distinction between the projects (python-tuf and TUF) more clear to new readers? Anyway, it's already good :heavy_check_mark: .

lukpueh:
> I wonder if we want to be more explicit in naming TUF as the specification (as in reference _"TUF specification website"_ instead of _"TUF website"_) to make the distinction between the projects (python-tuf and TUF) more clear to new readers? 

IMO it's okay as it is because python-tuf is the official TUF reference implementation, which makes it less of an autonomous project than other TUF implementations. 🤷

jku:
I was more worried about confusion of type _"a link to TUF website when I'm already on the TUF website, that's just silly"_ -- but I agree it's pretty clear

PR #1774, opened 2022-01-17
Allow compressed metadata in flight

### LINKED ISSUE #1251: Client explicitly asks server to not compress data

After the acceptance of [TAP10 ](https://github.com/theupdateframework/taps/blob/master/tap10.md) the framework no longer handles compressed metadata.  Instead, 

> the specification should instead recommend that compression be implemented at the presentation layer of the OSI protocol stack, which is better suited to handle decompression of data. Once decompressed, metadata can be validated against the hashes and signatures listed in the Snapshot file.

However, the implementation explicitly [asks server to not compress data]( https://github.com/theupdateframework/tuf/blob/9fdbcd5b4bc3233b6583a611676b76f062a0b01d/tuf/download.py#L237) to avoid potential issues with some server configurations.

Two questions coming from this observation:
- Do all the calculations up to now (for example the ones in PEP458) consider uncompressed metadata?
- Should the implementation match tap10 and allow handling of compressed metadata at a lower level?

**Current behavior**:
Client explicitly asks server to not compress data

**Expected behavior**:
Allow handling of compressed metadata at the presentation layer

jku:
from commit description:
> This commit tries to deal with two interests:
>     * metadata is highly repetitive and compressible: allowing compression
>       would be good
>     * there may be broken web servers (see
>       https://github.com/pypa/pip/blob/404838abcca467648180b358598c597b74d568c9/src/pip/_internal/download.py#L842)
>       that have problems with compression on already compressed target files
>     
>  We can make things better for that first interest while we have no real
>     data for the second interest -- our current workarounds to avoid
>     compression are based on hearsay, not testing.
>     
>  Now that individual fetchers are possible I suggest we simplify
>     ngclient and allow compression. As an example the pip Fetcher
>     could still use the pip response chunking code with all their
>     workarounds -- pip certainly has better capability to maintain
>     a mountain of workarounds and also has endless amounts of real-world
>     testing compared to python-tuf.
>     
>  Details:
>     * Stop modifying Accept-Encoding (Requests default includes gzip)
>     * Don't use response.raw in RequestsFetcher as there is no need anymore
>     * Fix issue in test_session_get_timeout(): it's not mocking the error that requests really raises in this case
> 

Fixes #1251

### DIFF
--- a/tests/simple_server.py
+++ b/tests/simple_server.py
@@ -3,64 +3,16 @@
 # Copyright 2012 - 2017, New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-"""
-<Program>
-  simple_server.py
-
-<Author>
-  Konstantin Andrianov.
-
-<Started>
-  February 15, 2012.
-
-<Copyright>
-  See LICENSE-MIT or LICENSE for licensing information.
-
-<Purpose>
-  This is a basic server that was designed to be used in conjunction with
-  test_download.py to test download.py module.
-
-<Reference>
-  SimpleHTTPServer:
-    http://docs.python.org/library/simplehttpserver.html#module-SimpleHTTPServer
-"""
+"""Simple HTTP server for python-tuf tests"""
 
 import socketserver
-import sys
 from http.server import SimpleHTTPRequestHandler
-from typing import Type, Union
-
-
-class QuietHTTPRequestHandler(SimpleHTTPRequestHandler):
-    """A SimpleHTTPRequestHandler that does not write incoming requests to
-    stderr."""
-
-    def log_request(
-        self, code: Union[int, str] = "-", size: Union[int, str] = "-"
-    ) -> None:
-        pass
-
-
-# NOTE: On Windows/Python2 tests that use this simple_server.py in a
-# subprocesses hang after a certain amount of requests (~68), if a PIPE is
-# passed as Popen's stderr argument. This problem doesn't emerge if
-# we silence the HTTP messages.
-# If you decide to receive the HTTP messages, then this bug
-# could reappear.
-
-# pylint: disable=invalid-name
-handler: Type[Union[SimpleHTTPRequestHandler, QuietHTTPRequestHandler]]
-
-if len(sys.argv) > 2 and sys.argv[2]:
-    handler = QuietHTTPRequestHandler
-else:
-    handler = SimpleHTTPRequestHandler
 
 # Allow re-use so you can re-run tests as often as you want even if the
 # tests re-use ports. Otherwise TCP TIME-WAIT prevents reuse for ~1 minute
 socketserver.TCPServer.allow_reuse_address = True
 
-httpd = socketserver.TCPServer(("localhost", 0), handler)
+httpd = socketserver.TCPServer(("localhost", 0), SimpleHTTPRequestHandler)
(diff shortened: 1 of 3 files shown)


kairoaraujo:
LGTM
I'm not sure about the third commit as well.
As we talked about before, ``requests`` support it, so we are testing something outside the TUF domain.

jku:
Marking non-draft with only these changes:
* rebased on develop
* removed the last commit (_tests: Add gzip support to simple_server_): we don't want to test every Requests feature that is essentially opaque to us. The commit is still conserved in another branch: https://github.com/jku/python-tuf/tree/gzip-simple-server

tests/simple_server.py
```diff
@@ -3,64 +3,16 @@
...
-    handler = QuietHTTPRequestHandler
-else:
-    handler = SimpleHTTPRequestHandler
 
 # Allow re-use so you can re-run tests as often as you want even if the
 # tests re-use ports. Otherwise TCP TIME-WAIT prevents reuse for ~1 minute
 socketserver.TCPServer.allow_reuse_address = True
 
-httpd = socketserver.TCPServer(("localhost", 0), handler)
+httpd = socketserver.TCPServer(("localhost", 0), SimpleHTTPRequestHandler)
```
lukpueh, line 15: 👍  Thanks for cleaning up!

As the comment above says, I introduced this workaround for testing on Windows in Python2 (https://github.com/theupdateframework/python-tuf/commit/7dbb30ae1006d0b5f506bba19d4b53398c44f139). So this has become irrelevant quite a while ago.

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -99,26 +99,12 @@ def _chunks(self, response: "requests.Response") -> Iterator[bytes]:
...
-                #  misbehave:
-                # https://github.com/pypa/pip/blob/404838abcca467648180b358598c597b74d568c9/src/pip/_internal/download.py#L547-L582
-                data = response.raw.read(self.chunk_size)
-
-                # We might have no more data to read, we signal
-                # that the download is complete.
-                if not data:
-                    break
-
+            for data in response.iter_content(self.chunk_size):
```
lukpueh, line 102: Could you remind me why we used to read from `response.raw`,  why it is not necessary anymore, and how it is related to compression?

lukpueh: Do we give up any needed control about the incoming data stream?

jku: oh, great question, should have talked about that.

I believe this was added to pip for this case:
* broken servers might slap "Content-encoding: gzip" when they serve a file that happens to be compressed (without further compressing it): they might do this even if client did not ask for compression
* if response includes "Content-encoding: gzip", requests will decompress (even if explicitly did not request compression)
* the application can workaround this by reading(non-decompressed) response.raw but obviously you can only do that if you 100% know that you did not request compression -- otherwise it makes no sense

lukpueh: Thanks for the insight. Would you mind adding a short note about that to the commit message, where you say "Don't use response.raw in RequestsFetcher"?

jku:
modified commit message with an explanation of why we no longer use requests.Response.raw (which is really a urllib3.response.HTTPResponse)

lukpueh (approved):

PR #1775, opened 2022-01-17
Add  a "validate" argument option to JSONSerializer 

### LINKED ISSUE #1696: Metadata API: Provide a way to validate object when serializing to dictionary

MVrachev:
Fixes #1696, #1788

We can say we are almost done with Metadata API validation during the initialization of `Signed` objects as summarized here: https://github.com/theupdateframework/python-tuf/issues/1140#issuecomment-971588922.

What we **didn't** focus on is validation when serializing the Metadata objects by calling `Metadata.to_dict()`.

That option can be useful for package managers who import "TUF" into their lifecycle by giving them an assurance that when there are multiple changes of a metadata object then serializing this object to a file and then back to an instance from a file is possible.
I have added a separate API call `Metadata.validate()` that could be called to give this assurance.

This function can also be used by passing a `validate` argument when creating a `JsonSerializer`.
By default, this argument is **False**.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -39,7 +39,7 @@ from tuf.api.metadata import (
     Targets,
     Timestamp,
 )
-from tuf.api.serialization import DeserializationError
+from tuf.api.serialization import DeserializationError, SerializationError
 from tuf.api.serialization.json import CanonicalJSONSerializer, JSONSerializer
 
 logger = logging.getLogger(__name__)
@@ -156,6 +156,13 @@ class TestMetadata(unittest.TestCase):
             self.assertDictEqual(md_obj.to_dict(), md_obj_2.to_dict())
 
             os.remove(path_2)
+
+    def test_serialize_with_validate(self) -> None:
+        # Assert that by changing one required attribute validation will fail.
+        root = Metadata.from_file(f"{self.repo_dir}/metadata/root.json")
+        root.signed.version = 0
+        with self.assertRaises(SerializationError):
+            root.to_bytes(JSONSerializer(validate=True))
 
     def test_to_from_bytes(self) -> None:
         for metadata in TOP_LEVEL_ROLE_NAMES:
--- a/tests/test_metadata_eq_.py
+++ b/tests/test_metadata_eq_.py
@@ -1,0 +1,243 @@
+#!/usr/bin/env python
+
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Test __eq__ implementations of classes inside tuf/api/metadata.py."""
+
+
+import copy
+import os
+import sys
+import unittest
+from typing import Any, ClassVar, Dict
+
+from tests import utils
+from tuf.api.metadata import (
+    TOP_LEVEL_ROLE_NAMES,
+    DelegatedRole,
+    Delegations,
+    Key,
+    Metadata,
+    MetaFile,
+    Role,
+    Root,
+    Snapshot,
+    TargetFile,
+    Targets,
+    Timestamp,
+)
+
+
+class TestMetadataComparisions(unittest.TestCase):
+    """Test __eq__ for all classes inside tuf/api/metadata.py."""
+
+    metadata: ClassVar[Dict[str, bytes]]
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.repo_dir = os.path.join(
+            utils.TESTS_DIR, "repository_data", "repository", "metadata"
+        )
+        cls.metadata = {}
+        for md in TOP_LEVEL_ROLE_NAMES:
+            with open(os.path.join(cls.repo_dir, f"{md}.json"), "rb") as f:
+                cls.metadata[md] = f.read()
+
+    def copy_and_simple_assert(self, obj: Any) -> Any:
+        # Assert that obj is not equal to an object from another type
+        self.assertNotEqual(obj, "")
+        result_obj = copy.deepcopy(obj)
+        # Assert that __eq__ works for equal objects.
+        self.assertEqual(obj, result_obj)
+        return result_obj
+
+    def test_metadata_eq_(self) -> None:
+        md = Metadata.from_bytes(self.metadata["snapshot"])
+        md_2: Metadata = self.copy_and_simple_assert(md)
+
+        for attr, value in [("signed", None), ("signatures", None)]:
+            setattr(md_2, attr, value)
+            self.assertNotEqual(md, md_2, f"Failed case: {attr}")
+
+    def test_md_eq_special_signatures_tests(self) -> None:
+        # Test that metadata objects with different signatures are not equal.
+        md = Metadata.from_bytes(self.metadata["snapshot"])
+        md_2 = copy.deepcopy(md)
+        md_2.signatures = {}
+        self.assertNotEqual(md, md_2)
+
+        # Test that metadata objects with empty signatures are equal
+        md.signatures = {}
+        self.assertEqual(md, md_2)
+
+        # Metadata objects with different signatures types are not equal.
+        md_2.signatures = ""  # type: ignore
+        self.assertNotEqual(md, md_2)
+
+    def test_signed_eq_(self) -> None:
+        md = Metadata.from_bytes(self.metadata["snapshot"])
+        md_2: Metadata = self.copy_and_simple_assert(md)
+
+        # We don't need to make "signed" = None as that was done when testing
+        # metadata attribute modifications.
+        for attr, value in [("version", -1), ("spec_version", "0.0.0")]:
+            setattr(md_2.signed, attr, value)
+            self.assertNotEqual(md.signed, md_2.signed, f"Failed case: {attr}")
+
+    def test_key_eq_(self) -> None:
+        key_dict = {
+            "keytype": "rsa",
+            "scheme": "rsassa-pss-sha256",
+            "keyval": {"public": "foo"},
+        }
+        key = Key.from_dict("12sa12", key_dict)
+        key_2: Key = self.copy_and_simple_assert(key)
+        for attr, value in [
+            ("keyid", "a"),
+            ("keytype", "foo"),
+            ("scheme", "b"),
+            ("keytype", "b"),
+        ]:
+            setattr(key_2, attr, value)
+            self.assertNotEqual(key, key_2, f"Failed case: {attr}")
+
+    def test_role_eq_(self) -> None:
+        role_dict = {
+            "keyids": ["keyid1", "keyid2"],
+            "threshold": 3,
+        }
+        role = Role.from_dict(role_dict)
+        role_2: Role = self.copy_and_simple_assert(role)
+
+        for attr, value in [("keyids", []), ("threshold", 10)]:
+            setattr(role_2, attr, value)
+            self.assertNotEqual(role, role_2, f"Failed case: {attr}")
+
+    def test_root_eq_(self) -> None:
+        md = Metadata.from_bytes(self.metadata["root"])
+        signed_copy: Root = self.copy_and_simple_assert(md.signed)
+
+        # Common attributes between Signed and Root doesn't need testing.
+        # Ignore mypy request for type annotations on attr and value
+        for attr, value in [  # type: ignore
+            ("consistent_snapshot", None),
+            ("keys", {}),
+            ("roles", {}),
+        ]:
+
+            setattr(signed_copy, attr, value)
+            self.assertNotEqual(md.signed, signed_copy, f"Failed case: {attr}")
+
+    def test_metafile_eq_(self) -> None:
+        metafile_dict = {
+            "version": 1,
+            "length": 12,
+            "hashes": {"sha256": "abc"},
+        }
+        metafile = MetaFile.from_dict(metafile_dict)
+        metafile_2: MetaFile = self.copy_and_simple_assert(metafile)
+
+        # Ignore mypy request for type annotations on attr and value
+        for attr, value in [  # type: ignore
+            ("version", None),
+            ("length", None),
+            ("hashes", {}),
+        ]:
+            setattr(metafile_2, attr, value)
+            self.assertNotEqual(metafile, metafile_2, f"Failed case: {attr}")
+
+    def test_timestamp_eq_(self) -> None:
+        md = Metadata.from_bytes(self.metadata["timestamp"])
+        signed_copy: Timestamp = self.copy_and_simple_assert(md.signed)
+
+        # Common attributes between Signed and Timestamp doesn't need testing.
+        setattr(signed_copy, "snapshot_meta", None)
+        self.assertNotEqual(md.signed, signed_copy)
+
+    def test_snapshot_eq_(self) -> None:
+        md = Metadata.from_bytes(self.metadata["snapshot"])
+        signed_copy: Snapshot = self.copy_and_simple_assert(md.signed)
+
+        # Common attributes between Signed and Snapshot doesn't need testing.
+        setattr(signed_copy, "meta", None)
+        self.assertNotEqual(md.signed, signed_copy)
+
+    def test_delegated_role_eq_(self) -> None:
+        delegated_role_dict = {
+            "keyids": ["keyid"],
+            "name": "a",
+            "terminating": False,
+            "threshold": 1,
+            "paths": ["fn1", "fn2"],
+        }
+        delegated_role = DelegatedRole.from_dict(delegated_role_dict)
+        delegated_role_2: DelegatedRole = self.copy_and_simple_assert(
+            delegated_role
+        )
+
+        # Common attributes between DelegatedRole and Role doesn't need testing.
+        for attr, value in [
+            ("name", ""),
+            ("terminating", None),
+            ("paths", [""]),
+            ("path_hash_prefixes", [""]),
+        ]:
+            setattr(delegated_role_2, attr, value)
+            msg = f"Failed case: {attr}"
+            self.assertNotEqual(delegated_role, delegated_role_2, msg)
+
+    def test_delegations_eq_(self) -> None:
+        delegations_dict = {
+            "keys": {
+                "keyid2": {
+                    "keytype": "ed25519",
+                    "scheme": "ed25519",
+                    "keyval": {"public": "bar"},
+                }
+            },
+            "roles": [
+                {
+                    "keyids": ["keyid2"],
+                    "name": "b",
+                    "terminating": True,
+                    "paths": ["fn2"],
+                    "threshold": 4,
+                }
+            ],
+        }
+        delegations = Delegations.from_dict(delegations_dict)
+        delegations_2: Delegations = self.copy_and_simple_assert(delegations)
+        # Ignore mypy request for type annotations on attr and value
+        for attr, value in [("keys", {}), ("roles", {})]:  # type: ignore
+            setattr(delegations_2, attr, value)
+            msg = f"Failed case: {attr}"
+            self.assertNotEqual(delegations, delegations_2, msg)
+
+    def test_targetfile_eq_(self) -> None:
+        targetfile_dict = {
+            "length": 12,
+            "hashes": {"sha256": "abc"},
+        }
+        targetfile = TargetFile.from_dict(targetfile_dict, "file1.txt")
+        targetfile_2: TargetFile = self.copy_and_simple_assert(targetfile)
+
+        # Common attr between TargetFile and MetaFile doesn't need testing.
+        setattr(targetfile_2, "path", "")
+        self.assertNotEqual(targetfile, targetfile_2)
+
+    def test_targets_eq_(self) -> None:
+        md = Metadata.from_bytes(self.metadata["targets"])
+        signed_copy: Targets = self.copy_and_simple_assert(md.signed)
+
+        # Common attributes between Targets and Signed doesn't need testing.
+        # Ignore mypy request for type annotations on attr and value
+        for attr, value in [("targets", {}), ("delegations", [])]:  # type: ignore
+            setattr(signed_copy, attr, value)
+            self.assertNotEqual(md.signed, signed_copy, f"Failed case: {attr}")
+
+
+# Run unit test.
+if __name__ == "__main__":
+    utils.configure_test_logging(sys.argv)
+    unittest.main()
--- a/tests/test_metadata_eq_.py
+++ b/tests/test_metadata_eq_.py
@@ -12,6 +12,8 @@ import sys
 import unittest
 from typing import Any, ClassVar, Dict
 
+from securesystemslib.signer import Signature
+
 from tests import utils
 from tuf.api.metadata import (
     TOP_LEVEL_ROLE_NAMES,
@@ -59,6 +61,23 @@ class TestMetadataComparisions(unittest.TestCase):
         for attr, value in [("signed", None), ("signatures", None)]:
             setattr(md_2, attr, value)
             self.assertNotEqual(md, md_2, f"Failed case: {attr}")
+
+    def test_md_eq_signatures_reversed_order(self) -> None:
+        # Test comparing objects with same signatures but different order.
+
+        # Remove all signatures and create new ones.
+        md = Metadata.from_bytes(self.metadata["snapshot"])
+        md.signatures = {"a": Signature("a", "a"), "b": Signature("b", "b")}
+        md_2 = copy.deepcopy(md)
+        # Reverse signatures order in md_2.
+        # In python3.7 we need to cast to a list and then reverse.
+        md_2.signatures = dict(reversed(list(md_2.signatures.items())))
+        # Assert that both objects are not the same because of signatures order.
+        self.assertNotEqual(md, md_2)
+
+        # but if we fix the signatures order they will be equal
+        md_2.signatures = {"a": Signature("a", "a"), "b": Signature("b", "b")}
+        self.assertEqual(md, md_2)
 
     def test_md_eq_special_signatures_tests(self) -> None:
         # Test that metadata objects with different signatures are not equal.
@@ -225,6 +244,48 @@ class TestMetadataComparisions(unittest.TestCase):
         # Common attr between TargetFile and MetaFile doesn't need testing.
         setattr(targetfile_2, "path", "")
         self.assertNotEqual(targetfile, targetfile_2)
+
+    def test_delegations_eq_roles_reversed_order(self) -> None:
+        # Test comparing objects with same delegated roles but different order.
+        role_one_dict = {
+            "keyids": ["keyid1"],
+            "name": "a",
+            "terminating": False,
+            "paths": ["fn1"],
+            "threshold": 1,
+        }
+        role_two_dict = {
+            "keyids": ["keyid2"],
+            "name": "b",
+            "terminating": True,
+            "paths": ["fn2"],
+            "threshold": 4,
+        }
+
+        delegations_dict = {
+            "keys": {
+                "keyid2": {
+                    "keytype": "ed25519",
+                    "scheme": "ed25519",
+                    "keyval": {"public": "bar"},
+                }
+            },
+            "roles": [role_one_dict, role_two_dict],
+        }
+        delegations = Delegations.from_dict(copy.deepcopy(delegations_dict))
+
+        # Create a second delegations obj with reversed roles order
+        delegations_2 = copy.deepcopy(delegations)
+        # In python3.7 we need to cast to a list and then reverse.
+        delegations_2.roles = dict(reversed(list(delegations.roles.items())))
+
+        # Both objects are not the equal because of delegated roles order.
+        self.assertNotEqual(delegations, delegations_2)
+
+        # but if we fix the delegated roles order they will be equal
+        delegations_2.roles = delegations.roles
+
+        self.assertEqual(delegations, delegations_2)
 
     def test_targets_eq_(self) -> None:
         md = Metadata.from_bytes(self.metadata["targets"])
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -127,6 +127,16 @@ class Metadata(Generic[T]):
         self.signatures = signatures
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Metadata):
+            return False
+
+        return (
+            self.signatures == other.signatures
+            and self.signed == other.signed
+            and self.unrecognized_fields == other.unrecognized_fields
+        )
+
     @classmethod
     def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata[T]":
         """Creates ``Metadata`` object from its dict representation.
@@ -490,6 +500,18 @@ class Signed(metaclass=abc.ABCMeta):
         self.version = version
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Signed):
+            return False
+
+        return (
+            self.type == other.type
+            and self.version == other.version
+            and self.spec_version == other.spec_version
+            and self.expires == other.expires
+            and self.unrecognized_fields == other.unrecognized_fields
+        )
+
     @abc.abstractmethod
     def to_dict(self) -> Dict[str, Any]:
         """Serialization helper that returns dict representation of self"""
@@ -597,6 +619,18 @@ class Key:
         self.scheme = scheme
         self.keyval = keyval
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Key):
+            return False
+
+        return (
+            self.keyid == other.keyid
+            and self.keytype == other.keytype
+            and self.scheme == other.scheme
+            and self.keyval == other.keyval
+            and self.unrecognized_fields == other.unrecognized_fields
+        )
 
     @classmethod
     def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "Key":
@@ -742,6 +776,16 @@ class Role:
         self.threshold = threshold
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Role):
+            return False
+
+        return (
+            self.keyids == other.keyids
+            and self.threshold == other.threshold
+            and self.unrecognized_fields == other.unrecognized_fields
+        )
+
     @classmethod
     def from_dict(cls, role_dict: Dict[str, Any]) -> "Role":
         """Creates ``Role`` object from its dict representation.
@@ -803,6 +847,17 @@ class Root(Signed):
             raise ValueError("Role names must be the top-level metadata roles")
 
         self.roles = roles
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Root):
+            return False
+
+        return (
+            super().__eq__(other)
+            and self.keys == other.keys
+            and self.roles == other.roles
+            and self.consistent_snapshot == other.consistent_snapshot
+        )
 
     @classmethod
     def from_dict(cls, signed_dict: Dict[str, Any]) -> "Root":
@@ -986,6 +1041,17 @@ class MetaFile(BaseFile):
         self.hashes = hashes
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, MetaFile):
+            return False
+
+        return (
+            self.version == other.version
+            and self.length == other.length
+            and self.hashes == other.hashes
+            and self.unrecognized_fields == other.unrecognized_fields
+        )
+
     @classmethod
     def from_dict(cls, meta_dict: Dict[str, Any]) -> "MetaFile":
         """Creates ``MetaFile`` object from its dict representation.
@@ -1066,6 +1132,14 @@ class Timestamp(Signed):
         super().__init__(version, spec_version, expires, unrecognized_fields)
         self.snapshot_meta = snapshot_meta
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Timestamp):
+            return False
+
+        return (
+            super().__eq__(other) and self.snapshot_meta == other.snapshot_meta
+        )
+
     @classmethod
     def from_dict(cls, signed_dict: Dict[str, Any]) -> "Timestamp":
         """Creates ``Timestamp`` object from its dict representation.
@@ -1118,6 +1192,12 @@ class Snapshot(Signed):
     ):
         super().__init__(version, spec_version, expires, unrecognized_fields)
         self.meta = meta
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Snapshot):
+            return False
+
+        return super().__eq__(other) and self.meta == other.meta
 
     @classmethod
     def from_dict(cls, signed_dict: Dict[str, Any]) -> "Snapshot":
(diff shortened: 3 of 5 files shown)


jku:
What I was trying to say earlier (and I think Lukas as well) is that the validation implementation should be in the JSONSerializer. Otherwise you end up serializing twice as this PR does. Metadata.validate() does not need to exist as far as I can see.

Is there a SSLIB bug to implement eq for signature?

JSONSerializer.serialize() is never tested when validate=True. I suppose we want most of our tests to use validate=True?

lukpueh:
> What I was trying to say earlier (and I think Lukas as well) is that the validation implementation should be in the JSONSerializer. Otherwise you end up serializing twice as this PR does. Metadata.validate() does not need to exist as far as I can see.

Exactly. The `validate` implementation is too JSONDe/Serializer-specific for the Metadata class. Plus Jussi's 2x-to_dict argument.
 
> Is there a SSLIB bug to implement eq for signature?

Does not look like it. @MVrachev, would you mind creating a ticket and referencing it in a TODO comment in `Metadata.__eq__()`?

> 
> JSONSerializer.serialize() is never tested when validate=True. I suppose we want most of our tests to use validate=True?

Agreed.

tuf/api/metadata.py
```diff
@@ -283,6 +311,24 @@ def to_file(
...
+        Raises:
+            ValueError: If self cannot be validated
+        """
+
+        temp_dict = self.to_dict()
+        new_self_obj = Metadata.from_dict(temp_dict)
+        try:
+            if self != new_self_obj:
+                raise ValueError("Objects are not equal")
+        except Exception as e:
```
lukpueh: What exceptions do you expect here? Should the `to_dict` and `from_dict` calls also be in the try block?

MVrachev: `Metadata.from_dict()` can raise `KeyError: The metadata dict format is invalid.`  or `ValueError: The metadata has an unrecognized signed._type field.`
I moved this part inside the `JSONSerializer.serliaze()` and you can have a look there.

MVrachev:
I rebased and updated this pr

> What I was trying to say earlier (and I think Lukas as well) is that the validation implementation should be in the JSONSerializer. Otherwise you end up serializing twice as this PR does. Metadata.validate() does not need to exist as far as I can see.

I removed `Metadata.validate()` and moved it into `JSONSerializer.serialize()`.

> Is there a SSLIB bug to implement eq for signature?

I proposed a pr with that: https://github.com/secure-systems-lab/securesystemslib/pull/383

> JSONSerializer.serialize() is never tested when validate=True. I suppose we want most of our tests to use validate=True?

I added a simple test for that and changed one of the places where we instantiate a JSONSerializer to use validation.
Many of the places where we call `to_bytes` we call it when we want to raise another specific error and It didn't make sense to use JSONSerializer with `validate = True`.
I also don't think it makes sense to comprehensively test this part with all of the classes and their attributes as essentially the when calling `JSONSerializer.serialize` with validation turned on we are relying on the classes initialization validation which already is tested extensively in `test_trusted_metadata_set.py`.

tuf/api/metadata.py
```diff
@@ -114,6 +114,36 @@ def __init__(self, signed: T, signatures: "OrderedDict[str, Signature]"):
...
 
+    def __eq__(self, other: "Any") -> bool:
+        if not isinstance(other, Metadata):
+            return False
+
+        # TODO: Implement "__eq__" method in securesystemslib Signature class.
+
+        # Verifies that the signatures for both objects are the same.
+        # Signatures dictionaries can't be directly compared as their values
+        # are securesystemslib Signature objects which don't implement __eq__.
```
```suggestion
        # Verifies that the signatures for both objects are the same.
        # # TODO: Use Signature.__eq__ (secure-systems-lab/securesystemslib#383)
```

tuf/api/metadata.py
```diff
@@ -114,6 +114,36 @@ def __init__(self, signed: T, signatures: "OrderedDict[str, Signature]"):
         self.signed: T = signed
         self.signatures = signatures
 
+    def __eq__(self, other: "Any") -> bool:
```
```suggestion
    def __eq__(self, other: Any) -> bool:
```
Same comment applies elsewhere in this PR.

tuf/api/metadata.py
```diff
@@ -114,6 +114,36 @@ def __init__(self, signed: T, signatures: "OrderedDict[str, Signature]"):
...
+        # Signatures dictionaries can't be directly compared as their values
+        # are securesystemslib Signature objects which don't implement __eq__.
+        if (self.signatures is not None and other.signatures is None) or (
+            self.signatures is None and other.signatures is not None
+        ):
+            return False
+        if self.signatures is None and other.signatures is None:
+            return self.signed == other.signed
+
+        # Here we know both "self" and other have "signatures"
```
lukpueh: We actually don't know that. We only know that either both or neither of them is `None`. This means that the calls to `keys()` and `values()` below could result in `AttributeErrors`.

lukpueh: So I'd probably check first that `self.signatures` and `other.signatures` have the same type, then I'd check if `self.signatures` is the type we expect. If it is, you can go on with your custom `Signature.__eq__` check here. If it isn't, I'd leave it to whatever type it is to implement `__eq__`, and just use that.

tuf/api/metadata.py
```diff
@@ -448,6 +478,18 @@ def __init__(
...
+    def __eq__(self, other: "Any") -> bool:
+        if not isinstance(other, Signed):
+            return False
+
+        return (
+            self.type == other.type
+            and self.version == other.version
+            and self.spec_version == other.spec_version
+            and self.expires == other.expires
+            and self.unrecognized_fields == other.unrecognized_fields
```
lukpueh, line 514: Is a shallow comparison enough for this field? Comment applies to all `self.unrecognized_fields == other.unrecognized_fields` in this PR.

lukpueh: On a more general note, could it be that we run into problems with this field when doing the `data==deserialize(serialize(data))` check? It seems we include the field in serialization (to_dict), but ignore it in deserialization (from_dict), is that correct?

tuf/api/serialization/json.py
```diff
@@ -47,19 +49,31 @@ class JSONSerializer(MetadataSerializer):
...
         """Serialize Metadata object into utf-8 encoded JSON bytes."""
+
+        md_dict = metadata_obj.to_dict()
+        if self.validate:
+            try:
+                new_md_obj = Metadata.from_dict(copy.deepcopy(md_dict))
+                if metadata_obj != new_md_obj:
+                    raise ValueError("Objects are not equal")
+            except Exception as e:
+                raise ValueError("Metadata content cannot be validated!") from e
```
lukpueh: I think all of this should remain in the original try block so that calls to `serialize` really only raise `SerializationError`s.

tuf/api/serialization/json.py
```diff
@@ -47,19 +49,31 @@ class JSONSerializer(MetadataSerializer):
...
 
     def serialize(self, metadata_obj: Metadata) -> bytes:
         """Serialize Metadata object into utf-8 encoded JSON bytes."""
+
+        md_dict = metadata_obj.to_dict()
+        if self.validate:
+            try:
+                new_md_obj = Metadata.from_dict(copy.deepcopy(md_dict))
+                if metadata_obj != new_md_obj:
+                    raise ValueError("Objects are not equal")
```
lukpueh: Is it worth to mention what objects are not equal? The caller might not know how we validate here. What about
```suggestion
                    raise ValueError("metadata_obj != deserialize(serialize(metadata_obj))")
```
Is that to technical?  Alternatively (or in addition) we could also explain what `validate` means above in the attribute docstring.

tests/test_api.py
```diff
@@ -146,6 +147,49 @@ def test_read_write_read_compare(self) -> None:
 
             os.remove(path_2)
 
+    def assertEqualRecursively(self, var: Any) -> None:
```
lukpueh: Minor nit: The name is a bit misleading, I briefly thought this was a builtin of `unittest.TestCase` akin to `assertEqual` and was confused, why it only received one argument.

 I suggest a name that indicates that this is a test helper, something like `_test_metadata_eq_methods` or so.

tests/test_api.py
```diff
@@ -146,6 +147,49 @@ def test_read_write_read_compare(self) -> None:
...
+                self.assertNotEqual(var, obj_2)
+
+                self.assertEqualRecursively(getattr(var, attr))
+
+    def test_md__eq__(self) -> None:
+        # Test that the "__eq__" implementations work for each of the top level
+        # metadata classes and for each of the helper classes.
+        for role in TOP_LEVEL_ROLE_NAMES:
+            md_obj = Metadata.from_file(f"{self.repo_dir}/metadata/{role}.json")
+            self.assertEqualRecursively(md_obj)
```
lukpueh: The implementation of assertEqualRecursively looks pretty smart. But wouldn't it be enough to just call `assertEqual` on `md_obj` and a `deepcopy` of it? If your hierarchy of `__eq__` weren't implemented this should also fail.

And regarding the `assertNotEqual`, if you look at the coverage, you'll see that not all branches in your `__eq__` implementations are covered. And at least to me it's not immediately clear which condition in `__eq__` should match after your `setattr(obj_2, attr, None)`.

Maybe it would be more obvious and yield better coverage if you just iterate through a table of unequal  (for each condition you test in `__eq__`) object pairs for each type and call `asserEqual` on them.

lukpueh (changes requested):
Cool stuff, @MVrachev! Please address my comments and we should be able to soon merge.

Regarding tests, I do think it would be good to add a few to `test_metadata_serialization.py` that call `serialize` with `self.validate` being `True` and a `metadata_ob` that triggers the different validation errors.

lukpueh:
I just merged #1783. Please make sure that we don't run into dict order problems you kindly pointed out in https://github.com/theupdateframework/python-tuf/pull/1783/commits/e3b267e2e0799673ac99ccfccd3631628013201c!

tuf/api/metadata.py
MVrachev: > Is a shallow comparison enough for this field? Comment applies to all self.unrecognized_fields == other.unrecognized_fields in this PR.

This is not a problem for us as we don't know what our users can put there. We don't handle `unrecognized_fields` anyway. If the `unrecognized_fields` is a dictionary using built-in types let's say `Dict[str, int]` we wouldn't have problems comparing it. If the user is adding object instances there he/she has the responsibility to implement his/her own `__eq__` method.

> On a more general note, could it be that we run into problems with this field when doing the data==deserialize(serialize(data)) check? It seems we include the field in serialization (to_dict), but ignore it in deserialization (from_dict), is that correct?

We don't ignore `unrecognized_fields`. We populate each of them inside the `from_dict()` methods and serialize them inside the `to_dict()` methods.
We comprehensively test them. See #1466 where we add them to the `test_metadata_serialization.py`.

tuf/api/metadata.py
```diff
@@ -735,6 +799,18 @@ def __init__(
...
+    def __eq__(self, other: "Any") -> bool:
+        if not isinstance(other, Root):
+            return False
+
+        return (
+            super().__eq__(other)
+            and self.keys == other.keys
+            and self.roles == other.roles
+            and self.consistent_snapshot == other.consistent_snapshot
+            and self.unrecognized_fields == other.unrecognized_fields
```
jku: unrecognized_fields is handled by super() already, correct?

This comment applies to all signed-derivatives

tuf/api/serialization/json.py
```diff
@@ -47,19 +49,31 @@ class JSONSerializer(MetadataSerializer):
     Attributes:
         compact: A boolean indicating if the JSON bytes generated in
                 'serialize' should be compact by excluding whitespace.
+        validate: Whether to validate when serializing.
```
jku: Can you try to improve this? It's not easy but the idea is to give the reader enough information to choose True or False without reading the source code: could try to mention that it exists to prevent programming errors from leading to changes in metadata serialize/deserialize cycle , but maybe also the cost.

tuf/api/serialization/json.py
jku: +1 for doscstring improvement: it doesn't even necessarily have to be in the `validate` argument docstring but could be in serialize() or JSONSerializer object docstring... wherever it seems to make sense

jku (changes requested):
Structure looks good to me. I left a few more comments in source.

MVrachev:
It's good to solve issue #1788 inside this pr as well.

MVrachev:
I rebased this pr on top of the latest `develop` changes and made the following changes:
1. As suggested by Jussi removed `unrecognized_fields` check inside all `__eq__` implementation in Signed class derivates as it's checked in `Signed.__eq__`
2. Inside `Metadata.__eq__` I realized there is no sense in testing if `self.signatures` is `None` or `other.signatures` is `None` or if both of them are `None`.
All of those checks can be replaced by `type(self.signatures) == type(other.signatures) == dict`.
3. I rewrote the tests. I created a separate `test_metadata_eq_` file where similarly to `test_metadata_serialization.py` I used table testing to write mostly simple tests. I checked if all cases are covered with local `coverall` runs.
4. I added made sure that the dictionaries order is taken into account when comparing signatures or delegated roles (see #1788)
5. Made sure only `SerializationError` will be thrown by `JSONSerializer.serialize()`
6. Changed the documentation regarding the `validate` argument.

I know the changes look huge, but that's mostly from `test_metadata_eq` which should be simple to read.

MVrachev:
I had to rebase a couple of times because GitHub showed changes from previous commits.

tests/test_metadata_eq_.py
```diff
@@ -0,0 +1,412 @@
...
+# pylint: disable=no-member
+class TestMetadataComparisions(unittest.TestCase):
+    """Test __eq__ for all classes inside tuf/api/metadata.py."""
+
+    metadata: ClassVar[Dict[str, bytes]]
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.repo_dir = os.path.join(
+            os.getcwd(), "repository_data", "repository", "metadata"
```
lukpueh: May I suggest to use [`TEST_DIR`](
https://github.com/theupdateframework/python-tuf/blob/dbc4bb615cdfdc2c9e09d249abb18c369aafec20/tests/utils.py#L41) instead of `getcwd`? This allows us to run the tests from anywhere, not only from inside the `tests` dir.

tuf/api/metadata.py
```diff
@@ -116,6 +116,33 @@ def __init__(self, signed: T, signatures: Dict[str, Signature]):
...
         self.signatures = signatures
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Metadata):
+            return False
+
+        # Verifies that the signatures for both objects are the same.
+        # TODO: Use Signature.__eq__ (secure-systems-lab/securesystemslib#383)
+        # pylint: disable=unidiomatic-typecheck
+        if not (type(self.signatures) == type(other.signatures) == dict):
```
lukpueh: I guess the idiomatic check would be

```suggestion
        if not (isinstance(self.signatures, dict) and isinstance(other.signatures, dict)):
```

tuf/api/metadata.py
```diff
@@ -116,6 +116,33 @@ def __init__(self, signed: T, signatures: Dict[str, Signature]):
...
+            return False
+
+        # Verifies that signatures keyids for both objects are the same
+        if self.signatures.keys() != other.signatures.keys():
+            return False
+
+        # Iterate over self.signatures and other.signatures at the same time
+        # as the signatures in the file format are ordered.
+        keyids = zip(self.signatures.keys(), other.signatures.keys())
```
lukpueh: Why zipping two lists you just checked to be equal?

tuf/api/metadata.py
```diff
@@ -1237,6 +1347,30 @@ def __init__(
...
+        if not isinstance(other, Delegations):
+            return False
+
+        # Assert that roles keys are the same before traversing them.
+        if self.roles.keys() != other.roles.keys():
+            return False
+
+        # Iterate over self.roles and other.roles at the same time
+        # as the order of the roles dictionaries is important.
+        role_names = zip(self.roles.keys(), other.roles.keys())
```
lukpueh: Why zipping two lists you just checked to be equal?

tuf/api/serialization/json.py
```diff
@@ -47,19 +49,35 @@ class JSONSerializer(MetadataSerializer):
     Attributes:
         compact: A boolean indicating if the JSON bytes generated in
                 'serialize' should be compact by excluding whitespace.
+        validate: Whether to validate when serializing. It's used to prevent
+            common programming mistakes when changing the metadata objects.
```
lukpueh: I think this is unnecessarily vague. What about:

```suggestion
        validate: Check that the metadata object can be deserialized again
            without change of contents. This validation might slow down
            serialization significantly.
```

lukpueh (changes requested):
Thanks for addressing my earlier comments and adding all these tests. I noticed a few more things. Please address and ping me for another round...

tuf/api/metadata.py
MVrachev: Yes, but it becomes too long more than 80 characters... When I format it with black it looks like this:

```python
        if not (
            isinstance(self.signatures, dict)
            and isinstance(other.signatures, dict)
        ):
```
which is ugly.
That's why I decided to use `type` as it's shorter and triple `==` even though it's not idiomatic, but it does the same job.
Still, if you prefer your suggestion I will replace it.

lukpueh: I do have a slight preference because you use `isinstance`-checks everywhere else in your `__eq__` implementations, so it seems consistent to also use it here. That said, I don't think that the subtle difference between `isinstance` and `== type` in this case will ever be a problem. So you decide.

tuf/api/serialization/json.py
MVrachev: I agree with you, but also I want to mention that this helps find common mistakes.
So, I made it: 
```
Check that the metadata object can be deserialized again
without change of contents and thus find common mistakes.
This validation might slow down serialization significantly.
```

tuf/api/metadata.py
MVrachev: Will follow your suggestion. :))

tuf/api/metadata.py
MVrachev: After the check `if self.signatures.keys() != other.signatures.keys():` we know that `self` and `other` have the same keys, but we don't know if they follow the same order.

When I zip them I make sure I traverse both of them following their corresponding insertion order.
If the `self.signatures` and `other.signatures` orders are different, then when I check (a couple of lines below) `self_keyid != other_keyid` will return False.

I tried without zipping, but it fails the test `test_md_eq_signatures_reversed_order` inside `test_metadata_eq_.py`.

tuf/api/metadata.py
```diff
@@ -116,6 +116,33 @@ def __init__(self, signed: T, signatures: Dict[str, Signature]):
...
+        if not isinstance(other, Metadata):
+            return False
+
+        # Verifies that the signatures for both objects are the same.
+        # TODO: Use Signature.__eq__ (secure-systems-lab/securesystemslib#383)
+        # pylint: disable=unidiomatic-typecheck
+        if not (type(self.signatures) == type(other.signatures) == dict):
+            return False
+
+        # Verifies that signatures keyids for both objects are the same
```
jku: this comment explains only the obvious -- "why" is what requires explanation (at least I assume the reason is to ensure the dict ordering as dict does not take care of it)

tuf/api/metadata.py
MVrachev: The same comment as for `Metadata.signatures`.

We need zipping in order to traverse both dictionaries in their corresponding order.
I just tried without zipping and my test `test_delegations_eq_roles_reversed_order` inside `test_metadata_eq_.py` fails.

tests/test_metadata_eq_.py
MVrachev: I agree, but don't know why I don't succeed in running the tests from the `tuf` folder.

tuf/api/metadata.py
jku: > After the check if self.signatures.keys() != other.signatures.keys(): we know that self and other have the same keys, but we don't know if they follow the same order.

but doesn't that check actually tell us that the order is correct: List eq does check the order?

Then that would mean you don't need the loop at all -- could just check `self.signatures != other.signatures` since now you know the order is the same

lukpueh: Ah right, good point, because `keys()` doesn't return a list anymore in Python3 but rather something set-like and thus unordered.

But still, you're comparing keyids twice here. You could instead only check the lengths at first, i.e.  `len(self.signatures.keys()) != len(other.signatures.keys())`, or just pass `strict=True` to `zip`, which does the same thing.

tuf/api/serialization/json.py
```diff
@@ -44,22 +46,39 @@ def deserialize(self, raw_data: bytes) -> Metadata:
...
 
     def serialize(self, metadata_obj: Metadata) -> bytes:
         """Serialize Metadata object into utf-8 encoded JSON bytes."""
+
         try:
+            md_dict = metadata_obj.to_dict()
+
+            if self.validate:
+                try:
+                    new_md_obj = Metadata.from_dict(copy.deepcopy(md_dict))
```
jku: is this the correct place to hook in? Now we go from Metadata to "intermediate dictionary format" back to Metadata: so we never actually serialize to json during the validation.

We could serialize all the way to json, then do `JSONDeserializer().deserialize()` to go back to metadata maybe?

tuf/api/metadata.py
lukpueh: >  could just check self.signatures != other.signatures

Not until https://github.com/secure-systems-lab/securesystemslib/pull/383 is merged and released.

lukpueh: > > could just check self.signatures != other.signatures
> 
> Not until [secure-systems-lab/securesystemslib#383](https://github.com/secure-systems-lab/securesystemslib/pull/383) is merged and released.

... maybe we should just go ahead and do that, before spending more thoughts on `zip` :D

MVrachev:
I closed it by mistake.

MVrachev:
I updated the pr after unrecognized fields support was added in securesystemslib version 0.22.0
The changes include:
1. I added `Metadata.unrecognized_fields` inside `Metadata._eq_`
2. I changed the way we check the dictionaries order of `Metadata.signatures` and `Delegations.roles`
3. I changed how we actually do validation inside `JSONSerializer.serialize()` by using `json_bytes` and calling `JSONDeserialize.deserialize()` to create a new metadata object for comparison. 

I think I addressed all comments by @jku and @lukpueh. It's ready for another round of reviews.

tuf/api/metadata.py
```diff
@@ -1331,6 +1433,25 @@ def __init__(
...
+            return False
+
+        # Check that the order of the roles is the same as dictionaries
+        # are insensitive to order in comparisons (see issue #1788).
+        if (
+            isinstance(self.roles, dict)
+            and isinstance(other.roles, dict)
+            and list(self.roles.keys()) != list(other.roles.keys())
+        ):
+            return False
```
jku: I assume the `list()` calls are here because dict_keys does not consider order in eq just like dict does not, so you need to construct a list? Annoying but makes sense.


I'm wondering about this: a few lines later we do `self.roles == other.roles` and that presumably does the first two isinstance checks above right? So if we move the list comparison after `self.roles == other.roles` then we would not need the first two checks or this if-clause at all...

I think with a good comment that would still be understandable and 5 fewer lines of code

MVrachev: When I call `dict.keys()` I don't receive a list:
image
that's why I have to cast it to a list that does take into account order.

> self.roles == other.roles and that presumably does the first two isinstance checks above right?

No, it doesn't replace the need for `isinstance` checks. 
`self.roles == other.roles` can return `True` when both `roles` attributes are other types like list, boolean, None...

I have thought about a few alternatives how can we replace the `isinstance`s and they are:
1. use `type(a) == type(b) == dict` check, but that is not `indomatic` and @lukpueh didn't like it: https://github.com/theupdateframework/python-tuf/pull/1775#discussion_r801435198
2. use `try` and `except` block where in the `except` we will catch `AttributeError`, but that's not beautiful either
3. use `all(isinstance(d, dict) for d in [self.signatures, other.signatures])`, but that's long again and `black` doesn't format it well
4. `isinstance(self.signatures, dict) and self.signatures == other.signatures`

None of the above points looked better than the current.
Maybe the last one is somehow okay, but it excludes the option of two equal `Metata` objects which has non-dict type signatures. That's probably good, but we don't enforce it inside `__init__`.

jku: > `self.roles == other.roles` can return `True` when both `roles` attributes are other types

that case doesn't sound like something eq() needs to care about though?

jku: Based on a chat it's clear that you are worried about the code raising an error later if the two roles are not dicts. I don't have a very strong opinion but it seems to me that eq() should be allowed to crash if the objects internal structure is invalid -- we can't be prepared for everything and should catch issues with roles type elsewhere _if_ we really don't want that to happen. eq() implementations should be as simple as possible

MVrachev: I kind of agree that we should allow it to crash.
What do you think @lukpueh?

lukpueh: Hm. Although I originally suggested checking the type, I agree with Jussi that `__eq__` shouldn't validate the internal structure of an object. As an extreme example we don't really want to avoid exceptions raised like this:
```python
>>> a = b = Delegations({}, {})
>>> del a.keys
>>> a == b 
...
AttributeError: 'Delegations' object has no attribute 'keys'
```

lukpueh: Here's a suggestion for a simplified `Delegations.__eq__`:

```python
def __eq__(self, other: Any) -> bool:
    if not isinstance(other, Delegations):
        return False

    return (
        self.keys == other.keys
        and list(self.roles.items()) == list(other.roles.items()) # order matters
        and self.unrecognized_fields == other.unrecognized_fields
    )
```

lukpueh: Discussion here is also relevant for `Metadata.__eq__`

tuf/api/serialization/json.py
```diff
@@ -44,17 +45,22 @@ def deserialize(self, raw_data: bytes) -> Metadata:
 class JSONSerializer(MetadataSerializer):
     """Provides Metadata to JSON serialize method.
 
-    Attributes:
+    Args:
         compact: A boolean indicating if the JSON bytes generated in
-                'serialize' should be compact by excluding whitespace.
+            'serialize' should be compact by excluding whitespace.
+        validate: Check that the metadata object can be deserialized again
+            without change of contents and thus find common mistakes.
```
lukpueh, line 52: I know you liked the part about finding common mistakes, but I must say I don't really know what that means. :)

tests/test_metadata_eq_.py
```diff
@@ -0,0 +1,412 @@
...
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.repo_dir = os.path.join(
+            utils.TESTS_DIR, "repository_data", "repository", "metadata"
+        )
+        cls.metadata = {}
+        for md in TOP_LEVEL_ROLE_NAMES:
+            with open(os.path.join(cls.repo_dir, f"{md}.json"), "rb") as f:
+                cls.metadata[md] = f.read()
```
lukpueh, line 47: You you don't seem to use the raw bytes other than feeding them to `Metadata.from_bytes` below. Any reason to not load the metadata right away?

tests/test_metadata_eq_.py
```diff
@@ -0,0 +1,412 @@
...
+        md = Metadata.from_bytes(self.metadata["snapshot"])
+
+        # Assert that md is not equal to an object from another type
+        self.assertNotEqual(md, "")
+
+        md_2 = copy.deepcopy(md)
+        self.assertEqual(md, md_2)
+
+        setattr(md_2, self.case_name, test_case_data)
+        self.assertNotEqual(md, md_2)
```
lukpueh: I see you use this pattern a lot in this module (everywhere where you use the `run_sub_tests_with_dataset` decorator):

```
for $DATA in $DATA_SET
  1. load container
  2. assert container not equal to empty string
  3. copy container and assert equal
  4. patch container copy with $DATA and assert unequal to original
```

I think this gives us good test coverage, but maybe you can restructure the pattern a bit? More specifically, 
steps 1-3 are don't really need to be repeated for every `$DATA` in `$DATA_SET`, when only step 4 varies. 

I think it should look something like this:
```
1. load container
2. assert container not equal to empty string
3. copy container and assert equal
4. for $DATA in $DATA_SET
    patch container copy with $DATA and assert unequal to original
```

WDYT?

tests/test_metadata_eq_.py
lukpueh: Oh, I see now, you modify the contents in some tests. Never mind my suggestion then.

tests/test_metadata_eq_.py
```diff
@@ -0,0 +1,412 @@
...
+
+        # Assert that both objects are not the same because of signatures order.
+        self.assertNotEqual(md, md_2)
+
+        # but if we fix the signatures order they will be equal
+        md_2.signatures = {}
+        for j in [0, 1, 2]:
+            md_2.signatures[str(j)] = signatures[j]
+
+        self.assertEqual(md, md_2)
```
lukpueh, line 80: A few coding style comments:
- You don't need real keys or signatures to test the order condition
- You only need to insert two signatures to test the order condition (not three)
- The Pythonic way of looping over [0,1,2] is to loop over `range(3)`, it also works the other way around `range(2, -1, -1)`

I think I'd write the whole test like this...
```python
md = Metadata.from_bytes(self.metadata["snapshot"])
md.signatures = {"a": Signature("a", "a"), "b": Signature("b", "b")}
md_2 = copy.deepcopy(md)

md_2.signatures = dict(reversed(md_2.signatures.items()))
self.assertFalse(md, md_2)

# You could re-revert and assert that they are equal again, but that doesn't seem necessary

```

tests/test_metadata_eq_.py
lukpueh: Same comment applies in other tests that use that decorator below.

tests/test_metadata_eq_.py
```diff
@@ -0,0 +1,412 @@
...
+
+    def test_md_eq_special_signatures_tests(self) -> None:
```
lukpueh, line 82: I think we can trust `securesystemslib` to properly implement and test `Signatures.__eq__`. So I suggest to remove this test.

tests/test_metadata_eq_.py
```diff
@@ -0,0 +1,412 @@
...
+
+        # Both objects are not the equal because of delegated roles order.
+        self.assertNotEqual(delegations, delegations_2)
+
+        # but if we fix the delegated roles order they will be equal
+        delegations_2.roles = {}
+        delegations_2.roles["a"] = copy.deepcopy(role_one)
+        delegations_2.roles["b"] = copy.deepcopy(role_two)
+
+        self.assertEqual(delegations, delegations_2)
```
lukpueh, line 288: I think you can simplify this test similarly to `test_md_eq_signatures_reversed_order` (see comment above)

lukpueh (changes requested):
Thanks for your efforts here, @MVrachev! It's getting there. I have a few more comments. This time I also took a closer look at the new test module.

tuf/api/metadata.py
MVrachev: I like how you placed the check for the order inside the final check in the return. 
I wanted to do something like it before but was hesitant because I wanted to have a comment, but short comment like this could do the work. :)

tuf/api/serialization/json.py
MVrachev: Common programming mistakes can be many things. Like using the wrong type for a field, using the wrong spec_version, using a wrong value (like a version below or equal to 0), or not knowing that each of the metadata roles should be listed inside `Root.roles`.
The list is long and I wanted to give a short reason why somebody will use that feature.

tests/test_metadata_eq_.py
MVrachev: Also, I think it's shorter if I load the data once in the begging of the class and then reuse it instead of using:
```python
            with open(os.path.join(cls.repo_dir, f"{<ROLE>}.json"), "rb") as f:
                cls.metadata[md] = f.read()
```
in multiple places.

lukpueh: Sure, although there is a `Metadata.from_file` method which is just as short. But yeah, it makes sense to read the files only once in setUpClass and I wasn't arguing against that.

tests/test_metadata_eq_.py
MVrachev: I agree that we are redoing steps 1-3 maany times and that it's probably to optimize them and I do agree with the second pseudocode and it should work.
There are two disadvantages to opting-out from the decorator: 
1. we don't receive an exact message which of the $DATA caused the effect as we do by using the decorator and a subtest
2. If one of the $DATA_SET cases fails then the other will do as well.

In a discussion with @lukpueh we remembered that the asserts provided `msg` option meaning that if an assert fails then the custom message will be appended. This will resolve **1** as a problem. For example, if we want to add `Failed case: {case}` to the message the result message of a failing equal looks like this: 
`<tuf.api.metadata.Metadata object at 0x7fdf6db05f00> != 'bbaba' : Failed case: signed`
meaning we still see a log of the given argument.

So, the only downside to using a for loop here is 2 which I think is okay given that if for one of the attributes the test fails most likely it will fail for the others as well.

tests/test_metadata_eq_.py
MVrachev: I like the simplification. It's neat how you can reverse a dict in one-liner.
I will only add that I think there is a sense in asserting that they are equal after we fix the signatures.

tests/test_metadata_eq_.py
MVrachev: I don't think I agree with this one.
Here we are testing the following cases:
1. Test that metadata objects with different signatures are not equal.
2. Test that metadata objects with empty signatures are equal
3. Metadata objects with different signatures types are not equal.

which doesn't test the same thing as https://github.com/secure-systems-lab/securesystemslib/blob/075043e46b1d017fc332cf44a2038558b3e246d8/tests/test_signer.py#L75,

MVrachev:
I rebased and made the following major changes:
1. simplified the tests a lot by following the @lukpueh suggestion and using a for loop instead of the decorator and adding a custom message to gather information when there is an error from which attribute is coming. 
2. simplified the tests related to order of signatures and delegated roles by using `reverse` (but because python3.7 didn't have reverse implemented for dictionaries I had to cast to a list first).
3. simplified the verification that objects with the same signatures, but in a different order are different. The same simplification was done for delegated role.

Thanks for the reviews, now it should be in a good shape I hope. :D

tests/test_metadata_eq_.py
lukpueh: We can still use the decorator. I think it makes the tests even cleaner (despite the repetitive first item in the data set rows). Here's an example how I aggregated the common parts of your `test_metadata_eq_`  and `test_signed_eq_`. You should be able to use these for most of your other test_*_eq functions as well. 

```python
   md_data: utils.DataSet = {
        "snapshot signed":       snapshot,           # test_metadata_eq_ (copy_and_simple_assert)
        "snapshot spec_version": snapshot["signed"], # test_signed_eq_ (copy_and_simple_assert)
        # add data for test_key_eq_, test_role_eq_, test_root_eq_, test_metafile_eq_, ...
        }

    @utils.run_sub_tests_with_dataset(md_data)
    def test_metadata__eq__(self, md: Any) -> None:
        self.assertNotEqual(md, "")
        md_2 = copy.deepcopy(md)
        self.assertEqual(obj, md)

    md_attrs_data: utils.DataSet = {
        "snapshot signed":       (snapshot, "signed", None),                     # test_metadata_eq_ (iteration 1)
        "snapshot signatures":   (snapshot, "signatures", None),                 # test_metadata_eq_ (iteration 2)
        "snapshot version":      (snapshot["signed"], "version", -1),            # test_signed_eq_ (iteration 1)
        "snapshot spec_version": (snapshot["signed"], "spec_version", "0.0.0"),  # test_signed_eq_ (iteration 2)
       # add data for test_key_eq_, test_role_eq_, test_root_eq_, test_metafile_eq_, ...
        }

    @utils.run_sub_tests_with_dataset(md_attrs_data)
    def test_metadata_attrs__eq__(self, test_case_data: Tuple) -> None:
        md, attr, val = test_case_data
        md_2 = copy.deepcopy(md)
        setattr(md_2, attr, val)
        self.assertNotEqual(md, md_2)
```

NOTE: I didn't include the initialization for the first argument (`md`), if you want to access what's in `cls.metadata`, you might need to define the test data sets in `setUpClass` as well.

MVrachev: Honestly, I think the way the tests are defined now I think is simpler than what you suggested.
We will have one big dataset with many cases instead of separating the attributes into logically separated datasets.

Additionally, the dataset won't be so beautiful when working `Key`, `Role,` `DelegatedRole`, `TargetFile` and `MetaFile` data. In order to reuse it we will have to define it inside `setupClass`, so the data can be reused.

The only real benefit I see is that the file will be smaller.

What do you think @jku?

tuf/api/metadata.py
```diff
@@ -1203,6 +1285,19 @@ def __init__(
...
+        if not isinstance(other, DelegatedRole):
+            return False
+
+        return (
+            super().__eq__(other)
+            and self.name == other.name
+            and self.terminating == other.terminating
+            and self.paths == other.paths
+            and self.path_hash_prefixes == other.path_hash_prefixes
+            and self.unrecognized_fields == other.unrecognized_fields
```
```suggestion
```
super() does this

jku (approved):
* it's a lot of new code but I think it's worth it and the eq() implementations are now simple
* I admit I am not able to evaluate if the tests are useful: `eq()` is really tricky. But I trust Martins judgement here: LGTM
* We only use this feature in one place in tests ATM, I wonder if we should do it more? 

Anyway, looks good to me, we can keep improving this after the fact -- e.g. this will be very nice to have for the static tests (#1806). Thanks for your work.

tests/test_metadata_eq_.py
jku: I honestly have problems figuring out if a specific test is even useful, eq() is such a special case. I don't have a strong opinion on the style

jku:
> We only use this feature in one place in tests ATM, I wonder if we should do it more?

If you have looked at the potential places (any `Metadata.to_files/to_bytes()` calls) already, and decided not to use validate please leave a comment so we know.

tests/test_metadata_eq_.py
lukpueh: Just wanted to say that your original idea of using the decorator was quite good, albeit with some restructuring. Apologies for causing any confusion. I'm really also fine either way. Thanks for your persistent efforts, @MVrachev!

lukpueh (approved):

MVrachev:
> * it's a lot of new code but I think it's worth it and the eq() implementations are now simple
> * I admit I am not able to evaluate if the tests are useful: `eq()` is really tricky. But I trust Martins judgement here: LGTM

When I was writing the tests I wanted to make sure we are:
1. calling each of the classes with its corresponding `__eq__` implementation
2. comparing an object to a different object (in our case with an empty string)
3. changing each of the attributes for each of the classes and then calling `__eq__`. 
That way we check each of the `==` comparisons in each of the `__eq__` functions.

While writing the tests I used coverall to check which lines in the `__eq__` implementations were tested.

> * We only use this feature in one place in tests ATM, I wonder if we should do it more?
> 
> Anyway, looks good to me, we can keep improving this after the fact -- e.g. this will be very nice to have for the static tests (#1806). Thanks for your work.

> If you have looked at the potential places (any Metadata.to_files/to_bytes() calls) already, and decided not to use validate please leave a comment so we know.

Well, we don't actually use validation in one test, but in multiple, because it's in the helper function `modify_metadata` inside `tests/test_trusted_metadata_set.py`. 
I didn't want to include this feature in multiple tests as I thought it can slow them down.
Let's keep it that way for now and if we decide we can always change that.

PR #1783, opened 2022-01-19
Drop python3.6 and remove the usage of OrderedDict

MVrachev:

Python version 3.6 was supported until December 23-rd 2021 meaning its
end of life has expired more than 20 days. See https://endoflife.date/python.
Dropping support for python version 3.6 will allow us to remove
OrderedDicts.

After a quick check I saw that Warehouse target python version 3.8.2:
- their docker file: https://github.com/pypa/warehouse/blob/main/Dockerfile#L47
- https://github.com/pypa/warehouse/blob/main/.python-version
- last pr updating pr version: https://github.com/pypa/warehouse/pull/7828
Pip supports python version 3.7+ as well. They dropped python 3.6 a couple of months ago:
https://github.com/pypa/pip/pull/10641

This means it shouldn't cause headache to our users if we drop python
version 3.6 too.

In all versions, python3.7+ dictionaries preserve the insertion order.
This means we can replace the usage of OrderedDict with a standard dictionary.
    
Something we have to keep in mind is that even though the insertion order is preserved
the equality comparison for normal dicts is insensitive for normal dicts compared to `OrderedDict`

For example:
```python
>>> OrderedDict([(1,1), (2,2)]) == OrderedDict([(2,2), (1,1)])
False
>>> dict([(1,1), (2,2)]) == dict([(2,2), (1,1)])
True
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       # Run regular TUF tests on each OS/Python combination, plus special tests
       # (sslib master) and linters on Linux/Python3.x only.
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-latest, macos-latest, windows-latest]
         toxenv: [py]
         include:
--- a/examples/repo_example/basic_repo.py
+++ b/examples/repo_example/basic_repo.py
@@ -25,7 +25,6 @@ NOTE: Metadata files will be written to a 'tmp*'-directory in CWD.
 """
 import os
 import tempfile
-from collections import OrderedDict
 from datetime import datetime, timedelta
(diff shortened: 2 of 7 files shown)


MVrachev:
Honestly, I am not sure why the CI started testing with python 3.6 when I removed it from ` .github/workflows/ci.yml`...

jku:
> not sure why the CI started testing with python 3.6 when I removed it from .github/workflows/ci.yml

It didn't: The UI does not make it very clear but current project settings define those tests as required for merge -- so this PR stays blocked because required tests haven't executed.

lukpueh (approved):

lukpueh:
> > not sure why the CI started testing with python 3.6 when I removed it from .github/workflows/ci.yml
> 
> It didn't: The UI does not make it very clear but current project settings define those tests as required for merge -- so this PR stays blocked because required tests haven't executed.

Seems like a good use case for my administrative super powers. 🦸

PR #1790, opened 2022-01-21
Rm all legacy

lukpueh:
Fixes #1745 

- Removes legacy code, tests and docs
- Updates remaining docs and some project metadata accordingly

**Preliminary work**
  - https://github.com/python/peps/pull/2240
  - https://github.com/theupdateframework/theupdateframework.io/pull/31
  - https://github.com/theupdateframework/python-tuf/pull/1773
  - https://github.com/theupdateframework/python-tuf/pull/1769
  - https://github.com/theupdateframework/python-tuf/pull/1768
  - https://github.com/theupdateframework/python-tuf/pull/1767
 - Updated references in [CII Badge App](https://bestpractices.coreinfrastructure.org/de/projects/1351)
 - Checked specification for stale references to python-tuf (found none)

**Note to reviewers**:
- Excluding deleted files from diff should make review easier
- I'd appreciate a quick look through the remaining files in the repo, if I missed anything.
- Check commit messages for details

**Note of appreciation**
Many thanks to everyone who helped python-tuf to evolve! ❤️ 

Most notably (based on prior [Acknowledgements](https://github.com/theupdateframework/python-tuf/blame/b2704a56a8c5023ffd7d377d8a8e72fb26a46a8a/README.md#L130-L132) and *git shortlog* and in alphabetical order):

*Sebastien Awwad, Justin Cappos, Geremy Condra, Vladimir Diaz, Zane Fisher, Pankhuri Goyal, Jussi Kukkonen, Trishank Kuppusamy, Joshua Lock, Marina Moore, Justin Samuel, Teodora Sechkova, Tian Tian, Santiago Torres-Arias, Martin Vrachev, Yuyu Zheng, Kairo de Araujo*

Issue #1745, opened 2021-12-22
Remove "legacy code"


lukpueh:
The recently released [v0.20.0](https://github.com/theupdateframework/python-tuf/releases/tag/v0.20.0) was the last release to include the legacy implementation. In [preparation for v1.0.0](https://github.com/theupdateframework/python-tuf/blob/v0.20.0/docs/1.0.0-ANNOUNCEMENT.md) it is time to actually say farewell to the legacy implementation. Here is a (likely non-exhaustive) list of related tasks:

- Identify and remove legacy code, tests and documentation
- Review and update continued resources as necessary
   - [README](https://github.com/theupdateframework/python-tuf/blob/develop/README.md) (e.g. 1.0.0 release and legacy deprecation announcements) 
   - some [python-tuf/docs/](https://github.com/theupdateframework/python-tuf/tree/develop/docs)
   - [theupdateframework.io](https://theupdateframework.io/)
   - [TUF specification](https://github.com/theupdateframework/specification)
   - 'passing', 'silver' and 'gold' criteria on [bestpractices.coreinfrastructure.org](https://bestpractices.coreinfrastructure.org/en/projects/1351)
   -  ...
- Consider renames
  - `ngclient` -> `client`
     *IMO the name `ngclient` makes most sense in relation to the "old" client and so will lose its meaning after the old client has gone, at least after a while* 
  - `api.metadata` -> `metadata`
     *IIRC the api directory was a bit of a crutch to develop new code alongside legacy code without mixing them up, which we won't need after removing the legacy code. IMO api should be designated by other means, e.g. on RTD*

jku:
> ngclient -> client

I prefer keeping `ngclient`: this way any old code `from tuf.client.updater import Updater` fails immediately which I think is a good thing now that the API is slightly different. I don't think `client` is significantly more discoverable either: users are going to read the API reference or example code anyway.

> api.metadata -> metadata

Fine by me (although I don't feel a strong need: imports look fine to me now). If we do this I would suggest we consider keeping the code where it is though (for git reasons) and solving the problem with clever imports.

MVrachev:
> * Consider renames
>   
>   * `ngclient` -> `client`
>     _IMO the name `ngclient` makes most sense in relation to the "old" client and so will lose its meaning after the old client has gone, at least after a while_
>   * `api.metadata` -> `metadata`
[...]

PR #1792, opened 2022-01-24
test: stop using unittest_toolbox in new tests

lukpueh:
Fixes parts of #1745

Update new test modules to stop using unittest_toolbox in preparation for its removal in #1790.

The tools provided by unittest_toolbox can easily (in a more obvious way) be replaced by using the standard library modules `tempfile` and `random` (no more used) directly.

In the case of tempdir and -file creation/removal, skipping the use of unittest_toolbox, which creates/removes a tmp dir for each test by default, also uncovers some test cleanup failures, which would occur when temporary test directories were removed while a test server hadn't released them.
(see `except OSError: pass` in unittest_toolbox's tearDown method)

**Change details**

**test_fetcher_ng.py:**
- Stop implicitly creating (setUp) and removing (tearDown) tmp test dirs.
- Move now manual creation of an exemplary targets file to setUpClass, as the same file is used by all tests. And remove it explicitly in tearDownClass after killing the server (see note about failure above).
- Trigger URL parsing error with a hardcoded invalid URL string instead of a random string.

**test_updater_ng.py**
- Stop implicitly creating (setUp) and removing (tearDown) tmp test dirs.
- Explicitly create tmp test dirs in setUp, but don't remove them in tearDown to avoid above mentioned failures. They will be removed all at once when removing the tmp root test dir in tearDownClass

--

unrelated change: define a TESTS_DIR constant for convenience

requirements**:

### DIFF
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -20,57 +20,48 @@ import requests
 import urllib3.exceptions
 
 from tests import utils
-from tuf import unittest_toolbox
 from tuf.api import exceptions
 from tuf.ngclient._internal.requests_fetcher import RequestsFetcher
 
 logger = logging.getLogger(__name__)
 
 
-class TestFetcher(unittest_toolbox.Modified_TestCase):
+class TestFetcher(unittest.TestCase):
     """Test RequestsFetcher class."""
 
     server_process_handler: ClassVar[utils.TestServerProcess]
 
     @classmethod
     def setUpClass(cls) -> None:
-        # Launch a SimpleHTTPServer (serves files in the current dir).
+        """
+        Create a temporary file and launch a simple server in the
+        current working directory.
+        """
         cls.server_process_handler = utils.TestServerProcess(log=logger)
+
+        cls.file_contents = b"junk data"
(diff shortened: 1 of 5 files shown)


jku (approved):
Looks like a good idea to me: some test case helpers might be a good idea but Modified_TestCase has not been a positive influence in our test code for some time...

tests/test_fetcher_ng.py
```diff
@@ -20,76 +20,65 @@
...
-        current_dir = os.getcwd()
-        target_filepath = self.make_temp_data_file(directory=current_dir)
-
-        with open(target_filepath, "r", encoding="utf8") as target_fileobj:
-            self.file_contents = target_fileobj.read()
-            self.file_length = len(self.file_contents)
+        cls.file_contents = b"junk data"
+        cls.file_length = len(cls.file_contents)
+        with tempfile.NamedTemporaryFile(
+            dir=os.getcwd(), delete=False
```
MVrachev, line 45: Didn't know you can close a `tempfile` and not delete it, cool.

MVrachev (approved):
I really like the changes!
They simplify the code a lot!
LGTM!

Issue #1764, opened 2022-01-11
Review all TODO items in the new implementation prior to 1.0.0


MVrachev:
**Description of issue or feature request**:
We have multiple `TODO` items in the source code inside the new implementation and it will be good to review them one by one and decide if we want to fix them now or leave them.
If we decide to leave them it will be good to create separate issues (where it makes sense) or group them in one summarization issue.

jku:
For Metadata API:
* 2 TODOs about `pylint: disable=too-many-arguments`. Let's remove the TODOs, in my opinion the lint is useful in most situations but not here so the disables are fine
* TODO about serialization error location: this could be an issue of its own, and relates to the other exception work going on. IMO the errors should be in the common exceptions file because metadata API users have to handle deserializationerror (it's not an internal error)

jku:
The ngclient TODOs should get handled by #1312, #1761 and #1714

jku:
Closing as #1794, #1805 are closed and  #1804 has been filed

PR #1794, opened 2022-01-25
Remove TODOs about too-many-arguments

jku:
The lint warning about argument count is useful in general but in these
two cases we want to break the rule: remove TODOs.

Improves #1764


MVrachev (approved):
Agree with you.
LGTM!

lukpueh (approved):

PR #1796, opened 2022-01-26
Metadata API: Accept X.Y spec_version

### LINKED ISSUE #1751: spec version incompatibility
datadogs metadata contains 
`"spec_version": "1.0",`

This is currently not compatible with Metadata API (deserializating the data fails). 

"1.0" seems to not be a strictly compliant spec_version (the specification says _format follows semantic versioning_) but we could potentially support this for backwards compat?

EDIT: also metadata generated with go-tuf (so for example sigstore metadata) looks like this.

jku:
All TUF implementations used to use "1.0" as the spec version and most
of them have never modified that value since.

Accept two-digit spec_version for legacy compatibility: it is strictly
speaking against the current spec (which requires semver) but there
should be no harm in doing this and it allows us to deserialize
metadata generated by e.g. go-tuf.

Fixes #1751

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -36,20 +36,20 @@ class TestSerialization(unittest.TestCase):
     # repository it's required that meta has at least one element inside it.
     invalid_signed: utils.DataSet = {
         "no _type": '{"spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "no spec_version": '{"_type": "signed", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "no version": '{"_type": "signed", "spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "no expires": '{"_type": "signed", "spec_version": "1.0.0", "version": 1, "meta": {}}',
+        "no spec_version": '{"_type": "snapshot", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no version": '{"_type": "snapshot", "spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no expires": '{"_type": "snapshot", "spec_version": "1.0.0", "version": 1, "meta": {}}',
         "empty str _type": '{"_type": "", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "empty str spec_version": '{"_type": "signed", "spec_version": "", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "empty str spec_version": '{"_type": "snapshot", "spec_version": "", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
         "_type wrong type": '{"_type": "foo", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "version wrong type": '{"_type": "signed", "spec_version": "1.0.0", "version": "a", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "invalid spec_version str": '{"_type": "signed", "spec_version": "abc", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "non-number spec_version": '{"_type": "signed", "spec_version": "1.2.a", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "one part spec_version": '{"_type": "signed", "spec_version": "1", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -431,16 +431,17 @@ def __init__(
...
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
+        # Accept semver (X.Y.Z) but also X.Y for legacy compatibility
         spec_list = spec_version.split(".")
-        if (
-            len(spec_list) != 3
-            or not all(el.isdigit() for el in spec_list)
-            or spec_list[0] != SPECIFICATION_VERSION[0]
+        if len(spec_list) not in [2, 3] or not all(
+            el.isdigit() for el in spec_list
```
jku, line 437: this is a weird construct but black wants to lay it out this way -- I'll take suggestions to make it clearer though

jku:
coverage claims none of the valueerrors are tested. I'll have to investigate that -- _EDIT: done, bug fix commit added._

joshuagl (commented):
The specification does suggest semver, but also leaves it up to clients to decide what is considered a match.

This seems fine to me.

tuf/api/metadata.py
MVrachev: I don't have a better suggestion as well, but it's not so bad.
black splits `all` brackets so that we can see clearly what are the arguments which are the complicated part.

tests/test_metadata_serialization.py
```diff
@@ -254,6 +254,8 @@ def test_invalid_timestamp_serialization(self, test_case_data: str) -> None:
     valid_timestamps: utils.DataSet = {
         "all": '{ "_type": "timestamp", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
             "meta": {"snapshot.json": {"hashes": {"sha256" : "abc"}, "version": 1}}}',
+        "two part spec_version": '{ "_type": "timestamp", "spec_version": "1.9000", "version": 1, "expires": "2030-01-01T00:00:00Z", \
```
MVrachev: Nitpick: Maybe better use `1.0` as described in the issue as problematic?
I am not sure there is a point in making spec_version = `1.9000` with so many zeroes helps.

tuf/api/metadata.py
```diff
@@ -431,16 +431,17 @@ def __init__(
...
         ):
-            raise ValueError(
-                f"Unsupported spec_version, got {spec_list}, "
-                f"supported {'.'.join(SPECIFICATION_VERSION)}"
-            )
+            raise ValueError(f"Failed to parse spec_version {spec_version}")
+
+        # major version must match
+        if spec_list[0] != SPECIFICATION_VERSION[0]:
+            raise ValueError(f"Unsupported spec_version {spec_version}")
```
MVrachev, line 443: So, finally, we don't want to log if there are differences in minor versions?

tests/test_metadata_serialization.py
```diff
@@ -36,20 +36,20 @@ class TestSerialization(unittest.TestCase):
     # repository it's required that meta has at least one element inside it.
     invalid_signed: utils.DataSet = {
         "no _type": '{"spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "no spec_version": '{"_type": "signed", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "no version": '{"_type": "signed", "spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
-        "no expires": '{"_type": "signed", "spec_version": "1.0.0", "version": 1, "meta": {}}',
+        "no spec_version": '{"_type": "snapshot", "version": 1, "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no version": '{"_type": "snapshot", "spec_version": "1.0.0", "expires": "2030-01-01T00:00:00Z", "meta": {}}',
+        "no expires": '{"_type": "snapshot", "spec_version": "1.0.0", "version": 1, "meta": {}}',
```
MVrachev, line 41: Good catch. The data can even be valid, but because of the `_type`, it will fail...

MVrachev (approved):
Overall, I agree with the changes even though it would have been better if the clients and implementations all stick to using all of the numbers in `semver`.

tests/test_metadata_serialization.py
jku: yeah -- one of those cases where parsing exception messages might have helped... but I'm still not willing to do that amount of work

tuf/api/metadata.py
jku: could do a debug log I suppose... but realistically a large portion of metadata that is deserialized is likely to not match exactly (think updated clients loading older targets metadata that was created with lower spec_version).

tests/test_metadata_serialization.py
jku: I suppose you're right

jku:
force pushed with one change:
* the new valid test case uses the common legacy value "1.0" instead of just some two part value

MVrachev (approved):

lukpueh (approved):
Patch looks good to me. Also kudos for catching the test bug!

PR #1799, opened 2022-01-26
ngclient: simplify storing a downloaded file

### LINKED ISSUE #1761: ngclient: throws securesystemslib errors
If ngclient fails to write a target file to disk, we let securesystemslib.exceptions.StorageError propagate.

This seems like a bug: we should either handle this error, or possibly not use the `securesystemslib.util.persist_temp_file()` at all (it's not super useful to us as we're not using the storage abstraction and the rest is a few lines of code that actually hides the error we'd really like to get: OSError which we already use for metadata files in the same situation).

MVrachev:
Fixes #1761

Replace the usage of `securesystemslib.util.persist_temp_file()` with
`shutil.copyfileobj()` as file system abstraction is not used in the
client.
This way we prevent `securesystemslib.exception.StorageError` from
leaking through client API calls.

Note: with those changes, we are no longer do fsync.

requirements**:

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -35,11 +35,10 @@ downloads target files is available in `examples/client_example
 
 import logging
 import os
+import shutil
 import tempfile
 from typing import Optional, Set
 from urllib import parse
-
-from securesystemslib import util as sslib_util
 
 from tuf.api import exceptions
 from tuf.api.metadata import (
@@ -218,8 +217,7 @@ class Updater:
             ValueError: Invalid arguments
             DownloadError: Download of the target file failed in some way
             RepositoryError: Downloaded target failed to be verified in some way
-            exceptions.StorageError: Downloaded target could not be written
-                to disk
+            OSError: Failed to write target to file
 
         Returns:
             Local path to downloaded file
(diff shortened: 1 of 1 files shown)


jku (approved):
This looks correct to me.

To make it more obvious you could mention in the commit message that this means we no longer do fsync. It believe that's fine in our use case but could be worth mentioning.

MVrachev:
@jku I updated pr, so it changes the exceptions docstring inside `updater.download_target()`.

Additionally, I made it dependent on https://github.com/theupdateframework/python-tuf/pull/1787 to make sure we end up with the correct exception docstring for `updater.download_target()`.

tuf/ngclient/updater.py
```diff
@@ -218,8 +217,7 @@ def download_target(
             ValueError: Invalid arguments
             DownloadError: Download of the target file failed in some way
             RepositoryError: Downloaded target failed to be verified in some way
-            exceptions.StorageError: Downloaded target could not be written
-                to disk
+            OSError: Cannot open "filepath" for writing.
```
jku: strictly speaking there are quite a few reasons for OSError here, not just failing to open file (like not enough space as an example).

jku (commented):
I think this is correct, left a comment about the docstring

tuf/ngclient/updater.py
MVrachev: Updated the message.
Do you think that now is better?

jku: I'm not sure there was anything wrong with the original message? Alternatively "Failed to write target to file" could work.

MVrachev: I understood you had some reservations about it.
I updated the pr with your suggestion.

jku (approved):
Looks good to me, thanks.

If another maintainer wants to chime in on the issue of not using `persist_temp_file()` (and as a result, not fsyncing) that would be welcome: I've tried to think the failure scenarios through and this seems like the simplest code that does what we want.

lukpueh (approved):

PR #1800, opened 2022-01-26
Metadata API: Document serialization "repro" issue

### LINKED ISSUE #1392: Document Metadata API: MetaFile hash matching issue with serialization
So far I've been assuming we can build the client and use the Metadata API to
* deserialize what we get from the remote repo, and then
* serialize the data to disk once we know it's valid

The issue here is "meta" in timestamp (and I suppose snapshot) can contain file hashes. If we only store on local disk what has been deserialized and then serialized these hashes will fail to match when they are next read from disk (because white space etc).

Fixing this in the client is easy but I wonder if there's something that needs to be done in metadata API to make this more obvious or a harder mistake to make.

jku:
It's not obvious to casual reader that reading metadata and then
writing it might not always produce the same file. It's also not
immediately obvious why this matters.

Document both concepts.

Fixes #1392

---

It's a little wordy but I think it may be useful.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -232,6 +232,13 @@ class Metadata(Generic[T]):
     ) -> bytes:
         """Return the serialized TUF file format as bytes.
 
+        Note that if bytes are first deserialized into ``Metadata`` and then
+        serialized with ``to_bytes()``, the two are not required to be
+        identical even though the signatures are guaranteed to stay valid. If
+        byte-for-byte equivalence is required (which is the case when content
+        hashes are used in other metadata), the original content should be used
+        instead of re-serializing.
+
         Arguments:
             serializer: A MetadataSerializer instance that implements the
                 desired serialization format. Default is JSONSerializer.
@@ -264,6 +271,13 @@ class Metadata(Generic[T]):
         storage_backend: Optional[StorageBackendInterface] = None,
     ) -> None:
         """Writes TUF metadata to file storage.
+
+        Note that if a file is first deserialized into ``Metadata`` and then
+        serialized with ``to_file()``, the two files are not required to be
+        identical even though the signatures are guaranteed to stay valid. If
+        byte-for-byte equivalence is required (which is the case when file
+        hashes are used in other metadata), the original file should be used
+        instead of re-serializing.
 
         Arguments:
             filename: The path to write the file to.


MVrachev (approved):
The comments look good.

I wonder do we want to add a comment somewhere that even though when deserializing and serializing we don't get byte-for-byte equivalence we do get a consistent signature no matter the whitespaces?

jku:
I can see that making the distinction could help... I'll try that (even if the paragraph is a bit long already)

jku:
force pushed: added the mention that signatures are guaranteed to stay valid in a deserialize-serialize cycle even if the file content itself is not guaranteed to stay the same.

jku:
force pushed: 
* rebased to get recent build fix
* added the mention that signatures are guaranteed to stay valid in a deserialize-serialize cycle even if the file content itself is not guaranteed to stay the same.

tuf/api/metadata.py
```diff
@@ -232,6 +232,13 @@ def to_bytes(
     ) -> bytes:
         """Return the serialized TUF file format as bytes.
 
+        Note that if bytes are first deserialized into ``Metadata`` and then
+        serialized with ``to_bytes()``, the two are not guaranteed to be the
+        same because of whitespace issues, even if the signatures are
+        guaranteed to stay valid. If byte-for-byte equivalence is required
```
MVrachev: A smart way to mention that signatures are guaranteed to stay valid.

MVrachev (approved):
LGTM!

joshuagl (approved):

lukpueh (commented):
I like the clarification, modulo two nits:

- the claim *"not guaranteed to be the same"*  depends on the serializer, which is configurable for both `to_file` and `to_bytes`. Maybe it's worth mentioning that?
- the reason *"because of whitespace issues"* is also specific to the default json serializer, and also only part of the problem (e.g. non-guaranteed dict key order is another problem). Maybe it's not necessary to mention a particular reason?

jku:
Agreed on both comments, thanks, will update.

jku:
I solved this by
* not mentioning whitespace as a reason
* just saying that serialization is not required to be byte-for-byte the same (the implication is that different serializers may make different decision, but that's not said)

lukpueh (approved):
Thanks for the update, @jku!

Issue #1793, opened 2022-01-24
Yearly maintainer list cleanup


jku:
Teodora stepping down as maintainer got me thinking that we should clean up the maintainers list periodically:
* I think both [docs/MAINTAINERS.txt](https://github.com/theupdateframework/python-tuf/blob/develop/docs/MAINTAINERS.txt) and especially the actual permissions lists (so github project, pypi, etc) should only list the maintainers who are actively using the elevated powers (so approving and merging PRs, creating releases, configuring services).
* this is basic security posture: anyones account can be compromised, so fewer elevated permissions leads to fewer project compromises
* all services should have > 2 active maintainers to minimize bus factor (and prevent project being locked out from critical services by accident)

Maybe a clearly separate "Emeritus maintainers" section in docs/MAINTAINERS.txt would be nice and would make moving psychologically easier? I would suggest a rough rule _if you've not used elevated permissions in the past year or don't plan to use them during the next year, you might be a Emeritus Maintainer_.

As an additional thing: it's not easy to check if the actual permissions are appropriate (as an example I don't think I can see who even has Github maintainer status) so maybe we should make a yearly issue to manually check them?

**2022-01-24 situation looks like this:**

* docs/MAINTAINERS.txt: @awwad, @mnm678,  @trishankatdatadog, @lukpueh, @SantiagoTorres, @joshuagl, @jku, @sechkova (+@JustinCappos as consensus builder)
* github: ???
* PyPI: @jku, @joshuagl, @JustinCappos, @lukpueh, @SantiagoTorres 
* ReadTheDocs: @jku, @sechkova, @joshuagl
* coveralls: ???

Teodora has already said she's moving to other things but how about others? E.g. @awwad, @SantiagoTorres any objections to becoming emeritus maintainers?

@lukpueh would you mind filling in the question marks on the list? Also would you like to make an account at readthedocs.org and let me know the account name so I can add you as maintainer?

trishankatdatadog:
> * this is basic security posture: anyones account can be compromised, so fewer elevated permissions leads to fewer project compromises

Agreed, I've already raised this issue before, but got radio silence.

JustinCappos:
I'm supportive of this.  Would like to see what @awwad and @SantiagoTorres
<https://github.com/SantiagoTorres> think.

Justin

SantiagoTorres:
[...]

PR #1803, opened 2022-01-27
CI: Add yearly reminder issue to review maintainers

jku:
This is easy to forget as 
 * there are multiple different critical services
 * some permissions are not visible to everyone

but review is important as every maintainer account increases attack surface.
So let's remind ourselves once a year.

This relates to #1793.

### DIFF
--- a/.github/workflows/maintainer-permissions-reminder.yml
+++ b/.github/workflows/maintainer-permissions-reminder.yml
@@ -1,0 +1,52 @@
+name: Maintainer review reminder
+
+on:
+  schedule:
+    - cron: '10 10 10 2 *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+jobs:
+  file-reminder-issue:
+    name: File issue to review maintainer permissions
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/github-script@v5
+      with:
+        script: |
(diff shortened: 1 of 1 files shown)


jku:
This is what it looks like https://github.com/jku/python-tuf/issues/36

mnm678 (approved):

jku:
Well, it seems passable even with fresh eyes... We could add more specific rules (like _if maintainer has not exercised elevated permissions in a year, permissions should be revoked_) but that seems unnecessary: situations change and the problem is hard to quantify. I think the the goal is just to make yearly review/discussion a habit for the project.

Maybe I should also explain the weird layout of the issue body: this is the only way I found to both make the generated issue look good and avoid the multiline js string style that requires concatenating individual strings for each line.

joshuagl (approved):
This LGTM, but I'd like @lukpueh to review also (particularly the minimum number of maintainers per service) as he's been looking at this space recently.

joshuagl:
Thank you for working on this @jku!

lukpueh (approved):
Great work, @jku! Comparing it with the issue-opening GHA cronjob you wrote for sslib this looks good. I also agree with the contents here.

jku:
:+1: we should get the first reminder next week so we can a) confirm it's working and b) document the current maintainership status

PR #1808, opened 2022-01-28
Add test coverage for delegated hash bins

### LINKED ISSUE #1639: RepositorySimulator: add support for delegated hash bins

Currently `repository_tool.py` provides an easy way to distribute a large number of target files over multiple delegated roles (hashed bins).  A feature described in PEP458 that would be used in warehouse.  https://github.com/theupdateframework/python-tuf/blob/2206fc917e46ff584dcae75e640e1144c8b9f078/tuf/repository_tool.py#L2494

Having such functionality in `RepositorySimulator` that uses `metadata api` will be useful for testing the newer code.

ivanayov:
This change adds tests coverage for `path_hash_prefixes` and
verifies that role names matching specific prefixed successfully
find and download the corresponding metadata files

Fixes #1639

### DIFF
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -32,7 +32,7 @@ class TestDelegation:
     keyids: List[str] = field(default_factory=list)
     threshold: int = 1
     terminating: bool = False
-    paths: List[str] = field(default_factory=lambda: ["*"])
+    paths: Optional[List[str]] = field(default_factory=lambda: ["*"])
     path_hash_prefixes: Optional[List[str]] = None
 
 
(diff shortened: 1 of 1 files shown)


jku (changes requested):
marking request changes so it's visible on PR list: we talked about this over zoom, and found two issues: ivana promised a refactor at some point

jku:
There are still some problems here:
1. roles_pref_map still seems unnecessary: `_init_repo()` call already adds the delegations defined in your test data so there seems to be no reason to then add them manually? If the data driven test setup (defining a Dataset outside the test) does not work in this case for some reason, then let's not use it: defining a Dataset outside of the test and then also generating the test data in the test code combines the worst of both worlds.
2. test comments say we _verify that no metadata file was downloaded as expected for non-existing path_, however: 
    * we do expect delegated metadata to be downloaded for non-existing targetpaths if and only if the prefix of the hash of the targetpath matches the one in the delegation: this what the test is trying to test
    * test does not verify whether metadata was downloaded or not: it should
3. The tests in TestDelegationsGraphs are so far concentrated on testing delegations only: TargetFiles are not needed. I think this is a good idea for your test as well as it makes the test data smaller and focuses the test better. Yes, we should add target download tests for hash prefixes case but it does not need to happen in this class and it does not need to happen in this PR.

The test should assert the same things `test_graph_traversal()` does for the "paths" case: assert that  after `get_targetinfo()`
correct metadata files are in cache and that the expected repository requests were made (see how visited_order is used to check both of those).

The complication with this is that while DelegationsTestCase defines `visited_order` as the expected delegated targets that are downloaded, it does not define the targetpath that is used: the targetpath is currently hard coded in the test, usually as "missingpath". I'm totally fine doing the same in this test for the first version but I do expect you to assert that visited_order is correct: currently it is not used at all.

---

_This is outside of scope for this PR but in case it makes the intent clearer: What DelegationsTestCase maybe should do is define a dict of target_paths and visited_orders:_
```
search_results: {
    "missingpath": ["A", "B", "C"]
    "noMatchInCPaths": ["A", "B"]
}
```
_in your path_hash_prefixes cases the search results would look like_
```
search_results: {
    "missingpath": ["someprefix"]
    "othermissingpath": ["otherprefix"]
}
```
_because with your test data you will only get a single delegated role for every search._

ivanayov:
Thank you for the review @jku ! I changed the test as you proposed and moved the old with target downloads to #1910 where your comments are addressed as well.

tests/test_updater_delegation_graphs.py
```diff
@@ -362,6 +362,96 @@ def test_safely_encoded_rolenames(self) -> None:
...
+
+            updater = self._init_updater()
+            # Call explicitly refresh to simplify the expected_calls list
+            updater.refresh()
+            self.sim.fetch_tracker.metadata.clear()
+            # Check that metadata dir contains only top-level roles
+            self._assert_files_exist(TOP_LEVEL_ROLE_NAMES)
+
+            # Looking for a non-existing targetpath forces updater
+            # to visit all possible delegated roles
```
jku: this is on the edge of misleading... yes, a lookup leads to visiting "all possible delegated roles" but with hash bins that always means _exactly one delegated role_.

tests/test_updater_delegation_graphs.py
```diff
@@ -362,6 +362,96 @@ def test_safely_encoded_rolenames(self) -> None:
...
+                    path_hash_prefixes=[
+                        "b8",
+                        "b9",
+                        "ba",
+                        "bb",
+                        "bc",
+                        "bd",
+                        "be",
+                        "bf",
+                    ],
```
jku: I suppose one way to make this shorter is to use e.g.
```suggestion
                    path_hash_prefixes=["8", "9", "a", "b"],
```
This way you could have four delegations (0-3, 4-7, 8-b, c-f) that actually cover all possible path hashes

jku (approved):
Thank you, this looks much better. I have two issues left but I'm starting to think they may be for follow up issues:
* The purpose of TestDelegation was to make the test case definition short. These are now ridiculously long, 14 lines per delegation or 50 lines for a single test case.
* The test does not actually check that target lookups lead to the correct metadata file downloads, it only checks that after three lookups, all three are downloaded (so we could e.g. have a bug where we download all three on the first lookup, and would not notice that)

I think we could leave those for later PRs... but I did make a suggestion on how to make the TestDelegation shorter in lines (it could also cover all hashes and not just a small subset). Let me know if you'd like to still handle that here or not.

tests/test_updater_delegation_graphs.py
ivanayov: +1 for this. Those linter requirements are sometimes really hard for reading.

jku (changes requested):
Oh yeah, much more readable. Thanks

Please fix the delegation role names, they don't currently make any sense (this does not affect the test functionality but I think the current names mislead the reader). This should trigger the the CI as well

jku (approved):

Issue #1802, opened 2022-01-27
Should we allow unrecognized fields for the signatures dictionaries?


MVrachev:
**Description of issue or feature request**:
In [ADR 8](https://github.com/theupdateframework/python-tuf/blob/develop/docs/adr/0008-accept-unrecognised-fields.md) we follow the [Document formats](https://theupdateframework.github.io/specification/latest/#document-formats) section of the specification and allow unrecognized fields at each level of the metadata files.
The question is do we want to allow unrecognized fields inside the `signatures` dictionaries? 
Are there possible security issues to that?

For context see the specification issue on the matter: https://github.com/theupdateframework/specification/issues/203

jku:
Lets decide this before 1.0 release (with the understanding that we might just decide to postpone)

jku:
Quoting myself in sslib issue: 

> It is true that we should be more cautious with this part of the metadata (not because they contain signatures but because they are not part of the signed portion of metadata). But: because of TUF design that requires canonicalization we end up parsing even the signed portion before we can check that it has been signed.

Because of this _I think_ allowing unrecognised fields in signature does not significantly increase risk


Oh and the reason why we might want to do this: I believe supporting fulcio certificates/signatures requires extending the metadata. If unrecognised fields are not supported that would mean python-tuf 1.0 might not be able to read metadata written by python-tuf 1.x with fulcio support

mnm678:
cc @asraa who may have insight into the Fulcio signatures.

When using canonical json, it's true that adding an additional field doesn't increase the risk, because we are already doing some somewhat risky parsing. If we switch to a format like DSSE, we'll want to avoid adding additional fields to the signature that could make the implementation vulnerable to a [canonicalization attack](https://soatok.blog/2021/07/30/canonicalization-attacks-against-macs-and-signatures/).

jku:
To complete the options here: If we allow unrecognised_fields in Signature then there's nothing preventing allowing them in Metadata itself as well.

asraa:
> cc @asraa who may have insight into the Fulcio signatures.

It was majorly convenient adding a cert field to the signature -- +1

MVrachev:
Together in a discussion with @jku and @lukpueh we agreed that it makes sense to implement unrecognized fields inside `Metadata`, `Metadata.signatures` and in the `Signatures` objects.
You can see read more about my reasoning here: https://github.com/theupdateframework/specification/issues/203#issuecomment-1031479746.

[...]

PR #1809, opened 2022-01-28
Add Metadata.signatures serialization tests

MVrachev:

In a conversation with @jku he pointed out we don't have serialization
tests for `Metadata.signatures`.
That's how we decided it's worth adding such tests inside
`test_metadata_serialization.py` and moving the
test case about duplicating signatures inside the same test file.

requirements**:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ packages = find:
 python_requires = ~=3.7
 install_requires =
   requests>=2.19.1
-  securesystemslib>=0.20.0
+  securesystemslib>=0.22.0
 
 [options.packages.find]
 exclude = tests
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -42,11 +42,16 @@ class TestSerialization(unittest.TestCase):
             { "_type": "timestamp", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
             "meta": {"snapshot.json": {"hashes": {"sha256" : "abc"}, "version": 1}}} \
         }',
+        "non unique duplicating signatures": b'{"signed": \
+                { "_type": "timestamp", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
+                "meta": {"snapshot.json": {"hashes": {"sha256" : "abc"}, "version": 1}}}, \
+            "signatures": [{"keyid": "id", "sig": "b"}, {"keyid": "id", "sig": "b"}] \
+        }',
     }
 
     @utils.run_sub_tests_with_dataset(invalid_metadata)
     def test_invalid_metadata_serialization(self, test_data: bytes) -> None:
-        # We expect a DeserializationError reraised from KeyError.
+        # We expect a DeserializationError reraised from ValueError or KeyError.
         with self.assertRaises(DeserializationError):
             Metadata.from_bytes(test_data)
 
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -122,7 +122,7 @@ class TestSerialization(unittest.TestCase):
     def test_invalid_signed_serialization(self, test_case_data: str) -> None:
         case_dict = json.loads(test_case_data)
         with self.assertRaises((KeyError, ValueError, TypeError)):
-            Snapshot.from_dict(copy.deepcopy(case_dict))
+            Snapshot.from_dict(case_dict)
 
     valid_keys: utils.DataSet = {
         "all": '{"keytype": "rsa", "scheme": "rsassa-pss-sha256", \
@@ -155,7 +155,7 @@ class TestSerialization(unittest.TestCase):
         case_dict = json.loads(test_case_data)
         with self.assertRaises((TypeError, KeyError)):
             keyid = case_dict.pop("keyid")
-            Key.from_dict(keyid, copy.copy(case_dict))
+            Key.from_dict(keyid, case_dict)
 
     invalid_roles: utils.DataSet = {
         "no threshold": '{"keyids": ["keyid"]}',
@@ -169,7 +169,7 @@ class TestSerialization(unittest.TestCase):
     def test_invalid_role_serialization(self, test_case_data: str) -> None:
         case_dict = json.loads(test_case_data)
         with self.assertRaises((KeyError, TypeError, ValueError)):
-            Role.from_dict(copy.deepcopy(case_dict))
+            Role.from_dict(case_dict)
 
     valid_roles: utils.DataSet = {
         "all": '{"keyids": ["keyid"], "threshold": 3}',
@@ -276,7 +276,7 @@ class TestSerialization(unittest.TestCase):
     def test_invalid_root_serialization(self, test_case_data: str) -> None:
         case_dict = json.loads(test_case_data)
         with self.assertRaises(ValueError):
-            Root.from_dict(copy.deepcopy(case_dict))
+            Root.from_dict(case_dict)
 
     invalid_metafiles: utils.DataSet = {
         "wrong length type": '{"version": 1, "length": "a", "hashes": {"sha256" : "abc"}}',
@@ -292,7 +292,7 @@ class TestSerialization(unittest.TestCase):
     def test_invalid_metafile_serialization(self, test_case_data: str) -> None:
         case_dict = json.loads(test_case_data)
         with self.assertRaises((TypeError, ValueError, AttributeError)):
-            MetaFile.from_dict(copy.deepcopy(case_dict))
+            MetaFile.from_dict(case_dict)
 
     valid_metafiles: utils.DataSet = {
         "all": '{"hashes": {"sha256" : "abc"}, "length": 12, "version": 1}',
@@ -316,7 +316,7 @@ class TestSerialization(unittest.TestCase):
     def test_invalid_timestamp_serialization(self, test_case_data: str) -> None:
         case_dict = json.loads(test_case_data)
         with self.assertRaises((ValueError, KeyError)):
-            Timestamp.from_dict(copy.deepcopy(case_dict))
+            Timestamp.from_dict(case_dict)
 
     valid_timestamps: utils.DataSet = {
         "all": '{ "_type": "timestamp", "spec_version": "1.0.0", "version": 1, "expires": "2030-01-01T00:00:00Z", \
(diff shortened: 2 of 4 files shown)


tests/test_metadata_serialization.py
```diff
@@ -29,9 +30,81 @@
...
+
+    @utils.run_sub_tests_with_dataset(invalid_signatures)
+    def test_invalid_signatures_serialization(self, test_data: str) -> None:
+        case_dict = json.loads(test_data)
+        # pylint: disable=no-member
+        if "unrecognized_fields" in self.case_name:
+            # When unrecognized fields exist the object creation will be
+            # successful, but the unrecognized fields won't be stored.
+            md = Metadata.from_dict(copy.deepcopy(case_dict))
+            self.assertNotEqual(case_dict, md.to_dict())
```
jku: unrecognised fields case is now listed as an invalid signature case, yet you are showing here how it passes deserialization: Metadata object is valid, and all the signatures in it are valid ...

If we allow this (successful deserialization where unknown data is dropped) to happen , it should not be part of invalid cases. 
* I think we can allow this to happen for now and decide in #1802 whether to handle unrecognised fields or to make this an error -- with the understanding that currently a deserialize-serialize cycle will destroy unrecognised_fields in both Signature or Metadata and that long term this is undesirable.
* a separate test might make sense, this is a very special case

tests/test_metadata_serialization.py
```diff
@@ -29,9 +30,81 @@
 logger = logging.getLogger(__name__)
 
 
+# pylint: disable=too-many-public-methods
 class TestSerialization(unittest.TestCase):
     """Test serialization for all classes in 'tuf/api/metadata.py'."""
 
+    invalid_signatures: utils.DataSet = {
```
jku: this test case definition is huge as it's not actually data set for invalid Signature, but for invalid Metadata. I get that you are doing this because you wanted to include the duplicate signatures case... but it's still test data for Metadata not Signature: at a minimum the name is wrong. 

In detail, three test cases out of five seem to be tests for Signature (would not need a Metadata), and two are tests for Metadata.  If you want to include the latter two, I think you should separate this into two separate test data sets: invalid_signatures and invalid_metadatas. It should be more readable as the Signature test data is vastly smaller.

tests/test_metadata_serialization.py
```diff
@@ -29,9 +30,81 @@
...
+        if "unrecognized_fields" in self.case_name:
+            # When unrecognized fields exist the object creation will be
+            # successful, but the unrecognized fields won't be stored.
+            md = Metadata.from_dict(copy.deepcopy(case_dict))
+            self.assertNotEqual(case_dict, md.to_dict())
+        else:
+            with self.assertRaises((KeyError, ValueError)):
+                md = Metadata.from_dict(copy.deepcopy(case_dict))
+
+    valid_signatures: utils.DataSet = {
```
jku, line 90: this is also really valid_metadatas. I think valid_signatures (with actual signatures in it) might be more useful

I don't object to also add valid_metadatas but the possibility space explodes there... we  won't be able to test many meaningful combinations. but cases like "multiple signatures", "no signatures" are still doable

jku (changes requested):
"signature" and "metadata" are used in test names in a confusing way if you compare to all existing tests... details in code

MVrachev:
@jku I addressed all of your comments:
- I removed the unrecognized_fields test for signatures as it's already tested inside `securesystemslib`: https://github.com/secure-systems-lab/securesystemslib/pull/387
- I split the signature and metadata datasets
- added real signatures inside the valid metadata set.

tests/test_metadata_serialization.py
```diff
@@ -29,9 +32,87 @@
...
+                } \
+              ] \
+        }',
+    }
+
+    @utils.run_sub_tests_with_dataset(invalid_metadata)
+    def test_invalid_metadata_serialization(self, test_case_data: str) -> None:
+        case_dict = json.loads(test_case_data)
+        with self.assertRaises((ValueError, KeyError)):
+            Metadata.from_dict(case_dict)
```
jku: What do you think about using the full serialization path for Metadata -- `Metadata.from_bytes() / Metadata.to_bytes()`?

It would be a bit different from other tests in this file -- but feels like a meaningful extension: if we could do a full serialization for other objects too, we would: it's just not possible for those cases.

This applies to the valid test case as well.

tests/test_metadata_serialization.py
```diff
@@ -29,9 +32,87 @@
...
+            "signatures": [ \
+                { \
+                    "keyid": "59a4df8af818e9ed7abe0764c0b47b4240952aa0d179b5b78346c470ac30278d", \
+                    "sig": "5b00100e9cf1c083f8347371ab840cf60124780305124ed7a53fe31bf43473c90b1d2c802ed2f11f5057ba21e6b7a05118b1907f737d2e29c9692aa3345f9801" \
+                }, \
+                { \
+                    "keyid": "8a1c4a3ac2d515dec982ba9910c5fd79b91ae57f625b9cff25d06bf0a61c1758", \
+                    "sig": "de0e16920f87bf5500cc65736488ac17e09788cce808f6a4e85eb9e4e478a312b4c1a2d7723af56f7bfb1df533c67d8c93b6f49d39eabe7fae391a08e1f72f01" \
+                } \
+            ] \
```
jku: opinions on making test data smaller in this way?

```suggestion
            "signatures": [ { "keyid": "a", "sig": "b" }, { "keyid": "c", "sig": "d" } ] \
```

I think the massive test data hides what the test does -- it looks like the keyids and signatures are important for the test when they are not. From that perspective the test data should be minimal to test the case that we want to test.

jku (commented):
You could add a "valid signatures" test set as well (with at least a default case and a case with unrecognised fields).  **EDIT:** oh I see you mentioned that in your comment: I think it wouldn't hurt to have here as well but it's your call.

Left a couple of more suggestions in source, but all of these are probably negotiable...

tests/test_metadata_serialization.py
MVrachev: You said you want me to use real signatures?
> this is also really valid_metadatas. I think valid_signatures (with actual signatures in it) might be more useful

from https://github.com/theupdateframework/python-tuf/pull/1809#discussion_r797441847

jku: so I did... I think I may have been wrong then because A) the serialization does not test the signatures themselves -- just whether the signature object is valid and B) 11 lines for the signatures looks so unreadable

My bad, sorry. Either way works for me.

MVrachev: I prefer if they are shorter, so yea, I will update it.

jku: Wait, I take that back.

I said _"valid_signatures (with actual signatures in it) might be more useful"_. This still sounds reasonable: I probably would not have objected to that.

But your test case here is **valid_metadata**. Adding 22 lines of unused signature data into the Metadata test (but not doing that in the signature test) is not as useful as it hides what is actually tested

MVrachev: Well, you can have a look at my latest changes and tell me how you finally want it :D.
If it's still doesn't look good I will rever my changes and add the valid signatures inside the signatures test.

MVrachev:
@jku I followed your suggestions. 

I also added a new commit updating `securesystemslib` to 0.22.0 which contains support for unrecognized fields from `Signature`.
I added that commit so unrecognized fields can be added as a test case for `Signature` serialization.

tests/test_metadata_serialization.py
```diff
@@ -25,13 +28,76 @@
...
+                "meta": {"snapshot.json": {"hashes": {"sha256" : "abc"}, "version": 1}}}, \
+            "signatures": [] \
+        }',
+    }
+
+    @utils.run_sub_tests_with_dataset(valid_metadata)
+    def test_valid_metadata_serialization(self, test_case_data: bytes) -> None:
+        md = Metadata.from_bytes(test_case_data)
+        input_dict = json.loads(test_case_data)
+        self.assertDictEqual(input_dict, md.to_dict())
```
jku, line 77: This is better as it runs the whole deserialize-pipeline ... It still doesn't test the whole serialize-pipeline or compare the final result to original input.

I can guess why, you are balancing between keeping the input test data readable and being able to compare the input data to expected output of `Metadata.to_bytes()`.

Did you try if the byte array could be kept made equal to expected output (compact mode or not) and still somewhat readable?

setup.cfg
```diff
@@ -37,7 +37,7 @@ packages = find:
 python_requires = ~=3.7
 install_requires =
   requests>=2.19.1
-  securesystemslib>=0.20.0
+  securesystemslib>=0.22.0
```
jku, line 40: hmm, this is probably correct but let's make sure we are doing things for right reasons.

It's true that passing the updated test suite requires this... but this is not really an absolute runtime requirement: python-tuf will work with older securesystemslib just as well as it did before this PR. This PR only changes tests  so runtime requirements couldn't possibly have changed: it's a bit weird to imply they did.

On the other hand I don't see an issue with bumping sslib requirement: sslib is practically a part of python-tuf and the bump likely won't make life harder for anyone.

jku (commented):
I think this is good...

I left a comment on the Metadata serialization tests: like I said I do think there's value in trying to test the full deserialize-serialize cycle (currently the PR never serializes to json). 
* Doing that would mean the input data would have to be in one of the output styles we support so possibly not as readable (and would be again a little more different from the other serialization tests)
* possibly the right choice is to store the Metadata test data in actual static files in git -- admit that full metadata is too big to include in test files themselves 

Maybe we just add some details to the existing issue we have for static test data, and accept the metadata tests in this PR for the time being? **EDIT** I left some thoughts in https://github.com/theupdateframework/python-tuf/issues/1806

tests/test_metadata_serialization.py
jku: as mentioned in PR comment, I think we can leave this as is but just plan to add better full-cycle serialization tests in near future

jku (approved):
I'll approve but maybe @lukpueh has opinions

setup.cfg
MVrachev: I will remove that to escape controversy and give us an option to decide when to bump it.

MVrachev: PS: I can update the pr later today. If you still think it's okay to bump the securesystemslib version here then go ahead and merge it.

tests/test_metadata_serialization.py
MVrachev: I can update the pr later today. If you still think it's okay to compare the data that way then merge it and I can update it in another pr.

lukpueh (approved):
Thanks, LGTM!

I'm also fine with bumping sslib runtime requirement. I do see your point Jussi, but I think we can let it go, given that we will require the bump for an upcoming PR and before the next release.

Issue #1784, opened 2022-01-19
Generic exceptions vs Specific exceptions


MVrachev:
**Description of issue or feature request**:
We want to use a shortlist of exceptions that will be thrown from our APIs, so that our users can handle them easily.
On another hand that doesn't stop us from creating other more specific exceptions that derive from those main exceptions.
This allows us to easily test that specific errors were thrown when we expected them.
For example, we are using `BadVersionError` and `LengthOrHashMismatchError` which derive from `RepositoryError`.

Still, there are two use cases that are worth discussing:

1. There are multiple places inside `TrustedMetadataSet` where in one API call we are throwing generic and specific exceptions.
That's happening inside each of the functions `update*` inside `TrustedMetadataSet` we throw `RepositoryError`, `ReplayedMetadataError`, `ExpiredMetadataError` and `UnsignedMetadataError` (from `verify_delegate()`) at the same time.
One idea could be to define new specific exceptions (which derive from `RepositoryError` and replace the instances of `RepositoryError` with them. 

Read and synchronize with issues:
- #1778 
- #1779 
- #1714
- #1312 

2. Inside the `TrustedMetadataSet` `update*` function docstrings we only document that we are throwing `RepositoryError` without specifying the specific exceptions. If none of them are documented in the Raises section of the method docstrings (they only mention the broad RepositoryError), do we expect anyone to handle them? If we do, we should probably document them?

_Inspired by https://github.com/theupdateframework/python-tuf/pull/1725#discussion_r787713776._

3. Inside `_get_session` in module `tuf/ngclient/_internal/request_fetcher.py` we currently throw `URLParsingError`: https://github.com/theupdateframework/python-tuf/blob/0e26364dfcb54bc7c0eee9f7d2ef7a67b9b53868/tuf/ngclient/_internal/requests_fetcher.py#L135
In my commit https://github.com/theupdateframework/python-tuf/pull/1725/commits/e8f26ebb1e7c68fc9c0ddd5b3a268c0a56a7a304 inside pr #1725 I am replacing it with `DownloadError`.
The question is do we want to keep the specific exception - `URLParsingError` and make it derive from `DownloadError`?
Does my change in this commit make sense?

jku:
> The question is do we want to keep the specific exception - URLParsingError and make it derive from DownloadError?
Does my change in this commit make sense?

[...]

PR #1810, opened 2022-01-31
ngclient: Make DownloadErrors more consistent

### LINKED ISSUE #1714: ngclient: decide what to do with requests (Fetcher impl) errors
I wonder if we should handle all exceptions thrown by Fetcher implementations (and throw a generic error that clients can actually handle)?

This issue can be tested with `examples/client_example/client_example.py`: if e.g. server is not running there's a requests `ConnectionError`. All requests errors do derive from `OSError` so it gets handled by the example code but another fetcher implementation might not do the same. 

Alternatively we can just say that Fetcher implementations should document what gets thrown (and requestsFetcher could just link to RequestException)

jku:
* Make fetch() wrap non-DownloadErrors in a DownloadError
* Try to be more consistent in docstrings
* rename FetcherHTTPError to DownloadHTTPError

As a result calling any public fetcher methods is now guaranteed to only raise DownloadError (or something derived from it). I think this looks cleaner, makes it easier understand and implementing a fetcher is not more complex: Fetcher implementations are allowed to raise whatever errors (like RequestsError in the default case) but `FetcherInterface.fetch()` will automatically wrap errors so that only DownloadErrors are raised to calling code.

This is strictly speaking an API change (because of the rename) but that error is meant for mostly internal use.

Fixes #1714

### DIFF
--- a/examples/client_example/client_example.py
+++ b/examples/client_example/client_example.py
@@ -10,7 +10,7 @@ import os
 import shutil
 from pathlib import Path
 
-from tuf.api.exceptions import RepositoryError
+from tuf.api.exceptions import DownloadError, RepositoryError
 from tuf.ngclient import Updater
 
 # constants
@@ -73,8 +73,8 @@ def download(target: str) -> bool:
         path = updater.download_target(info)
         print(f"Target downloaded and available in {path}")
 
-    except (OSError, RepositoryError) as e:
-        print(str(e))
+    except (OSError, RepositoryError, DownloadError) as e:
+        print(f"Failed to download target {target}: {e}")
         return False
 
     return True
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -60,7 +60,7 @@ class RequestsFetcher(FetcherInterface):
         Raises:
             exceptions.SlowRetrievalError: A timeout occurs while receiving
                 data.
-            exceptions.FetcherHTTPError: An HTTP error code is received.
+            exceptions.DownloadHTTPError: An HTTP error code is received.
 
         Returns:
             A bytes iterator
@@ -88,7 +88,7 @@ class RequestsFetcher(FetcherInterface):
         except requests.HTTPError as e:
             response.close()
             status = e.response.status_code
-            raise exceptions.FetcherHTTPError(str(e), status)
+            raise exceptions.DownloadHTTPError(str(e), status)
 
         return self._chunks(response)
 
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -121,7 +121,7 @@ class RequestsFetcher(FetcherInterface):
         parsed_url = parse.urlparse(url)
 
         if not parsed_url.scheme or not parsed_url.hostname:
-            raise exceptions.DownloadError("Failed to parse URL {url}")
+            raise exceptions.DownloadError(f"Failed to parse URL {url}")
 
         session_index = f"{parsed_url.scheme}+{parsed_url.hostname}"
         session = self._sessions.get(session_index)
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -51,7 +51,7 @@ class RequestsFetcher(FetcherInterface):
         self.socket_timeout: int = 4  # seconds
         self.chunk_size: int = 400000  # bytes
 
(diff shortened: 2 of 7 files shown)


jku:
linter failure is #1811

tuf/ngclient/fetcher.py
```diff
@@ -28,38 +28,66 @@ class FetcherInterface:
...
+            max_length: an integer value representing the length of
+                the file or an upper bound.
 
         Raises:
-          exceptions.DownloadLengthMismatchError: downloaded bytes exceed
-            'max_length'.
+            exceptions.DownloadError: An error occurred during download.
+            exceptions.DownloadLengthMismatchError: downloaded bytes exceed
+                'max_length'.
+            exceptions.FetcherHTTPError: An HTTP error code was received.
```
MVrachev: Here as well as inside the `download_bytes()` docstrings we document specific`DownloadError`s - `DownloadLengthMismatchError` and `FetcherHTTPError` next to the more general `DownloadError`. 

Do we think that's okay?
@lukpueh what do you think as you have been working on #1784.

I see why we do that here as those errors could be relevant to the user and handled by him/her, but wanted to make sure we discuss this.

MVrachev (commented):
Liked the changes overall.
I agree with the reasoning behind the rename of `FetcherHTTPError` to `DownloadHTTPError`.

One question and wanted to ask for a favor.
I just noticed I have forgotten to add an `f` for f-string when fixing the `DownloadError` here:
https://github.com/theupdateframework/python-tuf/blob/cb7bd6aff9be097facecbd27ee5998c48feeb52d/tuf/ngclient/_internal/requests_fetcher.py#L125 can you add it as a part of this pr?

tuf/ngclient/fetcher.py
jku: yeah, there's no great solutions here, I just picked the one I dislike least: 
* I don't want to create new sub-errors just for the sake of it 
* in this case it makes sense to point out the specific errors and not just the generic one (because callers will be interested)
* but the problem you mention does make the docstring less clear than it could be

jku:
force-pushed, only changes are:
* rebased on develop to get a clean build
* added a new commit to add f to an f-string

MVrachev (approved):
LGTM

tuf/ngclient/fetcher.py
```diff
@@ -28,38 +28,65 @@ class FetcherInterface:
...
+            A bytes iterator
+        """
+        # Ensure that fetch() only raises DownloadErrors, regardless of the
+        # fetcher implementation
+        try:
+            return self._fetch(url)
+        except Exception as e:
+            if isinstance(e, exceptions.DownloadError):
+                raise e
+            raise exceptions.DownloadError(f"Failed to download {url}") from e
```
lukpueh: Probably a matter of taste, but wouldn't this look more idiomatic?

```suggestion
        except exceptions.DownloadError:
            raise
        except Exception as e
            raise exceptions.DownloadError(f"Failed to download {url}") from e
```

examples/client_example/client_example.py
```diff
@@ -73,8 +73,8 @@ def download(target: str) -> bool:
         path = updater.download_target(info)
         print(f"Target downloaded and available in {path}")
 
-    except (OSError, RepositoryError) as e:
-        print(str(e))
+    except (OSError, RepositoryError, DownloadError) as e:
```
lukpueh, line 76: Do I understand correctly that any download related errors were just not caught here before, and that re-raising all of them as DownloadErrors allows us to easily handle them too here?

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -122,7 +121,7 @@ def _get_session(self, url: str) -> requests.Session:
         parsed_url = parse.urlparse(url)
 
         if not parsed_url.scheme or not parsed_url.hostname:
-            raise exceptions.DownloadError("Failed to parse URL {url}")
+            raise exceptions.DownloadError(f"Failed to parse URL {url}")
```
lukpueh, line 124: Thanks for catching this, @MVrachev!

lukpueh (approved):
I like this! Just added one minor coding style suggestion inline, feel free to ignore and merge as is.

examples/client_example/client_example.py
jku: they were handled, but only because requests errors all happen to be derived from OSError -- so this was relying on a fetcher implementation detail.

tuf/ngclient/fetcher.py
jku: oh definitely, I'll change this

jku:
rebased to get the build fix from develop, no changes.

PR #1815, opened 2022-02-01
Improve docstrings language in Metadata API

### LINKED ISSUE #1600: Metadata API: Improve docstrings language

ivanayov:
This change updates some obvious and unnecessary fields docs in the
Metadata API with more despriptive details, as well as unifies
syntax between different classes and methods

Addresses #1600

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -107,9 +107,9 @@ class Metadata(Generic[T]):
     instance attributes.*
 
     Args:
-        signed: The actual metadata payload, i.e. one of ``Targets``,
+        signed: Actual metadata payload, i.e. one of ``Targets``,
             ``Snapshot``, ``Timestamp`` or ``Root``.
-        signatures: An ordered dictionary of keyids to ``Signature`` objects, each
+        signatures: Ordered dictionary of keyids to ``Signature`` objects, each
             signing the canonical serialized representation of ``signed``.
     """
 
@@ -131,7 +131,7 @@ class Metadata(Generic[T]):
             Destroys the metadata dict passed by reference.
 
         Returns:
-            A TUF ``Metadata`` object.
+            TUF ``Metadata`` object.
         """
 
         # Dispatch to contained metadata class on metadata _type field.
@@ -174,11 +174,11 @@ class Metadata(Generic[T]):
         """Loads TUF metadata from file storage.
 
         Args:
-            filename: The path to read the file from.
-            deserializer: A ``MetadataDeserializer`` subclass instance that
+            filename: Path to read the file from.
+            deserializer: ``MetadataDeserializer`` subclass instance that
                 implements the desired wireline format deserialization. Per
                 default a ``JSONDeserializer`` is used.
-            storage_backend: An object that implements
+            storage_backend: Object that implements
                 ``securesystemslib.storage.StorageBackendInterface``.
                 Default is ``FilesystemBackend`` (i.e. a local file).
         Raises:
@@ -187,7 +187,7 @@ class Metadata(Generic[T]):
                 The file cannot be deserialized.
 
         Returns:
-            A TUF ``Metadata`` object.
+            TUF ``Metadata`` object.
         """
 
         if storage_backend is None:
@@ -205,7 +205,7 @@ class Metadata(Generic[T]):
         """Loads TUF metadata from raw data.
 
         Args:
-            data: metadata content.
+            data: Metadata content.
             deserializer: ``MetadataDeserializer`` implementation to use.
                 Default is ``JSONDeserializer``.
 
@@ -214,7 +214,7 @@ class Metadata(Generic[T]):
                 The file cannot be deserialized.
 
         Returns:
-            A TUF ``Metadata`` object.
+            TUF ``Metadata`` object.
         """
 
         if deserializer is None:
@@ -239,7 +239,7 @@ class Metadata(Generic[T]):
         instead of re-serializing.
 
         Args:
-            serializer: A ``MetadataSerializer`` instance that implements the
+            serializer: ``MetadataSerializer`` instance that implements the
                 desired serialization format. Default is ``JSONSerializer``.
 
         Raises:
@@ -279,10 +279,10 @@ class Metadata(Generic[T]):
         instead of re-serializing.
 
         Args:
-            filename: The path to write the file to.
-            serializer: A ``MetadataSerializer`` instance that implements the
+            filename: Path to write the file to.
+            serializer: ``MetadataSerializer`` instance that implements the
                 desired serialization format. Default is ``JSONSerializer``.
-            storage_backend: A ``StorageBackendInterface`` implementation. Default
+            storage_backend: ``StorageBackendInterface`` implementation. Default
                 is ``FilesystemBackend`` (i.e. a local file).
 
         Raises:
@@ -307,11 +307,11 @@ class Metadata(Generic[T]):
         """Creates signature over ``signed`` and assigns it to ``signatures``.
 
         Args:
-            signer: A securesystemslib.signer.Signer implementation.
+            signer: ``securesystemslib.signer.Signer`` implementation.
             append: ``True`` if the signature should be appended to
                 the list of signatures or replace any existing signatures. The
                 default behavior is to replace signatures.
-            signed_serializer: A ``SignedSerializer`` that implements the desired
+            signed_serializer: ``SignedSerializer`` that implements the desired
                 serialization format. Default is ``CanonicalJSONSerializer``.
 
         Raises:
@@ -320,7 +320,8 @@ class Metadata(Generic[T]):
             exceptions.UnsignedMetadataError: Signing errors.
 
         Returns:
-            ``securesystemslib.signer.Signature`` object that was added into signatures.
+            ``securesystemslib.signer.Signature`` object that was added into
+            signatures.
         """
 
         if signed_serializer is None:
@@ -357,8 +358,8 @@ class Metadata(Generic[T]):
 
         Args:
             delegated_role: Name of the delegated role to verify
-            delegated_metadata: The ``Metadata`` object for the delegated role
-            signed_serializer: serializer used for delegate
+            delegated_metadata: ``Metadata`` object for the delegated role
+            signed_serializer: Serializer used for delegate
                 serialization. Default is ``CanonicalJSONSerializer``.
 
         Raises:
@@ -411,9 +412,9 @@ class Signed(metaclass=abc.ABCMeta):
     instance attributes.*
 
     Args:
-        version: The metadata version number.
-        spec_version: The supported TUF specification version number.
-        expires: The metadata expiry date.
+        version: Metadata version number.
+        spec_version: Supported TUF specification version number.
+        expires: Metadata expiry date.
         unrecognized_fields: Dictionary of all unrecognized fields.
 
     Raises:
@@ -524,7 +525,7 @@ class Signed(metaclass=abc.ABCMeta):
         """Checks metadata expiration against a reference time.
 
         Args:
-            reference_time: The time to check expiration date against. A naive
+            reference_time: Time to check expiration date against. A naive
                 datetime in UTC expected. Default is current UTC date and time.
 
         Returns:
@@ -549,8 +550,8 @@ class Key:
         keyid: Key identifier that is unique within the metadata it is used in.
             Keyid is not verified to be the hash of a specific representation
             of the key.
-        keytype: key type, e.g. "rsa", "ed25519" or "ecdsa-sha2-nistp256".
-        scheme: signature scheme. For example:
+        keytype: Key type, e.g. "rsa", "ed25519" or "ecdsa-sha2-nistp256".
+        scheme: Signature scheme. For example:
             "rsassa-pss-sha256", "ed25519", and "ecdsa-sha2-nistp256".
         keyval: Opaque key content
         unrecognized_fields: Dictionary of all unrecognized fields.
@@ -614,7 +615,7 @@ class Key:
         removing the private key from keyval.
 
         Args:
-            key_dict: A key in securesystemlib dict representation.
+            key_dict: Key in securesystemlib dict representation.
         """
         key_meta = sslib_keys.format_keyval_to_metadata(
             key_dict["keytype"],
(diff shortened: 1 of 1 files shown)


tuf/api/metadata.py
```diff
@@ -966,7 +987,11 @@ class Timestamp(Signed):
         version: The metadata version number.
         spec_version: The supported TUF specification version number.
         expires: The metadata expiry date.
-        unrecognized_fields: Dictionary of all unrecognized fields.
+        unrecognized_fields: A Dictionary of all unrecognized attribute-value
+            fields. It guarantees that input containing unknown fields will
+            not break serialization from and deserialisation into a Metadta
+            object and thus aligns with the TUF specification implication that
+            unrecognized attribute-value fields should be supported
         snapshot_meta: Meta information for snapshot metadata.
```
ivanayov, line 1021: This is one I wondered whether to edit or not. It relates to obvious and unnecessary information, but on the other side it's simple and correct. I wondered what's your opinion about explaining why it's available in the API - do you find such clarification good or rather unnecessary?

tuf/api/metadata.py
```diff
@@ -118,14 +118,14 @@ def __init__(self, signed: T, signatures: Dict[str, Signature]):
 
     @classmethod
     def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata[T]":
-        """Creates Metadata object from its dict representation.
+        """Creates ``Metadata`` object from its dict representation.
```
jku, line 122: We say "dict representation" a lot as if it's obvious what that means :)

I don't have a great alternative suggestion, just commenting.

tuf/api/metadata.py
```diff
@@ -118,14 +118,14 @@ def __init__(self, signed: T, signatures: Dict[str, Signature]):
...
 
-        Arguments:
-            metadata: TUF metadata in dict representation.
+        Args:
+            metadata: A TUF metadata in dict representation.
 
         Raises:
             KeyError: The metadata dict format is invalid.
-            ValueError: The metadata has an unrecognized signed._type field.
+            ValueError: The metadata has an unrecognized ``signed._type`` field.
```
jku: ValueError docstring is likely wrong: there are many more ways to trigger it than wrong type
I think both errors should be documented with same vague message like "Invalid metadata argument".

tuf/api/metadata.py
```diff
@@ -339,18 +340,18 @@ def verify_delegate(
...
+            delegated_role: A name of the delegated role to verify
+            delegated_metadata: The ``Metadata`` object for the delegated role
             signed_serializer: serializer used for delegate
-                serialization. Default is CanonicalJSONSerializer.
+                serialization. Default is ``CanonicalJSONSerializer``.
 
         Raises:
-            UnsignedMetadataError: 'delegate' was not signed with required
-                threshold of keys for 'role_name'
+            UnsignedMetadataError: A ``delegate`` was not signed with required
```
jku: First, this refers to "delegate" but should refer to "delegated_role".
 
Second, here's an example of indefinite article looking inaccurate: The error is raised because **the specific value** of argument 'delegated_role' is invalid. I think the original (without article) is correct. I think this is true for each "raises" docstring that mentions an argument.

tuf/api/metadata.py
```diff
@@ -401,7 +402,11 @@ class Signed(metaclass=abc.ABCMeta):
         version: The metadata version number.
         spec_version: The supported TUF specification version number.
         expires: The metadata expiry date.
-        unrecognized_fields: Dictionary of all unrecognized fields.
+        unrecognized_fields: A Dictionary of all unrecognized attribute-value
+            fields. It guarantees that input containing unknown fields will
+            not break serialization from and deserialisation into a Metadta
+            object and thus aligns with the TUF specification implication that
+            unrecognized attribute-value fields should be supported
```
jku: * typo of "Metadata"
* Makes promises we might not be willing to keep (you can certainly put things into unrecognised_fields we won't serialize/deserialize successfully)
* quite long, especially if copied many times over: let's try to squeeze this into a couple of lines that describe the intent (but not necessarily mechanism) in a way that's useful to a developer. If a longer description is needed, let's do that in a single place (like module docstring)

I can try to improve this if it seems difficult

tuf/api/metadata.py
```diff
@@ -416,7 +421,7 @@ def _type(self) -> str:
     def expires(self) -> datetime:
         """The metadata expiry date::
 
-        # Use 'datetime' module to e.g. expire in seven days from now
+        # Use ``datetime`` module to e.g. expire in seven days from now
```
jku: This is not useful: ordinary comments are not included in ReadTheDocs where the double backtick is useful. 

This applies to all instance where non-docstring comments are modified like this

tuf/api/metadata.py
jku: (reply to another comment) 
Yeah documenting meta in a comprehensive manner requires a lot of specification context so maybe we leave it as is... The MetaFile type does clearly document its content after all.

tuf/api/metadata.py
```diff
@@ -87,7 +87,7 @@ class Metadata(Generic[T]):
     Provides methods to convert to and from dictionary, read and write to and
     from file and to create and verify metadata signatures.
 
-    Metadata[T] is a generic container type where T can be any one type of
+    ``Metadata[T]`` is a generic container type where T can be any one type of
     [Root, Timestamp, Snapshot, Targets]. The purpose of this is to allow
```
jku: I suppose you could double-backtick all of these as well to show that they are actual python types/classes?

tuf/api/metadata.py
jku: (this btw is why I'm a bit hesitant about committing to documenting ValueError/KeyError in a way that _looks_ comprehensive: the documentation is going to be either incomplete or vague. It's not practical to do the massive amount of work of actually trying to list all possible cases where object construction can fail when we rely on the core libraries/data structures raising errors so much)

jku (commented):
I wish this was in multiple commits so we could easily decide to only take some changes now and leave others for discussion... This issue should maybe have been split by change type to make that clearer that things like consistent double-backticks are easier to find a solution to than other issues (like consistent use of article where we need to decide what we want). The changes do conflict with each other so I can see why you did them all at once.

Anyway, I'll review but let's discuss if it makes sense to split this or to continue tweaking this PR.

--- 

There's a lot of good changes here, just commenting on potential issues.

I'm not sold on the use of indefinite article everywhere. It's fine in some places but looks quite bad in others:
`threshold: A number of keys required to...` 
`length: A length of the target file in bytes`
those do not look right

Also the use of indefinite article is not consistent: there are still places that use definite article instead and the commit even adds a few new definite articles (see TargetFile methods for examples). I don't think 100% consistency is required but I don't understand how the article is chosen...

Overall, I think the indefinite article does not work in many places: In my opinion we should not standardize on it. I would maybe suggest a default choice of no article (since it makes docstrings shorter) and diverging from that if the context requires it.

ivanayov:
@jku Thank you for the review! I've split to several commits, as you advice. It's a bit harder to address feedback, as it should go to a specific commit in a single-file update, but I agree it's a better history at the end.

tuf/api/metadata.py
```diff
@@ -1379,13 +1373,13 @@ def from_data(
...
-        """Verifies that length and hashes of "data" match expected values.
+        """Verifies that length and hashes of ``data`` match expected values.
 
         Args:
-            data: File object or its content in bytes.
+            data: Target file object or its content in bytes.
 
         Raises:
-            LengthOrHashMismatchError: Calculated length or hashes do not
+            LengthOrHashMismatchError: The calculated length or hashes do not
```
ivanayov: This one is probably left by mistake. Will revert it at the end while addressing other comments to save time.

tuf/api/metadata.py
```diff
@@ -120,18 +121,14 @@ def __init__(self, signed: T, signatures: Dict[str, Signature]):
...
         """Creates Metadata object from its dict representation.
 
-        Arguments:
+        Args:
             metadata: TUF metadata in dict representation.
 
-        Raises:
-            KeyError: The metadata dict format is invalid.
-            ValueError: The metadata has an unrecognized signed._type field.
-
```
MVrachev: @jku I think we agreed we want to document `KeyError` and `ValueError` from `from_dict()` and the constructors?
That's why I created #1819.

MVrachev: @ivanayov @jku has agreed we want to document those exceptions.
I will fix that when proposing a pr for #1819.
Could you remove that change?

ivanayov: @MVrachev yep, I'll revert then.

ivanayov:
Python setup failed on windows.

tuf/api/metadata.py
```diff
@@ -1140,7 +1150,7 @@ def to_dict(self) -> Dict[str, Any]:
 
     @staticmethod
     def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
-        """Determines whether "targetname" matches the "pathpattern"."""
+        """Determines whether ``targetname`` matches the ``pathpattern``."""
```
```suggestion
        """Determines whether ``targetpath`` matches the ``pathpattern``."""
```

tuf/api/metadata.py
```diff
@@ -788,14 +794,14 @@ def add_key(self, role: str, key: Key) -> None:
...
 
     def remove_key(self, role: str, keyid: str) -> None:
-        """Removes key from 'role' and updates the key store.
+        """Removes key from ``role`` and updates the key store.
 
         Args:
-            role: The name of the role, for which a signing key is removed.
-            key: The identifier of the key to be removed for 'role'.
+            role: Name of the role, for which a signing key is removed.
+            key: Identifier of the key to be removed for ``role``.
```
```suggestion
            keyid: Identifier of the key to be removed for ``role``.
```

tuf/api/metadata.py
```diff
@@ -1476,16 +1489,16 @@ def add_key(self, role: str, key: Key) -> None:
...
     def remove_key(self, role: str, keyid: str) -> None:
-        """Removes key from delegated role 'role' and updates the delegations
+        """Removes key from delegated role ``role`` and updates the delegations
         key store.
 
         Args:
-            role: The name of the role, for which a signing key is removed.
-            key: The identifier of the key to be removed for 'role'.
+            role: Name of the role, for which a signing key is removed.
+            key: Identifier of the key to be removed for ``role``.
```
```suggestion
            keyid: Identifier of the key to be removed for ``role``.
```

jku (changes requested):
thanks for the nice commits: I was expecting to suggest moving some commits to a later PR... but I think I actually agree with all of these, they seem to be clear improvements. Only thing that is a required fix is a rebase (merge is not currently clean)

There are things to improve in the docs but this is a big PR already and there's no need to fix everything at once, just leaving some comments: 
* I noticed some docstring errors and left comments: none of them seem to be your fault so feel free to either fix here or leave to next PR
* The unrecognised_fields docstring feels about the right amount of detail to me -- maybe we 'd like to explain general concept in one place somewhere but individual docstrings look ok
* Many arguments still explain "what" the argument is, not "why" it is: as an example we document signer argument with `securesystemslib.signer.Signer implementation` which does not explain why signer is an argument (what is it used for).

... but all of these can be handled in future improvements. I've re-run the tests after setup-python fixed itself: they look good.

lukpueh:
Thanks for working on this, @ivanayov and for reviewing, @jku! Let me know if you want more eyes.

tuf/api/metadata.py
```diff
@@ -788,14 +794,14 @@ def add_key(self, role: str, key: Key) -> None:
...
 
         Raises:
-            ValueError: If 'role' doesn't exist or if 'role' doesn't include
+            ValueError: If ``role' doesn't exist or if ``role`` doesn't include
```
jku: Oh found an error:
```suggestion
            ValueError: If ``role`` doesn't exist or if ``role`` doesn't include
```

ivanayov:
Thank you @jku ! I addressed your comments here, as they are related to the changes anyway. 
Changed 
```
signer: ``securesystemslib.signer.Signer`` implementation
```
 to 
```
signer: ``securesystemslib.signer.Signer`` implementation containing a
signature and a keyid of the key that signature was generated with.
```
The other changes are just the fixes that you proposed + moving some of the existing changes  to the right commit (I was a bit lazy to do that before addressing the comments 😁).

There are still some 
```
data: Metadata content.
...
filename: Path to write the file to.
etc.
```
but I think there's enough context around them and adding explanations could make the docs too verbose and might rather confuse the user by creating expectations of some deeper meaning behind. Do you think that those should be better explained? @lukpueh do you have some preference, too?

jku (changes requested):
Thank you. I think I would like to leave the signer argument improvements for later: It is significantly more difficult to explain "why" than "what" -- and this PR is long enough.

In practice:
```
signer: ``securesystemslib.signer.Signer`` implementation containing a
signature and a keyid of the key that signature was generated with.
```

this is slightly incorrect (signer does not contain a signature, it produces signatures) and still doesn't explain why there is a signer argument. Could you just leave the original description in and then we can continue tweaking these later?

tuf/api/metadata.py
lukpueh: Would it be better non-abbreviated, i.e. "dictionary representation"? I think I used dict to keep it on one line.
Or would "json/dict representation" help people better associate it with the format described in the spec?

lukpueh:
> There are still some
> 
> ```
> data: Metadata content.
> ...
> filename: Path to write the file to.
> etc.
> ```
> 
> but I think there's enough context around them and adding explanations could make the docs too verbose and might rather confuse the user by creating expectations of some deeper meaning behind. Do you think that those should be better explained? @lukpueh do you have some preference, too?

I think both are okay. Especially the latter is actually adds useful information not provided by the arg name alone.

ivanayov:
@jku I removed the change.

tuf/api/metadata.py
ivanayov: I think "json/dict representation" is good. "dict" vs. "dictionary" is quite similar, while "json representation" should be obvious for everyone.

jku (approved):

jku:
Thanks!

PR #1820, opened 2022-02-03
Add exceptions docs for __init__ and from_dict()

### LINKED ISSUE #1819: Document exceptions for __init__ and from_dict() methods in Metadata API
Document `ValueError` (and `KeyError` if needed) exceptions for `__init__` and `from_dict()` methods in Metadata API.

The reason this was not done is that this sort of makes it looks like the programmer should handle these errors... but as long as we are clear in the docstring that these are invalid argument errors, this should be fine

For context read https://github.com/theupdateframework/python-tuf/issues/1785

MVrachev:
Fixes #1819

Document `ValueError`, `KeyError` and `TypeError` exceptions for `__init__` and
`from_dict()` methods in Metadata API.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -124,8 +124,7 @@ class Metadata(Generic[T]):
             metadata: TUF metadata in dict representation.
 
         Raises:
-            KeyError: The metadata dict format is invalid.
-            ValueError: The metadata has an unrecognized signed._type field.
+            ValueError, KeyError, TypeError: Invalid arguments.
 
         Side Effect:
             Destroys the metadata dict passed by reference.
@@ -416,6 +415,9 @@ class Signed(metaclass=abc.ABCMeta):
         spec_version: The supported TUF specification version number.
         expires: The metadata expiry date.
         unrecognized_fields: Dictionary of all unrecognized fields.
+
+    Raises:
+        ValueError: Invalid arguments.
     """
 
     # type is required for static reference without changing the API
@@ -551,6 +553,9 @@ class Key:
             "rsassa-pss-sha256", "ed25519", and "ecdsa-sha2-nistp256".
         keyval: Opaque key content
         unrecognized_fields: Dictionary of all unrecognized fields.
+
+    Raises:
+        TypeError: Invalid type for an argument.
     """
 
     def __init__(
@@ -573,7 +578,11 @@ class Key:
 
     @classmethod
     def from_dict(cls, keyid: str, key_dict: Dict[str, Any]) -> "Key":
-        """Creates Key object from its dict representation."""
+        """Creates Key object from its dict representation.
+
+        Raises:
+            KeyError, TypeError: Invalid arguments.
+        """
         keytype = key_dict.pop("keytype")
         scheme = key_dict.pop("scheme")
         keyval = key_dict.pop("keyval")
@@ -680,6 +689,9 @@ class Role:
         keyids: The roles signing key identifiers.
         threshold: Number of keys required to sign this role's metadata.
         unrecognized_fields: Dictionary of all unrecognized fields.
+
+    Raises:
+        ValueError: Invalid arguments.
     """
 
     def __init__(
@@ -698,7 +710,11 @@ class Role:
 
     @classmethod
     def from_dict(cls, role_dict: Dict[str, Any]) -> "Role":
-        """Creates Role object from its dict representation."""
+        """Creates Role object from its dict representation.
+
+        Raises:
+            ValueError, KeyError: Invalid arguments.
+        """
         keyids = role_dict.pop("keyids")
         threshold = role_dict.pop("threshold")
         # All fields left in the role_dict are unrecognized.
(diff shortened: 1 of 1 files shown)


jku (commented):
I might leave this for lukas to handle, the only comment I have is that I have a strong sense that many docstrings claim things that we have not checked. An example I found in`Targets.from_dict()`:

    TypeError: Invalid type for a signed_dict value.

I'm pretty sure TypeError can also happen if "delegations" is not a dict. And if "targets" contains non-dict. And if if "delegations.roles" contains non-dicts. Maybe there are more cases? I don't think we should spend a lot of time figuring out the exact cases since no-one will care. If we are going to try and document this, let's try to make it non-specific enough that it may actually be correct and has a chance of staying correct

Would it be wrong to do something like this for each `from_dict()` for example:

    KeyError, TypeError, ValueError: Invalid arguments

lukpueh:
> 
> Would it be wrong to do something like this for each `from_dict()` for example:
> 
> ```
> KeyError, TypeError, ValueError: Invalid arguments
> ```


Works for me. 👍  @MVrachev, mind updating the PR as Jussi said and then ping me for review?

MVrachev:
@lukpueh you can review it now.
Not everywhere we threw all three of those `KeyError, TypeError, ValueError`, so I documented only those that I saw that are thrown.

tuf/api/metadata.py
```diff
@@ -889,6 +912,9 @@ class MetaFile(BaseFile):
         hashes: Dictionary of hash algorithm names to hashes of the metadata
             file content.
         unrecognized_fields: Dictionary of all unrecognized fields.
+
+    Raises:
+        ValueError: Invalid arguments.
```
lukpueh: I think this misses `TypeError` (from the `_validate_hashes` call).

tuf/api/metadata.py
```diff
@@ -402,6 +401,9 @@ class Signed(metaclass=abc.ABCMeta):
         spec_version: The supported TUF specification version number.
         expires: The metadata expiry date.
         unrecognized_fields: Dictionary of all unrecognized fields.
+
+    Raises:
+        ValueError: Invalid arguments.
```
lukpueh, line 420: Can you remind me, do we expect users to catch type errors by verifying our type annotations? If not, we'd also need to expect an `AttributeError` here in case the `spec_version` passed to the constructor does not have a `split` method, which we call in:
https://github.com/theupdateframework/python-tuf/blob/dd59adafef6fcb85aa910f4ec39564886cca91dc/tuf/api/metadata.py#L440

OTOH, I noticed a few `isinstance`-checks in other constructors in the module, that we probably wouldn't need if we  relied on users running mypy, e.g.:
https://github.com/theupdateframework/python-tuf/blob/dd59adafef6fcb85aa910f4ec39564886cca91dc/tuf/api/metadata.py#L549-L558

I guess the difference between those two example that in the former a user sees the issue right away, but not in the latter.

We don't have to solve this in this PR, I was just curious.

lukpueh (changes requested):
Thanks for your diligence, @MVrachev! I found one small mistake (see comment inline) and raised another not immediately related concern. I think we only need to fix the former to land this PR.

tuf/api/metadata.py
MVrachev: I don't think we have a strong stance on when should we check for types.
I think I agree that in the `Key` example it's harder to notice that these fields are not strings.
I don't think we can rely on our users `mypy` as it's not an official requirement to run python-tuf, but at the same time, we don't want to overdo it with the type checks as we are working with untyped language.
@jku any opinion on this?

jku: > I noticed a few isinstance-checks in other constructors in the module, that we probably wouldn't need if we relied on users running mypy

For deserialization code path (including constructors as that's how the code path is built, for better or worse) we can't tell users to rely on mypy: the input is after all just random objects coming from `json.loads()`.

I think in these cases documenting TypeError is as useful as documenting ValueError and KeyError... That is "not very useful at all" but if we do it we should try to have full coverage.

tuf/api/metadata.py
```diff
@@ -124,8 +124,7 @@ def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata[T]":
             metadata: TUF metadata in dict representation.
 
         Raises:
-            KeyError: The metadata dict format is invalid.
-            ValueError: The metadata has an unrecognized signed._type field.
+            ValueError, KeyError: Invalid arguments.
```
jku: It is hard to believe we can't get TypeError here, considering we potentially call every other `from_dict()`?

jku (changes requested):
I can't see any missing errors (apart from the suspicion in a code comment) but I imagine there may well be more: I'm fine with this PR otherwise though

tuf/api/metadata.py
MVrachev: Yes, you are correct. Just calling `Key.from_dict()` will lead to `TypeError`.

jku (approved):

jku:
I approved this: I'm not entirely convinced that the coverage really is 100% complete but I also don't think it makes sense to put more effort into it.

lukpueh (approved):

PR #1821, opened 2022-02-04
github: disable pip caching temporarily

jku:
This is a temporary workaround until Github fixes either setup-python or the Windows image... If we merge this, let's file an issue to revert the workaround later.

---

setup-python fails on Windows currently
(https://github.com/actions/virtual-environments/issues/5009)
Disable caching to workaround the failure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,8 +47,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'pip'
-          cache-dependency-path: 'requirements*.txt'
+          # cache: 'pip'
+          # cache-dependency-path: 'requirements*.txt'
 
       - name: Install dependencies
         run: |


MVrachev:
> This is a temporary workaround until Github fixes either setup-python or the Windows image... If we merge this, let's file an issue to revert the workaround later.
> 
> setup-python fails on Windows currently ([actions/virtual-environments#5009](https://github.com/actions/virtual-environments/issues/5009)) Disable caching to workaround the failure.
> 
> Signed-off-by: Jussi Kukkonen [jkukkonen@vmware.com](mailto:jkukkonen@vmware.com)

I saw your conversation on the issue about the `pip cache`, but I am not sure I understand.
How does that fix the problem? Can you elaborate more on this?

jku:
> How does that fix the problem? Can you elaborate more on this?

It does not fix anything but makes setup-python not try something (caching) that currently fails on the Windows image: I don't know what that is but it might be trying to create a cache directory without enough permissions or something like that.

lukpueh (approved):
Thanks for the quick troubleshooting and filing a bug upstream, @jku! I'll merge here so that we have a working CI. Let's re-enable with #1822.

PR #1823, opened 2022-02-04
README: Fix CI badge

jku:
The URL was referring to an older workflow name which lead to a badge
that no longer updates based on CI results. This commit also changes to
using "actions/workflows/FILENAME/badge.svg" as current documentation
suggests.


lukpueh (approved):
Cross-checked with [GitHub docs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge#using-the-workflow-file-name). Thanks for fixing, @jku!

PR #1824, opened 2022-02-04
Revert "github: disable pip caching temporarily"

jku:
This reverts commit 55d6cb47da9d2eb3bd6ebdbc3b93a5b8884d9454.

According to changelog setup-python v2.3.2 should include a workaround
for the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,8 +47,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-          # cache: 'pip'
-          # cache-dependency-path: 'requirements*.txt'
+          cache: 'pip'
+          cache-dependency-path: 'requirements*.txt'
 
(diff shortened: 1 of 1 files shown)


lukpueh:
Heureka! 🎉  Can you squash, signoff and we merge this?

jku:
I had to pin 2.3.2 but I only did that because they forgot to move v2 tag.

Now maybe we should pin actions versions (?) but let's not decide that in a workaround: let's wait until v2 exists and then I remove the last commit from here

joshuagl:
> Now maybe we should pin actions versions (?) but let's not decide that in a workaround: let's wait until v2 exists and then I remove the last commit from here

I believe we should pin actions versions (we do in the specification)

jku:
> I believe we should pin actions versions (we do in the specification)

I suppose you are right: #1826

setup-python v2 tag has been updated, this PR is good to go.

lukpueh (approved):

PR #1825, opened 2022-02-04
Update supported specification version to 1.0.28

### LINKED ISSUE #1763: specification has new version
It seems specification (https://github.com/theupdateframework/specification/blob/master/tuf-spec.md) has new version. 
Please review the version.

MVrachev:
Fixes #1763

I looked into all changes between our current version 1.0.19 and the
current version of the specification 1.0.28 and I agree with Jussi that
the only one not fully resolved is:
"8dafd00 (tag: v1.0.24) Clarify optional attributes" and more precisely
the changes from commit:
https://github.com/theupdateframework/specification/pull/165/commits/4dd279bc318afaea9c069b265c0468e235df0192

It doesn't make sense to have a target file without "paths" or
"path_hash_prefixes", so our `python-tuf` requirement to have at least
one of them set makes sense.

Both with Jussi we agreed that we can easily loosen this requirement if
when solving https://github.com/theupdateframework/specification/issues/200
it's decided that both of them can be omitted, but for now,
we decided it's better to stick to our current requirement to have one of them set.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -74,7 +74,7 @@ logger = logging.getLogger(__name__)
 
 # We aim to support SPECIFICATION_VERSION and require the input metadata
 # files to have the same major version (the first number) as ours.
-SPECIFICATION_VERSION = ["1", "0", "19"]
+SPECIFICATION_VERSION = ["1", "0", "28"]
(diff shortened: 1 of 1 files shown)


jku (approved):
Looks correct to me.

PR #1830, opened 2022-02-07
github: Pin actions hashes

### LINKED ISSUE #1826: github: start pinning actions versions
Let's start pinning github actions hashes in workflows like the specification already does (instead of just using tags like "v2"). Dependabot seems to be doing a decent job of alerting about updates and the updates do not seem awfully frequent.

jku:
This allows us to control when our workflows change.
Dependabot should now open PRs when the actions update.

This still leaves the actual OS image as a variable but Github does not
support pinning that: we'd have to start using our own containers (and
installing our own pythons, etc) to do that -- not worth the trouble.

Fixes #1826

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,10 +41,10 @@ jobs:
 
     steps:
       - name: Checkout TUF
-        uses: actions/checkout@v2
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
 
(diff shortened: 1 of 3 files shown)


joshuagl (approved):
Thank you! LGTM.

> This still leaves the actual OS image as a variable but Github does not
> support pinning that: we'd have to start using our own containers (and
> installing our own pythons, etc) to do that -- not worth the trouble.

Agreed.

PR #1832, opened 2022-02-08
Re-add setup.py to fix dependabot

### LINKED ISSUE #1828: Dependabot does not work
Several months ago we replaced setup.py with setup.cfg. This seems to have immediately broken dependabot, which has since then been complaining about missing setup.py -- it does not complain in an issue as you might have expected but in the easily discoverable location of _Insights -> Dependency graph -> Dependabot -> "requirements-dev.txt Last checked ..."_ 

I'm sending a PR of manual dependency updates soon but this issue is specifically for either fixing dependabot or jumping through whatever hoops we need to jump through to get dependabot working.

Related: https://github.com/dependabot/dependabot-core/issues/4483

jku:
This seems to fix the issue with dependabot (#1828) in my fork... There's also a unrelated trivial packaging fix included.

Also, reading the dependabot log I noticed there is one more dependency update I did not include in my latest dependency update PR (pycparser): I'll leave that unfixed for now, we can use it as confirmation that dependabot actually works.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,4 @@
+# This file exists to keep dependabot happy:
+# https://github.com/dependabot/dependabot-core/issues/4483
+from setuptools import setup
+setup()
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,7 +3,6 @@ include LICENSE*
 
 # Development and testing
 include *.txt
-global-include pylintrc
 include tox.ini
 graft tests
 


setup.py
```diff
@@ -0,0 +1,4 @@
+# This file exists to keep dependabot happy:
+# https://github.com/dependabot/dependabot-core/issues/4483
+from setuptools import setup
+setup()
```
lukpueh, line 4: Out of curiosity, does `setup()` read from setup.cfg?

lukpueh (commented):
Ready to un-draft?

setup.py
jku: It should, to my knowledge... but I want to still confirm this

jku: as far as I can tell this works fine: the wheels dist-info contains the correct METADATA.

(as a sidenote, we may be missing a license variable as metadata contains `License: UNKNOWN`, will file an issue)

lukpueh (approved):

PR #1840, opened 2022-02-08
Key.from_securesystemslib_key() raise ValueError

### LINKED ISSUE #1818: Metadata API: from_securesystemslib_key() should handle sslib error
from #1785

`from_securesystemslib_key()` should raise ValueError if securesystemslib.FormatError happens in format_keyval_to_metadata().

from_securesystemslib_key() is a helper so that repository code does not need to figure out how to safely get a Key from a securesystemslib private key. The securesystemslib call done there is a bit useless (we could just probe the dictionary ourselves) but since we call the function we should make sure to handle FormatError

MVrachev:
Fixes #1818 

If a `securesystemslib.FormatError` is raised inside
`Key.from_securesystemslib_key()` then reraise `ValueError`.
This is done so that our users don't have to import securesystemslib
in order to handle the error and because the securesystemslib error
itself is securesystemslib implementation-specific.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -379,6 +379,11 @@ class TestMetadata(unittest.TestCase):
         key = Key.from_securesystemslib_key(sslib_key)
         self.assertFalse("private" in key.keyval.keys())
 
+        # Test raising ValueError with non-existent keytype
+        sslib_key["keytype"] = "bad keytype"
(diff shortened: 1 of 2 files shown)


lukpueh (approved):
LGTM!

Issue #1806, opened 2022-01-28
tests: static test data generation


jku:
current state:
* most client tests use a repository that is generated live during the test (RepositorySimulator)
* there is (and should be) static test data as well, see tests/repository_data/
* test data was generated with scripts that used the legacy codebase but it's not clear if the scripts and the data are actually up to date (or if the data has been manually modified)
* scripts are now removed as legacy code is removed (#1790)

We will want to modify the static files at some point (e.g. we should make the files consistent_snapshot), so we should have _easy_ ability to do so. I believe that means we should 
* generate the static files with a script that uses metadata API
* possibly verify at test runtime that the script result is unchanged -- this would give us additional confirmation that we are not changing things (even superficially) in our serialisation format.

jku:
After seeing the 92% speedup :rocket:  in test runtime when we drop legacy tests: we could even expand these static test sets, we clearly have the test runtime budget now.

lukpueh:
Thanks for the effort, @jku! I meant to write a ticket with 16192902cef3cb669deb0e085c981341ced9c347 but then thought, let's cross the bridge when we get there, i.e. when we need to regenerate those fixtures. All the better when there's already a ticket.

jku:
Adding to this: Our serialization tests are pretty good, but they are very "unit testy". Making sure that the output of the whole deserialize-serialize pipeline does not change is not really easy -- and adding test data for complete Metadata objects really stretches the viability of the "test data within test code" approach.

So maybe we should have something like this:
* script that produces a bunch of metadata files (or maybe just bytes)
* test that ensures that deserializing and then serializing the data produces the same data
* test that the output of the script did not change unexpectedly:
  *  either the metadata files are stored in git (so humans can easily see what is tested) and the script just checks the produced content is still same
  * or the hashes of the metadata bytes are stored and verified: seeing the test data would require running the script manually


To be clear, this plan only solves the serialization testing issue -- but nothing prevents these same metadata files from also forming real repository versions.

MVrachev:
One question that comes to my mind is about the purpose of these generation scripts.
What priorities do we have for them:
1. simple easy to understand scripts that generate one set of new metadata sets
2. flexible scripts that can be used to generate a variety of metadata combinations
3. scripts that allow generating and publishing multiple metadata versions (but that does sound like something `RepositorySimulator` already does)

If 1 is the idea, then it’s clear.
If 2 is the focus here it’s good to think about what kind of metadata combinations do we want to generate.
[...]

PR #1860, opened 2022-02-12
Metadata test full serialization cycle

MVrachev:

Replace the usage of Metadata.to_dict inside
test_valid_metadata_serialization and instead use Metadata.to_bytes()
in order to test that the full serialization cycle is working as
expected:
Metadata.from_bytes -> Metadata.to_bytes

_Raised from Jussi's comment: https://github.com/theupdateframework/python-tuf/pull/1809#discussion_r804473334_

requirements**:

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -73,8 +73,15 @@ class TestSerialization(unittest.TestCase):
     @utils.run_sub_tests_with_dataset(valid_metadata)
     def test_valid_metadata_serialization(self, test_case_data: bytes) -> None:
         md = Metadata.from_bytes(test_case_data)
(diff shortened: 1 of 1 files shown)


jku (approved):
Neither option here is great IMO: we either don't test the full serialization cycle. or we end up hard coding the specific json options in the test (and then the test still won't notice if json module somehow changes how it works).

This isn't a complaint on the patch as such: I'm just saying we need to admit that this type of testing does not work well with complete metadata files and the complete serialization cycle: generating metadata (and verifying the results do not change) is a better approach for that: #1806 

Anyway, looks good to me, but let's not try to force a round peg into a square hole any more than this (unless there's some new _squaring-a-circle_ ideas of course)

PR #1867, opened 2022-02-15
build: pin test requirements for deterministic CI

### LINKED ISSUE #1813: build: let's start pinning dev requirements (?)
Issue #1811 is an example of build failing when it succeeded just moments before: A new black release was made that changes whitespace rules for a specific situation.

We may want to start pinning dev-requirements as well: this increases the amount of work (as we would start getting more dependabot PRs for these requirements) but it would mean our build would be less prone to spontaneously change. Whether we can pin the all transient build/test/release requirements is another question: there might be a lot of it?

@joshuagl opinions?

lukpueh:
Fixes #1813 (?, see discussion in issue)

Configures tox to use a pinned requirements file for deterministic
CI builds, i.e. our CI shouldn't start failing because of an
incompatible upstream release of any of our testing tools:

NOTE: pinned tuf runtime requirements were already were already
used for test builds before (included via `-r
requirements-pinned.txt` in 'requirements-test.txt'). Now they are
explicitly listed in 'requirements-test-pinnned.txt'.

'requirements-test-pinnned.txt' was generated semi-automatically by
running pip-compile over 'requirements-test.txt' for each
supported/tested Python version (see snippet below) and manually
merging the resulting per-Python version requirements files into
one, adding environment markers as needed.

```
for ver in 3.7.12 3.8.12 3.9.9 3.10.0; do
  pyenv virtualenv ${ver} tuf-env-${ver}
  pyenv activate tuf-env-${ver}
  python3 -m pip install -U pip pip-tools
  pip-compile --no-header --annotation-style line \
      -o requirements-test-pinned-${ver}.txt \
      requirements-test.txt
  pyenv deactivate
  pyenv uninstall -f tuf-env-${ver}
done
```

requirements**:

### DIFF
--- a/requirements-test-pinned.txt
+++ b/requirements-test-pinned.txt
@@ -1,0 +1,41 @@
+astroid==2.9.3            # via pylint
+bandit==1.7.2             # via -r requirements-test.txt
+black==22.1.0             # via -r requirements-test.txt
+certifi==2021.10.8        # via -r requirements-pinned.txt, requests
+cffi==1.15.0              # via -r requirements-pinned.txt, cryptography, pynacl
+charset-normalizer==2.0.12  # via -r requirements-pinned.txt, requests
+click==8.0.3              # via black
+coverage==6.3.1           # via -r requirements-test.txt
+cryptography==36.0.1      # via -r requirements-pinned.txt, securesystemslib
+gitdb==4.0.9              # via gitpython
+gitpython==3.1.26         # via bandit
+idna==3.3                 # via -r requirements-pinned.txt, requests
+importlib-metadata==4.11.1; python_version < '3.8'  # via click, stevedore
+isort==5.10.1             # via -r requirements-test.txt, pylint
+lazy-object-proxy==1.7.1  # via astroid
+mccabe==0.6.1             # via pylint
+mypy==0.931               # via -r requirements-test.txt
+mypy-extensions==0.4.3    # via black, mypy
+pathspec==0.9.0           # via black
+pbr==5.8.1                # via stevedore
+platformdirs==2.5.0       # via black, pylint
+pycparser==2.21           # via -r requirements-pinned.txt, cffi
+pylint==2.12.2            # via -r requirements-test.txt
+pynacl==1.5.0             # via -r requirements-pinned.txt, securesystemslib
+pyyaml==6.0               # via bandit
+requests==2.27.1          # via -r requirements-pinned.txt
+securesystemslib[crypto,pynacl]==0.22.0  # via -r requirements-pinned.txt
+smmap==5.0.0              # via gitdb
+stevedore==3.5.0          # via bandit
+toml==0.10.2              # via pylint
+tomli==2.0.1              # via black, mypy
+typed-ast==1.5.2; python_version < '3.8'  # via astroid, black, mypy
+types-requests==2.27.9    # via -r requirements-test.txt
+types-urllib3==1.26.9     # via types-requests
+typing-extensions==4.1.1  # via astroid, black, gitpython, importlib-metadata, mypy, pylint
+urllib3==1.26.8           # via -r requirements-pinned.txt, requests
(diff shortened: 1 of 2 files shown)


jku:
I think the core idea is correct. We want tests to be deterministic.

The only worry I have here is that requirements-test-pinned.txt is a lot of data that is not maintained in an automated fashion
* Updating versions for dependencies with `python_version` is likely a manual process
* updating the list for missing or unused dependencies is also manual

and this manual process does not seem very simple: installing half a dozen python versions, a tool to manage them and understanding the process are just the prerequisites -- you still have to generate the requirements files and understand how to merge them.

So I wonder if we're setting the bar a little high for ourselves and future maintainers?

---
Would it be easier to automate the process if we actually stored the `requirements-test-pinned-X.Y.txt` files (and used them in tests)?

MVrachev:
What about pinning only certain packages like our linters?
Is it possible and feasable?

jku:
> What about pinning only certain packages like our linters? Is it possible and feasable?

It's something to consider, yes: I expect most changes in determinism by far are going to be from the tool updates themselves (like black changing whitespace rules).

lukpueh:
> The only worry I have here is that requirements-test-pinned.txt is a lot of data that is not maintained in an automated fashion

Agreed, although the problem already exits for `requirements.txt->requirements-pinned.txt`. Doing the same for here does not really create a new problem. It might even give us more reason to fix the existing one ticketized in #1249.

> Would it be easier to automate the process if we actually stored the `requirements-test-pinned-X.Y.txt` files (and used them in tests)?

Probably.  But I suggest we think about that when working on #1249. Until then I'd rather use env markers than adding 4 new requirements files. 

Of course I'm also fine with someone working on  #1249 right away and not merge this suboptimal temporary solution.

lukpueh:
> What about pinning only certain packages like our linters? Is it possible and feasable?

If by "certain packages" you mean only direct or a subset of direct dependencies, then yes, we can pin those, dependabot auto-bumps versions, and removing/adding them is a manual process anyway. But if you mean a subset of direct dependencies plus the corresponding transitive dependencies, then it only makes automatic maintenance harder.

jku (approved):
I suppose we can try this?

I think long term maintaining this manually is not sustainable but if we plan to make #1249 a priority then we could merge this and see how it works in practice

lukpueh:
> I think long term maintaining this manually is not sustainable but if we plan to make #1249 a priority then we could merge this and see how it works in practice

Added `backlog` label to #1249, to fix this in a more sustainable way. Merging this tempfix in the meanwhile...

jku:
This does not seem to quite work as I thought it would (see current dependabot PRs). I started a discussion on slack but copying here as well:

since we are installing applications not libraries, those applications have actually pinned their dependencies. So if we pin these deps-of-deps too, it seems dependabot thinks we want to update them and files loads of updates?

---

EDIT: 

No, these apps generally do not pin dep versions.

if `pip-compile` (somehow) finds a requirements file that already pins versions that work for the requirements, it won't update the version. You need to delete the pinned files first or use `--upgrade` . At least this seems to explain 5/6 of the updates dependabot wants to do.

The last one, mccabe is not updated to 0.7 because pylint is explicitly asking for <0.7

lukpueh:
> if pip-compile (somehow) finds a requirements file that already pins versions that work for the requirements, it won't update the version.

Do you know which requirements file that was in this case? Does it consider files in dependencies?

jku:
so I was running this to see what happens:
`pip-compile --no-header --annotation-style line -o requirements-test-pinned.txt requirements-test.txt`

I don't know if it found requirements-test-pinned.txt because it was specified as the output file or just by guesswork but it clearly found that file and used the version numbers from there and did not update them if they fulfilled the requirements: so the new file _did not have the newest versions available_.

If I deleted the output file first, it does update the the versions. Likewise if I use `--upgrade`

PR #1872, opened 2022-02-18
Debian patches for debian branch

rzr:
cc: @joshuagl feel free to merge before releasing on latest release and keep the work on this base , to align debian packaging team


debian/control
```diff
@@ -1,19 +1,24 @@
...
+               python3-dateutil,
+               python3-nacl,
                python3-requests,
-               python3-securesystemslib
-Standards-Version: 4.5.0
-Vcs-Browser: https://github.com/theupdateframework/tuf
-Vcs-Git: https://github.com/theupdateframework/tuf.git
+               python3-securesystemslib,
+               tox
+Standards-Version: 4.5.1
```
eamanu, line 16: could be updated to 4.6.0.1

debian/control
```diff
@@ -1,19 +1,24 @@
...
+Rules-Requires-Root: no
+Vcs-Browser: https://salsa.debian.org/python-team/packages/tuf/-/tree/debian/master
```
eamanu, line 18: I recommend use `https://salsa.debian.org/python-team/packages/tuf` because here vcs-browser is pointing to a particular branch, and if that change in the future, that link would be incorrect.

debian/tests/testsuite
```diff
@@ -0,0 +1,17 @@
...
+set -efu
+set -x
+
+pys="$(py3versions -s 2> /dev/null)"
+
+cp -rva tests "$AUTOPKGTEST_TMP/"
+cd "$AUTOPKGTEST_TMP/tests"
+
+# Resolve "from tests import utils"
+PYTHONPATH=".."
```
eamanu, line 11: the purpose of autopkgtest is tests the packages as is installed. So, please remove PYTHONPATH.

eamanu (changes requested):
Hi, @rzr thanks for the work. I left some review, please consider apply it. Also move that to the debian package in salsa.

rzr:
Please add those changes in your branch to preserve history it will be easier to maintain debian's side too

lukpueh:
Thanks for your efforts, @rzr! It does not look like upstream maintainers currently want to allocate time to package python-tuf for Debian. So I suggest to maintain Debian metadata and apply the patches of this PR in a downstream fork/mirror of python-tuf. Closing here...

Issue #1862, opened 2022-02-14
example-client: better error handling


jku:
The example client has two conflicting goals:
* provide "production-like" error messages: don't dump a 300 line stack trace if the network is down 
* be a useful debugging tool

For an example of the second I modified the example client to use some real world repositories: the error messages I got were mostly useless -- I needed to remove the error handling in the example app to see what the errors were.

So I suggest a couple of things:
* DeserializationError needs an actual error message when it's raised -- failure to deserialize local root metadata currently looks like following, note the missing error message on last line (this can be reproduced by just modifying the local root.json file so it is invalid). _This should now be fixed with #1876_
  ```
  $ ./client_example.py  download x
  Found trusted root in /home/jku/.local/share/python-tuf-client-example
  Failed to download target x: 
  ```
* Maybe there should be a way to see the actual stack trace, and that should be documented in the error message. So a hypothetical normal failure would look like:
  ```
  $ ./client_example.py download x
  Found trusted root in /home/jku/.local/share/python-tuf-client-example
  Failed to download target x: Failed to transmogrify metadata
    (For detailed traceback use "-v")
  ```
  and then using -v would `raise e` instead of handling it? This would be a small change in client_example itself

ameypd:
I would like to take this up as my first PR.

ameypd:
@jku 
I have set up the environment, please tell me how to proceed with this issue.

jku:
you could create a failure situation to experiment with: e.g. don't start the local webserver like the instructions tell you (so you get the one line error message when running `./client_example.py download file1.txt`). Then add a command line option that makes it so that you get a full stack trace when running e.g. `./client_example.py -v download file1.txt`. That would be useful for debugging

ameypd:
@jku 
![image](https://user-images.githubusercontent.com/113001748/201942808-cfa5871e-b4d3-49fd-b507-7b8cb215de3f.png)
Is this expected behaviour?

jku:
That looks like what I'd expect yeah.

What I'd like to have is an optional way to see the stack trace for the "failed to download...."  error -- I think this is just a matter of re-raising the error instead of handling it with the simple error message in the example code.

ameypd:
`    client_args.add_argument(
        "-v",
        "--verbose",
        help="Output verbosity level (-v, -vv, ...)",
        action="count",
        default=2,
    )`
I changed the default to 2 from 0, but that is not the way to do it right?

lukpueh:
[...]

PR #1876, opened 2022-02-21
Logging and error message improvements

### LINKED ISSUE #1875: Provide more information when Key.verify_signature() fails
**Current behavior**:
If we call `Metadata.verify_delegate()` and one of the keys fails to be verified the message they got is this:
https://github.com/theupdateframework/python-tuf/blob/fc9b42fa5d4e887821189055b9f44fa0f04f0885/tuf/api/metadata.py#L413
No logging is provided about why the key has failed verification.

**Expected behavior**:
Adding a log about the last exception message could be useful to provide context for what exactly happened.
This could be useful to identify if a keyvalue is malformed.

jku:
* Improve logging when signature verification fails for unexpected reasons (like key type or key content that securesystemslib can't read)
* Improve error message when deserialization fails

### DIFF
--- a/tuf/api/serialization/json.py
+++ b/tuf/api/serialization/json.py
@@ -36,7 +36,7 @@ class JSONDeserializer(MetadataDeserializer):
             metadata_obj = Metadata.from_dict(json_dict)
 
         except Exception as e:
-            raise DeserializationError from e
+            raise DeserializationError("Failed to deserialize JSON") from e
 
         return metadata_obj
 
@@ -66,7 +66,7 @@ class JSONSerializer(MetadataSerializer):
             ).encode("utf-8")
 
         except Exception as e:
-            raise SerializationError from e
+            raise SerializationError("Failed to serialize JSON") from e
 
         return json_bytes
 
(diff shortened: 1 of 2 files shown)


tuf/api/serialization/json.py
```diff
@@ -36,7 +36,7 @@ def deserialize(self, raw_data: bytes) -> Metadata:
             metadata_obj = Metadata.from_dict(json_dict)
 
         except Exception as e:
-            raise DeserializationError from e
+            raise DeserializationError("Failed to deserialize JSON") from e
```
lukpueh, line 39: Would it make sense to include the original error message? E.g.:
```suggestion
            raise DeserializationError(f"Failed to deserialize JSON: {e}") from e
```

Same comment applies below.

jku: The few cases that I tried did not end up very useful: the errors can be really varied and are often not informative without more context. This context is IMO only available with the stack trace and source code.

As an example if you remove "scheme" from a key, you would get this is in client example with your suggestion:
`Failed to deserialize JSON: 'scheme'`
there is context there (we're getting a KeyError while parsing the json dict) but this is not understandable from that message, only from the stack trace...

jku: I suggested a related improvement to the client example in  #1862 (optionally show the stack trace, so we usually get something human readable but can do developer readable output as well).

jku:
@lukpueh ping, do you dislike the error message or are we good to go with this one?

lukpueh (approved):

PR #1886, opened 2022-02-24
Add development blog

### LINKED ISSUE #1884: project blog: setup
It would be nice to have a place to publish project related things that are not quite documentation but something we want to be able to link to (maybe workflow examples or updates on PyPI work, etc ). This sounds like a blog so let's investigate how to have one.

I think the path of least resistance might be Github Pages right here in the python-tuf repo. Other options include reviving the 
theupdateframework.github.io (currently archived project in theupdateframework organization) or a third party blog platform.

Issues to investigate:
* GitHub Pages have a weird configuration: you can set the Pages root folder... but it looks like the UI only allows `/` and `/docs`: these might not work for us

jku:
I'm pushing this out for comments as it seems to be the best I can do:
* this should be working config for a project blog
* I added our 1.0 announcement as seed content (to check that it works)
* it's a bit ugly with two more files in `docs/` but I really don't see another way that's as easy as this.
* This requires an org admin to enable Github Pages for the project and to set `docs/` as the root: nothing else needs to be configured in UI
* I did not add links to the blog anywhere: I figured we can try this out as "soft launch" and if it looks good can add links e.g. from the README when we get content...

I believe the URL is going to be https://theupdateframework.github.io/python-tuf

I've tested this at https://github.com/jku/github-pages-with-jekyll

Fixes #1884


lukpueh:
> I'm pushing this out for comments as it seems to be the best I can do:

> 
> * this should be working config for a project blog
> * I added our 1.0 announcement as seed content (to check that it works)
> * it's a bit ugly with two more files in `docs/` but I really don't see another way that's as easy as this.
> * This requires an org admin to enable Github Pages for the project and to set `docs/` as the root: nothing else needs to be configured in UI
> * I did not add links to the blog anywhere: I figured we can try this out as "soft launch" and if it looks good can add links e.g. from the README when we get content...
> 
> I believe the URL is going to be https://theupdateframework.github.io/python-tuf
> 
> I've tested this at https://github.com/jku/github-pages-with-jekyll
> 
> Fixes #1884

lukpueh:
> * This requires an org admin to enable Github Pages for the project and to set `docs/` as the root: nothing else needs to be configured in UI

Done!

docs/_posts/2022-02-21-release-1-0-0.md
```diff
@@ -0,0 +1,47 @@
...
+title: "Python-TUF reaches version 1.0.0"
+author: Jussi Kukkonen and Lukas Pühringer
+---
+
+The Python-TUF community is proud to announce the release of Python-TUF 1.0.0.
+The release, which is available on [PyPI](https://pypi.org/project/tuf/) and
+[GitHub](https://github.com/theupdateframework/python-tuf/), introduces new
+stable and more ergonomic APIs.
+
+<img align="right" src="../../../tuf-icon-color.png" width="200"/>
```
lukpueh: Isn't this
```suggestion
<img align="right" src="../tuf-icon-color.png" width="200"/>
```
?

Also  https://github.com/theupdateframework/python-tuf/blob/develop/docs/tuf-icon-color.png is in favicon resolution, so I guess that's a bit small.

jku: oops, favicon was not the intention.

The path I used _should_ work, although it might not be the best way to handle it: The issue with asset paths is that they are for the final URL (jekyll does not mangle them) so we either need to use absolute paths, which means `/python-tuf/FILENAME` or we use relative paths which is what I did: the post path is `/python-tuf/2022/02/21/POST-TITLE.html` so we need to go up three levels to `../../../FILENAME`.

There may be a jekyll config that allows defining a the asset lookup path but I've not found one yet

lukpueh: > oops, favicon was not the intention.

We could copy [the image I used for the ssl blog](https://github.com/lukpueh/ssl-site/blob/bb0b59634c2807c52d926bce3e0783c400d56664/img/blog/tuf-icon.png), it has an appropriate resolution. FYI: The original is taken from [cncf/artwork#tuf-icon-color.png](https://github.com/cncf/artwork/blob/master/projects/tuf/icon/color/tuf-icon-color.png).
> 
> The path I used _should_ work, although it might not be the best way to handle it: The issue with asset paths is that they are for the final URL (jekyll does not mangle them) so we either need to use absolute paths, which means `/python-tuf/FILENAME` or we use relative paths which is what I did: the post path is `/python-tuf/2022/02/21/POST-TITLE.html` so we need to go up three levels to `../../../FILENAME`.
> 
> There may be a jekyll config that allows defining a the asset lookup path but I've not found one yet

Thanks for the explanation!

jku:
I added the icon lukas suggested and renamed the existing one for clarity.

lukpueh (approved):

PR #1887, opened 2022-02-28
docs: Clean up blog header

jku:
Minima theme for Github Pages by default adds all files in blog root (docs/) as links in
the header. This looks ridiculous in our case: let's just have a link to
blog front page.

---

Tested this fix in https://github.com/jku/github-pages-with-jekyll / https://jku.github.io/github-pages-with-jekyll/

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -9,5 +9,8 @@ github_username:  theupdateframework
 
 show_excerpts: true # set to false to remove excerpts on the homepage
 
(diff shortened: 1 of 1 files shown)


jku:
FYI the blog is up at https://theupdateframework.github.io/python-tuf/ (it's not linked from anywhere since it's still a bit half baked like this)

lukpueh (approved):
Thanks for the quick fix!

PR #1895, opened 2022-03-03
Single source version number

jku:
As of setuptools 46.4.0, one can accomplish single source version number with

    version = attr: package.__version__

in setup.cfg: As long as setuptools simplified AST parser is able to
read the file, this works without actually importing the module.

---

Fix the release documentation on this aspect, and make some minor improvements to the doc.

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -5,14 +5,18 @@
   tracked by git
 * Ensure `docs/CHANGELOG.md` contains a one-line summary of each [notable
   change](https://keepachangelog.com/) since the prior release
-* Update `setup.cfg` and `tuf/__init__.py` to the new version number vA.B.C
+* Update `tuf/__init__.py` to the new version number "A.B.C"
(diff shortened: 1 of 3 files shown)


joshuagl (approved):
Love to see cleanup like this, thank you!

docs/RELEASE.md
```diff
@@ -24,11 +28,10 @@
   (ensure commands invoking `python` below are using Python 3)
   * Remove existing dist build dirs
   * Create source dist and wheel `python3 -m build`
-  * Sign source dist `gpg --detach-sign -a dist/tuf-vA.B.C.tar.gz`
-  * Sign wheel `gpg --detach-sign -a dist/tuf-vA.B.C-py3-none-any.whl`
-  * Upload to test PyPI `twine upload --repository testpypi dist/*`
-  * Verify the uploaded package https://testpypi.python.org/pypi/tuf/
```
jku, line 30: Forgot to mention this in comment: this was removed because normal maintainers cannot upload same package twice so this is impossible to do if the previous steps were done

Issue #1269, opened 2021-01-28
Verifiably reproducible build artefacts


joshuagl:
**Description of issue or feature request**:

We can give users of our release artefacts (tarballs and wheels) greater confidence in the integrity of the artefacts and our development processes if they can verify that the artefacts we produced correspond to the signed tagged source code for the release.

We can achieve this through implementing [reproducible builds](https://reproducible-builds.org).

**Current behavior**:

Tarball (sdist) and wheels (bdist_wheel) generated for a release __are not__ verifiably reproducible.

> $ diff dist-a dist-b
> Binary files dist-a/tuf-0.16.0-py2.py3-none-any.whl and dist-b/tuf-0.16.0-py2.py3-none-any.whl differ
> Binary files dist-a/tuf-0.16.0.tar.gz and dist-b/tuf-0.16.0.tar.gz differ

**Expected behavior**:

Tarball (sdist) and wheels (bdist_wheel) generated for a release __are__ verifiably reproducible.

joshuagl:
Ensuring [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) is set (i.e. [this patch](https://github.com/joshuagl/tuf/commit/d116795becbd7f0a805c59a1b137ebab501e5877)) enables us to create verifiably reproducible wheels, but the tarball (sdist) is still not verifiably reproducible.

> $ diff dist-r-a dist-r-b
> Binary files dist-r-a/tuf-0.16.0.tar.gz and dist-b-r/tuf-0.16.0.tar.gz differ

joshuagl:
**Note**: I tried this on macOS and Fedora 33. I don't think the sdist being non-deterministic is host tool related.

mnm678:
tar isn't deterministic by default, do they still differ when generated with [`--mtime`](https://wiki.debian.org/ReproducibleBuilds/TimestampsInTarball)?

joshuagl:
Quite right, thanks for the link! Our sdist tarballs are generated by setuptools, which (so far as I could tell in the relatively brief time I spent looking today) has no option to specify mtime. 

It's possible we just have to brute force this; unpack and re-pack the tarball with --mtime, use [`strip-nondeterminism`](https://salsa.debian.org/reproducible-builds/strip-nondeterminism), or similar.

I intend to spend some more time on this in the next couple of weeks, unless someone gets to it first.

sethmlarson:
[...]

PR #1896, opened 2022-03-04
Update package metadata

ofek:

## Background

Hello there! This was requested in https://github.com/in-toto/in-toto/pull/490#issuecomment-1057923077

The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://www.python.org/dev/peps/pep-0517/)/[PEP 660](https://www.python.org/dev/peps/pep-0660/)) and the format for metadata declaration ([PEP 621](https://www.python.org/dev/peps/pep-0621/)/[PEP 631](https://www.python.org/dev/peps/pep-0631/)). As a result, the use of `setup.py` files is now [heavily discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

## Summary of changes

This implements [PEP 621](https://www.python.org/dev/peps/pep-0621/), obviating the need for `setup.cfg` and `MANIFEST.in`. Support has not landed in `setuptools`, so builds will now use `hatchling`. It is quite stable and actively maintained, the only reason the version is 0.x is because 1.0.0 will drop support for Python 2. It's also on the major distribution channels such as [conda-forge](https://anaconda.org/conda-forge/hatchling), [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), etc.

I've done this for such projects as [pipx](https://github.com/pypa/pipx/blob/2ecc668acf472ad6956cc682499c077c1130d17e/pyproject.toml), all [Datadog Agent integrations](https://github.com/DataDog/integrations-core/pulls?q=is%3Apr+author%3Aofek+in%3Atitle+Add+pyproject.toml+file), etc.

## Notes

- The empty `setup.py` remains for Dependabot


jku:
before you spend more time on implementation details, let's discuss if this is something we want. I mentioned this in the in-toto issue:

> We ... would likely need good reasons to move to a less known build tool: I'm not trying to belittle hatchling, I'm just saying we try to be careful about pipeline security. Getting rid of MANIFEST.in might not be a good enough reason to make a major build tool change.

You provided a comprehensive [comment](https://github.com/in-toto/in-toto/pull/490#issuecomment-1058128344) that explains really well why you are building your project: I am convinced new build tools are a good idea. I'm not convinced there are enough advantages for python-tuf to migrate:
* our build is already declarative so improvements in that area seem really minor (do correct if this is wrong)
* hatchling is seemingly a two months old project and has one developer

I want to support cool python infrastructure work but I also take build pipeline choices seriously.

ofek:
> our build is already declarative so improvements in that area seem really minor

Declarative yes, but not in the standard format, and therefore no support for new standards like [licensing](https://www.python.org/dev/peps/pep-0639/).

> hatchling is seemingly a two months old project

Since we use it at Datadog for shipping business critical things, our CI would be the first to yell at me if something goes awry 😄

We also were the first to use tuf + in-toto at scale, which I assume has contributed to their current stability.

> and has one developer

As mentioned in the comment you linked, `setuptools` is almost exclusively maintained by one person and the codebase is an order of magnitude larger.

-----

I rebased; would you mind triggering the CI with that button below again?

joshuagl:
Thanks for the proposal @ofek. 

For context: our most recent round of changes to package metadata were to bring it inline with what is recommended in the [Python Packaging User Guide](https://packaging.python.org/en/latest/), specifically [Packaging and distributing projects](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/). That tutorial does suggest considering alternatives, but does not motivate _why_ one should. I appreciate your related remarks here and in the linked in-toto issue comparing alternatives. We did briefly look at flit in the past, but ultimately didn't feel the need to switch away from setuptools yet. One reason to consider alternatives is our desire to have verifiably reproducible build artefacts (#1269).

Aside from discussion of switching packaging tool, which I'll defer to more active maintainers on, there are some good unrelated fixes in this PR (tox isolated builds, gitignore fixes, redundant comment about version in tuf/__init__.py). @ofek would you mind if I pull those out into separate patches and credit you with `Co-authored-by:` in the commit message? See: https://github.com/joshuagl/tuf/commits/joshuagl/build-nits

ofek:
> would you mind if I pull those out into separate patches

Sure!

> One reason to consider alternatives is our desire to have verifiably reproducible build artefacts (https://github.com/theupdateframework/python-tuf/issues/1269).

Hatchling does that by default https://ofek.dev/hatch/latest/config/build/#reproducible-builds, including everything mentioned in that linked issue like file ordering

jku:
> Hatchling does that by default https://ofek.dev/hatch/latest/config/build/#reproducible-builds, including everything mentioned in that linked issue like file ordering

Now this is interesting and a reason to consider this, thanks.

joshuagl:
> Hatchling does that by default https://ofek.dev/hatch/latest/config/build/#reproducible-builds, including everything mentioned in that linked issue like file ordering

A quick glance at the source code tells me that includes producing reproducible sdist. Nice!

ofek:
I updated the branch.

> A quick glance at the source code tells me that includes producing reproducible sdist. Nice!

Yup you can test here by doing:

```
$ docker run --rm -v $PWD:/home/proj -w /home/proj python bash -c "pip install -qq build;python -m build -o /tmp/dist;python -c \"import hashlib,os,pathlib;print(os.linesep.join(hashlib.sha256(p.read_bytes()).hexdigest() for p in pathlib.Path('/tmp/dist').iterdir()))\""
...
a467c36feab72b9f1464b446eb6ec87bc2eeff40861c6c088c30f812900f7e6e
3e548ba8c530d6ff2b04737a24ae42073361927560e30c95aa4ae3f7d3bc4ab0
$ docker run --rm -v $PWD:/home/proj -w /home/proj python:alpine ash -c "pip install -qq build;python -m build -o /tmp/dist;python -c \"import hashlib,os,pathlib;print(os.linesep.join(hashlib.sha256(p.read_bytes()).hexdigest() for p in pathlib.Path('/tmp/dist').iterdir()))\""
...
a467c36feab72b9f1464b446eb6ec87bc2eeff40861c6c088c30f812900f7e6e
3e548ba8c530d6ff2b04737a24ae42073361927560e30c95aa4ae3f7d3bc4ab0
```

lukpueh:
Thanks for your efforts, @ofek! Hatch/hatchling looks really cool and getting reproducible builds for free is quite a compelling argument.

I admit that I share the reservations @jku and @joshuagl have expressed, wrt preferring a more "established" tool. But this patch is simple enough to revert (or migrate to another PEP 621 compatible build system) if needed, which makes the bus factor of your project less worrisome. And the integrations you mentioned along with the lightweightness of the hatchling builders, make me confident that issues are detected and probably also fixed swiftly.

So I lean towards giving this a try.

jku:
Agreed, potentially reproducible builds is a game changer, something I'm willing to take a bit of risk for. 

Merging this doesn't get us quite to reproducible builds (doesn't pin the hatch/hatchling version we use for official builds for a start) but getting that right does not need to happen in this PR.

pyproject.toml
```diff
@@ -1,7 +1,83 @@
...
+]
+# Keep in sync with /docs/MAINTAINERS.txt
+maintainers = [
+  { name = "Justin Cappos", email = "jcappos@nyu.edu" },
+  { name = "Marina Moore", email = "mm9693@nyu.edu" },
+  { name = "Trishank Karthik Kuppusamy", email = "trishank@nyu.edu" },
+  { name = "Lukas Puehringer", email = "lukas.puehringer@nyu.edu" },
+  { name = "Joshua Lock", email = "jlock@vmware.com" },
+  { name = "Jussi Kukkonen", email = "jkukkonen@vmware.com" },
+]
```
jku: I think this is unneeded:
* it's more maintenance work to keep updated
* I believe this will result in pypi showing all of our contact info in the project home page, I don't see that being useful: we don't really want anyone contacting _only_ maintainers (except for security issues but that process is documented in README).

We already have theupdateframework@googlegroups.com listed as "author", this sounds fine to me as it gives an email contact point.

pyproject.toml
```diff
@@ -1,7 +1,83 @@
...
+  "requests>=2.19.1",
+  "securesystemslib>=0.22.0",
+]
+dynamic = ["version"]
+
+[project.urls]
+Documentation = "https://theupdateframework.readthedocs.io/en/stable/"
+Homepage = "https://www.updateframework.com"
+Issues = "https://github.com/theupdateframework/python-tuf/issues"
+Source = "https://github.com/in-toto/in-toto"
```
jku:
```suggestion
Source = "https://github.com/theupdateframework/python-tuf"
```

pyproject.toml
```diff
@@ -1,7 +1,83 @@
...
+path = "tuf/__init__.py"
+
+[tool.hatch.build.targets.sdist]
+include = [
+  "/docs",
+  "/tests",
+  "/tuf",
+  "/requirements*.txt",
+  "/tox.ini",
+]
```
jku:
```suggestion
[tool.hatch.build.targets.sdist]
include = [
  "/docs",
  "/tests",
  "/tuf",
  "/requirements*.txt",
  "/tox.ini",
  "/setup.py",
]
```

pyproject.toml
```diff
@@ -1,7 +1,83 @@
...
 [build-system]
-requires = ["setuptools>=46.4.0", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling>=0.18.0"]
+build-backend = "hatchling.build"
+
+[project]
+name = "tuf"
+description = "A secure updater framework for Python"
+readme = "README.md"
```
jku, line 9: does this lead to pypi using the README as project long description?

jku (commented):
Looks good overall: left a question and a few specific comments in source.
* Are the readthedocs.yaml changes unrelated? 
* If you could rebase  (so we don't end up with merge commits in this branch), that would be nice

jku:
Actually one more question: since we have the setup.py workaround (for dependabot), how does it even work?
```
from setuptools import setup
setup()
```

...but we no longer depend on setuptools?


Or is it so that we _know_ that nothing in the build/install pipeline ever runs setup.py -- it just needs to exist for dependabot?

pyproject.toml
ofek: yes

ofek:
Addressed and rebased!

> Are the readthedocs.yaml changes unrelated?

Yes because by default it does `python setup.py install`

> since we have the setup.py workaround (for dependabot), how does it even work? [..] it just needs to exist for dependabot?

Yes it's just for dependabot for now https://github.com/dependabot/dependabot-core/issues/3290

jku (approved):
I can't think of anything else to add:
* I suppose we could make the dummy setup.py error out (to make sure no-one is running it by accident)... but that does not have to be in this PR
* thinking the reproducible build aspect through (e.g. do we want to pin the build tools) is important but again not required for this PR

Thanks for pushing the issue here ofek, I'm looking forward to trying out build reproducibility.

Approving. I wouldn't mind if another maintainer wants to have a look at the build results before merging.

lukpueh (approved):

PR #1898, opened 2022-03-08
Tests: provide a way to generate a simple metadata set

MVrachev:
Related to: #1806

I created a new script called "generate_md.py" which can be used
to easily generate a repository. Additionally, I created a new
test file making sure that the locally stored metadata files and
the newly generated metadata roles are the same.
This will allow us to test that we are not changing the metadata
file structure when making changes.

requirements**:

### DIFF
--- a/tests/generated_data/generate_md.py
+++ b/tests/generated_data/generate_md.py
@@ -1,0 +1,140 @@
+"""Script for generating new metadata files."""
+
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+import copy
+import os
+import sys
+from datetime import datetime
+from typing import Dict, List, Optional
+
+from securesystemslib.signer import SSlibSigner
+
+from tests import utils
+from tuf.api.metadata import (
+    SPECIFICATION_VERSION,
+    TOP_LEVEL_ROLE_NAMES,
+    Key,
+    Metadata,
+    MetaFile,
+    Role,
+    Root,
+    Snapshot,
+    Targets,
+    Timestamp,
+)
+from tuf.api.serialization.json import JSONSerializer
+
+# Hardcode keys and expiry time to achieve reproducibility.
+public_values: List[str] = [
+    "b11d2ff132c033a657318c74c39526476c56de7556c776f11070842dbc4ac14c",
+    "250f9ae3d1d3d5c419a73cfb4a470c01de1d5d3d61a3825416b5f5d6b88f4a30",
+    "82380623abb9666d4bf274b1a02577469445a972e5650d270101faa5107b19c8",
+    "0e6738fc1ac6fb4de680b4be99ecbcd99b030f3963f291277eef67bb9bd123e9",
+]
+private_values: List[str] = [
+    "510e5e04d7a364af850533856eacdf65d30cc0f8803ecd5fdc0acc56ca2aa91c",
+    "e6645b00312c8a257782e3e61e85bafda4317ad072c52251ef933d480c387abd",
+    "cd13dd2180334b24c19b32aaf27f7e375a614d7ba0777220d5c2290bb2f9b868",
+    "7e2e751145d1b22f6e40d4ba2aa47158207acfd3c003f1cbd5a08141dfc22a15",
+]
+keyids: List[str] = [
+    "5822582e7072996c1eef1cec24b61115d364987faa486659fe3d3dce8dae2aba",
+    "09d440e3725cec247dcb8703b646a87dd2a4d75343e8095c036c32795eefe3b9",
+    "3458204ed467519c19a5316eb278b5608472a1bbf15850ebfb462d5315e4f86d",
+    "2be5c21e3614f9f178fb49c4a34d0c18ffac30abd14ced917c60a52c8d8094b7",
+]
+
+keys: Dict[str, Key] = {}
+for i in range(4):
+    keys[f"ed25519_{i}"] = Key.from_securesystemslib_key(
+        {
+            "keytype": "ed25519",
+            "scheme": "ed25519",
+            "keyid": keyids[i],
+            "keyval": {
+                "public": public_values[i],
+                "private": private_values[i],
+            },
+        }
+    )
+
+expires_str = "2050-01-01T00:00:00Z"
+EXPIRY = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+SPEC_VERSION = ".".join(SPECIFICATION_VERSION)
+OUT_DIR = "generated_data/ed25519_metadata"
+if not os.path.exists(OUT_DIR):
+    os.mkdir(OUT_DIR)
+
+SERIALIZER = JSONSerializer()
+
+
+def verify_generation(md: Metadata, path: str) -> None:
+    """Verify that newly generated file equals the locally stored one.
+
+    Args:
+        md: Newly generated metadata object.
+        path: Path to the locally stored metadata file.
+    """
+    with open(path, "rb") as f:
+        static_md_bytes = f.read()
+        md_bytes = md.to_bytes(SERIALIZER)
+        if static_md_bytes != md_bytes:
+            raise ValueError(
+                f"Generated data != local data at {path}. Generate a new "
+                + "metadata with 'python generated_data/generate_md.py'"
+            )
+
+
+def generate_all_files(
+    dump: Optional[bool] = False, verify: Optional[bool] = False
+) -> None:
+    """Generate a new repository and optionally verify it.
+
+    Args:
+        dump: Wheter to dump the newly generated files.
+        verify: Whether to verify the newly generated files with the
+            local staored.
+    """
+    roles = {role_name: Role([], 1) for role_name in TOP_LEVEL_ROLE_NAMES}
+    root = Root(1, SPEC_VERSION, EXPIRY, {}, roles, True)
+    root.add_key("root", keys["ed25519_0"])
+    root.add_key("timestamp", keys["ed25519_1"])
+    root.add_key("snapshot", keys["ed25519_2"])
+    root.add_key("targets", keys["ed25519_3"])
+
+    md_root: Metadata[Root] = Metadata(root, {})
+
+    timestamp = Timestamp(1, SPEC_VERSION, EXPIRY, MetaFile(1))
+    md_timestamp: Metadata[Timestamp] = Metadata(timestamp, {})
+
+    meta: Dict[str, MetaFile] = {"targets.json": MetaFile(1)}
+    snapshot = Snapshot(1, SPEC_VERSION, EXPIRY, meta)
+    md_snapshot: Metadata[Snapshot] = Metadata(snapshot, {})
+
+    targets = Targets(1, SPEC_VERSION, EXPIRY, {})
+    md_targets: Metadata[Targets] = Metadata(targets, {})
+
+    i = 0
+    for md in [md_root, md_timestamp, md_snapshot, md_targets]:
+        assert isinstance(md, Metadata)
+        key_info = copy.deepcopy((keys[f"ed25519_{i}"].to_securesystemslib_key()))
+        key_info["keyval"]["private"] = private_values[0]
+        signer = SSlibSigner(key_info)
+        md.sign(signer)
+        path = os.path.join(OUT_DIR, f"{md.signed.type}_with_ed25519.json")
+        if verify:
+            verify_generation(md, path)
+
+        if dump:
+            md.to_file(path, SERIALIZER)
+
+        i += 1
+
+
+if __name__ == "__main__":
+    utils.configure_test_logging(sys.argv)
+    # To generate a new set of metadata files this script is supposed to be run
+    # from the "tests" folder.
+    generate_all_files(dump=True)
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
-# Files that will always have LF line endings on checkout.
-tests/repository_data/** text eol=lf
-
+# All JSON files will always have LF line endings on checkout.
+# This prevents git replacing line endings with CRLF on Windows.
+*.json text eol=lf
--- a/tests/generated_data/ed25519_metadata/root_with_ed25519.json
+++ b/tests/generated_data/ed25519_metadata/root_with_ed25519.json
@@ -1,0 +1,71 @@
+{
+ "signatures": [
+  {
+   "keyid": "5822582e7072996c1eef1cec24b61115d364987faa486659fe3d3dce8dae2aba",
+   "sig": "e2cfb42fe355843c5284ca0bd58acb742a232c2c7b93a87c34ed086cfe51c41a21f77bb8d4ed0c74f9fb3f76f48f7005488347450c28468d8bb288e5988ae201"
+  }
+ ],
+ "signed": {
+  "_type": "root",
+  "consistent_snapshot": true,
+  "expires": "2050-01-01T00:00:00Z",
+  "keys": {
+   "09d440e3725cec247dcb8703b646a87dd2a4d75343e8095c036c32795eefe3b9": {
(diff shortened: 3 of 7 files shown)


MVrachev:
I am not sure how to resolve the mypy issue: https://github.com/theupdateframework/python-tuf/runs/5467473728?check_suite_focus=true
@jku any suggestions?

jku:
> I am not sure how to resolve the mypy issue: https://github.com/theupdateframework/python-tuf/runs/5467473728?check_suite_focus=true @jku any suggestions?

Directories do not become importable packages automatically. You need at least an empty `__init__.py`

MVrachev:
@jku I fixed the mypy issues.
Additionally, I had to add the second commit as on the previous CI run the test failed on all Windows venvs https://github.com/theupdateframework/python-tuf/runs/5467473050?check_suite_focus=true).
The second commit aims to achieve consistency on all operating systems.

The downside to this is that the metadata files won't be human-readable.

jku:
> The second commit aims to achieve consistency on all operating systems.
> 
> The downside to this is that the metadata files won't be human-readable.

That sounds highly undesirable... not to mention that it sounds like a bug somewhere. What is the inconsistency?

jku (changes requested):
Hmm, turns out I have opinions about this. I think we can both
* reduce complexity and
* make this more prepared for future additions


Some suggestions:
* keys that are meant for use in multiple places should not be named for their role (this is the reason I don't want to use the legacy test keys). they should be named for the type of key they are (like "ed25519_1") so that individual tests, or in this case generating functions, can decide what they use the key for.
* The CLI complexity is unused: let's not have arguments unless we really need them. Currently neither argument is needed
* The generated data (files/roles) are now hard coded in four different functions: if I next want to add a single file, I would have to modify all four functions. This cannot be the optimal design. I think the object does not need attributes at all: `generate_repository()` could just do everything it needs to (either write the files it wants to write or validate the same files, depending on the argument): once the function is finished, no-one else needs to know which files it touched.
* generate_repository() calls the data it generates a repository and I get the aim of providing examples of python-tuf repositories... but I think it may be premature:
  * Most importantly **we don't actually check that it's a valid repository**. It looks like one but I can only tell because it's the simplest possible, anything more complex is going to be hard to estimate. The filenames are at least incorrect for a consistent_snapshot repository
  * what if I want to add e.g. a file that tests all keytypes we support? Do I have to create a complete repository to add that one file? Looking at the metadata types, timestamp probably only needs one test case but targets might need multiple different files...
  * Would a better approach be to have one or more methods to generate just a bunch of independent metadata files that are not called a repository, and later we add repository support when we can reasonably test it?

MVrachev:
@jku I think I have simplified the test script a lot.
The test doesn't require context and it should be a lot better.

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,153 @@
...
+
+        meta: Dict[str, MetaFile] = {"targets.json": MetaFile(1)}
+        snapshot = Snapshot(1, self.spec_version, self.expiry, meta)
+        md_snapshot: Metadata[Snapshot] = Metadata(snapshot, {})
+
+        targets = Targets(1, self.spec_version, self.expiry, {})
+        md_targets: Metadata[Targets] = Metadata(targets, {})
+
+        for md in [md_root, md_timestamp, md_snapshot, md_targets]:
+            assert isinstance(md, Metadata)
```
MVrachev: This check is required from mypy otherwise there is a warning:
`
tests/generated_data/generate_md.py:142: error: Argument 1 to "verify_generation" of "MetadataGenerator" has incompatible type "object"; expected "Metadata[Any]"  [arg-type]
`

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,153 @@
...
+    MetaFile,
+    Role,
+    Root,
+    Snapshot,
+    Targets,
+    Timestamp,
+)
+from tuf.api.serialization.json import JSONSerializer
+
+# Hardcode keys and expiry time to achieve reproducibility.
```
lukpueh, line 29: Maybe hardcode expiry here to match the comment?

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,153 @@
...
+    "ed25519_4": {
+        "keytype": "ed25519",
+        "scheme": "ed25519",
+        "keyid": "2be5c21e3614f9f178fb49c4a34d0c18ffac30abd14ced917c60a52c8d8094b7",
+        "keyval": {
+            "public": "0e6738fc1ac6fb4de680b4be99ecbcd99b030f3963f291277eef67bb9bd123e9",
+            "private": "7e2e751145d1b22f6e40d4ba2aa47158207acfd3c003f1cbd5a08141dfc22a15",
+        },
+    },
+}
```
lukpueh: You could condense this a bit, but having a list of (keyid, public val, private val)-tuples and construct the dicts below in a loop. You could even omit keyid in the tuple and compute it in the loop.

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,153 @@
...
+
+keys: Dict[str, Key] = {
+    "ed25519_1": Key.from_securesystemslib_key(keys_dict["ed25519_1"]),
+    "ed25519_2": Key.from_securesystemslib_key(keys_dict["ed25519_2"]),
+    "ed25519_3": Key.from_securesystemslib_key(keys_dict["ed25519_3"]),
+    "ed25519_4": Key.from_securesystemslib_key(keys_dict["ed25519_4"]),
+}
```
lukpueh: Looks like a use case for a loop / dict comprehension.

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class MetadataGenerator:
```
lukpueh: Why use a class? Will there be multiple `MetadataGenerator` objects alongside each other? Wouldn't it be enough to make `generate_all_files` a normal function that takes all constant values from module globals (it already does for keys) and all variable values from function arguments?

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,153 @@
...
+                    + "metadata with 'python generated_data/generate_md.py'"
+                )
+
+    def generate_all_files(
+        self, dump: Optional[bool] = False, verify: Optional[bool] = False
+    ) -> None:
+        """Generate a new repository and optionally verify it.
+
+        Args:
+            dump: Wheter to dump the newly generated files inside self.dir.
```
lukpueh: Is there a use case where the files are not dumped?

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,153 @@
...
+        self.expiry = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+        self.dir = out_dir if out_dir else "ed25519_metadata"
+        if not os.path.exists(self.dir):
+            os.mkdir(self.dir)
+
+        self.spec_version = ".".join(SPECIFICATION_VERSION)
+        # Dump metadata files in a human readable format.
+        self.serializer = JSONSerializer()
+
+    def verify_generation(self, md: Metadata, path: str) -> None:
```
lukpueh: 🎉  I like this function. It's a good example for factoring out functionality that could be in `generate_all_files`, so that it can also be used re-used by other metadata generating functions as well.

lukpueh (commented):
I commented a couple of observations/questions, mostly about further simplifying the script.
But maybe I'm missing the big picture. In what direction should the script scale? 

I'd probably make it just a collection of `generate_*_metadata` functions, plus helper functions to be re-used by those generation function, like  your verification function.

tests/generated_data/generate_md.py
MVrachev: Yes, when we are testing that functionality https://github.com/MVrachev/tuf/blob/45f20beaa83bd2b8b6dbd44164e1a1b7883bd741/tests/test_metadata_generation.py#L20.

tests/generated_data/generate_md.py
MVrachev: I used class mostly because I can reuse variables.
I just wanted to escape from the idea of using script globals that are not hardcoded and generally for me it's easier to follow knowing which variables are reused across functions when you have the `self` prefix.
We have discussed with @jku that it's most likely we will enhance this script generating more variations of the same roles and that could be another reason why a class could be better.

If you still prefer to remove the class I will.

MVrachev:
> I commented a couple of observations/questions, mostly about further simplifying the script. But maybe I'm missing the big picture. In what direction should the script scale?
> 
> I'd probably make it just a collection of `generate_*_metadata` functions, plus helper functions to be re-used by those generation function, like your verification function.

I did it that way before and @jku mentioned that there is no sense in having separate functions now when we generate only one variation per role.
If we enhance this script to generate more variations (as we plan to) it will make sense to have separate `generate_*_metadata` functions.
@jku has mentioned something like this in his last points here https://github.com/theupdateframework/python-tuf/issues/1806#issuecomment-1060417793.

tests/generated_data/generate_md.py
MVrachev: I followed your suggestions and made three lists `public_values`, `private_values` and `keyids`.
I don't want to calculate the `keyids` as that is a possible place where a change inside `securesystemslib` will result in our tests failing which I am not sure we want.

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,146 @@
...
+]
+keyids: List[str] = [
+    "5822582e7072996c1eef1cec24b61115d364987faa486659fe3d3dce8dae2aba",
+    "09d440e3725cec247dcb8703b646a87dd2a4d75343e8095c036c32795eefe3b9",
+    "3458204ed467519c19a5316eb278b5608472a1bbf15850ebfb462d5315e4f86d",
+    "2be5c21e3614f9f178fb49c4a34d0c18ffac30abd14ced917c60a52c8d8094b7",
+]
+
+keys: Dict[str, Key] = {}
+for i in range(0, 4):
```
lukpueh: Hint: `range(0, 4) == range(4)`

tests/generated_data/generate_md.py
lukpueh: > If you still prefer to remove the class I will.

I do because I can't see an advantage of using a class. But I leave it up to you.

lukpueh:
> > I'd probably make it just a collection of `generate_*_metadata` functions, plus helper functions to be re-used by those generation function, like your verification function.
> 
> I did it that way before and @jku mentioned that there is no sense in having separate functions now when we generate only one variation per role.

Oh, I probably wasn't clear. I didn't mean that you should split `generate_all_files` now. I was talking about the future when you have multiple functions that each generate a set of metadata files for a specific use case. So I think we are on the same page here.

tests/generated_data/generate_md.py
MVrachev: I decided to remove the class.

MVrachev:
@lukpueh I updated the pr with removing the class and a couple of small simplifications.
@jku and @lukpueh it's ready for another review.

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,137 @@
...
+
+expires_str = "2050-01-01T00:00:00Z"
+EXPIRY = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
+SPEC_VERSION = ".".join(SPECIFICATION_VERSION)
+OUT_DIR = "generated_data/ed25519_metadata"
+if not os.path.exists(OUT_DIR):
+    os.mkdir(OUT_DIR)
+
+SERIALIZER = JSONSerializer()
+key_info = copy.deepcopy(keys["ed25519_0"].to_securesystemslib_key())
```
lukpueh: Why deepcopy?

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,137 @@
...
+    meta: Dict[str, MetaFile] = {"targets.json": MetaFile(1)}
+    snapshot = Snapshot(1, SPEC_VERSION, EXPIRY, meta)
+    md_snapshot: Metadata[Snapshot] = Metadata(snapshot, {})
+
+    targets = Targets(1, SPEC_VERSION, EXPIRY, {})
+    md_targets: Metadata[Targets] = Metadata(targets, {})
+
+    for md in [md_root, md_timestamp, md_snapshot, md_targets]:
+        assert isinstance(md, Metadata)
+        md.sign(SIGNER)
```
lukpueh: Why do you sign all metadata files with the root key? Why not sign timestamp with timestamp, snapshot with snapshot and targets with targets keys?

lukpueh (commented):
Cool! Thanks for the updates, Martin! I think it's easier to review and use now, without making it much harder to scale later. Here are two more comments about signing...

tests/generated_data/generate_md.py
MVrachev: Because at `keys["ed25519_0"]` we have the dictionary information of a dict containing a dictionary of a dictionary.
More precisely the keyval of a key is again a dictionary and a simple `copy.copy()` doesn't work here.

tests/generated_data/generate_md.py
MVrachev: Changed that.
Initially (while there was a class), it was to reuse the signer.

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,140 @@
...
+    md_timestamp: Metadata[Timestamp] = Metadata(timestamp, {})
+
+    meta: Dict[str, MetaFile] = {"targets.json": MetaFile(1)}
+    snapshot = Snapshot(1, SPEC_VERSION, EXPIRY, meta)
+    md_snapshot: Metadata[Snapshot] = Metadata(snapshot, {})
+
+    targets = Targets(1, SPEC_VERSION, EXPIRY, {})
+    md_targets: Metadata[Targets] = Metadata(targets, {})
+
+    i = 0
```
lukpueh, line 119: hint: `enumerate()` gives you a running index.

tests/generated_data/generate_md.py
lukpueh: What do you mean by "doesn't work here"? And why do you need to copy it at all?

tests/generated_data/generate_md.py
```diff
@@ -0,0 +1,140 @@
...
+    for md in [md_root, md_timestamp, md_snapshot, md_targets]:
+        assert isinstance(md, Metadata)
+        key_info = copy.deepcopy((keys[f"ed25519_{i}"].to_securesystemslib_key()))
+        key_info["keyval"]["private"] = private_values[0]
```
lukpueh, line 123:
```suggestion
        key_info["keyval"]["private"] = private_values[i]
```
?

lukpueh: I think you could avoid `deepcopy`, `to_securesystemslib_key` and the assignment of `private_values[i]` if you use the private key dictionary defined above (before passing it to `from_securesystemslib_key`).

tests/generated_data/generate_md.py
lukpueh: I am not sure I understand. So initially you wanted to sign with one key on purpose?

MVrachev: Yea, because the files are not supposed to be read as a stand-alone repository, but as separate independent files, and as such we can reuse the same key as many times as we want.
The only reason I defined 4 keys is because of the root roles.

tests/generated_data/generate_md.py
MVrachev: I remembered it just when I sent it, thanks.

MVrachev:
Not sure why, but GitHub doesn't see my update which exists on the same branch on my fork...
Will close and reopen the pr.

MVrachev:
@lukpueh because of a GitHub bug the new changes inside test-static-data-generation branch of my fork
weren't noticed. I closed it, but there was no option to reopen it.
That's why I created #1915.

PR #1910, opened 2022-03-18
Add test coverage for delegated hash bins target download

### LINKED ISSUE #1639: RepositorySimulator: add support for delegated hash bins

ivanayov:
This change adds tests coverage for `path_hash_prefixes` and
verifies that role names matching specific prefixed successfully
find and download the corresponding metadata files and do not
succeed to find existing, but non-matching files

Fixes https://github.com/theupdateframework/python-tuf/issues/1639

Follow-up of #1808 

### DIFF
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -32,7 +32,7 @@ class TestDelegation:
     keyids: List[str] = field(default_factory=list)
     threshold: int = 1
     terminating: bool = False
-    paths: List[str] = field(default_factory=lambda: ["*"])
+    paths: Optional[List[str]] = field(default_factory=lambda: ["*"])
     path_hash_prefixes: Optional[List[str]] = None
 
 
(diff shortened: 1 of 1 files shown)


jku:
I've not given this the review it needed, sorry... This vaguely looks like we could do better but I can't immediately see the better way: did you consider having the same test setup as in the existing downloads test `test_targetfile_search()` -- so there would be
* staticly defined metadata structure: it could be the same delegations_tree already defined in the file, just added with a targets metadata that does hashed bin delegations
* test cases would then only have to define a target path to search for, search result and the expected visited order

This would not make the current test much shorter but it might mean that adding a new test case would be maybe 2 lines instead of 40 lines of test data that it is now?

tests/test_updater_delegation_graphs.py
```diff
@@ -438,6 +438,84 @@ def test_targetfile_search(self, test_data: TargetTestCase) -> None:
...
+                    b"fff0838676c1dacb8827b02511e89f85b43814a74dd2c3ffb31d952a1db94311",
+                    "non_matching_path_to_prefix.ext",
+                ),
+                TestTarget(
+                    "20-27",
+                    b"22992ae806e6d652835f3c84e169f05127da69e599f902e4dbdc6ff27ce46d01",
+                    "delegate_1.ext",
+                ),
+            ],
+            visited_order=["f8-ff", "20-27"],
```
MVrachev, line 487: I just noticed you are not using `visited_order` anywhere in your test.

lukpueh:
Is this superseded by #2031? Should we think about deprecating `path_hash_prefixes` now that TAP 15 has landed? (see see https://github.com/theupdateframework/taps/issues/132#issuecomment-1115175226)

jku:
> Is this superseded by #2031? 

I think so. I wouldn't object to a tight client test specifically for hashed bins but I don't think it's super important, and this PR is 80 lines of test code for very limited (new) test functionality.

PR #1913, opened 2022-03-22
build: Add verify-release script

jku:
verify-release
* Builds a release from current commit
* Warns if git describe does not match built version (in other words we're building from a commit that is not release tagged)
* Warns if built version is not the latest GitHub or PyPI version
* Errors if GitHub or PyPI release artifacts for this version do not match the built
  release artifacts

This should be useful after release as any developer (or a CI job) can
easily verify that the release matches the sources in git.

Note that the last checks currently fail as the 1.0 build was not
reproducible. They should succeed after next release.

---

Some comments:
* this commit adds a file to root directory, I don't like it... but I didn't know where else to put it
* signatures are not checked because it's so painful. This could be a future improvement if someone has good ideas...
* if the release files do not match, there is an error but no details are printed and the downloaded files are not left available: I did not want to add this complexity to the script

### DIFF
--- a/verify_release
+++ b/verify_release
@@ -1,0 +1,159 @@
+#!/usr/bin/env python
+
+# Copyright 2022, TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""verify_release - verify that published release matches a locally built one
+
+Builds a release from current commit and verifies that the release artifacts
+on GitHub and PyPI match the built release artifacts.
+"""
+
+import json
+import os
+import subprocess
+import sys
+from filecmp import dircmp
+from tempfile import TemporaryDirectory
+
+import requests
+
+# Project variables
+# Note that only these project artifacts are supported:
+# [f"{PYPI_PROJECT}-{VER}-none-any.whl",  f"{PYPI_PROJECT}-{VER}.tar.gz"]
+GITHUB_ORG = "theupdateframework"
+GITHUB_PROJECT = "python-tuf"
+PYPI_PROJECT = "tuf"
+
+
+def build(build_dir: str) -> str:
+    """Build release locally. Return version as string"""
+    cmd = ["python3", "-m", "build", "--outdir", build_dir]
+    subprocess.run(cmd, stdout=subprocess.DEVNULL, check=True)
+    build_version = None
+    for filename in os.listdir(build_dir):
+        prefix, postfix = f"{PYPI_PROJECT}-", ".tar.gz"
+        if filename.startswith(prefix) and filename.endswith(postfix):
+            build_version = filename[len(prefix) : -len(postfix)]
+    assert build_version
+    return build_version
+
+
+def get_git_version() -> str:
+    """Return version string from git describe"""
+    cmd = ["git", "describe"]
+    process = subprocess.run(cmd, text=True, capture_output=True, check=True)
+    assert process.stdout.startswith("v") and process.stdout.endswith("\n")
+    return process.stdout[1:-1]
+
+
+def get_github_version() -> str:
+    """Return version string of latest GitHub release"""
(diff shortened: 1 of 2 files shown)


verify_release
```diff
@@ -0,0 +1,159 @@
...
+    with TemporaryDirectory() as pypi_dir:
+        cmd = ["pip", "download", "--no-deps", "--dest", pypi_dir]
+        target = f"{PYPI_PROJECT}=={version}"
+        binary_download = cmd + [target]
+        source_download = cmd + ["--no-binary", ":all:", target]
+
+        subprocess.run(binary_download, stdout=subprocess.DEVNULL, check=True)
+        subprocess.run(source_download, stdout=subprocess.DEVNULL, check=True)
+
+        return not dircmp(pypi_dir, compare_dir).diff_files
```
jku: maybe this should check common_files instead of diff_files to also cover things like a completely missing  release artifact...

verify_release
```diff
@@ -0,0 +1,159 @@
...
+    """Return version string of latest GitHub release"""
+    release_json = f"https://api.github.com/repos/{GITHUB_ORG}/{GITHUB_PROJECT}/releases/latest"
+    releases = json.loads(requests.get(release_json).content)
+    return releases["tag_name"][1:]
+
+
+def get_pypi_version() -> str:
+    """Return latest version string available on PyPI"""
+    # pip is sad: it can't even tell what the newest available version is...
+    # FIXME: Maybe think of something better than downloading latest tarball here?
```
lukpueh: You could use the [PyPI JSON API project endpoint](https://warehouse.pypa.io/api-reference/json.html#project): `https://pypi.org/pypi/tuf/json`

lukpueh (commented):
Cool stuff, @jku! I tried it out and it works nicely.

Are we supposed to run it with development requirements installed? It only needs `requests` and `build` to be installed. Maybe it's worth mentioning requirements somewhere, e.g. in the doc header or in RELEASE.md?

Will get back with an actual code review.

verify_release
jku: True. I intentionally wanted to find the version that pip considers current since that's what matters in the end (obviously the two should be the same but...).

I suppose the method name is a bit misleading in that sense

lukpueh: Did you consider using experimental `pip index versions tuf` or parsing error output of `pip install tuf==`? (quoting [SO](https://stackoverflow.com/questions/4888027/python-and-pip-list-all-versions-of-a-package-thats-available/26664162#26664162))

jku: well, I have not found a way to find the answer to _"what is **current** version according to pip?"_ I think none of those answer this question.

lukpueh: Okay, that all makes sense. :)

jku: maybe I've left out the reasoning for this: A malicious maintainer could upload tuf version 99. In this case it's not enough that we check that locally built 1.0 matches pypi version 1.0: end users are still getting version 99 if they do `pip install tuf`.

jku: I've improved naming and comments, did not modify the code

verify_release
jku: yes, this is a bug: if files are only found in one of the directories, then diff_files is empty. So should rather test that `common_files` list contains the two expected filenames.

lukpueh: Are you going to change it (here and above)? Also, did you verify that `common_files` recurses into subdirs?

lukpueh (changes requested):
LGTM, modulo the bug you discovered yourself.

verify_release
jku: I'll fix this yes.

jku: the correct field is actually `same_files`. I hard coded the expected list of files (1 tarball, 1 wheel) that must match when comparing releases.

jku:
Issue is fixed. The testing is obviously not quite 100.0% (since we dont have a positive test case on pypi.org yet) but I have done the manual testing I could.

lukpueh (approved):
🎉

PR #1915, opened 2022-03-22
Tests: provide a way to generate a simple metadata set 

MVrachev:
**Continuation from #1898.**
Because of a GitHub bug the new changes inside `test-static-data-generation` branch of my fork
weren't noticed.
That's why I created a new pr.

Related to: #1806

I created a new script called "generate_md.py" which can be used
to easily generate a repository. Additionally, I created a new
test file making sure that the locally stored metadata files and
the newly generated metadata roles are the same.
This will allow us to test that we are not changing the metadata
file structure when making changes.

requirements**:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
-# Files that will always have LF line endings on checkout.
-tests/repository_data/** text eol=lf
-
+# All JSON files will always have LF line endings on checkout.
+# This prevents git replacing line endings with CRLF on Windows.
+*.json text eol=lf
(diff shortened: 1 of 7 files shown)


MVrachev:
@lukpueh because of a GitHub bug the new changes inside `test-static-data-generation` branch of my fork
weren't noticed. I closed it, but there was no option to reopen it.
That's why I created this new pr.

jku (approved):
This looks a lot better and shorter, thanks!

* The key & signer handling could probably be better (why does the generator function need to construct the signers?) but there's no need to polish everything before we know what the code is going to be like...
* The error message could mention that new expected results should only be generated **if** the change in results was expected
* script only works if run from a specific working directory
* two separate arguments (dump/verify) is weird: that leads to 4 distinct modes but only two of them make sense...

These are all either trivial or just thoughts for future: I think it's a good start for these tests, approved.

PR #1918, opened 2022-03-23
build: pin direct test dependencies only

### LINKED ISSUE #1899: Rethink strategy for pinning test dependencies

Since 5643cecf we pin test dependencies in order to improve the determinism of our testing and to prevent test dependency updates from causing CI failures without the PR being tested changing.

Over approximately two weeks we have 9 open PRs from dependabot for test dependencies.

This relatively long list could potentially obfuscate dependot updates in python-tuf dependencies, which we care to address with higher priority.

Furthermore, the level of review rigour required for test dependency updates differs from the that we apply to python-tuf dependency updates.

It may be desirable to alter our approach to pinning test depependencies such that we can reduce the PR load/review implications while preserving the CI/CD determinism.

**Current behavior**:

Modest PR load on test dependencies which we cannot reasonably review with the same level of depth and rigour as we apply to our library dependencies.

**Expected behaviour**:

??

**Ideas**:
- @jku suggested we might only pin top-level test dependencies
- we could label test dependencies differently, so that they are easier to filter, by having a different [`updates`](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#configuration-options-for-updates) entries constrained to the test dependency requirements file.

lukpueh:
Fixes #1899
Reverts #1867

In https://github.com/theupdateframework/python-tuf/pull/1867 we started pinning direct and transitive test
dependencies for stable test results, i.e. to not have an unnoticed
update of a used test tool (or their dependencies) break our tests.

This resulted in a dependabot updates inundating our PR tracker,
potentially obfuscating updates, which we care to address with
higher priority.

As a compromise we now only pin direct test dependencies, which
should still give us relatively stable test runs, while reducing
the spam.

requirements**:


jku (approved):
Looks good to me.

PR #1922, opened 2022-03-24
Add default args to Signed constructors

### LINKED ISSUE #1459: Create constructor(s) for creating Metadata from scratch

jku:
Tweak the constructor signatures to make creating brand new Metadata easier (the constructors were originally designed for deserialization purposes). Modify examples and tests to use the new constructor args.

The changes are backwards compatible. Resulting API is much easier to use for creating new Metadata. The API is still a bit unusual but not difficult to learn IMO:

    targets = Metadata(Targets())

There are some noteworthy cases:
* `expires` does not have a reasonable default value and I planned to leave it a required argument. That would have broken backwards compat however as order of arguments would have had to change. So expires now defaults to an arbitrary 1 day in future _EDIT: changed to `utcnow()` so default is expired metadata_
* `consistent_snapshot` defaults to True but None is still possible to indicate Metadata does not contain the field at all
* `spec_version` defaults to whatever is the current library supported version

Only the core metadata classes we're updated. There are other classes that would probably benefit from a better constructor, like Delegations, but those were not handled yet. 

Fixes #1459

requirements**:

### DIFF
--- a/tests/generated_data/generate_md.py
+++ b/tests/generated_data/generate_md.py
@@ -16,7 +16,6 @@ from tuf.api.metadata import (
     TOP_LEVEL_ROLE_NAMES,
     Key,
     Metadata,
-    MetaFile,
     Role,
     Root,
     Snapshot,
@@ -97,23 +96,15 @@ def generate_all_files(
         verify: Whether to verify the newly generated files with the
             local staored.
     """
-    root = Root(1, SPEC_VERSION, EXPIRY, {}, ROLES, True)
-    root.add_key("root", keys["ed25519_0"])
-    root.add_key("timestamp", keys["ed25519_1"])
-    root.add_key("snapshot", keys["ed25519_2"])
-    root.add_key("targets", keys["ed25519_3"])
+    md_root = Metadata(Root(expires=EXPIRY))
+    md_timestamp = Metadata(Timestamp(expires=EXPIRY))
+    md_snapshot = Metadata(Snapshot(expires=EXPIRY))
+    md_targets = Metadata(Targets(expires=EXPIRY))
 
-    md_root: Metadata[Root] = Metadata(root, {})
-
-    timestamp = Timestamp(1, SPEC_VERSION, EXPIRY, MetaFile(1))
-    md_timestamp: Metadata[Timestamp] = Metadata(timestamp, {})
-
-    meta: Dict[str, MetaFile] = {"targets.json": MetaFile(1)}
-    snapshot = Snapshot(1, SPEC_VERSION, EXPIRY, meta)
-    md_snapshot: Metadata[Snapshot] = Metadata(snapshot, {})
-
-    targets = Targets(1, SPEC_VERSION, EXPIRY, {})
-    md_targets: Metadata[Targets] = Metadata(targets, {})
+    md_root.signed.add_key("root", keys["ed25519_0"])
+    md_root.signed.add_key("timestamp", keys["ed25519_1"])
+    md_root.signed.add_key("snapshot", keys["ed25519_2"])
+    md_root.signed.add_key("targets", keys["ed25519_3"])
 
     for i, md in enumerate([md_root, md_timestamp, md_snapshot, md_targets]):
         assert isinstance(md, Metadata)
--- a/tests/generated_data/generate_md.py
+++ b/tests/generated_data/generate_md.py
@@ -11,17 +11,7 @@ from typing import Dict, List, Optional
 from securesystemslib.signer import SSlibSigner
 
 from tests import utils
-from tuf.api.metadata import (
-    SPECIFICATION_VERSION,
-    TOP_LEVEL_ROLE_NAMES,
-    Key,
-    Metadata,
-    Role,
-    Root,
-    Snapshot,
-    Targets,
-    Timestamp,
-)
+from tuf.api.metadata import Key, Metadata, Root, Snapshot, Targets, Timestamp
 from tuf.api.serialization.json import JSONSerializer
 
 # Hardcode keys and expiry time to achieve reproducibility.
@@ -60,13 +50,11 @@ for index in range(4):
 
 expires_str = "2050-01-01T00:00:00Z"
 EXPIRY = datetime.strptime(expires_str, "%Y-%m-%dT%H:%M:%SZ")
-SPEC_VERSION = ".".join(SPECIFICATION_VERSION)
 OUT_DIR = "generated_data/ed25519_metadata"
 if not os.path.exists(OUT_DIR):
     os.mkdir(OUT_DIR)
 
 SERIALIZER = JSONSerializer()
-ROLES = {role_name: Role([], 1) for role_name in TOP_LEVEL_ROLE_NAMES}
 
 
 def verify_generation(md: Metadata, path: str) -> None:
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -32,7 +32,7 @@ import fnmatch
 import io
 import logging
 import tempfile
-from datetime import datetime, timedelta
+from datetime import datetime
 from typing import (
     IO,
     Any,
@@ -500,7 +500,7 @@ class Signed(metaclass=abc.ABCMeta):
 
         self.spec_version = spec_version
 
-        self.expires = expires or datetime.utcnow() + timedelta(days=1)
+        self.expires = expires or datetime.utcnow()
 
         if version is None:
             version = 1
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -32,7 +32,7 @@ import fnmatch
 import io
 import logging
 import tempfile
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import (
     IO,
     Any,
@@ -120,11 +120,11 @@ class Metadata(Generic[T]):
     def __init__(
         self,
         signed: T,
-        signatures: Dict[str, Signature],
+        signatures: Optional[Dict[str, Signature]] = None,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
         self.signed: T = signed
-        self.signatures = signatures
+        self.signatures = signatures or {}
         self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
 
     def __eq__(self, other: Any) -> bool:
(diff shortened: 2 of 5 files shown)


jku:
> * [ ]  Tests have been added for the bug fix or new feature

The generated tests and serialization tests do a good job testing this I believe.

> * [x]  Docs have been added for the bug fix or new feature

Metadata docstring should maybe be improved to show an example of creating a new Metadata...

tuf/api/metadata.py
```diff
@@ -120,11 +120,11 @@ class Metadata(Generic[T]):
     def __init__(
         self,
         signed: T,
-        signatures: Dict[str, Signature],
+        signatures: Optional[Dict[str, Signature]] = None,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ):
         self.signed: T = signed
-        self.signatures = signatures
+        self.signatures = signatures or {}
```
jku: this construct is a bit problematic in a specific edge case of the argument container being empty:

```python
sigs = {}
root = Metadata(Root(), sigs)
assert sigs is root.signatures
```

so maybe I should check `is None` instead

MVrachev: Yea, I think this:
`self.signatures = signatures if signatures is not None else {}`
would be better.

tuf/api/metadata.py
```diff
@@ -497,11 +499,15 @@ def __init__(
             raise ValueError(f"Unsupported spec_version {spec_version}")
 
         self.spec_version = spec_version
-        self.expires = expires
 
-        if version <= 0:
+        self.expires = expires or datetime.utcnow() + timedelta(days=1)
```
MVrachev: My personal preference is if we use a default value of one year for example.
I am a little worried for an instance when somebody forgets passing `expires` and `expires` becomes only one day...

MVrachev: Also, I think it's worth noting in the docstring that if no expires is given a default value of `X` amount of days/years will be used.
Or maybe we don't want to promote that expires is optional and not mention it?

tuf/api/metadata.py
```diff
@@ -497,11 +499,15 @@ def __init__(
...
+
+        if version is None:
+            version = 1
```
MVrachev, line 517: The same question about the docstring and default values applies to all attributes.
Do we want to mention the default values?

tuf/api/metadata.py
```diff
@@ -1660,15 +1668,15 @@ class Targets(Signed):
...
+        version: Optional[int] = None,
+        spec_version: Optional[str] = None,
+        expires: Optional[datetime] = None,
+        targets: Optional[Dict[str, TargetFile]] = None,
         delegations: Optional[Delegations] = None,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
     ) -> None:
         super().__init__(version, spec_version, expires, unrecognized_fields)
-        self.targets = targets
+        self.targets = targets or {}
```
MVrachev: Do you think it would be a good idea to check for `None` again instead of a truthy value the same way as you suggested about `signatures`?
Maybe a user could want to just pass `{}` targets?

tests/generated_data/generate_md.py
```diff
@@ -97,23 +96,15 @@ def generate_all_files(
...
-    timestamp = Timestamp(1, SPEC_VERSION, EXPIRY, MetaFile(1))
-    md_timestamp: Metadata[Timestamp] = Metadata(timestamp, {})
-
-    meta: Dict[str, MetaFile] = {"targets.json": MetaFile(1)}
-    snapshot = Snapshot(1, SPEC_VERSION, EXPIRY, meta)
-    md_snapshot: Metadata[Snapshot] = Metadata(snapshot, {})
-
-    targets = Targets(1, SPEC_VERSION, EXPIRY, {})
-    md_targets: Metadata[Targets] = Metadata(targets, {})
+    md_root = Metadata(Root(expires=EXPIRY))
```
MVrachev, line 87: I think because of the use of default values for Metadata and Root attributes you can remove `ROLES` as well (line 69).

tests/generated_data/generate_md.py
```diff
@@ -97,23 +96,15 @@ def generate_all_files(
...
+    md_timestamp = Metadata(Timestamp(expires=EXPIRY))
+    md_snapshot = Metadata(Snapshot(expires=EXPIRY))
+    md_targets = Metadata(Targets(expires=EXPIRY))
```
MVrachev, line 90: I really like you don't have to create temporary `Signed` objects but you can't directly instantiate a Metadata object with populated default `Signed` fields as well. 
Really cool!

MVrachev (commented):
Overall, the changes are great with a few comments.

I am wondering whether we should simplify and use the default attribute values in the `examples` scripts?
For those examples, I think we all agree the most important priority is that the code is easy to understand.
Otherwise, the simplifications are great, no doubt.

tuf/api/metadata.py
jku: There is no good default value: it's not clear at all that a years expiry is better than one days expiry -- at least a short one won't be used in production for a year before the mistake is noticed. The safe thing would be to set expiry to `utcnow()` so metadata is immediately expired, that's certainly an option... Maybe an option I like the most.

The decision is not very important I think? More important is that we show what should be done in examples and docs.

tuf/api/metadata.py
jku: yeah all these constructs are a bit suspect when the field is a container, checking for None is probably best

tests/generated_data/generate_md.py
jku: right, same is likely true for SPEC_VERSION. 

Lint won't complain about unused variables because they're globals so they get exported as module variables. That's a bit annoying, I wonder if there's a best practice here...

tuf/api/metadata.py
MVrachev: If we do want to maintain the compatibility of this feature, yes, it makes sense to make it expire as soon as possible.

jku:
I believe all comments we're resolved with additional commits:
* more docs for the default values
* more care taken with empty containers as args (this issue is theoretically there for unrecognised_fields as well btw)
* expires default is `utcnow()` (so expired out of the gate)
* some unused variables cleaned up

jku:
> I am wondering whether we should simplify and use the default attribute values in the `examples` scripts?

I believe this is done, if you have specific place you're thinking of let me know.

tuf/api/metadata.py
```diff
@@ -1183,12 +1199,14 @@ class Snapshot(Signed):
...
-        spec_version: Supported TUF specification version number.
-        expires: Metadata expiry date.
+        version: Metadata version number. Default is 1.
+        spec_version: Supported TUF specification version. Default is the
+            version currently supported by the library.
+        expires: Metadata expiry date. Default is current date and time.
         unrecognized_fields: Dictionary of all attributes that are not managed
             by TUF Metadata API
         meta: Dictionary of target metadata filenames to ``MetaFile`` objects.
+            Default is a dictionary with a Metafile for "snapshot.json".
```
```suggestion
            Default is a dictionary with a Metafile for "snapshot.json" with version 1.
```

MVrachev:
> > I am wondering whether we should simplify and use the default attribute values in the `examples` scripts?
> 
> I believe this is done, if you have specific place you're thinking of let me know.

I am sorry, I think I was unclear.
I meant do we want to change anything inside the `examples` folders at all?
My point was that the primary purpose of all examples scripts is to be easily understandable and this new syntax using the default values could hide some details.

jku:
> examples scripts is to be easily understandable and this new syntax using the default values could hide some details.

If there is a specific case where this is true I'm willing  to discuss. I just don't see one, everything seems to have a fairly obvious default value apart from expiry.

tuf/api/metadata.py
```diff
@@ -104,6 +104,14 @@ class Metadata(Generic[T]):
     ``[Root]`` is not validated at runtime (as pure annotations are not available
     then).
 
+    New Metadata instances can created from scratch with::
```
```suggestion
    New Metadata instances can be created from scratch with::
```

tuf/api/metadata.py
```diff
@@ -444,9 +453,10 @@ class Signed(metaclass=abc.ABCMeta):
     instance attributes.*
 
     Args:
-        version: Metadata version number.
-        spec_version: Supported TUF specification version number.
-        expires: Metadata expiry date.
+        version: Metadata version number. Default is 1.
+        spec_version: Supported TUF specification version. Default is the
+            version currently supported by the library.
+        expires: Metadata expiry date. Default is current date and time.
```
lukpueh: Is it worth mention that these are optional in the child classes only? I was confused for a split-second, because of the `Optional` type hint but no default value in the constructor signature (I understand now that [`Optional[X]` just means `X | None`](https://docs.python.org/3/library/typing.html#typing.Optional) regardless of it being an optional arg). What about something like...

```suggestion
        expires: Metadata expiry date. If None is passed the current date and time is assigned.
```
(same for `version` and `spec_version`)

lukpueh (changes requested):
I love the change! Please fix typo (see inline) and this is good to go. 

The comment about the docstring is not a blocker nor is this one:
Re beeing more careful with empty containers (d8c0f3b3f39d11a406883164b0d32d4346ba711f), should we also update all the `unrecognized_fields or {}` in Metadata constructors?

tuf/api/metadata.py
jku: Maybe that would be better -- the small but important difference between Optional and the argument actually being optional is indeed annoying and hard to keep clear

jku:
Force pushed with the documentation commit amended with three fixes for issues pointed out in comments (the last three resolved discussions).

> Re beeing more careful with empty containers (https://github.com/theupdateframework/python-tuf/commit/d8c0f3b3f39d11a406883164b0d32d4346ba711f), should we also update all the unrecognized_fields or {} in Metadata constructors?

I'll file an issue: it's pretty minor but strictly speaking yes we should

tuf/api/metadata.py
```diff
@@ -444,9 +453,11 @@ class Signed(metaclass=abc.ABCMeta):
     instance attributes.*
 
     Args:
-        version: Metadata version number.
-        spec_version: Supported TUF specification version number.
-        expires: Metadata expiry date.
+        version: Metadata version number. If None, 1 one is assigned.
```
```suggestion
        version: Metadata version number. If None, 1 is assigned.
```
?

jku: damn it :facepalm:

lukpueh: 🤷  happens

jku:
Once more, force-pushed with a fix for latest resolved review comment

lukpueh (approved):
Thanks!

PR #1926, opened 2022-03-25
verify_release: Warn about missing requirements

jku:
This is mostly useful for build module as it's not imported otherwise:
we explicitly call "python -m build" so everything works like in a
real release build.

---

Forgot this from the initial PR: it's not critical but I think makes sense. verify_release should be something that's easy to run for anyone, not just core developers who have requirements-dev.txt installed.


lukpueh (approved):
Thanks

PR #1932, opened 2022-03-30
Add workflow for basic codeql analysis

joshuagl:
Fixes #NA

Adds a workflow to leverage CodeQL analysis

requirements**:

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,30 @@
+name: "CodeQL Analysis"
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
+  schedule:
+    - cron: '30 0 * * 2'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@27ea8f8fe5977c00f5b37e076ab846c5bd783b96
+      with:
+        languages: 'python'
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@27ea8f8fe5977c00f5b37e076ab846c5bd783b96


jku (commented):
I've looked at codeql checks at some point: very few of them seem to be relevant to our code base but I suppose there's nothing wrong with enabling this just in case?

* workflow would be more readable if generic comments and the unused abstractions  were removed (like a "strategy" and a "matrix" for our one single language, or the autobuild task for a non-compiled language... ). I'm fine with merging as is but more readable is probably better, especially considering following point
* what are the pros/cons of keeping this as separate workflow and not as just another job in ci.yml?
* what happens if the cron triggered workflow fails?

lukpueh:
@joshuagl, do you want to finalize this PR? Should someone take over?

> I've looked at codeql checks at some point: very few of them seem to be relevant to our code base but I suppose there's nothing wrong with enabling this just in case?

@jku, do you know which checks could be relevant for our code base? Regardless, I agree that we probably can just enable it and see what it does.

> * workflow would be more readable if generic comments and the unused abstractions  were removed (like a "strategy" and a "matrix" for our one single language, or the autobuild task for a non-compiled language... ). I'm fine with merging as is but more readable is probably better, especially considering following point

I'm a bit against merging as is. We should at least 
- get the branch names right
- remove commented out code and irrelevant comments
- pin the used `codeql-action`s to sha refs

> * what are the pros/cons of keeping this as separate workflow and not as just another job in ci.yml?

I think the only reason to add the codeql job to the ci workflow would be to trigger on the same events (which would mean no cron) and execute it on the same runner (which has no clear advantage)

> * what happens if the cron triggered workflow fails?

Maybe display an alert?
https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-alerts

joshuagl:
Thanks for the feedback both. I'd just taken the sample workflow and opened a draft PR to see if it provided any useful results. I can clean up the PR and mark it "ready for review" as we have consensus that it's worth trying.

joshuagl:
I _think_ this is ready for review.

AIUI the point of the scheduled analysis is so that there's an up-to-date baseline for PR's to compare against, then the analysis run on the PR can inform submitters that there are M alerts compared to N on the main branch.

jku (approved):
> @jku, do you know which checks could be relevant for our code base?

I don't remember details, just that most checks were relevant for e.g. web services and such. I'm sure there are things useful for us as well and there's no telling how the test set evolves in future.

Anyway, looks fine to me.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,30 @@
+name: "CodeQL Analysis"
+
+on:
+  push:
+    branches: [ develop, '*' ]
```
lukpueh: IIUC this means trigger on push to develop branch and all branches that don't contain a slash. Is this intended?

lukpueh (approved):
Thanks for cleaning this up, @joshuagl! I left one question about the trigger condition inline, but I'm fine merging as is.

.github/workflows/codeql-analysis.yml
joshuagl: You're right (see [filter pattern cheat sheet](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)). I had misremembered and thought we had a branch for the pre 1.0 state, but we do not. I think we only want to do CodeQL analysis on the `develop` branch.

joshuagl:
I had to rebase on the latest develop, so squashed commits into one. The only code change since the force push was to update the branches we trigger on per https://github.com/theupdateframework/python-tuf/pull/1932#discussion_r897745187

PR #1933, opened 2022-03-30
Add GH workflow to build and release on PyPI (WIP)

### LINKED ISSUE #1550: Automated release builds

Automate releases to reduce potential for human error and protect against developer machine compromise.

**Current behavior**:

Current release process is very manual and runs on the local machine of a developer making the release: https://github.com/theupdateframework/tuf/blob/7731738590f291041c06b3c525e248d245cb1f33/docs/RELEASE.md

**Expected behavior**:

Release process is automated as much as possible and, ideally, happens in a hermetic environment on a hosted build service.

At a minimum we should script releases, but even better would be to define a plan to meet SLSA 2 or above https://slsa.dev/levels

lukpueh:
Fixes #1550 

This PR adds a workflow with two jobs to **build** and **release on PyPI**. The release job waits for the build job and uses a custom release environment, which can be configured to require review.

To share the build artifacts between the jobs and to make them available for intermediate review, we use [actions/upload-artifact](https://github.com/actions/upload-artifact) (in build) and [actions/download-artifact](https://github.com/actions/download-artifact) (in release) (also see [GitHub docs](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts)).

To upload the build artifacts to PyPI, we use [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) as [recommend by PyPA](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/).

A [demo is available in my fork](https://github.com/lukpueh/tuf/actions/runs/2065688693).

**Caveat**
The URL to grab the build artifacts created in a workflow run requires knowledge of the corresponding action ID and artifact ID, plus a login token (w/o special permissions). This makes it a bit cumbersome to fetch the artifacts with a script in order to compare them to a local build. It might be easier to instruct reviewer to download the [artifacts manually using the GitHub UI](https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts), and tailor the [local verification script ](https://github.com/theupdateframework/python-tuf/pull/1913) accordingly.

**TODO**
- [ ] Change `repository_url` to release on PyPI instead of Test PyPI and update the token
- [ ] Require successful CI run for a pushed tag to trigger the workflow, instead of just a push (see [`workflow_run` docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow))
- [ ] Integrate workflow with [local verification](https://github.com/theupdateframework/python-tuf/pull/1913) (see caveat above)
- [ ] Add "Release on GitHub" job to workflow
   Note, this could be akin to and in parallel with the "release-on-pypi" job, waiting for the same review approval, and uploading the same build artifacts as release assets. Alternatively, we could create the release as part of the build job and upload the build artifacts as release assets right away, bypassing the download/upload caveat from above. It would be nice if we could set the release as draft at first, but that has the same caveat, i.e. it requires a token and special knowledge to download the assets.
- [ ] Update documentation
  - [ ] [RELEASE.md](https://github.com/theupdateframework/python-tuf/blob/434730fc3354574d548a9b947bdc5314ec5a2a3b/docs/RELEASE.md)
  - [ ] [INSTALLATION.html#verify-release-signatures](https://theupdateframework.readthedocs.io/en/latest/INSTALLATION.html#verify-release-signatures)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,53 @@
+name: CD
+concurrency: cd
+
+# Trigger workflow on release tag push
+on:
+  push:
+    # TODO: Should we restrict to vX.Y.Z tags?
+    tags: v*
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout release tag
+        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
+
+      - name: Set up Python
+        uses: actions/setup-python@0ebf233433c08fb9061af664d501c3f3ff0e9e20
+        with:
+          python-version: '3.x'
+
+      - name: Install build dependency
+        run: python3 -m pip install --upgrade pip build
+
+      - name: Build binary wheel and source tarball
+        run: python3 -m build --sdist --wheel --outdir dist/ .
+
+      - name: Store build artifacts for review and release
+        uses: actions/upload-artifact@6673cd052c4cd6fcf4b4e6e60ea986c889389535
+        with:
+          name: build-artifacts
+          path: dist
+
+  release-on-pypi:
+    name: Release on PyPI
+    runs-on: ubuntu-latest
+    needs: build
+    environment: release
+    steps:
+      - name: Fetch build artifacts
+        uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741
+        with:
+          name: build-artifacts
+          path: dist
+
+      - name: Publish binary wheel and source tarball on PyPI
+        uses: pypa/gh-action-pypi-publish@717ba43cfbb0387f6ce311b169a825772f54d295
+        with:
+          user: __token__
+          # TODO: Change to PyPI and update token
+          repository_url: https://test.pypi.org/legacy/
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -71,8 +71,8 @@ jobs:
       - name: Finalize GitHub release
         uses: actions/github-script@9ac08808f993958e9de277fe43a64532a609130e
         with:
-        script: |
-          console.log(${{needs.build.outputs.release_id}})
+          script: |
+            console.log(${{needs.build.outputs.release_id}})
 #           await github.rest.repos.updateRelease({
 
 
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,6 +11,7 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    if: ${{ github.ref_type == 'tag' }}
     steps:
       - name: Checkout release tag
         uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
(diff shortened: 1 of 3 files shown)


lukpueh:
Re "Release on GitHub":
Unfortunately, the GitHub-native [action/create-release](https://github.com/actions/create-release) is unmaintained. But there is a third-party [softprops/action-gh-release](https://github.com/softprops/action-gh-release), which seems popular. If we use that we'd have to make an exception in our "only allow actions created by GitHub"-policy.

jku:
> The URL to grab the build artifacts created in a workflow run requires knowledge of the corresponding action ID and artifact ID, plus a login token

:open_mouth: wow. They really are making artifacts hidden from unauthenticated users, how disappointing of GitHub. This limits the number of people able to download the artifacts to a mere 70 million accounts: a security improvement indeed :roll_eyes:

This was so close to Just Working but then... not quite.

So, assuming we do want to integrate the GitHub release into the process and also want to have the local verification as automatic as possible, a potential solution for both is:
* add code to do the GH release (either using a 3rd party action or just adding the code, it doesn't look _horribly bad_ for our simple case, 1 call to create release, 1 call for each asset)
* do the GH release during the "build" job (as in before the maintainer deploy permission), but name the release like "1.1.0-rc" or something
* now verify_release would be able to check that the release (candidate) matches a local build
* after maintainer approval, the "deploy" job would take the release artifacts and push them to pypi. GH release can be renamed to "1.1.0"

It's definitely more work than I expected but maybe doable? We would not need to handle edge cases since we could just say e.g. that maintainers need to delete failed -rc releases before trying to re-release

jku:
I think using the `pypa/gh-action-pypi-publish` action seems fine... but also it literally just calls `twine check` and `twine upload` and that's it. We could do the same manually (and this would allow running `twine check` in the build phase instead of failing in deploy)

jku:
> release environment, which can be configured to require review.

To clarify, this means configured in GitHub project/environment settings, not in the workflow? 

Are we defining that all maintainers can deploy? I think this might be fine: Current pypi.org project maintainers is a subset of all maintainers but that makes sense as you can upload _anything_ with pypi.org access. With deploy permission you can only release from git sources...

lukpueh:
> 😮 wow. They really are making artifacts hidden from unauthenticated users, how disappointing of GitHub. This limits the number of people able to download the artifacts to a mere 70 million accounts: a security improvement indeed 🙄

Yep, [others too have expressed their doubt](https://github.community/t/public-read-access-to-actions-artifacts/17363) about the security benefit of this restriction.

> This was so close to Just Working but then... not quite.

Yep, I was really excited when I discovered the download/upload feature... and then quickly somewhat disappointed.

> So, assuming we do want to integrate the GitHub release into the process and also want to have the local verification as automatic as possible, a potential solution for both is:
>  [...]

Thanks for thinking this through with me! That's one of the approaches I had in mind. I'll explore in that direction then.

lukpueh:
> I think using the `pypa/gh-action-pypi-publish` action seems fine... but also it literally just calls `twine check` and `twine upload` and that's it. We could do the same manually (and this would allow running `twine check` in the build phase instead of failing in deploy)

Right, I forgot to mention that this is a 3rd-party action too. I was less worried since it seems endorsed by the pypa. But yeah, if it just calls twine, we might as call it ourselves. 

I'll prepare a POC using `pypa/gh-action-pypi-publish` and `softprops/action-gh-release`, and in a next iteration will try to replace them with custom code.

lukpueh:
> > release environment, which can be configured to require review.
> 
> To clarify, this means configured in GitHub project/environment settings, not in the workflow?

Yes!  Environments and protection rules like "requires review" are configured in the GitHub repo settings (see [GH docs](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#environment-protection-rules)).

> 
> Are we defining that all maintainers can deploy? I think this might be fine: Current pypi.org project maintainers is a subset of all maintainers but that makes sense as you can upload _anything_ with pypi.org access. With deploy permission you can only release from git sources...

I haven't thought this far. I probably would have started with requiring a review from one of the maintainers who currently also has pypi.org access. I am not sure if the git sources constraint is a good argument for broadening the subset, because git sources can be _anything_.

jku:
One more thing (that is probably future work as well): is signing the release still relevant? 

I'm not sure if anyone uses our signatures for anything... but we could:
* have a signing key in an environment secret and sign with gpg or
* do what the cool kids are doing:  https://blog.chainguard.dev/zero-friction-keyless-signing-with-github-actions/

lukpueh:
> One more thing (that is probably future work as well): is signing the release still relevant?
> 
> I'm not sure if anyone uses our signatures for anything... but we could:
> 
> * have a signing key in an environment secret and sign with gpg or
> * do what the cool kids are doing:  https://blog.chainguard.dev/zero-friction-keyless-signing-with-github-actions/

I'm in favor of the latter, GPG isn't relevant for PyPI anymore, and IIUC this is also the right direction towards generating in-toto attestations for our supply chain (I really need to catch up on the latest in-toto/sigstore developments).

jku:
Even more future improvements (there's a lot of details in this one PR/discussion, no need to fix everything in one go): 

For standard auditability reasons, the build log should include the hashes (and/or signatures ) of the artifacts.

lukpueh:
Damn, I think can't re-open the PR after having deleted the remote branch in my fork, even though I have re-pushed it. Will need to check. Otherwise, I'll create a new one migrating the key points in the discussion here.

PR #1936, opened 2022-04-01
tests: Small refactor of a test

### LINKED ISSUE #1753: tests: test_update_root_new_root_fail_threshold_verification is misleading

jku:
Test was supposed to test a threshold that is higher than number of
signatures, but it actually was just using completely unsigned metadata.

This still doesn't test the case where _trusted_ metadata defines a
threshold that new metadata does not reach: only the case where new
metadata defines threshold that it does not meet (this case is covered
in updater tests though).

---

Fixes #1753

### DIFF
--- a/tests/test_trusted_metadata_set.py
+++ b/tests/test_trusted_metadata_set.py
@@ -240,12 +240,14 @@ class TestTrustedMetadataSet(unittest.TestCase):
         self.trusted_set.update_root(root)
 
     def test_update_root_new_root_fail_threshold_verification(self) -> None:
-        # new_root data with threshold which cannot be verified.
-        root = Metadata.from_bytes(self.metadata[Root.type])
-        # remove root role keyids representing root signatures
-        root.signed.roles[Root.type].keyids.clear()
+        # Increase threshold in new root, do not add enough keys
+        def root_threshold_bump(root: Root) -> None:
+            root.version += 1
+            root.roles[Root.type].threshold += 1
+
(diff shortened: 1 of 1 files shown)


tests/test_trusted_metadata_set.py
```diff
@@ -240,12 +240,14 @@ def root_new_version_modifier(root: Root) -> None:
         self.trusted_set.update_root(root)
 
     def test_update_root_new_root_fail_threshold_verification(self) -> None:
-        # new_root data with threshold which cannot be verified.
-        root = Metadata.from_bytes(self.metadata[Root.type])
-        # remove root role keyids representing root signatures
-        root.signed.roles[Root.type].keyids.clear()
```
lukpueh, line 246: Not really relevant anymore, but I don't understand why the old test didn't do `root.signatures.clear()` here.

lukpueh (approved):
LGTM

PR #1941, opened 2022-04-04
Add tricky test case blog post

### LINKED ISSUE #1885: blog: "how we test ngclient compliance"
Assuming we find a working blog platform (#1884), we should start publishing things.

On idea I had was _"Testing a tricky edge case in a TUF client"_ (very much a working title)
  - show how cool RepositorySimulator is and what it does
  - document usage of `--dump` to see the "live" test repository state for debugging purposes
  - document a specific edge case in TUF spec so we can refer to it later

Maybe that's too much content for one post already... I ended with that list because I'm specifically thinking of one of the cases in #1707 : these are really tricky corner cases and the tooling we have makes them testable -- but only if you know about the test components and how to work with Metadata

assigning to ivana: it's your PR so maybe you're interested

ivanayov:
This blog post explains details around the use of respository
simulator, `--dump` option and test cases with expired metadata

Fixes #1885

### DIFF
--- a/docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
+++ b/docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
@@ -1,0 +1,120 @@
+---
+title: "Testing Tricky Edge Cases in a TUF Client"
+author: Ivana Atanasova
+---
+
+Usually the TUF Specification creates an impression of simple and straightforward approach to address software update systems security gaps. In the next few paragraphs we'll try to convince you that the devil is in the details.
+
+With the [v1.0.0 release](https://blogs.vmware.com/opensource/2022/02/22/python-tuf-reaches-version-1-0-0/) we can say that the current reference implementation is finally in a good place, although it wouldn’t be so trustworthy without all the awesome test functionality it provides. Therein lies some interesting surprises, for the conformance tests reflect use cases and tricky details that wouldn’t easily come to mind. TUF, in fact, is capable of managing some tricky business!
+
+Before looking into them, let’s first introduce the test functionality itself.
+
+## Some repository simulator magic
+
+The test suite is heavily based on [RepositorySimulator](https://github.com/theupdateframework/python-tuf/blob/develop/tests/repository_simulator.py), which allows you to play with repository metadata by modifying it, signing and storing new roles versions, while serving older ones in the client test code. You can also simulate downloading new metadata from a remote without the need of file access or network connections, and modify expiry dates and time.
+
+Even though `RepositorySimulator` hosts repos purely in memory, you can supply the `--dump` flag to write its contents to a temporary directory on the local filesystem with "/metadata/..." and "/targets/..." URL paths that host metadata and targets respectively in order to audit the metadata. The test suite provides you with the ability to see the "live" test repository state for debugging purposes.
+
+Let’s cite a specific example with testing expired metadata to demonstrate the cool thing the `RepositorySimulator` provides, i.e. the capability to simulate real repository chains of updates as suggested by the spec, and not just modify individual metadata.
+
+More specifically, we would like to simulate a workflow in which a [targets](https://theupdateframework.github.io/specification/latest/#targets) version is being increased and a [timestamp](https://theupdateframework.github.io/specification/latest/#timestamp) expiry date is being changed. We are going to elaborate below on how this can be used to test the `Updater` above all programmatically. Now, let's just focus on how to verify that the `RepositorySimulator` did what we expected.
+
+Let's assume we did the following:
+* Upgraded `targets` to v2
+* Changed `timestamp` v2 expiry date
+
+We can verify that the metadata looks as expected, without the need to implement file access.
+
+First, we need to find the corresponding temporary directory:
+```
+$ python3 test_updater_top_level_update.py TestRefresh.test_expired_metadata --dump
+Repository Simulator dumps in /var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_
+```
+
+Once we know it, we can verify that the metadata has 2 cached versions:
+
+```
+$ tree /var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_/test_expired_metadata
+/var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_/test_expired_metadata
+├── 1
+│   ├── 1.root.json
+│   ├── snapshot.json
+│   ├── targets.json
+│   └── timestamp.json
+└── 2
+    ├── 2.root.json
+    ├── snapshot.json
+    ├── targets.json
+    └── timestamp.json
+```
+
+And now we can also see that after bumping the version and moving timestamp v2 expiry date two weeks forward from v1, the v2 corresponding timestamp metadata has recorded that expiry date correctly:
+
+Timestamp v1:
+<pre><code>$ cat /var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_/test_expired_metadata/1/timestamp.json 
+{
+ "signatures": [{...}],
+ "signed": {
+  "_type": "timestamp",
+  <b>"expires": "2022-03-30T00:18:31Z"</b>,
+  "meta": { "snapshot.json": {"version": 1}},
+  "spec_version": "1.0.28",
+  "version": 1
+ }}
+</code></pre>
+
+Timestamp v2:
+
+<pre><code>$ cat /var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_/test_expired_metadata/2/timestamp.json 
+{
+ "signatures": [{...}],
+ "signed": {
+  "_type": "timestamp",
+  <b>"expires": "2022-04-13T00:18:31Z"</b>,
+  "meta": { "snapshot.json": {"version": 2}},
+  "spec_version": "1.0.28",
+  "version": 2
+ }}
+</code></pre>
+
+As you can see, the first date is 30 Mar and the second - 13 Apr, which is exactly 14 days later. This is a great way to observe what the tests really do and check if they do it successfully.
+
+## When we talk about security, edge cases are the norm
+
+Now, let’s take a closer look at two edge cases, using in this test the cool things the RepositorySimulator provides:
+
+### Example with expired metadata:
+
+Imagine that we have performed an update and stored metadata in a cache. And the locally stored timestamp/snapshot has expired. But we still need it to perform an update from remote by verifying the signatures and we need to use the expired timestamp.
+
+We can play with versions and expiry to verify that this scenario not explicitly mentioned in the spec works correctly and safely. By using the simulator, we can do the following:
+1. Set the timestamp expiry one week ahead (to day 7)
+2. On the very first day (day 0) download, verify, and load metadata for the [top-level roles](https://theupdateframework.github.io/specification/latest/#roles-and-pki) following the TUF specification order. This is done by simply calling `updater.refresh()`.
+3. Then we bump [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) versions to v2 in the repository on the same day (day 0)
+4. Set v2 expiry dates three weeks ahead (to day 21)
+5. Travel in time somewhere between day 7 and day 21
+6. Perform a successful `refresh` (with `updater.refresh()` call) with the expired locally cached timestamp
+7. Check that the final repository version of the snapshot and targets roles is v2.
+
+This is a not so obvious use-case to keep in mind when thinking about updates. You can see how it looks in practice in the [reference implementation](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_metadata).
+
+### Example rollback protection check with expired metadata:
+
+Now let’s see if a rollback attack protection can be performed when the local timestamp has expired. In this case we need at least two timestamp and snapshot versions, an expired older version of timestamp, and a verification that a rollback check is performed with the old version.
+
+For a timestamp rollback, the case is pretty similar to the use of expired metadata. We can do the following:
+1. Set timestamp v1 expiry one week ahead (to day 7)
+2. Perform `updater.refresh()` on the very first day
+3. Publish timestamp v2 in the repository with expiry three weeks ahead (to day 21)
+4. Perform `updater.refresh()` somewhere between day 7 and day 21
+5. Verify that rollback check uses the expired timestamp v1. (For reference, see the implementation [example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_version_rollback)).
+
+A similar approach can be used when testing both timestamp and snapshot rollback protection. We just need to guarantee that after the last snapshot update, the snapshot version is not the latest in order to verify a rollback check is performed both with expired timestamp and an older snapshot. Sounds complicated, but it’s pretty easy with the simulator and [this example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_snapshot_rollback) illustrates it pretty well. 
+
+
+## The devil is in the details
+
+One of the great things about a reference implementation is that one can learn a lot about the TUF specification by looking at the tests, which are full of examples that would hardly come to mind when you read the abstract straightforward workflow explained in the spec. And those tests most likely do not cover everything… 
+
+Do you have a comment about the TUF spec or the cited examples? An idea? Please share it with us!
+


docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "Testing a tricky edge case in a TUF client"
+author: Ivana Atanasova
+---
+
+Usually the TUF Specification creates an impression of simple and straightforward approach to address software update systems security gaps. In the next few paragraphs we'll try to convince you that the devil is in the details.
```
MVrachev, line 6: I am intrigued in hearing from the rest of the team if they think the specification is simple and straightforward. :D
Maybe it's not bad in using a little advertainment language.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,90 @@
...
+  "version": 2
+ }}
+</code></pre>
+
+## Edge cases are a norm when we talk about security
+
+The TUF spec says the following:
+“Record the time at which the update began as the fixed update start time. Time is fixed at the beginning of the update workflow to allow an application using TUF to effectively pause time, in order to ensure that metadata which has a valid expiration time at the beginning of an update does not fail an expiration check later in the update workflow.”
+
+Now imagine that we have performed an update and stored metadata in cache. And the locally stored timestamp/snapshot has expired. But we still need it to perform an update from remote by verifying the signatures and we need to use the expired timestamp. Here the repository simulator can do a great job, as we’d like to somehow update expiry without explicitly touching the file system from the tests.
```
MVrachev:
```suggestion
Now imagine that we have performed an update and stored metadata in a cache. And the locally stored timestamp/snapshot has expired. But we still need it to perform an update from remote by verifying the signatures and we need to use the expired timestamp. Here the repository simulator can do a great job testing this scenario, as we’d like to somehow update expiry without explicitly touching the file system from the tests.
```
I think it should be underlined clearer that the `RepositorySimulator` can be useful for testing this scenerio, not actually performing it.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,90 @@
...
+
+We can play with versions and expiry to verify that this scenario not explicitly mentioned in the spec works correctly and safely. By using the simulator, we can set the timestamp expiry one week ahead. Then on the very first day we download, verify, and load metadata for the [top-level roles](https://theupdateframework.github.io/specification/latest/#roles-and-pki) following the specified order and implementing all the checks required in the TUF client workflow (which we simply call a refresh). Then we bump [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) versions to v2 on the same day and set v2 expiry dates three weeks ahead. If we travel in time somewhere between day 7 and day 21, we’d like to perform a successful refresh with the expired locally cached timestamp and the final version of the snapshot and targets roles to be v2.
```
MVrachev: Can you enumerate the steps you are describing here and split them one by one, so they could be easier to follow?
It's a little hard for me to follow.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,90 @@
...
+This is a not so obvious use-case to keep in mind when thinking about updates. You can see how it looks in practice in the [reference implementation](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_metadata).
+
+Now let’s see if a rollback attack protection can be performed when the local timestamp has expired. In this case we need at least two timestamp and snapshot versions, an expired older version of timestamp, and a verification that a rollback check is performed with the old version.
+For a timestamp rollback, the case is pretty similar to the use of  expired metadata. We can set timestamp v1 expiry one week ahead, perform refresh on the very first day, publish timestamp v2 with expiry three weeks ahead and perform a refresh somewhere between day 7 and day 21. All we need is to verify that rollback check uses the expired timestamp v1. (For reference, see the implementation [example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_version_rollback)).
```
MVrachev: Maybe you add a new space to actually divide the two paragraphs because markdown merges them as if there is no newline between lines number 83 and 84?

MVrachev: Again as above, I recommend using numbers and writing them in separate steps.

MVrachev (changes requested):
Good blogpost!
I have a couple of suggestions.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
ivanayov: The key is in "creates an impression" :D

ivanayov:
Thank you @MVrachev ! I've pushed few updates.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Example 1:
+We can play with versions and expiry to verify that this scenario not explicitly mentioned in the spec works correctly and safely. By using the simulator, we can do the following:
+* Set the timestamp expiry one week ahead (Day 7)
+* On the very first day download, verify, and load metadata for the [top-level roles](https://theupdateframework.github.io/specification/latest/#roles-and-pki) following the specified order and implementing all the checks required in the TUF client workflow (which we simply call a refresh)
+* Then we bump [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) versions to v2 on the same day and set v2 expiry dates three weeks ahead (Day 21)
+
+Then if we travel in time somewhere between Day 7 and Day 21, we’d like to perform a successful refresh with the expired locally cached timestamp and we want the final version of the snapshot and targets roles to be v2.
```
MVrachev: I think you can make the last sentence bullets as well and use numbers instead of just bullets.
Then you can add comments referencing the steps.
Also, I think you can split some of the bulltes.
For example, you can do something like this:
```suggestion
1. Set the timestamp expiry one week ahead (day 7).
2. On day 1 download, verify, and load metadata for the [top-level roles](https://theupdateframework.github.io/specification/latest/#roles-and-pki) following the TUF specification order. This is done by simply calling `RespositorySimulator.refresh()`.
3.  On the same day, set [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) expiration date three weeks ahead (day 21).
4. Bump [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) versions to v2 on the same day.
5. Travel in time somewhere between day 7 and day 21.
6. Perform a successful `refresh` (with `RepositorySimulator.refresh()` call).
7. Check that the final version of [snapshot].(https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) is 2.

Keep in mind that when performing the second `refresh` on step 6 [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) has expired, but still, we manage to make a successful refresh to version 2.
```

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,104 @@
...
+  "version": 2
+ }}
+</code></pre>
+
+## Edge cases are a norm when we talk about security
+
+The TUF spec says the following:
+“Record the time at which the update began as the fixed update start time. Time is fixed at the beginning of the update workflow to allow an application using TUF to effectively pause time, in order to ensure that metadata which has a valid expiration time at the beginning of an update does not fail an expiration check later in the update workflow.”
+
+Now imagine that we have performed an update and stored metadata in a cache. And the locally stored timestamp/snapshot has expired. But we still need it to perform an update from remote by verifying the signatures and we need to use the expired timestamp. Here the repository simulator can do a great job testing this scenario, as we’d like to somehow update expiry without explicitly touching the file system from the tests.
```
MVrachev: I think this should be moved under `### Example 1:` or line 80.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,104 @@
...
+### Example 2:
+Now let’s see if a rollback attack protection can be performed when the local timestamp has expired. In this case we need at least two timestamp and snapshot versions, an expired older version of timestamp, and a verification that a rollback check is performed with the old version.
+
+For a timestamp rollback, the case is pretty similar to the use of expired metadata. We can do the following:
+* Set timestamp v1 expiry one week ahead (Day 7)
+* Perform refresh on the very first day
+* Publish timestamp v2 with expiry three weeks ahead (Day 21)
+* Perform a refresh somewhere between Day 7 and Day 21
+
+All we need is to verify that rollback check uses the expired timestamp v1. (For reference, see the implementation [example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_version_rollback)).
```
MVrachev: I think it will be better if you move this sentence to the bullet points.
With the usage of bullet points (or numbers), one will expect that all of the test steps are described, but here they are not. There is one additional step to check which timestamp is used which is mentioned here.

MVrachev (commented):
I like that you moved the examples under their own headers and that you simplified them with bullets.
Still, I think it could be simplified more.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
ivanayov: Thank you! I like your proposal, looks easier to understand.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
ivanayov: This is an introduction and is valid for all examples. May be I should remove the `### Example` headers, or rename them. I it just goes under Ex. 1, then Ex. 2 would be missing some context.

Thanks for noticing.

ivanayov:
Pushed some wording updates.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
+---
+title: "Testing Tricky Edge Cases in a TUF Client"
+author: Ivana Atanasova
+---
+
+Usually the TUF Specification creates an impression of simple and straightforward approach to address software update systems security gaps. In the next few paragraphs we'll try to convince you that the devil is in the details.
+
+With the [v1.0.0 release](https://blogs.vmware.com/opensource/2022/02/22/python-tuf-reaches-version-1-0-0/) we can say that the current reference implementation is finally in a good place, although it wouldn’t be so trustworthy without all the awesome test functionality it provides. Therein lies some interesting surprises, for the test coverage reflects use cases that wouldn’t respectively come to mind. TUF, in fact, is capable of managing some tricky business! And that’s why it provides more than just test coverage - it is a way to analyze and validate that the implementation strictly follows the TUF spec with all its corner cases and tricky details that are far from obvious.
```
lukpueh: I am unsure if the word "respectively" fits here (cc @jhdalek55).

lukpueh: Could you try to rephrase:
> TUF, in fact, is capable of managing some tricky business! And that’s why it provides more than just test coverage - it is a way to analyze and validate that the implementation strictly follows the TUF spec with all its corner cases and tricky details that are far from obvious.

 "managing some tricky business" reads a bit overly casual and vague. Also, I don't understand why the consequence is that "it provides more than just test coverage". What is "it" and what do you mean by "provide test coverage"?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+Before looking into them, we need to introduce the test functionality itself.
+
+
+## Some Repository Simulator Magic
+
+The great thing about [RepositorySimulator](https://github.com/theupdateframework/python-tuf/blob/develop/tests/repository_simulator.py), is it allows you to play with repository metadata by modifying it, signing and storing new roles versions, while serving older ones in the client test code. You can also simulate downloading new metadata from a remote without the need of file access or network connections, and modify expiry dates and time.
```
lukpueh: Premature comma: 
```suggestion
The great thing about [RepositorySimulator](https://github.com/theupdateframework/python-tuf/blob/develop/tests/repository_simulator.py) is, it allows you to play with repository metadata by modifying it, signing and storing new roles versions, while serving older ones in the client test code. You can also simulate downloading new metadata from a remote without the need of file access or network connections, and modify expiry dates and time.
```

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+---
+
+Usually the TUF Specification creates an impression of simple and straightforward approach to address software update systems security gaps. In the next few paragraphs we'll try to convince you that the devil is in the details.
+
+With the [v1.0.0 release](https://blogs.vmware.com/opensource/2022/02/22/python-tuf-reaches-version-1-0-0/) we can say that the current reference implementation is finally in a good place, although it wouldn’t be so trustworthy without all the awesome test functionality it provides. Therein lies some interesting surprises, for the test coverage reflects use cases that wouldn’t respectively come to mind. TUF, in fact, is capable of managing some tricky business! And that’s why it provides more than just test coverage - it is a way to analyze and validate that the implementation strictly follows the TUF spec with all its corner cases and tricky details that are far from obvious.
+Before looking into them, we need to introduce the test functionality itself.
+
+
+## Some Repository Simulator Magic
+
```
lukpueh, line 13: I wouldn't start this sentence with "The great thing about RepositorySimulator is ...". The reader doesn't know yet what RepositorySimulator is and why they should care about it. Maybe you can set the stage for the tool before you praise it. Something along the lines of "The test suite is heavily based on RepositorySimulator, which allows you to ..."

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+The great thing about [RepositorySimulator](https://github.com/theupdateframework/python-tuf/blob/develop/tests/repository_simulator.py), is it allows you to play with repository metadata by modifying it, signing and storing new roles versions, while serving older ones in the client test code. You can also simulate downloading new metadata from a remote without the need of file access or network connections, and modify expiry dates and time.
+
+When you execute a test in python-tuf, it creates a local temporary directory serving for cache, with "/metadata/..." and "/targets/..." URL paths that host metadata and targets respectively.
```
lukpueh: When you execute a test ... **with repository simulator**?

Also, "cache" seems a bit like an understatement. RepositorySimulator simulates a TUF repository (for metadata and target files), right?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+
+The test suite provides you with the ability to see the "live" test repository state for debugging purposes. You just need to use the `--dump` option, so that you can see the current tests corresponding path.
```
lukpueh: I'd remove the second sentence here. It's a usage detail, which you explain below anyway.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+5. Verify that rollback check uses the expired timestamp v1. (For reference, see the implementation [example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_version_rollback)).
+
+A similar approach can be used when testing both timestamp and snapshot rollback protection. We just need to guarantee that after the last snapshot update, the snapshot version is not the latest in order to verify a rollback check is performed both with expired timestamp and an older snapshot. Sounds complicated, but it’s pretty easy with the simulator and [this example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_snapshot_rollback) illustrates it pretty well. 
+
+
+## The Devil is in the Details
+
+One of the great things about a reference implementation is that one can learn a lot about the TUF specification by looking at the tests, which are full of examples that would hardly come to mind when you read the illusory straightforward workflow explained in the spec. And those tests most likely do not cover everything…
```
lukpueh: Maybe say "abstract" instead of "illusory"?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+ }}
+</code></pre>
+
+As you can see, the first date is 30 Mar and the second - 13 Apr, which is exactly 14 days later. This is a great way to observe what the tests really do and check if they do it successfully.
+
+## When We Talk about Security, Edge Cases are the Norm
+
+Now, let’s take a closer look at two edge cases, using in this test the two cool things the RepositorySimulator provides:
+
+The TUF spec says the following:
```
lukpueh: Can you add a bit of context, when/where/why the TUF spec says this?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+
+The great thing about [RepositorySimulator](https://github.com/theupdateframework/python-tuf/blob/develop/tests/repository_simulator.py), is it allows you to play with repository metadata by modifying it, signing and storing new roles versions, while serving older ones in the client test code. You can also simulate downloading new metadata from a remote without the need of file access or network connections, and modify expiry dates and time.
+
+When you execute a test in python-tuf, it creates a local temporary directory serving for cache, with "/metadata/..." and "/targets/..." URL paths that host metadata and targets respectively.
+
+The test suite provides you with the ability to see the "live" test repository state for debugging purposes. You just need to use the `--dump` option, so that you can see the current tests corresponding path.
+
+Let’s cite a specific example with testing expired metadata to demonstrate two of the cool things the RepositorySimulator provides:
+* Upgrade [targets](https://theupdateframework.github.io/specification/latest/#targets) to v2
+* Change [timestamp](https://theupdateframework.github.io/specification/latest/#timestamp) v2 expiry date
```
lukpueh: I think the cool thing here is the bigger picture, i.e. the whole repository update story `test_expired_metadata` tells. 

In other words, I would try to emphasize on RepositorySimulator's capability to simulate real repository chains of updates as suggested by the spec, and not just modify individual metadata.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+
+So how can we verify that the metadata looks as expected, without the need to implement file access?
```
lukpueh: I feel like you are putting a lot of emphasis on the `--dump` feature, which is cool but not the core of the tool. I'd talk more about the specific workflow that RepositorySimulator simulates here and how that can be used to test the Updater above all programmatically, and only then mention how we can pass the handy `--dump` flag to audit what the simulator is doing.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+  <b>"expires": "2022-04-13T00:18:31Z"</b>,
+  "meta": { "snapshot.json": {"version": 2}},
+  "spec_version": "1.0.28",
+  "version": 2
+ }}
+</code></pre>
+
+As you can see, the first date is 30 Mar and the second - 13 Apr, which is exactly 14 days later. This is a great way to observe what the tests really do and check if they do it successfully.
+
+## When We Talk about Security, Edge Cases are the Norm
```
lukpueh: Minor nit: Please choose a consistent [capitalization style](https://en.wikipedia.org/wiki/Capitalization#Names_of_capitalization_styles) (also review other titles)

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+
+Now, let’s take a closer look at two edge cases, using in this test the two cool things the RepositorySimulator provides:
+
+The TUF spec says the following:
+
+“Record the time at which the update began as the fixed update start time. Time is fixed at the beginning of the update workflow to allow an application using TUF to effectively pause time, in order to ensure that metadata which has a valid expiration time at the beginning of an update does not fail an expiration check later in the update workflow.” Let's see the following examples where this is not so obvious.
+
+Let's consider the following examples where this is not so obvious.
```
lukpueh: Duplicate sentence.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Example with expired metadata:
+
+Imagine that we have performed an update and stored metadata in a cache. And the locally stored timestamp/snapshot has expired. But we still need it to perform an update from remote by verifying the signatures and we need to use the expired timestamp. Here the repository simulator can do a great job testing this scenario, as we’d like to somehow update expiry without explicitly touching the file system from the tests.
+
+We can play with versions and expiry to verify that this scenario not explicitly mentioned in the spec works correctly and safely. By using the simulator, we can do the following:
+1. Set the timestamp expiry one week ahead (Day 7)
+2. On the very first day download, verify, and load metadata for the [top-level roles](https://theupdateframework.github.io/specification/latest/#roles-and-pki) following the TUF specification order. This is done by simply calling `RespositorySimulator.refresh()`.
```
lukpueh: You mean `updater.refresh`, right? (same below)

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
lukpueh: Having read the examples below, I don't understand why we need this paragraph from the spec here.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+3. Then we bump [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) versions to v2 on the same day
+4. Set v2 expiry dates three weeks ahead (Day 21)
+5. Travel in time somewhere between Day 7 and Day 21
+6. Perform a successful `refresh` (with `RepositorySimulator.refresh()` call) with the expired locally cached timestamp
+7. Check that the final version of the snapshot and targets roles is v2.
+This is a not so obvious use-case to keep in mind when thinking about updates. You can see how it looks in practice in the [reference implementation](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_metadata).
```
lukpueh, line 99: I must say, the docstring from `test_expired_metadata` describes this chain of events slightly clearer:
- it mentions what happens in the repo and what happens on the client
- it mentions for each event when it happens in a uniform way

lukpueh: (same applies in example below)

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,117 @@
...
+
+For a timestamp rollback, the case is pretty similar to the use of expired metadata. We can do the following:
+1. Set timestamp v1 expiry one week ahead (Day 7)
+2. Perform refresh on the very first day
+3. Publish timestamp v2 with expiry three weeks ahead (Day 21)
+4. Perform a refresh somewhere between Day 7 and Day 21
+5. Verify that rollback check uses the expired timestamp v1. (For reference, see the implementation [example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_version_rollback)).
+
+A similar approach can be used when testing both timestamp and snapshot rollback protection. We just need to guarantee that after the last snapshot update, the snapshot version is not the latest in order to verify a rollback check is performed both with expired timestamp and an older snapshot. Sounds complicated, but it’s pretty easy with the simulator and [this example](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py#:~:text=test_expired_timestamp_snapshot_rollback) illustrates it pretty well. 
+
```
lukpueh, line 113: Are these two examples really an edge cases? Aren't they (vaguely) defined by the client workflow in the spec?

I do agree that they are tricky to implement in a client and tricky to test, and that the RepositorySimulator does a great job in achieving the latter.

lukpueh (changes requested):
Great blog post, @ivanayov! I like the overall structure and I mostly like the casual/fun style of writing. :) 

Let me know if any of my comments are unclear.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
jku: oh yeah this is actually incorrect: RepositorySimulator does not create directories, it serves everything from memory

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
jku: I probably described it as an edge case :grimacing: with the logic that 
* we want to ensure that we only use valid metadata
* expired metadata is not valid
* **except** when were validating new metadata with an older metadata (in this case specifically old metadata can be expired but must be otherwise valid)

this is defined in spec, that's true

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
ivanayov: so probably it should stay just "in python-tuf"?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
ivanayov: Do you think it's better to cite the docstring, or link to it, or cite and elaborate?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
lukpueh: This could actually make for a good transition to advertising the `--dump` flag. Something along the lines of, "Although, RepositorySimulator hosts repos purely in memory, you can supply the `--dump` flag to write its contents to a temporary directory on the local filesystem in order to audit the metadata."  (feel free to rephrase as you see fit)

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
lukpueh: Cite and elaborate probably. :) I just wanted to point out that currently it is not immediately obvious if something happens in the repo or on the client. If you polish that a bit, I'm happy. My second comment about the time is less important.

ivanayov:
Thank you for all the useful feedback, @lukpueh and @jku !

I need to think once again about the examples, so will write a message after I push an update for your last 2 comments.

My be, if those are not explained as edge cases, the whole message should be changed to just showing a test examples, which is fine. I still think that being defined in the spec does not necessarily contradict to being an edge case, as the test gives more detail to the specific workflow, which is not very straightforward, and the spec doesn't talk directly about the connection between local metadata and expiry dates. That connection follows rather implicitly after a deeper review.

ivanayov:
Updated the examples to distinguish repo from client.

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,113 @@
+---
+title: "Testing Tricky Edge Cases in a TUF Client"
+author: Ivana Atanasova
+---
+
+Usually the TUF Specification creates an impression of simple and straightforward approach to address software update systems security gaps. In the next few paragraphs we'll try to convince you that the devil is in the details.
+
+With the [v1.0.0 release](https://blogs.vmware.com/opensource/2022/02/22/python-tuf-reaches-version-1-0-0/) we can say that the current reference implementation is finally in a good place, although it wouldn’t be so trustworthy without all the awesome test functionality it provides. Therein lies some interesting surprises, for the test coverage reflects use cases that wouldn’t easily come to mind. TUF, in fact, is capable of managing some tricky business! And that’s why it provides more than just test coverage - it is a way to analyze and validate that the implementation strictly follows the TUF spec with all its corner cases and tricky details that are far from obvious.
+Before looking into them, we need to introduce the test functionality itself.
```
lukpueh: I still don't fully understand what you are trying to say here (see https://github.com/theupdateframework/python-tuf/pull/1941#discussion_r871088189).

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,113 @@
...
+
+Let’s cite a specific example with testing expired metadata to demonstrate the cool thing the `RepositorySimulator` provides, i.e. the capability to simulate real repository chains of updates as suggested by the spec, and not just modify individual metadata.
+
+More specifically, we would like to simulate a workflow in which a [targets](https://theupdateframework.github.io/specification/latest/#targets) version is being increased and a [timestamp](https://theupdateframework.github.io/specification/latest/#timestamp) expiry date is being changed. We are going to elaborate in more detail in the sections below about an actual use-case this can be used to test the `Updater` above all programmatically. Now, let's just focus on how to verify that the `RepositorySimulator` did what we expected.
```
lukpueh: Minor wording nit: "...elaborate in more detail in the sections below about an actual use-case this can be used to test..." reads a bit convoluted. What about "...elaborate below on how this can be used to test ..."?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,113 @@
...
+  "spec_version": "1.0.28",
+  "version": 2
+ }}
+</code></pre>
+
+As you can see, the first date is 30 Mar and the second - 13 Apr, which is exactly 14 days later. This is a great way to observe what the tests really do and check if they do it successfully.
+
+## When we talk about security, edge cases are the norm
+
+Now, let’s take a closer look at two edge cases, using in this test the two cool things the RepositorySimulator provides:
```
lukpueh: What are "the two cool things the RepoSimulator provides"? Also, further above you say "demonstrate the cool thing the RepositorySimulator provides". Does it provide one or two cool things now? :P

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,113 @@
...
+Once we know it, we can verify that the metadata has 2 cached versions:
+
+```
+$ ls /var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_/test_expired_metadata/
+1	2
+```
+```
+$ ls /var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_/test_expired_metadata/1
+1.root.json	snapshot.json	targets.json	timestamp.json
+```
```
lukpueh, line 49: Just an idea: Maybe it would be worth to `tree /var/folders/pr/b0xyysh907s7mvs3wxv7vvb80000gp/T/tmpzvr5xah_/test_expired_metadata`, instead of `ls`ting it twice?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,113 @@
...
+
+As you can see, the first date is 30 Mar and the second - 13 Apr, which is exactly 14 days later. This is a great way to observe what the tests really do and check if they do it successfully.
+
+## When we talk about security, edge cases are the norm
+
+Now, let’s take a closer look at two edge cases, using in this test the two cool things the RepositorySimulator provides:
+
+### Example with expired metadata:
+
+Imagine that we have performed an update and stored metadata in a cache. And the locally stored timestamp/snapshot has expired. But we still need it to perform an update from remote by verifying the signatures and we need to use the expired timestamp. Here the repository simulator can do a great job testing this scenario, as we’d like to somehow update expiry without explicitly touching the file system from the tests.
```
lukpueh: Why is it important to not touch the file system?

docs/_posts/testing_a_tricky_edge_case_in_a_tuf_client.md
```diff
@@ -0,0 +1,113 @@
...
+
+We can play with versions and expiry to verify that this scenario not explicitly mentioned in the spec works correctly and safely. By using the simulator, we can do the following:
+1. Set the timestamp expiry one week ahead (to day 7)
+2. On the very first day (day 0) download, verify, and load metadata for the [top-level roles](https://theupdateframework.github.io/specification/latest/#roles-and-pki) following the TUF specification order. This is done by simply calling `updater.refresh()`.
+3. Then we bump [snapshot](https://theupdateframework.github.io/specification/latest/#update-snapshot) and [targets](https://theupdateframework.github.io/specification/latest/#targets) versions to v2 in the repository on the same day (day 0)
+4. Set v2 expiry dates three weeks ahead (to day 21)
+5. Travel in time somewhere between day 7 and day 21
+6. Perform a successful `refresh` (with `updater.refresh()` call) with the expired locally cached timestamp
+7. Check that the final repository version of the snapshot and targets roles is v2.
```
lukpueh, line 97: Minor format nit: I think you want a blank line before the next sentence, otherwise it looks like the summary is part of step 7.

lukpueh (commented):
Thanks for addressing my comments! This version looks a lot better. I still have a few comments, but nothing too important.

lukpueh (approved):
Thanks for addressing all my comments. I think this is good to go live!

Issue #529, opened 2017-11-10
Add in-toto metadata to releases to validate the integrity of its software supply chain


vladimir-v-diaz:
**Description of issue or feature request**:
Project releases should include [in-toto](https://in-toto.github.io/) metadata that can be used to validate the integrity of the project's software supply chain.

**Current behavior**:
Developer signatures can be provided for each release of the project, both on GitHub and PyPI.  However, these signatures do not guarantee that some part of the source->release process was compromised.

**Expected behavior**:
The packaged release should include metadata and a way to verify that the project was packaged as intended.  All steps of the source->release procedure should be properly signed and confirmed to be valid, as defined by the project developers.

jku:
I'm going to remove "good first issue": The description may be clear to an in-toto expert but as an example I wouldn't have any idea where to start implementing this.

Also editing the title to what I think the suggestion is

lukpueh:
With python-tuf builds becoming reproducible (see https://github.com/theupdateframework/python-tuf/issues/1269) we can provide multiple in-toto links for any given release build each signed with a different maintainer key, and create a corresponding in-toto layout that encodes the key authorization and a signature threshold requirement. 

See [`apt-transport-in-toto`](https://github.com/in-toto/apt-transport-in-toto) for a detailed description of this scenario (note, the tool deals with Debian packages and therefor includes a lot of code that hooks into `apt`, but the in-toto metadata scaffolding would be alike for Python wheels)

PR #1946, opened 2022-04-07
Add GH workflow to build and release on GH and PyPI

### LINKED ISSUE #1550: Automated release builds

lukpueh:
Fixes #1550
Supersedes #1933

Adds a new GitHub workflow (cd), which is triggered on a successful run of the automated tests (ci), but only if ci was triggered by a release tag push.

The new cd workflow builds wheel and sdist at the pushed tag, creates a preliminary GitHub release (X.Y.Z-rc), uploading the release artifacts as assets, and pauses for [review](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments), which may include locally running `verify_release` (see RELEASE.md for details). 

Upon approval, the release artifacts are published on PyPI and the Github release is finalized.
 
This PR further:
- updates RELEASE.md to match the new workflow
- updates `verify_release` script to allow intermediate review, in order to approve the deployment.
- (unrelated) configures tox to also lint `verify_release`

Please see commits messages for details!

**Demo**
See instructions in https://github.com/lukpueh/tuf/commit/82b7b122a99d0b39c0ddff7239ee33fcb391018d, also note that:
- `cd.yml` needs to be on your [default branch for the workflow to trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)
- Please don't publish vX.Y.Z versions for demo purposes. Although, we can remove them, we can't reuse the version numbers for actual release testing. (see [PEP 440](https://peps.python.org/pep-0440/#developmental-releases) for development release version numbers)

**TODOs** (likely in follow-up PRs)
-  (critical) re-enable release signing -- Prior to this PR RELEASE.md instructed maintainers to sign release artifacts with GPG as part of the manual release process. The [installation docs](https://theupdateframework.readthedocs.io/en/latest/INSTALLATION.html#verify-release-signatures) still mention release signatures. Some ideas to fix this inconsistency:
   - quick-fix 1: don't require signing (update docs)
   - quick-fix 2: sign in GitHub action
   - quick-fix 3: sign locally and upload signatures to release assets manually (integrate with `verify_release` script?)
   - long-term fix: e.g. in-toto ([#529](https://github.com/theupdateframework/python-tuf/issues/529))
- replace 3rd-party actions with custom code -- This PR adds two 3rd-party actions `softprops/action-gh-release` and `pypa/gh-action-pypi-publish`, which goes against our "only allow actions created by GitHub"-policy. It does not seem like a big deal, because the former action is popular and recommended by the archived GitHub-native `action/create-release`, and the latter is hosted by the pypa, which also hosts `twine`, the tool we usually use to pubish on pypi. Still, we should at least see how hard it would be to implement the required logic for those actions ourselves (see [custom finalize release action](https://github.com/lukpueh/tuf/blob/37cb272a437c2a2455a1172d74e57055c6056ae6/.github/workflows/cd.yml#L84-L94) in cd.yml for inspiration)
- make ci/cd setup more robust (see [code comment](https://github.com/lukpueh/tuf/blob/37cb272a437c2a2455a1172d74e57055c6056ae6/.github/workflows/cd.yml#L16-L25))
- replace upload/download artifacts with downloading artifacts from GitHub release assets (see [code comment](https://github.com/lukpueh/tuf/blob/37cb272a437c2a2455a1172d74e57055c6056ae6/.github/workflows/cd.yml#L55-L60))
- (nice to have) auto-generate release notes for GitHub release page (see [GH docs](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes))
- Configure release job to not add GitHub auto-created release artifacts ([requested by Jussi](https://github.com/theupdateframework/python-tuf/pull/1946#discussion_r845774429))
- CD workflow could have a permissions section

requirements**:

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,6 +11,8 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    outputs:
+      release_id: ${{ steps.gh-release.outputs.id }}
     steps:
       - name: Checkout release tag
         uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
@@ -26,14 +28,29 @@ jobs:
       - name: Build binary wheel and source tarball
         run: python3 -m build --sdist --wheel --outdir dist/ .
 
-      - name: Store build artifacts for review and release
+      - id: gh-release
+        name: Publish GitHub release candiate
+        uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5
+        with:
+          name: ${{ github.ref_name }}-rc
+          tag_name: ${{ github.ref }}
+          body: "Release waiting for review..."
+          files: dist/*
+
+      - name: Store build artifacts
+        # NOTE: The release job could download the assets from the GitHub release page,
+        # published in the previous step. But using the GitHub upload/download actions
+        # seems more robust as there is no need to compute download URLs.
+        # NOTE: (2) action-gh-release returns download URLSs as output, which could be
+        # propagated to next job along with release_id (see above)
+        # https://github.com/softprops/action-gh-release#outputs
         uses: actions/upload-artifact@6673cd052c4cd6fcf4b4e6e60ea986c889389535
         with:
           name: build-artifacts
           path: dist
 
-  release-on-pypi:
-    name: Release on PyPI
+  release:
+    name: Release
     runs-on: ubuntu-latest
     needs: build
     environment: release
@@ -49,3 +66,15 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
+
+      - name: Finalize GitHub release
+        uses: actions/github-script@9ac08808f993958e9de277fe43a64532a609130e
+        with:
+          script: |
+            await github.rest.repos.updateRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              release_id: '${{ needs.build.outputs.release_id }}',
+              name: '${{ github.ref_name }}',
+              body: 'See [CHANGELOG.md](https://github.com/'+ context.repo.owner +'/'+ context.repo.repo +'/blob/${{ github.ref_name }}/docs/CHANGELOG.md) for details.'
+            })
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,21 +1,35 @@
 name: CD
 concurrency: cd
 
-# Trigger workflow on release tag push
+# Trigger workflow on completed CI (further checks below)
 on:
-  push:
-    # TODO: Should we restrict to vX.Y.Z tags?
-    tags: v*
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
 
 jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    # Skip unless CI was successful and ran on a ref starting with 'v' (release tag)
+    if: ${{ github.event.workflow_run.conclusion == 'success' && startsWith(github.event.workflow_run.head_branch, 'v') }}
+    # NOTE: This works because we currently only trigger CI on a push to the 'develop'
+    # branch or a 'v*'-tag, but it seems rather brittle.
+    # Unfortunately, there is not much more info we get from the CI workflow
+    # ('workflow_run') than the ref name. No ref, ref_type, etc., so we don't even know
+    # if a tag or a branch was pushed. :(
+    # See https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run
+    # NOTE: (2) An alternative solution might be to restructure workflows, so that all
+    # test logic from 'ci.yml' is moved to a separate workflow file '_test.yml', that
+    # can be included in both CI (triggered on push to 'develop'-branch) and CD
+    # (triggered on push to 'v*'-tag) workflows.
     outputs:
       release_id: ${{ steps.gh-release.outputs.id }}
     steps:
       - name: Checkout release tag
         uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
+        with:
+          ref: ${{ github.event.workflow_run.head_branch }}
 
       - name: Set up Python
         uses: actions/setup-python@0ebf233433c08fb9061af664d501c3f3ff0e9e20
@@ -32,8 +46,8 @@ jobs:
         name: Publish GitHub release candiate
         uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5
         with:
-          name: ${{ github.ref_name }}-rc
-          tag_name: ${{ github.ref }}
+          name: ${{ github.event.workflow_run.head_branch }}-rc
+          tag_name: ${{ github.event.workflow_run.head_branch }}
           body: "Release waiting for review..."
           files: dist/*
 
@@ -75,6 +89,6 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               release_id: '${{ needs.build.outputs.release_id }}',
-              name: '${{ github.ref_name }}',
-              body: 'See [CHANGELOG.md](https://github.com/'+ context.repo.owner +'/'+ context.repo.repo +'/blob/${{ github.ref_name }}/docs/CHANGELOG.md) for details.'
+              name: '${{ github.event.workflow_run.head_branch }}',
+              body: 'See [CHANGELOG.md](https://github.com/'+ context.repo.owner +'/'+ context.repo.repo +'/blob/${{ github.event.workflow_run.head_branch }}/docs/CHANGELOG.md) for details.'
             })
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,7 +1,7 @@
 name: CD
 concurrency: cd
 
-# Trigger workflow on completed CI (further checks below)
+# Trigger workflow on any completed CI (see further checks below)
 on:
   workflow_run:
     workflows: [CI]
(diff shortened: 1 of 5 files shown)


docs/RELEASE.md
```diff
@@ -1,39 +1,29 @@
 # Release process
 
-* Ensure you have a backup of all working files and then remove files not tracked by git
-  `git clean -xdf`. **NOTE**: this will delete all files in the tuf tree that aren't
-  tracked by git
```
jku, line 5: removing this seems right, but there is a related issue: https://github.com/theupdateframework/python-tuf/issues/1942 (build results for verify_release are still affected by randome files in tree).

I can have a look at the workaround (making a clean checkout before running build).

.github/workflows/cd.yml
```diff
@@ -0,0 +1,94 @@
...
+        run: python3 -m pip install --upgrade pip build
+
+      - name: Build binary wheel and source tarball
+        run: python3 -m build --sdist --wheel --outdir dist/ .
+
+      - id: gh-release
+        name: Publish GitHub release candiate
+        uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5
+        with:
+          name: ${{ github.event.workflow_run.head_branch }}-rc
```
jku, line 42: so the release made here will have a "title" `vA.B.C-rc` but the download URLs will be the final ones already? This seems ok to me.

Does the release contain the annoying GitHub created "source code" zips as well or do they not get created with `action-gh-release`? (this is not important just wondering)

.github/workflows/ci.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - develop
+    tags:
+      # TODO: Should we restrict to vX.Y.Z tags?
+      - v*
```
jku: seems fine even without the TODO

docs/RELEASE.md
```diff
@@ -1,39 +1,29 @@
...
+6. Push the tag to GitHub `git push origin vA.B.C`
+
+  *A push triggers the [CI workflow](.github/workfows/ci.yml), which, on success, triggers
+  the [CD worfklow](.github/workfows/cd.yml), which builds source dist and wheel,
+  creates a preliminary GitHub release under `vA.B.C-rc`, and pauses for review.*
+
+7. Run `verify_release --skip-pypi` locally to make sure a build on your machine matches
+  the preliminary release artifacts published on GitHub.
+
+&#10132; [Review *deployemnt*](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments) on GitHub
```
jku:
```suggestion
&#10132; [Review *deployment*](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments) on GitHub
```

docs/RELEASE.md
```diff
@@ -1,39 +1,29 @@
...
+
+  *An approval resumes the CD workflow to publish the release on PyPI, and to finalize the
+  GitHub release (removse `-rc` suffix and updates release notes).*
```
jku:
```suggestion
  *Approval resumes the CD workflow to publish the release on PyPI, and to finalize the
  GitHub release (removes `-rc` suffix and updates release notes).*
```

docs/RELEASE.md
```diff
@@ -1,39 +1,29 @@
...
+
+8. `verify_release` may be used again to make sure the release artifacts PyPI.
```
jku:
```suggestion
8. `verify_release` may be used again to make sure the PyPI release artifacts match local build as well.
```

.github/workflows/cd.yml
```diff
@@ -0,0 +1,94 @@
...
+        uses: actions/upload-artifact@6673cd052c4cd6fcf4b4e6e60ea986c889389535
+        with:
+          name: build-artifacts
+          path: dist
+
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    needs: build
+    environment: release
```
jku, line 60: is `environment` only needed in one of the steps? Could move it there if that is the case

jku (changes requested):
This is really great
 * workflow triggering is indeed a bit hacky but I don't have better suggestions: looks good enough to me
 * CD workflow could have a `permissions` section but leaving that as future improvement seems ok
 * Can you document what needs to happen in project configuration? 
   * A "pypi" environment with an application secret that allows pushing releases is added?
   * There's a list of maintainers who are allowed to approve the use of the pypi environment
   * anything else?
   
left some minor comments in code

jku:
On your follow up work:
> replace upload/download artifacts with downloading artifacts from GitHub release assets

On second thought, the github release assets are modifiable by any maintainer I think. So maybe using the job artifacts is actually correct? :shrug:

docs/RELEASE.md
jku: made #1947 for this

lukpueh: Thanks, I considered leaving that line, but secretly hoped that we could find a different fix (`git clean -xdf` already killed my pre commit hook once). https://github.com/theupdateframework/python-tuf/pull/1947 looks great for this.

.github/workflows/cd.yml
lukpueh: > so the release made here will have a "title" `vA.B.C-rc` but the download URLs will be the final ones already? This seems ok to me.

Correct. The -rc suffix is only in the title. The URL uses the release tag, which is whatever's pushed.

> 
> Does the release contain the annoying GitHub created "source code" zips as well or do they not get created with `action-gh-release`? (this is not important just wondering)

Yes it does. I haven't checked if there's a way to prevent GitHub from doing this.  Will add a quick note in the list of TODO's above, which I intend to populate tickets from.

docs/RELEASE.md
lukpueh: 🤦 Apologies!

.github/workflows/cd.yml
lukpueh: Not sure what you mean. Environments work on the job level: From the [docs](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#using-an-environment):
> Each job in a workflow can reference a single environment. Any protection rules configured for the environment must pass before a job referencing the environment is sent to a runner.

lukpueh:
> On your follow up work:
> 
> > replace upload/download artifacts with downloading artifacts from GitHub release assets
> 
> On second thought, the github release assets are modifiable by any maintainer I think. So maybe using the job artifacts is actually correct? 🤷

Good thought. I had a feeling that it is more "contained",  but didn't think of access permissions. Again, ideally these jobs are protected using something like in-toto, which signs materials and products of each job, and links them together, so that nobody can change them in transit, but that's for a later iteration. I think we can go ahead with this PR with the job artifact feature currently used.

lukpueh:
Thanks for the thorough review and fixing my typos! Regarding your high-level comments:

> * workflow triggering is indeed a bit hacky but I don't have better suggestions: looks good enough to me

I think what I proposed in [this comment](https://github.com/lukpueh/tuf/blob/37cb272a437c2a2455a1172d74e57055c6056ae6/.github/workflows/cd.yml#L22-L25) is definitely better (@rdimitrov did something similar in theupdateframework/go-tuf#234). However, as long as we don't accept other triggers for ci.yml we are safe. So let's maybe add a ticket and a comment in ci.yml and go ahead for the time being?

> * CD workflow could have a `permissions` section but leaving that as future improvement seems ok

👍 

> * Can you document what needs to happen in project configuration?
>   
>   * A "pypi" environment with an application secret that allows pushing releases is added?
>   * There's a list of maintainers who are allowed to approve the use of the pypi environment
>   * anything else?

No that's basically it. I describe it in https://github.com/lukpueh/tuf/commit/82b7b122a99d0b39c0ddff7239ee33fcb391018d. Do you think this could go into RELEASE.md as prerequisite?


> 
> left some minor comments in code

❤️

.github/workflows/cd.yml
jku: I thought an environment could be exposed to just a single step in in a job, but reading through the syntax it seems cleat that's not true: I must have confused "environment" with "env".

rdimitrov:
> I think what I proposed in [this comment](https://github.com/lukpueh/tuf/blob/37cb272a437c2a2455a1172d74e57055c6056ae6/.github/workflows/cd.yml#L22-L25) is definitely better (@rdimitrov did something similar in [theupdateframework/go-tuf#234](https://github.com/theupdateframework/go-tuf/pull/234)). However, as long as we don't accept other triggers for ci.yml we are safe. So let's maybe add a ticket and a comment in ci.yml and go ahead for the time being?

Let me know if I there's something I can help you with once you want to address that 👍

docs/RELEASE.md
jku: 1947 is merged, it's all good now I think :crossed_fingers:

joshuagl:
This is really cool work! I can try and provide more detailed review next week, but I really like what I'm seeing and would be happy to see it merge and continue to iterate in tree.

To comment on some of the TODOs:
* I am open to using existing 3rd-party actions, especially the PyPA ones, where they seem reputable and we've taken some steps to develop trust in the code and maintainer(s).
* I prefer the use of [download|upload]-artifact over staging release assets – as @jku already mentioned, this keeps the artefact publishing pipeline out of reach of developer (account owner) tampering.
* Of the quick fix signing options, I think I favour 3 and developer signing locally. Longer-term I'd like to explore in-toto and/or SLSA to bind the code to the repo and the GitHub accounts, rather than developer controlled long-lived secrets.

lukpueh:
Thanks for the reviews, @jku and @joshuagl! I just pushed two commits that fix typos, polish code comments and add instructions for the one-time repo setup in RELEASE.md.

As far as I'm concerned we can merge and start using this. I'll ticketize the remaining TODOs from [above](https://github.com/theupdateframework/python-tuf/pull/1946#issue-1195886368).

jku (approved):
Seems correct to me, only a release will tell for sure :)

joshuagl (approved):
I'm excited to see this in action. Awesome work!

lukpueh:
Many thanks for the re-reviews! Regarding todos...

> - re-enable release signing

https://github.com/theupdateframework/python-tuf/issues/1966

> - replace 3rd-party actions with custom code

https://github.com/theupdateframework/python-tuf/issues/1967

> - make ci/cd setup more robust

https://github.com/theupdateframework/python-tuf/issues/1961

> - replace upload/download artifacts with downloading artifacts from GitHub release assets

Won't fix. See discussion above and corresponding [code comment](https://github.com/theupdateframework/python-tuf/blob/b99d0432a7f0a803c492580085e85f81908e0dc1/.github/workflows/cd.yml#L49-L51).

> - (nice to have) auto-generate release notes for GitHub release page

https://github.com/theupdateframework/python-tuf/issues/1968

> - Configure release job to not add GitHub auto-created release artifacts

Won't fix. It does not seem possible to exclude or even remove these archives from the assets at the moment.

> - CD workflow could have a permissions section

https://github.com/theupdateframework/python-tuf/issues/1969

Issue #1909, opened 2022-03-18
TAP 15 (succinct hashed bin delegations) implementation


MVrachev:
**Description of issue or feature request**:
[TAP 15](https://github.com/theupdateframework/taps/blob/master/tap15.md) was created on June 23-rd 2020 and was last modified in July 6-th 2020.
Since then the TAP has been put to `draft` status meaning it needs a prototype implementation before it's accepted as a future specification change.
Given that this TAP underlines an efficient way of handling hash bin delegations, meaning it would be really useful when `python-tuf` is integrated into `Warehouse`, it's logical that we should not only create a prototype but directly work to integrate it in `python-tuf`.

The result I am expecting from this issue is:
1. Support reading, updating, and saving target metadata files that are using `succinct hashed bin delegations`.
2. An exemplary document inside `examples/` showcasing how `succinct hashed bin delegations` can be utilized in practice.

I have come up with the following steps:

- [x] Add and update the necessary attributes to support reading and writing from/to a target file with `succinct hashed bin delegations`
- [x] Add simple testing for target files using `succinct hashed bin delegations`.
- [x] Make sure the client parses the succinct hash bin delegated metadata
- [x] Test that the client parses the succinct hash bin delegated metadata and visits the expected bin for a specific target
- [x] Test actually downloading a target
- [x] Add a script inside `examples/`
- [x] Testing the `examples/` script

lukpueh:
This is exactly what I had in mind for this issue! Plus, the client also needs to be able to parse and interpret the new metadata, in order to find the correct bin for a target.

MVrachev:
> This is exactly what I had in mind for this issue! Plus, the client also needs to be able to parse and interpret the new metadata, in order to find the correct bin for a target.

Agree, added it to the TODO list.

MVrachev:
After #2031 and #2038 are merged I think we are ready to close this issue.
I don't see any other todo items left.

MVrachev:
Closing as there is no work left on that issue.

Issue #1943, opened 2022-04-06
Relying on mandatory DelegatedRole.name attribute when TAP 15 describes it as optional


MVrachev:
**Description of issue or feature request**:
We are working to implement [TAP 15](https://github.com/theupdateframework/taps/blob/master/tap15.md) (see #1909).
There it's said that:
>  If the succinct_hash_delegations field is present in a delegation, the name field will not be used as a rolename, and so is not required.

meaning that the `DelegatedRole.name` attribute is not required.
That seems to be a problem for us as until now we are relying on its existence to uniquely identify roles:
(see `Delegations.from_dict()` implementation)
https://github.com/theupdateframework/python-tuf/blob/10f9feaead407cf65b7b29ecc4d28b0e2190cb45/tuf/api/metadata.py#L1472

One solution to this problem could be providing an additional `ID` number that will serve as a unique identifier for roles.
This approach is widely used by databases.

But then we have the next question what to do about `DelegatedRole.to_dict()` where we currently provide `name`:
https://github.com/theupdateframework/python-tuf/blob/10f9feaead407cf65b7b29ecc4d28b0e2190cb45/tuf/api/metadata.py#L1359

In summary what we should do is:

- [ ] decide what unique identifier to use for delegated roles inside `Delegations`
- [ ] decide what to do with `DelegatedRole.to_dict`

lukpueh:
Yes, we definitely still need distinct names for each "bin-n" role, otherwise the client can't download the targets metadata needed for a target in a bin. The challenge with TAP 15 is that the delegating targets metadata does not list all its delegatees anymore, but rather a number (`delegation_hash_prefix_len`) that allows to compute how many delegatees (bins) there are.

Currently, the idea is to compute the name for each bin using a configured prefix `bin_name_prefix` together with the bin index. See https://github.com/theupdateframework/taps/issues/132 for some remaining discussions about the name.

MVrachev:
> Currently, the idea is to compute the name for each bin using a configured prefix `bin_name_prefix` together with the bin index. See [theupdateframework/taps#132](https://github.com/theupdateframework/taps/issues/132) for some remaining discussions about the name.

The problem is that it's possible to have `DelegatedRole`s objects without a name attribute set and the calculation of the bin index requires a target path...

lukpueh:
I think the problem is that `DelegatedRole` so far was used for exactly one delegated role, but with TAP 15 there would only be one `DelegatedRole` object for all bin-n roles.

[...]

PR #1948, opened 2022-04-08
Add support for TAP 15

MVrachev:
Related to issue: #1909

[TAP 15](https://github.com/theupdateframework/taps/blob/master/tap15.md) was created on June 23-rd 2020 and was last modified in July 6-th 2020.
Since then the TAP has been put to `draft` status meaning it needs a prototype implementation before it's accepted as a future specification change.
Given that this TAP underlines an efficient way of handling hash bin delegations, meaning it would be really useful when `python-tuf` is integrated into `Warehouse`, it's logical that we should not only create a prototype but directly work to integrate it in `python-tuf`.

The outcome of the TAP 15 implementation is
1. Support reading, updating, and saving target metadata files that are using `succinct hashed bin delegations`.
2. Support in ngclient for calculating the delegated role name on a given target path.
3. An exemplary document inside `examples/` showcasing how `succinct hashed bin delegations` can be utilized in practice.

This pr covers points 1 and 2.
**Point 3 will be done in another pr.** 

requirements**:

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -357,17 +357,22 @@ class RepositorySimulator(FetcherInterface):
         # Create delegation
         if delegator.delegations is None:
             delegator.delegations = Delegations({}, {})
+
+        role_name: str = role.name if role.name is not None else ""
+        if role.succinct_hash_info:
+            role_name = role.succinct_hash_info.get_bin_name()
+
         # put delegation last by default
-        delegator.delegations.roles[role.name] = role
+        delegator.delegations.roles[role_name] = role
 
         # By default add one new key for the role
         key, signer = self.create_key()
-        delegator.add_key(role.name, key)
-        self.add_signer(role.name, signer)
+        delegator.add_key(role_name, key)
+        self.add_signer(role_name, signer)
 
         # Add metadata for the role
-        if role.name not in self.md_delegates:
-            self.md_delegates[role.name] = Metadata(targets, {})
+        if role_name not in self.md_delegates:
+            self.md_delegates[role_name] = Metadata(targets, {})
 
     def write(self) -> None:
         """Dump current repository metadata to self.dump_dir
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -346,7 +346,10 @@ class RepositorySimulator(FetcherInterface):
         self.target_files[path] = RepositoryTarget(data, target)
 
     def add_delegation(
-        self, delegator_name: str, role: DelegatedRole, targets: Targets
+        self,
+        delegator_name: str,
+        role: DelegatedRole,
+        targets: Optional[Targets],
     ) -> None:
         """Add delegated target role to the repository."""
         if delegator_name == Targets.type:
@@ -359,20 +362,31 @@ class RepositorySimulator(FetcherInterface):
             delegator.delegations = Delegations({}, {})
 
         role_name: str = role.name if role.name is not None else ""
+        key, signer = self.create_key()
+
         if role.succinct_hash_info:
+            # Add target metadata for all bins.
+            for delegated_name in role.succinct_hash_info.get_all_bin_names():
+                self.md_delegates[delegated_name] = Metadata(
+                    Targets(expires=self.safe_expiry)
+                )
+
+                self.add_signer(delegated_name, signer)
+
             role_name = role.succinct_hash_info.get_bin_name()
+
+        else:
+            # By default add one new key for the role
+            self.add_signer(role_name, signer)
+
+            # Add metadata for the role
+            if targets is not None:
+                if role_name not in self.md_delegates:
+                    self.md_delegates[role_name] = Metadata(targets)
 
         # put delegation last by default
         delegator.delegations.roles[role_name] = role
-
-        # By default add one new key for the role
-        key, signer = self.create_key()
         delegator.add_key(role_name, key)
-        self.add_signer(role_name, signer)
-
-        # Add metadata for the role
-        if role_name not in self.md_delegates:
-            self.md_delegates[role_name] = Metadata(targets, {})
 
     def write(self) -> None:
         """Dump current repository metadata to self.dump_dir
--- a/tests/test_metadata_eq_.py
+++ b/tests/test_metadata_eq_.py
@@ -213,6 +213,7 @@ class TestMetadataComparisions(unittest.TestCase):
             ("terminating", None),
             ("paths", [""]),
             ("path_hash_prefixes", [""]),
+            ("succinct_hash_info", ""),
         ]:
             setattr(delegated_role_2, attr, value)
             msg = f"Failed case: {attr}"
--- a/tests/test_metadata_eq_.py
+++ b/tests/test_metadata_eq_.py
@@ -25,6 +25,7 @@ from tuf.api.metadata import (
     Role,
     Root,
     Snapshot,
+    SuccinctHashDelegations,
     TargetFile,
     Targets,
     Timestamp,
@@ -193,6 +194,19 @@ class TestMetadataComparisions(unittest.TestCase):
         # Common attributes between Signed and Snapshot doesn't need testing.
         setattr(signed_copy, "meta", None)
         self.assertNotEqual(md.signed, signed_copy)
+
+    def test_hash_delegations_eq(self) -> None:
+        hash_deleg = SuccinctHashDelegations(8, "delegated")
+        hash_deleg_2: SuccinctHashDelegations = self.copy_and_simple_assert(
+            hash_deleg
+        )
+
+        for attr, value in [("prefix_bit_len", 0), ("bin_name_prefix", "")]:
+            setattr(hash_deleg_2, attr, value)
+            msg = f"Failed case: {attr}"
+            self.assertNotEqual(hash_deleg, hash_deleg_2, msg)
+            # Restore the old value of the attribute.
+            setattr(hash_deleg_2, attr, getattr(hash_deleg, attr))
 
     def test_delegated_role_eq_(self) -> None:
         delegated_role_dict = {
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -32,7 +32,7 @@ class TestDelegation:
     keyids: List[str] = field(default_factory=list)
     threshold: int = 1
     terminating: bool = False
-    paths: List[str] = field(default_factory=lambda: ["*"])
+    paths: Optional[List[str]] = field(default_factory=lambda: ["*"])
     path_hash_prefixes: Optional[List[str]] = None
 
 
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -32,7 +32,7 @@ class TestDelegation:
     keyids: List[str] = field(default_factory=list)
     threshold: int = 1
     terminating: bool = False
-    paths: Optional[List[str]] = field(default_factory=lambda: ["*"])
+    paths: List[str] = field(default_factory=lambda: ["*"])
     path_hash_prefixes: Optional[List[str]] = None
 
 
@@ -344,8 +344,8 @@ class TestDelegationsGraphs(TestDelegations):
         for rolename in roles_to_filenames:
             delegations.append(TestDelegation("targets", rolename))
 
-        delegated_rolenames = DelegationsTestCase(delegations)
-        self._init_repo(delegated_rolenames)
+        delegated_names = DelegationsTestCase(delegations)
+        self._init_repo(delegated_names)
         updater = self._init_updater()
         updater.refresh()
 
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -20,6 +20,7 @@ from tuf.api.metadata import (
     SPECIFICATION_VERSION,
     TOP_LEVEL_ROLE_NAMES,
     DelegatedRole,
+    SuccinctHashDelegations,
     Targets,
 )
 from tuf.ngclient import Updater
@@ -422,6 +423,99 @@ class TestDelegationsGraphs(TestDelegations):
             # order and the corresponding metadata files were persisted
             self.assertListEqual(self.sim.fetch_tracker.metadata, exp_calls)
             self._assert_files_exist(exp_files)
+        finally:
+            self.teardown_subtest()
+
+    @dataclass
+    class SuccinctHashTestCase:
+        prefix_bit_len: int
+        target_path: str
+        expected_visited_bin: str
+
+    succinct_bins_graph: utils.DataSet = {
+        "case 1": SuccinctHashTestCase(
+            prefix_bit_len=1,
+            target_path="othermissingpath",
+            expected_visited_bin="delegated-0",
+        ),
+        "case 2": SuccinctHashTestCase(
+            prefix_bit_len=1,
+            target_path="missingpath",
+            expected_visited_bin="delegated-1",
+        ),
+        "case 3": SuccinctHashTestCase(
+            prefix_bit_len=2,
+            target_path="foo",
+            expected_visited_bin="delegated-0",
+        ),
+        "case 4": SuccinctHashTestCase(
+            prefix_bit_len=2,
+            target_path="doo",
+            expected_visited_bin="delegated-1",
+        ),
+        "case 5": SuccinctHashTestCase(
+            prefix_bit_len=2,
+            target_path="missingpath",
+            expected_visited_bin="delegated-2",
+        ),
+        "case 6": SuccinctHashTestCase(
+            prefix_bit_len=2,
+            target_path="bar",
+            expected_visited_bin="delegated-3",
+        ),
+        "many bins case": SuccinctHashTestCase(
+            prefix_bit_len=8,
+            target_path="bar",
+            expected_visited_bin="delegated-fc",
+        ),
+    }
+
+    @utils.run_sub_tests_with_dataset(succinct_bins_graph)
+    def test_succinct_hash_bins_graph_traversal(
+        self, test_data: SuccinctHashTestCase
+    ) -> None:
+        # Test traversing the delegation tree when succinct hash delegation is
+        # used. For a successful traversal all top level metadata files plus the
+        # expected bin should exist locally and only the bin must be downloaded.
+
+        try:
+            exp_files = [*TOP_LEVEL_ROLE_NAMES, test_data.expected_visited_bin]
+            exp_calls = [(test_data.expected_visited_bin, 1)]
+
+            self.sim = RepositorySimulator()
+            delegated_role = DelegatedRole(
+                None,
+                [],
+                1,
+                False,
+                succinct_hash_info=SuccinctHashDelegations(
+                    test_data.prefix_bit_len, "delegated"
+                ),
+            )
+            # Add "delegated_role" as delegation and create target metadata
+            # instances for all bins.
+            self.sim.add_delegation("targets", delegated_role, None)
+            self.sim.update_snapshot()
+
+            self.setup_subtest()
+
+            updater = self._init_updater()
+            # Call explicitly refresh to simplify the expected_calls list
+            updater.refresh()
+            self.sim.fetch_tracker.metadata.clear()
+            # Check that metadata dir contains only top-level roles
+            self._assert_files_exist(TOP_LEVEL_ROLE_NAMES)
+
+            # Looking for a non-existing targetpath forces updater
+            # to visit a correspondning delegated role
+            targetfile = updater.get_targetinfo(test_data.target_path)
+            self.assertIsNone(targetfile)
+
+            # Check that the delegated roles were visited in the expected
+            # order and the corresponding metadata files were persisted
+            self.assertListEqual(self.sim.fetch_tracker.metadata, exp_calls)
+            self._assert_files_exist(exp_files)
+
         finally:
             self.teardown_subtest()
 
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1351,6 +1351,18 @@ class SuccinctHashDelegations:
             **self.unrecognized_fields,
         }
 
+    def get_bin_name(self, bin_number: int = 0) -> str:
+        """Return the name of bin with number ``bin_number``.
+        The ``bin_number`` default is 0 as that is used to uniquely identify
+        delegated roles containing succinct hash delegations from the rest.
+
+        Args:
+            bin_number: Number of the bin that we want the name for.
+        """
+        # Add zero padding if necessary and cast to hex the suffix.
+        suffix = f"{bin_number:0{self.suffix_len}x}"
+        return f"{self.bin_name_prefix}-{suffix}"
+
     def _find_bin_for_bits(self, hash_bits_representation: str) -> str:
         """Helper function for find_delegation calculating the actual rolename.
 
@@ -1653,9 +1665,13 @@ class Delegations:
         roles_res: Dict[str, DelegatedRole] = {}
         for role_dict in roles:
             new_role = DelegatedRole.from_dict(role_dict)
-            if new_role.name in roles_res:
+            role_name: str = new_role.name if new_role.name is not None else ""
+            if new_role.succinct_hash_info:
+                role_name = new_role.succinct_hash_info.get_bin_name()
+
+            if role_name in roles_res:
                 raise ValueError(f"Duplicate role {new_role.name}")
-            roles_res[new_role.name] = new_role
+            roles_res[role_name] = new_role
         # All fields left in the delegations_dict are unrecognized.
         return cls(keys_res, roles_res, delegations_dict)
 
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1273,6 +1273,90 @@ class Snapshot(Signed):
         return snapshot_dict
 
 
+class SuccinctHashDelegations:
+    """A container with information about a succinct hash delegation.
+
+    Succinct hash delegation describes delegations for all possible target paths
+    to a variable number of bins, the names of which are generated based on the
+    two class attributes.
+
+    Args:
+        prefix_bit_len: Number of bits between 1 and 32 used to separate the
+            different bins.
+        bin_name_prefix: Prefix of all bin names.
+        unrecognized_fields: Dictionary of all attributes that are not managed
+            by TUF Metadata API.
+
+    Raises:
+            ValueError, TypeError, AttributeError: Invalid arguments.
+    """
+
+    def __init__(
+        self,
+        prefix_bit_len: int,
+        bin_name_prefix: str,
+        unrecognized_fields: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        if prefix_bit_len <= 0 or prefix_bit_len > 32:
+            raise ValueError("prefix_bit_len must be between 1 and 32")
+        if not isinstance(bin_name_prefix, str):
+            raise ValueError("bin_name_prefix must be a string")
+
+        self.prefix_bit_len = prefix_bit_len
+        self.bin_name_prefix = bin_name_prefix
+
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, SuccinctHashDelegations):
+            return False
+
+        return (
+            self.prefix_bit_len == other.prefix_bit_len
+            and self.bin_name_prefix == other.bin_name_prefix
+            and self.unrecognized_fields == other.unrecognized_fields
+        )
+
+    @classmethod
+    def from_dict(
+        cls, succinct_hash_info: Dict[str, Any]
+    ) -> "SuccinctHashDelegations":
+        """Creates ``SuccinctHashDelegations`` object from its json/dict
+        representation.
+
+        Raises:
+            ValueError, KeyError, TypeError, AttributeError: Invalid arguments.
+        """
+        prefix_bit_len = succinct_hash_info.pop("delegation_hash_prefix_len")
+        bin_name_prefix = succinct_hash_info.pop("bin_name_prefix")
+        # All fields left in the succinct_hash_info are unrecognized.
+        return cls(prefix_bit_len, bin_name_prefix, succinct_hash_info)
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Returns the dict representation of self."""
+        return {
+            "delegation_hash_prefix_len": self.prefix_bit_len,
+            "bin_name_prefix": self.bin_name_prefix,
+            **self.unrecognized_fields,
+        }
+
+    def _find_bin_for_bits(self, hash_bits_representation: str) -> str:
+        """Helper function for find_delegation calculating the actual rolename.
+
+        Args:
+            hash_bits_representation: binary bit representation of the target
+                hash.
+        """
+
+        bit_length = self.prefix_bit_len
+        # Get the first bit_length of bits and then cast them to decimal.
+        bin_number = int(hash_bits_representation[:bit_length], 2)
+        return f"{self.bin_name_prefix}-{bin_number}"
+
+
 class DelegatedRole(Role):
     """A container with information about a delegated role.
 
@@ -1318,7 +1402,7 @@ class DelegatedRole(Role):
         terminating: bool,
         paths: Optional[List[str]] = None,
         path_hash_prefixes: Optional[List[str]] = None,
-        succinct_hash_info: Optional[Dict[str, Any]] = None,
+        succinct_hash_info: Optional[SuccinctHashDelegations] = None,
         unrecognized_fields: Optional[Dict[str, Any]] = None,
     ):
         super().__init__(keyids, threshold, unrecognized_fields)
@@ -1341,15 +1425,6 @@ class DelegatedRole(Role):
             not isinstance(p, str) for p in path_hash_prefixes
         ):
             raise ValueError("Path_hash_prefixes must be strings")
-        if succinct_hash_info is not None:
-            if not isinstance(succinct_hash_info, Dict):
-                raise ValueError("succinct_hash_info must be a dict")
-            if "delegation_hash_prefix_len" not in succinct_hash_info:
-                raise ValueError(
-                    "succinct_hash_info requires delegation_hash_prefix_len"
-                )
-            if "bin_name_prefix" not in succinct_hash_info:
-                raise ValueError("succinct_hash_info requires bin_name_prefix")
 
         self.paths = paths
         self.path_hash_prefixes = path_hash_prefixes
@@ -1373,7 +1448,7 @@ class DelegatedRole(Role):
         """Creates ``DelegatedRole`` object from its json/dict representation.
 
         Raises:
-            ValueError, KeyError: Invalid arguments.
+            ValueError, KeyError, TypeError: Invalid arguments.
         """
         name = role_dict.pop("name", None)
         keyids = role_dict.pop("keyids")
@@ -1381,7 +1456,12 @@ class DelegatedRole(Role):
         terminating = role_dict.pop("terminating")
         paths = role_dict.pop("paths", None)
         path_hash_prefixes = role_dict.pop("path_hash_prefixes", None)
-        succinct_hash_info = role_dict.pop("succinct_hash_delegations", None)
+        succinct_hash_dict = role_dict.pop("succinct_hash_delegations", None)
+        succinct_hash_info = None
+        if succinct_hash_dict is not None:
+            succinct_hash_info = SuccinctHashDelegations.from_dict(
+                succinct_hash_dict
+            )
         # All fields left in the role_dict are unrecognized.
         return cls(
             name,
@@ -1397,19 +1477,19 @@ class DelegatedRole(Role):
     def to_dict(self) -> Dict[str, Any]:
         """Returns the dict representation of self."""
         base_role_dict = super().to_dict()
-        res_dict: Dict[str, Any] = {
+        res: Dict[str, Any] = {
             "terminating": self.terminating,
             **base_role_dict,
         }
         if self.name is not None:
-            res_dict["name"] = self.name
+            res["name"] = self.name
         if self.paths is not None:
-            res_dict["paths"] = self.paths
+            res["paths"] = self.paths
         elif self.path_hash_prefixes is not None:
-            res_dict["path_hash_prefixes"] = self.path_hash_prefixes
+            res["path_hash_prefixes"] = self.path_hash_prefixes
         elif self.succinct_hash_info is not None:
-            res_dict["succinct_hash_delegations"] = self.succinct_hash_info
-        return res_dict
+            res["succinct_hash_delegations"] = self.succinct_hash_info.to_dict()
+        return res
 
     @staticmethod
     def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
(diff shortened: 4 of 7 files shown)


MVrachev:
This pr is a draft because of warnings given by `mypy`.
Those warnings are connected with the issue discussed in https://github.com/theupdateframework/python-tuf/issues/1943.

I am not sure what is the problem for python3.7 https://github.com/theupdateframework/python-tuf/runs/5888251078?check_suite_focus=true

tuf/api/metadata.py
```diff
@@ -1274,12 +1274,14 @@ class DelegatedRole(Role):
     instance attributes.*
 
     Args:
-        name: Delegated role name.
```
jku: this is a backwards incompatible API change, we might not want that

tuf/api/metadata.py
```diff
@@ -1289,32 +1291,59 @@ class DelegatedRole(Role):
...
     def __init__(
         self,
-        name: str,
         keyids: List[str],
         threshold: int,
         terminating: bool,
+        name: Optional[str] = None,
         paths: Optional[List[str]] = None,
         path_hash_prefixes: Optional[List[str]] = None,
+        succinct_hash_info: Optional[Dict[str, Any]] = None,
```
jku: is there a reason for using a different name than the TAP does?

tuf/api/metadata.py
```diff
@@ -1384,6 +1419,46 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
...
+
+        Args:
+            hash_bits_representation: binary bit representation of the target
+                hash.
+        """
+        if self.succinct_hash_info is None:
+            return self.name
+
+        # number of bins
+        bit_length = self.succinct_hash_info["delegation_hash_prefix_len"]
```
jku: Dictionaries that pretend to be data structures is a code smell to me: it prevents static typing checks and makes typos deadly. There's multiple other ways to handle this and I think we should consider them:
* maybe hash_prefix_len and bin_name_prefix are just instance variables in DelegatedRole? 
* Maybe succinct_hash_delegations is a data class of it's own -- bit overkill but fits the Metadata API design?
* Maybe DelegatedRole is actually three separate classes for paths, prefixes and succinct prefixes cases? This also sounds like OO-overkill but I'll mention it anyway: all the classes would just implement `get_rolename()` and the de/serialization separately

tuf/ngclient/updater.py
```diff
@@ -438,11 +438,10 @@ def _preorder_depth_first_walk(
                 # delegated roles.
                 for child_role in targets.delegations.roles.values():
                     if child_role.is_delegated_path(target_filepath):
-                        logger.debug("Adding child role %s", child_role.name)
+                        child_name = child_role.get_rolename(target_filepath)
```
jku: Is is_delegated_path() useful for anything now? It seems we only call it in combination with get_rolename(). 

So we could just have get_rolename() return `Optional[str]` and just have a single public method in DelegatedRole: no need to call is_delegated_path() anymore?

tuf/api/metadata.py
```diff
@@ -1384,6 +1419,46 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
...
+        # Get the first bit_length of bits and then cast them to decimal.
+        bin_number = int(hash_bits_representation[:bit_length], 2)
+
+        name_prefix = self.succinct_hash_info["bin_name_prefix"]
+        return f"{name_prefix}-{bin_number}"
+
+    def get_rolename(self, target_filepath: str) -> str:
+        """Get the delegated rolename based on the given ``target filepath``.
```
jku: the name is techincally correct but feels a bit off (we already have a `rolename` property as well). maybe `find_delegation()` or something to show that we're looking for a delegation for the specific target_filepath?

I mention this elsewhere but this should probably return `Optional[str]` since it may be be called for a normal delegation and I think returning a name if the delegation does not match feels wrong -- this would also mean is_delegated_path() could just be deprecated as this method returns the answer to that as well.

tuf/api/metadata.py
```diff
@@ -1289,32 +1291,59 @@ class DelegatedRole(Role):
...
-        if paths is None and path_hash_prefixes is None:
-            raise ValueError("One of paths or path_hash_prefixes must be set")
+        if (
+            paths is None
+            and path_hash_prefixes is None
+            and succinct_hash_info is None
+        ):
+            raise ValueError(
+                "paths, path_hash_prefixes or succinct_hash_info required"
+            )
```
jku: if you want to remove some code, something like this could work:
```python
exclusive_vars = [paths, path_hash_prefixes, succinct_hash_info]
if sum(1 for var in exclusive_vars if var is not None) != 1:
    raise ValueError("One of (paths, path_hash_prefixes, succinct_hash_info) must be set")
```

jku (changes requested):
Nice work. 
* I left a complaint about using dict to "hide" API, that one I fee strongly about, let's figure out a solution for it. 
* changing API in incompatible ways (like arg order in `DelegatedRole.__init__()`) may be inevitable, but let's at least try to avoid it
* other comments are not strictly change requests.

Is the current implementation using None as the key in `Delegations.roles` dictionary? I'm quite surprised mypy does not complain about it in Delegations.from_dict() ...

tuf/api/metadata.py
MVrachev: Yes, it boils down to that it's just shorter.
`succinct_hash_delegations` is 25 characters and because of our characters per line restriction, I had to often split lines.
`succinct_hash_info` is 18 characters which save a good amount of characters while still hinting at the purpose of this argument.

tuf/api/metadata.py
MVrachev: Yea, you are correct. We should start thinking as a real library and be careful with such changes.
Because of Python limitations, I had to move the `name` argument if it has a default value after all attributes without a default value.
I think maybe it will be better to require a value for `name` even if it's `None` and not move it.
Will try with that.

tuf/api/metadata.py
MVrachev: Loved this simplification.
Great suggestion!

MVrachev:
I have made multiple changes:
1. Made `DelegatedRole.name` as the first attribute in the initialization list (the same way as it's now in the develop branch) and as thus removing the incompatible change
2. Renamed `DelegatedRole.get_rolename` to `Delegated.find_delegations` and used  `find_delegations` replacing the need for `DelegatedRole.is_delegated_path`.
3. Simplified checks with `sum` suggestions by @jku 
4. Experimented with creating a new container helper class `HashDelegations` containing the information about succinct hash delegations. This allowed me to make sure the requirements from TAP 15 are met inside the class without making `DelegatedRole` even more complicated and to add support for unrecognized fields inside the `succinct_hash_delegations` dict.

@jku it's ready for another review.

tuf/api/metadata.py
```diff
@@ -1313,8 +1380,16 @@ def __init__(
         ):
             raise ValueError("Path_hash_prefixes must be strings")
 
+        succinct_info = None
+        if succinct_hash_info is not None:
+            if not isinstance(succinct_hash_info, Dict):
+                raise ValueError("succinct_hash_info must be a dict")
+
+            succinct_info = HashDelegations.from_dict(succinct_hash_info)
```
jku: you're calling `from_dict()` from a constructor, this is not how any existing code does it. from_dict() is a helper for the deserialization path but constructors can be called elsewhere too: constructors don't take unprocessed json dictionaries as argument, they take deserialized values/objects (like your new HashDelegations) as arguments.

tuf/api/metadata.py
```diff
@@ -1384,6 +1465,58 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
...
+            # Get the bit representation of the hash.
+            hash_bits = "".join(
+                f"{one_byte:08b}" for one_byte in hasher.digest()
+            )
+
+            return self._calculate_delegation(hash_bits)
+
+        return None
+
     def is_delegated_path(self, target_filepath: str) -> bool:
```
jku, line 1598: So with this design:
* is_delegated_path() could be deprecated: find_delegation() answers that question and supports succinct delegations
* however is_delegated_path() is still available and currently returns wrong answer if used with a succinct delegation

I think we should document is_delegated_path as deprecated in the docstring and maybe add the couple of lines needed to fix the second issue as well

tuf/api/metadata.py
```diff
@@ -1384,6 +1465,58 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
...
+    def find_delegation(self, target_filepath: str) -> Optional[str]:
+        """Find the delegated rolename based on the given ``target filepath``.
+
+        If the given ``target_filepath`` is not in any of the paths that
+        ``DelegatedRole`` is trusted to provide it returns None.
+
+        If succinct_hash_info is used the result will be in the format
+        "{bin_name_prefix}-{bin_number}" where "bin_name_prefix" is information
+        provided by succinct_hash_info and "bin_number" is the corresponding
+        bin number calculated based on the ``target_filepath``.
```
jku: Nit: Looks complex and too detailed for this docstring, maybe not needed at all?

tuf/api/metadata.py
```diff
@@ -1384,6 +1465,58 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
 
         return True
 
+    def _calculate_delegation(self, hash_bits_representation: str) -> str:
```
jku: this method now seems like it should be part of the new class (HashDelegations.find_delegation() maybe)?

I also think the argument to the function should be `targetpath` : this way the arguments and output make sense to a reader: you feed in the targetpath and get the real delegated role name as output -- The fact that this specific implementation uses a string of ones and zeros as an intermediate value is not important and you only need to move a few lines of code to get this better abstraction.

Then `DelegatedRole.find_delegation()` ends up being quite clean, something like
```
if self.succinct_hash_info:
    return self.succinct_hash_info.find_delegation(target_filepath)
elif self.is_delegated_path(target_filepath):
    return self.name
return None
```

Opinions?

tuf/api/metadata.py
```diff
@@ -1258,6 +1258,64 @@ def to_dict(self) -> Dict[str, Any]:
         return snapshot_dict
 
 
+class HashDelegations:
+    """A container with information about a succinct hash delegation.
```
jku: I think I like this: I imagined it would be OO-overkill but maybe not: 

We could move some functionality here 
* like the delegation math you need, see other comment: I think it'll look just fine
* but also the math required to create new succinct delegations -- where @kairoaraujo can say something like _"I want 2**14 bins, give me the delegation but also give me a generator for all the bin names so I can create the metadata for those"._

We'll want to be careful with the name though, there are so many delegation related names. We now have three completely different names for the same concept, this does not feel correct: 
* spec calls it `succinct_hash_delegations`
* DelegatedRole calls it `succinct_hash_info`
* here it's called `HashDelegations`

tuf/api/metadata.py
```diff
@@ -1274,12 +1332,15 @@ class DelegatedRole(Role):
     instance attributes.*
 
     Args:
-        name: Delegated role name.
+        name: Delegated role name. If succinct_hash_info is passed, then a None
+            value should be given to ``name``.
```
jku: Note that there is quite a bit of text on the mutual exclusivity of arguments above this: probably makes sens to mention this name issue there and fix that text in general.

tuf/api/metadata.py
```diff
@@ -1289,22 +1350,28 @@ class DelegatedRole(Role):
...
         self,
-        name: str,
+        name: Optional[str],
         keyids: List[str],
         threshold: int,
         terminating: bool,
         paths: Optional[List[str]] = None,
         path_hash_prefixes: Optional[List[str]] = None,
+        succinct_hash_info: Optional[Dict[str, Any]] = None,
         unrecognized_fields: Optional[Mapping[str, Any]] = None,
```
jku: strictly speaking this is still breaking API.  
`DelegatedRole("a", [], 1, False, [], None, {})`
used to be valid  code. Now it will fail I assume. 

New arguments should generally speaking be added in the end of the list... I'm not saying we can't break API here: it's not a major breakage by any means but maybe I'd still vote for adding the new arg in the end.

jku (commented):
Thanks! I think the direction in this version can work. It's a new class just to hold two variables: this feels like overkill at first but there's some specific problems (the bin math plus the handling unrecognized fields in the new dict) that we can contain within that class so I think it works out.

tuf/api/metadata.py
MVrachev: I agree about the naming.
Honestly I wasn't sure how to call it at all.
Maybe `SuccinctHashInfo`? But I don't like the suffix `info` as it doesn't bring any new context...
On another hand, `SuccinctHashDelegations` becomes too much...

tuf/api/metadata.py
MVrachev: I didn't add it to the end as in my mind `unrecognized_fields` are at the end of each class, but maybe that's not so important.

tuf/api/metadata.py
jku: On the other hand, class name is not used often so that could still work if you can keep the variable name as something clearly related but a little shorter, something like:

succinct_delegations = SuccinctHashDelegations(3, "bins")

Admittedly that is a bit longer than your current variable name

tuf/api/metadata.py
jku: To make this a little more complex: we don't have an API for someone who
* uses non-succinct hash delegations
* wants to find out which bin might provide (or shoudl provide) a targetpath

This API clearly needs to be above the DelegatedRole level -- at Delegations maybe, but internally could use some of the same code

This question is not really TAP 15 related... but makes sense to solve (or at least discuss) at same time (and is related to #1952 and warehouse work)

MVrachev: The reason I don't pass `targetpath` as an argument is the testing.
I want to test the actual bin calculation math and if I pass the `targetpath` this means that the hash calculation has to be moved inside `_calculate_delegation` and I won't be able to pass the binary representation in a string as do now. 
That way the test inside `test_api.py` won't be clear for the reader if we do that.
Given that this is an internal API I think it's allowed to require a little more specific argument for easier testing.

jku: hmm, I think I'll leave this to you and @lukpueh as I think you've been talking about this part of the code.

Personally I'm not fond of making production code harder to read based on unit testing needs, or the basic idea of building a string of ones and zeros just to turn it into an int later ... but I can live with those so I'll leave these for you

tuf/api/metadata.py
MVrachev: Maybe you are right. Will remove it.

MVrachev:
@jku I updated the pr once again making the following changes (not in order):
- renaming `HashDelegations` to `SuccinctHashDelegations` - please leave a comment what do you think
- left a deprecation messaged for `is_delegated_path` and made `is_delegated_path` work when `succinct_hash_delegations` is used
- removed a complex comment inside `from_delegations`
- moved the helper function `_calculate_delegation` inside `SuccinctHashDelegations`
- added additional comments about the exclusive arguments in the `DelegatedRole` docstring

MVrachev:
Made an update to one of the commits to remove the `f` string feature that caused problems for all `python3.7` CI runs.

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,99 @@ def to_dict(self) -> Dict[str, Any]:
         return snapshot_dict
 
 
+class SuccinctHashDelegations:
+    """A container with information about a succinct hash delegation.
```
jku, line 1289: I think this still needs an explanation: "succinct hash delegation" is probably word salad for anyone except about 20 people on the planet.

It does not need to explain every detail but the core idea: that it desribes delegations of all possible target paths to variable number of bins, the names of which are generated based on the two fields.

tests/test_metadata_eq_.py
```diff
@@ -182,6 +183,23 @@ def test_snapshot_eq_(self) -> None:
         setattr(signed_copy, "meta", None)
         self.assertNotEqual(md.signed, signed_copy)
 
+    def test_hash_delegations_eq(self) -> None:
+        succinct_hash_info_dict = {
+            "delegation_hash_prefix_len": 8,
+            "bin_name_prefix": "delegated",
+        }
+        hash_deleg = SuccinctHashDelegations.from_dict(succinct_hash_info_dict)
```
jku: not terribly important but I think it's weird to use `from_dict()` outside of serialization testing: from_dict() is a serialization implementation detail, and using the constructor should be less lines of code as well.

This style is used in multiple tests

tuf/api/metadata.py
```diff
@@ -1326,21 +1418,31 @@ def __eq__(self, other: Any) -> bool:
...
         keyids = role_dict.pop("keyids")
         threshold = role_dict.pop("threshold")
         terminating = role_dict.pop("terminating")
         paths = role_dict.pop("paths", None)
         path_hash_prefixes = role_dict.pop("path_hash_prefixes", None)
+        succinct_hash_dict = role_dict.pop("succinct_hash_delegations", None)
+        succinct_hash_info = None
+        if succinct_hash_dict is not None:
+            if not isinstance(succinct_hash_dict, Dict):
+                raise ValueError("succinct_hash_delegations must be a dict")
```
jku: maybe not needed: the constructor will fail if it's not a dict

tuf/api/metadata.py
```diff
@@ -1384,8 +1490,39 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
...
+            )
+
+            # pylint: disable=protected-access
+            return self.succinct_hash_info._calculate_delegation(hash_bits)
+
+        return None
+
     def is_delegated_path(self, target_filepath: str) -> bool:
-        """Determines whether the given ``target_filepath`` is in one of
+        """This method is depricated, use find_delegation instead!
```
```suggestion
        """This method is deprecated, use ``find_delegation()`` instead.
```

fixed the typo, but while I was doing that also removed the exclamation: no need to shout in docs. Same typo in a warning message.

There might be some fancy way to product nice note/warning layout in RTD... but I think this is fine, no need to overdo it

tuf/api/metadata.py
```diff
@@ -1399,6 +1536,9 @@ def is_delegated_path(self, target_filepath: str) -> bool:
                 targets URL.
         """
 
+        msg = "is_delegated_path is depricated, use find_delegation instead!"
+        warnings.warn(msg, DeprecationWarning)
```
jku: it's not quite this simple if you want to keep using this method in find_delegation() for now: now calling find_delegation(), a non-deprecated function, will result in a deprecation warning.

So either we don't have the warning (this is fine IMO: switching from `is_delegated_path()` to `find_delegation()` does not require any preparation) or we prevent the warning in the non-deprecated code path.

jku (commented):
* SuccinctHashDelegations works better as a name I think: it's long but not used v commonly.
* moving `_calculate_delegation()` without modifying it otherwise probably does not make sense:  having to disable pylints is a good hint that something is not right. my opinion is still that it would make sense for SuccinctHashDelegations to expose an actual delegation lookup method that accepts target path as argument but I'll just leave that as opinion
* I think it mostly looks good: our naming is now not perfect as DelegatedRole can now represent multiple roles and as result we need `DelegatedRole.find_delegation()`  which really does not make sense as a name ... but it seems best we can do without breaking API: I'm fine with that solution

tuf/api/metadata.py
MVrachev: I added this as the error I got was an `AttributeError: 'str' object has no attribute 'pop'` ` and I thought that it made sense to give a more concrete message, but python developers will know what that message means, so it probably it doesn't make sense.
Will remove this check.

MVrachev:
I addressed all of the @jku comments.
I found a way to add a good looking warning by using the `sphinx` format: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html#warnings which looks like this: 
image
Additionally, I added the proposed solution by @jku inside #1943 to use the first bin name as the unique identifier inside `DelegatedRole.name` when succinct hash delegation is used.

@lukpueh and @jku it's ready for a review.

tuf/api/metadata.py
```diff
@@ -1384,8 +1510,33 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
...
+        If the given ``target_filepath`` is not in any of the paths that
+        ``DelegatedRole`` is trusted to provide it returns None.
+
+        Args:
+            target_filepath: URL path to a target file, relative to a base
+                targets URL.
+        """
+        # Check if the given target_filepath is in any of the "paths" or
+        # "path_hash_prefixes" that DelegatedRole is trusted to provide.
+        warnings.filterwarnings("ignore", category=DeprecationWarning)
```
jku: It's still not this simple. This filters all deprecation warnings within the application from this point forwards, this is incorrect for two reasons:
* we want to filter only warnings from this module
* we want to filter for only a single call, not all calls from now on

You would need to manually remove the filter, or better yet use the catchwarnings() context manager. 

This complexity is why I'm not necessarily convinced the deprecation warnings are worth it: they are surprisingly difficult to get right.

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """Returns the dict representation of self."""
+        return {
+            "delegation_hash_prefix_len": self.hash_prefix_len,
+            "bin_name_prefix": self.bin_name_prefix,
+            **self.unrecognized_fields,
+        }
+
+    def get_first_bin_name(self) -> str:
+        """Return the name of the first bin"""
+        return f"{self.bin_name_prefix}-0"
```
jku: This does not seem to need a method *if* it's this simple...

However, this result (and _calculate_delegation() return value) seemto not match TAP example or the previous implementation IIRC: I'm leaving a longer comment in _calculate_delegation

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+            hash_bits_representation: binary bit representation of the target
+                hash.
+        """
+
+        bit_length = self.hash_prefix_len
+        # Get the first bit_length of bits and then cast them to decimal.
+        bin_number = int(hash_bits_representation[:bit_length], 2)
+
+        name_prefix = self.bin_name_prefix
+        return f"{name_prefix}-{bin_number}"
```
jku: This does not seem to be in line with the TAP examples and if I recall correctly the previous implementation: 

TAP definition:
> The name will be structured as NAME_PREFIX-COUNT where COUNT is a hexadecimal value between 0 and 2^BIT_LENGTH-1 (inclusive) that represents the bin number

TAP example:
>      "name": alice.hbd-00,

 
Your rolename seems to be decimal and not padded with zeros. Is this correct?

jku (commented):
I have not looked at all of it but the naming scheme seems important enough to raise the issue now (see code comments)

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+class SuccinctHashDelegations:
+    """A container with information about a succinct hash delegation.
+
+    Succinct hash delegation describes delegations for all possible target paths
+    to a variable number of bins, the names of which are generated based on the
+    two class attributes.
+
+    Args:
+        hash_prefix_len: Number of bits between 1 and 32 used to separate the
+            different bins.
```
lukpueh, line 1297: I think this needs a better explanation, either here or in the function description above.

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+    def __init__(
+        self,
+        hash_prefix_len: int,
+        bin_name_prefix: str,
+        unrecognized_fields: Optional[Mapping[str, Any]] = None,
+    ) -> None:
+        if hash_prefix_len <= 0 or hash_prefix_len > 32:
+            raise ValueError("hash_prefix_len must be between 1 and 32")
+        if not isinstance(bin_name_prefix, str):
+            raise ValueError("bin_name_prefix must be a string")
```
lukpueh, line 1315: May I ask why you don't check if `hash_prefix_len` is an int, but do check if `bin_name_prefix` is a string?

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        self.hash_prefix_len = hash_prefix_len
+        self.bin_name_prefix = bin_name_prefix
+        self.unrecognized_fields: Mapping[str, Any] = unrecognized_fields or {}
```
lukpueh: See https://github.com/theupdateframework/python-tuf/pull/1963

tuf/api/metadata.py
lukpueh: I agree with @jku that this might not need a dedicated method. I think a `get_bin_name(self, bin_number)` method would be more useful, especially given the subtleties (zero-padded hex) about the name that Jussi mentioned.

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """Calculates the name of the bin responsible for ``target_filepath``.
+
+        Args:
+            target_filepath: URL path to a target file, relative to a base
+                targets URL.
+        """
+        hasher = sslib_hash.digest(algorithm="sha256")
+        hasher.update(target_filepath.encode("utf-8"))
+        # Get the bit representation of the hash.
+        hash_bits = "".join(f"{one_byte:08b}" for one_byte in hasher.digest())
```
lukpueh: Minor optimization nit: You only need the first `math.ceil(self.hash_prefix_len / 8)` bytes.

tuf/api/metadata.py
lukpueh: Minor style nit: Is it really necessary to alias these names here? Wouldn't it be okay to:
```python
# Cast relevant bit prefix to decimal to get the bin number
bin_number = int(hash_bits_representation[:self.hash_prefix_len], 2)
return f"{self.bin_name_prefix}-{bin_number}" # FIXME: Format as zero-padded hex
```

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        # Split hash calculation and from the rest for easier testing.
```
```suggestion
        # Split hash calculation from the rest for easier testing.
```

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """
+
+        bit_length = self.hash_prefix_len
+        # Get the first bit_length of bits and then cast them to decimal.
+        bin_number = int(hash_bits_representation[:bit_length], 2)
+
+        name_prefix = self.bin_name_prefix
+        return f"{name_prefix}-{bin_number}"
+
+    def find_delegation_bin(self, target_filepath: str) -> str:
```
lukpueh: Minor naming nit: Maybe `find_bin` is enough?

tuf/api/metadata.py
```diff
@@ -1258,17 +1259,122 @@ def to_dict(self) -> Dict[str, Any]:
...
+            "delegation_hash_prefix_len": self.hash_prefix_len,
+            "bin_name_prefix": self.bin_name_prefix,
+            **self.unrecognized_fields,
+        }
+
+    def get_first_bin_name(self) -> str:
+        """Return the name of the first bin"""
+        return f"{self.bin_name_prefix}-0"
+
+    def _calculate_delegation(self, hash_bits_representation: str) -> str:
```
lukpueh: Minor naming nit: `_calculate_delegation` sounds a little vague. Maybe `_find_bin_for_bits` makes it clearer that this is part of `find_bin` below?  🤷

lukpueh (commented):
Awesome work! I love how all of this can be implemented in the metadata classes and does not need any changes in `ngclient`.

I mostly reviewed the `SuccinctHashDelegations` class in this round and left a couple of minor comments, let me know what you think.

Regarding the changes in `Delegations` and `DelegatedRole`, I lean towards revisiting the TAP 15 metadata definition. See my suggestion for an alternative format that might make the implementation a little less "hackish" in: https://github.com/theupdateframework/python-tuf/issues/1943#issuecomment-1106390045

I haven't looked closely at the tests yet, but will follow up shortly. Do you plan to add a client test that validates a target delegated to a hash bin?

tuf/api/metadata.py
MVrachev: If `hash_prefix_len` is not a number a comparison like the one done on line 1283 will throw a `TypeError`, for example if `hash_prefix_len` is a string this comparison will throw `TypeError: '<' not supported between instances of 'str' and 'int'`
I am not convinced there is a point in checking that `hash_prefix_len` is an `int`, not a `float`.

tuf/api/metadata.py
MVrachev: Okay, for now, I will keep it simple and leave a docstring warning.
If we later decide to add a deprecation warning it could be worth it to explore, but probably for a more critical API than an API from a helper class.

tuf/api/metadata.py
MVrachev: I am not sure I understand your point here.
If `hash_prefix_len = 17` and I do that math it means `hash_bits` will be `3`.
Meaning I won't be comparing the bits with all of the bits I have defined in `succinct_hash_delegations["delegation_hash_prefix_len"]`.

Maybe I can look only into the `hash_prefix_len` amount of bytes without dividing by 8?

tuf/api/metadata.py
MVrachev: Better explanation about what is a hash deletion or specifically about succinct hash bin delegation?
Can you give some reference points to what you mean?

tuf/api/metadata.py
lukpueh: > If hash_prefix_len = 17 and I do that math it means hash_bits will be 3.

No, it means that you only need the first 3 bytes in order to look at the first 17 bits (all bits from first and second byte + the left-most bit from the third byte).

> Maybe I can look only into the hash_prefix_len amount of bytes without dividing by 8?

You don't need 17 bytes to get the left-most 17 bits.

This should give you the desired prefix of the hash:

```python
needed_bits = self.hash_prefix_len # aliased for emphasis
needed_bytes =  math.ceil(needed_bits / 8)

prefix = "".join(f"{one:08b}" for one in hasher.digest()[:needed_bytes])[:needed_bits] # oneliner might be a bit ugly
```

tuf/api/metadata.py
MVrachev: Agreed that `get_bin_name(self, bin_number)` sounds a lot better.
Will change this method.

MVrachev:
I have updated my pr and addressed all comments made by @jku and @lukpueh.
I am preparing another commit following @lukpueh suggestions in https://github.com/theupdateframework/python-tuf/issues/1943#issuecomment-1106390045 and I am will commit it in another branch.

tests/repository_simulator.py
```diff
@@ -346,9 +346,16 @@ def add_target(self, role: str, data: bytes, path: str) -> None:
...
+        self,
+        delegator_name: str,
+        role: DelegatedRole,
+        targets: Optional[Targets],
     ) -> None:
-        """Add delegated target role to the repository."""
+        """Add delegated target role to the repository.
+
+        If no "targets" are provided this indicates "role" uses a succinct hash
+        delegation and target metadata instances will be created for all bins.
```
jku: this is untrue according to code: whether targets are provided is unrelated to succinct-or-not question

tests/test_updater_delegation_graphs.py
```diff
@@ -425,6 +426,94 @@ def test_hash_bins_graph_traversal(
...
+        "case 5": SuccinctHashTestCase(
+            hash_prefix_len=2,
+            target_path="missingpath",
+            expected_visited_bin="delegated-2",
+        ),
+        "case 6": SuccinctHashTestCase(
+            hash_prefix_len=2,
+            target_path="bar",
+            expected_visited_bin="delegated-3",
+        ),
```
jku: nice test! Have you tried if it's slow with "real" numbers like if your length is 8 or 16?

i am asking because these artificially small test cases mean reader can't easily see if the number is hexadecimal or decimal (and the padding does not get tested)...

tuf/api/metadata.py
```diff
@@ -1258,17 +1267,173 @@ def to_dict(self) -> Dict[str, Any]:
...
+            raise ValueError("bin_name_prefix must be a string")
+
+        self.hash_prefix_len = hash_prefix_len
+        self.bin_name_prefix = bin_name_prefix
+
+        # Calculate the zero padding value based on the total number of bins.
+        # If hash_prefix_len = 8 then number_of_bins = 256 and zero_padding = 3
+        # meaning the third bin will have a suffix of "003".
+        self.number_of_bins = 2**hash_prefix_len
+        self.zero_padding = int(math.log10(self.number_of_bins)) + 1
```
lukpueh: Ha that's smart!  Should give the same result as `len(str(self.number_of_bins))`, right? I wonder if using `len` doesn't make it slightly easier to understand (and might also be more performant). But I'm fine either way.

Minor naming nit: Maybe something like `suffix_width` or so makes more sense? Because the amount of needed zero padding actually is `(int(math.log10(self.number_of_bins)) + 1) - len(bin_index)`.

jku: it's still decimal though and not hex...

I'm sorry if I sound like a broken record but I'm not sure if this has been addressed anywhere.  Maybe I'm missing something but I'd like to see this addressed in a comment at least: the spec seems to define the bin names containing hexadecimal yet the PR uses decimal?

lukpueh: You're right, TAP 15 says "*NAME_PREFIX-COUNT where COUNT is a **hexadecimal** value between 0 and 2^BIT_LENGTH-1*". So my suggestion would be `len(f"{self.number_of_bins:x}"))`.

tuf/api/metadata.py
```diff
@@ -421,6 +421,13 @@ def verify_delegate(
 
             keys = self.signed.delegations.keys
             role = self.signed.delegations.roles.get(delegated_role)
+            if role is None:
+                for r in self.signed.delegations.roles.values():
+                    if (
+                        r.succinct_hash_info is not None
+                        and r.succinct_hash_info.is_bin(delegated_role)
+                    ):
+                        role = r
```
jku: I think this makes a good point for possibly simplifying the TAP: this is probably close to the best we can do with the current design but it really isn't obvious why we do this in the order we do and why we loop through all delegated roles

(the alternative implementation that avoids looping through all roles for the current TAP design would take `delegated_role`, try to parse it as a succinct delegation, translate "prefix-NNN" to "prefix-0", try to lookup "prefix-0" in `delegations.roles`, etc: still not simple or obvious)

jku: and if it's not obvious to the reader: the verifying code needs a way to find a succinct delegation based on a rolename, assuming the succinct delegation delegates to that role name... this is not trivial when there can be other, normal, delegations and multiple succinct delegations in the same Delegations structure.

tuf/api/metadata.py
```diff
@@ -1258,17 +1267,173 @@ def to_dict(self) -> Dict[str, Any]:
...
+        # calculated for a small number of symbols required by hash_prefix_len.
+        needed_bits = self.hash_prefix_len  # aliased for emphasis
+        needed_bytes = math.ceil(needed_bits / 8)
+        hash_bits = "".join(
+            f"{one_byte:08b}" for one_byte in hasher.digest()[:needed_bytes]
+        )
+        needed_hash_bits = hash_bits[:needed_bits]
+
+        # Split hash calculation from the rest for easier testing.
+        return self._find_bin_for_bits(needed_hash_bits)
```
jku: I've not played with this yet but after looking at this and talking to Lukas, I think I finally understood what's being done here: it's essentially a right shift to get hash_prefix_len leftmost bits. I think if this is the case then I would prefer seeing the actual math here without the string-of-1-and-0.

```python
# we can't ever need more than 4 bytes (32 bits)
hash_bytes = hasher.digest()[:4]

# right shift so that we only have the leftmost hash_prefix_len bits we care about
bin_number = int.from_bytes(hash_bytes, byteorder="big") >> (32 - self.hash_prefix_len)
        
return self.get_bin_name(bin_number)
```

**EDIT** I tested and that does work (after I fixed the confusion I was having about digest() return value)

tuf/api/metadata.py
```diff
@@ -1258,17 +1267,173 @@ def to_dict(self) -> Dict[str, Any]:
...
+                targets URL.
+        """
+        hasher = sslib_hash.digest(algorithm="sha256")
+        hasher.update(target_filepath.encode("utf-8"))
+        # Get the bit representation of the hash. The bit representation is
+        # calculated for a small number of symbols required by hash_prefix_len.
+        needed_bits = self.hash_prefix_len  # aliased for emphasis
+        needed_bytes = math.ceil(needed_bits / 8)
+        hash_bits = "".join(
+            f"{one_byte:08b}" for one_byte in hasher.digest()[:needed_bytes]
```
jku: i'm not sure if this is a bug or just bad naming but a hexadecimal digit is not equal to a byte (but see my other comment: I'm proposing removing this code altogether)

jku: oh I may be misunderstanding the hashlib use here -- digest() maybe actually returns the bytes and not hex: in that case never mind me, I'm the one making a mistake

EDIT: Yes, my bad: I'm marking this resolved

tuf/api/metadata.py
lukpueh: 🤦  "Obviously"  bit shifting is the way to do this.

tuf/api/metadata.py
MVrachev: Oh...totally missed that...

tuf/api/metadata.py
MVrachev: You are correct that directly shifting the bits is easier to understand.
Will apply this. Thanks for the detailed comment!

tuf/api/metadata.py
MVrachev: I will leave a comment to be a little clearer.

tuf/api/metadata.py
MVrachev: I realized that besides the calculation of the zero-padding I will have to change the calculation of `suffix` in `get_bin_name` API call in order to save the suffix as HEX.

jku: I assume the reason for using hex here is that it makes the binary math a little easier for humans to verify: if your targetpath hash begins with "a1d8...", then in a 16 bit succinct delegation the correct bin name should be "prefix-a1d8" -- at least that's how I've understood it

MVrachev:
I think I have applied all suggestions by @jku  and @lukpueh.
Most notably I changed:
- the bin names suffix is now hex
- calculating the bin names is done by bit shifting
- `test_find_delegation_with_succinct_hash_info` is changed, so it's testing the whole process of calling `find_delegations` to reciving the actual bin name
- added multiple bin names case inside `test_succinct_hash_bins_graph_traversal` to make sure all bin files are existing in cases where zero padding has to be used (the bins are more than 10).

MVrachev:
I have implemented @lukpueh suggestion from the comment here https://github.com/MVrachev/tuf/tree/new-suggestion.
For those interested, please comment on issue #1943.

MVrachev:
Together with @jku, @lukpueh, and the TAP author - @mnm678 we met to discuss if we should change the design on TAP 15 based on multiple suggestions.
We end up deciding that there are multiple changes that we want to make which are all listed in my comment [here](https://github.com/theupdateframework/taps/pull/154#pullrequestreview-973613582).
Marina updated the TAP 15 content in her pr https://github.com/theupdateframework/taps/pull/154 and there were multiple changes that I had to do in this pr to be up to date.
That's why I decided it's probably best if I close this pr, make the changes from scratch (reusing big chunks of this code of course) and present a cleaner history in a new pr.
This pr is superseded by https://github.com/theupdateframework/python-tuf/pull/2010

PR #1949, opened 2022-04-09
Updated tuf/api/metadata.py

### LINKED ISSUE #1938: unrecognized_fields is annotated as Mapping but maybe should be just Dict
I'm pretty sure I suggested annotating unrecognized_fields as a Mapping (so "immutable dict") at some point. I was probably thinking about the client use case but clearly in the repository case this annotation can be incorrect: we may actually want to modify the content.

`unrecognised_fields` throughout the API (tuf/api/metadata.py) should be annotated as Dict.

abs007:
Signed-off-by: Abhisman Sarkar abhisman.sarkar@gmail.com

Fixes #1938

Annotated 'unrecognized_fields' as Dict instead of Mapping

requirements**:

- [ ✔] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [-] Tests have been added for the bug fix or new feature
- [-] Docs have been added for the bug fix or new feature

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1,4 +1,4 @@
-
+# Copyright New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
 """
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -1,4 +1,4 @@
-# Copyright New York University and the TUF contributors
+
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
 """
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -130,11 +130,11 @@ class Metadata(Generic[T]):
         self,
         signed: T,
         signatures: Optional[Dict[str, Signature]] = None,
-        unrecognized_fields: Optional[Mapping[str, Any]] = None,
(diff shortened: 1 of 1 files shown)


jku:
* The DCO check is likely failing because the bot requires every commit to have Signed-off-by.
* squashing your changes into one commit and force pushing this branch is fine by me (so is a new branch with a new PR, totally up to you): result should be a branch with just a single commit on top of `develop`.
* I would prefer a more descriptive commit message (one that answer question _"Why?"_). Something like this would work
    ```
    Metadata API: Annotate 'unrecognized_fields' as Dict
    
    Annotating them Mapping seems to be wrong as we may actually want to
    modify the contents in the repository code
    
    Signed-off-by: Abhisman Sarkar abhisman.sarkar@gmail.com
    ```


As advice, I suggest using a branch per PR (so not using your `abs007:develop` branch): it probably makes things easier. That said, technically using `abs007:develop` works just as well -- it just means if you modify or update your develop branch, this PR changes when you push

Also, I agree with Lukas: new tests are not be required (since annotations do not affect runtime and mypy already checks annotations during lint). The annotations automatically become part of the docs so that's handled as well.

abs007:
Thanks for the help. I'll go on and make a new pr since as you said that further changes on develop would change this pr.

PR #1953, opened 2022-04-13
Tests: restore objects to initial state after test

MVrachev:
Inside `test_metadata_eq_.py` we test the `__eq__` implementations of all
classes. In order to do this, we change the attribute of the object and
then compare them to the unchanged version of those objects.
Usually, we do it in the following steps:
1. create an initial version `a`
2. create a copy of `a` called `b`
3. iterate all attributes inside `b` and change them to a given value
4. check that `a` and `b` are different

We do however forget to restore the object `b` to its initial state
which means we don't check the `__eq__` correctly as we stop on the
first, the found difference which could be of an older attribute changed
in one of the past iterations.

requirements**:

### DIFF
--- a/tests/test_metadata_eq_.py
+++ b/tests/test_metadata_eq_.py
@@ -61,6 +61,8 @@ class TestMetadataComparisions(unittest.TestCase):
         for attr, value in [("signed", None), ("signatures", None)]:
             setattr(md_2, attr, value)
             self.assertNotEqual(md, md_2, f"Failed case: {attr}")
+            # Restore the old value of the attribute.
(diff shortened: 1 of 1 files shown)


lukpueh (approved):
Good catch, @MVrachev! I must say though, a change like this would be easier if the test was table-driven, as we [discussed on the original PR](https://github.com/theupdateframework/python-tuf/pull/1775#discussion_r812938781). :P

As a matter of fact, if you had used the decorator instead of a for loop it wouldn't be necessary to restore the object at all.

PR #1954, opened 2022-04-14
Replaced manual path construction with os.path.join

### LINKED ISSUE #1903: Replace manual path construction with "os.path.join"
I found a couple of places in the tests where we have used `f` strings and manually joined paths instead of using `os.path.join()`.
This can lead to a problem where on Linux and Mac OS this behavior will work, but on Windows won't.
Some examples:
- https://github.com/theupdateframework/python-tuf/blob/e7037cf8c42022ea20f3b7891a50aa3ed563521b/tests/test_api.py#L236
- https://github.com/theupdateframework/python-tuf/blob/e7037cf8c42022ea20f3b7891a50aa3ed563521b/tests/test_api.py#L162

When discussing with @jku why don't we get CI failures from this behavior he found out that:
> File I/O in the Windows API converts forward slashes ("/") to backslashes ('\'), before passing the request to the native API

Even though that's the case, it's better if we replace those instances of manual path construction with the preferred `os.path.join`.
We should not forget we are developing `python-tuf` for Windows users as well. :)

aribasch:
Fixes #1903

**Description of the changes being introduced by the pull request:**
Replaces the use of f-strings to manually join paths with os.path.join() in the test suite. All updated tests adhere to style guidelines.

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -134,7 +134,9 @@ class TestMetadata(unittest.TestCase):
 
         # Test serializing to a file with bad filename
         with self.assertRaises(exceptions.StorageError):
-            md = Metadata.from_file(f"{self.repo_dir}/metadata/root.json")
+            md = Metadata.from_file(
+                os.path.join(self.repo_dir, "metadata", "root.json")
+            )
             md.to_file("")
 
     def test_compact_json(self) -> None:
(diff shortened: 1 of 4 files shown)


jku (changes requested):
Thanks for taking a stab at this! 

Two issues:
* We require a Developer Certificate of Origin (meaning Signed-off-by line in commit message). See details in the end of https://github.com/theupdateframework/python-tuf/blob/develop/docs/CONTRIBUTING.rst#dco
* Changes to local file system paths look correct but I believe URLs were actually correct already: URL separator does not change depending on Operating System, it's always "/"

aribasch:
Thanks for the pointers @jku! I reverted the URL construction back to f-strings and signed the commit message. Let me know what you think.

lukpueh (approved):
Thanks, @aribasch!

PR #1960, opened 2022-04-19
Update supported spec version

mnm678:
Solves #1959


jku (approved):
looks good, thanks. I assume the generated test update process was reasonable and not annoying? I'm asking since I think this is the first time the test triggered in unrelated changes

mnm678:
> looks good, thanks. I assume the generated test update process was reasonable and not annoying? I'm asking since I think this is the first time the test triggered in unrelated changes

It was pretty straightforward. I especially appreciated the error message that told me exactly how to fix it.

PR #1963, opened 2022-04-21
Checking for None instead of falsyness

### LINKED ISSUE #1937: unrecognised_fields default value vs empty container
I believe this code currently does the unexpected thing:

```python
custom = {}
role = Role([], 1, custom)
custom["customfield"] = "value"
assert "customfield" in role.unrecognized_fields
```

As in the roles unrecognised fields now does not include "customfield". This is because the initialization `self.unrecognized_fields = unrecognized_fields or {}` acts surprisingly when the input container is empty. We should check for None instead of falsyness.

Same thing applies throughout the API where unrecognized fields are used.

abs007:
Metadata API: Checking for None instead of falsyness for unrecognized_fields

Fixes #1937

Initialization of unrecognized_fields acts surprisingly when the input
container is empty. Hence, We're checking for None instead of falsyness.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -134,10 +134,10 @@ class Metadata(Generic[T]):
     ):
         self.signed: T = signed
         self.signatures = signatures if signatures is not None else {}
-        if unrecognized_fields is not None:
-            self.unrecognized_fields = unrecognized_fields
-        else:
-            self.unrecognized_fields = {}
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Metadata):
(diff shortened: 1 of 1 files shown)


lukpueh (commented):
Thanks for the patch, @abs007! Two style comments:

1. There is no need to add PR template wording such as *"Description of the changes being introduced by the pull request:"* to the commit message
2. IMO the following construct is a bit easier on the eyes...

```python
if unrecognized_fields is None:
    unrecognized_fields = {}

self.unrecognized_fields = unrecognized_fields
```

abs007:
Agreed. That does look better. I'll make the changes

abs007:
I had a formatting issue earlier but I resolved that. For some reason, the signing error persists even though I've added my signature. 
Could you look into this, thanks.

abs007:
Closing this pr and opening a new one for this issue as I think that might solve the signature issue.

lukpueh:
Hi @abs007, usually there is no need to close/re-open pull requests, as long as you don't need to switch the branch of the PR. I recommend reading our [git history guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/git-history.md). It has some information about how to rewrite the history in a given git branch.

abs007:
Thanks for the info. 
I unfortunately couldn't understand a part in the example provided by @awwad for rebasing commits, made on Dec 6, 2018. 
Specifically, the third paragraph "I've fixed this using `git rebase -i` ..."  

Inkedgit_LI

Could you provide resources to understand this, thanks.

lukpueh:
> Could you provide resources to understand this, thanks.

Did you check out the resources about `git rebase` in the git history guide I shared with you above?

lukpueh:
Don't worry too much about @awwad's comment, it is not supposed to be a general tutorial for `git rebase`.

abs007:
> Did you check out the resources about git rebase in the git history guide I shared with you above?

Haven't gone through that. Will do.

PR #1970, opened 2022-04-21
Update build config to include examples dir

### LINKED ISSUE #1901: examples not included in sdist
Examples should be part of sdist but they seem to be missing from the MANIFEST.

See also #1896 for proposed changes to build metadata

lukpueh:
Fixes #1901

Explicitly include `examples` dir in sdist.

The same would be achieved, by removing explicit includes, which
currently would also add these files/dirs:

```
/gitattributes
/github
/mypy_cache
/readthedocs.yaml
/verify_release
```

Maybe we should instead of defining includes, explicitly exclude
(some of) these files? The advantage of a blacklist approach is
that it becomes less likely to forget including files that should
be in included.

See hatch docs for:
- what files should be in sdist
https://ofek.dev/hatch/latest/plugins/builder/#source-distribution

- what files get into sdist by default:
https://ofek.dev/hatch/latest/plugins/builder/#default-file-selection_1

- how to configure what files get into sdist:
https://ofek.dev/hatch/latest/config/build/#file-selection

requirements**:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ path = "tuf/__init__.py"
 [tool.hatch.build.targets.sdist]
 include = [
   "/docs",
+  "/examples",
   "/tests",
   "/tuf",
   "/requirements*.txt",


joshuagl (approved):
LGTM, thanks.

I am in favour of configuring a deny/exclude list, rather than an include list.

Issue #1967, opened 2022-04-21
cd: consider replacing 3rd-party GitHub actions


lukpueh:
**Description of issue or feature request**:
https://github.com/theupdateframework/python-tuf/pull/1946 adds a CD workflow to release build artifacts on PyPI and GH upon successful completion of the CI workflow for a pushed release tag.

The PR adds 3rd-party actions `softprops/action-gh-release` and `pypa/gh-action-pypi-publish`, which goes against our *"only allow actions created by GitHub"*-policy. It does not seem like a big deal, because the former action is popular and recommended by the archived GitHub-native [action/create-release](https://github.com/actions/create-release), and the latter is hosted by the [PyPA](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/), which also hosts twine, the tool we usually use to publish on PyPI.

We could at least see how hard it would be to implement the required logic for those actions ourselves 

**Current behavior**:
`cd.yml` uses 3rd-party GitHub actions

**Expected behavior**:
Assess required effort to replace 3rd-party GitHub actions with custom code, e.g. a [shell script](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun) that runs `twine` for pypi uplaod, and a github script that uses [`octokit` to create a release](https://octokit.github.io/rest.js/v18#repos-create-release) (see [custom finalize release action](https://github.com/theupdateframework/python-tuf/blob/b99d0432a7f0a803c492580085e85f81908e0dc1/.github/workflows/cd.yml#L74-L87) for inspiration for the latter).

jku:
this was improved on at some point: we only use `pypa/gh-action-pypi-publish` and githubs actions now: this seems like the right set.

PR #1971, opened 2022-04-22
Refactor ci/cd workflows to make more robust

### LINKED ISSUE #1961: ci/cd: make workflow pipeline more robust
#1946 adds a CD workflow to release build artifacts on PyPI and GH upon successful completion of the CI workflow for a pushed release tag.

Unfortunately, the information about CI that we can access from CD is [vague](  https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run
) and thus makes the workflow pipeline a bit brittle (see details below). 

**Current behavior**:
CD triggers for each completed run of CI, regardless of the ref that triggered CI, but then skips execution if the CI run was not successful or the ref (branch or tag) that triggered CI does not start with a 'v'. In order to not trigger CD for a branch that starts with 'v', we rely on the configuration of CI, which currently does not trigger on other branches than 'develop'. 

**Expected behavior**:
CD triggers for completed and successful run of CI but only if CI was triggered by a release tag, regardless of CIs trigger conditions.

**Alternative solution**:
Restructure workflows, so that all test logic from 'ci.yml' is moved to a separate workflow file '_test.yml', that can be included in both the CI workflow, which then only triggers on push to 'develop'-branch, and the CD workflow, which then only triggers on on push to 'v*'-tag.

### LINKED ISSUE #1969: cd: workflow could have a permissions section
**Current behavior**:
https://github.com/theupdateframework/python-tuf/pull/1946 adds a CD workflow to release build artifacts on PyPI and GH upon successful completion of the CI workflow for a pushed release tag.

The newly added workflow does not yet configure  [`permissions`](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs).

**Current behavior**:
No permission config in cd.yml

**Expected behavior**:
Permission config cd.yml

lukpueh:
Fixes #1961,
Fixes #1969 

*I recommend reviewing commit by commit!*

Prior to this change, ci triggered cd depending on the event that triggered ci. Due to the vague information about that event available to cd, the workflow pipeline was a bit brittle. This change disassociates ci and cd workflows to allow for an independent configuration of trigger events. The test jobs, which used to be defined in ci, are now in a separate workflow file _test.yml that can be included in both ci and cd workflows.

**Changes in ci**
- Only defines trigger events and permissions, the "meat" of ci is defined in the called _test.yml now.
- No longer triggers on tag pushes, this was only needed for cd.

**Changes in cd**
- Now triggers directly on tag pushes instead of (cd)-workflow_run.
- Calls _test.yml, and require successful run before build/release. (`needs: test` replaces `if: ...`)
- Changes variable names about pushed tag that triggered the event.

requirements**:


lukpueh:
Looks like we need to update the required checks in the repo settings, because the names of jobs have changed with this PR.

We also need allow 3rd-party actions for the CD workflow until we fix #1967.

lukpueh:
Also note that CD is still triggered by each CI run, and fails due to the 3rd-party action permission issue, even if its execution would be skipped based on what triggered ci: https://github.com/theupdateframework/python-tuf/actions/workflows/cd.yml

After this PR CD is really only triggered on tag push.

joshuagl (approved):
Very nice cleanup, thanks!

jku (approved):
Looks good, I have nothing to add.

lukpueh:
Thanks for the reviews, I rebased and force pushed to (auto-)resolve a conflict. See note about non-completed checks above https://github.com/theupdateframework/python-tuf/pull/1971#issuecomment-1106204097.

Will merge and update required checks.

PR #1974, opened 2022-04-24
chore: Dependency Review Action

naveensrinivasan:
Dependency review is a tool that helps you identify and fix vulnerabilities in your dependencies. By checking the dependency reviews in a pull request and changing any dependencies that are flagged as vulnerable, the project can avoid vulnerabilities being added to your project. https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,20 @@
+# Dependency Review Action
+#
+# This Action will scan dependency manifest files that change as part of a Pull Reqest, surfacing known-vulnerable versions of the packages declared or updated in the PR. Once installed, if the workflow run is marked as required, PRs introducing known-vulnerable packages will be blocked from merging.
+#
+# Source repository: https://github.com/actions/dependency-review-action
+# Public documentation: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v1
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -15,6 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@v3
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
       - name: 'Dependency Review'
-        uses: actions/dependency-review-action@v1
+        uses: actions/dependency-review-action@a9c83d3af6b9031e20feba03b904645bb23d1dab


jku:
So 
* Dependency Review seems to be a brand new action from GitHub.
* It warns about dependency changes that have vulnerabilities -- so the exact same thing Dependabot does, but on PRs instead of a cron job?
* Dependency review includes details of changes to indirect dependencies -- (this _should_ not affect us as we try to keep everything listed in requirements files). It seems to do this by using the dependency graph api however, so uses a different mechanism: we'd be using a Github service to get our dependency graph instead of just running software developed by github...

What it actually changes in the workflow is buried three links deep in the docs but the result seems to be this in the review UI:
image

lukpueh:
> What it actually changes in the workflow is buried three links deep in the docs but the result seems to be this in the review UI:
...

The feature referenced here does not come with `actions/dependency-review-action`, but is already enabled by default for all public repos (see e.g. the [rich diff for `requirements-test.txt` in a recent Dependabot PR](https://github.com/theupdateframework/python-tuf/pull/2013/files?short_path=685da80#diff-685da804fbcac569d75387e475e57d1de687a54c6c41b3aa4057694cfb5abc4b)). 

The /=dependency-review-action basically makes any vulnerabilities shown in that diff more visible by displaying them in the checks box on the front page of a PR, with the option to block merge if the action fails.

I agree with @jku that it probably does not add much value since we already use [`Dependabot alerts`, which are sent whenever](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts#detection-of-vulnerable-dependencies):
> - A new vulnerability is added to the GitHub Advisory Database.
> - The dependency graph for a repository changes.

OTOH, there is probably not much harm in enabling the action (it is hosted by GitHub, only needs read permission), other than probably making builds take slightly longer and adding another config file that we need to maintain. Furthermore, the Dependabot alerts docs cited above further say:
> Additionally [besides Dependabot alerts], GitHub can review any dependencies added, updated, or removed [...] This allows you to spot and deal with vulnerable dependencies before, rather than after, they reach your codebase.

I have no strong preference about this, but lean a bit towards merging and see how it goes.

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,20 @@
...
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v1
```
jku: For both "uses" lines: we're trying to use hashes for actions instead of tags. This is because tags are mutable and in practice GitHub does move the tags so our workflows end up changing without us knowing that they change. 

This might not be that important for this workflow (since it's read permissions only) but for consistency, could you update this commit so that it refers to the specific commit hashes instead of tags? There are examples in our other workflows.

jku (changes requested):
> I have no strong preference about this, but lean a bit towards merging and see how it goes.

Works for me, let's try it out. I have one change request, will leave comment about that.

.github/workflows/dependency-review.yml
naveensrinivasan: I have updated them to SHA. PTAL. Thanks

.github/workflows/dependency-review.yml
```diff
@@ -15,6 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@v3
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
       - name: 'Dependency Review'
-        uses: actions/dependency-review-action@v1
```
jku: wow: I think your changes are correct and I'll approve but... look at where "v1" actually points to: the initial commit event though there are several 1.0.x releases after that! Further proof that actions versioning is really not well thought out.

jku (approved):

.github/workflows/dependency-review.yml
jku: they fixed this now... apparently the project had 1600 users and no-one had noticed they were all running the initial commit

PR #1975, opened 2022-04-25
Checking for None instead of falsyness

### LINKED ISSUE #1937: unrecognised_fields default value vs empty container

abs007:
Metadata API: Checking for None instead of falsyness for unrecognized_fields

Fixes #1937

Initialization of `unrecognized_fields` acts surprisingly when the input
container is empty. Hence, We're checking for `None` instead of falsyness.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -134,7 +134,10 @@ class Metadata(Generic[T]):
     ):
         self.signed: T = signed
         self.signatures = signatures if signatures is not None else {}
-        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields or {}
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Metadata):
@@ -519,7 +522,10 @@ class Signed(metaclass=abc.ABCMeta):
             raise ValueError(f"version must be > 0, got {version}")
         self.version = version
 
-        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields or {}
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Signed):
@@ -639,7 +645,10 @@ class Key:
         self.keytype = keytype
         self.scheme = scheme
         self.keyval = keyval
-        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields or {}
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Key):
@@ -797,7 +806,10 @@ class Role:
             raise ValueError("threshold should be at least 1!")
         self.keyids = keyids
         self.threshold = threshold
-        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields or {}
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Role):
@@ -1068,7 +1080,10 @@ class MetaFile(BaseFile):
         self.version = version
         self.length = length
         self.hashes = hashes
-        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields or {}
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, MetaFile):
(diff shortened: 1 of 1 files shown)


abs007:
Same signature error persists here too ;_;

jku:
Thanks. The change itself looks good, just some tweaking required for the CI to pass
* Our linter black calls itself the _"uncompromising code formatter"_ and you seem to have found out it really is uncompromising... Sorry about that. If you've installed the development requirements you can run all CI tests locally with `tox` or just the linters with `tox -e lint`. If black complains about trivial whitespace issues you can tell black to just fix them (`black tuf/api/metadata.py`) and add the changes into your commit after checking that they look correct. 
* I would ask you to combine all changes into a single commit as they are logically one change (force pushing to the same branch is totally fine)
* make sure the commit (in fact all commits, if you end up with several) has a signed-off-by

jku:
> Same signature error persists here too ;_;

If you're wondering how this appeared: at least one of the six commits does not contain a signed-off-by (all commits are required to have it).

There's a couple of way to make this easier, this is what I do:
* I have a file `.git/signoff` that contains an empty line and then my signed-off-by line
* I have run `git config commit.template .git/signoff`

This way the message for every _new_ commit  starts with my signed-off-by line already in there

tuf/api/metadata.py
```diff
@@ -134,7 +134,10 @@ def __init__(
     ):
         self.signed: T = signed
         self.signatures = signatures if signatures is not None else {}
-        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields or {}
+        if unrecognized_fields is None:
+            unrecognized_fields = {}
+
+        self.unrecognized_fields = unrecognized_fields
```
MVrachev, line 140: I think it would be better if you add some information about the expected type of `unrecognized_fields` like:
```suggestion
        self.unrecognized_fields: Dict[str, Any] = unrecognized_fields
```
That applies to all places where you have used `self.unrecognized_fields = unrecognized_fields`

jku: I wouldn't object to that but I'm not going to require it either: static checkers will correctly deduce the type as `Dict[str, Any]` here (because of the original argument type) and that's enough for me.

abs007:
> * Our linter black calls itself the _"uncompromising code formatter"_ and you seem to have found out it really is uncompromising... 

Wish it could've compromised just this time 😿
Had a question when going through the different tools used for automatic testing. Is linting used for debugging and code formatters for beautifying the code?

> * I would ask you to combine all changes into a single commit as they are logically one change (force pushing to the same branch is totally fine)

Thinking of running `git rebase -i HEAD~6` and then squashing the commits. Would I need to `git fetch -all` and `git rebase <upstream>/<main branch>` before running this?

lukpueh:
> Had a question when going through the different tools used for automatic testing. Is linting used for debugging and code formatters for beautifying the code?

I guess you can put it that way. A linter usually only analyzes the code, e.g. if it has a correct format, or if it uses certain constructs that are not recommended. A code formatter, on the other hand, actually applies a certain format to the code, but it does not deal with non-format related issues. Note that our code formatter black can also be used in lint mode, where it only checks if the format is correct, but does not change anything.

> Thinking of running git rebase -i HEAD~6 and then squashing the commits. Would I need to git fetch -all and git rebase <upstream>/<main branch> before running this?

Depends on what you are trying to achieve. If you only want to squash your commits, you don't have to fetch and rebase on `<upstream>/<main branch>`.

jku:
> Had a question when going through the different tools used for automatic testing. Is linting used for debugging and code formatters for beautifying the code?

I use "lint" as an umbrella term for finding both stylistic and actual programming errors. But black is definitely just a "beautifier": it's an attempt to avoid arguing about the correct style :) 

> > * I would ask you to combine all changes into a single commit as they are logically one change (force pushing to the same branch is totally fine)
> 
> Thinking of running `git rebase -i HEAD~6` and then squashing the commits. Would I need to `git fetch -all` and `git rebase <upstream>/<main branch>` before running this?

rebase and squash all sounds good: the commit message is the only thing that should need work then. 

You can even combine the fetch & rebase-squash:
```
git fetch <upstream>
git rebase -i <upstream>/develop
```
that should give you an interactive rebase on top of current develop with your commits listed

abs007:
Finally got it ☺. Got to learn a lot from all the info and tips. Thanks @jku and @lukpueh.

lukpueh (approved):
Thanks for the fix, @abs007!

abs007:
> There's a couple of way to make this easier, this is what I do:
> 
> * I have a file `.git/signoff` that contains an empty line and then my signed-off-by line
> * I have run `git config commit.template .git/signoff`
> 
> This way the message for every _new_ commit starts with my signed-off-by line already in there

Would this add the signature to say something like, merge commit messages.

jku:
I think it would if you made a merge? I'm not even 100% sure about that since the way python-tuf is developed we don't normally end up with manual merge commits in tree: PRs are linear (so no merges in the PR) and they are then typically merged to develop branch via the GitHub web UI

abs007:
Got it.

PR #1989, opened 2022-04-28
build: Pin hatchling version

### LINKED ISSUE #1984: build: reproducibility issue with hatch version
I'm currently seeing our wheel file contents changing without changes to source:

```
diff -r locally-built-whl/tuf-1.1.0.dist-info/WHEEL github-whl/tuf-1.1.0.dist-info/WHEEL
2c2
< Generator: hatch 0.23.0
---
> Generator: hatch 0.22.0

```

So yes, we need to pin hatch version. This makes a lot of sense to me -- why would we not pin the build tool when we pin all  other dependencies?

I'm also a bit worried about that 0.22? Where did that come from -- I don't see it on pypi?

jku:
Pin hatchling version. This version should be kept up-to-date: my working assumption is that Dependabot will handle it.

I think we want to do something like this: IMO our build artifacts should not change if hatchling releases a new version (which is what happened right after we released 1.1.0: now `verify_release` will say the artifacts do not match because WHEEL file includes the hatchling version number).

Whether this exact change is the correct one, I'm not sure. Does dependabot update `build-system.requires`? I'm hoping so but I don't know :shrug:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
-# Build-system section
 [build-system]
-requires = ["hatchling>=0.21.0"]
+# hatchling pinned for reproducibility: version should be kept up-to-date
+requires = ["hatchling==0.22.0"]
 build-backend = "hatchling.build"
 
 [project]


pyproject.toml
```diff
@@ -1,6 +1,6 @@
-# Build-system section
 [build-system]
-requires = ["hatchling>=0.21.0"]
+# hatchling pinned for reproducibility: version should be kept up-to-date
+requires = ["hatchling==0.22.0"]
```
jku, line 3: Actually current version is 0.23.0 now... but maybe this is a good way to test what dependabot actually does (set version to less than current, expect dependabot to file an issue) ?

joshuagl: Clever :-)

joshuagl (approved):
> I think we want to do something like this: IMO our build artifacts should not change if hatchling releases a new version (which is what happened right after we released 1.1.0: now `verify_release` will say the artifacts do not match because WHEEL file includes the hatchling version number).

💯 – very strongly agree.

Change LGTM, lets see if dependabot does pick it up.

lukpueh (approved):
Thanks, @jku

ofek:
Not yet https://github.com/dependabot/dependabot-core/issues/3290

PR #1991, opened 2022-04-29
lint: Stop using requests annotations

### LINKED ISSUE #1988: reconsider types-requests dependency
We get a lot of dependency bumps from dependabot for types-requests: The dependency is in requirements-test.txt, I assume it brings in annotation stubs for requests so that mypy does not complain during lint.

These upgrades are annoying because typeshed (the project producing this package) does not make real releases. They don't even tag the releases in source: the packages just appear in pypi.org. So there is absolutely no way to know if the new version is something we want or what source version it represents -- meaning no way to review a dependency upgrade.

Our use of requests is pretty stable: no changes in months. I wonder if we should just tell mypy to ignore requests and stop depending on types-requests?

jku:
requests project does not maintain annotations: typeshed project tries
to do it for them, and releases the annotations as "types-requests".

There's two main problems:
* typeshed releases constantly: this means a lot of test dependency
  updates
* typeshed releases are not tagged in git: updates are impossible to
  review

The benefit we get from types-requests is minimal as there is very
little requests-related code and it does not change often.

Remove annotations to lower the test dependency update churn.

---

Fixes #1988

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -149,6 +149,7 @@ disable_error_code = ["attr-defined"]
 
 [[tool.mypy.overrides]]
 module = [
+  "requests.*",
   "securesystemslib.*",
   "urllib3.*"
 ]
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -10,4 +10,3 @@ isort==5.10.1
 pylint==2.13.7
 mypy==0.942
 bandit==1.7.4
-types-requests==2.27.20


mnm678 (approved):

jku:
types-requests has had 6 updates in 7 days proving that we do need a solution to this... I'll merge this but I am open to other suggestions if someone has them

PR #1997, opened 2022-05-07
[BugBash] fix: Incompatible parameter type passed to MetaFile

dhavalgshah:
Fixes Pyre warning from Sonatype Lift

Signed-off-by: Dhaval Shah <30974879+dhavalgshah@users.noreply.github.com>

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -313,9 +313,13 @@ class RepositorySimulator(FetcherInterface):
         if self.compute_metafile_hashes_length:
             hashes, length = self._compute_hashes_and_length(Snapshot.type)
 
-        self.timestamp.snapshot_meta = MetaFile(
-            self.snapshot.version, length, hashes
-        )
+        metafile_dict = {
+            "version": self.snapshot.version,
+            "length": length,
+            "hashes": hashes,
+        }
+
+        self.timestamp.snapshot_meta = MetaFile.from_dict(metafile_dict)
 
         self.timestamp.version += 1
 
@@ -326,9 +330,14 @@ class RepositorySimulator(FetcherInterface):
             length = None
             if self.compute_metafile_hashes_length:
                 hashes, length = self._compute_hashes_and_length(role)
+            metafile_dict = {
+                "version": delegate.version,
+                "length": length,
+                "hashes": hashes,
+            }
 
-            self.snapshot.meta[f"{role}.json"] = MetaFile(
-                delegate.version, length, hashes
+            self.snapshot.meta[f"{role}.json"] = MetaFile.from_dict(
+                metafile_dict
             )
 
         self.snapshot.version += 1


tests/repository_simulator.py
```diff
@@ -313,9 +313,13 @@ def update_timestamp(self) -> None:
...
-        self.timestamp.snapshot_meta = MetaFile(
-            self.snapshot.version, length, hashes
-        )
+        metafile_dict = {
+            "version": self.snapshot.version,
+            "length": length,
+            "hashes": hashes,
+        }
+
+        self.timestamp.snapshot_meta = MetaFile.from_dict(metafile_dict)
```
jku, line 322: from_dict() exists as a helper to deserialization, I think using them as constructors is a code smell.

Can you explain what the incompatibility is in the original code (it's not immediately clear to me)?

dhavalgshah: ### Reason for commit
Pyre throws the following warning:
`Expected int for 1st positional only parameter to call MetaFile.__init__ but got Optional[int].`

### Why use from_dict?
I used `MetaFile.from_dict()` thinking it is the convention as other tests use a similar approach.
https://github.com/theupdateframework/python-tuf/blob/c5ca38f0ae86cadad81e8adf99df1566cdee54da/tests/test_metadata_eq_.py#L162-L167

### Other approach
Use a temporary var for `self.snapshot.version` and guard it with the if None or if X clause.

jku: That's interesting... My first instinct is that this might be a bug in Pyre: `self.snapshot.version` should not be Optional and should always be an int and this is how mypy (another static type checker) handles it. I can see how a static analyzer could be confused (Snapshot constructor does accept None as version) but the internal variable is never set to None.

Assuming my analysis is correct (and this is a bug in Pyre), I don't want more complex code like in this commit: I wouldn't object to changes to e.g. `Signed.__init__()` (where Snapshot.version is set) if they don't make the code harder to read and at the same time make life easier for static checkers.

You are correct that other tests do sometimes use "`from_dict()` as constructor" approach. I don't think it's good style (except when testing deserialization) and we should rather improve constructors if they are not good enough.

jku:
I had never tried Pyre before but some tests seem to imply it's not really at mypy level in intelligence yet... or is not trying to have the sort of unassisted typing that mypy can do: it seems to require explicitly typing everything.

[Example](https://pyre-check.org/play?input=%0Afrom%20typing%20import%20*%0A%0Aclass%20Test%3A%0A%20%20%20%20def%20__init__(self%2C%20i%3A%20Optional%5Bint%5D)%20-%3E%20None%3A%0A%20%20%20%20%20%20%20%20if%20i%20is%20None%3A%0A%20%20%20%20%20%20%20%20%20%20%20%20i%20%3D%200%0A%20%20%20%20%20%20%20%20self.i%20%3D%20i%0A%0Areveal_type(Test().i)%0A): This is very close to what python-tuf does in this case: it's clear to human reader (and mypy) that Test.i is always an int but Pyre does not see that.

tests/repository_simulator.py
dhavalgshah: There's a lot going in `Signed.__init__()` that I find interesting.

https://github.com/theupdateframework/python-tuf/blob/c5ca38f0ae86cadad81e8adf99df1566cdee54da/tuf/api/metadata.py#L495-L502

Do we need `Optional` type in `version` and `spec_versions` arguments in `__init__`? Both of them are immutable types.

~Can this work?~

Update: This will not work. Working on a different approach.
````
    def __init__(
        self,
        version: int = 1,
        spec_version: str = ".".join(SPECIFICATION_VERSION),
        expires: Optional[datetime],
        unrecognized_fields: Optional[Dict[str, Any]],
    ):
````

dhavalgshah:
I guess you're right. I'll close this PR and focus on good-first-bugs instead.

tests/repository_simulator.py
jku: Yeah, it's a bit tricky. It's indeed possible that there is a design where Optional is not required for version, spec_version or even expires (immutability isn't required by python itself as you could set the default to None and then handle that in code) but it's not quite as simple as that as you probably found out... 

Basically:
* we want easy to use constructors for the signed-derivatives that don't require spelling out all arguments
* we want to not write same code twice for constructor and the deserializing code paths -- which is why we currently call constructors from `from_dict()` helpers
* preferably this all happens without breaking current API

The current implementation mostly gives us these benefits **and** mypy can understand it...

jku:
> I guess you're right. I'll close this PR and focus on good-first-bugs instead.

Thanks for filing though: Good annotations and type safety is something we're quite interested in... There's almost certainly possible improvements in that area

dhavalgshah:
Putting [this here](https://github.com/facebook/pyre-check/issues/612) for cross-referencing the discussion above wrt Pyre not correctly narrowing the types. Based on what I have seen, both mypy and Pyre have quirks around how they infer the types. And sometimes it gets confusing, especially to newbies like me, or when code is provided without a proper context.

In the instance where class attributes are not annotated beforehand of the assignment in `__init__` function, mypy correctly infers the type from the assignment context whereas Pyre relies on param level type definition.

Issue #1966, opened 2022-04-21
cd: re-enable release signing


lukpueh:
**Description of issue or feature request**:
https://github.com/theupdateframework/python-tuf/pull/1946 adds a CD workflow to release build artifacts on PyPI and GH upon successful completion of the CI workflow for a pushed release tag.

The PR also removes instructions from RELEASE.md to [gpg sign release artifacts](https://github.com/theupdateframework/python-tuf/blame/31ca674aa0923e54d7486e5b4a068efb82a34641/docs/RELEASE.md#L31-L32) and add them to the GitHub release assets as part of the previously manual release process. However, the [installation docs](https://theupdateframework.readthedocs.io/en/latest/INSTALLATION.html#verify-release-signatures) still mention release signatures.

**Current behavior**:
No instructions / release process integration to sign release artifacts

**Expected behavior**:
Add instructions to sign release artifacts and integrate with release process

**Ideas**:
- quick-fix 1: sign in GitHub CD action
- quick-fix 2: sign locally and upload signatures to release assets manually (bonus: integrate signing/uploading with `verify_release` script)
- long-term fix: e.g. in-toto (https://github.com/theupdateframework/python-tuf/issues/529)

jku:
> quick-fix 1: sign in GitHub CD action

This is possible with a GPG key of course (where private key is in an environment secret) but Sigstore cosign has _experimental_ support for this as well: https://blog.chainguard.dev/zero-friction-keyless-signing-with-github-actions/
* requires the cosign-installer action (or otherwise access to cosign tool)
* uses GitHub as the OIDC authenticator (authenticates a specific repo and the deployment environment)
* uses an ephemeral key to sign the release
* uploads the signature to a public log

We'd need to at least document how to verify this signature (as signing/verifying non-container images is not as widely used in cosign). The actual signing command is also not listed in the blog post  (again as non-containers are a bit of an after thought)

It's also possible this is still a little too experimental, and we should reconsider this solution in six months.

jku:
This space seems to move very rapidly at the moment so just documenting a few things for myself if not others:
* lukas has a hand crafted in-toto solution linked above: it's neat but IMO not very easy to understand or currently to verify
* SLSA is building a "packaged" solution in [slsa-github-generator](https://github.com/slsa-framework/slsa-github-generator) -- this is new and seems to still be under active development
[...]

PR #2000, opened 2022-05-09
PoC: Release tuf with in-toto attestations and supply chain definition

### LINKED ISSUE #529: Add in-toto metadata to releases to validate the integrity of its software supply chain

lukpueh:
that is provided, the better.

Fixes #529
POC for "long-term fix" for #1966 

- Provides additional release instructions to locally generate in-toto links (attestations) for the 'git tag' step and each 'build' step (sdist and wheel).
- Configures cd build job on GitHub to also generate attestations for sdist and wheel builds.
- Adds script to generate generic in-toto layouts (supply chain definition) to verify the release: 
  This means for each build artifact that the release was tagged, and the tagged sources match the inputs to both local and cd builds, and the build outputs match the actual build artifacts available to the verifier. Moreover, the attestations must have been signed by the keys authorized in the layout.
- The PR also includes instructions on how to verify the build artifacts against the corresponding layout using the created attestations.

See [RELEASE_with_in-toto.md](https://github.com/lukpueh/tuf/blob/017031b761fe08e6e037c89fad7e6c42ab5d1eb0/docs/RELEASE_with_in-toto.md) for details.

requirements**:

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -31,10 +31,55 @@ jobs:
           python-version: '3.x'
 
       - name: Install build dependency
-        run: python3 -m pip install --upgrade pip build
+        run: python3 -m pip install --upgrade pip build in-toto[pynacl]
 
       - name: Build binary wheel and source tarball
-        run: python3 -m build --sdist --wheel --outdir dist/ .
+        env:
+            IN_TOTO_KEY: ${{ secrets.IN_TOTO_KEY }}
+            IN_TOTO_KEY_PW: ${{ secrets.IN_TOTO_KEY_PW }}
+        run: |
+          #######################################################
+          # Build and generate signed attestions with in-toto CLI
+
+          # Make signing key available to in-toto commands
+          echo -n "$IN_TOTO_KEY" > .in_toto/key
+
+          # Define patterns for files that need not be recorded as materials below
+          exclude=('__pycache__' 'build' 'htmlcov' '.?*' '*~' '*.egg-info' '*.pyc')
+
+          # Grab TUF version to construct build artifact names for product recording
+          version=$(python3 -c 'import tuf; print(tuf.__version__)')
+
+          # Build sdist and record all files in CWD as materials and the build artifact
+          # as product in a signed attestation 'sdist.<signing key id>.link'.
+          in-toto-run \
+            --step-name sdist \
+            --key .in_toto/key \
+            --key-type ed25519 \
+            --password "$IN_TOTO_KEY_PW" \
+            --materials . \
+            --products dist/tuf-${version}.tar.gz \
+            --exclude ${exclude[@]} \
+            --metadata-directory .in_toto \
+            --verbose \
+            -- python3 -m build --sdist --outdir dist/ .
+
+          # Build wheel and record all files in CWD as materials and the build artifact
+          # as product in a signed attestation 'wheel.<signing key id>.link'.
+          in-toto-run \
+            --step-name wheel \
+            --key .in_toto/key \
+            --key-type ed25519 \
+            --password "$IN_TOTO_KEY_PW" \
+            --materials . \
+            --products dist/tuf-${version}-py3-none-any.whl \
+            --exclude ${exclude[@]} dist/tuf-${version}.tar.gz \
+            --metadata-directory .in_toto \
+            --verbose \
+            -- python3 -m build --wheel --outdir dist/ .
+
+          # Remove signing key file
+          rm .in_toto/key
 
       - id: gh-release
         name: Publish GitHub release candiate
@@ -43,7 +88,9 @@ jobs:
           name: ${{ github.ref_name }}-rc
           tag_name: ${{ github.ref }}
           body: "Release waiting for review..."
-          files: dist/*
+          files: |
+            dist/*
+            .in_toto/*.link
 
       - name: Store build artifacts
         uses: actions/upload-artifact@6673cd052c4cd6fcf4b4e6e60ea986c889389535
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,10 @@ tests/htmlcov/
 .pre-commit-config.yaml
 .vscode
 
+# Ignore in-toto metadata
+.in_toto/*
+!.in_toto/create_layout.py
+
 # Debian generated files
 debian/.debhelper/
 debian/*-stamp
--- a/.in_toto/create_layout.py
+++ b/.in_toto/create_layout.py
@@ -1,0 +1,79 @@
+"""Script to generate two generic in-toto layouts to verify wheel & sdist independently.
+
+The layouts define two steps and an inspection:
+- step 1 (tag): used as initial reference for the source code
+- step 2 (build): requires its inputs to match the tagged sources
+- inspect: requires the actual final product available to the verifier, i.e. sdist or
+  wheel, to match the outputs of the build step (this is, where the two layouts differ).
+
+In addition the layouts define, which keys are authorized to provide attestations, and
+how many are required:
+- tag: any one maintainer
+- build: at least two of maintainers and online build job
+
+
+Usage:
+    # Create signing key pair for CD ('filepath' must be CD_KEY_PATH defined below)
+    python -c 'import securesystemslib.interface as i;\
+        i.generate_and_write_ed25519_keypair_with_prompt(filepath="cd_key")'
+
+    # Create unsigned layout files 'wheel.layout' and 'sdist.layout'
+    python create_layout.py
+
+    # Sign layout with maintainer key
+    in-toto-sign --gpg <gpg key id> -f wheel.layout
+    in-toto-sign --gpg <gpg key id> -f sdist.layout
+
+"""
+from in_toto.models.layout import Inspection, Layout, Step
+from in_toto.models.metadata import Metablock
+from securesystemslib.interface import import_ed25519_publickey_from_file
+
+MAINTAINER_KEYIDS = [
+    "e9c059ec0d3264fab35f94ad465bf9f6f8eb475a",  # Justin Cappos
+    "1343c98fab84859fe5ec9e370527d8a37f521a2f",  # Jussi Kukkonen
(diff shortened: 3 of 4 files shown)


lukpueh:
Although this PR is fully functional, I labeled it as PoC because it has a couple of UX pain points:

**Key distribution**
  - Signing key for CD build job attestation is [created locally and stored as GitHub secret (along with decryption password) in a one-time setup](https://github.com/theupdateframework/python-tuf/pull/2000/files#diff-265d46f53832a6abbc40020a5ead91dbfb350f008220f267e4bef87be349ffd4R14-R18), and [piped to a file during each CD build](https://github.com/theupdateframework/python-tuf/pull/2000/files#diff-ea3ea8c9932adc7ba8161ceda844fedd43b006848ef1140c050cbd7ea0788a18R45), in order for [in-toto to use it](https://github.com/theupdateframework/python-tuf/pull/2000/files#diff-ea3ea8c9932adc7ba8161ceda844fedd43b006848ef1140c050cbd7ea0788a18R57), which feels cumbersome and risky (layout and maintainer attestation keys, otoh, use gpg and thus benefit from existing key server network).
  - While public keys and authorization mappings for attestation signatures are published through the layouts, the layout signing keys and thresholds need to be defined and distributed out-of-band. This PoC assumes that the layout is signed with one maintainer gpg key (see note about threshold below). How to establish root trust in layout signing keys is still unclear. Relying on fingerprints from maintainers.txt in the repo to protect the repo does not seem ideal.
 
**Metadata distribution**
- CD build job attestations are published automatically on the GitHub release page, from where they must be downloaded by the verifier.
- There is no mechanism to easily distribute layouts and maintainer attestations for verification. This PoC assumes that the verifier has created layouts and maintainer attestations and they are available locally during verification
- There is also no mechanism to easily distribute layouts for threshold signing. In this PoC the layouts are signed only once.

**in-toto quirks**

In order for the verifier to link the paper trail of attestations to the actual final product, i.e. the build artifacts, a [layout must define a so-called inspection](https://github.com/theupdateframework/python-tuf/pull/2000/files#diff-f6e8b48adb2ba0b43943aa82410cd91714f66e35cdb3b6d3307c73e5e335ec83R66-R74), which in turn must define a shell command that is executed upon verification, even if no command is needed. Besides unnecessarily running a command, this can also fail the entire verification, if the command defined in the layout is not available to the verifier.

Ideally, the inspection command is relevant to the verification. In our case this could be an untar/unzip command to open up the build artifact and, by defining further artifact rules in the layout, match its contents to the tagged sources earlier in the supply chain. To not further bloat the PR, I only defined the dummy inspection as described above.

Another minor ux nuisance is that when the verifier runs the inspection command, it records all files from cwd as materials and products into an inspection attestation, which can then be used to link the final product to the products reported by the build attestation. For this to work, verification has to be run from a [clean directory that contains only the build artifact to be verified](https://github.com/theupdateframework/python-tuf/pull/2000/files#diff-265d46f53832a6abbc40020a5ead91dbfb350f008220f267e4bef87be349ffd4R138).

See in-toto/docs#27 for a discussion about (dummy) inspections and linking the final product.

lukpueh:
FYI: I also played around with a hatch plugin to generate in-toto attestations transparently by just calling `python -m build ...`, instead of wrapping the build command with `in-toto-run [opts] -- [cmd]`:  see https://github.com/lukpueh/tuf/commit/f89dcbdf84ac5ac81c218238360a089a78baa7cd

For this PoC it made sense to just use the verbose `in-toto-run` command, but a hatch plugin could benefit the broader community. 

(cc @ofek, author of hatch and in-toto adoption veteran)

lukpueh:
One more thing I'd like to add to the setup is creating attestations for the cd test runs and mandate them by the layout. But I'll likely do this in a follow-up PR.

jku:
> FYI: I also played around with a hatch plugin to generate in-toto attestations transparently by just calling python -m build ..., instead of wrapping the build command with in-toto-run [opts] -- [cmd]

I'm still ingesting the whole thing so this is just a comment on this specifically:

The build system plugin feels like a good idea but I do have a fundamental issue with injecting intoto into the python-tuf build system in general like the example does: we're adding a build dependency (an unversioned one even, although I'm sure that's because it's a PoC) for no benefit to the actual build.

I'm not saying the in-toto maintainers are not trustworthy :stuck_out_tongue:, but I'm not sure that an attestation mechanism should be in the actual build pipeline. I thought I could just use it to document inputs and outputs, without giving the tool the ability to compromise our build system. Is there a real advantage to letting in-toto run the whole build? It seems to me like the build command is just another input to attest -- and in fact in our case it's already attested if cd.yml is part of the source source attestation

jku:
Example attestations seem to be here
https://github.com/lukpueh/tuf/releases/tag/v1.2.dev26

lukpueh:
> I thought I could just use it to document inputs and outputs, without giving the tool the ability to compromise our build system. Is there a real advantage to letting in-toto run the whole build?

No real advantage except for easier invocation maybe, and the ability to automatically record "byproducts" such as standard streams in the attestations, which I am not doing at the moment.

To record an attestation with in-toto, without having it run the command, you can do: `in-toto-record start && python -m build && in-toto-record stop`, which is what I also tried out at some point, while preparing this PR, because it gives better control over which materials/products should be included in the attestation: see [lukpueh/tuf@release-attestations-with-record](https://github.com/lukpueh/tuf/blob/release-attestations-with-record/.github/workflows/cd.yml#L45)).

adityasaky:
cc @SolidifiedRay

trishankatdatadog:
Heck, yes! Now I wonder if you could use TUF (e.g., PEP 480 and/or SigStore Trust Delegations) to somehow bootstrap the whole thing (pointing to in-toto metadata inside)...

adityasaky:
We should combine efforts on this and https://github.com/in-toto/in-toto/pull/444 to get them both over the line, especially with bootstrapping in-toto metadata using TUF.

jku:
I've given this feedback verbally but maybe should do it in writing too: the core requirement for any build integrity improvement is a  _verification process that can be and is used in the real world_. If the verification is not easily usable and and understandable by humans, then the build integrity work may be cool but is in the end academic. This is the reason I've not made a proposal of e.g. sigstore signing yet: I could easily sign but what's the point if verification requires a crypto & certificate specialist? 

The core idea of a formal way to require CD _and_ maintainer attestations is really cool but still, I would like to see a way to easily verify at least some of this in a practically useful way before we merge... Otherwise it feels like adding a lot of new process (and code to maintain) right now for vague potential value some time in the future.

JustinCappos:
Can you say more about what you mean by "practically useful"?

When you in-toto verify something, you use a previously validated
root-of-trust to check a (potentially updated) layout file and validate
link metadata based upon the keys contained in this.  Assuming your
root-of-trust is valid, the verification process provides you cryptographic
validation that the parties with the specified keys signed link metadata
that attested to only permitted operations.

Also, FYI: this sort of validation is used in the real world now by some
organizations (e.g.,
https://www.datadoghq.com/blog/engineering/secure-publication-of-datadog-agent-integrations-with-tuf-and-in-toto/
).

jku:
> Can you say more about what you mean by "practically useful"?

What I mean is that 
1. there is a plan for how this verification really happens -- so we don't just enable verification in theory, we ensure it happens or at least make it likely to happen. Do we have verification automation somewhere? If we don't, will someone else have that automation? clear instructions for contributors and users to verify?
2. the fairly long list of verification todo items in RELEASE_with_in-toto.md  is somehow handled -- I don't necessarily mean everything needs to be polished but I'd like to see at least a practical plan for these items

If we merge like this, to me it looks like verification still requires a considerable amount of effort for an in-toto specialist, and is very difficult for a random developer.

trishankatdatadog:
> I we merge like this, to me it looks like verification still requires a considerable amount of effort for an in-toto specialist, and is very difficult for a random developer.

Good questions, and this is precisely the kind of thing we should at least de facto standardise with, say, PEP 480. Let's actually take something like the Datadog work I did above (which partly inspired the upcoming Scudo, a variant of Uptane with in-toto attestations), and try it with a prototype of PEP 480.

ofek:
Hey just now seeing this. Yeah a build hook would work. For example, Project Jupyter & IPython are almost done [switching](https://github.com/jupyterlab/team-compass/issues/146) their entire packaging ecosystem to be based on a Hatch build hook plugin. I tried to link to most of the changes here: https://hatch.pypa.io/latest/users/

One thing I added with all of you in mind was the [`extra-metadata`](https://hatch.pypa.io/1.2/plugins/builder/#options) wheel option. Basically a dedicated `*.dist-info/extra_metadata/` directory to stuff whatever users might want, like signatures 😄

trishankatdatadog:
> One thing I added with all of you in mind was the [`extra-metadata`](https://hatch.pypa.io/1.2/plugins/builder/#options) wheel option. Basically a dedicated `*.dist-info/extra_metadata/` directory to stuff whatever users might want, like signatures 😄

So we can validate and verify in-toto links using layout keys distributed out of band for now 🙂

lukpueh:
Converted this back to draft to be used as reference for other release signing solutions discussed in #1966, most notably wrt to UX and security properties.

python-tuf maintainers currently don't feel confident to merge because of the perceived potential for human error in the setup and the complexity of validating its correctness.

jku:
for reference, we now get a minimal build attestation for free (example is tuf-on-ci but python-tuf will have the same as soon as we do next release):

```
$ pypi-attestations verify pypi --repository https://github.com/theupdateframework/tuf-on-ci \
     https://files.pythonhosted.org/packages/5f/0f/c922408cb12ffde512458906804f9800b55e09a9e27e32f911672cf93002/tuf_on_ci_sign-0.16.0-py3-none-any.whl
OK: tuf_on_ci_sign-0.16.0-py3-none-any.whl
$ 
```

PR #2001, opened 2022-05-09
chore: update the workflow responsible for notifying of new TUF spec release

rdimitrov:

The following PR deprecates the use of the existing workflow for notifying of new TUF releases and moves to one that is hosted in the https://github.com/theupdateframework/specification/ repository.

**Note:** Dependent on - https://github.com/theupdateframework/specification/pull/224

requirements**:


.github/workflows/specification-version-check.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: |
+          python3 -m pip install -e .
+          script="from tuf.api.metadata import SPECIFICATION_VERSION; \
+                  print(f\"v{'.'.join(SPECIFICATION_VERSION)}\")"
+          ver=$(python3 -c "$script")
+          echo "::set-output name=version::$ver"
+  # Get the latest TUF specification release and open an issue (if needed)
+  specification-bump-check:
+    needs: get-supported-tuf-version
+    uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
```
jku, line 31: this does bring to surface a small issue with this approach: We do not want `@master` in our workflows... but on the other hand, getting dependabot updates to workflows on every spec commit sounds bad (I assume that's how it's going to work).

Hmm...

rdimitrov: It's a possible drawback, indeed. Unfortunately, I think the only way to change that is to bundle it as a Github Action and create a separate repository just for it.

rdimitrov: In that sense, I'll update the PR and add permissions to the job calling the workflow in order to protect the project from any vulnerabilities coming from the `specification` repo 👍

.github/workflows/specification-version-check.yml
```diff
@@ -0,0 +1,30 @@
...
+name: Specification version check
+jobs:
+  # Get the version of the TUF specification the project states it supports
+  get-supported-tuf-version:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+      - uses: actions/setup-python@98f2ad02fd48d057ee3b4d4f66525b231c3e52b6
```
lukpueh: Quoting @jku from https://github.com/theupdateframework/python-tuf/issues/2089#issuecomment-1230261767

> We should definitely set python-version to "3.x" in this workflow and everywhere else where we don't use a specific version.

lukpueh:
@rdimitrov: This is a friendly takeover of your PR to
1. rebase and solve conflicts
2. update action/setup-python to latest version
3. pin major version to be used to 3.x
4. upgrade pip before using it 

(3) and (4) were suggested by @jku in #2089

Given that the upstream dependent (https://github.com/theupdateframework/specification/pull/224) has been merged and permissions concerns addressed (https://github.com/theupdateframework/python-tuf/pull/2001#discussion_r868881109), I suggest this PR is good to go.

This PR also fixes parts of #2089.

jku (approved):
Looks good I suppose? Using master as the version still does not set a great example but I agree it seems safe.

Thanks!

jku:
oops, pip now complains: `no such option: -u`: I left #2089 open.

PR #2002, opened 2022-05-09
Appending Fetcher docs to state the method to be implemented.

### LINKED ISSUE #1916: Fetcher docs should make it clear what needs to be implemented
Should add a sentence in FetcherInterface docstring that explains that the public API of the interface is implemented already: implementations of the interface only need to implement the internal

    def _fetch(self, url: str) -> Iterator[bytes]

abs007:
Fetcher.py: Appending class docstring

Fixes #1916

The class docstring for `FetcherInterface` needed to clearly state that
only `_fetch()` had to be implemented in it's implementation. This is
because the public API of the interface is implemented already.

### DIFF
--- a/tuf/ngclient/fetcher.py
+++ b/tuf/ngclient/fetcher.py
@@ -23,6 +23,9 @@ class FetcherInterface:
     By providing a concrete implementation of the abstract interface,
     users of the framework can plug-in their preferred/customized
     network stack.
+
+    Implementations of FetcherInterface only need to implement ``_fetch()``.
+    The public API of the class is already implemented.
     """
 
     __metaclass__ = abc.ABCMeta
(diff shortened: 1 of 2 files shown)


tuf/ngclient/fetcher.py
```diff
@@ -23,6 +23,9 @@ class FetcherInterface:
     By providing a concrete implementation of the abstract interface,
     users of the framework can plug-in their preferred/customized
     network stack.
+
+    Implementaions of FetcherInterface only need to implement ``_fetch()``.
+    The public API of the class is already implemented.
```
MVrachev:
```suggestion
    Implementations of FetcherInterface only need to implement ``_fetch()``.
    The public API of the class is already implemented.
```

MVrachev (commented):
One typo, but otherwise looks good.

abs007:
Seemed like the 'Commit Suggestion' option left a signing error. Hence had to manually force push the change for fixing the typo.

jku (approved):
Nice, thanks!

PR #2003, opened 2022-05-09
fix:  ngclient: temp_file could be undefined #1999

### LINKED ISSUE #1999: ngclient: temp_file could be undefined
From sonatype lift results: 

> Local variable temp_file is undefined, or not always defined.
> tuf/ngclient/updater.py 290:15

This seems like it could happen if NamedTemporaryFile() fails to create a file.

dhavalgshah:
Fixes ngclient: temp_file could be undefined #1999

Signed-off-by: Dhaval Shah <30974879+dhavalgshah@users.noreply.github.com>

Fixes #1999

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -37,7 +37,7 @@ import logging
 import os
 import shutil
 import tempfile
-from typing import Optional, Set
+from typing import Any, Optional, Set
 from urllib import parse
 
 from tuf.api import exceptions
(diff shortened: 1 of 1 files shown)


dhavalgshah:
Logs are empty for the failed CI run and tox is running fine on my machine as well as on other CI runs.

Is there any hint on what might have gone wrong here? @jku

lukpueh:
If you scroll through the list of checks below, you'll find the job that failed. Its log tells you what went wrong.

jku:
Unfortunately a lot of the python base library  predates type annotations: you likely can't annotate temp_file in a useful way.

jku (approved):
It's annoying that we have to make code like this more complicated... but I can't see a way to avoid it either.

Thanks, looks good to me.

PR #2010, opened 2022-05-20
Add support for succinct roles (TAP 15)

MVrachev:
Related to the previous pr: https://github.com/theupdateframework/python-tuf/pull/1948 and issue #1909

**This pr includes 11 breaking changes:**

In `Delegations` class from `tuf/api/metadata.py` :
1. roles argument is made optional
2. unrecognized_fields argument becomes the 4-th rather than the 3-rd
as it used to be

In `Root` class from `tuf/api/metadata.py`:
1. The "role" and "key" arguments in "Root.add_key()" are in reverse
order - "key" becomes first and "role" second.
2. "Root.remove_key()" has been renamed to "Root.remove_key()"
3. The "role" and "keyid" arguments in "Root.revoke_key()" are in
reverse order - "keyid" becomes first and "role" second.

In `Targets` class from `tuf/api/metadata.py`:
1. The "role" and "key" arguments in "Targets.add_key()" are in reverse
order - "key" becomes first and "role" second
4. "Targets.remove_key()" has been renamed to "Targets.revoke_key()"
5. The "role" and "keyid" arguments in "Targets.revoke_key()" are in
reverse order - "keyid" becomes first and "role" second.
6.  In both methods "Targets.add_key()" and "Targets.revoke_key()" the
"role" argument becomes an optional with a default value of None

Those changes are made in an effort to make those two methods logical
for both cases when standard roles and succinct_roles are used.

[TAP 15](https://github.com/theupdateframework/taps/blob/master/tap15.md) was created on June 23-rd 2020 and was last modified on July 6-th 2020.
Since then the TAP has been put to `draft` status meaning it needs a prototype implementation before it's accepted as a future specification change.
Given that this TAP underlines an efficient way of handling succinct hash bin delegations, meaning it would be really useful when `python-tuf` is integrated into `Warehouse`, it's logical that we should not only create a prototype but directly work to integrate it in `python-tuf`.

The outcome of the TAP 15 implementation is
1. Support reading, updating, and saving target metadata files that are using `succinct_roles`.
3. Support in ngclient for calculating the delegated role name on a given target path.
4. An exemplary document inside `examples/` showcasing how `succinct_roles` can be utilized in practice.

This pr covers points 1 and 2.
Point 3 will be done in another pr.

requirements**:

### DIFF
--- a/examples/repo_example/hashed_bin_delegation.py
+++ b/examples/repo_example/hashed_bin_delegation.py
@@ -162,6 +162,7 @@ roles["bins"].signed.delegations = Delegations(
 #              10-17                       10 11 12 13 14 15 16 17
 #              ...                         ...
 #              f8-ff                       f8 f9 fa fb fc fd fe ff
+assert roles["bins"].signed.delegations.roles is not None
 for bin_n_name, bin_n_hash_prefixes in generate_hash_bins():
     # Update delegating targets role (bins) with delegation details for each
     # delegated targets role (bin_n).
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -354,9 +354,17 @@ class RepositorySimulator(FetcherInterface):
         else:
             delegator = self.md_delegates[delegator_name].signed
 
+        if (
+            delegator.delegations is not None
+            and delegator.delegations.succinct_roles is not None
+        ):
+            raise ValueError("Can't add a role when succinct_roles is used")
+
         # Create delegation
         if delegator.delegations is None:
-            delegator.delegations = Delegations({}, {})
+            delegator.delegations = Delegations({}, roles={})
+
+        assert delegator.delegations.roles is not None
         # put delegation last by default
         delegator.delegations.roles[role.name] = role
 
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -169,7 +169,7 @@ class RepositorySimulator(FetcherInterface):
         self.signers[role].clear()
         for _ in range(0, self.root.roles[role].threshold):
             key, signer = self.create_key()
-            self.root.add_key(role, key)
+            self.root.add_key(key, role)
             self.add_signer(role, signer)
 
     def _initialize(self) -> None:
@@ -182,7 +182,7 @@ class RepositorySimulator(FetcherInterface):
 
         for role in TOP_LEVEL_ROLE_NAMES:
             key, signer = self.create_key()
-            self.md_root.signed.add_key(role, key)
+            self.md_root.signed.add_key(key, role)
             self.add_signer(role, signer)
 
         self.publish_root()
@@ -370,7 +370,7 @@ class RepositorySimulator(FetcherInterface):
 
         # By default add one new key for the role
         key, signer = self.create_key()
-        delegator.add_key(role.name, key)
+        delegator.add_key(key, role.name)
         self.add_signer(role.name, signer)
 
         # Add metadata for the role
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -67,6 +67,7 @@ from tuf.api.metadata import (
     MetaFile,
     Root,
     Snapshot,
+    SuccinctRoles,
     TargetFile,
     Targets,
     Timestamp,
@@ -334,12 +335,16 @@ class RepositorySimulator(FetcherInterface):
         self.snapshot.version += 1
         self.update_timestamp()
 
+    def _get_delegator(self, delegator_name: str) -> Targets:
+        """Given a delegator name return, its corresponding Targets object."""
+        if delegator_name == Targets.type:
+            return self.targets
+
+        return self.md_delegates[delegator_name].signed
+
     def add_target(self, role: str, data: bytes, path: str) -> None:
         """Create a target from data and add it to the target_files."""
-        if role == Targets.type:
-            targets = self.targets
-        else:
-            targets = self.md_delegates[role].signed
+        targets = self._get_delegator(role)
 
         target = TargetFile.from_data(path, data, ["sha256"])
         targets.targets[path] = target
@@ -349,10 +354,7 @@ class RepositorySimulator(FetcherInterface):
         self, delegator_name: str, role: DelegatedRole, targets: Targets
     ) -> None:
         """Add delegated target role to the repository."""
-        if delegator_name == Targets.type:
-            delegator = self.targets
-        else:
-            delegator = self.md_delegates[delegator_name].signed
+        delegator = self._get_delegator(delegator_name)
 
         if (
             delegator.delegations is not None
@@ -376,6 +378,38 @@ class RepositorySimulator(FetcherInterface):
         # Add metadata for the role
         if role.name not in self.md_delegates:
             self.md_delegates[role.name] = Metadata(targets, {})
+
+    def add_succinct_roles(
+        self, delegator_name: str, bit_length: int, name_prefix: str
+    ) -> None:
+        """Add succinct roles info to a delegator with name "delegator_name".
+
+        Note that for each delegated role represented by succinct roles an empty
+        Targets instance is created.
+        """
+        delegator = self._get_delegator(delegator_name)
+
+        if (
+            delegator.delegations is not None
+            and delegator.delegations.roles is not None
+        ):
+            raise ValueError(
+                "Can't add a succinct_roles when delegated roles are used"
+            )
+
+        key, signer = self.create_key()
+        succinct_roles = SuccinctRoles([], 1, bit_length, name_prefix)
+        delegator.delegations = Delegations({}, None, succinct_roles)
+
+        # Add targets metadata for all bins.
+        for delegated_name in succinct_roles.get_roles():
+            self.md_delegates[delegated_name] = Metadata(
+                Targets(expires=self.safe_expiry)
+            )
+
+            self.add_signer(delegated_name, signer)
+
+        delegator.add_key(key)
 
     def write(self) -> None:
         """Dump current repository metadata to self.dump_dir
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -35,6 +35,7 @@ from tuf.api.metadata import (
     Metadata,
     Root,
     Snapshot,
+    SuccinctRoles,
     TargetFile,
     Targets,
     Timestamp,
@@ -694,6 +695,31 @@ class TestMetadata(unittest.TestCase):
             role = DelegatedRole("", [], 1, False, None, hash_prefixes)
             self.assertFalse(role.is_delegated_path("a/non-matching path"))
             self.assertTrue(role.is_delegated_path("a/path"))
+
+    def test_is_delegated_role_in_succinct_roles(self) -> None:
+        succinct_roles = SuccinctRoles([], 1, 5, "bin")
+        false_role_name_examples = ["foo", "bin-", "bin-s", "bin-20", "bin-100"]
+        for role_name in false_role_name_examples:
+            msg = f"Error for {role_name}"
+            self.assertFalse(succinct_roles.is_delegated_role(role_name), msg)
+
+        # delegated role name suffixes are in hex format.
+        true_name_examples = ["bin-00", "bin-0f", "bin-1f"]
+        for role_name in true_name_examples:
+            msg = f"Error for {role_name}"
+            self.assertTrue(succinct_roles.is_delegated_role(role_name), msg)
+
+    def test_get_roles_in_succinct_roles(self) -> None:
+        succinct_roles = SuccinctRoles([], 1, 16, "bin")
+        # bin names are in hex format and 4 hex digits are enough to represent
+        # all bins between 0 and 2^16 - 1 meaning suffix_len must be 4
+        expected_suffix_length = 4
+        self.assertEqual(succinct_roles.suffix_len, expected_suffix_length)
+        for bin_numer, role_name in enumerate(succinct_roles.get_roles()):
+            # This adds zero-padding if the bin_numer is represented by a hex
+            # number with a length less than expected_suffix_length.
+            expected_bin_suffix = f"{bin_numer:0{expected_suffix_length}x}"
+            self.assertEqual(role_name, f"bin-{expected_bin_suffix}")
 
 
 # Run unit test.
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -361,7 +361,7 @@ class TestMetadata(unittest.TestCase):
 
         # Add a key to snapshot role, make sure the new sig fails to verify
         ts_keyid = next(iter(root.signed.roles[Timestamp.type].keyids))
-        root.signed.add_key(Snapshot.type, root.signed.keys[ts_keyid])
+        root.signed.add_key(root.signed.keys[ts_keyid], Snapshot.type)
         snapshot.signatures[ts_keyid] = Signature(ts_keyid, "ff" * 64)
 
         # verify succeeds if threshold is reached even if some signatures
@@ -390,7 +390,7 @@ class TestMetadata(unittest.TestCase):
         with self.assertRaises(ValueError):
             Key.from_securesystemslib_key(sslib_key)
 
-    def test_root_add_key_and_remove_key(self) -> None:
+    def test_root_add_key_and_revoke_key(self) -> None:
         root_path = os.path.join(self.repo_dir, "metadata", "root.json")
         root = Metadata[Root].from_file(root_path)
 
@@ -410,8 +410,12 @@ class TestMetadata(unittest.TestCase):
         self.assertNotIn(keyid, root.signed.roles[Root.type].keyids)
         self.assertNotIn(keyid, root.signed.keys)
 
+        # Assert that add_key with old argument order will raise an error
+        with self.assertRaises(ValueError):
+            root.signed.add_key(Root.type, key_metadata)  # type: ignore
+
         # Add new root key
-        root.signed.add_key(Root.type, key_metadata)
+        root.signed.add_key(key_metadata, Root.type)
 
         # Assert that key is added
         self.assertIn(keyid, root.signed.roles[Root.type].keyids)
@@ -423,30 +427,30 @@ class TestMetadata(unittest.TestCase):
 
         # Try adding the same key again and assert its ignored.
         pre_add_keyid = root.signed.roles[Root.type].keyids.copy()
-        root.signed.add_key(Root.type, key_metadata)
+        root.signed.add_key(key_metadata, Root.type)
         self.assertEqual(pre_add_keyid, root.signed.roles[Root.type].keyids)
 
         # Add the same key to targets role as well
-        root.signed.add_key(Targets.type, key_metadata)
+        root.signed.add_key(key_metadata, Targets.type)
 
         # Add the same key to a nonexistent role.
         with self.assertRaises(ValueError):
-            root.signed.add_key("nosuchrole", key_metadata)
+            root.signed.add_key(key_metadata, "nosuchrole")
 
         # Remove the key from root role (targets role still uses it)
-        root.signed.remove_key(Root.type, keyid)
+        root.signed.revoke_key(keyid, Root.type)
         self.assertNotIn(keyid, root.signed.roles[Root.type].keyids)
         self.assertIn(keyid, root.signed.keys)
 
         # Remove the key from targets as well
-        root.signed.remove_key(Targets.type, keyid)
+        root.signed.revoke_key(keyid, Targets.type)
         self.assertNotIn(keyid, root.signed.roles[Targets.type].keyids)
         self.assertNotIn(keyid, root.signed.keys)
 
         with self.assertRaises(ValueError):
-            root.signed.remove_key(Root.type, "nosuchkey")
+            root.signed.revoke_key("nosuchkey", Root.type)
         with self.assertRaises(ValueError):
-            root.signed.remove_key("nosuchrole", keyid)
+            root.signed.revoke_key(keyid, "nosuchrole")
 
     def test_is_target_in_pathpattern(self) -> None:
         # pylint: disable=protected-access
@@ -507,9 +511,13 @@ class TestMetadata(unittest.TestCase):
         }
         key = Key.from_dict("id2", key_dict)
 
+        # Assert that add_key with old argument order will raise an error
+        with self.assertRaises(ValueError):
+            targets.add_key("role1", key)  # type: ignore
+
         # Assert that delegated role "role1" does not contain the new key
         self.assertNotIn(key.keyid, targets.delegations.roles["role1"].keyids)
-        targets.add_key("role1", key)
+        targets.add_key(key, "role1")
 
         # Assert that the new key is added to the delegated role "role1"
         self.assertIn(key.keyid, targets.delegations.roles["role1"].keyids)
@@ -519,45 +527,88 @@ class TestMetadata(unittest.TestCase):
 
         # Try adding the same key again and assert its ignored.
         past_keyid = targets.delegations.roles["role1"].keyids.copy()
-        targets.add_key("role1", key)
+        targets.add_key(key, "role1")
         self.assertEqual(past_keyid, targets.delegations.roles["role1"].keyids)
 
         # Try adding a key to a delegated role that doesn't exists
         with self.assertRaises(ValueError):
-            targets.add_key("nosuchrole", key)
+            targets.add_key(key, "nosuchrole")
 
         # Add the same key to "role2" as well
-        targets.add_key("role2", key)
+        targets.add_key(key, "role2")
 
         # Remove the key from "role1" role ("role2" still uses it)
-        targets.remove_key("role1", key.keyid)
+        targets.revoke_key(key.keyid, "role1")
 
         # Assert that delegated role "role1" doesn't contain the key.
         self.assertNotIn(key.keyid, targets.delegations.roles["role1"].keyids)
         self.assertIn(key.keyid, targets.delegations.roles["role2"].keyids)
 
         # Remove the key from "role2" as well
-        targets.remove_key("role2", key.keyid)
+        targets.revoke_key(key.keyid, "role2")
         self.assertNotIn(key.keyid, targets.delegations.roles["role2"].keyids)
 
         # Try remove key not used by "role1"
         with self.assertRaises(ValueError):
-            targets.remove_key("role1", key.keyid)
+            targets.revoke_key(key.keyid, "role1")
 
         # Try removing a key from delegated role that doesn't exists
         with self.assertRaises(ValueError):
-            targets.remove_key("nosuchrole", key.keyid)
+            targets.revoke_key(key.keyid, "nosuchrole")
 
         # Remove delegations as a whole
         targets.delegations = None
-        # Test that calling add_key and remove_key throws an error
+        # Test that calling add_key and revoke_key throws an error
         # and that delegations is still None after each of the api calls
         with self.assertRaises(ValueError):
-            targets.add_key("role1", key)
+            targets.add_key(key, "role1")
         self.assertTrue(targets.delegations is None)
         with self.assertRaises(ValueError):
-            targets.remove_key("role1", key.keyid)
+            targets.revoke_key(key.keyid, "role1")
         self.assertTrue(targets.delegations is None)
+
+    def test_targets_key_api_with_succinct_roles(self) -> None:
+        targets_path = os.path.join(self.repo_dir, "metadata", "targets.json")
+        targets: Targets = Metadata[Targets].from_file(targets_path).signed
+        key_dict = {
+            "keytype": "ed25519",
+            "keyval": {
+                "public": "edcd0a32a07dce33f7c7873aaffbff36d20ea30787574ead335eefd337e4dacd"
+            },
+            "scheme": "ed25519",
+        }
+        key = Key.from_dict("id2", key_dict)
+
+        # Remove delegated roles.
+        assert targets.delegations is not None
+        assert targets.delegations.roles is not None
+        targets.delegations.roles = None
+        targets.delegations.keys = {}
+
+        # Add succinct_roles information.
+        targets.delegations.succinct_roles = SuccinctRoles([], 1, 8, "foo")
+        self.assertEqual(len(targets.delegations.keys), 0)
+        self.assertEqual(len(targets.delegations.succinct_roles.keyids), 0)
+
+        # Add a key to succinct_roles and verify it's saved.
+        targets.add_key(key)
+        self.assertIn(key.keyid, targets.delegations.keys)
+        self.assertIn(key.keyid, targets.delegations.succinct_roles.keyids)
+        self.assertEqual(len(targets.delegations.keys), 1)
+
+        # Try adding the same key again and verify that noting is added.
+        targets.add_key(key)
+        self.assertEqual(len(targets.delegations.keys), 1)
+
+        # Remove the key and verify it's not stored anymore.
+        targets.revoke_key(key.keyid)
+        self.assertNotIn(key.keyid, targets.delegations.keys)
+        self.assertNotIn(key.keyid, targets.delegations.succinct_roles.keyids)
+        self.assertEqual(len(targets.delegations.keys), 0)
+
+        # Try removing it again.
+        with self.assertRaises(ValueError):
+            targets.revoke_key(key.keyid)
 
     def test_length_and_hash_validation(self) -> None:
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -495,6 +495,7 @@ class TestMetadata(unittest.TestCase):
             }
         )
         assert isinstance(targets.delegations, Delegations)
+        assert isinstance(targets.delegations.roles, Dict)
         targets.delegations.roles["role2"] = delegated_role
 
         key_dict = {
--- a/tests/test_metadata_eq_.py
+++ b/tests/test_metadata_eq_.py
@@ -23,6 +23,7 @@ from tuf.api.metadata import (
     Metadata,
     MetaFile,
     Role,
+    SuccinctRoles,
     TargetFile,
 )
 
@@ -55,6 +56,7 @@ class TestMetadataComparisions(unittest.TestCase):
         cls.objects["Role"] = Role(["keyid1", "keyid2"], 3)
         cls.objects["MetaFile"] = MetaFile(1, 12, {"sha256": "abc"})
         cls.objects["DelegatedRole"] = DelegatedRole("a", [], 1, False, ["d"])
+        cls.objects["SuccinctRoles"] = SuccinctRoles(["keyid"], 1, 8, "foo")
         cls.objects["Delegations"] = Delegations(
             {"keyid": cls.objects["Key"]}, {"a": cls.objects["DelegatedRole"]}
         )
@@ -79,6 +81,7 @@ class TestMetadataComparisions(unittest.TestCase):
             "paths": [""],
             "path_hash_prefixes": [""],
         },
+        "SuccinctRoles": {"bit_length": 0, "name_prefix": ""},
         "Delegations": {"keys": {}, "roles": {}},
         "TargetFile": {"length": 0, "hashes": {}, "path": ""},
         "Targets": {"targets": {}, "delegations": []},
--- a/tests/test_metadata_eq_.py
+++ b/tests/test_metadata_eq_.py
@@ -169,6 +169,7 @@ class TestMetadataComparisions(unittest.TestCase):
         # Create a second delegations obj with reversed roles order
         delegations_2 = copy.deepcopy(delegations)
         # In python3.7 we need to cast to a list and then reverse.
+        assert isinstance(delegations.roles, dict)
         delegations_2.roles = dict(reversed(list(delegations.roles.items())))
 
         # Both objects are not the equal because of delegated roles order.
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -491,6 +491,13 @@ class TestSerialization(unittest.TestCase):
                 {"keyids": ["keyid1"], "name": "b", "terminating": true, "paths": ["fn1"], "threshold": 3}, \
                 {"keyids": ["keyid2"], "name": "root", "terminating": true, "paths": ["fn2"], "threshold": 4} ] \
             }',
+        "roles and succinct_roles set": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
+                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
+            "roles": [ \
+                {"keyids": ["keyid"], "name": "a", "terminating": true, "paths": ["fn1"], "threshold": 3}, \
+                {"keyids": ["keyid2"], "name": "b", "terminating": true, "paths": ["fn2"], "threshold": 4} ], \
+            "succinct_roles": {"keyids": ["keyid"], "threshold": 1, "bit_length": 8, "name_prefix": "foo"}}',
     }
 
     @utils.run_sub_tests_with_dataset(invalid_delegations)
@@ -502,13 +509,17 @@ class TestSerialization(unittest.TestCase):
             Delegations.from_dict(case_dict)
 
     valid_delegations: utils.DataSet = {
-        "all": '{"keys": { \
+        "with roles": '{"keys": { \
                 "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
                 "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
             "roles": [ \
                 {"keyids": ["keyid"], "name": "a", "terminating": true, "paths": ["fn1"], "threshold": 3}, \
                 {"keyids": ["keyid2"], "name": "b", "terminating": true, "paths": ["fn2"], "threshold": 4} ] \
             }',
+        "with succinct_roles": '{"keys": { \
+                "keyid1" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}, \
+                "keyid2" : {"keytype": "ed25519", "scheme": "ed25519", "keyval": {"public": "bar"}}}, \
+            "succinct_roles": {"keyids": ["keyid"], "threshold": 1, "bit_length": 8, "name_prefix": "foo"}}',
         "unrecognized field": '{"keys": {"keyid" : {"keytype": "rsa", "scheme": "rsassa-pss-sha256", "keyval": {"public": "foo"}}}, \
             "roles": [ {"keyids": ["keyid"], "name": "a", "paths": ["fn1", "fn2"], "terminating": true, "threshold": 3} ], \
             "foo": "bar"}',
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -24,6 +24,7 @@ from tuf.api.metadata import (
     Role,
     Root,
     Snapshot,
+    SuccinctRoles,
     TargetFile,
     Targets,
     Timestamp,
@@ -407,6 +408,36 @@ class TestSerialization(unittest.TestCase):
         case_dict = json.loads(test_case_data)
         with self.assertRaises(ValueError):
             DelegatedRole.from_dict(case_dict)
+
+    valid_succinct_roles: utils.DataSet = {
+        # SuccinctRoles inherits Role and some use cases can be found in the valid_roles.
+        "standard succinct_roles information": '{"keyids": ["keyid"], "threshold": 1, \
+            "bit_length": 8, "name_prefix": "foo"}',
+        "succinct_roles with unrecognized fields": '{"keyids": ["keyid"], "threshold": 1, \
+            "bit_length": 8, "name_prefix": "foo", "foo": "bar"}',
+    }
+
+    @utils.run_sub_tests_with_dataset(valid_succinct_roles)
+    def test_succinct_roles_serialization(self, test_case_data: str) -> None:
+        case_dict = json.loads(test_case_data)
+        succinct_roles = SuccinctRoles.from_dict(copy.copy(case_dict))
+        self.assertDictEqual(case_dict, succinct_roles.to_dict())
+
+    invalid_succinct_roles: utils.DataSet = {
+        # SuccinctRoles inherits Role and some use cases can be found in the invalid_roles.
+        "missing bit_length from succinct_roles": '{"keyids": ["keyid"], "threshold": 1, "name_prefix": "foo"}',
+        "missing name_prefix from succinct_roles": '{"keyids": ["keyid"], "threshold": 1, "bit_length": 8}',
+        "succinct_roles with invalid bit_length type": '{"keyids": ["keyid"], "threshold": 1, "bit_length": "a", "name_prefix": "foo"}',
+        "succinct_roles with invalid name_prefix type": '{"keyids": ["keyid"], "threshold": 1, "bit_length": 8, "name_prefix": 1}',
+        "succinct_roles with high bit_length value": '{"keyids": ["keyid"], "threshold": 1, "bit_length": 50, "name_prefix": "foo"}',
+        "succinct_roles with low bit_length value": '{"keyids": ["keyid"], "threshold": 1, "bit_length": 0, "name_prefix": "foo"}',
+    }
+
+    @utils.run_sub_tests_with_dataset(invalid_succinct_roles)
+    def test_invalid_succinct_roles_serialization(self, test_data: str) -> None:
+        case_dict = json.loads(test_data)
+        with self.assertRaises((ValueError, KeyError, TypeError)):
+            SuccinctRoles.from_dict(case_dict)
 
     invalid_delegations: utils.DataSet = {
         "empty delegations": "{}",
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -176,6 +176,7 @@ class TestSerialization(unittest.TestCase):
         "no threshold": '{"keyids": ["keyid"]}',
         "no keyids": '{"threshold": 3}',
         "wrong threshold type": '{"keyids": ["keyid"], "threshold": "a"}',
+        "wrong keyids type": '{"keyids": 1, "threshold": 3}',
         "threshold below 1": '{"keyids": ["keyid"], "threshold": 0}',
         "duplicate keyids": '{"keyids": ["keyid", "keyid"], "threshold": 3}',
     }
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -425,6 +425,91 @@ class TestDelegationsGraphs(TestDelegations):
         finally:
             self.teardown_subtest()
 
+    @dataclass
+    class SuccinctRolesTestCase:
+        bit_length: int
+        target_path: str
+        expected_target_bin: str
+
+    # By setting the bit_length the total number of bins is 2^bit_length.
+    # In each test case target_path is a path to a random target we want to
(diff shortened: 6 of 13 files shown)


MVrachev:
I know the code is a lot, but a big chunk of it is testing.

tuf/api/metadata.py
```diff
@@ -1435,6 +1444,149 @@ def is_delegated_path(self, target_filepath: str) -> bool:
...
+                if (
+                    0 <= num < self.number_of_bins
+                    and len(suffix) == self.suffix_len
+                ):
+                    return True
+            except ValueError:
+                # suffix is not a number
+                return False
+
+        return False
```
jku: this would likely improve from early returns, something like this:

```suggestion
        if not role_name.startswith(desired_prefix):
            return False

        suffix = role_name[len(desired_prefix) :]
        if len(suffix) != self.suffix_len:
            return False

        try:
            # make sure suffix is hex value
            num = int(suffix, 16)
        except ValueError:
            return False

        return num >= 0 and num < self.number_of_bins
```

tuf/api/metadata.py
```diff
@@ -1457,18 +1611,24 @@ class Delegations:
...
     ):
         self.keys = keys
-
-        for role in roles:
-            if not role or role in TOP_LEVEL_ROLE_NAMES:
-                raise ValueError(
-                    "Delegated roles cannot be empty or use top-level role names"
-                )
+        if sum(1 for v in [roles, succint_roles] if v is not None) != 1:
+            raise ValueError("Only one of roles and succinct_roles must be set")
```
lukpueh: This error is also raised when neither are set. Then the error message might be surprising.

tuf/api/metadata.py
```diff
@@ -1643,25 +1665,78 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
...
+        self, target_filepath: str
+    ) -> Iterator[Tuple[str, bool]]:
+        """Given ``target_filepath`` get names and terminating status of all
+        delegations who are responsible for it.
+
+        Args:
+            target_filepath: URL path to a target file, relative to a base
+                targets URL.
+        """
+        if isinstance(self.roles, dict):
```
jku: There are multiple isinstance() calls that look a little fishy to me -- they imply there are cases where `self.roles` type is something other than Optional[Dict] which is not true. I think`self.roles is not None` is clearer and also matches the elif a few lines after it.

this comment likely applies most isinstance() calls in the actual code -- tests can do whatever they want.

tuf/api/metadata.py
```diff
@@ -1643,25 +1665,78 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
...
             **self.unrecognized_fields,
         }
+        if isinstance(self.roles, dict):
+            roles = [role_obj.to_dict() for role_obj in self.roles.values()]
+            res_dict["roles"] = roles
+        elif isinstance(self.succinct_roles, SuccinctRoles):
+            res_dict["succinct_roles"] = self.succinct_roles.to_dict()
+        return res_dict
+
+    def find_delegation(self, target_filepath: str) -> Optional[str]:
```
jku: is find_delegation() needed somewhere?

tuf/api/metadata.py
```diff
@@ -1603,18 +1611,24 @@ class Delegations:
     def __init__(
         self,
         keys: Dict[str, Key],
-        roles: Dict[str, DelegatedRole],
+        roles: Optional[Dict[str, DelegatedRole]] = None,
+        succint_roles: Optional[SuccinctRoles] = None,
```
jku: typo succinct

tuf/api/metadata.py
```diff
@@ -1603,18 +1611,24 @@ class Delegations:
...
         unrecognized_fields: Optional[Dict[str, Any]] = None,
```
jku: This is an API change -- it's likely an API change we want to do but this sort of things really should be pointed out in the commit message and PR cover letters.

The two changes here are moving unrecognised fields in constructor argument list and more importantly making the roles attribute optional -- this is mostly relevant to code that uses `Delegations.roles` as it now has to check for None (As far as I can tell you've done a great job with handling all of that code within our codebase but any tuf using code, e.g. repository-editor-for-tuf, now needs to add None-checks)

tuf/api/metadata.py
```diff
@@ -1593,6 +1599,8 @@ class Delegations:
             defines which keys are required to sign the metadata for a specific
             role. The roles order also defines the order that role delegations
             are considered during target searches.
+        succinct_roles: Contains succinct information about hash bin
+            delegations. For more information see: https://github.com/theupdateframework/taps/blob/master/tap15.md
```
jku: the fact that exactly one of [roles, succinct_roles] is not-None should be documented

tuf/api/metadata.py
```diff
@@ -1497,25 +1665,78 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
...
             **self.unrecognized_fields,
         }
+        if isinstance(self.roles, dict):
+            roles = [role_obj.to_dict() for role_obj in self.roles.values()]
+            res_dict["roles"] = roles
+        elif isinstance(self.succinct_roles, SuccinctRoles):
+            res_dict["succinct_roles"] = self.succinct_roles.to_dict()
+        return res_dict
+
+    def find_delegation(self, target_filepath: str) -> Optional[str]:
```
lukpueh: Can we remove this API function, given that it's nowhere used?

tuf/api/metadata.py
```diff
@@ -2012,6 +2012,31 @@ def add_key(self, role: str, key: Key) -> None:
 
         self.delegations.keys[key.keyid] = key
 
+    def add_key_in_succinct_roles(self, key: Key) -> None:
```
jku: I'm not a big fan of this one: it adds 50 lines of code (where 2 lines is new functionality) and two new public methods ... and reality is that the old methods could work just fine, the API just looks a bit odd with the unused rolename.

The PR as a whole is already an API change so let's consider modifying the API here as well, and let's also consider the option of just documenting that rolename can be an unused argument when calling `add_key()`.

jku: You say in the comment about this 

> clarify what the role name for succinct_roles should be (but this
> should happen in the TAP 15 itself and will lead to unnecessary
> complications)

I don't agree. `add_key()` is our API and unrelated to the TAP, and the problem is completely in our API decisions. We can choose how it works.

tests/repository_simulator.py
```diff
@@ -377,6 +378,44 @@ def add_delegation(
...
+        key, signer = self.create_key()
+        succinct_roles = SuccinctRoles([], 1, bit_length, name_prefix)
+
+        # Create delegation
+        if delegator.delegations is None:
+            delegator.delegations = Delegations(
+                {}, roles=None, succinct_roles=succinct_roles
+            )
+        else:
+            delegator.delegations.succinct_roles = succinct_roles
```
jku: style comment: I would make this shorter and easier to understand, like:

        key, signer = self.create_key()
        succinct_roles = SuccinctRoles([], 1, bit_length, name_prefix)
        delegator.delegations = Delegations({}, None, succinct_roles)
 
 it's a testing helper class so I don't think handling every weird case is possible or that useful. But less code is useful for readers

tuf/api/metadata.py
```diff
@@ -1435,6 +1444,149 @@ def is_delegated_path(self, target_filepath: str) -> bool:
...
+    def to_dict(self) -> Dict[str, Any]:
+        """Returns the dict representation of self."""
+        base_role_dict = super().to_dict()
+        return {
+            "bit_length": self.bit_length,
+            "name_prefix": self.name_prefix,
+            **base_role_dict,
+        }
+
+    def get_bin_name(self, bin_number: int = 0) -> str:
```
jku: is there a reason for this to be a public method?

tuf/api/metadata.py
```diff
@@ -1435,6 +1444,149 @@ def is_delegated_path(self, target_filepath: str) -> bool:
         return False
 
 
+class SuccinctRoles(Role):
+    """A container with succinct information about hash bin delegation for all
+    possible target paths. It's can be used inside ``Delegations``.
```
```suggestion
    possible target paths. It can be used inside ``Delegations``.
```

lukpueh: Minor doc style nit: The summary of a class/function should fit on one line. Same comment applies elsewhere in this PR.

jku (commented):
Overall this looks like a good direction I think and code does not seem to have issues (bugs). I have left some individual code comments (no bugs or anything, just opinions). These are the concerns:
* this adds a large amount of new API, 10 new methods or something like that: let's focus on whether all of that is actually used and needed, whether it needs to be public API and whether there are other options to adding the API (ie the `add_key()` case). I've left some comments 
* this is also an API change: I think it's one we want to do but let's take changes seriously

I've not checked if the test coverage is "complete": let's figure out the API size first.

tuf/api/metadata.py
jku: also the default value seems to not be useful or used.

tuf/api/metadata.py
```diff
@@ -1435,6 +1444,149 @@ def is_delegated_path(self, target_filepath: str) -> bool:
...
+    two class attributes.
+
+    Args:
+        keyids: Signing key identifiers.
+        threshold: Number of keys required to sign succinct roles metadata.
+        bit_length: Number of bits between 1 and 32 used to separate the
+            different bins.
+        name_prefix: Prefix of all bin names.
+        unrecognized_fields: Dictionary of all attributes that are not managed
+            by TUF Metadata API.
```
lukpueh: I still think the entire class docstring requires a lot of prior knowledge about succinct hash delegation, to a point where it isn't relevant anymore. 

Would do you think about the following suggestion. It is slightly wordier, but also provides more context from TAP 15:

```suggestion
    """Succinctly defines a hash bin delegation graph.

    A ``SuccinctRoles`` object describes a delegation graph that covers all targets,
    distributing them uniformly over the delegated roles (i.e. bins) in the graph.

    The total number of bins is 2 to the power of the passed ``bit_length``. And targets
    are assigned to bins by casting the left-most ``bit_length`` bits of the file path
    hash digest to int, using it as bin index between 0 and ``2**bit_length - 1``.

    Bin names are the concatenation of the passed ``name_prefix`` and a zero-padded hex
    representation of the bin index between separated by a hyphen.

    The same passed ``keyids`` and ``threshold`` is used for each bin, and each bin is
    'terminating'.

    See https://github.com/theupdateframework/taps/blob/master/tap15.md for details.

    Args:
        keyids: Signing key identifiers for any bin metadata.
        threshold: Number of keys required to sign any bin metadata.
        bit_length: Number of bits between 1 and 32.
        name_prefix: Prefix of all bin names.
        unrecognized_fields: Dictionary of all attributes that are not managed
            by TUF Metadata API.
```

lukpueh: Feel free to ignore/edit as you please.

tuf/api/metadata.py
```diff
@@ -1435,6 +1444,149 @@ def is_delegated_path(self, target_filepath: str) -> bool:
...
+            "bit_length": self.bit_length,
+            "name_prefix": self.name_prefix,
+            **base_role_dict,
+        }
+
+    def get_bin_name(self, bin_number: int = 0) -> str:
+        """Return the name of bin with number ``bin_number``.
+
+        The ``bin_number`` default is 0 as that is used to uniquely identify
+        delegated roles containing succinct_roles from the rest.
```
lukpueh: This seems like an artifact from the first iteration of the PR. Do we still need it? If not, I suggest to remove the default.

tuf/api/metadata.py
```diff
@@ -1435,6 +1444,149 @@ def is_delegated_path(self, target_filepath: str) -> bool:
...
+
+        Args:
+            bin_number: Number of the bin that we want the name for.
+        """
+        # Add zero padding if necessary and cast to hex the suffix.
```
lukpueh: Word order seems off in this comment

jku:
Documenting some possible options WRT `add_key()` and `remove_key()`:
* as in PR: add new methods for succinct case (so users need to remember to call different Targets method depending on what the value of Targets.delegations.succinct_delegations is)
* just document that `role` argument to the `*_key()` functions is unused whenever succinct delegations are used and make the existing functions work in succinct case
* Modify `*_key()` API so that instead of `self, role: str, key: Key` the arguments are `self, key: Key, role: Optional[str] = None`

tuf/api/metadata.py
```diff
@@ -1497,25 +1665,78 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
...
+        """
+        if isinstance(self.roles, dict):
+            for role in self.roles.values():
+                if role.is_delegated_path(target_filepath):
+                    yield role.name, role.terminating
+
+        elif self.succinct_roles is not None:
+            # We consider all succinct_roles as terminating.
+            # For more information read TAP 15.
+            yield self.succinct_roles.find_bin(target_filepath), True
```
lukpueh: Is it worth adding `else: raise RuntimeError`?

tuf/api/metadata.py
```diff
@@ -1773,13 +1994,47 @@ def add_key(self, role: str, key: Key) -> None:
...
+        Args:
+            key: Signing key to be added for succinct_roles.
+
+        Raises:
+            ValueError: If there are no delegations or if this method is called
+            and delegated roles are used or if succinct_roles are not used.
+        """
+        if (
+            self.delegations is None
+            or self.delegations.roles is not None
```
lukpueh: IMO this method shouldn't care about `self.delegations.roles`.

tuf/api/metadata.py
```diff
@@ -1791,16 +2046,57 @@ def remove_key(self, role: str, keyid: str) -> None:
...
-            raise ValueError(f"Key with id {keyid} is not used by {role}")
-        self.delegations.roles[role].keyids.remove(keyid)
-        for keyinfo in self.delegations.roles.values():
-            if keyid in keyinfo.keyids:
-                return
 
+        if self.delegations.succinct_roles is not None:
+            raise ValueError(
+                "Don't use remove_key when succinct_roles is in use"
+            )
```
lukpueh: IMO this method shouldn't care about `self.delegations.succinct_roles`.

tuf/api/metadata.py
```diff
@@ -1773,13 +1994,47 @@ def add_key(self, role: str, key: Key) -> None:
...
+                by this Target.
         """
-        if self.delegations is None or role not in self.delegations.roles:
+        if self.delegations is None:
             raise ValueError(f"Delegated role {role} doesn't exist")
-        if key.keyid not in self.delegations.roles[role].keyids:
-            self.delegations.roles[role].keyids.append(key.keyid)
+
+        if self.delegations.succinct_roles is not None:
+            raise ValueError("Don't use add_key when succinct_roles is in use")
```
lukpueh: IMO this method shouldn't care about `self.delegations.succinct_roles`.

tuf/api/metadata.py
```diff
@@ -1791,16 +2046,57 @@ def remove_key(self, role: str, keyid: str) -> None:
...
+            keyid: Identifier of the key to be removed for succinct_roles.
+
+        Raises:
+            ValueError: If there are no delegations or if this method is called
+            and delegated roles are used or if succinct_roles are not used or if
+            key with id ``keyid`` is not used by succinct roles.
+        """
+        if (
+            self.delegations is None
+            or self.delegations.roles is not None
```
lukpueh: IMO this method shouldn't care about `self.delegations.roles`.

lukpueh (changes requested):
This is really great, @MVrachev. I left a few comments inline, but nothing big a deal.

I also have a few more naming suggestions for `SuccinctRoles` methods, to consolidate with the rest of the API...
```
get_bin_name --> get_role(_for_index)
get_all_bin_names --> get_roles
find_bin --> get_role_for_target
is_bin --> is delegated_role
```
... and for `Delegations`:
```
get_all_delegations -> get_roles_for_target
```

What do you think?

tuf/api/metadata.py
MVrachev: Overall, I like the suggestion.
Most of the sentences I took as they and others I used with small modifications.

tuf/api/metadata.py
MVrachev: I was wondering whether I should leave this method as I thought it could be used by users of TUF.
But also if we see the need for it we can always add it.
Will remove it.

tuf/api/metadata.py
MVrachev: Agree and it looks more readable.
I heavily focused on what is necessary for the `role_name` to be a bin and ended up with long checks.
BTW: `0 <= num < self.number_of_bins` was suggested by `pylint` I think instead of `num >= 0 and num < self.number_of_bins`.

tuf/api/metadata.py
MVrachev: Currently, It's not used in the code, but my reasoning was that I thought that having a method that gives you the correct delegation now when we support succinct roles and standard roles seemed a must-have.
Now, when I think about it we have `get_all_delegations` which does actually the same thing, but instead of returning just one role, it returns all roles responsible for that `target_filepath` which could be many in the case of standard roles.

You are right, we have a replacement for this method, and given it's not currently used it's better to remove it.
Again, we can always add it later if the need arises.

tuf/api/metadata.py
MVrachev: Added a sentence in the commit message about those two API changes.

tuf/api/metadata.py
MVrachev: That's a good catch.
I thought that I need to verify that `self.roles` is actually a `Dict` in order for `mypy` to understands why I am using dictionary methods, but you are right that `self.roles` is defined clearly as `Optional[Dict]` and it shouldn't be something else rather than `None` or `Dict`.

tuf/api/metadata.py
MVrachev: Well, I rely on the idea that the user has created the `Delegations` object and then not deliberately changed the object in such a way that `roles` and `succinct_roles` are both `None`.
If a user has used the constructor and is in a situation where `succinct_roles` and `roles` is None, then he/she will be notified.

In short I am not sure we need that here.

MVrachev:
> get_bin_name --> get_role(_for_index)
> get_all_bin_names --> get_roles
> find_bin --> get_role_for_target
> is_bin --> is delegated_role

I understand where you come from and why those changes make sense for you.
I am worried about using the word `role` inside the `SuccinctRole` method names and especially the phrase `delegated role` because I don't want to confuse our users with `Delegations.roles` or the `DelegatedRole` class instances.
That's why I opt-in for a more concrete naming scheme including `bin` instead of role.
Also, after your docstring suggestion, it seems a lot clearer what a bin actually is.

After I update the pr with your docstring suggestions would you have a look again and tell me if you think those changes are still worth it?

MVrachev:
I have addressed all of your comments @jku and @lukpueh.
For `Targets.*_key()` methods I decided to change the order of the arguments and make `role` an optional.
One downside in this (besides it's a breaking change) is that those methods no longer have the same signature as `Root.*_key()` methods.

Also, I have mentioned in the commit messages about the breaking API changes and also updated the pr description with information about them.

The pr is smaller by a big margin and it's ready for another review.

lukpueh:
> > get_bin_name --> get_role(_for_index)
> > get_all_bin_names --> get_roles
> > find_bin --> get_role_for_target
> > is_bin --> is delegated_role
> 
> I understand where you come from and why those changes make sense for you. I am worried about using the word `role` inside the `SuccinctRole` method names and especially the phrase `delegated role` because I don't want to confuse our users with `Delegations.roles` or the `DelegatedRole` class instances. That's why I opt-in for a more concrete naming scheme including `bin` instead of role. Also, after your docstring suggestion, it seems a lot clearer what a bin actually is.
> 
> After I update the pr with your docstring suggestions would you have a look again and tell me if you think those changes are still worth it?

My idea was to point out exactly that connection, i.e. that a bin is a delegated targets role. But I won't insist. I'm fine with the names you picked as well.

What about my other suggestion `get_all_delegations -> get_roles_for_target`? The idea behind this one is that the function does not really return delegations but delegated roles (or actually their names, and their terminating property) 

I'm a bit pedantic about this, because roles and delegations are confused so often.

MVrachev:
> My idea was to point out exactly that connection, i.e. that a bin is a delegated targets role. But I won't insist. I'm fine with the names you picked as well.
> 
> What about my other suggestion `get_all_delegations -> get_roles_for_target`? The idea behind this one is that the function does not really return delegations but delegated roles (or actually their names, and their terminating property)
> 
> I'm a bit pedantic about this, because roles and delegations are confused so often.

Hmm... more and more when I think about it probably you have some point here.
I keep considering the `SuccinctRoles` instances as one delegated role which is not what is implied even from the class title... `SuccinctRoles` instance just represents a group of delegated roles instead of using a dictionary. Probably mentally it's easier to think `SuccinctRoles` is just one role as `keyids` and `threshold` is shared through all delegated roles represented by that instance.
Probably it will be useful in using the same `delegated role` phrase for `SuccinctRoles` roles as well instead of promoting a new `bin` term.

Will rename the methods as you have suggested here: 
> get_all_bin_names --> get_roles
> find_bin --> get_role_for_target
> is_bin --> is delegated_role

About `get_all_delegations -> get_roles_for_target`.
Yea, I missed it honestly as there were so many renames. I agree with that suggestion. Will do it.

MVrachev:
I followed @lukpueh advice and did the following renames:
- `get_all_bin_names` --> `get_roles`
- `find_bin` --> `get_role_for_target`
- `is_bin` --> `is delegated_role`
- `get_all_delegations` -> `get_roles_for_target`

EDIT: I forced pushed once more to fix the commit message mentioning the old methods name.

jku:
Looks good. The only thing that sticks out to me is the key API.

```
- def add_key(self, role: str, key: Key) -> None:
+ def add_key(self, key: Key, role: Optional[str] = None) -> None:
```

this looks ok to me -- it's easy to notice with static checks and we could even add a specific error message `if isinstance(role, Key):`...

```
- def remove_key(self, role: str, keyid: str) -> None:
+ def remove_key(self, keyid: str, role: Optional[str] = None) -> None:
```
... but I hadn't thought about this one all the way through: both argument types are str here. It really is the most annoying kind of API change, very easy to mess up.

MVrachev:
> Looks good. The only thing that sticks out to me is the key API.
> 
> ```
> - def add_key(self, role: str, key: Key) -> None:
> + def add_key(self, key: Key, role: Optional[str] = None) -> None:
> ```
> 
> this looks ok to me -- it's easy to notice with static checks and we could even add a specific error message `if isinstance(role, Key):`...

Maybe this could be useful. Preventing users from mistaking the arguments order.

> 
> ```
> - def remove_key(self, role: str, keyid: str) -> None:
> + def remove_key(self, keyid: str, role: Optional[str] = None) -> None:
> ```
> 
> ... but I hadn't thought about this one all the way through: both argument types are str here. It really is the most annoying kind of API change, very easy to mess up.

Yea... if only the types were different. I was thinking if we can use the insider knowledge we have about `keyid`.
For example that it will have a length longer than X number of symbols, but then what stops a user from having very long delegated role names? Nothing...

I hope we can make this API change as still there aren't so many users, especially in the 1.0.0 version of python-tuf and calls like `targets.add_key(None, key_obj)` and  ``targets.add_key(None, keyid)` are not really nice...

jku:
> I was thinking if we can use the insider knowledge we have about keyid.

I don't think we have any insider knowledge about it -- theoretically it's a hash but
* we've explicitly avoided using that in python-tuf as it seems like a mistake: it should be just a unique id 
* even if we know keyid is a hash, nothing prevents rolenames from also being hashes (in fact I could imagine it would be a good way to handle rolenames in some cases)

MVrachev:
@lukpueh any thoughts?

jku:
I think there's still several possibilities here that wouldn't require surprisingly breaking applications:
* keep current API (maybe allow None for rolename), just document it well enough. I think this is completely within reason. It's a wart in the API but that happens.
* `@overload` magic so that we could have both `remove_key(rolename, keyid)` and `remove_key(keyid)` -- the code will not be very pretty but should work.
* just rename the API: remove_key() stays as is for a few releases and we have a new method with different API -- the lovely linux naming tradition would give us `remove_key2()` :grimacing: 
* probably something else still?

lukpueh:
> I think there's still several possibilities here that wouldn't require surprisingly breaking applications:
> 
> * keep current API (maybe allow None for rolename), just document it well enough. I think this is completely within reason. It's a wart in the API but that happens.

This would be an okay solution.

> * `@overload` magic so that we could have both `remove_key(rolename, keyid)` and `remove_key(keyid)` -- the code will not be very pretty but should work.

I'm not sure about this one. It is definitely unidiomatic (~> surprising). And the code to do this would likely stain the otherwise elegant metadata module.

> * just rename the API: remove_key() stays as is for a few releases and we have a new method with different API -- the lovely linux naming tradition would give us `remove_key2()` 😬

I like this suggestion best, at least the first part. But I'd rather not have a `remove_key2` in our API. And I also don't think we need to support the old method for a few releases, if we make this release a breaking one.

So what about replacing `remove_key` right away with a name that also makes sense and that we like. Some ideas: `revoke_key`, `deauthorize`, `unauthorize`, *something with delegation in its name*, etc...

Some consistency considerations:
- Will the new name work as counterpart to `add_key` or do we need to rename that too?
- What about `Root`'s `{add, remove}_key` methods. Should we adopt a rename there too? And the argument order? This would make it a bigger cut, but API consistency seems important.

> * probably something else still?

Couldn't think of anything else.

jku:
Yes, agreed that providing the old deprecated method isn't terribly important if using the new method is basically a copy-paste fix

examples/repo_example/hashed_bin_delegation.py
```diff
@@ -162,6 +162,7 @@ def find_hash_bin(path: str) -> str:
 #              10-17                       10 11 12 13 14 15 16 17
 #              ...                         ...
 #              f8-ff                       f8 f9 fa fb fc fd fe ff
+assert isinstance(roles["bins"].signed.delegations.roles, dict)
```
lukpueh: Why assert this here?

tests/test_api.py
```diff
@@ -517,46 +519,89 @@ def test_targets_key_api(self) -> None:
...
+            "keyval": {
+                "public": "edcd0a32a07dce33f7c7873aaffbff36d20ea30787574ead335eefd337e4dacd"
+            },
+            "scheme": "ed25519",
+        }
+        key = Key.from_dict("id2", key_dict)
+
+        # Remove delegated roles.
+        assert isinstance(targets.delegations, Delegations)
+        assert targets.delegations.roles is not None
```
lukpueh, line 584: Why `assert` and not `self.assert*`?

tests/test_api.py
```diff
@@ -695,6 +740,38 @@ def test_is_delegated_role(self) -> None:
             self.assertFalse(role.is_delegated_path("a/non-matching path"))
             self.assertTrue(role.is_delegated_path("a/path"))
 
+    def test_zero_padding_in_succinct_roles(self) -> None:
+        bit_lengths = [2, 6, 8, 10, 16, 32]
```
lukpueh: Any reason for these exact bit lengths? Why not test `range(1, 32)`?

tests/test_api.py
```diff
@@ -695,6 +740,38 @@ def test_is_delegated_role(self) -> None:
...
+        expected_paddings = [1, 2, 2, 3, 4, 8]
```
lukpueh: Minor naming nit: These are not paddings.

tests/test_api.py
```diff
@@ -695,6 +740,38 @@ def test_is_delegated_role(self) -> None:
...
+
+        for bit_length, expected in zip(bit_lengths, expected_paddings):
+            succinct_delegation = SuccinctRoles([], 1, bit_length, "foo")
+            m = f"Error for {bit_length} expected padding {expected}"
+            self.assertEqual(succinct_delegation.suffix_len, expected, m)
+
+    def test_is_delegated_role_in_succinct_roles(self) -> None:
+        succinct_roles = SuccinctRoles([], 1, 5, "bin")
+        # delegated role name suffixes are in hex format.
```
lukpueh: "bin-20" and "bin-100" are valid hex numbers, so the comment feels a bit odd

tests/test_api.py
```diff
@@ -695,6 +740,38 @@ def test_is_delegated_role(self) -> None:
...
+        for role_name in true_name_examples:
+            msg = f"Error for {role_name}"
+            self.assertTrue(succinct_roles.is_delegated_role(role_name), msg)
+
+    def test_get_roles_in_succinct_roles(self) -> None:
+        succinct_roles = SuccinctRoles([], 1, 3, "bin")
+        bin_amount = 0
+        for i, role_name in enumerate(succinct_roles.get_roles()):
+            self.assertEqual(role_name, f"bin-{i}")
+            bin_amount = i
```
lukpueh: `bin_amount` does not seem very useful. :)

tests/test_api.py
```diff
@@ -695,6 +740,38 @@ def test_is_delegated_role(self) -> None:
...
+        true_name_examples = ["bin-00", "bin-0f", "bin-1f"]
+        for role_name in true_name_examples:
+            msg = f"Error for {role_name}"
+            self.assertTrue(succinct_roles.is_delegated_role(role_name), msg)
+
+    def test_get_roles_in_succinct_roles(self) -> None:
+        succinct_roles = SuccinctRoles([], 1, 3, "bin")
+        bin_amount = 0
+        for i, role_name in enumerate(succinct_roles.get_roles()):
+            self.assertEqual(role_name, f"bin-{i}")
```
lukpueh: This assert would become more interesting for a bit_length > 3, which would implicitly also test the padding and you wouldn't need `test_zero_padding_in_succinct_roles`.

tests/test_metadata_eq_.py
```diff
@@ -220,6 +221,19 @@ def test_delegated_role_eq_(self) -> None:
             # Restore the old value of the attribute.
             setattr(delegated_role_2, attr, getattr(delegated_role, attr))
 
+    def test_succinct_roles_eq(self) -> None:
```
lukpueh: Let's merge #2017 first and update this test accordingly, shall we?

lukpueh (changes requested):
Great tests, @MVrachev. I just reviewed:

```
docs/api/tuf.api.metadata.supporting.rst
docs/repository-library-design.md
examples/repo_example/hashed_bin_delegation.py
tests/test_api.py
tests/test_metadata_eq_.py
```

I'll review the rest tomorrow.

MVrachev:
I think `revoke_key` and the counterpart `add_key` sounds okay.

If we decide to do a rename to `remove_key` then does this mean that we are okay with its counterpart - `add_key` and its changes after I add a check from the type `if isinstance(role, Key)` as Jussi has suggested preventing mistakes?

lukpueh:
Any thoughts about also changing `Root.{add, remove}_key`?

tests/repository_simulator.py
```diff
@@ -354,21 +355,60 @@ def add_delegation(
...
+            and delegator.delegations.succinct_roles is not None
+        ):
+            raise ValueError("Can't add a role when succinct_roles is used")
+
         # Create delegation
         if delegator.delegations is None:
-            delegator.delegations = Delegations({}, {})
+            delegator.delegations = Delegations({}, roles={})
+
+        assert delegator.delegations.roles is not None
```
lukpueh, line 369: Is an `AssertionError` here better than a `TypeError` in the next line?

tests/repository_simulator.py
```diff
@@ -354,21 +355,60 @@ def add_delegation(
...
+        ):
+            raise ValueError(
+                "Can't add a succinct_roles when delegated roles are used"
+            )
+
+        key, signer = self.create_key()
+        succinct_roles = SuccinctRoles([], 1, bit_length, name_prefix)
+        delegator.delegations = Delegations({}, None, succinct_roles)
+
+        # Add target metadata for all bins.
```
lukpueh: Nit: 
```suggestion
        # Add targets metadata for all bins.
```

tests/repository_simulator.py
```diff
@@ -354,21 +355,60 @@ def add_delegation(
...
         key, signer = self.create_key()
-        delegator.add_key(role.name, key)
+        delegator.add_key(key, role.name)
         self.add_signer(role.name, signer)
 
         # Add metadata for the role
         if role.name not in self.md_delegates:
             self.md_delegates[role.name] = Metadata(targets, {})
 
+    def add_succinct_roles_to_delegations(
```
lukpueh: You usually advocate for short names... I think  `add_succinct_roles` would be good enough.

tests/repository_simulator.py
```diff
@@ -354,21 +355,60 @@ def add_delegation(
...
+        self, delegator_name: str, bit_length: int, name_prefix: str
+    ) -> None:
+        """Add succinct roles info to a delegator with name "delegator_name"."""
+        if delegator_name == Targets.type:
+            delegator = self.targets
+        else:
+            delegator = self.md_delegates[delegator_name].signed
```
lukpueh: Maybe worth adding a helper for this if/else to be used here, in `add_target` and `add_delegation`.

tests/repository_simulator.py
```diff
@@ -354,21 +355,60 @@ def add_delegation(
...
         self.add_signer(role.name, signer)
 
         # Add metadata for the role
         if role.name not in self.md_delegates:
             self.md_delegates[role.name] = Metadata(targets, {})
 
+    def add_succinct_roles_to_delegations(
+        self, delegator_name: str, bit_length: int, name_prefix: str
+    ) -> None:
+        """Add succinct roles info to a delegator with name "delegator_name"."""
```
lukpueh: I'd probably add a note that you are also bootstrapping empty bins metadata in this function.

tests/test_metadata_serialization.py
```diff
@@ -174,6 +175,7 @@ def test_invalid_key_serialization(self, test_case_data: str) -> None:
     invalid_roles: utils.DataSet = {
         "no threshold": '{"keyids": ["keyid"]}',
         "no keyids": '{"threshold": 3}',
+        "wrong keyids type": '{"keyids": 1, "threshold": 3}',
```
lukpueh: This seems unrelated, but I guess it's okay to include it in this PR.

MVrachev:
> Any thoughts about also changing `Root.{add, remove}_key`?

Probably it's best if we rename them as well, right?
As they are logically the same - adding or removing a key of a delegated role.
Any thoughts @jku on the above discussion and `Root.*_key()` API renames?

PS: Before any new updates on this pr I will wait for the full tests review and decision on what to do with the `Root` and `Targets` key API methods.

jku:
I think all *_key() methods should have the same logic (so yes root methods should change argument order and name if the targets methods do, even if they don't practically need to). 

Renaming to revoke_key() is acceptable I suppose? seems like best idea so far.

tests/test_updater_delegation_graphs.py
```diff
@@ -425,6 +425,90 @@ def test_hash_bins_graph_traversal(
...
+            # Check that metadata dir contains only top-level roles
+            self._assert_files_exist(TOP_LEVEL_ROLE_NAMES)
+
+            # Looking for a non-existing targetpath forces updater
+            # to visit a correspondning delegated role
+            targetfile = updater.get_targetinfo(test_data.target_path)
+            self.assertIsNone(targetfile)
+
+            # Check that the delegated roles were visited in the expected
+            # order and the corresponding metadata files were persisted
```
lukpueh:
```suggestion
            # order and the corresponding metadata files were persisted
```

MVrachev:
Okay, when @lukpueh signals me that he has reviewed all of the tests and after merging #2017 I will update the pr addressing all comments and renaming the `{Root/Targets}*_key` methods.

tests/test_updater_delegation_graphs.py
```diff
@@ -425,6 +425,90 @@ def test_hash_bins_graph_traversal(
...
+            bit_length=2,
+            target_path="bar",
+            expected_visited_bin="delegated-3",
+        ),
+        "many bins case": SuccinctRolesTestCase(
+            bit_length=8,
+            target_path="bar",
+            expected_visited_bin="delegated-fc",
+        ),
+    }
```
lukpueh, line 473: Very cool test! I must admit though that it took me a while to understand what it's doing. I have some ideas to make this a bit easier. Feel free to adopt/ignore as you please.

- `expected_visited_bin` implied (to me) that it is some node along the tree traversal, but the tree is actually pretty shallow (root + leaves). Maybe we can just call it  `expected_bin`, as it is the bin we expect the target to fall into.
- Using `delegated-` as name prefix makes sense, but if you call it `bin-` it would further reduce complexity (e.g.: `expected_bin="bin-0"`) 
- Some `target_path` values ("missingpath", "othermissingpath") convey an irrelevant meaning, others don't ("foo", "doo", ...). This is confusing. Maybe you can use all placeholder names?
- A few lines of comments would make my suggestions from above less important. E.g. how `bit_length` determines both the total number of bins and the bin a `target_path` falls into, based on its hash prefix...
- Similarly/alternatively, some of this information could be encoded in the test case names, which are currently not very descriptive, e.g. instead of "case 6" it could say "bin count: 4, target bin index: 3"

lukpueh (changes requested):
Part 2/2 of tests review. Includes
```
tests/repository_simulator.py
tests/test_metadata_serialization.py
tests/test_updater_delegation_graphs.py
```

tests/test_updater_delegation_graphs.py
lukpueh: Oops wrong line.

tests/test_updater_delegation_graphs.py
```diff
@@ -425,6 +425,90 @@ def test_hash_bins_graph_traversal(
...
+
+            updater = self._init_updater()
+            # Call explicitly refresh to simplify the expected_calls list
+            updater.refresh()
+            self.sim.fetch_tracker.metadata.clear()
+            # Check that metadata dir contains only top-level roles
+            self._assert_files_exist(TOP_LEVEL_ROLE_NAMES)
+
+            # Looking for a non-existing targetpath forces updater
+            # to visit a correspondning delegated role
```
lukpueh: That's the one with the typo:
```suggestion
            # to visit a corresponding delegated role
```

tests/test_updater_delegation_graphs.py
lukpueh: Please ignore this one.

examples/repo_example/hashed_bin_delegation.py
MVrachev: Because of `mypy` I have to assert that `roles["bins"].signed.delegations.roles` is None as a few lines later I have an assignment. 
Will change the assert to verify that `roles["bins"].signed.delegations.roles is not None` as if it's not none it can only be a dict.

tests/test_metadata_serialization.py
MVrachev: Oh, I just noticed we are missing it and added it.
Will add this change to the commit with changes related to `DelegatedRole`.

tests/test_api.py
MVrachev: Agree, will rename it to `expected_suffix_lengths`.

tests/test_api.py
MVrachev: I wanted to choose a subset of key `bit_lenghts` in which the expected suffix length will be different if I want to test against `range(1, 32)` I will have to come up with a formula about the expected suffix length (which even if I do it won't be so readable and maintainable) or have a list of 32 `bit_lengths` and 32 expected suffix lengths which I am not sure it makes sense.

tests/test_api.py
MVrachev: Okay, will move this comment before the `true_name_examples` and then it will make more sense.

tests/test_api.py
MVrachev: Okay, will remove it.

tests/test_api.py
MVrachev: Okay, will do it.

tests/repository_simulator.py
MVrachev: I do this because of mypy. Otherwise, it gives a warning:
`error: Unsupported target for indexed assignment ("Optional[Dict[str, DelegatedRole]]")  [index]`

tests/test_api.py
MVrachev: `mypy` doesn't recognize `self.assert*` methods.
For example, if I replace `assert isinstance(targets.delegations, Delegations)` with `self.assertIsNotNone(targets.delegations)` then I multiple warnings like:
`error: Item "None" of "Optional[Delegations]" has no attribute "roles"  [union-attr]`

tests/repository_simulator.py
MVrachev: Yep, I do advocate for shorter names.
You are correct, that would work.

MVrachev:
I tried addressing all comments made by @lukpueh.
Hope it's better now :)).
Ready for review.

jku:
> I tried addressing all comments made by @lukpueh.

I'm not against force-pushing large PRs, that's often still the best path forward... but if that is chosen (vs adding commits and potentially squashing after review), a clear list of changes since the last review (especially for the API changes) would be nice.

This is a bit of a review pain point in github as it's really hard to see changes between PR versions. I _think_ in this case it's this https://github.com/theupdateframework/python-tuf/compare/a1fb89c..c04ec75

MVrachev:
Yes, you are right that probably some summary of my changes will be good.
All of them were addressing Lukas comments, so those comments can give you some idea what I changed + I renamed the Root/Targets.remove_key() to Root/Targets.revoke_key() as we had discussed.

What I did ca be summarized like this:
1. Removed the `zero_padding` test in favor of a more complicated `test_get_roles_in_succinct_roles` test.
2. Moved a comment about bin names inside `test_is_delegated_role_in_succinct_roles` a few lines below.
3. Updated the test about `__eq__` implementation for SuccinctRoles after the merge of #2017.
4. Some typo fixes.
5. Renamed `RepositorySimulator. add_succinct_roles_to_delegations()` to `RepositorySimulator.add_succinct_roles()` and added a better docstring.
6. Added a helper `RepositorySimulator._get_delegator()` which removes code dublication in `add_target`, `add_delegation` and `add_succinct_roles` in `RepositorySimulator`.
7. Updated `test_succinct_roles_graph_traversal` with many suggestions by @lukas from this comment https://github.com/theupdateframework/python-tuf/pull/2010#discussion_r892305989
8. Renamed `Root/Targets.remove_key()` to `Root/Targets.revoke_key()` plus updated the pr description and commit description mentioning those API changes.

tuf/api/metadata.py
```diff
@@ -954,8 +961,8 @@ def add_key(self, role: str, key: Key) -> None:
             self.roles[role].keyids.append(key.keyid)
         self.keys[key.keyid] = key
 
-    def remove_key(self, role: str, keyid: str) -> None:
-        """Removes key from ``role`` and updates the key store.
+    def revoke_key(self, role: str, keyid: str) -> None:
```
jku: this (and add_key()) are now named exactly like the Targets methods, but have exactly opposite argument order...

MVrachev: Yes, and I was wondering should I change them just for the sake of consistency?
We changed the order just because we want to make `role` optional, but don't want to do that in `Root`.
So, what do you think?

lukpueh: > Yes, and I was wondering should I change them just for the sake of consistency?

Yes, that seems to be what we agreed on [above](https://github.com/theupdateframework/python-tuf/pull/2010#issuecomment-1149807399).

jku: Me earlier:
> I think all *_key() methods should have the same logic (so yes root methods should change argument order and name if the targets methods do, even if they don't practically need to).

I think it's bad if we have methods that look very similar but are not. So IMO either we don't change any argument order or we change all of them.

MVrachev: Okay, will make the arguments in the same order.

MVrachev:
I have updated the pr by making `Root.*_key()` API methods with consistent naming and argument order as the ones in `Targets.*_key()`.

tuf/api/metadata.py
```diff
@@ -1285,6 +1297,8 @@ class DelegatedRole(Role):
         ``paths`` and ``path_hash_prefixes`` are mutually exclusive: both cannot be
         set, at least one of them must be set.
 
+    Exactly one of ``paths`` and ``path_hash_prefixes`` must be set.
```
jku: nitpick: I think this is not needed now: the previous sentence seems to say the same thing.

tuf/api/metadata.py
```diff
@@ -1435,6 +1449,149 @@ def is_delegated_path(self, target_filepath: str) -> bool:
...
+    bits of the file path hash digest to int, using it as bin index between 0
+    and ``2**bit_length - 1``.
+
+    Bin names are the concatenation of the passed ``name_prefix`` and a
+    zero-padded hex representation of the bin index separated by a hyphen.
+
+    The passed ``keyids`` and ``threshold`` is used for each bin, and each bin
+    is 'terminating'.
+
+    For details: https://github.com/theupdateframework/taps/blob/master/tap15.md
```
jku: I think there's still a lot to improve in this documentation: it contains loads of detail (like _"casting the left-most bit_length of bits of the file path hash digest to int, using it as bin index between 0 and 2**bit_length - 1"_), but probably does not help the user that much.

That said, I don't think we need to polish it in this PR, and should look at this more holistically: maybe the doc improvements should happen elsewhere.

tuf/api/metadata.py
```diff
@@ -1497,25 +1672,58 @@ def from_dict(cls, delegations_dict: Dict[str, Any]) -> "Delegations":
...
-        return {
+        res_dict: Dict[str, Any] = {
             "keys": keys,
-            "roles": roles,
             **self.unrecognized_fields,
         }
+        if self.roles is not None:
+            roles = [role_obj.to_dict() for role_obj in self.roles.values()]
+            res_dict["roles"] = roles
+        elif self.succinct_roles is not None:
```
jku, line 1707: This is about an edge case but... should this be just `else:`?

because I think we might prefer an exception at serialization if both roles and succinct_roles are None: that should not have happened.

tuf/api/metadata.py
```diff
@@ -1765,42 +1973,73 @@ def to_dict(self) -> Dict[str, Any]:
...
         Raises:
-            ValueError: If there are no delegated roles or if ``role`` is not
-                delegated by this Target.
+            ValueError: If the argument order is wrong or if there are no
+                delegated roles or if ``role`` is not delegated by this Target.
         """
-        if self.delegations is None or role not in self.delegations.roles:
+        # Verify that our users are not using the old argument order.
+        if isinstance(role, Key):
+            raise ValueError("Argument order has changed")
```
jku: nitpick: we shouldn't make guesses about how the error happened, we should just report what the situation is

jku (approved):
I think I'm happy with this -- left some nitpicks but they're not critical. The main question I have is documentation:
* The documentation currently does not make it clear that this feature is not a TUF specification feature: maybe it should?

As TODO items (that might exist already):
* we should improve the succinct delegation docs and update the example script -- this is something we should do before we release this
* for completeness, actual download tests should contain succinct delegations: it may make sense to refactor the download tests at that point

tuf/api/metadata.py
lukpueh: Haha, yes, I take the blame for this one, and I absolutely agree that it should become more readable. But I also think that the user of this interface needs to know how``bit_length`` functions, in order to choose an appropriate value.

Ideally the user isn't bothered with these details at all and just passes the expected amount of target files, and the tool figures out an appropriate ``bit_length``.

MVrachev: Honestly, it was hard for me to describe what `SuccinctRoles` is without mentioning `bit_length` definition and what it represents. 
I feel like a short summary of how a target is assigned to a bin should be here or `get_role_for_target`. 
Now when I think about this... I think it does make more sense to be inside `get_role_for_target` as that's the actual API doing that.
Here it should be a more general description of `SuccinctRoles` and not so much how it does its magic.

tuf/api/metadata.py
MVrachev: I agree with you, but `mypy` wants to know that `succinct_roles is not None` in order to call `succinct_roles.to_dict()` otherwise it complains.
It's absolutley the same situation in `DelegatedRole.to_dict()`.
Do you want me to add an additional `else` and throw an error if they are missing `paths/path_hash_prefixes` or `roles/succinct_roles` respectivly? 
I don't think it's necessary given we have a check in `__init__`.

MVrachev:
Updated the pr with the following changes:
- added a note that `SuccinctRoles` is not supported by the spec
- addressed all other changes mentioned by @jku

tuf/api/metadata.py
```diff
@@ -1447,28 +1603,40 @@ class Delegations:
...
             role. The roles order also defines the order that role delegations
             are considered during target searches.
+        succinct_roles: Contains succinct information about hash bin
+            delegations. For more information see: https://github.com/theupdateframework/taps/blob/master/tap15.md
         unrecognized_fields: Dictionary of all attributes that are not managed
             by TUF Metadata API
 
+    Exactly one of ``roles`` and ``succinct_roles`` must be set.
+
+    Note: succinct_roles is a feature not yet part of the specification.
```
jku: API documentation should be direct: why are we telling this to the user of this API, and what can they do about it? Something like 

```suggestion
    Note that succinct roles is not a TUF specification feature yet and setting
    `succinct_roles` to a value makes the resulting metadata non-compliant.
    The metadata will not be accepted as valid by specification compliant 
    clients such as those built with python-tuf <= 1.1.0.
```

Could even move this to the actual argument description :shrug:

MVrachev:
Accepted @jku suggested how to document that `succinct_roles` is not in the spec yet and moved that documentation next to the `succinct_roles` argument description in `Delegations`.

jku:
Yes, looks good to me. Thanks everyone! The rough plan here is:
* Merge this PR: succinct delegations are then usable on "develop" branch
* Get any remaining polish done on TAP-15 and move it to "accepted" state https://github.com/theupdateframework/taps/pull/148
* release python-tuf with TAP-15 support (likely means v2.0.0)
* Maybe TAP-15 gets integrated into spec at some point?


@lukpueh, would you like to do a final review of the PR as the system thinks your review is pending? Checking  _"merge without waiting for requirements to be met"_ makes me feel criminal...

@MVrachev would you mind having a last look to identify any issues we may need to file for future work? 
* Updating the example code is one (if it does not exist already)
* Maybe also one for better parameterised target download tests that use succinct bins and normal bins -- there is a PR for the latter but I think we can do better

MVrachev:
> @MVrachev would you mind having a last look to identify any issues we may need to file for future work?
> 
> * Updating the example code is one (if it does not exist already)
> * Maybe also one for better parameterised target download tests that use succinct bins and normal bins -- there is a PR for the latter but I think we can do better

We have an issue https://github.com/theupdateframework/python-tuf/issues/1909 where we keep track of that stuff.
Only three items are left on the list and I am not sure there is a need for separate issues.

tests/test_updater_delegation_graphs.py
```diff
@@ -425,6 +425,91 @@ def test_hash_bins_graph_traversal(
         finally:
             self.teardown_subtest()
 
+    @dataclass
+    class SuccinctRolesTestCase:
+        bit_length: int
+        target_path: str
+        expected_target_bin: str
+
+    # By setting the bit_length the total number of is 2^bit_length.
```
lukpueh:
```suggestion
    # By setting the bit_length the total number of bins is 2^bit_length.
```

tuf/api/metadata.py
```diff
@@ -1435,6 +1447,150 @@ def is_delegated_path(self, target_filepath: str) -> bool:
...
+
+    def get_roles(self) -> Iterator[str]:
+        """Yield the names of all different delegated roles one by one."""
+        for i in range(0, self.number_of_bins):
+            suffix = f"{i:0{self.suffix_len}x}"
+            yield f"{self.name_prefix}-{suffix}"
+
+    def is_delegated_role(self, role_name: str) -> bool:
+        """Determines whether the given ``role_name`` is in one of
+        the delegated roles that ``SuccinctRoles`` represent.
```
```suggestion
        the delegated roles that ``SuccinctRoles`` represents.
```

lukpueh (approved):
LGTM! 💯

lukpueh (approved):

PR #2024, opened 2022-06-13
ngclient: pick old timestamp if new.version is equal

### LINKED ISSUE #2023: ngclient: used trusted timestamp if new has equal version

The [specification was changed](https://github.com/theupdateframework/specification/compare/v1.0.29..v1.0.30) to say wrt client timestamp update:
> In case they [versions] are equal, discard the new
timestamp metadata and abort the update cycle.
This is normal and it shouldn't raise any error

To conform with spec v1.0.30, ngclient needs to be updated. See #1992 for discussion.

**Current behavior**:

Update cycle fails if new timestamp version is less (`<`) than trusted timestamp version (L224). Otherwise (`>=`) , update cycle continues and the new timestamp becomes the trusted timestamp (L241).

https://github.com/theupdateframework/python-tuf/blob/46979bb46d3054ca37fc0b9258574a126d975fcc/tuf/ngclient/_internal/trusted_metadata_set.py#L222-L242

**Expected behavior**:

Update cycle fails if new timestamp version is less (`<`) than trusted timestamp version. Otherwise (`>=`), update cycle continues. If new and old timestamp versions are equal (`=`), new timestamp is ignored and old timestamp remains trusted.  If new timestamp version is higher (`>`) new timestamp becomes the trusted timestamp.

MVrachev:
Fixes #2023

In the spec version `1.0.30`, a new change has been added considering what
should happen if there is a new timestamp with the same version.
It says the following:
```
In case they [versions] are equal, discard the new
timestamp metadata and abort the update cycle.
This is normal and it shouldn't raise any error.
```

In other words, if there is a new timestamp with the same version, then
stop the update process and use the old timestamp.

Those changes reflect these latest specification modifications.

requirements**:

### DIFF
--- a/tests/test_trusted_metadata_set.py
+++ b/tests/test_trusted_metadata_set.py
@@ -274,9 +274,25 @@ class TestTrustedMetadataSet(unittest.TestCase):
         with self.assertRaises(exceptions.BadVersionNumberError):
             self.trusted_set.update_timestamp(self.metadata[Timestamp.type])
 
+    def test_update_timestamp_with_same_timestamp(self) -> None:
+        # Test that timestamp is NOT updated if:
+        # new_timestamp.version == trusted_timestamp.version
+        self.trusted_set.update_timestamp(self.metadata[Timestamp.type])
+        initial_timestamp = self.trusted_set.timestamp
+
+        # Update timestamp with the same version.
+        with self.assertRaises(exceptions.EqualVersionNumberError):
+            self.trusted_set.update_timestamp((self.metadata[Timestamp.type]))
+
+        # Every object has a unique id() if they are equal, this means timestamp
+        # was not updated.
+        self.assertEqual(id(initial_timestamp), id(self.trusted_set.timestamp))
+
     def test_update_timestamp_snapshot_ver_below_current(self) -> None:
         def bump_snapshot_version(timestamp: Timestamp) -> None:
             timestamp.snapshot_meta.version = 2
+            # The timestamp version must be increased to initiate a update.
+            timestamp.version += 1
 
         # set current known snapshot.json version to 2
         timestamp = self.modify_metadata(Timestamp.type, bump_snapshot_version)
@@ -382,6 +398,8 @@ class TestTrustedMetadataSet(unittest.TestCase):
     def test_update_snapshot_successful_rollback_checks(self) -> None:
         def meta_version_bump(timestamp: Timestamp) -> None:
             timestamp.snapshot_meta.version += 1
+            # The timestamp version must be increased to initiate a update.
+            timestamp.version += 1
 
         def version_bump(snapshot: Snapshot) -> None:
             snapshot.version += 1
--- a/tuf/api/exceptions.py
+++ b/tuf/api/exceptions.py
@@ -29,6 +29,10 @@ class BadVersionNumberError(RepositoryError):
     """An error for metadata that contains an invalid version number."""
 
 
+class EqualVersionNumberError(BadVersionNumberError):
+    """An error for metadata containing a previously verified version number."""
+
+
 class ExpiredMetadataError(RepositoryError):
     """Indicate that a TUF Metadata file has expired."""
 
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -226,6 +226,10 @@ class TrustedMetadataSet(abc.Mapping):
                     f"New timestamp version {new_timestamp.signed.version} must"
                     f" be >= {self.timestamp.signed.version}"
                 )
+            # Keep using old timestamp if versions are equal.
+            if new_timestamp.signed.version == self.timestamp.signed.version:
+                raise exceptions.EqualVersionNumberError()
+
             # Prevent rolling back snapshot version
             snapshot_meta = self.timestamp.signed.snapshot_meta
             new_snapshot_meta = new_timestamp.signed.snapshot_meta
--- a/tests/test_updater_top_level_update.py
+++ b/tests/test_updater_top_level_update.py
@@ -804,6 +804,27 @@ class TestRefresh(unittest.TestCase):
         updater = self._init_updater()
         updater.refresh()
 
+    def test_timestamp_eq_versions_check(self) -> None:
+        # Test that a modified timestamp with different content, but the same
+        # version doesn't replace the valid locally stored one.
+
+        # Make a successful update of valid metadata which stores it in cache
+        self._run_refresh()
+        initial_timestamp_meta_ver = self.sim.timestamp.snapshot_meta.version
+
+        # Change timestamp without bumping its version in order to test if a new
+        # timestamp with the same version will be persisted.
+        self.sim.timestamp.snapshot_meta.version = 100
+        self._run_refresh()
+
+        # If the local timestamp md file has the same snapshot_meta.version as
+        # the initial one, then the new modified timestamp has not been stored.
+        timestamp_path = os.path.join(self.metadata_dir, "timestamp.json")
+        timestamp: Metadata[Timestamp] = Metadata.from_file(timestamp_path)
+        self.assertEqual(
+            initial_timestamp_meta_ver, timestamp.signed.snapshot_meta.version
+        )
+
 
 if __name__ == "__main__":
     if "--dump" in sys.argv:
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -336,7 +336,13 @@ class Updater:
         data = self._download_metadata(
             Timestamp.type, self.config.timestamp_max_length
         )
-        self._trusted_set.update_timestamp(data)
+        try:
+            self._trusted_set.update_timestamp(data)
+        except exceptions.EqualVersionNumberError:
+            # If the new timestamp version is the same as current, discard the
+            # new timestamp. This is normal and it shouldn't raise any error.
+            return
+
         self._persist_metadata(Timestamp.type, data)
 
     def _load_snapshot(self) -> None:


tests/test_trusted_metadata_set.py
```diff
@@ -274,9 +274,24 @@ def version_modifier(timestamp: Timestamp) -> None:
...
+        # new_timestamp.version == trusted_timestamp.version
+        self.trusted_set.update_timestamp(self.metadata[Timestamp.type])
+        initial_timestamp = self.trusted_set.timestamp
+
+        # Update timestamp with the same version.
+        self.trusted_set.update_timestamp((self.metadata[Timestamp.type]))
+
+        # Every object has a unique id() if they are equal, this means timestamp
+        # was not updated.
+        self.assertEqual(id(initial_timestamp), id(self.trusted_set.timestamp))
```
lukpueh, line 289: This works because we know that `update_timestamp` assigns a new object to `self.trusted_set.timestamp` in case of an update. But the test wouldn't detect a change in the same object. E.g.

```python
class Timestamp: pass

# Create initial object
timestamp = Timestamp()

# Copy reference
initial_timestamp = timestamp

# Actually update initial object
timestamp.attr_change = "not detected"
 
# Change not detected
assert id(initial_timestamp) == id(timestamp)
```

tests/test_trusted_metadata_set.py
```diff
@@ -274,9 +274,24 @@ def version_modifier(timestamp: Timestamp) -> None:
...
+
     def test_update_timestamp_snapshot_ver_below_current(self) -> None:
         def bump_snapshot_version(timestamp: Timestamp) -> None:
             timestamp.snapshot_meta.version = 2
+            # The timestamp version must be incresed to initiate a updade.
```
lukpueh:
```suggestion
            # The timestamp version must be increased to initiate a update.
```

tests/test_trusted_metadata_set.py
```diff
@@ -382,6 +397,8 @@ def snapshot_expired_modifier(snapshot: Snapshot) -> None:
     def test_update_snapshot_successful_rollback_checks(self) -> None:
         def meta_version_bump(timestamp: Timestamp) -> None:
             timestamp.snapshot_meta.version += 1
+            # The timestamp version must be incresed to initiate a updade.
```
lukpueh:
```suggestion
            # The timestamp version must be increased to initiate a update.
```

tests/test_trusted_metadata_set.py
lukpueh: IMO it's safer to make a deep copy and assertEqual the objects. 🤷

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -226,6 +226,10 @@ def update_timestamp(self, data: bytes) -> Metadata[Timestamp]:
                     f"New timestamp version {new_timestamp.signed.version} must"
                     f" be >= {self.timestamp.signed.version}"
                 )
+            # If new_timestamp.version == trusted_timestamp.version ignore it.
+            if new_timestamp.signed.version == self.timestamp.signed.version:
```
lukpueh, line 230: Minor comment nits: 
- Copy-pasting the commented code is not very useful, and actually confusing if it's slightly modified
-  What is 'it'? What is ignored?

What about something like
```suggestion
            # Keep using old timestamp if versions are equal
```

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -226,6 +226,10 @@ def update_timestamp(self, data: bytes) -> Metadata[Timestamp]:
...
+                return self.timestamp
```
lukpueh: If we keep using the old timestamp, we should check its expiry (see self._check_final_timestamp()). cc @jku

lukpueh (changes requested):
Thanks for the quick patch, @MVrachev! I love that it's such a minimal change.

tuf/ngclient/_internal/trusted_metadata_set.py
jku: this happens in the beginning of update_snapshot() -- when we know we should have final timestamp.

jku:
Can you double check how the updater code paths work with this: we now return "success" from update_timestamp() even though we didn't quite accept the update -- how does this affect what gets written to disk?

tuf/ngclient/_internal/trusted_metadata_set.py
lukpueh: I see, and we check the new one here too, so that we don't persist an already expired one?

lukpueh:
> Can you double check how the updater code paths work with this: we now return "success" from update_timestamp() even though we didn't quite accept the update -- how does this affect what gets written to disk?

Ouch. It looks looks like we still persist the new timestamp:

https://github.com/theupdateframework/python-tuf/blob/73e064186846380fac143b50657604452960e812/tuf/ngclient/updater.py#L336-L340

In L339 we ignore `data`, i.e. new ts, because old ts has the same version, but L340 takes that ignore as green light to persist `data`

lukpueh:
This PR also has an effect on the first invocation of `update_timestamp` in  this function ...

https://github.com/theupdateframework/python-tuf/blob/73e064186846380fac143b50657604452960e812/tuf/ngclient/updater.py#L328-L333

Prior to this PR, we'd make the newly loaded timestamp the trusted timestamp, if a trusted timestamp with same version was already in memory. After this PR, the one in memory remains the trusted one.

tuf/ngclient/_internal/trusted_metadata_set.py
jku: The timestamp case doesn't actually need the earlier check for anything but it does this for consistency with other methods: it's done so that Updater can 
* successfully load the local metadata for rollback checks even if metadata is not valid as final but also
* avoid fetching the remote metadata completely if local metadata is valid as final

both cases are relevant for snapshot and targets, but we don't actually need the second case for timestamp.

tests/test_trusted_metadata_set.py
MVrachev: Yes, you are correct as using assignment just creates another reference to the same object.
Will use `deepcopy` as you have suggested.

tuf/ngclient/_internal/trusted_metadata_set.py
MVrachev: Yes, you are right. I did that as I was struggling to find a short way to describe that the new_timestamp version is the same as the old one, but as you showed in the comment that's not needed as the check shows that.

tests/test_trusted_metadata_set.py
MVrachev: I just realized that if we do use deepcopy() we are creating a new object for initial_timestamp whose id is not equal to the original trusted_set.timestamp id.
This means that even if our code works perfectly fine and doesn't update the trusted_set.timestamp when the new_timestamp has the same version then the assert self.assertEqual(id(initial_timestamp), id(self.trusted_set.timestamp)) will not be true.

We need exactly a reference to an object which we create with a normal assignment because if the initial_timestamp reference points to the same object as the trusted_set.timestamp reference means that an update was not finished.

I think it should be safe to use the normal assignments as we want to verify not that the content of the initial_timestamp is not changed, but that it's actually the same object as the trusted_set.timestamp after the update.

MVrachev:
> Prior to this PR, we'd make the newly loaded timestamp the trusted timestamp, if a trusted timestamp with same version was already in memory. After this PR, the one in memory remains the trusted one.

I think that's what it was meant in the specification when at step `5.4.3` they said:
```
In case they [versions] are equal, discard the new
timestamp metadata and abort the update cycle.
```
and given that the persistence of the new timestamp happens two steps later at `5.4.5` this to me indicates that the new timestamp **should not** replace the old timestamp if they have the same version.

MVrachev:
> This PR also has an effect on the first invocation of `update_timestamp` in this function ...
> 
> https://github.com/theupdateframework/python-tuf/blob/73e064186846380fac143b50657604452960e812/tuf/ngclient/updater.py#L328-L333
> 
> Prior to this PR, we'd make the newly loaded timestamp the trusted timestamp, if a trusted timestamp with same version was already in memory. After this PR, the one in memory remains the trusted one.

I begin questioning whether we can replace the return type of `trusted_set.update_timestamp()` to `bool` instead of `Metadata[Timestamp]` and the `bool` value can be whether there was actually an update. 

I saw that currently, we don't use the return value of `trusted_set.update_timestamp()` anywhere.
I think that's the case for the rest of the `trusted_set.update_*` calls with one exception for `trusted_set.update_targets()` inside the `test_update_metadata_output` test.

lukpueh:
> > Prior to this PR, we'd make the newly loaded timestamp the trusted timestamp, if a trusted timestamp with same version was already in memory. After this PR, the one in memory remains the trusted one.
> 
> I think that's what it was meant in the specification when at step `5.4.3` they said:
> 
> ```
> In case they [versions] are equal, discard the new
> timestamp metada
> ```

No, the spec does not define this workflow. The spec paragraph you quote is about how to handle a new timestamp **downloaded from repo** with same version, handled in L336-340. But I was referring to L329-330, where old trusted timestamp is **newly loaded from local storage** and compared to old trusted timestamp in memory.

That said, I don't think this specific change makes a difference in the overall update workflow, I just wanted to point out that there was a change.

lukpueh:
> I begin questioning whether we can replace the return type of `trusted_set.update_timestamp()` to `bool` instead of `Metadata[Timestamp]` and the `bool` value can be whether there was actually an update.
> 
> I saw that currently, we don't use the return value of `trusted_set.update_timestamp()` anywhere. I think that's the case for the rest of the `trusted_set.update_*` calls with one exception for `trusted_set.update_targets()` inside the `test_update_metadata_output` test.

I think that makes sense, although I am not very fond of different return types for the `update_*` functions. The good thing is that it is an internal interface... I'll have to think about it some more.

tuf/ngclient/_internal/trusted_metadata_set.py
lukpueh: The check does save us a write operation, in case remote metadata is already expired, right? IIUC, without the check we'd persist remote metadata regardless of expiration.

MVrachev:
> > > Prior to this PR, we'd make the newly loaded timestamp the trusted timestamp, if a trusted timestamp with same version was already in memory. After this PR, the one in memory remains the trusted one.
> > 
> > 
> > I think that's what it was meant in the specification when at step `5.4.3` they said:
> > ```
> > In case they [versions] are equal, discard the new
> > timestamp metada
> > ```
> 
> No, the spec does not define this workflow. The spec paragraph you quote is about how to handle a new timestamp **downloaded from repo** with same version, handled in L336-340. But I was referring to L329-330, where old trusted timestamp is **newly loaded from local storage** and compared to old trusted timestamp in memory.
> 
> That said, I don't think this specific change makes a difference in the overall update workflow, I just wanted to point out that there was a change.

I would argue that this workflow when there is already loaded `timestamp` in memory and `update_timestamp` is called with a new data loaded from local storage doesn't exist.
The reason is that we have designed the `TrustedMetadataSet` and the `Updater` class in such a way that `Updater.referesh()` can be called exactly once.
This means that `_load_timestamp` can be called exactly once as well, meaning when calling `self._trusted_set.update_timestamp(data)` on line 330 `self.trusted_set.timestamp` will be None and on lines 336-340 a new timestamp will be downloaded possibly with the same version in which case it will happen as I have quoted.

>> I begin questioning whether we can replace the return type of trusted_set.update_timestamp() to bool instead of Metadata[Timestamp] and the bool value can be whether there was actually an update.
>>
>> I saw that currently, we don't use the return value of trusted_set.update_timestamp() anywhere. I think that's the case for the rest of the trusted_set.update_* calls with one exception for trusted_set.update_targets() inside the test_update_metadata_output test.

> I think that makes sense, although I am not very fond of different return types for the update_* functions. The good thing is that it is an internal interface... I'll have to think about it some more.

I meant that we can change all `TrustedMetadataSet.update_*` methods to return `boolean` (even if for `update_root` it doesn't make a lot of sense) and `test test_update_metadata_output` will no longer be relevant and can be removed or changed.

jku:
> I saw that currently, we don't use the return value of trusted_set.update_timestamp() anywhere.
> I think that's the case for the rest of the trusted_set.update_* calls with one exception for trusted_set.update_targets() inside the test_update_metadata_output test.

No issue with changing the return types (that's why it's internal)  but the return type was added for a reason: this way _preoder_depth_first_walk() can just call `targets = self._load_targets(...)` instead of first loading the targets, then immediately looking up the same targets from TrustedMetadataSet -- not a critical thing and it can be changed: just wanted to point out it's not unused and you will have to do something about that if you remove the return values.

I think we want to be careful about adding more "success/failure return paths" to the update_* functions -- exceptions already do that for various failure cases (and semi-failure when _check_final_*() methods raise) so we can easily end up with complex code if we also use return value for some sort of success/failure... It may be the right solution, but let's think it through really well.

jku:
So an alternative approach: 
* I think from TrustedMetadataSet POV this situation (trying to load a version it already has loaded) is an error, and should raise an error
* Updater could (in the timestamp case) handle this specific error and continue processing

In practice this could be a new error derived from BadVersionNumberError.

MVrachev:
> So an alternative approach:
> 
> * I think from TrustedMetadataSet POV this situation (trying to load a version it already has loaded) is an error, and should raise an error
> * Updater could (in the timestamp case) handle this specific error and continue processing
> 
> In practice this could be a new error derived from BadVersionNumberError.

Maybe that's the best solution not requiring any significant changes.

MVrachev:
I updated the pr by using the suggestion given by @jku.
Suggestions about the exception name are welcome.

I have checked if the new exception is raised only for the second `update_timestamp` call in `tuf/ngclient/updater._load_timestamp()` as I don't see how during the first call there could be an already loaded timestamp as I have explained in my comment here https://github.com/theupdateframework/python-tuf/pull/2024#issuecomment-1158790977

tuf/api/exceptions.py
```diff
@@ -29,6 +29,10 @@ class BadVersionNumberError(RepositoryError):
     """An error for metadata that contains an invalid version number."""
 
 
+class VerifiedVersionNumberError(RepositoryError):
+    """An error for metadata containing a previously verified version number."""
```
jku: Hmm, the error is not that it's a _verified_ number, the error is that it's equal to current version in a case where equal version is not allowed... so `EqualVersionNumberError` maybe. I think it makes sense to derive from BadVersionNumberError (the error is in practice purely internal but I think this would be clearer to users reading exceptions.py)

jku (changes requested):
Yes, thank you, I like this.

The correct functionality of Updater should be tested since we change how it works.
* the basic test would, I suppose, be just two refresh calls without repository changes: the second one should not fail since updater handles the new error. This might well be tested in an existing test already?
* a more involved test would modify the timestamp (without modifying version) in between refreshes and would check that client still uses original version after second refresh... but this is maybe a little too much work

requesting changes just for the error name. For the tests, please have a look and decide yourself if they are adequate.

tuf/api/exceptions.py
MVrachev: I used `Verified` to underline that this timestamp has been already verified and that's why it doesn't need a new version.
I don't have strong preferences and will change the error name.

MVrachev:
> Yes, thank you, I like this.
> 
> The correct functionality of Updater should be tested since we change how it works.
> 
> * the basic test would, I suppose, be just two refresh calls without repository changes: the second one should not fail since updater handles the new error. This might well be tested in an existing test already?
> * a more involved test would modify the timestamp (without modifying version) in between refreshes and would check that client still uses original version after second refresh... but this is maybe a little too much work
> 
> requesting changes just for the error name. For the tests, please have a look and decide yourself if they are adequate.

Added a test following the @jku second suggestion. 
Now the coverage should be good.

jku (approved):
Looks correct to me, thanks. @lukpueh opinions on the solution?

tests/test_trusted_metadata_set.py
lukpueh: > we want to verify not that the content of the initial_timestamp is not changed, but that it's actually the same object as the trusted_set.timestamp after the update.

Okay, in that case it makes sense to use `id`.

lukpueh (approved):

lukpueh:
Looks great. Thanks for figuring this out, @MVrachev and @jku!

PR #2031, opened 2022-06-17
Tests: download a target with succinct_roles enabled

MVrachev:
Related to issue #1909 

Add test downloading a target file when `succonct_roles` is used and as
such test the whole updater downloading workflow.

requirements**:

### DIFF
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -510,6 +510,66 @@ class TestDelegationsGraphs(TestDelegations):
         finally:
             self.teardown_subtest()
 
+    succinct_roles_targets_test_cases: utils.DataSet = {
+        "bin amount = 2, taget bin index 0": SuccinctRolesTestCase(
+            bit_length=1,
+            target_path="boo",
+            expected_target_bin="bin-0",
+        ),
+        "bin amount = 4, taget bin index 0": SuccinctRolesTestCase(
+            bit_length=2,
+            target_path="foo",
+            expected_target_bin="bin-0",
+        ),
+        "bin amount = 256, taget bin index fc": SuccinctRolesTestCase(
+            bit_length=8,
+            target_path="bar",
+            expected_target_bin="bin-fc",
+        ),
+    }
+
+    @utils.run_sub_tests_with_dataset(succinct_roles_targets_test_cases)
+    def test_download_targets_with_succinct_roles(
+        self, test_data: SuccinctRolesTestCase
+    ) -> None:
+
+        try:
+            self.setup_subtest()
+            exp_files = [*TOP_LEVEL_ROLE_NAMES, test_data.expected_target_bin]
+            exp_calls = [(test_data.expected_target_bin, 1)]
+            exp_path = os.path.join(self.targets_dir, test_data.target_path)
+
+            self.sim = RepositorySimulator()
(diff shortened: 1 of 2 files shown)


jku (approved):
So, I'm happy to merge this, it's better than what we have. However it feels like it's just the delegation graph test test_succinct_roles_graph_traversal with `download_target()` call added to the end...

Some potential ways to improve:
* avoid duplicating all delegation graph checks: those checks are complex and I feel like they harm the readability of this test. Maybe it makes sense to verify the delegated role (exp_calls) in the end but even that is debatable
* actually check that the target content is the expected one, stop using the same content for all targets
* it's very minimal WRT repository content: e.g. there is never more than one target file in a delegated targets metadata, or even the whole repository. I think we could get more real coverage if we setup a repository with succinct delegation to lots of generated targets (like maybe targets could be every two character combination as file content -- 26*26 files, and filename could be the same as file content), and then the subtests could be a dozen target downloads from that repo?

I'm marking approve, but let me know your opinion

MVrachev:
I rewrote the whole test and used a different approach.
When I decided to follow your advice and generate more target files I realized that I don't need to use the decorator at all anymore. 
That's why I instead used for loops which in this case seems like a reasonable choice.

If you don't like that approach I can rever the new commit.

tests/test_updater_delegation_graphs.py
```diff
@@ -510,6 +510,45 @@ def test_succinct_roles_graph_traversal(
...
+                # Verify that the target info was successfully found.
+                target_info = updater.get_targetinfo(f"target-{i}")
+                assert target_info is not None
+                target_full_path = updater.download_target(target_info)
+
+                # Verify that the target content is the same as the target name.
+                with open(target_full_path, encoding="utf-8") as target:
+                    target_content = target.readline()
+                    self.assertEqual(
+                        target_content, os.path.basename(target_full_path)
```
jku: the content you write is the target name `f"target-{i}"`. I think it makes more sense to compare target_content to that string then to `os.path.basename(target_full_path)`

tests/test_updater_delegation_graphs.py
```diff
@@ -510,6 +510,45 @@ def test_succinct_roles_graph_traversal(
         finally:
             self.teardown_subtest()
 
+    def test_download_targets_with_succinct_roles(self) -> None:
+        try:
+            self.setup_subtest()
```
jku: is there a point to this if you are not using subtests?

jku (approved):
I like this, very nice! 1000 files sounds like a lot but if it's not slow I'm not complaining.

Left two comments but this is great as is :+1:

jku:
> if it's not slow I'm not complaining.

looking at just results for this CI run, this seems to increase complete test suite runtime by ~50%

Please test this before writing 1000 files.

jku:
Actually (since martin may be on a holiday) I can push a commit here myself that fixes the minor issues: will merge after that

tests/test_updater_delegation_graphs.py
jku: removed this try-except

tests/test_updater_delegation_graphs.py
jku: removed os.path.basename()

MVrachev:
Thank you for pushing the changes @jku!
I was indeed on a holiday.

PR #2038, opened 2022-06-24
Add an example script about succinct roles usage

MVrachev:
Related to issue https://github.com/theupdateframework/python-tuf/issues/1909

Add a basic example script showing all features of the succinct hash bin
delegations and the available API calls of `SuccinctRoles`.

The explanations are used to promote the usage of succinct hash bin
delegations by explaining it well enough so our users can understand
the API limitations and how to use them and at the same time I tried not
going into too many details of the SuccinctRoles math as its
implementation is inside tuf/api/metadata.py and there there are
explanations about that.

requirements**:


jku:
Higher-level question (answer may be "let's handle in another issue"):
Are we happy with the level of usefulness of these hashed bin examples?

What I mean is that a basic test of repository metadata is "can I use this metadata with a client?": currently it seems like I'd have to do a lot of work to go from this example to something that actually works in a meaningful way.

As a straw man proposal: what if the `succinct_....py` script actually imported `basic_repo.py` in the beginning and just added metadata to an already complete repository using the same keys etc? Importing scripts from other scripts is bad form of course but it might give us a usable repository with minimal hassle.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,233 @@
...
+target_bin = targets.signed.delegations.succinct_roles.get_role_for_target(
+    target_path
+)
+
+# In our example with NUMBER_OF_BINS = 32 and the current file as target_path
+# the target_bin is "delegated_bin-0d"
+
+# If we want to verify that this role is indeed a delegated role (bin)
+# represented by а SuccinctRoles instance we can use is_delegated_role() API.
+assert targets.signed.delegations.succinct_roles.is_delegated_role(target_bin)
```
jku: this is not needed IMO: we just got target_bin from succinct_roles.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,233 @@
...
+PRETTY = JSONSerializer(compact=False)
+TMP_DIR = tempfile.mkdtemp(dir=os.getcwd())
+
+# Generate a key for targets we haven't added one up to this point.
+_, targets_signer = create_key()
+targets.sign(targets_signer)
+targets.to_file(os.path.join(TMP_DIR, "targets.json"), serializer=PRETTY)
+
+for bin_name, bin_target_role in delegated_bins.items():
+    file_name = f"{bin_name}.json"
```
jku: Not a bug as such but if we have a goal of making metadata that can just be served to clients, then these (and the parent metadata) should have versioned filenames.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,233 @@
...
+NAME_PREFIX = "delegated_bin"
+#
+# The suffix "bin_number" is a zero-padded hexidentical number of that
+# particular bin.
+# Zero-padded means that if NUMBER_OF_BINS is a three-digit number like 256 then
+# all bin numbers must be represented by 3 digits and if there is a bin number
+# requiring less than 3 digits an additional zeros will be padded in the
+# begging. In this example, where NUMBER_OF_BINS = 256, the first bin will have
+# a name "delegated_bin-001", the tenth bin will have a name "delegated_bin-010"
+# and the hundred bin will have the name "delegated_bin-100".
```
jku: Please double check this paragraph: e.g. I would expect 256 bins to fit into 2 hex digits, and the tenth bin is likely "delegated_bin-0a"?

Maybe removing most of this is fine too: I don't think the reader needs all the details here

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,233 @@
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+"""
+A TUF succinct hash bin delegation example using the low-level TUF Metadata API.
+
+The example code in this file demonstrates how to *manually* perform succinct
+hash bin delegation using the low-level Metadata API.
+Succinct hash bin delegation achieves a similar result as using a standard hash
+bin delegation, but it requires less memory overhead.
```
jku: less memory overhead is likely true as well but I think the advantage we're looking for is smaller delegating json files: fewer bytes to transfer and and to parse.

jku (commented):
Looks nice. It feels a bit verbose (the file is larger than hashed_bin_delegation.py even though it has less code): It might be better _example code_ if the comments were shorter and more focused "because we want X, we do Y then Z" but I also think it's perfectly fine as is.

The other comments are just opinions but please doublecheck the paragraph abnout hexadecimal in bin names.

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: I added it to showcase this method as well as in this example I am showing all of the public API and didn't want to miss something.

jku: :+1:  I'm fine with that. 

The reason I commented is that I think generally example code is best when it focuses on accomplishing a specific task (vs documentation that is best when it's comprehensive)

MVrachev:
> Higher-level question (answer may be "let's handle in another issue"): Are we happy with the level of usefulness of these hashed bin examples?
> 
> What I mean is that a basic test of repository metadata is "can I use this metadata with a client?": currently it seems like I'd have to do a lot of work to go from this example to something that actually works in a meaningful way.
> 
> As a straw man proposal: what if the `succinct_....py` script actually imported `basic_repo.py` in the beginning and just added metadata to an already complete repository using the same keys etc? Importing scripts from other scripts is bad form of course but it might give us a usable repository with minimal hassle.

My only real concern is that this script will no longer be self-sufficient. 
You have a point in that those scripts are located under `examples/repo_example` (which by the way probably should be renamed to `repo_examples`) suggesting that those scripts are actually creating repositories.

Maybe we already make a lot of references to `basic_repo.py` and we can directly utilize it.
In `hash_bin_delegation.py` we mention `basic_repo.py` 5 times, in this script we mention it 4.
@lukpueh any opinions on this as you are heavily connected with those scripts.

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: 256 in decimal is equal to 100 in hex. 
Two hex numbers can cover all numbers between 0 and 255, but you do have a point that this probably is too many details as I am explaining what is zero-padding in the `SuccinctRoles` class itself: https://github.com/theupdateframework/python-tuf/blob/8a03abfdeb5a6b6bc892859f65425be0834893a8/tuf/api/metadata.py#L1497-L1499
Will remove that part to make the document shorter.

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: Yea, still you have a point.
This shouldn't be a showdown of all `SuccinctRoles` can do (which in the future can have more API methods), but instead should be goal-focused.
Will remove that part.

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: You have a point and this is a problem inside `hash_bin_delegation.py` as well.
Will fix it there too.

MVrachev:
> Looks nice. It feels a bit verbose (the file is larger than hashed_bin_delegation.py even though it has less code): It might be better _example code_ if the comments were shorter and more focused "because we want X, we do Y then Z" but I also think it's perfectly fine as is.

I think I had to give some context about succinct hash bin delegations and didn't want just to point out steps with no explanation.

examples/repo_example/succinct_hash_bin_delegations.py
jku: If there are 256 bins then I believe the bin numbers will be 0-255

MVrachev: Yes, you are correct.
Will fix the comment in this pr directly.

tuf/api/metadata.py
```diff
@@ -1495,8 +1495,8 @@ def __init__(
         self.name_prefix = name_prefix
 
         # Calculate the suffix_len value based on the total number of bins in
-        # hex. If bit_length = 8 then number_of_bins = 256 or 100 in hex
-        # and suffix_len = 3 meaning the third bin will have a suffix of "003"
+        # hex. If bit_length = 10 then number_of_bins = 1024 or "3ff" in hex
+        # and suffix_len = 2 meaning the third bin will have a suffix of "003".
```
jku: You are correct that there's a logic mistake in the current comment (suffix len should have been 2, and the example suffix "03" to make it correct)... but your fix is not correct either (both number of bins in hex and suffix len are wrong).

MVrachev: Ohh, yea, I thought about changing `suffix_len` to 2 or changing `bit_length` to a bigger value and accidentally did both.

jku: 1024 is still not equal to 0x3ff

MVrachev: Yes, the wording wasn't correct. The last bin name was incorrect. The `bin_names` suffix will be between `000` and `3ff`.

jku:
I wonder if lukas is around next week? Could get his opinion on the general direction these scripts should be going to... but this PR looks good to me, I think we can just file more issues for other changes

MVrachev:
> I wonder if lukas is around next week? Could get his opinion on the general direction these scripts should be going to... but this PR looks good to me, I think we can just file more issues for other changes

I think he should be @lukpueh can you have a look when you have the time?

@jku are there any particular issues you want me to file?

jku:
> @jku are there any particular issues you want me to file?

well no, this is just me asking again _Are we happy with the level of usefulness of these hashed bin examples_?
Basically, do we have hopes that these examples produce usable repositories (that could be tested), or are we content that they are just snippets of code? I could choose either one, was hoping Lukas has an opinion :)

lukpueh:
> Are we happy with the level of usefulness of these hashed bin examples?

I don't have a very strong opinion. My idea was to make the scripts as simple as possible to show the most basic interaction with the Metadata API to create TUF repositories. Making them importable with re-usuable functionality will likely increase their complexity. So, I suppose it's a matter of priorities. Mine was to inspire and not to provide repository tooling. Maybe we can find a middle-way, but I would rather spend that extra designing/engineering effort on actual repository tooling, which should then also allow for simpler example scripts.

Regarding code comments,  for traditional hash bin delegation there is actually no other documentation than the example script, except for a paragraph about the [`path_hash_prefixes`](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes) field in the spec, hence the excessive commenting in the script. With succinct hash bin delegation this is a bit different, because there is TAP 15 and also a new documented management class in the metadata API.

joshuagl:
> for traditional hash bin delegation there is actually no other documentation than the example script, except for a paragraph about the [`path_hash_prefixes`](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes) field in the spec

We would like to change that https://github.com/theupdateframework/specification/issues/210

MVrachev:
> With succinct hash bin delegation this is a bit different, because there is TAP 15 and also a new documented management class in the metadata API.

@lukpueh will appreciate an opinion on where the documentation can be shortened and is too detailed.

examples/repo_example/hashed_bin_delegation.py
```diff
@@ -200,8 +198,10 @@ def find_hash_bin(path: str) -> str:
 
 # Sign and persist
 # ----------------
-# Sign all metadata and persist to temporary directory at CWD for review
-# (most notably see 'bins.json' and '80-87.json').
+# Sign all metadata and persist to temporary directory at CWD for review using
+# versioned file names. Most notably see 'bins.json' and '80-87.json'.
```
lukpueh:
```suggestion
# versioned file names. Most notably see '1.bins.json' and '1.80-87.json'.
```

examples/repo_example/hashed_bin_delegation.py
```diff
@@ -200,8 +198,10 @@ def find_hash_bin(path: str) -> str:
...
+# For more information on versioned file names see:
+# https://theupdateframework.github.io/specification/latest/#writing-consistent-snapshots
```
lukpueh: The comment below refers to the "Persist metadata" section, which has the same comment about naming that you added here. I think you can remove it here. 🤷

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+"""
+A TUF succinct hash bin delegation example using the low-level TUF Metadata API.
+
+The example code in this file demonstrates how to *manually* perform succinct
+hash bin delegation using the low-level Metadata API.
```
lukpueh: IMO succinct hash bin delegation is less "manual" work than traditional hash bin delegation, given that there is dedicated API for it.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+Succinct hash bin delegation achieves a similar result as using a standard hash
+bin delegation, but the delegated json files are smaller resulting in fewer
```
lukpueh:
```suggestion
bin delegation, but the delegating metadata is smaller resulting in fewer
```
?

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+    Key,
+    Metadata,
+    SuccinctRoles,
+    TargetFile,
+    Targets,
+)
+from tuf.api.serialization.json import JSONSerializer
+
+# Succinct hash bin delegation
+# ===================
```
lukpueh: Minor code comment formatting nit:
```suggestion
# Succinct hash bin delegation
# ============================
```

lukpueh: Same comment applies for titles below.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+# communication.
+#
+# The assignment of target files to a target's metadata is done automatically,
+# based on the byte digest of the target file name. More precisely, only a
+# prefix of the target file name byte digest is needed to assign it to the
+# correct hash bin.
+#
+# In succinct hash bin delegation all delegated roles (bins) are represented by
+# one single "succinct_roles" dictionary or when a targets file is loaded with a
+# SuccinctRoles instance from tuf/api/metadata.py
```
lukpueh: - I think *"all delegated roles (bins) are represented by one single "succinct_roles" dictionary"* could be misinterpreted. E.g. do you mean that each bin metadata is in this dictionary?
- I don't exactly know what you mean by: *"or when a targets file is loaded with a SuccinctRoles instance from tuf/api/metadata.py"*
- I don't think you need to mention ` tuf/api/metadata.py` when you talk about `SuccinctRoles`, because the class is imported into local scope. (same comment applies elsewhere).

Can we omit the whole paragraph here? The first instance of `SuccinctRoles` we create is another 100 lines away. Maybe we can talk about it only then.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+#
+# The number of bins, name prefix for all bins and key threshold are all
+# attributes that need to be configured.
+
+# Number of bins, bit length and bin number computation
+# -----------------
+# The right number of bins depends on the expected number of target files in a
+# repository. The maximum number of bins is 2 of the power of 32.
```
lukpueh: Not sure if you can say "of the power".
```suggestion
# repository. The maximum number of bins is 2 to the power of 32.
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+# For the purpose of this example we choose:
+NUMBER_OF_BINS = 32
+#
+# The bit length is the number of bits that will be used to calculate the
+# corresponding bin for each target path and it's directly related to the
```
lukpueh:
```suggestion
# corresponding bin for each target path and is directly related to the
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+# a number of bins:
+BIT_LENGTH = int(math.log2(NUMBER_OF_BINS))
+#
+# To calculate the appropriate bin number for a particular target the path to
+# that target is signed using a cryptographic algorithm (like "sha256") which
```
lukpueh:
```suggestion
# that target is hashed using a cryptographic algorithm (like "sha256") which
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+
+# Delegated role (bin) name format
+# -----------------
+# Each bin has a name in the format of f"{NAME_PREFIX}-{bin_number}".
+#
+# Name prefix is the common prefix of all delegated target roles (bins) in a
+# particular SuccinctRoles object. For our example it will be:
+NAME_PREFIX = "delegated_bin"
+#
+# The suffix "bin_number" is a zero-padded hexidentical number of that
```
lukpueh:
```suggestion
# The suffix "bin_number" is a zero-padded hexadecimal number of that
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+# particular bin.
```
lukpueh, line 72: I suggest to drastically shorten the comments until here. There are plenty of resources to refer to,
- hashed_bin_delegation.py
- `SuccinctRoles` code documentation
- TAP 15.

(The reason why `hashed_bin_delegation.py` is so verbosely commented is that none of these resources were available when I created the example script)

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+# Keys and threshold
+# -----------------
+# Given that the primary concern of succinct hash bin delegation is to reduce
+# network overhead it was decided that all bins will be signed by the same
+# set of one or more signing keys.
+#
+# Before generating the keys a decision has to be made about the number of keys
+# required for signing the bins or in other words the value of the threshold.
+# Threshold is a number with the same meaning as defined in the spec:
+# https://theupdateframework.github.io/specification/latest/#threshold
```
lukpueh: Is it worth putting so much emphasis on multiple signing keys and thresholds here? I have a feeling that it will always be 1 for any hash bin delegation.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+    return Key.from_securesystemslib_key(sslib_key), SSlibSigner(sslib_key)
+
+
+# Now, we can generate a THRESHOLD number of keys:
+succinct_roles_keys: List[Key] = []
+succinct_roles_signers: List[SSlibSigner] = []
+for i in range(0, THRESHOLD):
+    key, signer = create_key()
+    succinct_roles_keys.append(key)
+    succinct_roles_signers.append(signer)
```
lukpueh: The loop will only have one iteration. You can reduce complexity by removing the loop. And, for the sake of this example, use only one key.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+
+# Top level targets instance with succinct hash bin delegation
+# -----------------
+# NOTE: See "Targets" and "Targets delegation" paragraphs in 'basic_repo.py'
+# example for more details about the Targets object.
+#
+# Now we have all the ingredients needed to create a Targets instance using
+# succinct hash bin delegation.
+#
+# First, we would create a Targets metadata instance without any delegations.
```
lukpueh:
```suggestion
# First, we create a Targets metadata instance without any delegations.
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+
+# We would define expire as 7 days from today.
```
lukpueh:
```suggestion
# We define expire as 7 days from today.
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,226 @@
...
+expiration_date = datetime.utcnow().replace(microsecond=0) + timedelta(days=7)
+targets = Metadata(Targets(expires=expiration_date))
+
+# Then, we want to add delegations and with it information about the succinct
+# hash bin delegations which are represented by SuccinctRoles instance.
+#
+# Using succinct hash bin delegations has two restrictions:
+# 1) no other delegated roles have to be used
+# 2) only one succinct hash bin delegations can exist for one targets role
```
lukpueh:
```suggestion
# 2) only one succinct hash bin delegation can exist for one targets role
```

lukpueh (commented):
Nice work, Martin! Here's a first batch of comments. As argued somewhere inline, I think you could shorten the comments especially in the beginning of the workflow quite a bit.

Want to give it a try and I'll do another round of review?

Btw. I here are a few [dev doc writing tricks](https://mason.dev/docs/docs-devs-love.pdf) I learned at EuroPython last week. (hat tip to @masonegger)

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: I wanted to give an option to set a different threshold.
If you set a different threshold than 1 it will no longer work.
I will do as you have suggested but add a short comment that if `TRESHOLD` is changed to something larger than 1 the user should make sure to generate and sign all of the bins.

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: Yes, my mistake here.

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: Yes, will remove the paragraph as it's far from an actual `SuccinctRoles` usage.

examples/repo_example/succinct_hash_bin_delegations.py
MVrachev: I simplified it.

examples/repo_example/hashed_bin_delegation.py
MVrachev: Agree, no need to say it twice in a different way.

MVrachev:
@lukpueh I updated the pr as you had suggested.
I tried simplifying it as much as I could.

Ready for the next batch of comments.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+    SuccinctRoles,
+    TargetFile,
+    Targets,
+)
+from tuf.api.serialization.json import JSONSerializer
+
+# Succinct hash bin delegation
+# ============================
+# Succinct hash bin delegation aims to distribute a large number of target files
+# over multiple delegated targets metadata roles (bins). The consequence is a
```
lukpueh:
```suggestion
# over multiple delegated targets metadata roles (bins). The consequence is
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# attributes that need to be configured.
+
+# Number of bins, bit length and bin number computation
+# -----------------------------------------------------
+# The right number of bins depends on the expected number of target files in a
+# repository. For the purpose of this example we choose:
+NUMBER_OF_BINS = 32
+#
+# The bit length is the number of bins that will be used to calculate the bin
+# for each target path. It can be calculated directly from NUMBER_OF_BINS:
```
lukpueh: I find it hard to parse this sentence. I think I liked it better as it was before. Or what about e.g.:

"The number of bins determines how many bits in a target path are considered to assign the target to a bin."

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+
+# Keys and threshold
+# ------------------
+# Given that the primary concern of succinct hash bin delegation is to reduce
+# network overhead it was decided that all bins will be signed by the same
+# set of one or more signing keys.
+#
+# Before generating the keys a decision has to be made about the number of keys
+# required for signing the bins or in other words the value of the threshold.
+# For the purpose of this example we choose the threshold to be:
```
lukpueh: You said you simplified this paragraph? --> https://github.com/theupdateframework/python-tuf/pull/2038#discussion_r927032721 

Looks the same to me as before minus the link to the spec.

lukpueh: The "it was decided" and the "decision has to be made" sounds a bit mysterious to me

What about

```suggestion
# Succinct hash bin delegation uses the same key(s) to sign all bins. This is
# acceptable because the primary concern of this type of delegation is to reduce
# network overhead. For the purpose of this example only one key is required.
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+THRESHOLD = 1
+# Note: If THRESHOLD is changed to more than 1 the example should be modified.
```
lukpueh: I think you can omit this comment. Verification is not in the scope of this example.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# succinct hash bin delegation.
+#
+# First, we create a Targets metadata instance without any delegations.
+
+# We define expire as 7 days from today.
+expiration_date = datetime.utcnow().replace(microsecond=0) + timedelta(days=7)
+targets = Metadata(Targets(expires=expiration_date))
+
+# Then, we want to add delegations and with it information about the succinct
+# hash bin delegations which are represented by SuccinctRoles instance.
```
lukpueh:
```suggestion
# hash bin delegations, which are represented by a SuccinctRoles instance.
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+#
+# Using succinct hash bin delegations has two restrictions:
+# 1) no other delegated roles have to be used
+# 2) only one succinct hash bin delegation can exist for one targets role
```
lukpueh: I think you can omit this extra info. It is documented in TAP15 and in the metadata API. And not relevant for the example.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+    name_prefix=NAME_PREFIX,
+)
+
+# Now we will populate the keyids by using the succinct_roles_keys list.
+delegations_keys_info: Dict[str, Key] = {}
+succinct_roles.keyids.append(key.keyid)
+delegations_keys_info[key.keyid] = key
+
+# We are ready to define the Delegations instance which we will add to targets.
+# As mentioned, standard roles are not allowed together with succinct_roles.
```
lukpueh: See comment above about non-relevant extra info.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+
+# Now we can add the current target to the bin responsible for it.
+delegated_bins[target_bin].signed.targets[target_path] = target_file_info
+
+# Sign and persist
+# ----------------
+# Sign all metadata and persist to a temporary directory at CWD for review using
+# versioned file names. Most notably see 'targets.json' and
+# 'delegated_bin-0d.json'. For more information on versioned file names see:
+# https://theupdateframework.github.io/specification/latest/#writing-consistent-snapshots
```
lukpueh: https://github.com/theupdateframework/python-tuf/pull/2038#discussion_r924581694 applies here too.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# the target_bin is "delegated_bin-0d"
+
+# Now we can add the current target to the bin responsible for it.
+delegated_bins[target_bin].signed.targets[target_path] = target_file_info
+
+# Sign and persist
+# ----------------
+# Sign all metadata and persist to a temporary directory at CWD for review using
+# versioned file names. Most notably see 'targets.json' and
+# 'delegated_bin-0d.json'. For more information on versioned file names see:
```
lukpueh:
```suggestion
# versioned file names. Most notably see '1.targets.json' and
# '1.delegated_bin-0d.json'. For more information on versioned file names see:
```

?

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# https://theupdateframework.github.io/specification/latest/#writing-consistent-snapshots
+
+# NOTE: See "Persist metadata" paragraph in 'basic_repo.py' example for more
+# details about serialization formats and metadata file name convention.
+PRETTY = JSONSerializer(compact=False)
+TMP_DIR = tempfile.mkdtemp(dir=os.getcwd())
+
+# Generate a key for targets we haven't added one up to this point.
```
lukpueh: Does this phrase a conjunction? Should it be two sentences?

lukpueh (changes requested):
Thanks for the updates, @MVrachev! This has become a lot more concise. I'd appreciate a final review of a native speaker before we merge. cc @joshuagl, @jhdalek55

jhdalek55:
I'll take a look later this morning.

jhdalek55:
@lukpueh Do you wish a review just on this PR or on the TAP as a whole? If its the latter, send me the correct link?

lukpueh:
> @lukpueh Do you wish a review just on this PR or on the TAP as a whole? If its the latter, send me the correct link?

Just the new file in this PR. You can comment here: https://github.com/theupdateframework/python-tuf/pull/2038/files#diff-8b760bfeb4c14beace00509e9a55c8486510b4516e7ba9e10adee18782673616

The document is a Python program but with extensive code comments to serve as sort of tutorial. I'd appreciate if you could take a look if the code comments are sound English. No need to comment on the overall structure. I think that is fine as is. Thanks!

jhdalek55:
Ok. I'll take a look.

examples/repo_example/hashed_bin_delegation.py
```diff
@@ -200,8 +198,8 @@ def find_hash_bin(path: str) -> str:
 
 # Sign and persist
 # ----------------
-# Sign all metadata and persist to temporary directory at CWD for review
-# (most notably see 'bins.json' and '80-87.json').
+# Sign all metadata and persist to temporary directory at CWD for review using
```
jhdalek55: I'm assuming this is just field-specific terminology, but "persist to" is an awkward term in this context.

examples/repo_example/hashed_bin_delegation.py
```diff
@@ -200,8 +198,8 @@ def find_hash_bin(path: str) -> str:
...
+# versioned file names. Most notably see '1.bins.json' and '1.80-87.json'.
 
 # NOTE: See "Persist metadata" paragraph in 'basic_repo.py' example for more
 # details about serialization formats and metadata file name convention.
```
jhdalek55: co
```suggestion
# details about serialization formats and metadata file name conventions.
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+"""
+A TUF succinct hash bin delegation example using the low-level TUF Metadata API.
+
+The example code in this file demonstrates how to perform succinct hash bin
+delegation using the low-level Metadata API.
+Succinct hash bin delegation achieves a similar result as using a standard hash
+bin delegation, but the delegating metadata is smaller resulting in fewer bytes
```
jhdalek55:
```suggestion
bin delegation, but the delegating metadata is smaller, resulting in fewer bytes
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# ============================
+# Succinct hash bin delegation aims to distribute a large number of target files
+# over multiple delegated targets metadata roles (bins). The consequence is a
+# smaller metadata files and thus a lower network overhead for repository-client
+# communication.
+#
+# The assignment of target files to a target's metadata is done automatically,
+# based on the byte digest of the target file name.
+#
+# The number of the bins, name prefix for all bins and key threshold are all
```
jhdalek55:
```suggestion
# The number of bins, name prefix for all bins and key threshold are all
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# attributes that need to be configured.
+
+# Number of bins, bit length and bin number computation
+# -----------------------------------------------------
+# The right number of bins depends on the expected number of target files in a
```
jhdalek55:
```suggestion
# Determining the correct number of bins is dependent on the expected number of target files in a
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# repository. For the purpose of this example we choose:
+NUMBER_OF_BINS = 32
+#
+# The bit length is the number of bins that will be used to calculate the bin
```
jhdalek55: Slight modification of Lukas' suggestion below.

```suggestion
# The number of bins will determine the number of bits in a target path considered in assigning the target to a bin.
```

examples/repo_example/succinct_hash_bin_delegations.py
jhdalek55: See my revision above.

examples/repo_example/succinct_hash_bin_delegations.py
jhdalek55: Agree with your edits, @lukpueh. The original text was way too verbose.

examples/repo_example/succinct_hash_bin_delegations.py
jhdalek55: I agree.

examples/repo_example/succinct_hash_bin_delegations.py
jhdalek55:
```suggestion
# hash bin delegations, which are represented by SuccinctRoles instance.
```

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,201 @@
...
+# We can get all bin names for a SuccinctRoles instance with get_roles()
+delegated_bins: Dict[str, Metadata[Targets]] = {}
+for delegated_bin_name in targets.signed.delegations.succinct_roles.get_roles():
+    delegated_bins[delegated_bin_name] = Metadata(
+        Targets(expires=expiration_date)
+    )
+
+# Add target file inside a delegated role (bin)
+# ---------------------------------------------
+# For the purpose of this example we will protect the integrity of this very
```
jhdalek55: "very" is not needed.
```suggestion
# For the purpose of this example we will protect the integrity of this 
```

examples/repo_example/succinct_hash_bin_delegations.py
jhdalek55: Agreed. It's a weird phrase.
```suggestion
# Generate a key for targets that  up to this point have not received one.
```

jhdalek55 (changes requested):
Most of the changes here are minor, but needed to clarify the text.

examples/repo_example/hashed_bin_delegation.py
lukpueh: Ha. I always thought this was field terminology. But a quick web search does not yield many results. I'll change to "write"
```suggestion
# Sign all metadata and write to temporary directory at CWD for review using
```

examples/repo_example/succinct_hash_bin_delegations.py
lukpueh:
```suggestion
```

examples/repo_example/succinct_hash_bin_delegations.py
lukpueh:
```suggestion
```

examples/repo_example/succinct_hash_bin_delegations.py
lukpueh, line 130:
```suggestion
```

lukpueh:
```suggestion
```

examples/repo_example/succinct_hash_bin_delegations.py
lukpueh:
```suggestion
# versioned file names. Most notably see '1.targets.json' and
# '1.delegated_bin-0d.json'.
```

examples/repo_example/succinct_hash_bin_delegations.py
lukpueh: I suggest to move the `create_key` call up to where we created the other key. Then we don't have to explain why we all of a sudden create another key.

examples/repo_example/succinct_hash_bin_delegations.py
```diff
@@ -0,0 +1,190 @@
...
+
+def create_key() -> Tuple[Key, SSlibSigner]:
+    """Generates a new Key and Signer."""
+    sslib_key = generate_ed25519_key()
+    return Key.from_securesystemslib_key(sslib_key), SSlibSigner(sslib_key)
+
+
+key, signer = create_key()
+
+# Top level targets instance with succinct hash bin delegation
```
lukpueh: Btw. the delegator does not have to be the top-level targets role. I think we can just omit that piece of information.

lukpueh: If you don't mind, I'll also try to shorten the comments in this and the next section a bit.

lukpueh: Addressed in 50960c40761a0624f7f2c828135cd50d9c875ee6

lukpueh:
@MVrachev, I took the liberty the push two commits, addressing my and @jhdalek55's comments. Feel free to revert if you don't agree. Otherwise, I'm fine with this PR. Can we maybe get another set of eyes? This PR is the last blocker for our next release.

cc @joshuagl, @jku

jku (approved):
Looks good to me. 

As a nitpick, the succinct role creation would likely be a little neater and safer if you don't try to insert the keys during object construction (and use `add_key()` afterwards instead). I filed #2071 to improve that area anyway

lukpueh (approved):

MVrachev:
I don't mind the changes you added.
I am not a native speaker and really appreciate the help of @jhdalek55 and all of the reviews.

PR #2039, opened 2022-06-24
SuccinctRoles.is_delegated_role() add a test case

MVrachev:

Add a test case when there is a bin name with the desired prefix, but
which cannot be cast to a hexadecimal number.

requirements**:


jku (approved):
Looks good. 

Looking at these two cases:

        "bin-s",
        "bin-0t",

is the difference that first one fails because of length already and second one fails at hex parsing stage?

MVrachev:
> Looks good.
> 
> Looking at these two cases:
> 
> ```
>     "bin-s",
>     "bin-0t",
> ```
> 
> is the difference that first one fails because of length already and second one fails at hex parsing stage?

Yes, exactly that is the difference.

PR #2048, opened 2022-07-05
ngclient: Prevent automatic decoding of gzip files

### LINKED ISSUE #2047: DownloadLengthMismatchError because requests automatically decodes gzip files

If target is a gzip file, served with a `Content-Encoding: gzip` header, `requests` will automatically decode the file.

**Current behavior**:

As a result, the [default fetcher][2] for tuf ngclient raises a `DownloadLengthMismatchError`:

```
tuf.api.exceptions.DownloadLengthMismatchError: Downloaded x bytes exceeding the maximum allowed length of y
```

**Expected behavior**:

From the [requests docs][1]:

>An important note about using `Response.iter_content` versus `Response.raw`. `Response.iter_content` will automatically decode the `gzip` and `deflate` transfer-encodings. `Response.raw` is a raw stream of bytes – it does not transform the response content. If you really need access to the bytes as they were returned, use `Response.raw`.

I would expect tuf to access the `Reponse.raw` instead of `Response.iter_content` here:

https://github.com/theupdateframework/python-tuf/blob/b5e952cee3edb1c3363a8a7859cb548b13ab71f3/tuf/ngclient/_internal/requests_fetcher.py#L99

Note that `Response.raw` is a [urllib3.response.HTTPResponse][3], which also decodes `gzip` encoded content by default, but `requests` [disables this][4] by setting `decode_content=False`.

[1]: https://requests.readthedocs.io/en/latest/user/quickstart/#raw-response-content
[2]: https://github.com/theupdateframework/python-tuf/blob/b5e952cee3edb1c3363a8a7859cb548b13ab71f3/tuf/ngclient/_internal/requests_fetcher.py#L77
[3]: https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html#urllib3.response.HTTPResponse
[4]: https://github.com/psf/requests/blob/bda7f0171f8bba17989d3a2c28dfa9a9261b1b65/requests/adapters.py#L497

dennisvang:
Fixes #2047

- Replace `response.iter_content` by `response.raw` in `RequestsFetcher._chunks()`. This prevents requests/urllib3 from automatically decoding files with `Content-Encoding` headers (`gzip` or `deflate`).
- Add a `CustomHTTPRequestHandler` to `tests.utils`, and use this in `simple_server.py` (instead of `SimpleHTTPRequestHandler`). The custom request handler adds the ability to specify additional headers in an HTTP request, which are then added to the corresponding HTTP response. This is used to test the "Content-Encoding" issue.
- Add a regression test for issue 2047 which verifies that files served with a "Content-Encoding: gzip" header are not decoded.

Some open questions:

- Is it sufficient to catch `ReadTimeoutError` in `_chunks()`, or should we catch e.g. [IncompleteRead][1] too?

requirements**:

[1]: https://github.com/urllib3/urllib3/blob/43c372f6eb9f9f94848c7b7645ad19ebf6c047c5/src/urllib3/response.py#L734

### DIFF
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -17,6 +17,7 @@ from typing import Any, ClassVar, Iterator
 from unittest.mock import Mock, patch
 
 import requests
+from requests.models import ReadTimeoutError  # actually from urllib3
 
 from tests import utils
 from tuf.api import exceptions
@@ -109,18 +110,20 @@ class TestFetcher(unittest.TestCase):
     # Response read timeout error
     @patch.object(requests.Session, "get")
     def test_response_read_timeout(self, mock_session_get: Any) -> None:
-        mock_response = Mock()
+        # from urllib3.connectionpool import ConnectionPool
+        # dummy_pool = ConnectionPool("dummy")
+        mock_response = Mock(raw=Mock())
         attr = {
-            "iter_content.side_effect": requests.exceptions.ConnectionError(
-                "Simulated timeout"
+            "read.side_effect": ReadTimeoutError(
+                None, None, "Simulated timeout"
             )
         }
-        mock_response.configure_mock(**attr)
+        mock_response.raw.configure_mock(**attr)
         mock_session_get.return_value = mock_response
 
         with self.assertRaises(exceptions.SlowRetrievalError):
             next(self.fetcher.fetch(self.url))
-        mock_response.iter_content.assert_called_once()
+        mock_response.raw.read.assert_called_once()
 
     # Read/connect session timeout error
     @patch.object(
@@ -170,6 +173,14 @@ class TestFetcher(unittest.TestCase):
             # Force download_file to execute and raise the error since it is a
             # context manager and returns Iterator[IO]
             yield self.fetcher.download_file(self.url, self.file_length - 4)
+
+    # Download a file with Content-Encoding gzip (or deflate)
+    def test_download_file_content_encoding(self):
+        # todo:
+        #  - enable custom headers in simple_server.py, so we can return a
+        #    file with Content-Encoding header
+        #  - serve a gzipped file (in TestFetcher.setUpClass?)
+        pass
 
 
 # Run unit test.
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -114,7 +114,7 @@ class TestFetcher(unittest.TestCase):
         content_encoding_header = json.dumps({"Content-Encoding": "gzip"})
         headers = {utils.REQUEST_RESPONSE_HEADERS: content_encoding_header}
         get_with_headers = partialmethod(requests.Session.get, headers=headers)
-        target = 'tuf.ngclient._internal.requests_fetcher.requests.Session.get'
+        target = "tuf.ngclient._internal.requests_fetcher.requests.Session.get"
         # The test file content does not represent a real gzip file,
         # so we can expect an error to be raised if requests/urllib3 tries to
         # decode the file (urllib3 uses zlib for this).
@@ -126,7 +126,7 @@ class TestFetcher(unittest.TestCase):
                 temp_file.seek(0)
                 fetched_data = temp_file.read()
         except requests.exceptions.ContentDecodingError as e:
-            self.fail(f'fetch() raised unexpected decoding error: {e}')
+            self.fail(f"fetch() raised unexpected decoding error: {e}")
         # If all is well, decoding has *not* been attempted, and the fetched
         # data matches the original file contents
         self.assertEqual(self.file_contents, fetched_data)
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -6,7 +6,6 @@
 """Unit test for RequestsFetcher.
 """
 
-from functools import partialmethod
 import io
 import json
 import logging
@@ -15,6 +14,7 @@ import os
 import sys
 import tempfile
 import unittest
+from functools import partialmethod
 from typing import Any, ClassVar, Iterator
 from unittest.mock import Mock, patch
 
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -6,7 +6,9 @@
 """Unit test for RequestsFetcher.
 """
 
+from functools import partialmethod
 import io
+import json
 import logging
 import math
 import os
@@ -95,6 +97,40 @@ class TestFetcher(unittest.TestCase):
             # Check that we calculate chunks as expected
             self.assertEqual(chunks_count, expected_chunks_count)
 
+    # Fetch data with Content-Encoding gzip (or deflate)
+    def test_fetch_content_encoding(self):
+        """
+        Regression test for issue #2047
+
+        By default, requests/urllib3 will automatically try to decode data
+        that is served with a "Content-Encoding: gzip" header (or "deflate").
+        As the length of the decoded data typically would be different from
+        the expected number of bytes specified in the TUF targets metadata,
+        this would give rise to a DownloadLengthMismatchError later on. Thus,
+        we need to ensure that a file served with the "Content-Encoding"
+        header is *not* decoded by the RequestsFetcher.
+        """
+        # Serve dummy file with "Content-Encoding: gzip" header
+        content_encoding_header = json.dumps({"Content-Encoding": "gzip"})
+        headers = {utils.REQUEST_RESPONSE_HEADERS: content_encoding_header}
+        get_with_headers = partialmethod(requests.Session.get, headers=headers)
+        target = 'tuf.ngclient._internal.requests_fetcher.requests.Session.get'
+        # The test file content does not represent a real gzip file,
+        # so we can expect an error to be raised if requests/urllib3 tries to
+        # decode the file (urllib3 uses zlib for this).
+        try:
+            with tempfile.TemporaryFile() as temp_file:
+                with patch(target, get_with_headers):
+                    for chunk in self.fetcher.fetch(self.url):
+                        temp_file.write(chunk)
+                temp_file.seek(0)
+                fetched_data = temp_file.read()
+        except requests.exceptions.ContentDecodingError as e:
+            self.fail(f'fetch() raised unexpected decoding error: {e}')
+        # If all is well, decoding has *not* been attempted, and the fetched
+        # data matches the original file contents
+        self.assertEqual(self.file_contents, fetched_data)
+
     # Incorrect URL parsing
     def test_url_parsing(self) -> None:
         with self.assertRaises(exceptions.DownloadError):
@@ -171,14 +207,6 @@ class TestFetcher(unittest.TestCase):
             # Force download_file to execute and raise the error since it is a
             # context manager and returns Iterator[IO]
             yield self.fetcher.download_file(self.url, self.file_length - 4)
-
-    # Download a file with Content-Encoding gzip (or deflate)
-    def test_download_file_content_encoding(self):
-        # todo:
-        #  - enable custom headers in simple_server.py, so we can return a
-        #    file with Content-Encoding header
-        #  - serve a gzipped file (in TestFetcher.setUpClass?)
-        pass
 
 
 # Run unit test.
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -110,8 +110,6 @@ class TestFetcher(unittest.TestCase):
     # Response read timeout error
     @patch.object(requests.Session, "get")
     def test_response_read_timeout(self, mock_session_get: Any) -> None:
-        # from urllib3.connectionpool import ConnectionPool
-        # dummy_pool = ConnectionPool("dummy")
         mock_response = Mock(raw=Mock())
         attr = {
             "read.side_effect": ReadTimeoutError(
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -112,7 +112,7 @@ class TestFetcher(unittest.TestCase):
         """
         # Serve dummy file with "Content-Encoding: gzip" header
         content_encoding_header = json.dumps({"Content-Encoding": "gzip"})
-        headers = {utils.REQUEST_RESPONSE_HEADERS: content_encoding_header}
+        headers = {utils.DESIRED_RESPONSE_HEADERS: content_encoding_header}
         get_with_headers = partialmethod(requests.Session.get, headers=headers)
         target = "tuf.ngclient._internal.requests_fetcher.requests.Session.get"
         # The test file content does not represent a real gzip file,
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -17,7 +17,7 @@ from typing import Any, ClassVar, Iterator
 from unittest.mock import Mock, patch
 
 import requests
-from requests.models import ReadTimeoutError  # actually from urllib3
+from requests.adapters import ReadTimeoutError  # this is a urllib3 exception
 
 from tests import utils
 from tuf.api import exceptions
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -19,7 +19,7 @@ from typing import Any, ClassVar, Iterator
 from unittest.mock import Mock, patch
 
 import requests
-from requests.adapters import ReadTimeoutError  # this is a urllib3 exception
+from urllib3.exceptions import ReadTimeoutError
 
 from tests import utils
 from tuf.api import exceptions
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -98,7 +98,7 @@ class TestFetcher(unittest.TestCase):
             self.assertEqual(chunks_count, expected_chunks_count)
 
     # Fetch data with Content-Encoding gzip (or deflate)
-    def test_fetch_content_encoding(self):
+    def test_fetch_content_encoding(self) -> None:
         """
         Regression test for issue #2047
 
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -48,9 +48,10 @@ TEST_HOST_ADDRESS = "127.0.0.1"
 # DataSet is only here so type hints can be used.
 DataSet = Dict[str, Any]
 
-# May be used in a request to add headers to the response that is returned by
-# the customized http request handler.
-REQUEST_RESPONSE_HEADERS = "Request-Response-Headers"
+# If the test-server receives an HTTP-request with this custom header,
+# it will add the specified headers to the HTTP-response. The corresponding
+# value must be a JSON object with header name/value pairs.
+DESIRED_RESPONSE_HEADERS = "Desired-Response-Headers"
 
 
 # Test runner decorator: Runs the test as a set of N SubTests,
@@ -375,11 +376,11 @@ class TestServerProcess:
 
 class CustomHTTPRequestHandler(SimpleHTTPRequestHandler):
     def end_headers(self) -> None:
-        # If REQUEST_RESPONSE_HEADERS is found in the request headers, add the
+        # If DESIRED_RESPONSE_HEADERS is found in the request headers, add the
         # specified headers to the response.
-        requested_headers = self.headers.get(REQUEST_RESPONSE_HEADERS)
-        if requested_headers:
-            headers = json.loads(requested_headers)
+        desired_headers = self.headers.get(DESIRED_RESPONSE_HEADERS)
+        if desired_headers:
+            headers = json.loads(desired_headers)
             for key, value in headers.items():
                 self.send_header(keyword=key, value=value)
         super().end_headers()
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -22,7 +22,6 @@
 
 import argparse
 import errno
-from http.server import SimpleHTTPRequestHandler
 import json
 import logging
 import os
@@ -35,6 +34,7 @@ import time
 import unittest
 import warnings
 from contextlib import contextmanager
+from http.server import SimpleHTTPRequestHandler
 from typing import IO, Any, Callable, Dict, Iterator, List, Optional
 
 logger = logging.getLogger(__name__)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -375,7 +375,7 @@ class TestServerProcess:
 
 class CustomHTTPRequestHandler(SimpleHTTPRequestHandler):
     def end_headers(self) -> None:
-        # If Custom-Response-Headers is found in the request headers, add the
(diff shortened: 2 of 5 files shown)


tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -19,6 +20,7 @@
 # Globals
 logger = logging.getLogger(__name__)
 
+
```
MVrachev, line 23: I am not sure we need that.

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -11,6 +11,7 @@
 
 # Imports
 import requests
+from requests.models import ReadTimeoutError  # todo: import directly from urllib3?
```
MVrachev: I agree with you that it makes sense to directly import it from `urlib3` given that `requests.models` just reexports it.

tests/test_fetcher_ng.py
```diff
@@ -17,7 +17,7 @@
 from unittest.mock import Mock, patch
 
 import requests
-from requests.models import ReadTimeoutError  # actually from urllib3
+from requests.adapters import ReadTimeoutError  # this is a urllib3 exception
```
MVrachev: We already depend on urlib3:
https://github.com/theupdateframework/python-tuf/blob/c12be4857d5388eb8e3c2e692fa6cce4f95cff35/requirements-pinned.txt#L10
so I don't think you should worry about this.
Appreciate the thoughtfulness. 👍

tests/utils.py
```diff
@@ -46,6 +48,11 @@
 # DataSet is only here so type hints can be used.
 DataSet = Dict[str, Any]
 
+# May be used in a request to add headers to the response that is returned by
+# the customized http request handler.
+REQUEST_RESPONSE_HEADERS = "Request-Response-Headers"
```
MVrachev: I saw that later on, you set `REQUEST_RESPONSE_HEADERS` in your test in `test_fetcher.py` in order to test the specific situation.
Can you instead of setting a global constant during the test, create an `__init__` for `CustomHTTPRequestHandler` and set the information you need there in  a variable?
Then you can use that variable in `CustomHTTPRequestHandler.end_headers()` instead of using the global `REQUEST_RESPONSE_HEADERS`.

tests/test_fetcher_ng.py
```diff
@@ -94,6 +97,40 @@ def test_fetch_in_chunks(self) -> None:
...
+        As the length of the decoded data typically would be different from
+        the expected number of bytes specified in the TUF targets metadata,
+        this would give rise to a DownloadLengthMismatchError later on. Thus,
+        we need to ensure that a file served with the "Content-Encoding"
+        header is *not* decoded by the RequestsFetcher.
+        """
+        # Serve dummy file with "Content-Encoding: gzip" header
+        content_encoding_header = json.dumps({"Content-Encoding": "gzip"})
+        headers = {utils.REQUEST_RESPONSE_HEADERS: content_encoding_header}
+        get_with_headers = partialmethod(requests.Session.get, headers=headers)
```
MVrachev, line 116: As I am new to `functools` and `partialmethod` what is the bonus of creating a `partialmethod` object instead of just calling `response.Session.get` directly in line 123 with the headers you want?

tests/test_fetcher_ng.py
```diff
@@ -94,6 +97,40 @@ def test_fetch_in_chunks(self) -> None:
...
+        target = "tuf.ngclient._internal.requests_fetcher.requests.Session.get"
+        # The test file content does not represent a real gzip file,
+        # so we can expect an error to be raised if requests/urllib3 tries to
+        # decode the file (urllib3 uses zlib for this).
```
MVrachev, line 120: Have you tried running this script before replacing `response.iter_content` with `response.raw`? 
Did `request` raise an error with this test then?

MVrachev (changes requested):
Thanks, @dennisvang for your efforts!
I really liked how you approached the problem, that you give a detailed description, suggested a solution with links to support it, and added a regression test to make sure it doesn't occur again!

Maybe it's obvious, but I want to say it out loud and confirm I understand it: you created the `CustomHTTPRequestHandler` helper class in `utils.py` in order to support adding that `Content-Encoding` header you need for the test.
I am not sure where exactly do you envoke `CustomHTTPRequestHandler`? I am sure I am missing something and I wanted to clarify it.

Also, there are linting errors shown in the CI run, will appreciate it if you fix them.

PS: Sorry that it took us so long to give you feedback on your pr.
I will not be coding or reviewing the upcoming week, so if you update the pr or comment on something I will see it after a week or more.

tuf/ngclient/_internal/requests_fetcher.py
dennisvang: Sorry, I got a PEP8 warning there: "expecting 2 blank lines"

tests/test_fetcher_ng.py
dennisvang: Yes, I did try that, and I can confirm that the error was raised.

tests/test_fetcher_ng.py
dennisvang: Calling `response.Session.get` directly with the `headers` argument would certainly simplify things, but, in this case, I wanted to test `RequestsFetcher.fetch`. This implies that `response.Session.get` is called indirectly, viz. through [RequestsFetcher._fetch][1].

So, instead of calling `response.Session.get` directly, I patch it to make sure it is called with the desired `headers` argument. Basically, I use `functools.partialmethod` to supply a default value for `headers`. Now, when `RequestsFetcher._fetch` calls `get`, the `headers` arg is automatically set.

I am aware [`patch`][2] also accepts additional kwargs, but I believe these are only applied if `patch` creates a `Mock` object.

If there is a simpler way to do this, I am all for it. :)

[1]: https://github.com/theupdateframework/python-tuf/blob/c12be4857d5388eb8e3c2e692fa6cce4f95cff35/tuf/ngclient/_internal/requests_fetcher.py#L77
[2]: https://docs.python.org/3/library/unittest.mock.html#patch

tuf/ngclient/_internal/requests_fetcher.py
dennisvang: I'll fix that.

dennisvang:
@MVrachev  Thanks for the review. :)

> Maybe it's obvious, but I want to say it out loud and confirm I understand it: you created the `CustomHTTPRequestHandler` helper class in `utils.py` in order to support adding that `Content-Encoding` header you need for the test. 

Yes, that is correct. This seemed like the least intrusive way to get a response with the desired header. It does not affect the existing tests, as far as I know.

A much simpler alternative would be to bypass the test-server and just send a request to e.g. https://httpbin.org/gzip. However, that's an external dependency, and would require an actual internet connection.

Another alternative, perhaps, would be to patch `requests` or `urrlib3` (in the test), in order to modify the response at some appropriate point. However, I have a feeling that could also get quite complex.

> I am not sure where exactly do you envoke `CustomHTTPRequestHandler`? I am sure I am missing something and I wanted to clarify it.

The handler class is passed into the `socketserver.TCPServer` here:

https://github.com/theupdateframework/python-tuf/blob/5a572d812fe32bda6787b05712f6319341d610a1/tests/simple_server.py#L16

And this is used in the `TestServerProcess` class: 

https://github.com/theupdateframework/python-tuf/blob/c12be4857d5388eb8e3c2e692fa6cce4f95cff35/tests/utils.py#L185

tests/utils.py
dennisvang: Perhaps I'm misinterpreting your question, but I don't think passing the header into `CustomHTTPRequestHandler.__init__` would be an option, as the `CustomHTTPRequestHandler` is instantiated by `socketserver.TCPServer` (see [python source][1]):

https://github.com/theupdateframework/python-tuf/blob/5a572d812fe32bda6787b05712f6319341d610a1/tests/simple_server.py#L16

I could make `REQUEST_RESPONSE_HEADERS` a class attribute of `CustomHTTPRequestHandler`, instead of a global. Would that be preferable?

[1]: https://github.com/python/cpython/blob/81dca70d704d0834d8c30580e648a973250b2973/Lib/socketserver.py#L361

tuf/ngclient/_internal/requests_fetcher.py
joshuagl: I prefer it the original way (importing from requests), that way our requests_fetcher module only depends on the requests interface and it doesn't matter what underlying connectivity API requests uses. With the change to importing from urllib3 directly, we now have an explicit dependency on urllib3.

I'd like to hear from a more active maintainer than me on this, though – @lukpueh or @jku any thoughts?

dennisvang: @joshuagl  That was my original feeling also, as mentioned in https://github.com/theupdateframework/python-tuf/pull/2048/commits/eeba7c74add818f5e71f42ee9b5b1142ef2bf90e

On the other hand, my fix relies on `requests.Response.raw`, which [is in fact][1] a `urllib3.response.HTTPResponse`. So it actually relies on (part of) the urllib3 response interface as well, viz. the `read()` method.

[1]: https://github.com/psf/requests/blob/bda7f0171f8bba17989d3a2c28dfa9a9261b1b65/requests/adapters.py#L489

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -96,12 +98,16 @@ def _chunks(self, response: "requests.Response") -> Iterator[bytes]:
         download."""
 
         try:
-            for data in response.iter_content(self.chunk_size):
+            while True:
```
dennisvang, line 101: Alternatively, we could perhaps replace this `while` loop by the following:

```python
yield from response.raw.stream(amt=self.chunk_size, decode_content=False)
```

See `urrlib3` source: https://github.com/urllib3/urllib3/blob/1f2f6a5817fd420d65e280e671d2b579791aef09/src/urllib3/response.py#L748

tuf/ngclient/_internal/requests_fetcher.py
lukpueh: I'd also import the error from `requests` for said reasons. 

And when using `requests.Response.raw`, we do still rely on [documented](https://requests.readthedocs.io/en/latest/user/quickstart/#raw-response-content) `requests` API, and we should expect that they will make an effort to not break it.

dennisvang: @lukpueh and @joshuagl Allright, sounds good. I'll revert to `from requests.adapters import ...`.

tuf/ngclient/_internal/requests_fetcher.py
dennisvang: @lukpueh @joshuagl @MVrachev What do you think about this?

lukpueh:
It looks like this PR reverts https://github.com/theupdateframework/python-tuf/pull/1774. Did you happen to read the rationale behind switching from `raw` to `iter_content` presented in b8759a99378a5bdc20e78ec6d56e539bc7998595?

dennisvang:
> It looks like this PR reverts #1774. Did you happen to read the rationale behind switching from `raw` to `iter_content` presented in [b8759a9](https://github.com/theupdateframework/python-tuf/commit/b8759a99378a5bdc20e78ec6d56e539bc7998595)?

@lukpueh  No, I was not aware of those. Thanks for pointing them out. I'll have a look.

dennisvang:
@lukpueh You are right, the `_chunks` part does indeed look like a reversion of https://github.com/theupdateframework/python-tuf/commit/b8759a99378a5bdc20e78ec6d56e539bc7998595, almost down to the smallest detail. 

However, I do wonder: Have the changes introduced in https://github.com/theupdateframework/python-tuf/commit/b8759a99378a5bdc20e78ec6d56e539bc7998595 ever been tested with actual compressed files? Issue #2047 appears to suggest otherwise. Or perhaps it was tested with metadata files but not with target files?

lukpueh:
> However, I do wonder: Has the iter_content implementation introduced in https://github.com/theupdateframework/python-tuf/commit/b8759a99378a5bdc20e78ec6d56e539bc7998595 ever been tested with actual compressed files? Issue https://github.com/theupdateframework/python-tuf/issues/2047 appears to suggest otherwise. Or perhaps it was tested with metadata files but not with target files?

I wonder too. :) At least #1774 does not seem to add any tests with compressed target files. Let me look, I would think that @jku tested this.

dennisvang:
@lukpueh 

Perhaps it *was* only intended for metadata, judging from @jku's https://github.com/theupdateframework/python-tuf/issues/1251#issuecomment-754550336:

> ...
>I lean to the same solution... but I'll still document the alternative though (sechkova already said this but just for emphasis): allow compression *only for metadata*. ...

I suppose #2047 does not arise for *metadata* (although I haven't tested this), because the number of downloaded bytes is compared with the maximum allowed metadata file size from `config`, instead of the actual expected file size, as can be seen e.g. here:

https://github.com/theupdateframework/python-tuf/blob/e07fa2776283ebafc8fbab4cd386aee0a971fc13/tuf/ngclient/updater.py#L312-L314

unless an optional file `length` is included in the snapshot and/or timestamp metadata, as can be seen here:

https://github.com/theupdateframework/python-tuf/blob/e07fa2776283ebafc8fbab4cd386aee0a971fc13/tuf/ngclient/updater.py#L389-L394

lukpueh:
Looks like @jku's *"Minimal TUF Playground Client"* (see https://github.com/jku/repository-playground/pull/16) is able to successfully download a `*.tar.gz`-target:

```bash
git clone git@github.com:jku/repository-playground.git -b tuf-minimal
cd repository-playground/playground/client/
pip install tuf click
python playground_client.py download tuf-spec/tarball
# Downloads https://jku.github.io/playground-tuf-minimal/repository/content/tuf-spec/tarball/1.0.30/648dc8c3be0a7de08d654e38556349580926b3e024e22b4f2828eab8895b71c5.specification-1.0.30.tar.gz

```

lukpueh:
In above example the server serves metadata as `'application/json'` and gzipped target file as `'application/gzip'`. And only sends the `'Content-Encoding': 'gzip'` header for the metadata but not for the already gzipped target file. This seems like the correct behavior for a server.


See debugger output for details:
```python
> /Users/lukp/.pyenv/versions/min-tuf/lib/python3.8/site-packages/tuf/ngclient/_internal/requests_fetcher.py(92)_fetch()
-> return self._chunks(response)
(Pdb) session.headers
{'User-Agent': 'tuf/1.1.0 python-requests/2.28.1', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive'}
(Pdb) url
'https://jku.github.io/playground-tuf-minimal/repository//metadata/timestamp.json'
(Pdb) response.headers
{'Connection': 'keep-alive', 'Content-Length': '328', 'Server': 'GitHub.com', 'Content-Type': 'application/json; charset=utf-8', 'permissions-policy': 'interest-cohort=()', 'Last-Modified': 'Thu, 30 Jun 2022 06:16:55 GMT', 'Access-Control-Allow-Origin': '*', 'ETag': 'W/"62bd3fd7-1dc"', 'expires': 'Tue, 19 Jul 2022 12:51:55 GMT', 'Cache-Control': 'max-age=600', 'Content-Encoding': 'gzip', 'x-proxy-cache': 'MISS', 'X-GitHub-Request-Id': '3358:A2C4:8A7C6E:9658CB:62D6A693', 'Accept-Ranges': 'bytes', 'Date': 'Tue, 19 Jul 2022 14:01:26 GMT', 'Via': '1.1 varnish', 'Age': '373', 'X-Served-By': 'cache-muc13969-MUC', 'X-Cache': 'HIT', 'X-Cache-Hits': '1', 'X-Timer': 'S1658239286.277420,VS0,VE0', 'Vary': 'Accept-Encoding', 'X-Fastly-Request-ID': '70fb24cb5491e50d1eb5ff8f12a09d4468b7f150'}
(Pdb) c
> /Users/lukp/.pyenv/versions/min-tuf/lib/python3.8/site-packages/tuf/ngclient/_internal/requests_fetcher.py(92)_fetch()
-> return self._chunks(response)
(Pdb) response.headers
{'Connection': 'keep-alive', 'Content-Length': '49485', 'Server': 'GitHub.com', 'Content-Type': 'application/gzip', 'permissions-policy': 'interest-cohort=()', 'x-origin-cache': 'HIT', 'Last-Modified': 'Thu, 30 Jun 2022 06:16:55 GMT', 'Access-Control-Allow-Origin': '*', 'ETag': '"62bd3fd7-c14d"', 'expires': 'Tue, 19 Jul 2022 12:52:17 GMT', 'Cache-Control': 'max-age=600', 'x-proxy-cache': 'MISS', 'X-GitHub-Request-Id': 'EC3A:237B:BA3A:C458:62D6A6A9', 'Accept-Ranges': 'bytes', 'Date': 'Tue, 19 Jul 2022 14:02:08 GMT', 'Via': '1.1 varnish', 'Age': '386', 'X-Served-By': 'cache-muc13969-MUC', 'X-Cache': 'HIT', 'X-Cache-Hits': '1', 'X-Timer': 'S1658239329.841205,VS0,VE1', 'Vary': 'Accept-Encoding', 'X-Fastly-Request-ID': '7dee8b09d77bb81445bc41c771514782e89960d8'}
```

dennisvang:
> In above example the server serves metadata as `'application/json'` and gzipped target file as `'application/gzip'`. And only sends the `'Content-Encoding': 'gzip'` header for the metadata but not for the already gzipped target file. This seems like the correct behavior for a server.

@lukpueh You are right, that sounds like what is described in [rfc7231][1] and [here][2]:

>Note that the original media/content type is specified in the `Content-Type` header, and that the `Content-Encoding` applies to the representation, or "coded form", of the data. If the original media is encoded in some way (e.g. a zip file) then this information would *not* be included in the `Content-Encoding` header.

[1]: https://httpwg.org/specs/rfc7231.html#header.content-encoding
[2]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

dennisvang:
@lukpueh  So, that would mean the current implementation is correct, and #2047 is not a bug, but is probably caused by an incorrect server configuration. Which would imply I've been leading everyone on a wild goose chase...

lukpueh:
> @lukpueh So, that would mean the current implementation is correct, and https://github.com/theupdateframework/python-tuf/issues/2047 is not a bug, but is probably caused by an incorrect server configuration.

I think it is. Your effort is still appreciated and I apologize for not having picked up your issue earlier! 

(also 💯 for [verbose commit messages](https://github.com/theupdateframework/python-tuf/pull/1774#discussion_r789522544))

dennisvang:
@lukpueh No problem. I should have done a much more thorough searched of old issues. https://github.com/theupdateframework/python-tuf/pull/1774#discussion_r789522544 does indeed describe what is going wrong on my side.

For what it's worth:

#2047 occurred on my side when downloading a `.tar.gz` file from two different servers:

1. a local [Django development server][3]

   It turns out the Django development server guesses both the `Content-Type` and the `Content-Encoding` using [mimetypes.guess_type][1], as follows ([source][2]):

   ```python
   content_type, encoding = mimetypes.guess_type(str(fullpath))
   ```
   For a `.tar.gz` file, this finally results in a combination of `Content-Type: application/x-tar` and `Content-Encoding: gzip`, instead of the `Content-Type: application/gzip` (and no content-encoding) that would be expected according to [rfc7231][4].

2. direct download from an AWS S3 bucket

   I'm still looking into what is going wrong with the AWS S3 case, but I suspect it has to do with the arguments supplied to [`PutObject`][5] in order to upload the file to the S3 bucket.

   EDIT: see [update][6] below


[1]: https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type
[2]: https://github.com/django/django/blob/70147215bb848539273b42617b32773013bcdc47/django/views/static.py#L47
[3]: https://docs.djangoproject.com/en/stable/ref/django-admin/#runserver
[4]: https://httpwg.org/specs/rfc7231.html#header.content-encoding
[5]: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestParameters
[6]: https://github.com/theupdateframework/python-tuf/pull/2048#issuecomment-1190394156

lukpueh:
Good to know. Django folks are pretty explicit about not wanting to improve their dev web server. But AWS S3 bucket seems worth troubleshooting. Out of curiosity do you have that setup open-sourced?

lukpueh:
Either way, I'd say we close this PR...

dennisvang:
@lukpueh  

> ... But AWS S3 bucket seems worth troubleshooting. Out of curiosity do you have that setup open-sourced?

TLDR: It looks like the issue is caused by `S3Boto3Storage` from [`django-storages`][1].

### Details

The test setup uses Django's [`FileField`][6] with `S3Boto3Storage` from [`django-storages`][1] to upload `.tar.gz` files to an S3 bucket with default settings.

As it turns out, `django-storages` uses the same approach as the [django dev server][4] to guess the content-type and content-encoding for a file, before uploading that file to an S3 bucket. This happens in [`S3Boto3Storage._get_write_parameters`][2]:

```python
_type, encoding = mimetypes.guess_type(name)
```

The resulting `Content-Type` and `Content-Encoding` headers are finally passed along to the [S3 API][5] in a `PutObject` call in [`s3transfer.upload.PutObjectTask`][3].

The `Content-Encoding: gzip` header then ends up in the [S3 object metadata][7], so it is included in the response headers when the object is downloaded from S3.

And that's where #2047 arises.

### Note

I also tried creating a fresh S3 bucket with default settings, and uploading a `.tar.gz` file manually, using the AWS S3 web-console (so without `django-storages`). As expected, in this case the download response headers only contained the `Content-Type: application/x-gzip` header, without a `Content-Encoding` header.

[1]: https://github.com/jschneier/django-storages/blob/7dd36fb820b796ec02bcd35be19e32617f55c7d3/storages/backends/s3boto3.py#L235

[2]: https://github.com/jschneier/django-storages/blob/7dd36fb820b796ec02bcd35be19e32617f55c7d3/storages/backends/s3boto3.py#L503

[3]: https://github.com/boto/s3transfer/blob/7a0d98013197d02077c0ba71bbcd5f31d971a313/s3transfer/upload.py#L745

[4]: https://github.com/django/django/blob/70147215bb848539273b42617b32773013bcdc47/django/views/static.py#L47

[5]: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax

[6]: https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.FileField.storage

[7]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#object-metadata

dennisvang:
And, of course, it turns out there's already an open issue: https://github.com/jschneier/django-storages/issues/917

UPDATE:

This has been fixed in [django-storages 1.14][1] (see https://github.com/jschneier/django-storages/pull/1281 for details)

[1]: https://github.com/jschneier/django-storages/commit/d1cd2db41fd3bec01fe97d4b095159f974b91673

Issue #2040, opened 2022-06-24
Implement TAP 14


mnm678:
**Description of issue or feature request**:
Implement [TAP 14: Managing TUF versions](https://github.com/theupdateframework/taps/blob/master/tap14.md)

**Current behavior**:
TUF clients will not be able to safely update once the repository moves to a new major specification version.

**Expected behavior**:
TAP 14 is implemented, giving clients a period of time in which to update to the new specification version

jku:
I'm not confident this is something the python-tuf library can meaningfully provide as a generic feature -- I'm not against someone trying that, just warning that it seems difficult. Somehow automatically producing all versions at the same time sounds like a maintenance and testing problem.

I can see how a mechanism to support multiple incompatible client versions could be useful, but my hunch is that it's a _repository feature (and not a library decision)_: A specific repository implementation could decide to start providing two separate versions of the same metadata for a specific new TUF feature, to allow for a period when old clients keep working while new clients benefit from new features. 

In practice I would imagine this just looks like two separate repositories that happen to use some of the same keys and get roughly the same content updates... and since updating to the new incompatible repo by definition requires a client update, there's no problem in changing the repository url when that update is done. So I'm not sure the spec needs complications for this? Maybe I just don't see the case and a practical example would help?

mnm678:
The proposal in the TAP should be specific enough that python-tuf can have a feature to parse the new version metadata from a repository (ie finding the 2.0.0 folder and using the metadata there once it exists), and make some tests for various repository use cases. 

I agree that actual use of this feature will require repositories to make decisions about which versions to support, etc. I'm hoping that eventually we can explore this kind of thing in the repository-sandbox. But it would be nice to have the client support in place before we ever have a repository that needs to update versions so that this can be done quickly if there's ever a real need (security patch, etc).

mnm678:
Replying to a comment in [another issue](https://github.com/theupdateframework/python-tuf/pull/2049#issuecomment-1207448368):

The real world use case for this client workflow is when the client isn't controlled by the same entity as the repository. This is possible for clients that use multiple repositories, and for other areas when the client and repository implementations are more separate, like in the case of the many Sigstore TUF clients.

In either of these cases, the client will need this mechanism to determine a compatible supported version, and ensure the client upgrades as soon as possible after the repository.

jku:
Thanks, this is useful detail.

[...]

PR #2049, opened 2022-07-06
Tap 14 draft pr

abs007:
• Commits `fc5bd5c` to `88c140a` are based around setting up a sample metadata structure inside the `repository` folder (which is inside `repository_data`) for testing

• `774a039` is for reverting `repository` back to its original state and `8845ebd` stores the metadata inside a new `TAP 14` folder

• `df46e38` and `0f50945` add test functions that check inside the TAP 14 folder

• Currently working on implementing changes to the client update process inside the `updater.py` file


abs007:
Should I squash commits `fc5bd5c` to `774a039` ?

MVrachev:
@abs007 as the changes are a lot could you chime in and explain your thoughts on the TAP 14 implementation?
Now as you have some experience with that you have a valuable perspective on that and as it seems there is no full consensus on that feature https://github.com/theupdateframework/python-tuf/issues/2040 it will be good to hear your opinion.

abs007:
So currently I along with @mnm678 and @znewman01 are working on implementing the changes to how a client would be downloading the metadata according to the format in the TAP 14 specification. This means tinkering with the `updater.py` file to add new functionality.

tests/repository_data/repository/1.0.0/role1.json
```diff
@@ -34,7 +34,7 @@
    ]
   },
   "expires": "2030-01-01T00:00:00Z",
-  "spec_version": "1.0.0",
```
MVrachev: Please keep in mind that you can't just replace the value of any of the fields in the `signed` dictionary without re-signing the files again or the files will fail verification.

abs007: This I believe was made as a temporary change rather and the original state of `repository_data/repository` was brought back again in `774a039`. 

That being said, all of the metadata was copied over inside the `repository_data/TAP 14/1.0.0` & `repository_data/TAP 14/2.0.0` folders and metadata inside the `2.0.0` had the `"spec_version"` changed to `2.0.0` in a similar manner (just to show that this was metadata belonging to 2.0.0).
That seems like something I'll have to take a look at, thanks.

jku:
I know this is still a draft but I'll leave a quick comment: I don't understand why the client needs this functionality and before this PR is made ready I would like to see a comment in #2040 that explains the real world use case: _how is this useful for python-tuf users?_ 

Background for my confusion: If a repository starts offering multiple versions of the repository, why can't clients decide to "upgrade" using a plain old software update on the client, so that the client just switches from old to new repository URL? (I'm obviously assuming that the python-tuf version used supports both versions of the spec -- but so does this proposal I think)

In practice, if a future TUF-enabled pypi.org made available a second repository using a newer TUF spec features I think it would be totally appropriate for pip, the client, to simply change the used repo URL to the new repository in a pip software update. What advantage is there in making this repository selection dynamic?

znewman01:
Thanks for the question,  will get back to you soon. But in the meantime, can we move this discussion back to the associated Issue from this PR, since you’re raising questions about the utility of the feature itself?

tuf/ngclient/updater.py
```diff
@@ -85,6 +88,7 @@ def __init__(
         fetcher: Optional[FetcherInterface] = None,
         config: Optional[UpdaterConfig] = None,
     ):
+        self.spec_version = None  # spec_version is the last used version by the client to get metadata
```
znewman01: Make this private: `self._spec_version`

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,38 @@ def refresh(self) -> None:
...
         self._load_snapshot()
         self._load_targets(Targets.type, Root.type)
 
+    def _get_repository_versions(self) -> List[str]:
+        """Returns a list of all the repository versions."""
+
+        url = f"{self._metadata_base_url}supported-versions.json"
+
+        with self._fetcher.download_file(
+            url, "length placeholder"
```
znewman01: probably want `self.config.supported_versions_max_length` in place of `"length placeholder"`. Need to add a line to `tuf/ngclient/config.py` as well.

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,38 @@ def refresh(self) -> None:
...
+        ) as target_file:
+            repository_versions = json.loads(target_file)
+
+        return repository_versions["supported_versions"]
```
znewman01: I think you should just return `repository_versions`. We don't need to put it inside a JSON dictionary

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
 def _ensure_trailing_slash(url: str) -> str:
     """Return url guaranteed to end in a slash"""
     return url if url.endswith("/") else f"{url}/"
+
+
+def _get_spec_version(
+    repository_versions: List[str],
+    spec_version: str,
+    supported_versions: List[str],
+) -> str:
```
znewman01: Oh, we need to return the actual spec version too!

Maybe make the return type `Tuple[Optional[str], Optional[str]]`.

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+    """Returns the specification version to be used."""
```
znewman01: Make sure to mention the return type, which is a little funky. When does it return the message?

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+
+    repository_versions = [int(i) for i in repository_versions]
```
znewman01, line 542: This may throw an exception; please note in the docstring

```
Raises:
   ValueError: supported_versions, repository_version or spec_version contains an invalid entry (not parseable as ``int()``)
```

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+    supported_versions = [int(i) for i in supported_versions]
+    spec_version = int(spec_version)
+
+    # The client determines the latest version available on the repository by looking
+    # for the directory with the largest version number.
+    latestrep_ver = max(repository_versions)
```
znewman01: Rename to `latest_repo_version` for consistency

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+
+    # If the latest version on the repository is lower than the previous specification
+    # version the client used from this repository, the client should report an error
+    # and terminate the update.
+    if spec_version:
+        if latestrep_ver < spec_version:
+            raise exceptions.DownloadError(
```
znewman01: Note this in the docstring.

Also, should be a `RepositoryError`

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+                "The repository version is lower than the last used spec version"
```
znewman01: Please include the version information:

```python
f"The latest repository version ({latest_repo_version}) is lower than the last used spec version ({spec_version})."

File: tuf/ngclient/updater.py [hunk 11]
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+            )
+
+    # If the latest version on the client is found on the repository or vice versa, the
```
znewman01: Is this a typo? I can't find it in https://github.com/theupdateframework/taps/blob/840bbdaef85198e4b80e567b40b92fa592701bec/tap14.md and the phrase "latest version on the client" is confusing.

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+    # client will use this directory to download the metadata.
+    try:
+        spec_version = max(set(repository_versions) & set(supported_versions))
+    except ValueError:
+        raise exceptions.DownloadError(
+            f"""No matching specification version found. Found {repository_versions} in
```
znewman01: Don't use a multiline string; instead, use multiple one-line strings:

```python
raise exceptions.RepositoryError(
   f"No matching specification version found. Found {repository_versions} in "
   f"repository and {supported_versions} in client."
)
```
Python will automatically combine them.

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+            raise exceptions.DownloadError(
+                "The repository version is lower than the last used spec version"
+            )
+
+    # If the latest version on the client is found on the repository or vice versa, the
+    # client will use this directory to download the metadata.
+    try:
+        spec_version = max(set(repository_versions) & set(supported_versions))
+    except ValueError:
+        raise exceptions.DownloadError(
```
znewman01: Should be `RepositoryError`.

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
...
+            f"""No matching specification version found. Found {repository_versions} in
+                                              repository and {supported_versions} in client."""
+        )
+    # Checks for the highest matching version between the client and the repository
```
znewman01: This comment should go above the previous couple lines of code.

tuf/ngclient/updater.py
znewman01: Ahh, we decided to combine them, right.

I think you should copy all three relevant bullets:

> - If the latest version on the repository is equal to that of the client, it will use this directory to download metadata.
> - If the latest version pre-dates the client specification version, it may ...[use] a previous client version to download the metadata.
> If the latest version on the repository is higher than the client spec version, the client should ...perform the update with the directory that corresponds with the latest client specification version, if available. If no such directory exists, the client terminates the update.

tuf/ngclient/updater.py
```diff
@@ -471,7 +503,67 @@ def _preorder_depth_first_walk(
         # If this point is reached then target is not found, return None
         return None
 
+    def _download_directory(self, folder: str) -> None:
```
znewman01: I think this is dead code now and can be deleted.

tuf/ngclient/updater.py
```diff
@@ -262,10 +292,12 @@ def _download_metadata(
     ) -> bytes:
         """Download a metadata file and return it as bytes"""
         encoded_name = parse.quote(rolename, "")
-        if version is None:
-            url = f"{self._metadata_base_url}{encoded_name}.json"
+
+        self.spec_version += ".0.0"
```
znewman01: You shouldn't need to do this. We may want to update the test data instead.

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,38 @@ def refresh(self) -> None:
             RepositoryError: Metadata failed to verify in some way
             DownloadError: Download of a metadata file failed in some way
         """
+        #                                     json.load()
+        # Load self.spec_version using self._load_local_metadata() here from disk
+        # Exception would be raised if spec_version is not found locally. Set to "1" in that case
+        repository_versions = self._get_repository_versions()
+
+        # Updating self.spec_version
+        specver_message = _get_spec_version(
```
znewman01: Use [tuple unpacking](https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences):

```python
spec_version, message =  _get_spec_version(...)
```

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,38 @@ def refresh(self) -> None:
...
+            repository_versions, self.spec_version, SUPPORTED_VERSIONS
+        )
+
+        if specver_message(1):
+            logger.warning(specver_message(1))
+        self.spec_version = specver_message(0)
```
znewman01: You use this with an implicit slash in `download_metadata` below, so make sure to add it. Also, if `spec_version` is `None`, we should put an empty string so that the URL formatting works.

```python
self.spec_version = f"{spec_version}/" or ""
```

tests/test_updater_ng.py
```diff
@@ -326,6 +333,69 @@ def test_non_existing_target_file(self) -> None:
...
+            )
+
+        # Checking specific folders inside TAP 14
+        foldernames = ["targets", "1.0.0"]
+        for folder in foldernames:
+            self.assertTrue(
+                os.path.isdir(os.path.join(self.tap14_directory, folder))
+            )
+
+    def test_get_spec_version1(self) -> None:
```
znewman01: Call this `test_get_spec_version_supported` or something. And put the comment below in a docstring.

tests/test_updater_ng.py
```diff
@@ -326,6 +333,69 @@ def test_non_existing_target_file(self) -> None:
...
+        # This uses the default SUPPORTED_VERSIONS variable from updater.py
+        with self.assertRaises(exceptions.DownloadError):
```
znewman01: `assertRaises` takes an optional second argument: a message. This is really nice way to tell readers of your test code what you're checking:

`self.assertRaises(exceptions.DownloadError, "4 is not a supported version")`.

Go through and do that for every `assertRaises` or `assertEqual` call

tuf/ngclient/updater.py
```diff
@@ -54,6 +56,7 @@
 from tuf.ngclient.fetcher import FetcherInterface
 
 logger = logging.getLogger(__name__)
+SUPPORTED_VERSIONS = ["1", "2", "3"]
```
znewman01: This should actually be `SUPPORTED_VERSIONS = [SPECIFICATION_VERSION]` for now.

tests/test_updater_ng.py
```diff
@@ -326,6 +333,69 @@ def test_non_existing_target_file(self) -> None:
...
+            )
+
+        self.assertEqual(
+            _get_spec_version(
+                ["1", "2", "3"], "3", ngclient.updater.SUPPORTED_VERSIONS
+            ),
+            ("3", None),
+        )
+
+    def test_get_spec_version2(self) -> None:
```
znewman01: Call this just `test_get_spec_version`

tests/test_updater_ng.py
```diff
@@ -326,6 +333,69 @@ def test_non_existing_target_file(self) -> None:
...
+        warningchecker = "Not using the latest specification version available on the repository"
+        # Checks with different values
```
znewman01, line 379: When you have this many test cases, I recommend putting them in a list:

```python
test_cases = [
    (["1", "2", "3"], "3", ["3", "5", "6"],  "3", False),
    (["3", "5", "6"], "3", ["1", "2", "3", "4"], "3", True),
]
for repo_versions, spec_version, supported_versions, expected_version, should_have_warning in test_cases:
    actual_version, warning = _get_spec_version(...)
    self.assertEqual(actual_version, expected_version)
    self.assertEqual(bool(warning), should_have_warning)
# Do something similar for error_test_cases
```

znewman01: Then you can have a comment for each test case explaining why you have it.

znewman01 (commented):
Great work! I have a lot of comments, but you're making progress

tuf/ngclient/updater.py
abs007: But I'm guessing that it'll have to change and be read from the disk?

znewman01: No—`SUPPORTED_VERSIONS` is a property of the *client library* (that is, `python-tuf`).

The "last used specification version" is what we write to/read from disk.

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,38 @@ def refresh(self) -> None:
...
         self._load_timestamp()
         self._load_snapshot()
         self._load_targets(Targets.type, Root.type)
 
+    def _get_repository_versions(self) -> List[str]:
+        """Returns a list of all the repository versions."""
+
+        url = f"{self._metadata_base_url}supported-versions.json"
+
+        with self._fetcher.download_file(
```
znewman01: If this fails, we probably want to return `["1"]`

znewman01: Or `[""]`

tuf/ngclient/updater.py
abs007: Got it.

tuf/ngclient/updater.py
abs007: Could you reiterate on what you mean by "updating the test data"? Do you mean updating the data in `tests/repository_data/` ?
Also one more reason to not use this would be because major versions are being referenced with the digit only, eg: 1.0.0 is being referenced as 1.

tuf/ngclient/updater.py
abs007: Wouldn't this set `self.spec_version` to `None/` in case `spec_version` turns out to be `None`?
I'm guessing, a better alternative would be :
```py
self.spec_version = f"{spec_version}/" if spec_version is not None else ""
```

znewman01: Good catch! That's totally right

tuf/ngclient/updater.py
abs007: Also, should I use `download_bytes()` instead because afai can see, `download_bytes()` just uses `download_file()` inside of `fetcher.py`

tuf/ngclient/updater.py
abs007: Considering my comment above, should `supported_versions_max_length` have the length in bytes equivalent?

tuf/ngclient/updater.py
znewman01: Oh, good call—do that!

tuf/ngclient/updater.py
znewman01: Yes, it should use 'bytes' as the unit

tuf/ngclient/updater.py
abs007: From what I can understand, `supported-versions.json` is a JSON dictionary itself : 
```
{ "supported_versions" : [VERSION, ...],  //From the TAP14 page
```
So, If I return `repository_versions` instead then won't that return the dictionary itself? What I was trying to do was index the `[VERSION, ...]` list using the `supported_versions` key.

tuf/ngclient/updater.py
abs007: How about `Tuple[[str], Optional[str]]` ? `spec_version` is supposed to be either `""` or `[versions, ...]` whereas `warning` can be `None`.

tuf/ngclient/updater.py
abs007: How about? 
```
Returns the specification version to be used and displays a warning if chosen spec_version
is lower than the highest repository version.
```

tuf/ngclient/updater.py
znewman01: Oh, you're right! I misremembered the proposed update to the specification.

tuf/ngclient/updater.py
znewman01: > Do you mean updating the data in tests/repository_data/ ?

Yes, exactly.

tuf/ngclient/updater.py
znewman01: Looks great! I might add "specification version to be used, following the rules of TAP-14"

tuf/ngclient/updater.py
abs007: Cool, will do that

tuf/ngclient/updater.py
znewman01: I see `return (str(spec_version), warning)` below, which suggests:

`Tuple[str, Optional[str]]`.

tuf/ngclient/updater.py
abs007: What could be a possible value for `supported_versions_max_length` in that case?

tuf/ngclient/updater.py
abs007: Also, shouldn't I mention the return type in the function definition above and keep the docstring to just an explanation?

tuf/ngclient/updater.py
abs007: Ah, I meant that, oops. Will add that then.

tuf/ngclient/updater.py
abs007: I'm also guessing that I'm not supposed to keep this in a `try-catch` block then and keep it as it is but just add the Exception to the docstring

tuf/ngclient/updater.py
znewman01: Think about an example `supported-versions.json` file:

```json
{"supported_versions": ["1", "2", "3", "4"]}
```

That's got 45 bytes in it. So let's maybe round up to 1000 bytes? That's a pretty huge margin of error but well short of anything that could cause issues

tuf/ngclient/updater.py
znewman01: Good point—you don't typically want to repeat the type signature in documentation. However, it's not obvious to me from reading what `Tuple[str, Optional[str]]` means.

There are two ways to fix that:

1. Make two [`NewType`s](https://docs.python.org/3/library/typing.html#newtype) :

    ```python
    SpecificationVersion = NewType('SpecificationVersion', str)
    WarningMessage = NewType('WarningMessage', str)
    ```
    
    so the return type becomes `Tuple[SpecificationVersion, Optional[WarningMessage]]`. That's self-documenting, so there's no need to repeat it in the docstring.
    
2. Explain *what* the `str` and `Optional[str]` values are in the docstring

I might actually prefer (1), because I think in the medium term we want to make a full `SpecificationVersion` class to encapsulate ordering (so you don't have to convert to/from `int` in `_get_spec_version`), parsing, and conversion to URL parts. However, (2) is much easier so it's fine to do that for now.

tuf/ngclient/updater.py
znewman01: Yes, you can keep it as-is. No need to re-raise. Just document that sometimes `ValueError` happens.

tuf/ngclient/updater.py
znewman01: Can you push changes that address my comments before marking them as resolved? Makes it easier for me to track what we've been talking about.

abs007: Sure. Will do that

tuf/ngclient/updater.py
abs007: What could be the approx len() of the list in case of 1000 bytes?

tuf/ngclient/updater.py
abs007: I'm a bit confused. If `SpecificationVersion` would essentially be derived from `str` then won't there be `TypeError` be thrown when I'm still comparing it with `latest_repo_version` and performing int type functions on it?

tests/test_updater_ng.py
abs007: Fyi, `test_get_spec_version1` now `test_get_spec_version_supported` doesn't pass because `SUPPORTED_VERSIONS` isn't defined correctly.

tests/test_updater_ng.py
abs007: So, should I add a `msg` or leave a comment? I guess doing both would be repetitive.

znewman01: Yeah, if you're going to put a `msg` in the asserts you don't need a comment.

tests/test_updater_ng.py
znewman01: That's fine—in fact, really `test_get_spec_version` should only handle major version 1. So let's update the test cases

tuf/ngclient/updater.py
znewman01: Right—if you just make `SpecificationVersion` as a `NewType`, you can't compare them directly. It's the same as what you have now.

*Eventually*, you'd make a full `class SpecificationVersion` which doesn't require conversion to/from int.

tuf/ngclient/updater.py
znewman01: Well, a list with 1000 elements would definitely serialize to have > 1000 bytes. The real threshold would probably be a couple hundred. 

Why do you ask?

tests/test_updater_ng.py
abs007: My bad actually. I'll have to put comments if I'll be using the for loop.

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,50 @@ def refresh(self) -> None:
             RepositoryError: Metadata failed to verify in some way
             DownloadError: Download of a metadata file failed in some way
         """
+        try:
+            self._spec_version = self._load_local_metadata("spec_version")
+        except OSError:
+            self._spec_version = "1"
```
znewman01, line 132: This should probably be: `None`

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,50 @@ def refresh(self) -> None:
             RepositoryError: Metadata failed to verify in some way
             DownloadError: Download of a metadata file failed in some way
         """
+        try:
+            self._spec_version = self._load_local_metadata("spec_version")
+        except OSError:
```
znewman01, line 131: Maybe check the errno:

```python
except OSError as err:
    import errno # this should be at top, of course
    if err.errno != errno.ENOENT:
        raise
    self._spec_version = ...
```

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,50 @@ def refresh(self) -> None:
...
+        # Updating self.spec_version
+        spec_version, message = _get_spec_version(
+            repository_versions, self._spec_version, SUPPORTED_VERSIONS
+        )
+
+        if message:
+            logger.warning(message)
+        self._spec_version = (
+            f"{spec_version}/"
+            if spec_version is not None or spec_version == "1"
```
znewman01, line 145: IMO if your ternary operator spans 3 lines, better to just do an if/else

znewman01, line 145: `spec_version is not None or spec_version == "1"` <- doesn't the latter condition imply the former? if `spec_version == "1"` then it's definitely not `None`

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,50 @@ def refresh(self) -> None:
...
+
+            with self._fetcher.download_bytes(
+                url, self.config.supported_versions_max_length
+            ) as target_file:
+                repository_versions = json.loads(target_file)
+                return repository_versions["supported_versions"]
+
+        # If supported-versions.json is not found, then look through the root directory to find supported versions
+        except exceptions.DownloadHTTPError as e:
+            return ["1"]
```
znewman01, line 171: So, we need to decide how to represent:

1. spec version 1, everything lives in the root `/*`
2. spec version 1, everything lives under `/1/*`

These should be different. I think (1) should be `None` and (2) should be `"1"`.

That means we should return `[None]`.

tuf/ngclient/updater.py
```diff
@@ -122,12 +126,50 @@ def refresh(self) -> None:
...
+            url = f"{self._metadata_base_url}supported-versions.json"
+
+            with self._fetcher.download_bytes(
+                url, self.config.supported_versions_max_length
+            ) as target_file:
+                repository_versions = json.loads(target_file)
+                return repository_versions["supported_versions"]
+
+        # If supported-versions.json is not found, then look through the root directory to find supported versions
+        except exceptions.DownloadHTTPError as e:
```
znewman01, line 170: Maybe worth checking `e.status_code == 404`; otherwise it could be a real error.

tests/test_updater_ng.py
```diff
@@ -326,6 +333,109 @@ def test_non_existing_target_file(self) -> None:
...
+
+        # Checking specific folders inside TAP 14
+        foldernames = ["targets", "1.0.0"]
+        for folder in foldernames:
+            self.assertTrue(
+                os.path.isdir(os.path.join(self.tap14_directory, folder))
+            )
+
+    def test_get_spec_version_supported(self) -> None:
+        """This uses the default SUPPORTED_VERSIONS variable from updater.py"""
```
znewman01, line 359: I think it's a good idea to check that `SUPPORTED_VERSIONS == ["1"]`; it won't change for a long time.

tests/test_updater_ng.py
```diff
@@ -326,6 +333,109 @@ def test_non_existing_target_file(self) -> None:
...
+
+        self.assertEqual(
+            _get_spec_version(
+                ["1", "2", "3"], "3", ngclient.updater.SUPPORTED_VERSIONS
+            ),
+            ("3", None),
+            "3 is selected as the spec version and no warning ensues",
+        )
+
+    def test_get_spec_version(self) -> None:
```
znewman01, line 377: I think now you need to account for the `None` case for `repository_versions`

This is actually starting to get a little confusing: there's "version/path pairs" (which come from the repository versions, and then there are just "versions" (from supported versions). `_get_spec_version` should pick a matching (version, path) pair.

It might be worth encoding that in the type system.

tuf/ngclient/updater.py
```diff
@@ -475,3 +519,64 @@ def _preorder_depth_first_walk(
...
+        ValueError: supported_versions, repository_version or spec_version contains an
+        invalid entry (not parseable as ``int()``)
+        RepositoryError: Latest repository version lower than the last used version from
+        this repository
+        RepositoryError: No matching version found between supported_versions and repository_versions
+    """
+    SpecificationVersion = NewType("SpecificationVersion", str)
+    WarningMessage = NewType("WarningMessage", str)
+
+    repository_versions = [int(i) for i in repository_versions]
```
znewman01, line 542: This is gonna be subtle when we have to worry about `None` too (just a warning)

tests/test_updater_ng.py
znewman01: Or maybe even just have a function from "path" to version and use that in `_get_spec_version`.

lukpueh:
#2114 claims that it supersedes this PR. Can we close here?

abs007:
Sure

PR #2075, opened 2022-08-11
Supported `file` scheme without netloc.

### LINKED ISSUE #2074: `requests_fetcher` fails on `file:///` URIs because of `urlparse` fails on them
**Current behavior**:

`DownloadError: Failed to parse URL file:///`

**Expected behavior**:

No error.

KOLANICH:
that is provided, the better.

Fixes #2074

requirements**:

### DIFF
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -6,7 +6,7 @@
 """
 
 import logging
-from typing import Dict, Iterator
+from typing import Dict, Iterator, Tuple
 from urllib import parse
 
 # Imports
@@ -43,7 +43,7 @@ class RequestsFetcher(FetcherInterface):
         # improve efficiency, but avoiding sharing state between different
         # hosts-scheme combinations to minimize subtle security issues.
         # Some cookies may not be HTTP-safe.
-        self._sessions: Dict[str, requests.Session] = {}
+        self._sessions: Dict[Tuple[str, str], requests.Session] = {}
 
         # Default settings
         self.socket_timeout: int = 4  # seconds
@@ -118,10 +118,10 @@ class RequestsFetcher(FetcherInterface):
         # reuse connections while minimizing subtle security issues.
         parsed_url = parse.urlparse(url)
 
-        if not parsed_url.scheme or not parsed_url.hostname:
+        if not parsed_url.scheme:
             raise exceptions.DownloadError(f"Failed to parse URL {url}")
 
-        session_index = f"{parsed_url.scheme}+{parsed_url.hostname}"
+        session_index = (parsed_url.scheme, parsed_url.hostname)
         session = self._sessions.get(session_index)
 
         if not session:
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -121,7 +121,7 @@ class RequestsFetcher(FetcherInterface):
         if not parsed_url.scheme:
             raise exceptions.DownloadError(f"Failed to parse URL {url}")
 
-        session_index = (parsed_url.scheme, parsed_url.hostname)
+        session_index = (parsed_url.scheme, parsed_url.hostname or "")
         session = self._sessions.get(session_index)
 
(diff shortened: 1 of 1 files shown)


tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -118,7 +120,7 @@ def _get_session(self, url: str) -> requests.Session:
         # reuse connections while minimizing subtle security issues.
         parsed_url = parse.urlparse(url)
 
-        if not parsed_url.scheme or not parsed_url.hostname:
+        if not parsed_url.scheme or (not parsed_url.hostname and parsed_url.scheme not in self.SCHEMES_WITHOUT_HOSTNAME):
```
jku: this is unlikely to pass our linter (line looks too long). you can use `tox` to run linters and tests locally.

That said, I wonder if the check is even useful? IIRC it was added just to replace some earlier URL sanity checks, and it's certainly not "real" URL validation: The intent is only to make sure the session index lookup a few lines later works as intended and it seems to me that it does... I think removing the host name check (or even the whole if-clause) may be fine.

jku (changes requested):
Yeah I agree with the intent, this seems good.

I believe this will not pass linting test though, and I think the whole check could maybe be removed (see comment).

If you do modify the commit, please consider adding a commit message describing why this is done. I assume the reason is something like _"This enables using ngclient with a repository accessible via file:// URLs"_ ?

jku:
oh signoff is also missing, please check "DCO" here: https://github.com/theupdateframework/python-tuf/blob/develop/docs/CONTRIBUTING.rst#dco

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -43,7 +43,7 @@ def __init__(self) -> None:
         # improve efficiency, but avoiding sharing state between different
         # hosts-scheme combinations to minimize subtle security issues.
         # Some cookies may not be HTTP-safe.
-        self._sessions: Dict[str, requests.Session] = {}
+        self._sessions: Dict[Tuple[str, Optional[str]], requests.Session] = {}
```
jku: nitpick: no need for Optional: hostname is guaranteed to be a string (it may be an empty string but it's not None).

jku:
DCO failure is likely because your commit author email and the signoff email are not the same. I assume using the noreply address works (if that's what you wanted to do) as long as it's used consistently but I have never tried myself

tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -6,7 +6,7 @@
 """
 
 import logging
-from typing import Dict, Iterator
+from typing import Dict, Iterator, Tuple, Optional
```
jku: linter is complaining about the order here but `Optional` is also unused: could be removed entirely.

Sorry, we have a lot of lint checks -- the code used to be a mishmash of styles and we decided being strict is better than dealing with that...

jku:
> nitpick: no need for Optional: hostname is guaranteed to be a string (it may be an empty string but it's not None).

Argh, it looks like the typeshed annotations for requests disagree (they look buggy to me, but :shrug:). Very sorry, that is entirely on me.

I'm fine with your original use of "Optional" or alternatively this probably works too as a oneline change:
`session_index = (parsed_url.scheme, parsed_url.hostname or "")`

jku:
If you want to run the linter locally, `tox -e lint` works once you've [installed](https://theupdateframework.readthedocs.io/en/latest/INSTALLATION.html#install-for-development) the dev requirements.

That said, if you feel you are done with this PR, let me know: I can continue this as well later this week since the latest issue is really mine.

jku:
There's a fix in https://github.com/jku/python-tuf/commits/netloc -- I think I'm not currently allowed to push the fix to this branch but options are:
* I can make a new PR with your commit and mine
* You can include that fix in this PR
* you can check _Allow edits and access to secrets by maintainers_ in this PR so I can push the fix here

any way works for me

jku (approved):
LGTM: the empty hostname leads to a session per scheme, so e.g. all "file://" urls will use the same sessions

PR #2097, opened 2022-09-06
Update my maintainer email

jku:
The email in MAINTAINERS file already bounces: this one should work :)


lukpueh (approved):
I confirm that this is the same person.

PR #2114, opened 2022-09-22
PoC: Tap 14

### LINKED ISSUE #2040: Implement TAP 14

mnm678:
supersedes #2049

This pr is the proof of concept for TAP 14 for the TAP approval process. It should not be merged until that TAP is finalized.

that is provided, the better.

Fixes #2040

requirements**:


tests/test_updater_ng.py
```diff
@@ -167,14 +171,26 @@ def test_refresh_and_download(self) -> None:
         # top-level metadata is in local directory already
         self.updater.refresh()
         self._assert_files(
-            [Root.type, Snapshot.type, Targets.type, Timestamp.type]
+            [
+                Root.type,
+                Snapshot.type,
+                "spec_version",
```
znewman01, line 178: obligatory "we should probably make a type for `spec_version`" 🙂

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
             RepositoryError: Metadata failed to verify in some way
             DownloadError: Download of a metadata file failed in some way
         """
+        try:
```
znewman01, line 149: probably worth refactoring into its own method

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+            version_bytes = self._load_local_metadata("spec_version")
+            self._spec_version = json.loads(version_bytes.decode("utf-8"))[
```
znewman01, line 151: move this line into an `else` clause?

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+    def _get_repository_versions(self) -> List[str]:
+        """Returns a list of all the repository versions and paths."""
+
+        try:
+            url = f"{self._metadata_base_url}supported-versions.json"
+
+            target_bytes = self._fetcher.download_bytes(
+                url, self.config.supported_versions_max_length
+            )
+            repository_versions = json.loads(target_bytes)
```
znewman01: I would love to pull the parsing/types/etc. out into a `SupportedVersions` type that could be tested independently

znewman01: That's probably a more natural home for e.g. `_get_spec_versions`

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+            "spec_version",
+            json.dumps({"version": spec_version}).encode("utf-8"),
+        )
+
+        self._load_root(ordered_version_paths)
         self._load_timestamp()
         self._load_snapshot()
         self._load_targets(Targets.type, Root.type)
 
+    def _get_repository_versions(self) -> List[str]:
```
znewman01: `List[str]` may be an out-of-date type now

tuf/ngclient/updater.py
```diff
@@ -262,10 +339,16 @@ def _download_metadata(
     ) -> bytes:
         """Download a metadata file and return it as bytes"""
         encoded_name = parse.quote(rolename, "")
-        if version is None:
-            url = f"{self._metadata_base_url}{encoded_name}.json"
+
+        if self._spec_version_dir == "":
```
znewman01: IMO this is getting really hard to read. I think it would benefit a lot from a:

```python
def _format_url(self, base: str, parts: List[str]):
    return f"{base.strip('/')}/{'/'.join([p.strip('/') for p in parts])"

def _format_metadata_basename(rolename: str, version: Optional[int]) -> str:
    encoded_name = parse.quote(rolename, "")
    if version is None:
        return f"{encoded_name.json}"
    return f"{version}.{encoded_name}.json"
```

So you could build the URL up as:

```python
url = self._format_url(filter(None, [self._spec_version_dir, _format_basename(encoded_name, version)]))
```

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+
+        ordered_version_paths = []
 
-        self._load_root()
+        def filter_fn(a):
+            return a["version"] == version and version <= int(spec_version)
+
+        for version in sorted(repository_versions):
+            path = list(
+                filter(
```
znewman01: IMO clearer as a list comprehension with an `if` clause

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+        )
+
+        if message:
+            logger.warning(message)
+        self._spec_version = spec_version
+
+        ordered_version_paths = []
 
-        self._load_root()
+        def filter_fn(a):
```
znewman01: Can we give this a more meaningful name?

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+            repository_versions, self._spec_version, self._supported_versions
+        )
+
+        if message:
+            logger.warning(message)
+        self._spec_version = spec_version
+
+        ordered_version_paths = []
 
-        self._load_root()
```
znewman01, line 126: Preference to pull this out into its own method.

tuf/ngclient/updater.py
```diff
@@ -475,3 +570,61 @@ def _preorder_depth_first_walk(
 def _ensure_trailing_slash(url: str) -> str:
     """Return url guaranteed to end in a slash"""
     return url if url.endswith("/") else f"{url}/"
+
+
+def _get_spec_version(
+    repository_versions: List[str],
+    spec_version: str,
+    supported_versions: List[str],
+) -> Tuple[str, Optional[str]]:
```
znewman01: Might be nice to introduce a `SpecVersion` class (basically a `version`/`path` tuple)

znewman01: This would make the `int` conversion lines a little more bulletproof

tuf/ngclient/updater.py
```diff
@@ -307,21 +390,33 @@ def _load_root(self) -> None:
...
-                self._persist_metadata(Root.type, data)
+        save_repo_version_dir = self._spec_version_dir
 
-            except exceptions.DownloadHTTPError as exception:
-                if exception.status_code not in {403, 404}:
-                    raise
-                # 404/403 means current root is newest available
-                break
+        for next_version in range(lower_bound, upper_bound):
+            for version_repo in supported_version_repos:
```
znewman01: I think that as-written this would accept `/1.root.json`, `/2/2.root.json`, `3.root.json` which feels wrong.

znewman01: Maybe we want to say:

```python
for next_version in range(lower_bound, upper_bound):
    while supported_version_repos:
        if it's in supported_version_repos[0]:
            handle it!
            break
        else:
            supported_version_repos.pop()
```

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+            if i["version"] == spec_version:
+                self._spec_version_dir = i["path"]
+
+        # persist the version number
+        self._persist_metadata(
+            "spec_version",
+            json.dumps({"version": spec_version}).encode("utf-8"),
+        )
+
+        self._load_root(ordered_version_paths)
```
znewman01: Thinking about this: it's totally possible that v2 root format will look totally different, so it seems wrong to try to load them the same way.

I've thought about implementing TAP-14 as a *wrapper* around the Updater. Basically, it would expose the same interface and keep a dict of `{MajorVersion: Updater}`.

Then, its job is to (1) parse the `supported-versions` and (2) delegate appropriately. This keeps the complexity of TAP-14 out of the Updater itself, and provides a more natural way to support a very-different v2.

tuf/ngclient/updater.py
mnm678: I also thought about moving some of this into the metadata api, especially if we finalize the decision to sign this file in theupdateframework/taps#158

tuf/ngclient/updater.py
mnm678: I agree in general, but `_load_root` is unique in that it has to look in directories for multiple supported versions in order to ensure that no root metadata is missed. If the root format is changed, then this function could be updated as part of the client update to use the correct format for the current root file that is being checked.

joshuagl: > I've thought about implementing TAP-14 as a _wrapper_ around the Updater. Basically, it would expose the same interface and keep a dict of `{MajorVersion: Updater}`.

This is a great idea. Legacy python-tuf implemented [TAP 4: Multiple repository consensus on entrusted targets](https://github.com/theupdateframework/taps/blob/master/tap4.md) (mapping file) support with a wrapper around the updater (see [`MultiRepoUpdater`](https://github.com/theupdateframework/python-tuf/blob/febe6c3dc30195078a5487d68d7db708e7de8479/tuf/client/updater.py#L161)). It could be useful to experiment with multiple wrappers around the updater and how they would interact, so that we can determine whether we need to come up with another pattern so that TAP implementations aren't mutually exclusive.

znewman01: That's a great point. We're hoping to take that approach in go-tuf for TAP 4 as well!

Interactions sound a little tricky...I *think* we're okay, we'd put the TAP-4 wrapper as the outermost layer?

We can cross that bridge when we come to it.

tuf/ngclient/updater.py
mnm678: Just added a test, this does ignore `3.root.json` in your example. The complication is that both the `supported_version_repos` look and the `next_version` loop have to break early in some (different) cases, which is where the additional complexity comes in. I'm happy to simplify if there's an easier way to represent this.

mnm678:
Thanks for the feedback! I made some changes. I'm going to do a larger refactor that creates types for `SpecVersion` and `RepositoryVersions` pending the discussion in theupdateframework/taps#158

tuf/ngclient/updater.py
joshuagl: Yeah, my hunch is that ordering with TAP 4 at the outermost layer makes sense for TAP 4 + TAP 14.

JustinCappos:
Does it make sense to have each piece of ?root? metadata on a repo always
contain a list of the other spec_versions of metadata that are available on
that repo?  This way a client can better avoid being downgraded...

mnm678:
> Does it make sense to have each piece of ?root? metadata on a repo always contain a list of the other spec_versions of metadata that are available on that repo? This way a client can better avoid being downgraded...

I don't want to require anything of existing v1 implementations, but this is something we could explore for v2++. Another, similar idea discussed in theupdateframework/taps#158 is to have the `supported-versions.json`, which lists the supported versions on the repo file signed by root keys.

All of this is just to ensure that clients can find new versions as soon as they are available. The client won't downgrade spec versions once they use a higher spec version with a given repo.

trishankatdatadog:
I'm still [not convinced](https://github.com/theupdateframework/taps/pull/158#issuecomment-1261325613) that there is a use case for clients dynamically choosing which major spec version they want to support.

JustinCappos:
> I'm still [not convinced](https://github.com/theupdateframework/taps/pull/158#issuecomment-1261325613) that there is a use case for clients dynamically choosing which major spec version they want to support.

I read the linked comment.  Are you arguing that a client should support exactly one spec version or that a client should not decide per repo? 

I would say that a client supporting multiple spec versions may be necessary if it interacts with different repos that use different versions.  I do agree with the latter, that a client should always use the latest version.

trishankatdatadog:
> I would say that a client supporting multiple spec versions may be necessary if it interacts with different repos that use different versions.

Is there such a client? Why design for a need that does not yet exist? I think we should be careful not to overgeneralise.

tuf/ngclient/updater.py
```diff
@@ -85,6 +91,8 @@ def __init__(
         fetcher: Optional[FetcherInterface] = None,
         config: Optional[UpdaterConfig] = None,
     ):
+        self._spec_version = None  # spec_version is the last used version by the client to get metadata
```
MVrachev: Nit: Please provide a type for `_spec_version` as I suspect that `mypy` won't be happy.
Something like:
`self._spec_version: Optional[str] = None`

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
 
-        self._load_root()
+        self._load_root(ordered_version_paths)
         self._load_timestamp()
         self._load_snapshot()
         self._load_targets(Targets.type, Root.type)
 
+    def _set_spec_version(self) -> List[str]:
+        try:
+            version_bytes = self._load_local_metadata("spec_version")
```
MVrachev, line 150: Where can we have a look into an example `spec_version.json` file?

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+        )
+
+        return ordered_version_paths
+
+
+    def _get_repository_versions(self) -> List[object]:
+        """Returns a list of all the repository versions and paths."""
+
+        try:
+            url = f"{self._metadata_base_url}supported-versions.json"
```
MVrachev: Are there recommendations or documentation somewhere on what format will the `supported-versions.json` file have?

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+        # persist the version number
+        self._persist_metadata(
+            "spec_version",
+            json.dumps({"version": spec_version}).encode("utf-8"),
+        )
+
+        return ordered_version_paths
+
+
+    def _get_repository_versions(self) -> List[object]:
```
MVrachev: Actually, you know the exact format of your return type (as we see below in the except cause) and I think it will be better if you describe it as:
`List[Dict[Any]]`

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+            ]
+
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+            self._spec_version = "1"
+
+        repository_versions_and_paths = self._get_repository_versions()
+
+        repository_versions = [
```
MVrachev: Nit, but still makes it more readable:
```suggestion
        repository_versions: List[str] = [
```

tuf/ngclient/updater.py
```diff
@@ -475,3 +568,61 @@ def _preorder_depth_first_walk(
 def _ensure_trailing_slash(url: str) -> str:
     """Return url guaranteed to end in a slash"""
     return url if url.endswith("/") else f"{url}/"
+
+
+def _get_spec_version(
```
MVrachev: There are a lot of improvements I can suggest here to fit our style-guidelines, but will restrain myself for now as this is a proof of concept pr.

tuf/ngclient/updater.py
MVrachev: I think I agree with @znewman01 if there are other use cases where you will need that tuple, but it seems as here the tuple doesn't mean `versions/path`, but `version/warning` where `warning` is even optional.

MVrachev: Honestly, not sure why we return the `warning` and not directly print it.
This will save us from the need to return it, make a variable, etc.

tuf/ngclient/updater.py
```diff
@@ -122,12 +131,80 @@ def refresh(self) -> None:
...
+        )
+
+        if message:
+            logger.warning(message)
+        self._spec_version = spec_version
+
+        ordered_version_paths = []
+
+        for version in sorted(repository_versions):
+            path = [i["path"] for i in repository_versions_and_paths if version <= int(spec_version) and i["version"] == version]
```
MVrachev: I think this is too complicated to be a one-liner.

tuf/ngclient/updater.py
```diff
@@ -296,7 +376,7 @@ def _persist_metadata(self, rolename: str, data: bytes) -> None:
                     pass
             raise e
 
-    def _load_root(self) -> None:
+    def _load_root(self, supported_version_repos) -> None:
```
MVrachev: Annotation is needed for `supported_version_repos`.

tuf/ngclient/updater.py
```diff
@@ -307,21 +387,34 @@ def _load_root(self) -> None:
...
-        for next_version in range(lower_bound, upper_bound):
-            try:
-                data = self._download_metadata(
-                    Root.type,
-                    self.config.root_max_length,
-                    next_version,
-                )
-                self._trusted_set.update_root(data)
-                self._persist_metadata(Root.type, data)
+        save_repo_version_dir = self._spec_version_dir
```
MVrachev: I see why you save the original content of `self._spec_version_dir` because you change `self._spec_version_dir` multiple times later, but I had to read it a couple of times.

I suggest instead passing the `version_repo` variable to `download_metadata` and with this you remove:
- `save_repo_version_dir` variable
- `to_break` flag
- the `if` check for `to_break` at the end of the inner cycle and you can directly return after you add to `supported_version_repos` list

MVrachev (changes requested):
I have reviewed the core TAP 14 logic in `updater.py` and in `config.py`.
I will review the test files next week.

mnm678:
I updated this pr to reflect theupdateframework/taps#158, and to address the mypy type errors.

There is one remaining mypy error that I can't seem to resolve, so I'd appreciate some advice from reviewers :)

tests/repository_simulator.py
```diff
@@ -205,10 +205,15 @@ def _fetch(self, url: str) -> Iterator[bytes]:
         if path.startswith("/metadata/") and path.endswith(".json"):
             # figure out rolename and version
             ver_and_name = path[len("/metadata/") :][: -len(".json")]
+            # inside a version folder
+            if "/" in ver_and_name:
+                ver_and_name = ver_and_name.split("/")[1]
```
znewman01: What happens if there's multiple nested folders? `ver_and_name = "/foo/bar.json"`

I like 

```python
>>> "foo/bar/baz.json".partition("/")
('foo', '/', 'bar/baz.json')
>>> "foo/bar/baz.json".rpartition("/") 
('foo/bar', '/', 'baz.json')
```

For making intent a little clearer there.

You can also `assert len(ver_and_name.split("/")  <= 2`

tests/repository_simulator.py
```diff
@@ -205,10 +205,15 @@ def _fetch(self, url: str) -> Iterator[bytes]:
...
             version_str, _, role = ver_and_name.partition(".")
             # root is always version-prefixed while timestamp is always NOT
             if role == Root.type or (
-                self.root.consistent_snapshot and ver_and_name != Timestamp.type
+                self.root.consistent_snapshot
+                and ver_and_name != Timestamp.type
+                and ver_and_name != "supported-versions"
```
znewman01: This isn't a thing anymore, right?

znewman01 (commented):
Enjoy the comments I left before I realized this hadn't been updated to match the new TAP 😛 

I think for mypy you can manually annotate the type of `test_cases`:

```python
test_cases: List[Tuple[List[str], str, List[str], str]] = ...
```

It kinda seems like a bug in mypy, though. Not sure where it's getting that shape for test_cases...

mnm678:
Thanks @znewman01, this now reflects the latest version of the TAP.

jku:
TAP is now deferred. 

I'll close this since there has been no action and it seems unlikely to get merged without major work. Please reopen if I'm being too hasty

PR #2137, opened 2022-10-12
Fix: allow `length` to be zero

### LINKED ISSUE #2136: Incorrect metapath length validation

As per [TUF metapath specification](https://theupdateframework.github.io/specification/latest/#metapath-length), `length` attribute of metadata is an integer numerical value. However, the [current](https://github.com/theupdateframework/python-tuf/blob/28b027815e2e6490c4ec7cc21d8f73373cbe2670/tuf/api/metadata.py#L1055-L1056) validation states that length has to be greater than zero. So I believe this is a slight inconsistency with the specification?

There are legitimate use-cases where signed metadata is an empty file which length is 0. For example, storing git empty files results in their length being 0. 

**Current behavior**:

Signed metadata with length 0 raises a `ValueError('Length must be > 0, got 0')`

**Expected behavior**:

Length can be 0.

n-dusan:
* As per TUF specification, length attribute is a numerical value (which can include 0) - https://theupdateframework.github.io/specification/latest/#metapath-length

fix: update tests

Fixes #2136 

### DIFF
--- a/tests/test_metadata_serialization.py
+++ b/tests/test_metadata_serialization.py
@@ -297,7 +297,6 @@ class TestSerialization(unittest.TestCase):
     invalid_metafiles: utils.DataSet = {
         "wrong length type": '{"version": 1, "length": "a", "hashes": {"sha256" : "abc"}}',
         "version 0": '{"version": 0, "length": 1, "hashes": {"sha256" : "abc"}}',
-        "length 0": '{"version": 1, "length": 0, "hashes": {"sha256" : "abc"}}',
         "length below 0": '{"version": 1, "length": -1, "hashes": {"sha256" : "abc"}}',
         "empty hashes dict": '{"version": 1, "length": 1, "hashes": {}}',
         "hashes wrong type": '{"version": 1, "length": 1, "hashes": 1}',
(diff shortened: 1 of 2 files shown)


jku:
Thanks. I think I agree with this. @MVrachev any comment?

I suppose it might make sense to also test that `TargetFile.verify_length_and_hashes()` really supports the "empty input" (I'm pretty sure it does but :shrug:)?  I mean either a manual check and confirmation comment here or an actual test in tests/test_api.py. That said I think this PR is correct even if there was a bug in `verify_length_and_hashes()` so that doesn't really need to happen in this PR...

MVrachev (commented):
The pr itself is correct.
I don't see why anyone would use `length` with a value of `0`.
I am in favor of clarifying the spec, so that `length` is >=0 than making the changes in `python-tuf`.

lukpueh (approved):
> I am in favor of clarifying the spec, so that length is >=0

IIUC this PR claims that the spec already suggests that length must be `>=0`, whereas current behavior in `python-tuf` is `>=1`.

I agree with @n-dusan and @jku that it makes sense sense to allow zero-length target files, and that, while there is no use case for zero-length metadata files, it is okay to allow them.  (see https://github.com/theupdateframework/python-tuf/issues/2136#issuecomment-1276043558)

FWIW it does not seem more reasonable to restrict metadata file lengths to 1, which is the current behavior.

MVrachev (approved):
Yea, now I see the context and why that was decided.
LGTM!

n-dusan:
Thank you all! @jku I can definitely confirm after manually testing the changes that `targetinfo.verify_length_and_hashes()` works with a targetinfo file of length 0.
Example of `targetinfo`:
```
{'length': 0, 'hashes': {'sha256': 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855', 'sha512': 'cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e'}}
```

Is there anything else needed on my end?

PR #2157, opened 2022-10-27
build: Enable Python 3.11 in test matrix

jku:
3.11 is out, let's see how it runs...


joshuagl (approved):
Thanks for keeping up with Python releases!

PR #2158, opened 2022-10-29
updated hatchling to current latest 1.11.1

### LINKED ISSUE #2030: deps: hatchling update
We pin hatchling in pyproject.toml: `build-system.requires = ["hatchling==0.22.0"]`. I think this is the correct choice: otherwise the build is not really reproducible.

It looks like Dependabot does not notice this as a dependency: so we're currently several versions behind.

Tasks:
* Update hatchling to current
* Figure out how we stay up-to-date?

UDITBALUJA:
New Pull Request to fix #2030
Updated hatchling version to the latest 1.11.1

Signed-off-by: UDITBALUJA <uditbaluja1000@gmail.com>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 # hatchling pinned for reproducibility: version should be kept up-to-date
-requires = ["hatchling==0.22.0"]
+requires = ["hatchling==1.11.1"]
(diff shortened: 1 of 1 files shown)


jku (approved):
Thanks for persevering, this looks good to me.

PR #2159, opened 2022-10-31
Github workflows: Permissions tweaks

### LINKED ISSUE #2116: review token permissions (based on openssf scorecard recommendations)
https://deps.dev/project/github/theupdateframework%2Fpython-tuf
* expand token-permissions
* click "show details"

We're getting 0/10 on the openssf scorecard for _Token-Permissions_. I think some of those may be flaws in the scorecard tool (output even mentions "known issues") but I think this warrants a closer read and possibly filing issues for fixing the problems.

jku:
Fixes #2116 

Tweak workflow permissions based on OSSF scorecard suggestions
* set workflow-level permissions to none
* use job-level permissions to minimize exposure

Also:
* skip pypi-upload step on forks (this allows testing most of the CD workflow in forks). This is not really relevant to other commits and can be removed.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -82,6 +82,8 @@ jobs:
           path: dist
 
       - name: Publish binary wheel and source tarball on PyPI
+        # Only attempt pypi upload in upstream repository
+        if: github.repository == 'theupdateframework/python-tuf'
         uses: pypa/gh-action-pypi-publish@37f50c210e3d2f9450da2cd423303d6a14a6e29f
         with:
           user: __token__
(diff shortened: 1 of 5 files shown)


.github/workflows/cd.yml
```diff
@@ -36,6 +33,30 @@ jobs:
       - name: Build binary wheel and source tarball
         run: python3 -m build --sdist --wheel --outdir dist/ .
 
+      - name: Store build artifacts
+        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb
```
joshuagl, line 37: Note: If we want to further reduce the TCB we could record the digest of the generated artefacts, store that in an output, and compare it when retrieving the artefacts from the file store. This would remove the need to trust the artefact storage.

joshuagl (approved):
Really nice set of changes, thank you!

PR #2165, opened 2022-11-04
Move (most of) Key to Securesystemslib

jku:
This updates python-tuf sources for latest securesystemslib signer api changes. Most of the changes are in tests (since we test Key quite extensively) but verify_delegate() is also significantly rewritten. The commits are mostly logical -- the one exception is that the original refactoring of role/key lookup in Metadata was not great: it is fixed in a later commit by moving the functionality from Metadata to Targets and Root where it makes more sense.

Changelist:
* Key moves to Securesystemslib and the Key API changes:
  * `Key.from_securesystemslib_key()` is now `SSlibKey.from_securesystemslib_key()`
  * `Key.verify_signature()` arguments and exceptions change (it used to take Metadata as arg)
*  `Metadata.verify_delegate()` has to be rewritten because of `Key.verify_signature()` changes
* Introduced some new API calls in Targets and Root: `get_delegated_role()` and `get_key()`: these simplify the role and key lookup in `Metadata.verify_delegate()`

WRT API change: 
* this is clearly an API change (as Key changes method signatures), but not something users should suffer badly from
* Key, Signer and Signature are still technically part of the metadata API so no regression there: they just come from a securesystemslib module now.
* The Key implementations (like SSlibKey and SSlibSigner) are not currently importable through Metadata API but I don't think that is a problem?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,7 +34,7 @@ allowlist_externals = python3
 # Must to be invoked explicitly with, e.g. `tox -e with-sslib-master`
 [testenv:with-sslib-master]
 commands_pre =
-    python3 -m pip install git+https://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib[crypto,pynacl]
+    python3 -m pip install --force-reinstall git+https://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib[crypto,pynacl]
 
 commands =
     python3 -m coverage run aggregate_tests.py
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,8 +61,6 @@ html_favicon = "tuf-icon-32.png"
 # -- Autodoc configuration ---------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
 
-autodoc_mock_imports = ["securesystemslib"]
-
 # Tone down the "tuf.api.metadata." repetition
 add_module_names = False
 python_use_unqualified_type_names = True
--- a/examples/manual_repo/basic_repo.py
+++ b/examples/manual_repo/basic_repo.py
@@ -27,13 +27,12 @@ from pathlib import Path
 from typing import Any, Dict
 
 from securesystemslib.keys import generate_ed25519_key
-from securesystemslib.signer import SSlibSigner
+from securesystemslib.signer import SSlibKey, SSlibSigner
 
 from tuf.api.metadata import (
     SPECIFICATION_VERSION,
     DelegatedRole,
     Delegations,
-    Key,
     Metadata,
     MetaFile,
     Root,
@@ -157,7 +156,7 @@ roles["root"] = Metadata(Root(expires=_in(365)))
 for name in ["targets", "snapshot", "timestamp", "root"]:
     keys[name] = generate_ed25519_key()
     roles["root"].signed.add_key(
-        Key.from_securesystemslib_key(keys[name]), name
+        SSlibKey.from_securesystemslib_key(keys[name]), name
     )
 
 # NOTE: We only need the public part to populate root, so it is possible to use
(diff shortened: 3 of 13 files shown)


jku:
Unsure what to do with the documentation: I don't want to reproduce the docs in python-tuf... but I don't know if we can import the Key docs from securesystemslib?

lukpueh:
>  but I don't know if we can import the Key docs from securesystemslib?

I think that's a good idea. It's what we for in-toto's key utilities:
https://in-toto.readthedocs.io/en/latest/api.html#key-utilities

jku:
> > but I don't know if we can import the Key docs from securesystemslib?
> 
> I think that's a good idea. It's what we for in-toto's key utilities: https://in-toto.readthedocs.io/en/latest/api.html#key-utilities

oh nice: we should be doing that for Signer and Signature as well -- basically for the "modern" securesystemslib api.

it's interesting how that autodoc magic seems to work without ever importing securesystemslib...

jku:
this is now out of date since I modified the securesystemslib PR -- it's still roughly correct, but I am not sure if a python-tuf Key is even needed anymore, maybe not. Will update next week

tox.ini
```diff
@@ -31,7 +31,7 @@ install_command = python3 -m pip install {opts} {packages}
 # Must to be invoked explicitly with, e.g. `tox -e with-sslib-master`
 [testenv:with-sslib-master]
 commands_pre =
-    python3 -m pip install git+https://github.com/secure-systems-lab/securesystemslib.git@master#egg=securesystemslib[crypto,pynacl]
+    python3 -m pip install --force-reinstall git+https://github.com/jku/securesystemslib.git@privkeyuri#egg=securesystemslib[crypto,pynacl]
```
jku: this is testing leftover -- but it does lead to one passing test :champagne:

lukpueh:
Just tried out signing with post quantum keys (sphincs+) using the new signer API.  Works like a charm and out of the box! 🎉  cc @rugo 
<details>
<summary>see snippet</summary>

```bash
# Install `tuf` from this branch and `securesystemslib` from secure-systems-lab/securesystemslib#456
python3 -m pip install git+https://github.com/jku/python-tuf.git@no-key#egg=tuf
python3 -m pip install --force-reinstall git+https://github.com/jku/securesystemslib.git@privkeyuri#egg=securesystemslib[crypto,pynacl,PySPX]
```

POC: Verify that root metadata is signed by at least one sphincs+ key as required
```python
import os

from securesystemslib.keys import generate_sphincs_key
from securesystemslib.signer import Signer, SSlibKey, SSlibSigner

from tuf.api.metadata import Metadata, Root

# Generate pq keys
pq_key = generate_sphincs_key()
pq_pubkey = SSlibKey.from_securesystemslib_key(pq_key)
pq_signer = SSlibSigner(pq_key)

# # Generic solution: Use `Signer.from_priv_key_uri` instead of `SSlibSigner`,
# #     this also works:
#
# os.environ["PQ_KEY"] = pq_key["keyval"]["private"]
# pq_signer = Signer.from_priv_key_uri("envvar:PQ_KEY", pq_pubkey)

# Generate root role and configure signature requirement
root = Root()
root.add_key(pq_pubkey, Root.type)

# Wrap root in metadata envelope and add signature
root_metadata = Metadata(root)
root_metadata.sign(pq_signer)

# Assert `verify_delegate` indeed needs at least one sig from above key
assert root_metadata.signed.roles["root"].threshold == 1
assert root_metadata.signed.roles["root"].keyids == [pq_pubkey.keyid]

root_metadata.verify_delegate(Root.type, root_metadata)
```

</details>

jku:
Current state:
* Works with securesystemslib master
* Should not be merged before there is a securesystemslib release (and we start depending on it)
* the main commit really should be split into a few separate ones
* the `verify_delegate()` refactor needs a good review

jku:
Ok, PR is ready for review I think -- not for merge though, we do need a securesystemslib release before that. Updated description

jku:
Rebased

jku:
Rebased and then added a commit to fix #2272

jku:
Rebased and then added 
* commit to fix #2272
* one more fix to key usage in examples 
* tweak docs build so Key is still documented

jku:
TODO: Package should depend on securesystemslib >= 0.26 now

lukpueh (changes requested):
> TODO: Package should depend on securesystemslib >= 0.26 now

I think this is the only thing that's missing.

Otherwise LGTM. 👍

lukpueh (approved):
Thanks!

PR #2180, opened 2022-11-17
feat: generate hash-prefixed path names for target

### LINKED ISSUE #2166: Allow generation of hash-prefixed versions of `TargetFile`s

Allow generation of hash-prefixed versions of `TargetFile`s.

**Current behavior**:
Currently, python-tuf does not have the ability to generate hash-prefixed path names for targets files. One suggestion from @jku is to add a utility function in Metadata API's `TargetFile` like: 
```
Target.File.get_prefixed_paths() -> List[str]
```

**Expected behavior**:
The expected behavior is to able to get the hash-prefixed path names for any given targets file. 

Example: 
https://github.com/vmware-labs/repository-editor-for-tuf/blob/main/tufrepo/git_repo.py#L210

kamui-fin:
Fixes #2166 

I introduced a new method `get_prefixed_paths()` for class `TargetFile`, which generates hash-prefixed paths for the given target file path. I might be missing some additional requirements, so please let me know if any other changes are needed.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -731,6 +731,36 @@ class TestMetadata(unittest.TestCase):
         )
         self.assertEqual(sorted(target.get_prefixed_paths()), ["public/path/abc.file.ext", "public/path/def.file.ext"])
 
+        target = TargetFile(
+            100, {"sha256": "abc", "md5": "def"}, ""
+        )
+        self.assertEqual(sorted(target.get_prefixed_paths()), [])
+
+        target = TargetFile(
+            100, {"sha256": "abc", "md5": "def"}, "public/path/"
+        )
+        self.assertEqual(sorted(target.get_prefixed_paths()), [])
+
+        target = TargetFile(
+            100, {"sha256": "abc", "md5": "def"}, "file.ext"
+        )
+        self.assertEqual(sorted(target.get_prefixed_paths()), ["abc.file.ext", "def.file.ext"])
+
+        target = TargetFile(
+            100, {"sha256": "abc", "md5": "def"}, "public/path/.ext"
+        )
+        self.assertEqual(sorted(target.get_prefixed_paths()), ["public/path/abc..ext", "public/path/def..ext"])
+
+        target = TargetFile(
+            100, {"sha256": "abc"}, "/root/file.ext"
+        )
+        self.assertEqual(sorted(target.get_prefixed_paths()), ["/root/abc.file.ext"])
+
+        target = TargetFile(
+            100, {"sha256": "abc"}, "/"
+        )
+        self.assertEqual(sorted(target.get_prefixed_paths()), [])
+
     def test_is_delegated_role(self) -> None:
         # test path matches
         # see more extensive tests in test_is_target_in_pathpattern()
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -725,6 +725,12 @@ class TestMetadata(unittest.TestCase):
         targetfile_from_data = TargetFile.from_data(target_file_path, data)
         targetfile_from_data.verify_length_and_hashes(data)
 
+    def test_targetfile_hash_prefix_paths(self) -> None:
+        target = TargetFile(
+            100, {"sha256": "abc", "md5": "def"}, "public/path/file.ext"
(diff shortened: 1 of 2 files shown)


tuf/api/metadata.py
```diff
@@ -1892,6 +1893,17 @@ def verify_length_and_hashes(self, data: Union[bytes, IO[bytes]]) -> None:
...
 
+    def get_prefixed_paths(self) -> List[str]:
+        """
+        Returns hash-prefixed paths for the given target file path.
+        """
+        paths = []
+        path = pathlib.Path(self.path)
+        name, parent = path.name, path.parent
+        for hash in self.hashes.values():
+            paths.append(str(parent.joinpath(f"{hash}.{name}")))
```
jku: The paths we are dealing with here are URL path fragments, not file system paths: so using pathlib feels wrong, and the windows failures on CI are no doubt related. These tricky details are exactly why this method is likely a good idea :)

Using pathlib.PosixPath would likely give you technically correct results but I think in this case manually splitting and joining on "/" is fine too: that is exactly how URLs are expected to work.

You could also include the word "URL" somewhere in the docstring so readers could make the connection (that these are not filesystem paths but parts of a URL).

tests/test_api.py
```diff
@@ -725,6 +725,12 @@ def test_targetfile_from_data(self) -> None:
         targetfile_from_data = TargetFile.from_data(target_file_path, data)
         targetfile_from_data.verify_length_and_hashes(data)
 
+    def test_targetfile_hash_prefix_paths(self) -> None:
+        target = TargetFile(
+            100, {"sha256": "abc", "md5": "def"}, "public/path/file.ext"
+        )
```
jku, line 731: for completeness maybe also:
* path without separator "file.ext"
* path without filename "path/.ext"
* empty path ""
* empty filename "path/"
* (add other path shenanigans if you can think of any more)

jku: Oh since its likely not obvious: empty path and empty filename probably don't make sense: please still test that the code does not e.g. throw exceptions if that happens

jku:
Please see [CONTRIBUTING](https://github.com/theupdateframework/python-tuf/blob/develop/docs/CONTRIBUTING.rst#dco) for advice with the DCO failure (we require acceptance of Developer Certificate of Origin in the form of a "Signed-off-by" line)

tuf/api/metadata.py
```diff
@@ -1892,6 +1893,32 @@ def verify_length_and_hashes(self, data: Union[bytes, IO[bytes]]) -> None:
...
+            if not self.path:
+                raise ValueError
+            elif "/" not in self.path:
+                parent, name = None, self.path
+            else:
+                parent, name = self.path.rsplit("/", 1)
+                if name == "": 
+                    raise ValueError
+        except ValueError:
+            return paths
```
jku, line 1913: This started to look really familiar so I went searching and found a similar situation in tuf/ngclient/updater.py -- I think we can simplify this using the solution there: 4 levels of indent for splitting a string seems too much.

Can you try a solution like this (untested pseudo code):
```python
    parent, sep, name = self.path.rpartition("/")
    for hash in self.hashes.values():
        paths.append(f"{parent}{sep}{hash}.{name}")
```
The idea there is that empty string is considered a valid name, a valid separator and a valid parent: then you never need to handle Nones or empty strings as special cases.

I assume the test results are slightly different, but I think they might be fine -- in the sense that for each hash you actually should get a hash-prefix path, even if the path is a little weird for nonsensical original paths.

jku (changes requested):
Thanks for the tests! There is also a subtest decorator in utils that allows you to have a twoliner test with all test data defined separately -- but I'm not demanding you use it, this is readable.

I did leave a simplification suggestion you could try as the actual algorithm. I think it changes the test results but in a reasonable way: The returned list should IMO have as many items as the hashes dict.

jku:
Closing this as nothing happened in a month to keep PR list manageable: The PR is still welcome, feel free to reopen if there's a question or further changes

jku:
I've used this branch as the basis for a new one (rebasing these commits), I hope that's ok.

PR #2183, opened 2022-11-18
README: Remove limitations

jku:
There may be ways to unsafely use the client library but situation should be significantly better now with ngclient:
  * metadata writing is safer, more atomic
  * non-root cached metadata is never trusted (so inconsistent cached repository is not a security issue)
  * the cache locations are now clearly application decisions (they are required Updater constructor args)

I think the current unsafe usages are not surprising or worthy of mention in top level README: it's not surprising that modifying the app cache during the apps runtime may affect the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,12 +82,6 @@ Security Issues and Bugs
 
 See [SECURITY.md](docs/SECURITY.md)
 
-Limitations
------------
-
-The reference implementation may behave unexpectedly when concurrently
-downloading the same target files with the same TUF client.
-
 License
 -------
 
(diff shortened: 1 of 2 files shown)


jku:
As a better documentation of this, we could include a note in ngclient docs, something like "ngclient users should be 'single instance' applications: running multiple instances using the same cache directories is not recommended". 

I suppose I should add that to this PR

joshuagl:
> As a better documentation of this, we could include a note in ngclient docs, something like "ngclient users should be 'single instance' applications: running multiple instances using the same cache directories is not recommended".
> 
> I suppose I should add that to this PR

Please do :-)

lukpueh (approved):

PR #2186, opened 2022-11-19
feat: add from_data() and from_file() for MetaFile

### LINKED ISSUE #1893: MetaFile is missing from_bytes() or equivalent
We already have `TargetFile.from_data() / TargetFile.from_file()` to calculate hashes and length for a target.
We should optionally support hashes and length for Metafile too (but likely the API is not same, see below).

This hasn't come up so far because we haven't really used the optional hashes in our metafiles... it did come up in https://github.com/theupdateframework/go-tuf/pull/228.

Notes:
* presumably most of the code currently in `TargetFile.from_data()`can be shared somehow
* the situation is different in MetaFile and TargetFile for 2 reasons:
  * `MetaFile` needs version that comes from a `Metadata` -- obviously the implementation could parse the bytes it gets to construct Metadata... but that's not ideal because we can expect the Metadata is already in memory, reparsing is wasteful. The go-tuf generator code is a fine example of this.
  * A valid `MetaFile` only needs version: the API should not automatically lead to using hashes

potential API (_maybe_ add `from_file()` too to be consistent):

```python
class MetaFile(BaseFile):

    @classmethod
    def from_data(
        cls,
        version: int,
        data: Optional[Union[bytes, IO[bytes]]] = None,
        hash_algorithms: Optional[List[str]] = None,
    ) -> "MetaFile"
        # data is the Metadata bytes that the metafile represents.
        # Hashes and length should be included only if data is not None
```

This is consistent with TargetFile, but looks a bit odd: it's called `from_data()` but data argument can be None? Maybe data should not be optional: we should just document that `Metafile.from_data(ver, data)` constructor should only be used if hashes are wanted, by default `MetaFile(ver)` should be used.

kamui-fin:
Signed-off-by: Kamui <fin-kamui@pm.me>

Fixes #1893 

I added two `MetaFile` constructors `from_data` and `from_file` as per the issue details. 

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -33,6 +33,7 @@ from tuf.api.metadata import (
     Delegations,
     Key,
     Metadata,
+    MetaFile,
     Root,
     Snapshot,
     SuccinctRoles,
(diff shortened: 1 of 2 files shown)


jku (changes requested):
Thanks! Two comments:
* the code for from_data() looks solid but I think it is mostly the same as TargetFile.from_data(). I'm not against repeating code if it makes sense, but this seems avoidable... Can you try reusing the code -- as a possible solution, you could add a static method `_get_length_and_hashes(data: Union[bytes, IO[bytes]], hash_algorithms: List[str]) -> Tuple[int, Dict[str, str]]` to BaseFile class which both TargetFile and MetaFile derive from, and then call that from both `from_data()` methods?
* Now that I see `from_file()`, I think it may be unwise: there is a Metadata storage abstraction used for loading and storing metadata to disk... and this would kind of sidestep that abstraction by reading metadata straight from disk. I suggest dropping from_file() from the PR: we can figure that out later if it is important.

jku:
Closing this as nothing happened in a month to keep PR list manageable: The PR is still welcome, feel free to reopen if there's a question or further changes

PR #2188, opened 2022-11-21
pyproject: Stop disabling urllib3 import checks in mypy

jku:
Double reasoning for this one:
* urllib3 now does have annotations
* since we don't import requests annotations (to avoid depending on typeshed) urllib3 annotations are never needed: we don't use urllib3 directly


lukpueh (approved):
Thanks!

PR #2190, opened 2022-11-22
workflows: Add Scorecards workflow

### LINKED ISSUE #2160: Publish OSSF Scorecard results
https://github.com/ossf/scorecard

Would be nice to get these results visible. They have a GitHub Action that we could use.

The installation instructions say:
> One Scorecards check ([Branch-Protection](https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection)) requires authentication using a Personal Access Token (PAT). If you want all Scorecards checks to run on a public repository, you will need to follow the optional Authentication step. If you don't, all checks will run except Branch-Protection.

The token is described as "read-only" but I don't understand why it would be read-only... maybe it is but I don't get how.

So:
* the github action would be nice
* the action should work without a token: I'm fine with skipping the branch-protection check

jku:
This is a modifed version of the workflow from the project itself:
* Not using personal access tokens because I believe they are a security issue (this means Branch-Protection check result will be incorrect)
* Not uploading results to actions cache: Maybe there's a point but I don't see it as the SARIF files are not very human readable

This should give us some code scanning alerts in the security tab on Github. This is not really what I'm interested in though so I've enabled the upload to https://api.securityscorecards.dev/. The results json on there is not exactly readable but it is good enough to check what the current results are -- and deps.dev should use those results after some delay I believe.

--- 

Fixes #2160

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -1,0 +1,39 @@
+name: Scorecards analysis
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '21 6 * * 1'
+  push:
+    branches: [ develop ]
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  analysis:
+    name: Scorecards analysis
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write # for uploading to code-scanning dashboard
+      id-token: write # for signing results
+      actions: read
+      contents: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@99c53751e09b9529366343771cc321ec74e9bd3d
+        with:
+          results_file: results.sarif
+          # sarif format required by upload-sarif action
+          results_format: sarif
+          # "repo_token" not set because personal access tokens are dangerous.
+          # This means Branch-Protection check will not have correct results.
+          publish_results: true
+
+      - name: "Upload to code-scanning dashboard"
+        uses: github/codeql-action/upload-sarif@8aff97f12c99086bdb92ff62ae06dbbcdf07941b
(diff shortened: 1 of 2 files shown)


jku:
As usual with workflows the results can't be fully seen before it's merged: I have a test run on my fork https://github.com/jku/python-tuf/actions/runs/3523182162/jobs/5907032685 but obviously the results there are not very representative.

Some code scan alerts do appear in security tab: I suppose that's a good sign

.github/workflows/scorecards.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@99c53751e09b9529366343771cc321ec74e9bd3d
+        with:
+          results_file: results.sarif
+          results_format: sarif
```
joshuagl, line 31: Nit: I was surprised at the choice of Sarif format until I read more of the workflow and docs. Might be worth adding a comment like:
```suggestion
          # Sarif format is required to upload to the code-scanning dashboard
          results_format: sarif
```

.github/workflows/scorecards.yml
```diff
@@ -0,0 +1,38 @@
...
+          # "repo_token" not set because personal access tokens are dangerous.
+          # This means Branch-Protection check will not have correct results.
+          publish_results: true
+
+      - name: "Upload to code-scanning dashboard"
+        uses: github/codeql-action/upload-sarif@8aff97f12c99086bdb92ff62ae06dbbcdf07941b
+        with:
+          sarif_file: results.sarif
```
joshuagl: Nit: new-line at EOF to appease Git?
```suggestion
          sarif_file: results.sarif

```

joshuagl (approved):
Nice, thanks for working on this. I added a couple of minor nits as comments. Your call whether to address them or not.

PR #2191, opened 2022-11-23
Fixes #1862, added way to see full stack trace on error situations

### LINKED ISSUE #1862: example-client: better error handling

ameypd:
Passed verbosity as a parameter to download function. If exception occurs in the exception block, the value of verbosity is checked and if it is not equal to the default value(0), we print the stack trace using the traceback library.

### DIFF
--- a/examples/client_example/client_example.py
+++ b/examples/client_example/client_example.py
@@ -9,6 +9,7 @@ import logging
 import os
 import shutil
 from pathlib import Path
+import traceback
 
 from tuf.api.exceptions import DownloadError, RepositoryError
 from tuf.ngclient import Updater
@@ -39,7 +40,7 @@ def init() -> None:
         print(f"Found trusted root in {METADATA_DIR}")
 
 
-def download(target: str) -> bool:
+def download(target: str, error_level: int) -> bool:
     """
     Download the target file using ``ngclient`` Updater.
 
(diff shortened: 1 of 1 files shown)


lukpueh (changes requested):
Cool stuff!  This looks mostly good, but you don't need to pass the error level. You can also read this information from the current logger (i.e. `logging.root.level`, see docs for `logging.basicConfig`).

ameypd:
okay I will do the requested changes.

PR #2193, opened 2022-11-25
Repository module and example

### LINKED ISSUE #1136: Design repository library
*[Update 2/14/22]* 

- [ADR 10](https://github.com/theupdateframework/python-tuf/blob/develop/docs/adr/0010-repository-library-design.md) shows the updated direction of this issue including further reading (design document, prototype implementation)
- Comments in thread here below point to survey documents for existing repository tooling, based on which ADR 10 was developed

---

*[Original issue description]*

Identify what functionality from existing repository_tool/repository_lib/roledb/keydb modules should exist in a TUF repository library on top of a metadata model as sketched in #1112.

Start with API functionality.

Coordinate with #1134.

jku:
Calling this a _repository module_ sounds a little too grand (it's literally an abstract class that does almost nothing)... but I think it works so maybe we start with this.

I'm adding some reviewers for visibility. If someone wants to see what it does:
* for demo, please try https://github.com/jku/python-tuf/blob/repository-lib/examples/client_example/README.md
* the implementation example is in https://github.com/jku/python-tuf/blob/repository-lib/examples/repository
* the "library" is in https://github.com/jku/python-tuf/blob/repository-lib/tuf/repository

Note that the Repository class is just a "ergonomic metadata editing framework" -- it makes producing valid and logical next versions of TUF metadata within a repository easier. So it's not really a "repository" (module name makes sense as this is for repository side tooling but the class name could still be debated)

examples/repository should also be extended (or other examples added): 
* it should have a tiny developer API and there should be a developer tool (built using the same building blocks) that can sign and upload things to the example repository app
* some of the more complex repository details demonstrated in the manual examples should be implemented here as well (and eventually manual repo examples deleted)

Fixes #1136 maybe?


jku:
> for demo, please try https://github.com/jku/python-tuf/blob/repository-lib/examples/client_example/README.md

Forgot to mention: you may be tempted to run the server and use a browser to see the metadata being served. This sort of works but ultimately breaks things as at least Chrome does things that are incompatible with single thread servers like `http.server.HTTPServer`: it opens connections (just in case) and then does nothing.

jku:
Having played with this for about a week now, I think there is one "Repository implementation" we may want to include in the library itself. While the repositories themselves may be wildly different, the "developer tools" or "signing tools" are probably very similar:
* they have a root of trust like a downloader client
* they have a cache for local, valid metadata like a downloader client
* they _may_ update to current metadata by downloading from the remote repository like a downloader client would (but may also have other means like `git pull`)
* they do want to only operate on top of a valid metadata cache... but the definition of valid is not the same as in a downloader client: as an example they do want to load expired metadata so that they can make a new version of it...
* they likely never want to modify the metadata _in cache_: they'll want to modify the metadata to send a new version to the remote repository. The method of "sending" will vary (http post, git commit & push) but the idea is the same: the metadata cache should represent the remote repository state, and the modified metadata gets sent to that repository to integrate 

So the result should be a mix of ngclient.Updater and the SimpleRepository from this PR. This is clearly future work and will not happen in in this PR, I'm just writing the idea out there.

jku:
plain rebase, no changes

jku:
A potential change to snapshot/timestamp: 
- Current implementation requires "meta" as argument to both methods
- alternative is to require Repository to provide snapshot meta and timestamp meta as properties, this way the methods wouldn't need arguments.

I think this would be as flexible but would make calling snapshot()/timestamp() easier

**EDIT**: I've implemented this in later commit

tuf/repository/_repository.py
```diff
@@ -0,0 +1,139 @@
...
+            logger.debug("Snapshot update not needed")  # type: ignore[unreachable]
+        else:
+            logger.debug(
+                "Snapshot v%d, %d targets", snapshot.version, len(snapshot.meta)
+            )
+
+        version = snapshot.version if updated_snapshot else None
+        return version, removed
+
+    def timestamp(self, snapshot_meta: MetaFile) -> Optional[MetaFile]:
```
jku: one thing I don't like WRT snapshot/timestamp here (in addition to arguments mentioned in another comment):

timestamp() always creates a new version, snapshot does not.
* this is practically useful as timestamp does not get called "just in case" but snapshot might (a repository might not know if snapshot is needed or not: are there new targets or not?)
* on the other hand especially timestamp should get called ""without reason", just to bump the timestamp
* this is annoyingly unbalanced... so maybe should add a `force` argument to both to allow either mode -- "always create new version" or "only create if needed"

jku: modified this, resolving

examples/repository/README.md
```diff
@@ -0,0 +1,23 @@
+# TUF Repository Application Example
+
+
+This TUF Repository Application Example has following features:
```
lukpueh:
```suggestion
This TUF Repository Application Example has the following features:
```

examples/repository/README.md
```diff
@@ -0,0 +1,23 @@
...
+- Initializes a completely new repository on startup
+- Stores everything (metadata, targets, signing keys) in-memory
+- Serves metadata and targets on localhost (default port 8001)
+- Simulates a live repository by automatically adding a new target
+  file every 10 seconds.
+
+
+### Example with the repository example
```
lukpueh: Minor nit: The title sounds a bit weird with the double "example", what about something like
"Usage", or "How to run"?

examples/repository/README.md
```diff
@@ -0,0 +1,23 @@
...
+```console
+./repo
+```
+Your repository is now running and is accessible on localhost, See e.g.
+http://127.0.0.1:8001/metadata/1.root.json
+
+Note that because the example generates a new repository at startup,
+clients need to also re-initialize their trust root when the repository
+application is restarted. With the example client this is done with
+`./client tofu`.
```
lukpueh: Minor nit: I think you can remove this paragraph here, you already have a similar note in the client_example README.md, where it is more important.

examples/repository/_simplerepo.py
```diff
@@ -0,0 +1,137 @@
...
+class SimpleRepository(Repository):
+    """Very simple in-memory repository implementation
+
+    This repository keeps the metadata for all versions of all roles in memory.
+    It also keeps all target content in memory.
+
+
+    Attributes:
+        role_cache: Contains every historical metadata version of every role in
+            this repositorys. Keys are rolenames and values are lists of
```
lukpueh:
```suggestion
            this repository. Keys are role names and values are lists of
```

examples/repository/_simplerepo.py
```diff
@@ -0,0 +1,137 @@
...
+            Metadata
+        signer_cache: Contains all signers available to the repository. Keys
+            are rolenames, values are lists of signers
```
lukpueh:
```suggestion
            are role names, values are lists of signers
```

tuf/repository/_repository.py
```diff
@@ -0,0 +1,172 @@
...
+        with ones from all available keys. Keep snapshot_info and targets_infos
+        updated."""
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def targets_infos(self) -> Dict[str, MetaFile]:
+        """Returns the current targets version information
+
+        Not that there is a difference between this and the published snapshot
```
lukpueh:
```suggestion
        Note that there is a difference between this and the published snapshot
```

tuf/repository/_repository.py
```diff
@@ -0,0 +1,172 @@
...
+        meta: This dictionary reflects the targets metadata that currently
+        exists in the repository, but the dictionary published by snapshot()
+        will also include metadata that no longer exists in the repository.
+        """
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def snapshot_info(self) -> MetaFile:
+        """Returns the information matching current snapshot metadata"""
```
lukpueh: What is "the information matching current snapshot metadata"?

tuf/repository/_repository.py
```diff
@@ -0,0 +1,172 @@
...
+        raise NotImplementedError
+
+    @contextmanager
+    def edit(
+        self, role: str, init: bool = False
+    ) -> Generator[Signed, None, None]:
+        """Context manager for editing a roles metadata
```
lukpueh:
```suggestion
        """Context manager for editing a role's metadata
```

tuf/repository/_repository.py
```diff
@@ -0,0 +1,172 @@
...
+
+        Context manager user can raise AbortEdit from inside the with-block to
+        cancel the edit: in this case none of the changes are stored.
+        """
+        md = self.open(role, init)
+        with suppress(AbortEdit):
+            yield md.signed
+            self.close(role, md)
+
+    def sign(self, role: str) -> None:
```
lukpueh: Do you have a use case for this? Currently, I don't see it being used here or in the example repo app implementation.

tuf/repository/_repository.py
```diff
@@ -0,0 +1,172 @@
...
+        """sign without modifying content, or removing existing signatures"""
+        md = self.open(role)
+        self.close(role, md, sign_only=True)
+
+    def snapshot(self, force: bool = False) -> Tuple[bool, Dict[str, MetaFile]]:
+        """Update snapshot meta information
+
+        Updates the snapshot meta information according to current targets
+        metadata state and the current current snapshot meta information.
```
lukpueh:
```suggestion
        metadata state and the current snapshot meta information.
```

tuf/repository/_repository.py
```diff
@@ -0,0 +1,172 @@
...
+
+        Arguments:
+            force: should new snapshot version be created even if meta
+                information would not change?
+
+        Returns: Tuple of
+            - True if snapshot was created, False if not
+            - Meta information for targets metadata was removed from snapshot
```
lukpueh:
```suggestion
            - Meta information for targets metadata that was removed from snapshot
```

tuf/repository/_repository.py
```diff
@@ -0,0 +1,172 @@
...
+            if not update_version:
+                # prevent edit() from storing a new snapshot version
+                raise AbortEdit("Skip snapshot: No targets version changes")
+
+        if not update_version:
+            # this is reachable as edit() handles AbortEdit
+            logger.debug("Snapshot update not needed")  # type: ignore[unreachable]
+        else:
+            logger.debug(
+                "Snapshot v%d, %d targets", snapshot.version, len(snapshot.meta)
```
lukpueh: Minor nit: "targets" is so ambiguous. I've seen it being used to say multiple "target files" in this PR, here it means multiple "targets metadata *files*".

examples/repository/_simplerepo.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def open(self, role: str, init: bool = False) -> Metadata:
+        """Return current Metadata for role from 'storage' (or create a new one)"""
+
+        if init:
+            signed_init = _signed_init.get(role, Targets)
+            md = Metadata(signed_init())
+
+            # this makes version bumping in close() simpler
+            md.signed.version = 0
```
lukpueh, line 97: sneaky :P

examples/client_example/README.md
```diff
@@ -4,23 +4,43 @@
...
+# initialize the client with Trust-On-First-Use
+./client tofu
+
+# Then download example files from the repository:
+./client download file1.txt
+```
+
+Note that unlike normal repositories, the example repository only exists in
+memory and is re-generated from scratch at every startup: This means your
+client needs to run `tofu` everytime you restart the repository application.
```
lukpueh:
```suggestion
client needs to run `tofu` every time you restart the repository application.
```

lukpueh (commented):
Impressive work! I've given this a very close look, especially `tuf/repository` and `examples/repository/_simplerepo.py`, and only found some minor nits. I think we can get this merged and even released soon, but maybe note somewhere that this is more experimental than ngclient and metadata api.

tuf/repository/_repository.py
jku: no examples here, this really only comes to play with distributed threshold signing: you get metadata from elsewhere, and load it just so that you can sign (without bumping version etc) .

jku:
(I can't believe I'm asking for bikeshedding but here goes...)

One thing that I would like bring up for discussion is naming:
* `open()/close()` -- on one hand these are descriptive but it's not perfect: as an example, not closing a file descriptor is a bug but not closing a role in this api is fine: you just decided not to write a new version after all. I have also considered `open()/commit()` but so far the symmetry of the current naming has won. Maybe this is not even an important detail as most users should be using `edit()` which handles open/close.
* `snapshot_info` and `targets_infos`: these caches are really hard to name without making a mess in the code: A timestamp role contains snapshot meta and snapshot role contains targets meta, and these two should sort of map to those metas. `snapshot_info` and `targets_infos` represent the current state of stored metadata while the metas are a view of the published repository at a specific point in time -- including already removed roles which these caches do not contain. 

Another thing is the idea of the `edit()` contextmanager: does that make sense to others? Is the use of AbortEdit still understandable -- I imagine it won't be needed often but snapshot() does use it so it should be understandable...

tuf/repository/_repository.py
```diff
@@ -0,0 +1,173 @@
...
+    @property
+    @abstractmethod
+    def snapshot_info(self) -> MetaFile:
+        """Returns the information matching current snapshot metadata"""
+        raise NotImplementedError
+
+    @contextmanager
+    def edit(
+        self, role: str, init: bool = False
+    ) -> Generator[Signed, None, None]:
```
jku: the core usability complaint I have with edit() is that it doesn't handle better static typing:

```python
with edit("targets") as targets:
    # mypy sees targets as type Signed, not Targets
```

Most issues would just be solved by adding a typed helper:
```python
    def edit_targets(
        self, role: str, init: bool = False
    ) -> Generator[Targets, None, None]:
```

(not sure if the other types even need a helper: timestamp and snapshot at least don't need so much editing: Repository handles the usual cases already)

lukpueh: I am not very exited about adding a helper just for typing. What's the implication if mypy doesn't know the subtype?

jku: The implication is that more complex code in the with-block (like this [attempt at easy target removal](https://github.com/vmware-labs/repository-editor-for-tuf/blob/main/tufrepo/librepo/repo.py#L158)) won't have type suggestions in your editor and won't be type checked by mypy unless you do some extra work.

The workaround I use in that example makes VSCode suggestions work but doesn't make mypy happy... I think the least disruptive workaround is probably a `assert(isinstance(Targets, targets))` as the first line inside the with-block.

lukpueh:
> (I can't believe I'm asking for bikeshedding but here goes...)
> 
> One thing that I would like bring up for discussion is naming:

Fair. Naming is important, and hard.

> * `open()/close()` -- on one hand these are descriptive but it's not perfect: as an example, not closing a file descriptor is a bug but not closing a role in this api is fine: you just decided not to write a new version after all. I have also considered `open()/commit()` but so far the symmetry of the current naming has won. Maybe this is not even an important detail as most users should be using `edit()` which handles open/close.

I had similar thoughts. These names do have other connotations. What about more abstract terms, like  start/end, enter/exit, pre_edit/post_edit? ... otoh, those make less sense, when used outside of edit. Maybe open/close are just fine.

> * `snapshot_info` and `targets_infos`: these caches are really hard to name without making a mess in the code: A timestamp role contains snapshot meta and snapshot role contains targets meta, and these two should sort of map to those metas. `snapshot_info` and `targets_infos` represent the current state of stored metadata while the metas are a view of the published repository at a specific point in time -- including already removed roles which these caches do not contain.

Can't think of better names. Let's solve this with documentation. Maybe add a note to look at `timestamp` and `snapshot` methods, and e.g. `SimpleRepository`, when in doubt about how these are maintained. 

For that it might be helpful to rename the second return value in snapshot (e.g. to
replaced/updated/old/previous), to make it clear that this does not contradict with *"the dictionary published by snapshot() will also include metadata that no longer exists in the repository"*.

Btw., it also took me a bit to understand the magical defaultdict+lambda -powered assignment in SimpleRepository, when I tried to understand `targets_infos`. But maybe that's just me. And it's definitely a cool too. Maybe a comment would help?

> 
> Another thing is the idea of the `edit()` contextmanager: does that make sense to others? 

I think it does. One thing I was wondering was whether open/close arguments should also be left unspecified given that open and close can be anything the implementer wants them to be. E.g. they could both accept `**kwargs` passed on like so:
```
def edit(role: str, open_kwargs: Dict[str, Any] = None, close_kwargs: Dict[str, Any] = None)
``` 

> Is the use of AbortEdit still understandable -- I imagine it won't be needed often but snapshot() does use it so it should be understandable...

I think it's fine.

tuf/repository/_repository.py
jku: but just to be clear: I'm not suggesting to add a helper in this PR -- just bringing up potential areas of improvement

tuf/repository/_repository.py
jku: I will try to improve this...

I know you already know what it's trying to express but for others: snapshot_info represents the (version,hashes) of the current snapshot metadata -- this version may be newer than the one listed in timestamp meta.

jku:
> Btw., it also took me a bit to understand the magical defaultdict+lambda -powered assignment in SimpleRepository, when I tried to understand `targets_infos`. But maybe that's just me. And it's definitely a cool too. Maybe a comment would help?

yeah, that defaultdict is likely only useful once in in close() ... so the case of missing key could be handled manually there. But on the other hand role_cache and signer_cache use the same mechanism (it would even look the same if MetaFile() had a default version value).

I'll just add a comment

jku:
> For that it might be helpful to rename the second return value in snapshot (e.g. to
> replaced/updated/old/previous), to make it clear that this does not contradict with _"the dictionary published by snapshot() will also include metadata that no longer exists in the repository"_.

I've kept the names but I tweaked the docstring of the methods to make it clear that the returned values are metadata _versions_ that are no longer part of snapshot/timestamp meta.

I think I've now addressed everything so far.

jku:
From chat with lukas:

> open(self, role: str, init: bool = False)

this does not really need the init argument...

could also move sign() out of the ABC -- then close also does not need the sign_only argument.

jku:
Two changes:
* init argument was removed from open: it's just not necessary
* `sign()` (meaning signing an already existing version) was removed from the ABC: this should be done by implementations that really need this. It's not clear if the open/close concept works in that case at all

examples/repository/_simplerepo.py
lukpueh: Ping s/repositorys/repository

tuf/repository/_repository.py
```diff
@@ -0,0 +1,173 @@
...
+    """Abstract class for metadata modifying implementations
+
+    NOTE: The repository module is not considered part of the python-tuf
+    stable API yet.
+
+    This class is intended to be a base class used in any metadata editing
+    application, whether it is a real repository server or a developer tool.
+
+    Implementations must implement open() and close(), and can then use the
+    edit() contextmanager to implement actual operations. Not that signing
```
lukpueh:
```suggestion
    edit() contextmanager to implement actual operations. Note that signing
```

examples/repository/_simplerepo.py
```diff
@@ -0,0 +1,134 @@
...
+class SimpleRepository(Repository):
+    """Very simple in-memory repository implementation
+
+    This repository keeps the metadata for all versions of all roles in memory.
+    It also keeps all target content in memory.
+
+
+    Attributes:
+        role_cache: Every historical metadata version of every role in this
+            repositorys. Keys are role names and values are lists of Metadata
```
lukpueh:
```suggestion
            repository. Keys are role names and values are lists of Metadata
```

tuf/repository/_repository.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @abstractmethod
+    def close(self, role: str, md: Metadata) -> None:
+        """Write roles metadata into storage
+
+        If sign_only, then just append signatures of all available keys.
+
+        If not sign_only, update expiry and version and replace signatures
+        with ones from all available keys. Keep snapshot_info and targets_infos
+        updated."""
```
lukpueh: No more `sign_only`

lukpueh (changes requested):
Please fix two typos and one outdated docstring and this is good to go.

lukpueh (approved):

PR #2197, opened 2022-11-28
ngclient: Fail gracefully on missing role

### LINKED ISSUE #2195: ngclient should cope better with broken repos
* a delegation exists in a repository
* but the delegated roles metadata is not in the repository and not included in the snapshot
* Updater.get_targetinfo() is called with a path that is delegated to this nonexistent role

this currently leads to following error. We should fail with a RepositoryError of some kind instead.

```
  File "/home/jkukkonen/src/python-tuf/tuf/ngclient/updater.py", line 381, in _load_targets
    data = self._load_local_metadata(role)
  File "/home/jkukkonen/src/python-tuf/tuf/ngclient/updater.py", line 277, in _load_local_metadata
    with open(os.path.join(self._dir, f"{encoded_name}.json"), "rb") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/jkukkonen/.local/share/tuf-upload-example/d412c05c/secrole.json'

During handling of the above exception, another exception occurred:

  [... unrelated application stack race removed ...]
    updater.get_targetinfo(targetname)
  File "/home/jkukkonen/src/python-tuf/tuf/ngclient/updater.py", line 171, in get_targetinfo
    return self._preorder_depth_first_walk(target_path)
  File "/home/jkukkonen/src/python-tuf/tuf/ngclient/updater.py", line 436, in _preorder_depth_first_walk
    targets = self._load_targets(role_name, parent_role).signed
  File "/home/jkukkonen/src/python-tuf/tuf/ngclient/updater.py", line 392, in _load_targets
    metainfo = self._trusted_set.snapshot.signed.meta[f"{role}.json"]
KeyError: 'secrole.json'

```

jku:
If role is delegated but missing from snapshot, we currently raise a undocumented KeyError: a generic RepositoryError seems better as callers are expected to handle it (and adding a more specific error seems useless as this is a repository software bug, not just expired metadata or something).

The same check is also done later in TrustedMetadataSet but I think keeping the check in both is clearest.

Fixes #2195

---

### DIFF
--- a/tests/test_updater_fetch_target.py
+++ b/tests/test_updater_fetch_target.py
@@ -16,6 +16,7 @@ from typing import Optional
 from tests import utils
 from tests.repository_simulator import RepositorySimulator
 from tuf.api.exceptions import RepositoryError
+from tuf.api.metadata import DelegatedRole, Delegations
 from tuf.ngclient import Updater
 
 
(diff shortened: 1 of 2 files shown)


lukpueh (approved):
LGTM

PR #2199, opened 2022-11-28
fixes #1568 added import tuf and version = tuf.__version__ in conf.py

### LINKED ISSUE #1568: docs: python-tuf version should be visible in docs (readthedocs)

gkum99:
Signed-off-by: gkum99 <h20220014@goa.bits-pilani.ac.in>

that is provided, the better.

Fixes #1568 

Added import tuf and version = tuf.__version__ in conf.py

requirements**:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -75,3 +75,6 @@ autodoc_default_options = {
     'inherited-members': 'Exception', # excl. members inherited from 'Exception'
     'exclude-members': 'to_dict, from_dict'
 }
+
+# Version
+version = tuf.__version__


jku:
Annoyingly readthedocs seems to overwrite version in the PR doc builds (https://theupdateframework--2199.org.readthedocs.build/en/2199/) with the PR number... so the build result doesn't actually tell us if this works :joy: we'll just have to assume it does

jku (changes requested):
I think the code change probably works: let's try that.  But your current commit message is
> “fixes

Please replace that with a descriptive one, for example _Include python-tuf version in documentation_. You can force push the PR branch we don't mind that.

Optionally you could also move your code changes next to the other project information (copyright, author etc) where it makes sense -- but I can live without that change.

PR #2201, opened 2022-11-30
fixes #1568 Include python-tuf version in documentation

### LINKED ISSUE #1568: docs: python-tuf version should be visible in docs (readthedocs)

gkum99:
I have added code "version = tuf.__version__ " in conf.py and it is displaying version number in main documentation page.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,56 +12,56 @@
 #
 import os
 import sys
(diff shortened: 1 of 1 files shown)


jku (approved):
Thanks, let's try this and see how it works on readthedocs.

For next time: if you run autoformatting tools like black, having those changes in a separate commit would be nice.

PR #2208, opened 2022-12-06
Updater: add missing config docstring

MVrachev:
Updater: add missing config docstring.


jku (approved):
Looks fine to me, although we could just remove all the "Optionals" in the descriptions: it looks like a repeat of the annotation autodoc already adds.

PR #2215, opened 2022-12-09
changed depreciated __metaclass__ to metaclass

### LINKED ISSUE #2196: code uses python 2 "__metaclass__"
python-tuf and securesystemslib both use class variable `__metaclass__ = abc.ABCMeta` to implement abstract classes. This does nothing in Python3 -- classes can be instantiated even if they have that class variable set 

The actually supported methods are
```python
from abc import ABC
class MyABC(ABC)
```
and
```python
from abc import ABCMeta
class MyABC(metaclass=ABCMeta)
```
the latter may be safer (WRT multiple inheritance issues)

aditya-shrivastavv:
Fixes #2196 

In **Python 2**, the `__metaclass__ `attribute is used to specify the `metaclass` for a class. In **Python 3**, this syntax has been _changed_, and the metaclass keyword argument is used instead.

- [x] Tox test had ran successfully.

It's a simple fix which can be verified by anyone

### DIFF
--- a/tuf/ngclient/fetcher.py
+++ b/tuf/ngclient/fetcher.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 
 
 # Classes
-class FetcherInterface:
+class FetcherInterface(abc.ABC):
     """Defines an interface for abstract network download.
 
     By providing a concrete implementation of the abstract interface,
@@ -28,7 +28,7 @@ class FetcherInterface:
     The public API of the class is already implemented.
     """
 
-    __metaclass__ = abc.ABCMeta
+    metaclass = abc.ABCMeta
 
     @abc.abstractmethod
     def _fetch(self, url: str) -> Iterator[bytes]:
--- a/tuf/ngclient/fetcher.py
+++ b/tuf/ngclient/fetcher.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 
 
 # Classes
-class FetcherInterface(abc.ABC):
+class FetcherInterface(metaclass=abc.ABCMeta):
     """Defines an interface for abstract network download.
 
     By providing a concrete implementation of the abstract interface,
@@ -27,8 +27,6 @@ class FetcherInterface(abc.ABC):
     Implementations of FetcherInterface only need to implement ``_fetch()``.
     The public API of the class is already implemented.
     """
-
-    metaclass = abc.ABCMeta
 
     @abc.abstractmethod
     def _fetch(self, url: str) -> Iterator[bytes]:


jku (changes requested):
Thank you. Please have a look at https://docs.python.org/3/library/abc.html for an example of how `metaclass=abc.ABCMeta` keyword works.

You can verify your work by trying this:
```python
from tuf.ngclient import FetcherInterface
FetcherInterface()
```
this should fail as FetcherInterface should be an abstract class that can't be instantiated. Currently it does not fail.

Also, please make sure your commit includes
* message that describes what you are doing
* signed-off-by as described in docs/CONTRIBUTING.md

aditya-shrivastavv:
Hey jku ! 
I am extremely sorry for my mistakes in the PR.
I tried to correct my mistakes in the new commits, please have a look at them.
I am happy to say that have successfully implemented your "Changes Requested" now after running the commands:
```
from tuf.ngclient import FetcherInterface
FetcherInterface()
```
Error generates which says like - abstract class can't be instantiated. please have a look at it.
once again I want to apologize for causing problems. i am still learning, let me know if some more things need to be corrected.

references taken for solving the problems -
1.
https://www.quora.com/Why-do-we-use-abstract-class-Python-if-we-cannot-instantiate-it-it-is-only-inherited-Why-do-we-not-create-another-classes-instead-What-is-the-secret-behind-that-kind-of-classes
2.
https://stackoverflow.com/questions/67498595/in-python-can-an-abstract-class-be-instantiated

aditya-shrivastavv (commented):
requested changes - FetcherInterface should be an abstract class that can't be instantiated.
Fixed👍

jku (changes requested):
There is a weird extra commit in this branch: I believe it'll go away if you rebase this branch on current origin/develop (and force push the branch) since the change is in origin/develop already.

The PR message is correct and this code does work but 
* class now derives from class ABC instead of using the keyword "metaclass" like the PR message says
* there's a new unused class variable "metaclass": there's no need for that

aditya-shrivastavv:
i have removed unused line `metaclass = abc.ABCMeta` and inside `class FetcherInterface()`.
I have used `class FetcherInterface(metaclass=abc.ABCMeta):`

I think this solved the problems-
- FetcherInterface should be an abstract class that can't be instantiated.
- class now derives from class ABC instead of using the keyword "metaclass"
- removed new unused class variable "metaclass"

jku (approved):
Thanks, looks good to me.

I'll merge this as is but as advice for future: with the merge commit that I'll create we will end up with four commits for this two-line change, that's quite a lot. Your 3 commits could have been rebased and squashed into a single commit. Some maintainers don't mind unneeded merges and such but I try to keep the git log readable for future developers .

Thanks for the PR!

aditya-shrivastavv:
thanks jku, I will keep that in mind and be extra carefull next time. thank you for guiding me. Looking forward to more contributions😊

PR #2217, opened 2022-12-09
build: Change build dependency pinning strategy

### LINKED ISSUE #2216: windows tests fail
```
py: install_deps> python3 -m pip install -e . -r D:\a\python-tuf\python-tuf/requirements-test.txt
py: failed with python3 is not allowed, use allowlist_externals to allow it
```

This seems to be because we don't actually pin all build dependencies (and tox 4.0 has been released):
```
  python3 -m pip install --upgrade pip
  python3 -m pip install --upgrade tox coveralls
```

jku:
* don't autoupgrade pip: let's consider pip to be part of python platform?
* pin build and tox in new requirements-build.txt: this mostly prevents tox from going to 4.x before we're ready
* use requirements-build.txt as constraint when installing tox or build during CI & CD
* use requirements-build.txt in requiremenets-dev.txt

Note that coveralls is not pinned, not sure if it should be.

Fixes #2216 maybe

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -46,8 +46,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python3 -m pip install --upgrade pip
-          python3 -m pip install --upgrade tox coveralls
+          python3 -m pip install --constraint requirements-build.txt tox coveralls
 
       - name: Run tox (${{ env.TOXENV }})
         # See TOXENV environment variable for the testenv to be executed here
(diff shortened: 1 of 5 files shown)


jku:
> Constraints are only allowed to take the form of a package name and a version specifier

Tried to be too smart

joshuagl (approved):

jku:
Thanks. I'll merge this now so PR builds aren't all red: we might still want to revisit the specific solution

PR #2230, opened 2022-12-18
Fix pydocstyle D401: first line should be in imperative mood

fridex:

Make sure the codebase is more compliant with Python's docstring convention.


lukpueh:
See https://github.com/theupdateframework/python-tuf/pull/2231#pullrequestreview-1222490755

jku (approved):
I have no strong opinions on whether to have a linter for this but the fixes seem good and unlikely to be too annoying in git blame so I'm approving and merging these.

PR #2232, opened 2022-12-18
Adjust docstring to show example code

fridex:

Format example code in docstring.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -487,8 +487,8 @@ class Signed(metaclass=abc.ABCMeta):
     def expires(self) -> datetime:
         """The metadata expiry date::
 
-        # Use 'datetime' module to e.g. expire in seven days from now
-        obj.expires = utcnow() + timedelta(days=7)
+        >>> # Use 'datetime' module to e.g. expire in seven days from now
+        >>> obj.expires = utcnow() + timedelta(days=7)
         """
         return self._expires
 


tuf/api/metadata.py
```diff
@@ -487,8 +487,8 @@ def _type(self) -> str:
     def expires(self) -> datetime:
         """The metadata expiry date::
 
-        # Use 'datetime' module to e.g. expire in seven days from now
-        obj.expires = utcnow() + timedelta(days=7)
+        >>> # Use 'datetime' module to e.g. expire in seven days from now
+        >>> obj.expires = utcnow() + timedelta(days=7)
```
lukpueh, line 491: The double colon (`::`) plus blank line makes Sphinx format this as code block. E.g. from our [RTD page](https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.metadata.root.html#tuf.api.metadata.Root.expires):
Screen Shot 2022-12-19 at 09 33 51

What's the reason for this patch? 

(IIRC, the prompt is used by `doctest` to identify snippets to run, but we don't us that at the moment.)

fridex: Thanks for review - I was not aware of the syntax with double colon. I think we can close the PR then.

lukpueh: I also thought that indentation of the block would be needed (see last bullet point from our [docstrings guidelines](https://github.com/secure-systems-lab/code-style-guidelines/blob/master/python.md#docstrings), and the [sphinx docs they point to](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks))

But apparently it works without. 🤷

PR #2233, opened 2022-12-18
Provide __all__ for tuf.ngclient module

fridex:

Document public API for `tuf.ngclient` module.

### DIFF
--- a/tuf/ngclient/__init__.py
+++ b/tuf/ngclient/__init__.py
@@ -7,3 +7,9 @@
 from tuf.ngclient.config import UpdaterConfig
 from tuf.ngclient.fetcher import FetcherInterface
 from tuf.ngclient.updater import Updater
+
+__all__ = [
+    FetcherInterface.__name__,
+    Updater.__name__,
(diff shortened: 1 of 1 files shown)


fridex:
It looks like the public API is not documented in the whole python-tuf project. Would community be interested in providing `__all__` to explicitly document exported API (I can try to find some time to work on it)?

fridex:
Superseded by #2234.

fridex:
Reopening based on discussion in https://github.com/theupdateframework/python-tuf/issues/2235#issuecomment-1357332357

jku (approved):
I'm ok with this: I think the API does not change at all here (since the default is that "import *" gives you everything in `__init__.py`) but I think being explicit is fine.

Other further API clarifications we could do:
* export TargetFile along with Updater here (as it's a major part of Updater API)
* rename updater.py into `_updater.py` to make the API crystal clear

the first of those seems obviously correct... the second one is on the edge of "is it worth the unlikely but possible breakage"

PR #2243, opened 2022-12-20
tuf: move INFO logs to DEBUG or WARNING

woodruffw:
This should reduce some default logging noise when TUF's ngclient is used as a library in other applications (in particular, `sigstore-python`).

See: https://github.com/sigstore/sigstore-python/pull/351

cc @jku 

Signed-off-by: William Woodruff <william@trailofbits.com>

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -443,7 +443,7 @@ class Metadata(Generic[T]):
                 key.verify_signature(delegated_metadata, signed_serializer)
                 signing_keys.add(key.keyid)
             except exceptions.UnsignedMetadataError:
-                logger.info("Key %s failed to verify %s", keyid, delegated_role)
+                logger.warning("Key %s failed to verify %s", keyid, delegated_role)
 
         if len(signing_keys) < role.threshold:
             raise exceptions.UnsignedMetadataError(
@@ -776,7 +776,7 @@ class Key:
             SerializationError,
         ) as e:
             # Log unexpected failure, but continue as if there was no signature
-            logger.info("Key %s failed to verify sig: %s", self.keyid, str(e))
+            logger.warning("Key %s failed to verify sig: %s", self.keyid, str(e))
             raise exceptions.UnsignedMetadataError(
                 f"Failed to verify {self.keyid} signature"
             ) from e
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -443,7 +443,9 @@ class Metadata(Generic[T]):
                 key.verify_signature(delegated_metadata, signed_serializer)
                 signing_keys.add(key.keyid)
             except exceptions.UnsignedMetadataError:
-                logger.warning("Key %s failed to verify %s", keyid, delegated_role)
+                logger.warning(
+                    "Key %s failed to verify %s", keyid, delegated_role
+                )
 
         if len(signing_keys) < role.threshold:
             raise exceptions.UnsignedMetadataError(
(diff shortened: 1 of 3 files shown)


tuf/api/metadata.py
```diff
@@ -443,7 +443,7 @@ def verify_delegate(
                 key.verify_signature(delegated_metadata, signed_serializer)
                 signing_keys.add(key.keyid)
             except exceptions.UnsignedMetadataError:
-                logger.info("Key %s failed to verify %s", keyid, delegated_role)
+                logger.warning("Key %s failed to verify %s", keyid, delegated_role)
```
jku: warning is not required, debug is fine here (this just means one of the signing keys didn't sign the metadata: overall validity of the metadata will be handled two lines later)

jku (changes requested):
> This should reduce some default logging noise when TUF's ngclient is used as a library in other applications (in particular, sigstore-python).

Thanks. I think I agree with this... but I will note that 
a) sigstore-python is the one setting a non-default log level (WARNING is default IIRC)
b) maybe there is documentation on what should be considered INFO  and what DEBUG but I haven't found any. I think your interpretation is that INFO is something that can be exposed to end users and DEBUG is not: that sounds reasonable but it also means libraries then essentially have one log level available to them :shrug:

That one warning should be debug, otherwise LGTM.

woodruffw:
> a) sigstore-python is the one setting a non-default log level (WARNING is default IIRC)

Oh, you're completely right. I'm not sure why we set `INFO` as our default level; that might be an artifact of me originally intending to use the logger for more regular outputs.

tuf/api/metadata.py
woodruffw: Done!

tuf/api/metadata.py
```diff
@@ -776,7 +778,7 @@ def verify_signature(
             SerializationError,
         ) as e:
             # Log unexpected failure, but continue as if there was no signature
-            logger.info("Key %s failed to verify sig: %s", self.keyid, str(e))
+            logger.debug("Key %s failed to verify sig: %s", self.keyid, str(e))
```
jku: That's not actually the warning I commented about : this one I'd be fine as a warning -- here the situation could be a configuration error or a bug so a warning seems reasonable... but the Metadata as a whole may still be verified so we just raise UnsignedMetadataError for this specific sig so that verify_delegate can handle it.

tuf/api/metadata.py
```diff
@@ -443,7 +443,9 @@ def verify_delegate(
                 key.verify_signature(delegated_metadata, signed_serializer)
                 signing_keys.add(key.keyid)
             except exceptions.UnsignedMetadataError:
-                logger.info("Key %s failed to verify %s", keyid, delegated_role)
+                logger.warning(
```
jku: This should not be a warning, debug is fine: one of the keys just didn't have a valid sig for this metadata. The metadata may still be signed by enough keys

jku (changes requested):
I think you modified the wrong warning

tuf/api/metadata.py
woodruffw: Ugh, my bad. Will fix now.

jku (approved):
looks good to me. I'm not 100% sure what the correct way to deal with logger.info() is (as this approach essentially means that libraries are only allowed to log at `debug`, and `info` essentially does not exist for libraries: this feels wrong) but I agree this change is in line with other logger calls.

We likely debug-log a little too much at the moment in ngclient but these are all useful logs so that's another issue.

woodruffw:
Yeah, agreed -- I need to re-think how `sigstore-python` does loglevels more generally (it probably makes sense for `INFO` to be the default for first-party code, `DEBUG` for third-party, etc.) 

I'm going to file an issue for that.

PR #2246, opened 2022-12-22
Support DSSE in Metadata API

lukpueh:
**Notes to reviewer:**
- based on: [#2165](https://github.com/theupdateframework/python-tuf/pull/2165) and [sslib#487](https://github.com/secure-systems-lab/securesystemslib/pull/487)
  (tests should pass locally, when installing tuf and securesystemslib from these branches)
- relevant diff: https://github.com/lukpueh/tuf/compare/no-key...dsse
- leaving as draft until above PRs are merged and, in the case of securesystemslib, released
- the design is smart (h/t @PradyumnaKrishna), but adds quite some complexity to metadata API
- this needs better tests and documentation, but I wanted to get a first round of feedback before spending more time on it
- the commit messages have many details

Adds a TUF-specific [DSSE](https://github.com/secure-systems-lab/dsse) (`Envelope`) implementation and defines an abstract interface (`BaseMetadata`) for common `Envelope` and `Metadata` operations. `BaseMetadata` can be used independently of the underlying metadata type to read or write from and to file or bytes, and to  create or verify signatures. 

**caveat:** Optional arguments may be different for common `Envelope` and `Metadata` methods. This is because they require de/serialization operations at different stages. Envelope must deserialize, when accessing the payload (`get_signed`), but Metadata not. Metadata, OTOH, must serialize to canonical json when signing or verifying (`sign` and `verify_delegate`), but Envelope not.

This PR also adopts the new securesystemslib de/serialization infrastructure, which, most notably, 
- separates basic JSON bytes deserialization from custom object instantiation, 
- adds a `SerializationMixin`, which provides convenience methods to read or write from and to file or bytes 

**caveat:**  users of the serialization mixin must implement methods that return default de/serializers that can be used by the mixin.

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -52,8 +52,7 @@ from typing import (
 
 from securesystemslib import exceptions as sslib_exceptions
 from securesystemslib import hash as sslib_hash
-from securesystemslib import keys as sslib_keys
-from securesystemslib.signer import Signature, Signer
+from securesystemslib.signer import Key, Signature, Signer
 from securesystemslib.storage import FilesystemBackend, StorageBackendInterface
 from securesystemslib.util import persist_temp_file
 
@@ -61,7 +60,6 @@ from tuf.api.exceptions import LengthOrHashMismatchError, UnsignedMetadataError
 from tuf.api.serialization import (
     MetadataDeserializer,
     MetadataSerializer,
-    SerializationError,
     SignedSerializer,
 )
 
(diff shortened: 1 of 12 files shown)


tuf/api/serialization/__init__.py
```diff
@@ -15,45 +15,29 @@
...
 
 from tuf.api.exceptions import RepositoryError
 
 if TYPE_CHECKING:
     # pylint: disable=cyclic-import
     from tuf.api.metadata import Metadata, Signed
 
+MetadataSerializer: TypeAlias = BaseSerializer
+MetadataDeserializer: TypeAlias = BaseDeserializer
+SignedSerializer: TypeAlias = BaseSerializer
```
lukpueh, line 34: I created these aliases to not break the Metadata API.

tuf/api/metadata.py
```diff
@@ -83,7 +82,138 @@
...
+            self.signatures.clear()
+
+        return BaseEnvelope.sign(self, signer)
+
+    def verify_delegate(
+        self,
+        delegated_role: str,
+        delegated_metadata: "BaseMetadata",
+    ) -> None:
+        signed = self.get_signed(None)
```
lukpueh, line 204: The  `deserializer` argument in `get_signed` should be configurable, e.g. via optional `delegator_deserializer` argument  in `verify_delegate`.

Or maybe `verify_delegate` just should not be a method on the envelope, but rather on the already deserialized payload. At least for DSSE it does not seem like the right place to deserialize the payload here.

tuf/api/metadata.py
```diff
@@ -83,7 +82,138 @@
...
+class BaseMetadata(SerializationMixin, JSONSerializable, metaclass=abc.ABCMeta):
+    """A common metadata interface for Envelope (DSSE) and Metadata objects."""
+
+    @staticmethod
+    def _default_deserializer() -> BaseDeserializer:
+        """Default deserializer for Serialization Mixin."""
+        # pylint: disable=import-outside-toplevel
+        from tuf.api.serialization.json import JSONDeserializer
+
+        return JSONDeserializer()
```
lukpueh, line 94: Using the current implementation of `JSONDeserializer` as default deserializer for any `BaseMetadata` (i.e.  `Metadata` and `Envelope`) can lead to unexpected behavior, because it can deserialize both traditional metadata and dsse:

https://github.com/theupdateframework/python-tuf/blob/2bc97c41ecb844a52cd516f143678b9ba40dd821/tuf/api/serialization/json.py#L51-L55

This is the intended behavior for `BaseMetadata`, but a bug on either `Metadata` or `Envelope`:

```python
#  Each `from_file` call below uses the default `JSONDeserializer` returned
#  from `_default_deserializer()`, and thus can return both `Metadata` or `Envelope`
BaseMetadata.from_file("path/to/any/metadata")  # returning `Metadata` or `Envelope` is expected
Metadata.from_file("path/to/any/metadata")  # should only return `Metadata`
Envelope.from_file("path/to/any/metadata")  # should only return `Envelope` 
```

lukpueh (commented):
@jku raised a valid concern about whether there is any merit in a common base class for `Metadata` and `Envelope` in the python-tuf metadata API.

Coming from the [in-toto dsse integration work](https://github.com/in-toto/in-toto/pull/503) I had the following workflows in mind:
```
# metadata consumer
C1. load and verify metadata envelope
C2. extract payload and use it
```
```
# metadata producer
P1. create payload and wrap in metadata envelope
P2. sign and dump
```
A common base class is helpful for applications to reduce case handling for common operations (C1, P2). Users of the python-tuf metadata API, however, can be explicit about which type of metadata envelope they want to load, verify, sign, dump.

Also see inline comments for issues with common operations having different requirements...

lukpueh:
Closing in favour of #2385

PR #2250, opened 2022-12-27
ngclient: Remove "Optional" from helper props

### LINKED ISSUE #2247: ngclient: reconsider use of Optional
TrustedMetadataSet has properties like this:
```python
    @property
    def timestamp(self) -> Optional[Metadata[Timestamp]]:
        """Current timestamp ``Metadata`` or ``None``"""
        return self._trusted_set.get(Timestamp.type)

```
I think the return value should not be Optional: it leads to unneeded use of assert later on to get static typing working (see updater.py for examples). Instead 
* the property should raise a RuntimeError if the property is accessed when the value (in this case timestamp) is not available -- this is a programming error
* the cases where code checks `if trusted_set.timestamp`, it can use the dictionary directly and check for `if Timestamp.type in trusted_set` instead -- I think that should work

jku:
The properties in the internal TrustedMetadataSet are a bit difficult to use with static typing since they return Optional but in many cases we know the "None"-case is impossible.

Remove None from annotation: the idea is that calling the property getter too early is a programming error: it will result in KeyError which, even if not obvious, is consistent since the properties are short hand for the dictionary lookup:
 * trusted_set["timestamp"] raises KeyError if timestamp is not set
 * trusted_set.timestamp raises KeyError if timestamp is not set

Fixes #2247

requirements**:

### DIFF
--- a/tuf/ngclient/_internal/trusted_metadata_set.py
+++ b/tuf/ngclient/_internal/trusted_metadata_set.py
@@ -117,19 +117,19 @@ class TrustedMetadataSet(abc.Mapping):
         return self._trusted_set[Root.type]
 
     @property
-    def timestamp(self) -> Optional[Metadata[Timestamp]]:
-        """Get current timestamp ``Metadata`` or ``None``"""
-        return self._trusted_set.get(Timestamp.type)
+    def timestamp(self) -> Metadata[Timestamp]:
+        """Get current timestamp ``Metadata``"""
+        return self._trusted_set[Timestamp.type]
 
     @property
-    def snapshot(self) -> Optional[Metadata[Snapshot]]:
-        """Get current snapshot ``Metadata`` or ``None``"""
-        return self._trusted_set.get(Snapshot.type)
+    def snapshot(self) -> Metadata[Snapshot]:
+        """Get current snapshot ``Metadata``"""
+        return self._trusted_set[Snapshot.type]
 
     @property
-    def targets(self) -> Optional[Metadata[Targets]]:
-        """Get current targets ``Metadata`` or ``None``"""
-        return self._trusted_set.get(Targets.type)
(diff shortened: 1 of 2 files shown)


joshuagl (approved):
This is a nice improvement in terms of code readability, +1 from me. Perhaps we can leave the PR open until next week to allow for others to review?

lukpueh (approved):
Your rationale for this patch seems sound. LGTM!

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -149,7 +149,7 @@ def update_root(self, data: bytes) -> Metadata[Root]:
         Returns:
             Deserialized and verified root ``Metadata`` object
         """
-        if self.timestamp is not None:
+        if Timestamp.type in self._trusted_set:
```
lukpueh, line 152: Out of curiosity, have you considered
```suggestion
        if Timestamp.type in self:
```
? Given the custom `__getitem__`, we don't have to access the internal `_trusted_set` directly anywhere else in this class.

lukpueh: For context, this mildly put me off at first, because I thought your patch exposes the internal `_trusted_set`. But I realized that it does not and that the instance behaves just like the internal variable.

jku: yeah, I went back and forth with that one -- maybe it would have been clearer with just `self`

lukpueh: IMO it's fine to access the protected attribute inside the class. I was just confused, because updater instances also have a `_trusted_set`  attribute, but that one is actually a `TrustedMetadataSet` instance. :D

PR #2252, opened 2023-01-02
tox: Add python3 to allow list, bump tox to 4.x

### LINKED ISSUE #1791: tox warns on windows that python3 is not installed in testenv
The following warning can be found in current GitHub Windows runner build logs.

```
WARNING: test command found but not installed in testenv
  cmd: C:\hostedtoolcache\windows\Python\3.8.10\x64\python3.EXE
  env: D:\a\python-tuf\python-tuf\.tox\py
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above!
```

This is because we tell tox to run `python3` in the tox.ini commands section, see e.g.:
https://github.com/theupdateframework/python-tuf/blob/b2704a56a8c5023ffd7d377d8a8e72fb26a46a8a/tox.ini#L18

But Windows does not seem to install a `python3` command in the corresponding virtual environment, so it uses the system python3.exe, which [tox warns about if we don't whitelist it](https://tox.wiki/en/latest/config.html#conf-allowlist_externals).

**Possible fixes:**

- Configure GitHub windows runner to install python3 in virtual environment
  NOTE: I didn't check if we also get this warning for `python`(as we did before #1493), if we don't we could change back `python3` to `python` because that latter should run the python version for which the virtual environment was created anyway, which are only version 3 pythons.
- Whitelist python3 in tox.ini (I wouldn't do that, because we do want to use the python of the dedicated virtual environment and not any system python)

### LINKED ISSUE #2218: tests: Update to tox 4.x
I had to come up with `requirements-build.txt` to pin tox since windows test runs started failing: tox 4.x has been released and it says:
```
py: install_deps> python3 -m pip install -e . -r D:\a\python-tuf\python-tuf/requirements-test.txt
py: failed with python3 is not allowed, use allowlist_externals to allow it
```

So tox is now pinned to 3.27.1 (using "pip install --constraints"):
* should figure out what the right fix is to above error so we can upgrade tox
* then decide what to do with tox dependency and the other changes I did in https://github.com/theupdateframework/python-tuf/commit/b6c3b66ca68951dc127b2450a3323f46095ff551: I think it's correct but it was also a quick fix

jku:
This should enable tox4 and fix #1791, #2218

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,4 +1,4 @@
 # The build and tox versions specified here are also used as constraints
 # during CI and CD Github workflows
 build==0.9.0
-tox==3.28.0
+tox==4.1.2
--- a/tox.ini
+++ b/tox.ini
@@ -27,6 +27,9 @@ deps =
 
 install_command = python3 -m pip install {opts} {packages}
 
+# Workaround https://github.com/tox-dev/tox/issues/2801 (python3 not allowed in Windows)
+allowlist_externals = python3
+
 # Develop test env to run tests against securesystemslib's master branch
 # Must to be invoked explicitly with, e.g. `tox -e with-sslib-master`
 [testenv:with-sslib-master]


joshuagl (approved):
Nice. Thanks!

jku:
Based on the answer to the tox bug report I made (closed as not a bug): it sounds like the issue is that
* the "virtualenv" does not contain python3 on windows but does on other platforms
* tox developers reasonably don't consider this a bug in tox

I suppose I'll file an issue on virtualenv?

lukpueh (approved):

PR #2273, opened 2023-01-15
enhancement: Add from_data() method to MetaFile

### LINKED ISSUE #1893: MetaFile is missing from_bytes() or equivalent

VickyMerzOwn:
Signed-off-by: Satvik Vemuganti <vemugantisesha@iitbhilai.ac.in>

Fixes #1893 

There already exists a `TargetFile.from_data()` / `TargetFile.from_file()` to calculate hashes and length for a target.
This update enables support for hashes and length for Metafile too.
This PR introduces a static method in `BaseFile` as suggested by @jku in #2186. Also taking from the same comment, the `MetaFile.from_file()` method has not been implemented.

**The pull request fulfills the following
requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -34,6 +34,7 @@ from tuf.api.metadata import (
     Delegations,
     Key,
     Metadata,
+    MetaFile,
     Root,
     Signature,
     Snapshot,
@@ -724,6 +725,29 @@ class TestMetadata(unittest.TestCase):
         # Test with no algorithms specified
         targetfile_from_data = TargetFile.from_data(target_file_path, data)
         targetfile_from_data.verify_length_and_hashes(data)
+
+    def test_metafile_from_data(self) -> None:
+        data = b"Inline test content"
+
+        # Test with a valid hash algorithm
+        metafile = MetaFile.from_data(1, data, ["sha256"])
+        metafile.verify_length_and_hashes(data)
+
+        # Test with an invalid hash algorithm
+        with self.assertRaises(ValueError):
+            metafile = MetaFile.from_data(1, data, ["invalid_algorithm"])
+            metafile.verify_length_and_hashes(data)
+
+        self.assertEqual(
+            metafile,
+            MetaFile(
+                1,
+                19,
+                {
+                    "sha256": "fcee2e6d56ab08eab279016f7db7e4e1d172ccea78e15f4cf8bd939991a418fa"
+                },
+            ),
+        )
 
     def test_targetfile_get_prefixed_paths(self) -> None:
         target = TargetFile(100, {"sha256": "abc", "md5": "def"}, "a/b/f.ext")
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -900,6 +900,41 @@ class BaseFile:
         if length < 0:
             raise ValueError(f"Length must be >= 0, got {length}")
 
+    @staticmethod
+    def _get_length_and_hashes(
+        data: Union[bytes, IO[bytes]], hash_algorithms: Optional[List[str]]
+    ) -> Tuple[int, Dict[str, str]]:
+        """Calculate length and hashes of ``data``."""
+        if isinstance(data, bytes):
+            length = len(data)
+        else:
+            data.seek(0, io.SEEK_END)
+            length = data.tell()
+
+        hashes = {}
+
+        if hash_algorithms is None:
+            hash_algorithms = [sslib_hash.DEFAULT_HASH_ALGORITHM]
+
+        for algorithm in hash_algorithms:
+            try:
+                if isinstance(data, bytes):
+                    digest_object = sslib_hash.digest(algorithm)
+                    digest_object.update(data)
+                else:
+                    digest_object = sslib_hash.digest_fileobject(
+                        data, algorithm
+                    )
+            except (
+                sslib_exceptions.UnsupportedAlgorithmError,
+                sslib_exceptions.FormatError,
+            ) as e:
+                raise ValueError(f"Unsupported algorithm '{algorithm}'") from e
+
+            hashes[algorithm] = digest_object.hexdigest()
+
+        return (length, hashes)
+
 
 class MetaFile(BaseFile):
     """A container with information about a particular metadata file.
@@ -965,6 +1000,28 @@ class MetaFile(BaseFile):
 
         # All fields left in the meta_dict are unrecognized.
         return cls(version, length, hashes, meta_dict)
+
+    @classmethod
+    def from_data(
+        cls,
+        version: int,
+        data: Union[bytes, IO[bytes]],
+        hash_algorithms: List[str],
+    ) -> "MetaFile":
+        """Creates MetaFile object from bytes.
+        This constructor should only be used if hashes are wanted.
+        By default, MetaFile(ver) should be used.
+        Args:
+            version: Version of the metadata file.
+            data: Metadata bytes that the metafile represents.
+            hash_algorithms: Hash algorithms to create the hashes with. If not
+            specified, the securesystemslib default hash algorithm is used.
+        Raises:
+            ValueError: The hash algorithms list contains an unsupported
+            algorithm.
+        """
+        length, hashes = cls._get_length_and_hashes(data, hash_algorithms)
+        return cls(version, length, hashes)
 
     def to_dict(self) -> Dict[str, Any]:
         """Return the dictionary representation of self."""
@@ -1693,34 +1750,7 @@ class TargetFile(BaseFile):
             ValueError: The hash algorithms list contains an unsupported
                 algorithm.
         """
-        if isinstance(data, bytes):
-            length = len(data)
-        else:
-            data.seek(0, io.SEEK_END)
-            length = data.tell()
-
-        hashes = {}
-
-        if hash_algorithms is None:
-            hash_algorithms = [sslib_hash.DEFAULT_HASH_ALGORITHM]
-
-        for algorithm in hash_algorithms:
-            try:
-                if isinstance(data, bytes):
-                    digest_object = sslib_hash.digest(algorithm)
-                    digest_object.update(data)
-                else:
-                    digest_object = sslib_hash.digest_fileobject(
-                        data, algorithm
-                    )
-            except (
-                sslib_exceptions.UnsupportedAlgorithmError,
-                sslib_exceptions.FormatError,
-            ) as e:
-                raise ValueError(f"Unsupported algorithm '{algorithm}'") from e
-
-            hashes[algorithm] = digest_object.hexdigest()
-
+        length, hashes = cls._get_length_and_hashes(data, hash_algorithms)
         return cls(length, hashes, target_file_path)
 
     def verify_length_and_hashes(self, data: Union[bytes, IO[bytes]]) -> None:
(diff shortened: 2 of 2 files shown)


tests/test_api.py
```diff
@@ -725,6 +726,22 @@ def test_targetfile_from_data(self) -> None:
...
         targetfile_from_data.verify_length_and_hashes(data)
 
+    def test_metafile_from_data(self) -> None:
+         data = b"Inline test content"
+
+         # Test with a valid hash algorithm
+         metafile = MetaFile.from_data(
+             1, data, ["sha256"]
+         )
+         metafile.verify_length_and_hashes(data)
```
jku: I think this does not actually test that the length and hashes are really used  (as the method by design succeeds even if length and hashes are not set) so that might be useful to verify.

Maybe an additional assert that e.g. `metafile.verify_length_and_hashes(b"")` fails? 

(No need to do anything more extensive as `test_length_and_hash_validation()` is extensive: this would just confirm that length and hashes are there after `MetaFile.from_data()`)

jku (commented):
Looks good to me, thank you. I suggested one additional assert in the test, what do you think?

Let me know either way, I think this is fine as is if you don't feel like adding the assert.

tests/test_api.py
VickyMerzOwn: Okay, that makes sense. However, I'm not sure how I could add such an assertion. 
Would adding the following do the job? Simply:
```py
assert type(data) == bytes
assert data != b""
```
Pardon me, this is not at all ideal, and we're better off as it is, since `test_length_and_hash_validation()` anyways covers the said case.

jku: the case that isn't theoretically covered is if (now or in a future change), the Metafile actually did not store the length and hashes: then the call just succeeds regardless of the content.

My original suggestion was a sanity check to ensure the provided length/hashes are actually used in verify_length_and_hashes:
```python
# assert that provided length is actually used in verify 
with self.assertRaises(LengthOrHashMismatchError):
    metafile.verify_length_and_hashes(b"")
```

**But** the other option (that might actually be more readable in the test) would be to compare the metafile you produced to one made by the other constructor

```python
self. assertEqual(
    metafile,
    MetaFile(1, 19, {'sha256': 'fcee2e6d56ab08eab279016f7db7e4e1d172ccea78e15f4cf8bd939991a418fa'})
)
```

I think the latter one would make a lot of sense sense and should be understandable to the test reader. What do you think?

VickyMerzOwn: Okay, now I get the point.
Yes, the latter assertion seems more readable. At least, it helped me understand what was happening :-)
Do I update it or would you rather commit on top of my PR after it is merged? (since it was your idea)

jku:
I just noticed the failing docs build: 

> tuf/api/metadata.py:docstring of tuf.api.metadata.MetaFile.from_data:7: ERROR: Unexpected indentation.

Could you have a look at that -- it's likely sphincs being very strict about docstrings indents. You can run docs build on your machine with `tox -e docs` .

tests/test_api.py
jku: I was about to ...  but then noticed the failing docs build (see other comment). please go ahead and add the test as well if it's not too much trouble

VickyMerzOwn: No trouble at all. Doing it!

jku (changes requested):
Looks good to me but the there are still failing checks on CI:
* linter currently fails (just formatting issues), see `tox -e lint` 
* the second commit message needs a signed-off-by (feel free to `--amend` the commit and then use `--force` when pushing your branch)

VickyMerzOwn:
I apologize for the delays in this.
`pylint -j 0 --rcfile=pyproject.toml tuf examples tests verify_release` is returning the following:
```
************* Module examples.manual_repo.succinct_hash_bin_delegations
examples/manual_repo/succinct_hash_bin_delegations.py:28:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module examples.repository._simplerepo
examples/repository/_simplerepo.py:14:0: E0611: No name 'Key' in module 'securesystemslib.signer' (no-name-in-module)
examples/repository/_simplerepo.py:14:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module examples.manual_repo.basic_repo
examples/manual_repo/basic_repo.py:30:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module examples.manual_repo.hashed_bin_delegation
examples/manual_repo/hashed_bin_delegation.py:26:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module examples.uploader._localrepo
examples/uploader/_localrepo.py:15:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module tests.test_metadata_eq_
tests/test_metadata_eq_.py:15:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module tests.repository_simulator
tests/repository_simulator.py:57:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module tuf.api.metadata
tuf/api/metadata.py:56:0: E0611: No name 'Key' in module 'securesystemslib.signer' (no-name-in-module)
tuf/api/metadata.py:441:19: E1101: Module 'securesystemslib.exceptions' has no 'UnverifiedSignatureError' member (no-member)
tuf/api/metadata.py:905:4: C0116: Missing function or method docstring (missing-function-docstring)
************* Module tests.generated_data.generate_md
tests/generated_data/generate_md.py:11:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
************* Module tests.test_api
tests/test_api.py:27:0: E0611: No name 'SSlibKey' in module 'securesystemslib.signer' (no-name-in-module)
tests/test_api.py:212:31: E1101: Module 'securesystemslib.exceptions' has no 'VerificationError' member (no-member)
tests/test_api.py:233:31: E1101: Module 'securesystemslib.exceptions' has no 'VerificationError' member (no-member)
tests/test_api.py:267:31: E1101: Module 'securesystemslib.exceptions' has no 'VerificationError' member (no-member)
tests/test_api.py:275:31: E1101: Module 'securesystemslib.exceptions' has no 'VerificationError' member (no-member)
tests/test_api.py:283:31: E1101: Module 'securesystemslib.exceptions' has no 'VerificationError' member (no-member)
tests/test_api.py:288:31: E1101: Module 'securesystemslib.exceptions' has no 'UnverifiedSignatureError' member (no-member)

-------------------------------------------------------------------
Your code has been rated at 9.79/10 (previous run: 10.00/10, -0.21)```

However, all the above exceptions are available in the said file. Is it because the right version of`securesystemslib` isn't being picked? If so, how do I correct it?

**Comment by jku** [comment]
Yes, requirements have changed and the development setup may not follow automatically.

If you are running with tox (like `tox -e lint`) it should just work automatically but deleting the tox cache (rm -r .tox/) and re-running the command is a good way to resolve issues.

If you have installed the dependencies on your own environment, then re-installing should work: e.g. reinstalling dev requirements: `pip install -r requirements/dev.txt`

**Comment by VickyMerzOwn** [comment]
One particular action has been failing. It runs: `pylint -j 0 --rcfile=pyproject.toml tuf examples tests verify_release` but fails throwing an error with the message: `Error: Process completed with exit code 16.`
https://github.com/VickyMerzOwn/python-tuf/actions/runs/4282237925/jobs/7456261085

**Comment by jku** [comment]
the actual error is
`tuf/api/metadata.py:905:4: C0116: Missing function or method docstring (missing-function-docstring)`

What's interesting is that the other internal functions (that also do not have docstrings) don't seem to trigger this. I have no idea why.

Anyway: adding a docstring will fix this. After fixing this you will see a further linting issue: the `from_data()` method and the internal method are annotated to accept `List[str]` as the hash algorithms but they should likely accept `Optional[list[str]]`: this is what the method docstring seems to imply.

**Comment by VickyMerzOwn** [comment]
Cool! Would you like me to add a docstring for all the other methods too?

Also, would all my commits have to be signed? I'm not sure how to edit an older commit's message, but I'll try to figure it out if necessary. (I mean because there are merge commits too. The guidelines [here](https://github.com/theupdateframework/python-tuf/pull/2273/checks?check_run_id=11636122625) say I have to be the only author)

**Comment by jku** [comment]
Sorry, missed this yesterday.

> Cool! Would you like me to add a docstring for all the other methods too?

Definitely not in this PR. In general, if there are methods that are not obvious to a reader, then a PR with docstrings even for private methods would be welcome.

> Also, would all my commits have to be signed? I'm not sure how to edit an older commit's message, but I'll try to figure it out if necessary. (I mean because there are merge commits too. The guidelines here say I have to be the only author)

All commits with your contributions need a signed-off-by, merge commits do not (although we'd prefer no unnecessary merge commits in git history).

In this case I would suggest you squash all commits in this branch to a single commit (and make sure that one has a signed-off-by). 

Here's a potential way to do it:
```bash
# reset commits (--soft keeps changes in the index)
# this assumes origin is the  https://github.com/theupdateframework/python-tuf/ upstream repo
git reset --soft $(git merge-base origin/develop HEAD)
# now do the One Commit
git commit -s
# check results and then push your branch with `--force` (because you are overwriting your 11 commits with one)
```

lukpueh:
Hi @VickyMerzOwn, do you have bandwidth to wrap up this PR?

VickyMerzOwn:
Hi @lukpueh: Really sorry about this.
Is it okay if I create a new Pull Request?

VickyMerzOwn:
Please merge it if it's okay.

Another question:
I deleted my local clone of the repo a while back while clearing my machine.
So after cloning, I ran `python3 -m pip install -r requirements/dev.txt`
After this I ran `tox`, followed by `python3 aggregate_tests.py`

Not sure if it was a result of both commands, but a while later when I checked my `git status`, there were 190+ untracked files.

VickyMerzOwn:
There are merge commits included here. Is it alright?

lukpueh (approved):
Thanks, @VickyMerzOwn! LGTM

lukpueh:
Merging despite ci failure, which is unrelated, and fixed in #2447...

PR #2277, opened 2023-01-25
Expose default fetcher

### LINKED ISSUE #2268: ngclient `RequestsFetcher`: make timeout part of the public API

`TrustUpdater` in the `ngclient` APIs can be given a `RequestsFetcher` to use or, if one is not supplied, will construct a default one.

`RequestsFetcher` in turn has a default socket timeout (via `self.socket_timeout`) of 4 seconds, per TUF's mitigations against "slow retrieval"-style attacks. 

Unfortunately, this timeout might be too conservative for some real-world network scenarios: CI providers like GitHub Actions will occasionally de-prioritize network traffic, causing failures in tools like `sigstore-python` that use TUF to bootstrap their roots of trust.

We have a temporary workaround for this as part of https://github.com/sigstore/sigstore-python/pull/432, where we plan to use a timeout of 30s (under the operating theory that this is more than acceptably permissive for GitHub Actions and other CI providers, but still short enough/a hard-failing error mode.) However, that patch requires us to import an `_internal` API and set an attribute directly, which isn't ideal 🙂 

**Current behavior**:

`RequestFetcher` is a private API.

**Expected behavior**:

`RequestFetcher` becomes part of the `ngclient` public API surface, and (optionally) has a kwarg in its constructor for configuring the `socket_timeout`. The current approach of just modifying the attribute is also perfectly fine, given that it isn't currently marked as private.

jku:
This is useful for those who want to use the default fetcher but modify some attributes

The file itself could be moved to tuf/ngclient/ but this is not done yet as sigstore-python is using this internal module. Move can be done once sigstore-python 1.0 is no longer relevant.

Fixes #2268

requirements**:

### DIFF
--- a/tuf/ngclient/__init__.py
+++ b/tuf/ngclient/__init__.py
@@ -4,12 +4,18 @@
 """TUF client public API
 """
 
+
+# requests_fetcher is public but comes from _internal for now (because
+# sigstore-python 1.0 still uses the module from there). requests_fetcher
+# can be moved out of _internal once sigstore-python 1.0 is not relevant.
+from tuf.ngclient._internal.requests_fetcher import RequestsFetcher
 from tuf.ngclient.config import UpdaterConfig
 from tuf.ngclient.fetcher import FetcherInterface
 from tuf.ngclient.updater import Updater
 
 __all__ = [
     FetcherInterface.__name__,
+    RequestsFetcher.__name__,
     Updater.__name__,
     UpdaterConfig.__name__,
 ]
--- a/docs/api/tuf.ngclient.fetcher.rst
+++ b/docs/api/tuf.ngclient.fetcher.rst
@@ -1,6 +1,9 @@
 Fetcher
 ============
 
-.. automodule:: tuf.ngclient.fetcher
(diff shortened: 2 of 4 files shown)


jku:
[docs](https://theupdateframework--2277.org.readthedocs.build/en/2277/api/tuf.ngclient.fetcher.html)

tuf/ngclient/__init__.py
```diff
@@ -4,12 +4,18 @@
 """TUF client public API
 """
 
+
+# requests_fetcher is public but comes from _internal for now (because
+# sigstore-python 1.0 still uses the module from there). requests_fetcher
+# can be moved out of _internal once sigstore-python 1.0 is not relevant.
```
joshuagl, line 10: should a similar comment be made in the requests_fetcher module, for people looking at the code directly without seeing context here?

joshuagl (approved):
All looks good to me. I made a minor suggestion, but feel free to ignore.

tuf/ngclient/__init__.py
jku: Maybe yes

lukpueh (approved):

PR #2231, opened 2022-12-18
Fix pydocstyle D400: first line should end with a period

fridex:
**Description of the changes being introduced by the pull request**:

Make sure the codebase is more compliant with Python's docstring convention. There are more D400 warnings by pydocstyle left. To make this change easier to review, it tries to introduce more straightforward changes to the codebase.

PR #2278, opened 2023-01-26
Fix pydocstyle D400: first line should end with a period

fridex:

Make sure the codebase is more compliant with Python's docstring convention. There are more D400 warnings by pydocstyle left. To make this change easier to review, it tries to introduce more straightforward changes to the codebase.

Rebased #2231


lukpueh (approved):
Thanks!

PR #2279, opened 2023-01-26
Export TUF TargetFile in ngclient

fridex:
Signed-off-by: Fridolin Pokorny <fridolin.pokorny@datadoghq.com>

See https://github.com/theupdateframework/python-tuf/pull/2233#pullrequestreview-1230483501

### DIFF
--- a/tuf/ngclient/__init__.py
+++ b/tuf/ngclient/__init__.py
@@ -5,6 +5,8 @@
 """
 
 
+from tuf.api.metadata import TargetFile
+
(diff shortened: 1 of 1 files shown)


jku (approved):
Happy to have other maintainer opinions, but I think this makes sense: the ngclient API includes TargetFile (as a method return value and argument) so it makes sense that the ngclient module also "exports" TargetFile -- this way you can use ngclient without importing Metadata API.

PR #2282, opened 2023-01-26
Fix pydocstyle D400

fridex:
pydocstyle D400: First line should end with a period

Signed-off-by: Fridolin Pokorny <fridolin.pokorny@datadoghq.com>

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -397,7 +397,9 @@ class Metadata(Generic[T]):
         delegated_metadata: "Metadata",
         signed_serializer: Optional[SignedSerializer] = None,
     ) -> None:
-        """Verify that ``delegated_metadata`` is signed with the required
+        """Verify that ``delegated_metadata`` is signed with the required threshold.
+
+        Verify that ``delegated_metadata`` is signed with the required
         threshold of keys for the delegated role ``delegated_role``.
 
         Args:
@@ -564,8 +566,9 @@ class Signed(metaclass=abc.ABCMeta):
     def _common_fields_from_dict(
         cls, signed_dict: Dict[str, Any]
     ) -> Tuple[int, str, datetime]:
-        """Return common fields of ``Signed`` instances from the passed dict
-        representation, and returns an ordered list to be passed as leading
+        """Return common fields of ``Signed`` instances from the passed dict representation.
+
+        Returns an ordered list to be passed as leading
         positional arguments to a subclass constructor.
 
         See ``{Root, Timestamp, Snapshot, Targets}.from_dict`` methods for usage.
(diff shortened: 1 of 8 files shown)


tuf/api/metadata.py
```diff
@@ -1572,8 +1577,7 @@ def get_roles(self) -> Iterator[str]:
             yield f"{self.name_prefix}-{suffix}"
 
     def is_delegated_role(self, role_name: str) -> bool:
-        """Determine whether the given ``role_name`` is in one of
-        the delegated roles that ``SuccinctRoles`` represents.
+        """Determine whether the given ``role_name`` is in one of the delegated roles represented by ``SuccinctRoles``.
```
lukpueh, line 1580: This exceeds our usual line length of 80 quite a bit. Not sure if we have agreed on a limit for docstrings. Apparently we are not enforcing one. Can pydocstyle help with this? I don't have very strong feelings about the exact limit, but I think it makes sense to align(-ish) with the rest of the code.

tuf/api/metadata.py
```diff
@@ -397,7 +397,9 @@ def verify_delegate(
         delegated_metadata: "Metadata",
         signed_serializer: Optional[SignedSerializer] = None,
     ) -> None:
-        """Verify that ``delegated_metadata`` is signed with the required
+        """Verify that ``delegated_metadata`` is signed with the required threshold.
+
+        Verify that ``delegated_metadata`` is signed with the required
```
lukpueh, line 402: I'm not super fond of verbatim duplicates. Vertical space in source code files is valuable.  :) Maybe we can get a bit creative in cases like these.

tuf/api/metadata.py
fridex: I see black running which should have 88 chars default, but it looks like it works on docstrings in general. What about enabling [flake8](https://flake8.pycqa.org/en/2.5.5/index.html) as well? There are few violations reported, but could be fixable.

tuf/api/metadata.py
fridex: Sounds very reasonable. I'll try to push creativity. 👍🏻

But please, as I'm not that familiar with python-tuf codebase, it will require more in-depth review not to adjust semantics.

tuf/api/metadata.py
lukpueh: We configure 80 for black (see pyproject.toml). Also, black doesn't touch docstrings. I wonder why pylint doesn't complain. 

Regardless, I'm fine with adding another linter (flake8), but it's not a priority. I think the current combo of linters+formatters works quite well.

fridex: Okay, so let's keep it without flake8 in that case.

lukpueh:
Hi @fridex, do you have bandwidth to wrap up this PR?

jku:
Closing for now to cleanup issue list, please reopen if you get to it later.

PR #2289, opened 2023-02-02
Requirements refactor

jku:
* Split test and lint requirements (to allow lint tools to be more strict about Python versions than our testing is)
* move requirements files to a subdir (to keep top level dir clean)
* upgrade isort now that lint requirements are separate. Closes #2286 

Moving the requirements files is my preference but it's not something I feel strongly about: either way works.


jku:
My understanding is that dependabot will find the requirements files anywhere in the tree but we should double check the logs after merging

lukpueh (approved):
LGTM!

Minor nit: For the sake of horizontal density, should we go for something like...
```
requirements/
├── dev.txt
├── main.txt
└── ...
```
? (I've seen other projects do that)

jku:
> LGTM!
> 
> Minor nit: For the sake of horizontal density, should we go for something like...
> 
> ```
> requirements/
> ├── dev.txt
> ├── main.txt
> └── ...
> ```
> 
> ? (I've seen other projects do that)

sure why not. I assume you mean `requirements.txt` becomes `main.txt`.

jku:
rebased

jku:
> My understanding is that dependabot will find the requirements files anywhere in the tree but we should double check the logs after merging

Dependency Review actually prints these out: LGTM
https://github.com/theupdateframework/python-tuf/actions/runs/4103814736

PR #2290, opened 2023-02-03
build: Handle GH release manually

### LINKED ISSUE #2212: build: Stop depending on action-gh-release
We use softprops/action-gh-release to upload release artifacts to GitHub releases. This makes the action a fairly critical part of our release pipeline. The latest release of action-gh-release includes, among other things, thousands of lines of changes to package-lock.json: Yay npm.

So I'm wondering if we can replace the action.  I think  `await github.rest.repos.createRelease(...)` using actions/github-script should do it... but someone would have to go through the trouble of testing that.

jku:
Remove dependency on softprops/action-gh-release: instead do the GitHub release steps using the GitHub API and github-script.

The only difference should be that release name is not "<tag>-rc" first: instead the initial release is marked as draft in the API (and shows as draft in the GitHub UI).

Fixes #2212 

I've tested this in my fork, seems to work.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -49,7 +49,7 @@ jobs:
     permissions:
       contents: write # to modify GitHub releases
     outputs:
-      release_id: ${{ steps.gh-release.outputs.id }}
+      release_id: ${{ steps.gh-release.outputs.result }}
     steps:
       - name: Fetch build artifacts
         uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -58,14 +58,32 @@ jobs:
           path: dist
 
       - id: gh-release
-        name: Publish GitHub release candidate
-        uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5
+        name: Publish GitHub release draft
+        uses: actions/github-script@d556feaca394842dc55e4734bf3bb9f685482fa0
```
jku: this is probably the previous release of github-script

.github/workflows/cd.yml
```diff
@@ -93,12 +111,9 @@ jobs:
...
           script: |
-            await github.rest.repos.updateRelease({
+            github.rest.repos.updateRelease({
               owner: context.repo.owner,
               repo: context.repo.repo,
               release_id: '${{ needs.candidate_release.outputs.release_id }}',
-              name: '${{ github.ref_name }}',
-              body: 'See [CHANGELOG.md](https://github.com/' +
-                     context.repo.owner + '/' + context.repo.repo +
-                    '/blob/${{ github.ref_name }}/docs/CHANGELOG.md) for details.'
```
lukpueh, line 103: I hope not passing a body does not override the body passed above? I could not validate this assumption by looking at https://github.com/jku/test-releases/actions.

.github/workflows/cd.yml
```diff
@@ -93,12 +111,9 @@ jobs:
         uses: actions/github-script@98814c53be79b1d30f795b907e553d8679345975
         with:
           script: |
-            await github.rest.repos.updateRelease({
```
lukpueh, line 96: No need to wait, because we don't do anything with the return value?

lukpueh (approved):
Cool stuff!

.github/workflows/cd.yml
jku: It does work correctly AFAICT: I removed the testing tags I used in my fork because I'm afraid of accidentally pushing them to upstream (removing the tag seems to then remove the release), but I did run the release process and it looked good to me.

.github/workflows/cd.yml
jku: that's what I was thinking ... Now that you point it out, I wonder if this changes the return value in edge cases here -- if the call fails for a network error or something?

I'm fine with it either way: the only thing that happens in failure case is that the GH release would still be a draft: that's easy to notice and easy to fix

PR #2293, opened 2023-02-05
Updated the ci to fail if lint failed

### LINKED ISSUE #1248: GitHub Actions: Find and implement fail-fast strategy

#1242 added a GitHub Actions based CI workflow, which runs all configured builds in parallel, ignoring each others build results. However, GitHub Actions provides an option (i.e. `fail-fast`), to abort all builds if any build fails. This option might be useful for specific build relationships, where one failing build is likely to entail other failing builds.

**Current behavior**:
`fail-fast: false`

**Expected behavior**:
Consider `fail-fast: true` (default) but with specific conditions, e.g. prioritize `lint` build and abort any other build if that one fails.

shabeebk:
that is provided, the better.

Fixes #1248

The CI pipeline has been changed to run the lint tasks in the first and fail if the lint job fails.

requirements**:


jku:
There seems to be something wrong with the workflow. By the way, you can test CI changes in your own fork as well -- you just have to enable actions in the settings of your fork, then you can got the actions tab, choose the CI workflow, click "run workflow" and select "issue-1248" branch.

shabeebk:
@jku Thanks for reviewing it. I have fixed it now and ran successfully in my repository.

Screen Shot 2023-02-06 at 9 57 15 AM

jku:
I think I'll leave review for lukas (or I'll have to try to understand this better) but I will mention that the strange "waiting for status to be reported" is just because some checks are marked as required in project configuration: the check names have changed so the UI looks confused.

lukpueh:
Thanks for the patch, @shabeebk! Have you tried if this works as [you intended](https://github.com/theupdateframework/python-tuf/issues/1248#issuecomment-1418285289)? Specifically:
> 1. Run the lint job as the first job in the matrix.

I don't think that your patch changes anything about the order in which jobs run. IIUC they still run in parallel depending on available resources.

shabeebk:
Thanks, @lukpueh @jku for reviewing and for the feedback. I have updated the job to run lint separately before starting the test job. Added a snap of the job tree below.

The job is written in such a way that it will proceed only if lint succeeds.

eg: https://github.com/shabeebk/python-tuf/actions/runs/4301754458

Screenshot from 2023-03-01 01-03-45

.github/workflows/_test.yml
```diff
@@ -5,24 +5,54 @@ on:
...
 
 jobs:
+  lint-test:
+    name: Lint Test
+    strategy:
+      matrix:
+        python-version: [3.x]
+
+    env:
+      TOXENV: lint
```
lukpueh:
```suggestion
```
No need for this. You only run one toxenv on one python-version. You can specify them directly below.

.github/workflows/_test.yml
```diff
@@ -5,24 +5,54 @@ on:
...
+          python3 -m pip install --constraint requirements-build.txt tox coveralls
+
+      - name: Run tox (${{ env.TOXENV }})
+        # See TOXENV environment variable for the testenv to be executed here
+        run: tox
+
   tests:
     name: Tests
+    needs: lint-test
+    continue-on-error: ${{ matrix.failure-allowed }}
```
lukpueh, line 32: I don't think you need continue-on-error anymore, or do you?

.github/workflows/_test.yml
```diff
@@ -5,24 +5,54 @@ on:
...
     strategy:
-      fail-fast: false
       # Run regular TUF tests on each OS/Python combination, plus special tests
       # (sslib master) and linters on Linux/Python3.x only.
```
lukpueh: Comment is not fully accurate anymore (the "and linters" part)

.github/workflows/_test.yml
```diff
@@ -75,7 +103,7 @@ jobs:
     # Always run when all 'tests' jobs have finished even if they failed
     # TODO: Replace always() with a 'at least one job succeeded' expression
     if: always()
-    needs: tests
+    needs: [lint-test, tests]
```
lukpueh: Shouldn't this work as is by transitivity? If "coveralls-fin" needs "tests" and "tests" needs "lint-test", then 
"coveralls-fin" needs "lint-test" without specifying it explicitly, right?

lukpueh (changes requested):
Yeah, I think this is what we want. Many thanks for the continued efforts. I left a few minor comments, please address and we can merge this.

Oh, and please also sign-off all commits so that the DCO check passes.

.github/workflows/_test.yml
shabeebk: Yes, that is to keep the job running if anyone of them fails.

.github/workflows/_test.yml
shabeebk: Yes, you are right. Just fixed it.

.github/workflows/_test.yml
shabeebk: Updated. Thanks for pointing that out. :smiley:

.github/workflows/_test.yml
shabeebk: That's right. Updated to follow your thoughts.

shabeebk:
Closed in favour of a cleaner PR #2317

PR #2321, opened 2023-03-06
Repository annotations

### LINKED ISSUE #2311: potential repository design tweaks
Not sure how I'd like to solve these but I'll list two issues that I have with current repository API:
* lack of complete typing in `with Repository.edit(rolename) as role` is quite annoying -- I think I want 
  ```python
  with Repository.edit_root() as root
      # root is type Root
  with Repository.edit_targets(rolename) as targets
      # targets is type Targets
  ```
  versions for timestamp and snapshot could be added for completeness but aren't as important: they should mostly get edited through snapshot() and timestamp() anyway.
* it turns out that most of the time I'm not _editing_, I just want to lookup things in the metadata... but I would typically still like direct access to the signed property like `Repository.edit()` gives me. Maybe something like 
  ```python
  root = Repository.root()
  # root is type Root
  targets = Repository.targets(rolename)
  # targets is type Targets
  ``` 
This would create a nice balance with the context managers `with Repository.edit_root()` and `with Repository.edit_targets(rolename)`

jku:
Extend Repository API with
1. Read-only accessors for correctly typed Signed objects: 
   ```python
   root() -> Root
   timestamp() -> Timestamp
   snapshot() -> Snapshot
   targets(role: Optional[str]) -> Targets
   ```
1. typed versions of the edit() context manager:
   ```python
   edit_root()
   edit_timestamp()
   edit_snapshot()
   edit_targets(role: Optional[str])
   ```
This requires renaming old `timestamp()/snapshot()` to `do_timestamp()` and `do_snapshot()`.

The advantages are:
* less magic strings ("targets", "root") in application code
* more annotation coverage in application (and python-tuf) code
* mostly more readable code

Fixes #2311

requirements**:

### DIFF
--- a/tuf/repository/_repository.py
+++ b/tuf/repository/_repository.py
@@ -107,7 +107,7 @@ class Repository(ABC):
         """Context manager for editing root metadata. See edit()"""
         with self.edit(Root.type) as root:
             if not isinstance(root, Root):
-                raise RuntimeError("Unexpected Root type")
+                raise RuntimeError("Unexpected root type")
             yield root
 
     @contextmanager
@@ -115,7 +115,7 @@ class Repository(ABC):
         """Context manager for editing timestamp metadata. See edit()"""
         with self.edit(Timestamp.type) as timestamp:
             if not isinstance(timestamp, Timestamp):
-                raise RuntimeError("Unexpected Timestamp type")
+                raise RuntimeError("Unexpected timestamp type")
             yield timestamp
 
     @contextmanager
@@ -123,7 +123,7 @@ class Repository(ABC):
         """Context manager for editing snapshot metadata. See edit()"""
         with self.edit(Snapshot.type) as snapshot:
             if not isinstance(snapshot, Snapshot):
-                raise RuntimeError("Unexpected Snapshot type")
+                raise RuntimeError("Unexpected snapshot type")
             yield snapshot
 
     @contextmanager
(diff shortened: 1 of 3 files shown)


tuf/repository/_repository.py
```diff
@@ -94,7 +103,71 @@ def edit(self, role: str) -> Generator[Signed, None, None]:
...
+        self, rolename: str = Targets.type
+    ) -> Generator[Targets, None, None]:
+        """Context manager for editing targets metadata. See edit()"""
+        with self.edit(rolename) as targets:
+            if not isinstance(targets, Targets):
+                raise RuntimeError(f"Unexpected targets ({rolename}) type")
+            yield targets
+
+    def root(self) -> Root:
+        """Read current root metadata"""
```
jku, line 141: Not totally happy about mentioning metadata here but that seems unavoidable -- hopefully the lower case "metadata" and strict typing makes it clear what is being returned

jku:
rebased without changes (to update tests to what GH branch protection rules expect at the moment)

lukpueh (approved):
I haven't tried using this, but based on a pure code review this looks great.

Using the role names as getters and prefixing snapshot/timestamp action methods with `do_` definitely feels intuitive.

jku:
As a followup, the results look pretty good in my repository-playground. The remaining non-annotated cases are all usages of `Repository.open(rolename)` when the role is either being verified or is used as the verifier:
* verifier should not need to be a Metadata -- I think this was a small mistake in the Metadata: verify_delegate() should really be a Signer-level method. We should consider moving it (no problem leaving a deperecated version in Metadata for a while)
* the repository should maybe offer some verification functionality -- like `Repository.verify(rolename, delegating_rolename)`. This would hide the details from the application

PR #2338, opened 2023-03-20
Add TargetFile.get_prefixed_paths()

### LINKED ISSUE #2166: Allow generation of hash-prefixed versions of `TargetFile`s

jku:
Fixes #2166. 

This is the branch from #2180, with my improvements on top. It's a simple method but the edge cases are not obvious so it makes sense for python-tuf to handle this -- if applications do, then we end up with a many slightly different implementations.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -731,6 +731,36 @@ class TestMetadata(unittest.TestCase):
         )
         self.assertEqual(sorted(target.get_prefixed_paths()), ["public/path/abc.file.ext", "public/path/def.file.ext"])
 
+        target = TargetFile(
+            100, {"sha256": "abc", "md5": "def"}, ""
+        )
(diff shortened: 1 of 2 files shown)


lukpueh (approved):
I was surprised to see that we don't put any restrictions on target paths. Other than that, LGTM!

Issue #2370, opened 2023-04-27
Use PyPI Trusted publishing


jku:
https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/

We already do releases from GitHub CD workflow but this new PyPI feature allows us to remove the long lived secret from GitHub environment secrets: instead the GH action uses a short lived secret it receives from pypi.org using the OIDC identity of our workflow and release environment.

TODO:
- [x] add the release workflow as a trusted publisher in pypi.org
- [x] Add `id-token: write` permission to the publish task in the workflow (#2371)
- [x] Remove the secret use from the workflow (#2371)
- [x] Remove the secret from the environment
- [x] Remove the secret in PyPI (this is likely in my pypi account)

jku:
* release worked (or rather, the issues noticed were unrelated to this change)
* I've removed the token authorization from my pypi account 
* The token has also been deleted from the GH environment secrets

PR #2371, opened 2023-04-27
release: Use PyPI Trusted Publishing

jku:
Instead of using the secret stored in environment secrets, allow the publish action to use the OIDC identity to authenticate to pypi.org. This repository/workflow/environment combo has been marked as a "Trusted Publisher" in pypi.org: this means PyPI should give the publish action a short lived token to use for publishing.

This enables #2370: but the secret should still be removed before closing the issue (maybe after one successful release with Trusted Publishing).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -92,6 +92,7 @@ jobs:
     environment: release
     permissions:
       contents: write # to modify GitHub releases
+      id-token: write # to authenticate as Trusted Publisher to pypi.org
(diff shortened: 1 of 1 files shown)


lukpueh (approved):
LGTM! Cross-checked with [PyPI docs](https://docs.pypi.org/trusted-publishers/) and our PyPI project config.

This requires: https://github.com/theupdateframework/python-tuf/pull/2376 to work.

PR #2378, opened 2023-05-05
Move verify_delegate() to Root/Targets

### LINKED ISSUE #2361: move verify_delegate() from Metadata to Signed
verify_delegate() is currently an instance method of Metadata. It could be an instance method of Signed since it does not need anything from metadata and since this would lead to cleaner code.

* Generally code that uses tuf.repository can mostly look clean as it does not need to deal with the actual `Metadata`:
    ```
    root = repo.root()
    print (f"root version is {root.version}")
    ```
    This no longer works if one wants to verify a delegate just because of API design
* We could add a `Signed.verify_delegate()` and similar code should work:
    ```
    root = repo.root()
    root.verify_delegate("targets", metadata)
    ```
* nothing prevents leaving `Metadata.verify_delegate()` in place as a deprecated method

jku:
Fixes #2361 

* Add an internal `_Delegator` class that implements verify_delegate()
* Inherit that class in Root, Targets: the internal class is just an implementation detail for sharing the implementation, not something users of the library should care about
* Document Metadata.verify_delegate() as deprecated (and use the new implementation internally)
* Use the new method in tests, examples and ngclient

This will make both ngclient and external repository code slightly easier to read and write (but note that the ngclient cleanups are not included here as they touch quite a lot of lines even if they are simple).

There are several ways to add verify_delegate() to Root and Targets, I think this helper class is a decent compromise between modularity and simplicity: 
* the _Delegator class isn't intended to be part of a longer "inheritance path" (and does not have a matching element in the metadata format): it just provides an extra method for Root and Targets. If there's a naming style for this style of inheritance I'm happy to use it
* _Delegator is internal: users of the library don't need to care about it

requirements**:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -364,29 +364,49 @@ class TestMetadata(unittest.TestCase):
         role2 = role2_md.signed
 
         # test the expected delegation tree
-        root.verify_delegate(Root.type, root_md)
-        root.verify_delegate(Snapshot.type, snapshot_md)
-        root.verify_delegate(Targets.type, targets_md)
-        targets.verify_delegate("role1", role1_md)
-        role1.verify_delegate("role2", role2_md)
+        root.verify_delegate(
+            Root.type, root_md.signed_bytes, root_md.signatures
+        )
+        root.verify_delegate(
+            Snapshot.type, snapshot_md.signed_bytes, snapshot_md.signatures
+        )
+        root.verify_delegate(
+            Targets.type, targets_md.signed_bytes, targets_md.signatures
+        )
+        targets.verify_delegate(
+            "role1", role1_md.signed_bytes, role1_md.signatures
+        )
+        role1.verify_delegate(
+            "role2", role2_md.signed_bytes, role2_md.signatures
+        )
 
         # only root and targets can verify delegates
         with self.assertRaises(AttributeError):
-            snapshot.verify_delegate(Snapshot.type, snapshot_md)
+            snapshot.verify_delegate(
+                Snapshot.type, snapshot_md.signed_bytes, snapshot_md.signatures
+            )
         # verify fails for roles that are not delegated by delegator
         with self.assertRaises(ValueError):
-            root.verify_delegate("role1", role1_md)
+            root.verify_delegate(
+                "role1", role1_md.signed_bytes, role1_md.signatures
+            )
         with self.assertRaises(ValueError):
-            targets.verify_delegate(Targets.type, targets_md)
+            targets.verify_delegate(
+                Targets.type, targets_md.signed_bytes, targets_md.signatures
+            )
         # verify fails when delegator has no delegations
         with self.assertRaises(ValueError):
-            role2.verify_delegate("role1", role1_md)
+            role2.verify_delegate(
+                "role1", role1_md.signed_bytes, role1_md.signatures
+            )
 
         # verify fails when delegate content is modified
         expires = snapshot.expires
         snapshot.expires = expires + timedelta(days=1)
         with self.assertRaises(exceptions.UnsignedMetadataError):
-            root.verify_delegate(Snapshot.type, snapshot_md)
+            root.verify_delegate(
+                Snapshot.type, snapshot_md.signed_bytes, snapshot_md.signatures
+            )
         snapshot.expires = expires
 
         # verify fails if sslib verify fails with VerificationError
@@ -395,12 +415,16 @@ class TestMetadata(unittest.TestCase):
         good_sig = snapshot_md.signatures[keyid].signature
         snapshot_md.signatures[keyid].signature = "foo"
         with self.assertRaises(exceptions.UnsignedMetadataError):
-            root.verify_delegate(Snapshot.type, snapshot_md)
+            root.verify_delegate(
+                Snapshot.type, snapshot_md.signed_bytes, snapshot_md.signatures
+            )
         snapshot_md.signatures[keyid].signature = good_sig
 
         # verify fails if roles keys do not sign the metadata
         with self.assertRaises(exceptions.UnsignedMetadataError):
-            root.verify_delegate(Timestamp.type, snapshot_md)
+            root.verify_delegate(
+                Timestamp.type, snapshot_md.signed_bytes, snapshot_md.signatures
+            )
 
         # Add a key to snapshot role, make sure the new sig fails to verify
         ts_keyid = next(iter(root.roles[Timestamp.type].keyids))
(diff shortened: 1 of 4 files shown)


jku:
> If there's a naming style for this style of inheritance I'm happy to use it

A "Mixin" is likely what I'm trying to remember here?

jku:
I think this is ready for review. I have a branch [cleanup-after-verify-delegate-move](https://github.com/jku/python-tuf/commits/cleanup-after-verify-delegate-move) with changes based on this PR... but they are not required so I'll do those separately.

tuf/api/metadata.py
```diff
@@ -1926,7 +1962,7 @@ def get_delegated_role(self, delegated_role: str) -> Role:
     def get_key(self, keyid: str) -> Key:
         """Return the key object for the given keyid.
 
-        Raises ValueError if keyid is not found.
+        Raises ValueError if key is not found.
```
jku: this change just makes the two implementations have the same docstring.

tuf/api/metadata.py
```diff
@@ -675,7 +635,83 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """
+        raise NotImplementedError
+
+    def verify_delegate(
+        self,
+        delegated_role: str,
+        delegated_metadata: "Metadata",
+        signed_serializer: Optional[SignedSerializer] = None,
+    ) -> None:
```
jku: Maybe a meaningful note here is that this makes `Targets` and `Root` "aware" of `Metadata` now.

The alternative would be for the arguments to include payload bytes and signatures so something like 

    def verify_delegate(
        self,
        delegated_role: str,
        payload: bytes,
        signatures: Dict[str, Signature]
    )

Unfortunately I'm not sure how one would make that usable in calling code

tuf/api/metadata.py
```diff
@@ -396,56 +396,16 @@ def verify_delegate(
...
-            delegated_metadata: ``Metadata`` object for the delegated role
-            signed_serializer: Serializer used for delegate
-                serialization. Default is ``CanonicalJSONSerializer``.
-
-        Raises:
-            UnsignedMetadataError: ``delegated_role`` was not signed with
-                required threshold of keys for ``role_name``.
-            ValueError: no delegation was found for ``delegated_role``.
-            TypeError: called this function on non-delegating metadata class.
+        .. deprecated:: 2.2.0
```
lukpueh: Is this the correct version number?

tuf/api/metadata.py
```diff
@@ -675,7 +635,83 @@ def to_dict(self) -> Dict[str, Any]:
...
+        raise NotImplementedError
+
+    def verify_delegate(
+        self,
+        delegated_role: str,
+        delegated_metadata: Metadata,
+        signed_serializer: Optional[SignedSerializer] = None,
+    ) -> None:
+        """Verify that ``delegated_metadata`` is signed with the required
+        threshold of keys for the delegated role ``delegated_role``.
```
lukpueh: Off-topic and not important, but here's a chance for leaner docs, which also makes it consistent with the first part of the sentence.
```suggestion
        threshold of keys for the ``delegated_role``.
```

tuf/api/metadata.py
lukpueh: You could actually remove docstrings for `get_key` and `get_delegated_role` in `Targets` and `Root`, since they can be reused from their definitions in the parent class. Pylint would need to be told to not complain. So maybe not worth it, the docstrings are short enough.

tests/test_api.py
```diff
@@ -345,44 +345,83 @@ def test_metadata_verify_delegate(self) -> None:
         with self.assertRaises(ValueError):
             role2.verify_delegate("role1", role1)
 
+    # pylint: disable=too-many-locals,too-many-statements
```
lukpueh: I think this disables pylint downwards on class scope and not on function scope. Probably not what you intended, but also no big deal.

lukpueh (changes requested):
This change makes sense. I pointed out some nits that you may ignore. The only thing that should be changed is `test_api.TestMetadata.test_metadata_verify_delegate`, which still uses `Metadata.verify_delegate`. Otherwise, the tests pass even without that method.

Also, I think the version number in the deprecation notice is wrong (see inline comment).

tuf/api/metadata.py
jku: I was sure that wouldn't work with sphinx but I tried and you are right.

jku:
> `test_api.TestMetadata.test_metadata_verify_delegate`, which still uses `Metadata.verify_delegate`

I left a small test in place to make sure I didn't break that one -- just enough testing to make sure it's connected to the actual implementation in Signed. 

`test_signed_verify_delegate()` contains all of the tests that existed before.

jku:
```
lint: commands[1]> isort --check --diff tuf examples tests verify_release
ERROR: /home/runner/work/python-tuf/python-tuf/verify_release Imports are incorrectly sorted and/or formatted.
--- /home/runner/work/python-tuf/python-tuf/verify_release:before	2023-05-15 08:14:43.101068
+++ /home/runner/work/python-tuf/python-tuf/verify_release:after	2023-05-15 08:15:05.396047
@@ -19,9 +19,8 @@
 from typing import Optional
 
 try:
+    import build as _
     import requests
-
-    import build as _
 except ImportError:
     print("Error: verify_release requires modules 'requests' and 'build':")
     print("    pip install requests build")
lint: exit 1 (0.27 seconds) /home/runner/work/python-tuf/python-tuf> isort --check --diff tuf examples tests verify_release pid=1912
```

The code that is failing on CI is exactly the code that my isort is suggesting I use.

Same exact isort version
 
 (╯°□°)╯︵ ┻━┻

jku:
I've dropped the commit that would make lint work on my machine... CI doesn't seems to mind :shrug: 

Let's maybe not merge this yet though: I still feel like we might be designing ourselves into a corner wrt DSSE.

Marking draft while thinking about it...

lukpueh:
> The alternative would be for the arguments to include payload bytes and signatures so something like
> 
> ```
> def verify_delegate(
>     self,
>     delegated_role: str,
>     payload: bytes,
>     signatures: Dict[str, Signature]
> )
> ```
> 
> Unfortunately I'm not sure how one would make that usable in calling code

I have a patch for this on top of your (rebased) branch. It also adds an alias for default canonical signed bytes to Metadata. This makes the calling code only slightly more complicated: https://github.com/theupdateframework/python-tuf/compare/635a2870bde3c44b4eab9e66be61c77e08e47285...lukpueh:move-verify-delegate-v2

I think this is a reasonable path towards #2385. @jku, feel free to take over my branch for this PR, or let me know if you want me to force push.

lukpueh (approved):

lukpueh:
@jku, you may want to approve this pr too, now that it's a collaboration.

jku:
It looks like original author cannot click "approve"... but I think this looks correct and reasonable.

PR #2379, opened 2023-05-05
readthedocs: Specify build image

jku:
RTD docs build is failing because the default image has openssl that is incompatible with current urllib3: Specify a newer image.


jku:
I don't like hard coding os and python versions since we'll likely never update these... but we should be able to revert this commit at some point when RTD updates the default image.

lukpueh (approved):

PR #2387, opened 2023-05-09
Release python-tuf 3.0.0

### LINKED ISSUE #2382: New release
It's been a while since the last python-tuf release, so maybe time to bake one. I have not looked through the git log, not sure yet if this is a minor or point release.

Potential blockers:
* #2371 (and #2376) -- release infra improvement, would make sense to review & merge
* #2377 -- just a testing dependency update but a fairly major one
* anything else?

not blockers IMO:
* #2363 offline client work would be nice to get into a testable state but it's not ready yet
* #2378 It's close to ready but still a draft and so has not had review

jku:
* Update changelog
* Bump version

This fixes #2382

This is a major version change because of the `Key.from_securesystemslib_key()` removal and `Key.verify_signature()` method signature change: The latter is _almost_ an internal method but the former may be used by repository code.

I did not try to include every tiny change in the Changelog: but feel free to point out anything that you think should be mentioned.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,4 +1,37 @@
 # Changelog
+
+## v3.0.0
+
+The notable change in this release is #2165: The tuf.api.metadata.Key
+class implementation was moved to Securesystemslib with minor API
+changes. These changes require no action in tuf.ngclient users but may
+require small changes in tuf.api.metadata using repository
+implementations that create keys.
+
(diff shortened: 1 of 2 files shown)


lukpueh (approved):
Thanks for preparing this! I checked all PRs since the last release, and agree with your selection for the changelog.

Let's see if our updated release pipeline works... 🚀

PR #2389, opened 2023-05-10
github: Fix issue with draft releases

### LINKED ISSUE #2388: verify_release: Fix issue with draft releases no showing up
I think at some point I made the verify_release use "draft:true" when uploading the draft release, instead of modifying the name.
This was a mistake as already documented by lukas in #1933:

> It would be nice if we could set the release as draft at first, but that has the same caveat, i.e. it requires a token and special knowledge to download the assets.

so what happens is that while the release is a draft `./verify_release -- skip-pypi` does not notice that a new GitHub release exists:
```
./verify_release --skip-pypi
* Built release 3.0.0
* WARNING: GitHub latest version is 2.1.0
* ERROR: GitHub artifacts do not match built release
```

Easiest fix would be to revert using "draft:true".

jku:
Commit 707dc49 included a change where the release candidate was marked as draft. This was a mistake as draft releases are only visible to logged in maintainers. This leads to e.g. `./verify_release` script failing while the release is a draft.

Revert those changes:
* don't use "draft" attribute
* postfix the release name with "-rc" while the release waits for approval
* Only set the real description and name after release approval

Fixes #2388

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -66,12 +66,9 @@ jobs:
             res = await github.rest.repos.createRelease({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              name: '${{ github.ref_name }}',
+              name: '${{ github.ref_name }}-rc',
               tag_name: '${{ github.ref }}',
-              draft: true,
-              body: 'See [CHANGELOG.md](https://github.com/' +
-                     context.repo.owner + '/' + context.repo.repo +
-                    '/blob/${{ github.ref_name }}/docs/CHANGELOG.md) for details.'
+              body: 'Release waiting for review...',
             });
 
             fs.readdirSync('dist/').forEach(file => {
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -66,12 +66,9 @@ jobs:
...
               owner: context.repo.owner,
               repo: context.repo.repo,
-              name: '${{ github.ref_name }}',
+              name: '${{ github.ref_name }}-rc',
               tag_name: '${{ github.ref }}',
-              draft: true,
-              body: 'See [CHANGELOG.md](https://github.com/' +
-                     context.repo.owner + '/' + context.repo.repo +
-                    '/blob/${{ github.ref_name }}/docs/CHANGELOG.md) for details.'
+              body: "Release waiting for review...",
```
lukpueh: Minor consistency nit:
```suggestion
              body: 'Release waiting for review...',
```

lukpueh (approved):
This should work.

PR #2411, opened 2023-06-16
change Root.roles to Dict [Issue #2410]

### LINKED ISSUE #2410: Metadata API Root.roles should be a normal dict
Root.roles is currently annotated as `Mapping`. I think the idea was to document that you are not supposed to modify this dictionary.

In practice this just makes life hard in application code (and also Mapping is a purely `typing` module innovation: from 3.10 onwards all of the normal type annotations like dict can be used without the module).

I suggest we make Root.roles a `Dict` like the similar variables in e.g. Targets. This is a compatible change as `Dict` is a `Mapping`.

ryanchen99:
that is provided, the better.

Fixes #2410

Based of the issue in 2410, I change the Root.roles from Mapping to Dict.

requirements**:

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -42,7 +42,6 @@ from typing import (
     Generic,
     Iterator,
     List,
-    Mapping,
     Optional,
     Tuple,
     Type,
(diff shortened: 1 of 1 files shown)


jku (approved):
Looks good to me. This should be fully backwards compat: (Dict is a Mapping).

To recap the events:
* we (I, probably) used Mapping in places to indicate containers that should not be modified. This location is last remaining, others were changed already
* this protection is purely cosmetic of course and does not prevent many other methods of modifying these containers (like modifying the contained things or just swapping out the whole container)
* I now think Mapping is not a good idea for an API like this. It constrains application code in places where you actually would like to modify the container, even if temporarily
* Mapping is also not part of the "standard" type annotations (removing it means  once we require python 3.9 or so we could likely stop using typing module altogether)

PR #2292, opened 2023-02-03
POC: Use securesystemslib base+json de/serializer and mixin

lukpueh:
This PR is meant as a help for python-tuf maintainers to review the serialization-specific parts of https://github.com/secure-systems-lab/securesystemslib/pull/487, which are more tailored towards in-toto and DSSE.

Please see commit message for details.

PR #2436, opened 2023-08-09
Add basic DSSE equivalent for Metadata API and configurable DSSE support in ngclient

### LINKED ISSUE #2385: Support DSSE in client and metadata API.
[TAP 17](https://github.com/theupdateframework/taps/blob/master/tap17.md) proposes detaching the signature wrapper from the TUF specification, in order to allow using e.g. the *[Dead Simple Signing Envelope (DSSE)](https://github.com/secure-systems-lab/dsse/)* instead of the traditional canonicalisation-based container format. This should be supported by python-tuf, more specifically:

- there should be a low-level *metadata API* to read, verify, sign and write DSSE envelopes with TUF payloads, and
- *ngclient* should be able to consume and verify DSSE envelopes with TUF payloads

*NOTE: Below "Envelope" means DSSE and "Metadata" means traditional canonicalisation-based metadata container. "Payload" refers to the inner role metadata that can be wrapped by either Envelope or Metadata.*

**Note on different usage patterns**

In the current metadata API TUF Payloads are typically wrapped in Metadata throughout their lifetime. This means, users create, or load and deserialise the entire Metadata object, and then interact with the contained Payload. Signature verification and creation happens independently over an on-the-fly canonicalised Payload.

A deserialised Envelope object, on the other hand, contains the non-deserialised Payload, so that it can verify signatures over the verbatim bytes, before deserialising the untrusted Payload. Given that no canonicalisation is involved, a Payload cannot reliably be tied back to an existing Envelope and its signatures once it has been deserialised. This means, the user is likely to operate on the Payload alone, discarding the Envelope after signature verification, or only creating it right before signing respectively.

**References and Related work**
* [securesystemslib DSSE implementation](https://github.com/secure-systems-lab/securesystemslib/blob/main/securesystemslib/dsse.py)
* [PR: Adopt DSSE in Metadata API (rejected)](https://github.com/theupdateframework/python-tuf/pull/2246)
  PoC of the securesystemslib DSSE implementation, to be usable interchangeably for Metadata and Envelope for tuf and in-toto. It was rejected for being too over-engineered, but might still be useful as inspiration or anti-example.
* [FEATURE: DSSE bridge (not possible)](https://github.com/theupdateframework/python-tuf/issues/2093)
* [PR: Move verify_delegate() to Root/Targets](https://github.com/theupdateframework/python-tuf/pull/2378/)
  Not directly related to DSSE, but detaches TUF verification implementation from `Metadata`, and thus paves the way to signature wrapper agnosticism.

lukpueh:
*EDIT 2023/08/21: Rebased a.o. on top of #2361*
*EDIT 2023/10/12:*
  - *Updated commit reference after rebase on #2481*
  - *Updated link to demo gist*
  - *Updated PR description to reflect latest state*

*EDIT 2024/02/21:*
  - *Rebase*
  - *Remove client example changes and demo*

---

 
Fixes #2385

**Notes for reviewer**
* The largest chunk of the diff is a pure code move (503680ab580ff9bd9fdaafedc3cb2725f7c1d787)
* The commit messages provide detailed information about changes, but not all changes end up in the final diff
* There is a demo included below

**Background**
The search for a common container abstraction  for DSSE envelope and traditional Metadata has been unsuccessful in the past (see [#2246](https://github.com/theupdateframework/python-tuf/pull/2246), [#2292](https://github.com/theupdateframework/python-tuf/pull/2292)). Some reasons are:
* Although, they look very similar, their usage patterns are different enough to make any common abstraction overly complex
* The life time of the container is too short to justify an overly complex solution. 
* Even though application logic can be mostly agnostic to the container format. There are certain moments, where a deliberate decision for a container format should be made.

**Summary of changes**
- Adds basic DSSE equivalent for Metadata API called SimpleEnvelope API
- Make ngclient Updater configurable to deserialise TUF payloads either using Metadata API or SimpleEnvelope API

~~**Demo**~~ (removed in b6fa05b)

```bash
# Install python-tuf with dsse support
git clone https://github.com/lukpueh/tuf -b add-dsse
cd tuf && pip install -r requirements/dev.txt

# Create and serve a simple static dsse demo repo
curl https://gist.githubusercontent.com/lukpueh/91224e528af5332ef9c779075cf71e06/raw/eb9e5dc1943710fff938674f0eb144b1b9a7edd4/serve_simple_dsse_repo.py \
    | python -

# Run example client in another terminal (same env and dir)
cd examples/client
./client tofu
./client download --use-dsse file1.txt
```


lukpueh:
@jku: This is good for another round of review. The last 5 commits update the PR as we discussed (see commit messages for details). Remaining things I'd like to ticketize:

- Server-side tooling 
  We should figure out how we want to support DSSE in our repo-side tooling, e.g. repository simulator (for tests) and repository abstraction (used in examples). Supporting both simultaneously, as ngclient does with this PR, might not be worth the effort. Also, if we find a solution for repo abstraction first, maybe we can use that in repo simulator.

- API (Docs)
  Currently, payload classes (`Signed`, `Root`, `Targets`), etc. are imported into both Metadata API and Envelope API, to preserve backwards compatibility of Metadata API and make both APIs fully independently usable. If we keep it that way, we need to figure out if we want to document those classes in both namespaces on readthedocs.

examples/client/client
```diff
@@ -146,6 +152,13 @@ def main() -> None:
         help="Target file",
     )
 
+    download_parser.add_argument(
+        "--use-dsse",
+        help="Parse TUF metadata as DSSE",
+        default=False,
+        action="store_true",
+    )
```
jku: While it's really useful to see this client example for the review... I dislike making the client more complex. 

I wonder if we should just drop the example changes, at least until there is a repository example that this can be used against?

tuf/api/dsse.py
```diff
@@ -0,0 +1,158 @@
...
+
+    """
+
+    _DEFAULT_PAYLOAD_TYPE = "application/vnd.tuf+json"
+
+    @property
+    def signatures_dict(self) -> Dict:
+        """Convenience alias for ``self.signatures`` mapped to keyids."""
+        # TODO: Propose changing ``signatures`` list to dict upstream
+        return {sig.keyid: sig for sig in self.signatures}
```
jku, line 63: yeah, the TODO feels right -- even if the serialized format is a list, that's just `list(self.signatures.values())` away if self.signatures is a dict

tests/test_trusted_metadata_set.py
```diff
@@ -471,6 +494,53 @@ def target_expired_modifier(target: Targets) -> None:
 
     # TODO test updating over initial metadata (new keys, newer timestamp, etc)
 
+    def test_load_from_simple_envelope(self) -> None:
+        """Basic unit test for ``_load_from_simple_envelope`` helper.
+
+        TODO: Test via trusted metadata set tests like for traditional metadata
```
jku, line 500: should make a ticket about this I suppose -- or maybe more generally about testing all components with DSSE

jku (approved):
I think this is good to go. 
* I can't yet see how the repository side will work out with Envelope but I also don't think we will find out without merging this and then someone trying it out -- I suspect that unlike ngclient repositories are going to support Metadata or DSSE, not both.
* TrustedMetadataSet refactor is a good idea even without the rest of the PR
* Testing is lacking but I guess the first step towards improving that would be to write a RepositorySimulator for DSSE (or make the existing one support DSSE)

Thanks for the rewrites, I think it looks impressive:
```
$ git diff --stat origin/develop...lukpueh/add-dsse -- tuf/
...
2074 insertions(+), 1813 deletions(-)
```

lukpueh:
> Thanks for the rewrites, I think it looks impressive:

Thanks for the review! Your comments sound reasonable. Let me push another commit and ping for one last look.

> 
> ```
> $ git diff --stat origin/develop...lukpueh/add-dsse -- tuf/
> ...
> 2074 insertions(+), 1813 deletions(-)
> ```

Yes, sorry about that. I'm happy to move the initial commit, which is a pure code move, to a separate PR. Then it's a little less intimidating.

```
git diff --stat 503680a
...
 9 files changed, 564 insertions(+), 160 deletions(-)

```

jku:
I pasted the stats as a positive thing: 250 added LOC for this is not bad :) 

Sure, the _payload.py code move and especially the TrustedMetadataSet refactor could have been separate PRs but I think they're fine here.

tuf/ngclient/config.py
```diff
@@ -5,6 +5,20 @@
...
+@unique
+class EnvelopeType(Flag):
+    """Configures deserialization and verification mode of TUF metadata.
+
+    Args:
+        METADATA: Traditional canonical JSON -based TUF Metadata.
+        SIMPLE: Dead Simple Signing Envelope. (experimental)
+    """
+
+    METADATA = 1
```
trishankatdatadog, line 20:
```suggestion
    CJSON = 1
```

tuf/ngclient/config.py
```diff
@@ -5,6 +5,20 @@
...
+    SIMPLE = 2
```
trishankatdatadog, line 21:
```suggestion
    DSSE = 2
```

trishankatdatadog (commented):
Looks like excellent work (as usual), thanks, Lukas!

Do you think it might be worthwhile to drop support for Canonical JSON as an envelope, and make it DSSE-only by default going forward? We might need a MAJOR release for that and some deprecation notice. Then again, shouldn't people be free to support whatever envelope they like (e.g., JOSE)?

lukpueh:
Thanks for the comments, Trishank!

> Do you think it might be worthwhile to drop support for Canonical JSON as an envelope, and make it DSSE-only by default going forward?

That would be nice, but I fear it would be extremely disruptive for any existing TUF user, if python-tuf (most notably ngclient) just stops supporting the cjson metadata wrapper. 

> Then again, shouldn't people be free to support whatever envelope they like (e.g., JOSE)?

I thinks so yes. Right now users can choose between  "Envelope API" (dsse) or "Metadata API" (cjson) and configure the client to expect one or the other. I'm unsure if python-tuf can/should be changed to support arbitrary wrappers. I tried but couldn't find a good design for that.

jku:
> Then again, shouldn't people be free to support whatever envelope they like (e.g., JOSE)?

This has two really major disadvantages:
* each new option increases software complexity (which is bad for multiple reasons)
* each new option lowers chances of interoperability (an area where TUF has traditionally had significant issues)

These are downsides for DSSE as well, but DSSE balances that by improving other aspects of TUF (deserialization safety and ease of implementation).

The interop argument is also why dropping support for the TUF metadata format is unlikely to be a real option for a very long time.

trishankatdatadog:
Agreed with both of you, and trust that there is no good common API for CJSON and DSSE (let alone other envelopes) in python-tuf right now. We can push down support for other envelopes to other implementations, or later if need be here.

tests/test_trusted_metadata_set.py
lukpueh: see #2565

tuf/api/dsse.py
lukpueh: see #2564

lukpueh:
@jku, I addressed your comments and rebased on a pretty different `develop` compared to before.

Here are two git command that might be helpful to re-review and make sure I didn't mess up while resolving conflicts
```
# Check if the diffs to develop before and after the rebase roughly match 
git range-diff 6fed68b..c96ec55 develop..add-dsse

# Check if all new metadata.py changes indeed moved to _payload.py
git diff develop:tuf/api/metadata.py tuf/api/_payload.py

```

tuf/api/metadata.py
```diff
@@ -29,58 +29,49 @@
...
+    Signed,
+    Snapshot,
+    SuccinctRoles,
+    T,
+    TargetFile,
+    Targets,
+    Timestamp,
+    VerificationResult,
+)
+from tuf.api.exceptions import UnsignedMetadataError
```
jku, line 67: `tuf-api.metadata.LengthOrHashMismatchError` doesn't exist anymore but maybe that's fine

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -407,39 +401,32 @@ def update_delegated_targets(
         # does not match meta version in timestamp
         self._check_final_snapshot()
 
-        delegator: Optional[Metadata] = self.get(delegator_name)
+        delegator: Optional[Union[Root, Targets]] = self.get(delegator_name)
```
jku: nit
```suggestion
        delegator: Optional[Delegator] = self.get(delegator_name)
```

jku (approved):

tuf/api/metadata.py
lukpueh: Good eyes! Can't remember, if I left it out on purpose or forgot. Let me re-export it, just to be safe.

PR #2453, opened 2023-09-02
workflows: Includes version comments in GH action uses-lines

### LINKED ISSUE #2439: Include version comments in GH action uses-lines
Dependabot nowadays updates GH action version number if one is included as a comment. That would be useful.

So in all actions in .github/workflows/*.yml instead of using 

    uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f

we should use:

    uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4

Fixing all uses lines does not have to happen in a single PR

sumanth8495:
Fixes #2439 

- Includes version comments in GH action uses-lines

requirements**:

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -15,6 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v2.3.4
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
       - name: 'Dependency Review'
-        uses: actions/dependency-review-action@f6fff72a3217f580d5afd49a46826795305b63c7 # v2.3.4
+        uses: actions/@f6fff72a3217f580d5afd49a46826795305b63c7 # v3.0.8
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -17,4 +17,4 @@ jobs:
       - name: 'Checkout Repository'
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
       - name: 'Dependency Review'
-        uses: actions/@f6fff72a3217f580d5afd49a46826795305b63c7 # v3.0.8
+        uses: actions/dependency-review-action@f6fff72a3217f580d5afd49a46826795305b63c7 # v3.0.8
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -15,6 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v2.3.4
       - name: 'Dependency Review'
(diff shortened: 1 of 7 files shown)


jku (changes requested):
The version number in my example was meant as an example, not as the value that every action should have.

So when you did this:
> `actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v2.3.4`

I was expecting to see the version number for `actions/checkout` that matches `f43a0e5`. This is visible in e.g. https://github.com/actions/checkout/releases/: **v3.6.0**


`theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master` is a special case that I should have mentioned: no need to comment that one with a release

.github/workflows/scorecards.yml
```diff
@@ -21,10 +21,10 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@08b4669551908b1024bb425080c797723083c031
+        uses: ossf/scorecard-action@08b4669551908b1024bb425080c797723083c031 # v2.2.4
```
jku: bug: This looks like it should be  v2.2.0, no?

.github/workflows/specification-version-check.yml
```diff
@@ -14,8 +14,8 @@ jobs:
     outputs:
       version: ${{ steps.get-version.outputs.version }}
     steps:
-      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
-      - uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 
+      - uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1
```
jku: missing version numbers here

.github/workflows/specification-version-check.yml
```diff
@@ -31,6 +31,6 @@ jobs:
       contents: read
       issues: write
     needs: get-supported-tuf-version
-    uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
+    uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
```
jku: very minor nit: would prefer not to have the extra whitespace in the end

.github/workflows/dependency-review.yml
```diff
@@ -15,6 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
       - name: 'Dependency Review'
-        uses: actions/dependency-review-action@f6fff72a3217f580d5afd49a46826795305b63c7
+        uses: actions/@f6fff72a3217f580d5afd49a46826795305b63c7 # v3.0.8
```
jku: bug: this looks like a typo, missing the name "dependency-review-action"

jku (changes requested):
Thanks, looks pretty good. Looks like you figured out the codeql versioning which always  baffles me...

I found two potential bugs and a couple of missing version comments -- I'd like at least the bugs to be fixed before merging.

jku (approved):
Looks great, thanks for your persistence!

PR #2472, opened 2023-09-27
Offline mode

### LINKED ISSUE #2359: Implementation of Offline mode for TUF

jku:
Fixes #2359, supersedes #2363

### Summary

This is based 2363 code by Emile Baez, Emerson Rounds and omartounsi7 (see first commit). I've fixed some merge issues and linting problems, added tests and finally done some refactoring.

The PR exposes a new config option `offline: bool`. When offline, Updater behaviour changes in two ways:
1. nothing is downloaded during `refresh()`,  `get_target_info()` and `find_cached_target()`: the operations are done using locally available metadata only.
1. Metadata expiry is no longer respected. All other security checks are still in operation

What this means is that an offline Updater can be used when client
* has all the required metadata and targets cached locally (either because it has used Updater in online mode before or because some other process has filled the caches out-of-band)
* is fine with the compromise in security that using potentially revoked keys means

### Implementation notes

These are mostly about changes I did on top of 2363:

* The TrustedMetadataSet argument was renamed: as the component does not know anything about downloads, `offline` seemed less useful. The argument is now `respect_expiry` which I think is accurate. The expiry checks were consolidated in one method: if metadata has expired it either raises or logs a warning based on `respect_expiry` 
* Updater exceptions were standardized: If Updater would have to download something but is offline, it now raises DownloadError
* Testing feels solid now
  * There's a new set of tests for Updater with `offline=True`: these try to ensure nothing is downloaded while offline,  basic offline cases work and that offline errors are what we expect
  * TrustedMetadataSet tests were essentially doubled with subtests to cover both possible `respect_expiry` values 
  * It would be an option to do the same subtest treatment to Updater tests to ensure that `offline=True`  does not change anything we don't expect it to change... but I feel like this is a reasonable amount of tests
* I've moved the Updater `offline` checks to the `Updater._load_*()` methods: You can check the commit-before-last if you want to see what the alternative option looks like

### DIFF
--- a/tuf/ngclient/config.py
+++ b/tuf/ngclient/config.py
@@ -24,7 +24,13 @@ class UpdaterConfig:
             are used, target download URLs are formed by prefixing the filename
             with a hash digest of file content by default. This can be
             overridden by setting ``prefix_targets_with_hash`` to ``False``.
-
+        offline: When offline, Updater works normally except for two differences: First,
+            expired metadata is considered valid. Second, Updater will not
+            download metadata or targets (if a download would be needed, a DownloadError
+            is raised instead). This is useful for a client that has all required
+            metadata and targets cached locally and is willing to sacrifice timeliness of
+            metadata for offline functionality. Offline mode is not strictly compatible
+            with the TUF client workflow.
     """
 
     max_root_rotations: int = 32
--- a/tuf/ngclient/config.py
+++ b/tuf/ngclient/config.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass
 
 
 @dataclass
+# pylint: disable=too-many-instance-attributes
 class UpdaterConfig:
     """Used to store ``Updater`` configuration.
 
@@ -33,3 +34,4 @@ class UpdaterConfig:
     snapshot_max_length: int = 2000000  # bytes
     targets_max_length: int = 5000000  # bytes
     prefix_targets_with_hash: bool = True
+    offline: bool = False
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -126,22 +126,26 @@ class Updater:
         repository state.
 
         Raises:
-            OSError: New metadata could not be written to disk. In offline mode,
-                cached metadata could not be found.
+            OSError: New metadata could not be written to disk.
             RepositoryError: Metadata failed to verify in some way
             DownloadError: Download of a metadata file failed in some way
         """
 
         if self.config.offline:
-            # Try loading only local data
-            data = self._load_local_metadata(Timestamp.type)
-            self._trusted_set.update_timestamp(data)
-            data = self._load_local_metadata(Snapshot.type)
-            self._trusted_set.update_snapshot(data, trusted=True)
-            data = self._load_local_metadata(Targets.type)
-            self._trusted_set.update_delegated_targets(
-                data, Targets.type, Root.type
-            )
+            # load only local data
+            try:
+                data = self._load_local_metadata(Timestamp.type)
+                self._trusted_set.update_timestamp(data)
+                data = self._load_local_metadata(Snapshot.type)
+                self._trusted_set.update_snapshot(data, trusted=True)
+                data = self._load_local_metadata(Targets.type)
+                self._trusted_set.update_delegated_targets(
+                    data, Targets.type, Root.type
+                )
+            except (OSError, exceptions.RepositoryError) as e:
+                raise exceptions.DownloadError(
+                    "Cannot download metadata in offline mode"
+                ) from e
         else:
             self._load_root()
             self._load_timestamp()
@@ -174,8 +178,7 @@ class Updater:
                 that uniquely identifies the target within the repository.
 
         Raises:
-            OSError: New metadata could not be written to disk. In offline mode,
-                cached metadata could not be found.
+            OSError: New metadata could not be written to disk.
             RepositoryError: Metadata failed to verify in some way
             DownloadError: Download of a metadata file failed in some way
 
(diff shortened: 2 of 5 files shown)


jku:
I'm still considering whether the if-else paths in Updater are in the right places:
* currently they are in `refresh()` and `_preorder_depth_first_walk()`
* they could instead be a level lower: `_load_root()`, `_load_timestamp()`, `_load_snapshot()`, `_load_targets()`

I think the latter could be slightly more code but more understandable. Now that tests exist, I think I will try that option as well to see what it looks like.

**EDIT**: I've now implemented the latter option in the last commit, I think it's decent

tuf/ngclient/config.py
```diff
@@ -33,3 +34,4 @@ class UpdaterConfig:
     snapshot_max_length: int = 2000000  # bytes
     targets_max_length: int = 5000000  # bytes
     prefix_targets_with_hash: bool = True
+    offline: bool = False
```
jku, line 43: this needs a docstring (but the whole feature also needs some documentation)

tuf/ngclient/updater.py
```diff
@@ -229,6 +231,9 @@ def download_target(
             Local path to downloaded file
         """
 
+        if self.config.offline:
+            raise exceptions.DownloadError("Cannot download when offline")
```
jku, line 235: I didn't introduce a new exception subclass since the context seems to make this useful enough for apps: if your updater is offline, then DownloadError means _the requested action is not possible with the currently cached data_.

jku:
A possible complaint on this PR is that it exposes a very specific new mode: so for example expiry is either fully respected or not at all:
* caller can't know if some of the metadata was expired (apart from warnings in the log)
* caller can't say "I want offline mode but always respect expiry"

I felt like exposing more knobs in the API would make it more complicated to use, to test and to build in a safe manner. I'm happy to consider doing that though if there are real use cases where these knobs would make a difference.

tuf/ngclient/config.py
```diff
@@ -23,7 +24,13 @@ class UpdaterConfig:
             are used, target download URLs are formed by prefixing the filename
             with a hash digest of file content by default. This can be
             overridden by setting ``prefix_targets_with_hash`` to ``False``.
-
+        offline: When offline, Updater works normally except for two differences: First,
```
lukpueh, line 27: IMO "works normally" downplays the difference quite a bit. Wouldn't it be fairer to reverse the description and underline the difference a long the lines of ... ?

> when offline, Updater does not update metadata or target files anymore. It can still be used to retrieve and vet locally available target files against locally available metadata with the important caveat that metadata expiry is ignored.

tuf/ngclient/config.py
```diff
@@ -23,7 +24,13 @@ class UpdaterConfig:
...
+            expired metadata is considered valid. Second, Updater will not
+            download metadata or targets (if a download would be needed, a DownloadError
+            is raised instead). This is useful for a client that has all required
+            metadata and targets cached locally and is willing to sacrifice timeliness of
```
lukpueh, line 31: It's hard for a client to know, if they really have all the required metadata and targets cached locally, if they don't consult with the repository and ignore expirations.

Also, while I can see how it is useful for a client to use the metadata and targets cached locally, I am still sceptical about the usefulness of using the TUF client for this. In other words, what is the benefit of using the updater in offline mode vs. not using the updater at all?

We should make this clear somewhere. Probably not here in the docstring.

lukpueh: sacrifice timeliness AND compromise recovery

lukpueh (commented):
Great patch! Implementation-wise this looks good. I still think we should give clear answers to the following questions somewhere in the ngclient documentation:

- What exactly is the use of offline mode over not using ngclient at all when offline? 
- More specifically, what security guarantees exactly do we sacrifice in offline mode ("timeliness" alone feels to vague) and what guarantees do we preserve?

jku:
I have forgotten to update this one:

I discussed this with lukas and from sigstore perspective with William woodward and came these conclusions:
* this patch has value if the local metadata can be expected to be more trustworthy than targets cache: currently this is not the case, at least #1168 would be needed for that
* Adding complexity and new features into the library (with those features essentially being not useful until other work) does not seem useful
* sigstore can achieve the same results by just trusting the local target cache when offline

That should answer Lukas questions in the comment: this PR alone does not have clear security  benefits. We could get protection against maliciously modified local cache (both metadata and targets) if we can assume:
* this PR,
* 1168 (verifying local metadata from bootstrapped root),
* more secure initial metadata (e.g. the initial metadata is installed as non user modifiable, like if sigstore-python was installed as a Debian package)

So my instinct is to not merge this PR. We can re-evaluate if 1168 gets fixed.

PR #2479, opened 2023-10-02
workflows: Stop pinning actions that are not security relevant

jku:
These workflows have no real security relevance (runtime, build or test) in the sense that a compromise in the dependencies could not compromise python-tuf security:
* scorecards
* dependency-review
* codeql-analysis

There is also no significant reproducibility issue as these are not part of the test suite. These are all "tier 3" dependencies as documented in https://github.com/theupdateframework/python-tuf/issues/2014#issuecomment-1303524280

Stop pinning the actions used in these workflows (except the common actions that are used everywhere like actions/checkout: use the same version of those everywhere). The benefit here is fewer Dependabot PRs: If we had done this from the start we'd have skipped ~70 PRs by now.

The interesting permissions used in these workflows are
 * `security-events: write` This can add things onto the "Security" tab in GitHub
 * `id-token: write` This allows OIDC authentication, but only as this specific workflow

These permissions look completely acceptable to me.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -16,18 +16,19 @@ jobs:
     name: Analyze
     runs-on: ubuntu-latest
     permissions:
+      # NOTE: If you add security critical permissions, start pinning used actions
       actions: read
       contents: read
-      security-events: write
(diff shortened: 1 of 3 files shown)


joshuagl (approved):
This feels very pragmatic. +1

PR #2481, opened 2023-10-03
Metadata API: add get_verification_result method

### LINKED ISSUE #2449: Metadata API: add `_Delegator.get_delegate_signing_status()`
`_Delegator.verify_delegate` provides a binary answer for whether a delegated role has enough valid signatures to meet the threshold of keys defined by its delegator or not. 

While this is sufficient for client validation, TUF applications might also be interested in more granular information, e.g. when implementing a distributed signing feature for TUF metadata, so they can tell users how many signatures by what keys are still missing. 

Interesting information would be:
- set of valid signatures
- set of invalid signatures (this would require validating all signatures; `verify_threshold` only verifies until threshold is met)
- optional: eligible keys, used keys, remaining keys (maybe out of scope)

Given that threshold verification is such a crucial part of the Metadata API and can be done wrong so easily, adding a new public function --  e.g. `_Delegator.get_delegate_signing_status()` -- to Metadata API seems like a good idea.

lukpueh:
The method returns detailed information about signature verification of a delegated role metadata.

Its implementation is taken from the verify_delegate method and slightly updated. verify_delegate now is a thin wrapper on top of get_verification_result.

fixes #2449

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -47,6 +47,7 @@ from tuf.api.metadata import (
     TargetFile,
     Targets,
     Timestamp,
+    VerificationResult,
 )
 from tuf.api.serialization import DeserializationError, SerializationError
 from tuf.api.serialization.json import JSONSerializer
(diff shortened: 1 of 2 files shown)


jku (approved):
Looks right to me but would maybe be useful to get review from someone else too, I wrote a bit too much of this to be able to see it with fresh eyes...

kairoaraujo:
I will have a look at it.

kairoaraujo (approved):
LGTM.

PR #2506, opened 2023-11-03
CI: Run lint on oldest supported Python version

jku:
* This was suggested as best practice by a pylint developer
* Seems better than CI randomly breaking when GitHub updates Python version (and pylint starts applying new rules that we can't follow because that would break old Python versions)

--- 

In the long run I'd like to drop pylint and move to ruff (even if it's not as extensive), but this is just a quick fix to get the CI green again.

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -14,10 +14,10 @@ jobs:
       - name: Checkout TUF
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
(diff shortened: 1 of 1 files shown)


jku:
Note that we may have to deal with the lint warning anyway once the developers are linting with 3.12 ... but I have a feeling pylint might revert this decision (enabling this warning by default) before that

lukpueh:
➕ for moving to ruff. Pylint is quite the bottleneck in our tests

lukpueh (approved):

PR #2528, opened 2023-12-13
build: Upgrade hatchling to 1.20.0

lukpueh:


lukpueh:
IMO, we can wait for a couple of days before we merge in case there comes another patch release.

jku (approved):
quick manual test says this works, thanks (and the hatchling update seems fine too). 

It's annoying that `build`  does not document the isolated build environment versions in the log but it looks like it does the right thing.

lukpueh:
> It's annoying that build does not document the isolated build environment versions in the log but it looks like it does the right thing.

Yep, and there is no sane way of making it more verbose.  I did check if it works, by removing the stdout pipes in the [`subprocess` call used to run `pip install`](https://github.com/pypa/build/blob/de5b44b0c28c598524832dff685a98d5a5148c44/src/build/env.py#L64). 🙈

Issue #2014, opened 2022-06-01
document dependabot PR review strategy


jku:
This is what I believe makes sense:
* "test" dependencies are mostly pinned to make sure we know which versions are being used and that the current version doesnt't surprisingly start breaking builds: reviewing the contents of test dependency updates is not required. If the tests pass that should be fine. Major updates could warrant a look in the changelog
* actual run time dependency updates should be reviewed: what this means is case dependent, but as a minimum we should check the changelog. Reading the commit log or actual changes may be useful but can also be an unrealistic goal for some dependency updates.
* The purpose of the dependency review is two-fold: 
  * prevent depending on software that works differently than we expect (so API changes, other functionality changes, bugs)
  * prevent depending on software that is actually malicious (this is more relevant the newer the update is as a lot of malicious updates are noticed fairly quickly). It should be noted that pypi package can be malicious without the malicious code being in github: source code review only goes so far.

This isn't documented anywhere: it should be

rdimitrov:
I thought of touching that in the maintainer guidelines for go-tuf, so I'll be following 👍 One thing I had in mind is in go-tuf we have this policy where at least two people should review and approve PRs before they are merged, but there are dependabot cases where the change is nothing concerning and one approval should be enough too.

cc: @znewman01

lukpueh:
> This is what I believe makes sense:
> ...

Makes sense to me as well, and I have nothing really to add in terms of content. 

I do have a few ideas for where we could write this down:
- in python-tuf's [*Governance* doc](https://github.com/theupdateframework/python-tuf/blob/develop/docs/GOVERNANCE.md#contributions) (probably not the right place for wordy instructions, but rather a commitment to follow certain guidelines, that may be phrased out elsewhere)
- in a new *Dependency Review Guidelines*  document, either in python-tuf's [*docs*](https://github.com/theupdateframework/python-tuf/blob/develop/docs/GOVERNANCE.md), or in [secure-systems-lab/lab-guidelines](https://github.com/secure-systems-lab/lab-guidelines), which already hosts the [*Development Guidelines*](https://github.com/secure-systems-lab/lab-guidelines/blob/master/dev-workflow.md), which are cited in python-tuf's [*Instructions for contributors*](https://theupdateframework.readthedocs.io/en/latest/CONTRIBUTING.html)
[...]

Issue #2529, opened 2023-12-13
investigate alternative hatchling version pinning 


jku:
**Description of issue or feature request**:

Investigate if we could pin the hatchling version in a way that dependabot can manage.

Currently we do this in pyproject.toml:
`requires = ["hatchling==1.20.0"]`
And then call install like this:
`python3 -m pip install --constraint requirements/build.txt build`

What if instead we did 
`requires = ["hatchling"]`
and called build like this:
`PIP_CONSTRAINTS=requirements/build.txt python3 -m pip install build`
(with hatchling version added to build.txt)

I think this should work: The environment variable should carry over to the build isolation environment unlike the --constraint flag. 

**Current behavior**:

Dependabot does not update hatchling version.

**Expected behavior**:

Dependabot does update hatchling version when new versions are released

lukpueh:
Neat! Manual updates are definitely not sustainable. I can try this out.

joshuagl:
Very neat suggestion, thanks!

PR #2530, opened 2023-12-13
Dependabot: Use groups, update weekly

jku:
All dependencies are now checked weekly and those weekly updates are grouped into 4 groups:
  * critical python build/release deps
  * python test and lint deps (only pinned for test repro)
  * all other python dependencies
  * All github action dependencies

This is not quite the division that was hashed out in #2014, mostly for practical reasons:
* GitHub actions are already practically split by pinning strategy so they don't really need further groups:
  * Non-security-relevant actions are pinned by tags
  * Other actions are pinned by hash
* The dependency grouping is quite limited

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,13 +4,36 @@ updates:
 - package-ecosystem: "pip"
   directory: "/"
   schedule:
-    interval: "daily"
-    time: "10:00"
+    interval: "weekly"
   open-pull-requests-limit: 10
+  groups:
+    build-and-release-dependencies:
+      # Python dependencies known to be critical to our build/release security
+      patterns:
+        - "build"
+    test-and-lint-dependencies:
+      # Python dependencies that are only pinned to ensure test reproducibility
+      patterns:
+        - "bandit"
+        - "black"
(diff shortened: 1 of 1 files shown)


jku:
Example here: https://github.com/jku/python-tuf/pull/184

(note that dependabot seems a little confused as I added tox into this group after the PR existed and dependabot does not seem to cope with that)

.github/dependabot.yml
```diff
@@ -4,13 +4,36 @@ updates:
...
   schedule:
-    interval: "daily"
-    time: "10:00"
+    interval: "weekly"
   open-pull-requests-limit: 10
+  groups:
+    build-and-release-dependencies:
+      # Python dependencies known to be critical to our build/release security
+      patterns:
+        - "build"
```
jku, line 13: If #2529 works out, we should add hatchling here

joshuagl (approved):
This is a really neat improvement to maintainer workflows, LGTM!

.github/dependabot.yml
jku: added!

PR #2541, opened 2024-01-11
build: constrain version in verify_release script

lukpueh:
In #2528 we added a workaround in cd.yml, which allows pinning the build backend version AND having Dependabot autodupates for it.

This workaround also needs to be applied verify_release for reproducible builds verification.

that is provided, the better.

### DIFF
--- a/verify_release
+++ b/verify_release
@@ -45,8 +45,14 @@ def build(build_dir: str) -> str:
         git_cmd = ["git", "clone", "--quiet", orig_dir, src_dir]
         subprocess.run(git_cmd, stdout=subprocess.DEVNULL, check=True)
 
+        # patch env to constrain build backend version as we do in cd.yml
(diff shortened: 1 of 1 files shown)


verify_release
```diff
@@ -45,8 +45,14 @@ def build(build_dir: str) -> str:
         git_cmd = ["git", "clone", "--quiet", orig_dir, src_dir]
         subprocess.run(git_cmd, stdout=subprocess.DEVNULL, check=True)
 
+        # patch env to constrain build backend version as we do in cd.yml
+        env = os.environ.copy()
+        env["PIP_CONSTRAINT"] = "requirements/build.txt"
```
lukpueh, line 50: Note that this makes the script slightly less portable. Not sure if that matters.

jku (approved):
good call. Looks ok to me

Issue #932, opened 2019-10-10
Reinstate full protection against slow retrieval attacks


lukpueh:
**Description of issue or feature request**:

#781 updated tuf's downloader module to use [`requests`](https://github.com/psf/requests) instead of custom networking code, to fix issues with HTTPS proxies.

This change, however, deprived TUF of a way to prevent a certain kind of slow retrieval attack, where the attacker sends bytes below a minimum average speed.

This attack can not be prevented with the [timeout parameters provided by `requests`](https://requests.readthedocs.io/en/master/user/advanced/#timeouts), which are:
- connect timeout (max # of seconds to establish connection with server)
- read timeout (max # of seconds between bytes sent from the server)

**Current behavior**:
Cannot reliably enforce minimal average download speed and thus is vulnerable to slow retrieval attacks.

**Expected behavior**:
Reinstate full protection against slow retrieval attacks

Possible ways of fixing are
- report issue to `requests` and ask them to fix it, or
- switch to equally well-established and feature-rich `requests`-alternative that supports non-blocking I/O

*Note: The comments in #781 have more ways of addressing this issue and also discuss whey they are not viable. Consider reading up on the discussion when working on this issue.*

joshuagl:
I'd like to grab this, though it may be a a couple of weeks until I get round to it.

sharmaansh21:
@lukpueh Hi, I don't think anyone working on this, can I pick up this issue ?

govardhangdg:
Hello @lukpueh! I'd like to pick this up as my GSoC project.

lukpueh:
Thanks for your interest, @optimisticanshul and @govardhangdg! Make sure to turn in your GSoC application by the end of March.

Besides, you are welcome to join us on the [CNCF TUF slack channel](http://cloud-native.slack.com/messages/tuf) and tell us a little bit about you and why you are interested in working on TUF and this task in particular. :)

govardhangdg:
Thanks @lukpueh. Will do!

Urhengulas:
Hey @lukpueh,
I would also love to work on it as part of GSOC2020. Currently finishing my proposal and going to send a post in Slack.

jcstr:
Hello Everyone, 

Following the suggestion that @lukpueh says here, I've look around for a replacement for `requests`  as I'm taking the consideration to the properties that make a good replacement for it.

### Properties that fit for a good candidate for `requests` replacement
The library that will be the replacement has:
[...]

PR #2542, opened 2024-01-15
increase default network timeout

### LINKED ISSUE #2442: ngclient: Increase default network timeout
Our default RequestsFetcher sets this default value:

    self.socket_timeout: int = 4  # seconds

sigstore-python patches this:

    # NOTE: We poke into the underlying fetcher here to set a more reasonable timeout.
    # The default timeout is 4 seconds, which can cause spurious timeout errors on
    # CI systems like GitHub Actions (where traffic may be delayed/deprioritized due
    # to network load).
    fetcher = RequestsFetcher()
    fetcher.socket_timeout = 30
    return fetcher

I agree with sigstore-python on this:
* Trying to defend against slow-retrieval attacks is in my opinion impossible in practice, especially for a generic library that could be used in a large number of contexts (it _may_ be possible for a specific application). Even with 4 sec timeout a server could return 1 byte every 3 seconds without triggering this limit: that's a download speed of 30kB per day.
* This theoretical defense is creating real world bugs as documented above

Requests by default does not set a timeout at all:  We should set a default timeout (because indefinite hangs are also not cool) but we should set it without trying to solve slow-retrieval attacks at all: 30 secs sounds good to me.

NicholasTanz:
that is provided, the better.

Fixes #2442 

increase default network timeout from 4 sec to 30 sec. 
requirements**:

### DIFF
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -29,12 +29,11 @@ class RequestsFetcher(FetcherInterface):
 
     Attributes:
         socket_timeout: Timeout in seconds, used for both initial connection
-            delay and the maximum delay between bytes received. Default is
-            4 seconds.
+            delay and the maximum delay between bytes received.
         chunk_size: Chunk size in bytes used when downloading.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, socket_timeout: int = 30, chunk_size: int = 400000) -> None:
         # http://docs.python-requests.org/en/master/user/advanced/#session-objects:
         #
         # "The Session object allows you to persist certain parameters across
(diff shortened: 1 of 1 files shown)


NicholasTanz:
Hey @jku , can you take a look at this PR when you get a chance? thanks

jku (commented):
I think this is good, however:
 * There is a class docstring a few lines above this that now documents the wrong default value though -- I think that docstring sentence could be removed altogether so that the value isn't specified in two places (alternativelly the default settings could be made optional constructor arguments at which point their default values are documented by sphinx automatically)
* I'd really like it if the git commit message explained _why_ the change is done. e.g. 
    ```
    trying to defend against slow retrieval attacks in a generic library is impossible 
    but too low timeouts mean failures in high latency systems (like tests running
    on CI).
    ```


I'll leave the approval for someone else as this was my suggestion in the first place -- @lukpueh do you have an opinion?

lukpueh:
Works for me. AFAICS, we already acknowledged the problems with and the resulting lack of slow retrieval protection (#932). So I don't think this patch has any real security implications.

 @NicholasTanz, can you please address @jku's comments? Happy to then approve.

NicholasTanz:
> Works for me. AFAICS, we already acknowledged the problems with and the resulting lack of slow retrieval protection (#932). So I don't think this patch has any real security implications.
> 
> @NicholasTanz, can you please address @jku's comments? Happy to then approve.

yes sounds good, just made the changes @lukpueh

jku (approved):
Thanks, sorry we dropped the ball for two weeks.

changes look good. I might do a squash merge to get the commit message to be on the relevant commit... Let's see if this project allows that

PR #2551, opened 2024-02-01
Improve verification results

### LINKED ISSUE #2544: VerificationResult should include keys for keyids
The `get_verification_result` method returns a `VerificationResult`, which contains two sets of TUF keyids.

Typical usage of a VerificationResult entails retrieving the `Key` objects for those keyids from the corresponding root metadata keystore. It would be useful, if `VerificationResult` included Key objects right away.

Options are:
1. Use sets of Key objects  - This requires making the `Key` class hashable and immutable, which is possible but maybe overly complicated (see [discussion)](https://github.com/theupdateframework/python-tuf/issues/2449#issuecomment-1695601623))
2. Use lists of Key objects and promise that they only contain unique keys, and optionally provide tooling for set-like operations (see [wrapper workaround](https://github.com/theupdateframework/tuf-on-ci/blob/7899655c04248de8850ed3de2c584290a842cf5a/repo/tuf_on_ci/_repository.py#L59-L102))
3. Use "keyid: Key"- dictionaries  (see [wrapper workaround](https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/2db958ee9334499f474da527e13a0a9f865a1976/repository_service_tuf/cli/admin2.py#L56-L86))

jku:
This is an attempt at fixing the verification result handling that has been found to not work well. This API is useful mostly to repository implementations: clients should likely keep using `verify_delegate() `

**Modify VerificationResult**
* Remove `union()`: this is now handled by the new RootVerificationResult
* include threshold: This is useful and can be added now that union() does not exist
* Use `dict[str, Key]` instead of `set[str]` for signed and unsigned
* Note that the (broken metadata) case of role keyid that does not have a matching key, is now not counted as "unsigned" as it was before: it's just not counted at all 

**Add RootVerificationResult**
* Externally looks like VerificationResult (without threshold)
* It is actually two VerificationResults in a trenchcoat: `Root.get_root_verification_result()` is provided as a straightforward way to get one

Note that `RootVerificationResult.signed` and `RootVerificationResult.unsigned` assume that there are no keyid collisions between the two verifying roles (same keyid is not used for two different keys): This does **not** affect `verified` status in any way (only the `signed` and `unsigned` lists) so I think this is totally acceptable.

So far the usage looks quite nice but I think we want to try to use this in an app or two to avoid redoing the mistake of the first VerificationResult... 

Fixes #2544 

### DIFF
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -986,11 +986,16 @@ class Root(Signed, _DelegatorMixin):
             signatures: Signatures over payload bytes
 
         Raises:
-            ValueError: no delegation was found for ``delegated_role``.
+            ValueError: no delegation was found for ``root`` or given Root
+                versions are not sequential.
         """
 
         if previous is None:
             previous = self
+        elif self.version != previous.version + 1:
+            versions = f"v{previous.version} and v{self.version}"
+            raise ValueError(
+                f"Expected sequential root versions, got {versions}.")
 
         return RootVerificationResult(
             self.get_verification_result(Root.type, payload, signatures),
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -653,8 +653,8 @@ class VerificationResult:
 
     Attributes:
         threshold: Number of required signatures.
-        signed: dict of keyid:Key containing the keys that have signed.
-        unsigned: dict of keyid:Key containing the keys that have not signed.
+        signed: dict of keyid to Key, containing keys that have signed.
+        unsigned: dict of keyid to Key, containing keys that have not signed.
     """
 
     threshold: int
@@ -675,8 +675,8 @@ class RootVerificationResult:
     """Signature verification result for root metadata.
 
     Root must be verified by itself and the previous root version. This
-    dataclass represents that combination. For the edge case of first version
-    of root, the same VerificationResult can be used twice.
+    dataclass represents both results. For the edge case of first version
+    of root, these underlying results are identical.
 
     Note that `signed` and `unsigned` correctness requires the underlying
     VerificationResult keys to not conflict (no reusing the same keyid for
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -965,30 +965,36 @@ class Root(Signed, _DelegatorMixin):
 
     def get_root_verification_result(
         self,
-        other: "Root",
+        previous: Optional["Root"],
         payload: bytes,
         signatures: Dict[str, Signature],
     ) -> RootVerificationResult:
         """Return signature threshold verification result for two root roles.
 
-        Verify root metadata with two roles (the root role from `self` and
-        `other`). If you have only one role (in the case of root v1) you can
-        provide the same Root as both `self` and `other`.
+        Verify root metadata with two roles (`self` and optionally `previous`).
+
+        If the repository has no root role versions yet, `previous` can be left
+        None. In all other cases, `previous` must be the previous version of
+        the Root.
 
         NOTE: Unlike `verify_delegate()` this method does not raise, if the
         root metadata is not fully verified.
 
         Args:
-            other: The other `Root` to verify payload with
+            previous: The previous `Root` to verify payload with, or None
             payload: Signed payload bytes for root
             signatures: Signatures over payload bytes
 
         Raises:
             ValueError: no delegation was found for ``delegated_role``.
         """
+
+        if previous is None:
+            previous = self
+
         return RootVerificationResult(
             self.get_verification_result(Root.type, payload, signatures),
-            other.get_verification_result(Root.type, payload, signatures),
+            previous.get_verification_result(Root.type, payload, signatures),
         )
 
 
--- a/tuf/api/metadata.py
+++ b/tuf/api/metadata.py
@@ -652,7 +652,6 @@ class VerificationResult:
     """Signature verification result for delegated role metadata.
 
     Attributes:
-        verified: True, if threshold of signatures is met.
         threshold: Number of required signatures.
         signed: dict of keyid:Key containing the keys that have signed.
         unsigned: dict of keyid:Key containing the keys that have not signed.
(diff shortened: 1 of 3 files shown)


jku:
Still wondering if it would  be useful for the two result types to have a common base type, instead of being just duck typed to feel similar. 

The way I see a signing tool or repository UI using these is this:
* getting verification results happens separately for root and other roles (because it has to)
* but printing the results should be possible with the same code (with maybe a small tweak for root case)

So maybe it is fine for the app to have something like
```
def describe_result(result: VerificationResult | RootVerificationResult) -> str:
    ...
```

I would maybe rather not create some complicated inheritance (and inevitably large classes) when the dataclasses look like they might just work like that...

tuf/api/metadata.py
```diff
@@ -653,30 +652,61 @@ class VerificationResult:
...
-        """
-        return VerificationResult(
-            self.verified and other.verified,
-            self.signed | other.signed,
-            self.unsigned | other.unsigned,
-        )
+    @property
+    def signed(self) -> Dict[str, Key]:
+        """Dictionary of all signing keys that have signed, from both VerificationResults"""
+        return self.first.signed | self.second.signed
```
jku: this unfortunately requires python 3.9

Following likely does the same with dict unpacking but is even less readable
`return {**self.first.signed, **self.second.signed}`

tuf/api/metadata.py
```diff
@@ -935,6 +963,34 @@ def get_key(self, keyid: str) -> Key:  # noqa: D102
...
+        self,
+        other: "Root",
+        payload: bytes,
+        signatures: Dict[str, Signature],
+    ) -> RootVerificationResult:
+        """Return signature threshold verification result for two root roles.
+
+        Verify root metadata with two roles (the root role from `self` and
+        `other`). If you have only one role (in the case of root v1) you can
+        provide the same Root as both `self` and `other`.
```
lukpueh: I like this. It treats v1 as the edge case which it is. So far in my code it always looked like the non-v1 case was the special case.

tuf/api/metadata.py
```diff
@@ -935,6 +963,34 @@ def get_key(self, keyid: str) -> Key:  # noqa: D102
...
+            other: The other `Root` to verify payload with
+            payload: Signed payload bytes for root
+            signatures: Signatures over payload bytes
+
+        Raises:
+            ValueError: no delegation was found for ``delegated_role``.
+        """
+        return RootVerificationResult(
+            self.get_verification_result(Root.type, payload, signatures),
+            other.get_verification_result(Root.type, payload, signatures),
```
lukpueh: Is it premature optimisation to call get_verification_result only once, if self==other?

tuf/api/metadata.py
```diff
@@ -653,30 +652,63 @@ class VerificationResult:
...
+        return len(self.signed) >= self.threshold
+
+
+@dataclass
+class RootVerificationResult:
+    """Signature verification result for root metadata.
+
+    Root must be verified by itself and the previous root version. This
+    dataclass represents that combination. For the edge case of first version
+    of root, the same VerificationResult can be used twice.
```
lukpueh: I had to read `get_root_verification_result` to understand what this means. Maybe rephrase to something like:


```suggestion
    dataclass represents both results. For the edge case of first version
    of root, the results are identical.
```

tuf/api/metadata.py
```diff
@@ -653,30 +652,63 @@ class VerificationResult:
     """Signature verification result for delegated role metadata.
 
     Attributes:
-        verified: True, if threshold of signatures is met.
-        signed: Set of delegated keyids, which validly signed.
-        unsigned: Set of delegated keyids, which did not validly sign.
+        threshold: Number of required signatures.
+        signed: dict of keyid:Key containing the keys that have signed.
+        unsigned: dict of keyid:Key containing the keys that have not signed.
```
lukpueh: nit: Would you mind to change this?

```suggestion
        signed: dict of keyids to Keys that have signed.
        unsigned: dict of keyids to Keys that have not signed.
```

The second colon throws off the parser in my head. 😆

lukpueh (commented):
This looks very helpful. Let me try it in RSTUF and see if it really is.

tuf/api/metadata.py
jku: it probably is

tuf/api/metadata.py
```diff
@@ -935,6 +963,34 @@ def get_key(self, keyid: str) -> Key:  # noqa: D102
 
         return self.keys[keyid]
 
+    def get_root_verification_result(
+        self,
+        other: "Root",
```
lukpueh: It would be convenient to allow passing `None` for `other` and default to `self`. 

Plus we could check if the versions are correct:
```python
if self.version == 1 and other.version != 1 or self.version > 1 and other.version != self.version - 1:
   raise ValueError 
```

lukpueh: Similarly, if we decide to make the interface version-aware, I suggest to rename `other` to `previous`.

jku: that sounds reasonable

lukpueh:
I tried this in RSTUF and found it very helpful.

I think adding a common base class would be okay. The inheritance model could look something like this.
```python
class VerificationResult:
  verified: bool
  signed: dict[str, Key]
  unsigned: dict[str, Key]

class Single(VerificationResult):
  threshold: int

class Combined(VerificationResult):
  first: Single
  second: Single
```

But this does not add much value. The more interesting question is, if we want to provide other common behavior. Two things that I would have found useful:

- a list of missing signature infos:
   unique tuples of `threshold -  len(signed)` and `unsigned` if `verified` is `False` per Single result

- keys that effectively decrease the missing signature count
  union of `unsigned` if `verified` is `False` per Single result

jku:
The common base class probably looks more complicated: because Combined wants to override `verified`, `signed` and `unsigned` they probably would have to be properties in the parent class already.

jku:
> * a list of missing signature infos:
> unique tuples of threshold -  len(signed) and unsigned if verified is False per Single result
> * keys that effectively decrease the missing signature count
> union of unsigned if verified is False per Single result

I'm not completely sold on these because :
* this is only relevant for root and only when root signers change
* I think most repositories in practice would want to encourage everyone to sign (to make the signing events Proof-of-Ownership events at the same time) 
* Handling multliple tuples does not seem any easier than handling `result.first` and `result.second` (or if you think it is we can just make those a tuple instead of two fields) so you could just handle the underlying VerificationResults yourself

We could add this to `VerificationResult` if it makes things easier:
```
@property
def missing(self) -> int:
    return max(0, self.threshold - len(self.signed))
```

tuf/api/metadata.py
jku: oops, I didn't do any version checking yet. I will add some of it...

jku: Right, I added the check for sequential  versions. This seems uncontroversial.

I didn't check that "initial version" is v1. I don't think that's a spec requirement and I could even  imagine some semi-practical cases where the opposite could happen

tuf/api/metadata.py
```diff
@@ -719,33 +756,30 @@ def get_verification_result(
...
-        signed = set()
-        unsigned = set()
+        signed = {}
+        unsigned = {}
 
         for keyid in role.keyids:
             try:
                 key = self.get_key(keyid)
             except ValueError:
-                unsigned.add(keyid)
```
lukpueh, line 729: Interesting. This means that in the new VerificationResult there might be role keyids that are neither in signed nor unsigned. That's probably okay. But should we mention it somewhere. Maybe in the PR description?

lukpueh (approved):

tuf/api/metadata.py
jku: Yeah it is a change -- I think I mentioned it in the first commit message , can add to PR decription.

 I was also considering just raising at this point (because clearly the role is in a bad state)... but then that sounds like punishing the delegated role signing  for delegator metadata error -- an error that can still leave the delegated role capable of signing.

PR #2558, opened 2024-02-17
Replace most linting tools with ruff 

### LINKED ISSUE #2550: Replace most linting tools with ruff
Linting still takes so long in this project... I propose replacing these tools with `ruff check` and `ruff format`:
* black
* isort
* pylint
* pydocstyle

Result would be
* 3 fewer lint dependencies
* significantly faster linting

The only downside is that ruff linter selects very few linting rulesets by default (https://docs.astral.sh/ruff/linter/#rule-selection): we may want to enable few more.
But I think even modest collection of rules would be acceptable as a start.

NicholasTanz:
…pylint

that is provided, the better.

Fixes #2550 

Linting still takes so long in this project... I propose replacing these tools with ruff check and ruff format:

black
isort
pylint
pydocstyle

requirements**:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,19 +83,16 @@ line-length=80
 [tool.ruff.lint]
 select = [
     "D",  # pydocstyle
-    "I",  # isort
+    "E",  # pycodestyle
     "F",  # pyflakes
+    "I",  # isort
     "N",  # pep8-naming
-    "E"   # pycodestyle 
 ] 
 ignore = ["D400","D415","D213","D205","D202","D107","D407","D413","D212","D104","D406","D105","D411","D401","D200","D203"]
+
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["D", "E"]
 "examples/*/*" = ["D"] 
-"tuf/repository/__init__.py" = ["F401"]
-"tuf/api/exceptions.py" = ["F401"]
-"tuf/repository/_repository.py" = ["N818"]
-"verify_release" = ["F401", "D103", "E501"]
 
 # Pylint section
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,6 +83,10 @@ line-length=80
 [tool.ruff.lint]
 ignore = ["D400","D415","D213","D205","D202","D107","D407","D413","D212","D104","D406","D105","D411","D401","D200","D203"]
 
+[tool.ruff.lint.per-file-ignores] 
+"tuf/repository/__init__.py" = ["F401"]
+"tuf/api/exceptions.py" = ["F401"]
+
 # Pylint section
 
 # Minimal pylint configuration file for Secure Systems Lab Python Style Guide:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,8 +87,9 @@ select = [
     "F",  # pyflakes
     "I",  # isort
     "N",  # pep8-naming
+    "PL", # pylint
 ] 
-ignore = ["D400","D415","D213","D205","D202","D107","D407","D413","D212","D104","D406","D105","D411","D401","D200","D203"]
+ignore = ["D400","D415","D213","D205","D202","D107","D407","D413","D212","D104","D406","D105","D411","D401","D200","D203", "PLR0913", "PLR2004"]
 
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["D", "E"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,11 +81,21 @@ dev-mode-dirs = ["."]
 line-length=80
 
 [tool.ruff.lint]
+select = [
+    "D",  # pydocstyle
+    "I",  # isort
+    "F",  # pyflakes
+    "N",  # pep8-naming
+    "E"   # pycodestyle 
+] 
 ignore = ["D400","D415","D213","D205","D202","D107","D407","D413","D212","D104","D406","D105","D411","D401","D200","D203"]
-
-[tool.ruff.lint.per-file-ignores] 
+[tool.ruff.lint.per-file-ignores]
+"tests/*" = ["D", "E"]
+"examples/*/*" = ["D"] 
 "tuf/repository/__init__.py" = ["F401"]
 "tuf/api/exceptions.py" = ["F401"]
+"tuf/repository/_repository.py" = ["N818"]
+"verify_release" = ["F401", "D103", "E501"]
 
 # Pylint section
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,56 +95,6 @@ ignore = ["D400","D415","D213","D205","D202","D107","D407","D413","D212","D104",
 "tests/*" = ["D", "E"]
 "examples/*/*" = ["D"] 
 
-# Pylint section
-
-# Minimal pylint configuration file for Secure Systems Lab Python Style Guide:
-#     https://github.com/secure-systems-lab/code-style-guidelines
-#
-# Based on Google Python Style Guide pylintrc and pylint defaults:
-#     https://google.github.io/styleguide/pylintrc
-#     http://pylint.pycqa.org/en/latest/technical_reference/features.html
-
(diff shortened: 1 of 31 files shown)


NicholasTanz:
Hey @jku, I was wondering if you could take a look at this pr when you get a chance. I wasn't able to find a feature in Ruff that would check for specific regex patterns for functions, modules, classes, etc. like Pylint does - [pylint regex](https://pylint.pycqa.org/en/1.7/user_guide/options.html). So any input is appreciated - thanks

tox.ini
```diff
@@ -47,16 +47,14 @@ deps =
     --editable {toxinidir}
 lint_dirs = tuf examples tests verify_release
 commands =
-    black --check --diff {[testenv:lint]lint_dirs}
-    isort --check --diff {[testenv:lint]lint_dirs}
+    ruff check  --diff {[testenv:lint]lint_dirs}
```
jku: There is a difference between ruff check and ruff format here: we don't want `--diff` with ruff check as the docs state

> Implies `--fix-only`

Unlike `ruff format`, `ruff check` doesn't know how to fix most of the issues it finds, so this will silence most errors.

tox.ini
```diff
@@ -47,16 +47,14 @@ deps =
...
+    ruff format --diff {[testenv:lint]lint_dirs}
     pylint -j 0 --rcfile=pyproject.toml {[testenv:lint]lint_dirs}
 
     mypy {[testenv:lint]lint_dirs}
-
+    
     bandit -r tuf
 
-    pydocstyle tuf
```
lukpueh, line 58: Is this also covered by `ruff`?

tox.ini
```diff
@@ -47,16 +47,14 @@ deps =
...
 commands =
-    black --check --diff {[testenv:lint]lint_dirs}
-    isort --check --diff {[testenv:lint]lint_dirs}
+    ruff check  --diff {[testenv:lint]lint_dirs}
+    ruff format --diff {[testenv:lint]lint_dirs}
     pylint -j 0 --rcfile=pyproject.toml {[testenv:lint]lint_dirs}
 
     mypy {[testenv:lint]lint_dirs}
-
+
```
lukpueh: Please try to avoid unrelated changes and trailing new space. :)

requirements/lint.txt
```diff
@@ -6,9 +6,7 @@
 # Lint tools
 # (We are not so interested in the specific versions of the tools: the versions
 # are pinned to prevent unexpected linting failures when tools update)
-black==24.2.0
-isort==5.13.2
 pylint==3.0.3
+ruff==0.2.1
```
lukpueh, line 9: nit: can you please adopt these changes in test-and-lint-dependencies group in `dependabot.yml`?

lukpueh (commented):
Thanks for working on this. I left a few inline comments. Let's not worry about the regex rules, I think they only replicate pep8 naming convention, and add some exemptions for `unittest` methods.

pyproject.toml
```diff
@@ -75,17 +75,17 @@ include = [
...
 line-length=80
 
-# Isort section
-# Read more here: https://pycqa.github.io/isort/docs/configuration/config_files.html
-[tool.isort]
-profile="black"
-line_length=80
-known_first_party = ["tuf"]
+[tool.ruff.lint]
+ignore = ["D400","D415","D213","D205","D202","D107","D407","D413","D212","D104","D406","D105","D411","D401","D200","D203"]
```
jku: note that these currently do not get used at all since "D" (pydocstyle) is not in the default rule set.

Maybe it's fine to leave them here unused for now: Enabling pydocstyle probably requires figuring out how to limit that to tuf/ only: tests are likely to not pass this.

pyproject.toml
```diff
@@ -75,17 +75,17 @@ include = [
...
+
+[tool.ruff.lint.per-file-ignores] 
+"tuf/repository/__init__.py" = ["F401"]
+"tuf/api/exceptions.py" = ["F401"]
```
jku: I'd like these inline in the files if possible -- I'm assuming there aren't a massive amount of these

jku (commented):
> I wasn't able to find a feature in Ruff that would check for specific regex patterns for functions, modules, classes, etc. like Pylint does

I think we are not in love with the specific regex used, we just want two things:
* linter enforces a common and fairly strict python style (pep8 I suppose?)
* there's some escape hatch for things that might currently fail (this could include things like `setUpClass` that unittest requires)

So I'd say you have two options:
1. limit this PR so that it does not even aim to replace pylint yet
2. remove pylint and its config and make sure the ruff linter has some rules enabled to enforce naming style. If errors remain that look like we don't want to fix them (like mentioned setUpClass) add some specific config to allow them: I would assume these cases are pretty common so examples should be available 

If you choose option 2, there's probably some more work to do:
*  please set `select` in your tool.ruff.lint section (The default set of rules is very limited). Something like this could work as a start:
    ```
    select = [
        "E",   # pycodestyle errors
        "F",   # pyflakes
        "I",   # isort
        "N",   # pep8-naming
    ]
    ```
* I seem to remember that with the above setup there were a significant amount of "line too long" errors. I assume most of these are in docstrings (as black does not check them) so might be easy to fix. 
* If there's a small amount of issues in code, feel free to add inline ignore comments for those

tox.ini
jku: yes, as long as the relevant ruleset gets enabled (it's not by default)

pyproject.toml
jku: per-file-ignores accepts wildcards so if you enable the "D" ruleset (and see loads of errors in tests and examples), you could then try something like 

```
"tests/*" = ["D"]
"examples/*/*" = ["D"]
```

tuf/api/exceptions.py
```diff
@@ -11,6 +11,7 @@
 #### Repository errors ####
 
 # pylint: disable=unused-import
+# ruff: disable unused-inport
 from securesystemslib.exceptions import StorageError
```
jku: the way to handle this with ruff is 
```suggestion
from securesystemslib.exceptions import StorageError  # noqa: F401
```

then yuou can remove the file level ignore in pyproject.toml

jku: actually I'm sure the name works as well and might be more readable 
```suggestion
from securesystemslib.exceptions import StorageError  # noqa: unused-import
```

jku:
I this still a draft? 

Based on a quick look it seems like nice work. We could probably file new issues for most remaining nits (and new rulesets we could enable) to keep this PR reasonably sized.

NicholasTanz:
> I this still a draft?
> 
> Based on a quick look it seems like nice work. We could probably file new issues for most remaining nits (and new rulesets we could enable) to keep this PR reasonably sized.

Oh no, I forgot to change it back - and sounds good.

jku (approved):
The only hesitation I have is on how much better pylint is on some cases (pylint is type-aware and a mature project so does catch a lot of things): the differences are visible here in the amount of removed "pylint: disable" comments.

That said I think I'm ok with dropping pylint because of ruffs fantastic execution speed and also the speed of development: new rules and rulesets keep appearing constantly.

@lukpueh I'm fine with merging this but I'll leave it to you to decide.

Notes:
* should create a meta-issue to enable more rulesets (e.g. the fairly new flake8-bandit might allow removing one more rependency. A lovely end goal would be if we could opt out of rulesets instead opting in -- then we'd get new rulesets by default on ruff upgrades (and could opt-out or fix issues then if needed)
* Maybe file another issue to manage ignores
  * could use the rulenames instead of codes if that works ("too-many-arguments" instead of "PLR0913"), especially in in-line code comments
  * pycodestyle is disabled for tests: look at enabling it
* ruff check supports `--output-format=github ` for inline annotations -- could  test this

I can file these issues if this gets merged

lukpueh:
> @lukpueh I'm fine with merging this but I'll leave it to you to decide.

I have no objections. :)

jku:
Thanks @NicholasTanz!
  
  Filed #2566, #2567, #2568

PR #2561, opened 2024-02-21
gitattributes: Fix diff hunk headers

jku:
Git is really bad at identifying the correct Python function in the diff hunk headers (example is a change in Metadata.sign()):

    @@ -384,7 +384,7 @@ class Metadata(Generic[T]):

Amazingly there is much better context detection built-in, just not enabled. The same diff hunk headers with this commit looks like:

    @@ -384,7 +384,7 @@ def sign(

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 # Test data should not be modified on checkout, regardless of host settings
 *.json binary
+*.py diff=python


joshuagl (approved):
Awesome find. Yet another cool git feature which is missed by many, thanks for adding to .gitattributes!

PR #2569, opened 2024-02-21
lint: Enable pycodestyle in tests

### LINKED ISSUE #2568: linting: enable pycodestyle for tests
We're currently ignoring pycodestyle errors ruleset ("E") in pyproject.toml

```
"tests/*" = ["D", "E"]
"examples/*/*" = ["D"] 
```

We should
* not ignore "E" for tests
* fix the issues in tests or add  `# noqa: XXX` comments to silence specific places in code
* if it looks like there is a specific rule we don't want to follow and there are too many violations, we can add that rule to `per-file-ignores`

jku:
Only leave E501 (line-too-long) disabled: There is a lot of embedded test data that is not formatted according to the rules.

Fixes #2568


joshuagl (approved):
Very nice changes. I'm enjoying learning about ruff through the PRs here :-D

PR #2570, opened 2024-02-22
Lint use GitHub output format

### LINKED ISSUE #2566: Test "ruff check --output-format=github"
tox.ini contains a call to `ruff check`. Documentation talks about `--output-format=github` option that apparently might enable inline comments in github...

We should test this:
* add `--output-format=github` to the "ruff check" call
* in another temporary commit, introduce some linting errors (maybe by removing the whole `tool.ruff.lint.per-file-ignores` section in pyproject.toml) 
* make a draft PR to check if the error messages are visible inline in GitHub
* if they do, remove the temporary commit and force push (and make PR ready for review)

jku:
Use `RUFF_OUTPUT_FORMAT=github` when on GitHub: this gives inline linter annotations

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -26,6 +26,8 @@ jobs:
           python3 -m pip install --constraint requirements/build.txt tox coveralls
 
       - name: Run tox
+        env:
+          RUFF_OUTPUT_FORMAT: github
         run: tox -e lint
 
   tests:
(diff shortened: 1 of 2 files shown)


jku:
It's a little magical:
* the log still shows normal output https://github.com/theupdateframework/python-tuf/actions/runs/8005642339/job/21865562979?pr=2570
* but clearly the github output is used since the annotations are there: https://github.com/theupdateframework/python-tuf/pull/2570/commits/24c3dce8e43bcc7c5143c7aa3953179c76f45c22

but I guess we'll take that :)

lukpueh (approved):

PR #2573, opened 2024-02-26
change utcnow() to now(timezone.utc)

### LINKED ISSUE #2572: datetime.utcnow() is deprecated: stop using it
[utcnow() reference docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow): 
> Deprecated since version 3.12: Use [datetime.now()](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) with [UTC](https://docs.python.org/3/library/datetime.html#datetime.UTC) instead.

We have 23 uses of `utcnow()` (mostly tests and examples): all seem straight forward to change to `now()` with utc argument

Mansur908:
Fixes #2572

### DIFF
--- a/examples/manual_repo/basic_repo.py
+++ b/examples/manual_repo/basic_repo.py
@@ -23,7 +23,7 @@ NOTE: Metadata files will be written to a 'tmp*'-directory in CWD.
 
 import os
 import tempfile
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any, Dict
 
@@ -47,7 +47,7 @@ from tuf.api.serialization.json import JSONSerializer
 
 def _in(days: float) -> datetime:
     """Adds 'days' to now and returns datetime object w/o microseconds."""
-    return datetime.utcnow().replace(microsecond=0) + timedelta(days=days)
+    return datetime.now(timezone.utc).replace(microsecond=0) + timedelta(days=days)
 
 
 # Create top-level metadata
--- a/examples/manual_repo/basic_repo.py
+++ b/examples/manual_repo/basic_repo.py
@@ -47,8 +47,9 @@ from tuf.api.serialization.json import JSONSerializer
 
 def _in(days: float) -> datetime:
     """Adds 'days' to now and returns datetime object w/o microseconds."""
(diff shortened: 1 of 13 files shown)


lukpueh (changes requested):
Thanks for the patch! Please fix the [linter warnings](https://github.com/theupdateframework/python-tuf/actions/runs/8055330328/job/22004264585?pr=2573#step:5:16) (a few lines are too long now), and sign off your commits (see [DCO check](https://github.com/theupdateframework/python-tuf/pull/2573/checks?check_run_id=22001920416)), otherwise this looks great.

lukpueh:
FYI: You can run `ruff format .` to reformat the long lines

jku:
@Mansur908 if you have any suggestions on how our developer docs could be better to improve this "first PR flow", please comment in #2578

lukpueh (changes requested):
Lint is still failing.  Did you reformat with `ruff`? Btw. you can also [run tests locally](https://theupdateframework.readthedocs.io/en/latest/CONTRIBUTING.html) with `tox`, or  `tox -e lint` to run only the linter.

jku:
This might help you avoid the long lines in tests and examples: Instead of `from datetime import timezone` you could do `from datetime import UTC`  -- I think this alias looks fine for our case and makes usage short.
 
 (but the suggestion to run `tox` after any changes is still good)

jku:
Interesting, this leads to more code issues (serialization and deserialization), I believe:
* `datetime.strptime()` returns naive datetimes (at least for the input we give it): these will then not be compatible with non-naive `datetime.now()` results. We probably should set the timezone to UTZ immediately after reading the naive datetime with strptime
* serializing utc datetimes fails as our code `expires.isoformat() + "Z"` produces gibberish now: we should use an explicit strftime call here

I can have a look at this later but if you want to continue investigating:
* strptime() is used in tuf/api/_payload.py and  tests/generated_data/generate_md.py
* isoformat() is used in tuf/api/_payload.py

jku:
You seem to be allowing changes to your branch (thanks) so I'll push some code fixes on top: they should get this into a ready state

lukpueh (approved):

jku:
It's a bit of a scary change (with the serialization code tweaks) but I feel pretty good about our test suite... so let's go with this. Thanks @Mansur908

PR #2578, opened 2024-02-27
Improve dev docs

jku:
I'm trying to improve the experience for first time contributors: I think it must be bad since our success rate is probably below 10% for the first version of the first PR...

* Mention tests and DCO in PR template
* tweak contributing doc

Other suggestions welcome.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,8 @@
 <!--
 Before submitting a pull request:
-  * Please run linter and tests locally: "tox"
+  * Run linter and tests locally: Use "tox"
   * Ensure your commits are signed-off-by: Use "commit --signoff"
+  * Make sure new code has tests and is documented
   * For more info, see docs/CONTRIBUTING.rst
 
 Once commits are signed off and tested, describe the purpose and contents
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTING.rst
```diff
@@ -63,26 +71,9 @@ via PyPI).
 Auto-formatting
 ---------------
 
-CI/CD will check that new TUF code is formatted with `black
-<https://black.readthedocs.io/>`__ and `isort <https://pycqa.github.io/isort>`__.
-Auto-formatting can be done on the command line:
+The linter in CI/CD will check that new TUF code is formatted with
+`ruff<https://docs.astral.sh/ruff/>`__. Auto-formatting can be done on the
```
lukpueh: I think the whitespace is needed to render correctly. Also, the double-underscore suffix seems to be syntax for "anonymous hyperlink targets", which I don't think makes sense here.

```suggestion
`ruff <https://docs.astral.sh/ruff/>`_. Auto-formatting can be done on the
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,13 +1,22 @@
...
 
-**Please verify and check that the pull request fulfills the following
-requirements**:
 
+
+
+Fixes #<ISSUE NUMBER>
+
+
+<!-- Please check relevant options (you can remove the ones that do not apply) -->
```
lukpueh: I wonder if we should just remove the items below. They are usually ignored, and even if not, I never found them particularly helpful, neither as reviewer nor as contributor.

Maybe we can include a reminder in the bullet list above, something like:

> ensure changes are tested and code docs are updated

lukpueh (changes requested):
Thanks for the initiative, Jussi!

lukpueh (approved):

Issue #2567, opened 2024-02-21
linting: Enable more ruff rulesets 


jku:
ruff linter has 50 or so [rulesets](https://docs.astral.sh/ruff/rules/) that can be enabled. Currently we enable 6 good ones, but could enable more.

### Instructions

* Pick a ruleset that is not yet checked in the list below
* Add the ruleset in `select` array in pyproject.toml
* Run `tox -e lint`
* If there are any linter complaints:
  * If the complaints seems reasonable, try fixing the issue in code 
  * If there are only a small number of errors and they don't seem like they should be fixed, it may be appropriate to add an ignore comment to those lines (example: `# noqa: F401`)
  * If there are lots of failures, it may be an option to add a specific rule to `ignore` or `per-file-ignores` in pyproject.toml. Feel free to pick another ruleset as well: some are much trickier than others
  * If you are not sure how to proceed please file a new issue, mention which ruleset you are enabling and paste the `tox -e lint` output there to get advice
* Once there are no more linter errors, create a PR with your changes!

### Rulesets

Here's a list of rulesets that looked potentially interesting to me (checked ones are already enabled):
* [x] Pyflakes (F)
* [x] pycodestyle Error (E)
* [x] isort (I)
* [x] pep8-naming (N)
* [x] pydocstyle (D)
* [x] Pylint (PL)
* [x] pyupgrade (UP)
* [x] Ruff-specific rules (RUF)
* [x] pycodestyle Warning (W)
* [x] flake8-bandit (S)
* [x] flake8-annotations (ANN)
* [x] flake8-blind-except (BLE)
* [x] flake8-bugbear (B)
* [x] flake8-builtins (A)
* [x] flake8-datetimez (DTZ)
* [x] flake8-comprehensions (C4)
* [x] flake8-implicit-str-concat (ISC)
* [x] flake8-return (RET)
* [x] flake8-self (SLF)
* [x] flake8-simplify (SIM)
* [x] flake8-unused-arguments (ARG)

jku:
PR for flake8-bandit: #2571

rschulm:
Hello @jku , could you assign this issue to me? I am new to OSS and Git, and am looking to do my first few pull requests.

jku:
Thanks, PRs are welcome! 

I won't assign this issue since it's probably 10+ PRs altogether
* You can pick any ruleset from the list and try enabling that -- feel free to pick another if the first one seems too tricky (there's a lot of variance in complexity)
[...]

PR #2592, opened 2024-03-27
lint: Enable more ruff rulesets

jku:
Minor fixes were needed, the only possibly interesting one is the one in RequestsFetcher (use "yield from").

This is part of #2567

I expect this to currently fail tests for unrelated reasons,  #2591 should fix it

### DIFF
--- a/tuf/ngclient/_internal/requests_fetcher.py
+++ b/tuf/ngclient/_internal/requests_fetcher.py
@@ -106,8 +106,7 @@ class RequestsFetcher(FetcherInterface):
         """
 
         try:
-            for data in response.iter_content(self.chunk_size):
-                yield data
+            yield from response.iter_content(self.chunk_size)
         except (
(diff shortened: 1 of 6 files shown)


tuf/ngclient/_internal/requests_fetcher.py
```diff
@@ -106,8 +106,7 @@ def _chunks(self, response: "requests.Response") -> Iterator[bytes]:
         """
 
         try:
-            for data in response.iter_content(self.chunk_size):
-                yield data
+            yield from response.iter_content(self.chunk_size)
```
lukpueh, line 109: TIL

lukpueh (approved):

lukpueh:
> I expect this to currently fail tests for unrelated reasons, https://github.com/theupdateframework/python-tuf/pull/2591 should fix it

Just merged the fix. This PR should pass on rebase.

PR #2597, opened 2024-04-03
Enable linters

jku:
Enable linters:
* flake8-unused-arguments
* flake8-executable
* pygrep-hooks
* flake8-pie
* flake8-pyi
* flake8-raise
* flake8-self

These were all low hanging fruit where ruff can autofix or where there were only a few issues to fix.

This is part of  #2567


jku:
closing/reopening to trigger tests again

jku:
rebased on main again

lukpueh (approved):
These feel mostly nit picky. But I also won't argue with an auto-formatter. 😅

jku:
> These feel mostly nit picky. But I also won't argue with an auto-formatter. 😅

yeah some of the rules are very arguable... but for each ruleset there's typically a bunch of really good ones that just fit our code 100% so I figured the few weird ones we can live with or disable individually

PR #2599, opened 2024-04-03
Adopt upstream DSSE model change

lukpueh:
*[blocks on next securesystemslib release]*

Adopt `securesystemslib.dsse.Envelope.signatures` type change from list to dict (secure-systems-lab/securesystemslib/pull/743) in `tuf.api.dsse.SimpleEnvelope` subclass.

fixes #2564

PR #2600, opened 2024-04-03
Constrain securesystemslib dependency to <0.32.0

lukpueh:
There are several breaking changes coming up in securesystemslib on its way to 1.0.

To not disrupt tuf users this patch constrains securesystemslib to not update the current minor version..


lukpueh:
`with-sslib-main` tests fail for unrelated reasons and will be fixed in #2599

jku (approved):
This makes sense

PR #2601, opened 2024-04-04
Release v4.0.0

jku:
This is a major bump only because of Metadata API, ngclient remains compatible with 3.x.

---

The main motivation for doing the release right now is 
> Constrain securesystemslib dependency to <0.32.0 in preparation for future securesystemslib API changes

This allows securesystemslib to make API breaking releases without accidentally breaking python-tuf users who do not pin all of their dependencies.

Note that we had forgotten to "forward-port" the changelog and version bump from 3.1.1 branch so I did it here.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,37 @@
 # Changelog
 
+## v4.0.0
+
+This release is a small API change for Metadata API users (see below).
+ngclient API is compatible but optional DSSE support has been added.
+
+### Added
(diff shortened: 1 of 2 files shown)


jku:
For release announcement:
Thanks to @NicholasTanz for switching linting to ruff
Thanks to @Mansur908 for the datetime fixes

lukpueh (approved):
LGTM. I cross-checked with all PRs since 3.1.0, and this seems to capture the important things. You could mention "docs" under "various improvements" too, but I'm fine if you don't.

PR #2602, opened 2024-04-06
workflows: Add awk magic to GH changelog generation

jku:
* Create a changelog file with awk
* Add both "dist" and "changelog" to GH artifact
* This changes how the artifact is built: Now the dist directory is inside the artifact (instead of the contents of dist being in the artifact): this means the default path now works for `download-artifact`
* Dump changelog into the release body

This is obviously a bit more complicated... but I believe the typical failure mode is a successful release run, the changelog could just be empty (if e.g. CHANGELOG.md is not well formed).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -30,17 +30,18 @@ jobs:
       - name: Install build dependency
         run: python3 -m pip install --constraint requirements/build.txt build
 
-      - name: Build binary wheel and source tarball
-        run: PIP_CONSTRAINT=requirements/build.txt python3 -m build --sdist --wheel --outdir dist/ .
+      - name: Build binary wheel, source tarball and changelog
+        run: |
+          PIP_CONSTRAINT=requirements/build.txt python3 -m build --sdist --wheel --outdir dist/ .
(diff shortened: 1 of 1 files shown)


jku:
Adding a bit more detail: this is something I wanted for tuf-on-ci so did here as well. If others think this is unnecessary complexity that's fine by me.

.github/workflows/cd.yml
```diff
@@ -30,17 +30,20 @@ jobs:
       - name: Install build dependency
         run: python3 -m pip install --constraint requirements/build.txt build
 
-      - name: Build binary wheel and source tarball
-        run: PIP_CONSTRAINT=requirements/build.txt python3 -m build --sdist --wheel --outdir dist/ .
+      - name: Build binary wheel, source tarball and changelog
+        env:
+          TAG: ${{ github.ref_name }}
```
jku: I suppose this could be removed since `GITHUB_REF_NAME` should be in the environment already

jku:
tests will fail due to #2618

jku:
rebased on main

lukpueh (approved):
Very cool! 🪄

PR #2610, opened 2024-04-16
workflows: Simplify testing

### LINKED ISSUE #2604: should build job require prior test job to pass?

In #2317 we configured CI/CD tests to [`continue-on-error`](https://github.com/theupdateframework/python-tuf/blob/2d6fc743a487bb172c98a82c9a5a5c733eb134d4/.github/workflows/_test.yml#L36). I cannot remember (or guess from commits/PR/issue messages) why.

Moreover and IIUC this means that the build job, which [`needs`](https://github.com/theupdateframework/python-tuf/blob/2d6fc743a487bb172c98a82c9a5a5c733eb134d4/.github/workflows/cd.yml#L18) the test job to finish, will also trigger, if the tests fail. Is this intended?

See related in-toto/in-toto#732, which fixes the converse issue, where the build job unnecessarily required all test jobs to pass.

**Current behavior**:
build job triggers even if (all) tests fail

**Expected behavior**:
build job should maybe require tests to pass

**Bonus**
The [`experimental`](https://github.com/theupdateframework/python-tuf/blob/2d6fc743a487bb172c98a82c9a5a5c733eb134d4/.github/workflows/_test.yml#L48) setting in the test workflow is currently unused, and could be removed (or used like in the in-toto PR above).

jku:
* Don't try to handle sslib main test within the matrix
* Put it in a separate workflow
* Include the new workflow in CI but not in CD
* Bonus: Make cache-dependency-path more complete

Fixes #2604

---

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -9,7 +9,6 @@ jobs:
     name: Lint Test
     runs-on: ubuntu-latest
 
-
     steps:
       - name: Checkout TUF
         uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
@@ -19,7 +18,9 @@ jobs:
         with:
           python-version: 3.8
           cache: 'pip'
-          cache-dependency-path: 'requirements/*.txt'
+          cache-dependency-path: |
+            requirements/*.txt
+            pyproject.toml
 
       - name: Install dependencies
         run: |
@@ -33,23 +34,10 @@ jobs:
   tests:
     name: Tests
     needs: lint-test
-    continue-on-error: true
     strategy:
-      # Run regular TUF tests on each OS/Python combination, plus
-      # (sslib main) on Linux/Python3.x only.
       matrix:
-        toxenv: [py]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         os: [ubuntu-latest, macos-latest, windows-latest]
-        include:
-          - python-version: 3.x
-            os: ubuntu-latest
-            toxenv: with-sslib-main
-            experimental: true
-
-    env:
-      # Set TOXENV env var to tell tox which testenv (see tox.ini) to use
-      TOXENV: ${{ matrix.toxenv }}
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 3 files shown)


jku:
@lukpueh what do you think? 
* should fix the dependency issue
* workflow now does not need a Github PHD to modify
* PRs still get a _"some checks were not successful"_ which is a little annoying for external contributors but I think it's ok: we do want know if things break with sslib main...

jku:
Hmm, the check names now changed somehow -- it's different from what the expected check names are?

Maybe: 
* "CI" workflow only had one job previously: GitHub decides to not include "CI" in the name
* but now there are two jobs in "CI" so Github includes the name?

Github tries to be helpful and manages to be annoying.

lukpueh:
> Hmm, the names now changed somehow -- it's different from what the expected test names are?

Yes, we just have to uncheck old expected checks and check new ones. 

> 
> Maybe:
> 
> * "CI" workflow only had one job previously: GitHub decides to not include "CI" in the name
> * but now there are two jobs in "CI" so Github includes the name?

Could be. It also doesn't include the toxenv `py` in the name anymore.

jku:
> It also doesn't include the toxenv py in the name anymore.

Oh I missed that!

lukpueh (approved):
Looks great!

Do I understand correctly that we don't need to set `continue-on-error`, because default is `false` and so a failure in `tests`, which is included in `cd.yml` and needed by `build`, blocks `build`, but a failure in `sslib-main`, which is is not included or needed, does not?

lukpueh:
> > Hmm, the names now changed somehow -- it's different from what the expected test names are?
> 
> Yes, we just have to uncheck old expected checks and check new ones.


IIUC, right now, I can only un-require the old checks, but the settings UI won't show me the new ones until the PR is merged. I suggest we merge as is, bypassing the branch protection, and then I migrate the settings.

jku:
> Do I understand correctly that we don't need to set continue-on-error, because default is false and so a failure in tests, which is included in cd.yml and needed by build, blocks build, but a failure in sslib-main, which is is not included or needed, does not?

Yes that's my understanding.

For CI workflows:
* CI workflow will fail if _test_sslib_main.yml fails
* but that's ok since the "required statuses" feature uses specific status checks and not the workflow state

jku:
merged, I'll modify the required checks now...

PR #2612, opened 2024-04-19
Support app-specific user-agents

### LINKED ISSUE #2611: provide user-agent customization?
At some point we decided to not offer the ability to customize `RequestsFetcher`. The idea was that if someone wants to to customize, they can provide their own fetcher (possibly forking RequestsFetcher). In general this still seems reasonable.

There is one small customization I wish we had made easier: user-agent. It would be really useful for repository implementations to know which versions of which apps are connecting to the repository -- and it's a little much to expect apps to implement complete fetchers just to do that.

So maybe we should make extending the user-agent easier:
* maybe add user-agent field to UpdaterConfig and forward the value to RequestsFetcher
* alternatively make RequestsFetcher public and add an optional constructor argument user-agent 

I think I like the first option more since it only increases the UpdaterConfig API (the downside is that  only affects the default fetcher).

In both cases I think the argument should be _added_ to the default user agent so you end up with `MyApp/0.1 tuf/4.0.0 requests/1.2.3` in the repository logs

jku:
* application user-agent can be set with UpdaterConfig object
* Setting will affect the default fetcher only
* the application user-agent will be prefixed to the ngclient default user-agent

Fixes #2611

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -93,10 +93,14 @@ class Updater:
         else:
             self._target_base_url = _ensure_trailing_slash(target_base_url)
 
-        # Read trusted local root metadata
-        data = self._load_local_metadata(Root.type)
-        self._fetcher = fetcher or requests_fetcher.RequestsFetcher()
         self.config = config or UpdaterConfig()
+
+        if fetcher is not None:
+            self._fetcher = fetcher
+        else:
+            self._fetcher = requests_fetcher.RequestsFetcher(
+                app_user_agent=self.config.app_user_agent
+            )
 
         supported_envelopes = [EnvelopeType.METADATA, EnvelopeType.SIMPLE]
         if self.config.envelope_type not in supported_envelopes:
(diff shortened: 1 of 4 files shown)


jku:
I did not write a test as I couldn't figure out a reasonable way to do it... but I did test this manually: both default and modified user-agent are as expected.

tuf/ngclient/updater.py
```diff
@@ -93,18 +93,25 @@ def __init__(
...
+
         supported_envelopes = [EnvelopeType.METADATA, EnvelopeType.SIMPLE]
         if self.config.envelope_type not in supported_envelopes:
             raise ValueError(
                 f"config: envelope_type must be one of {supported_envelopes}, "
                 f"got '{self.config.envelope_type}'"
             )
 
+        # Read trusted local root metadata
+        data = self._load_local_metadata(Root.type)
```
lukpueh, line 113: Why was this line moved here?

lukpueh (approved):
Looks good. The purist in me leans towards option 2 in #2611, but your approach seems reasonable enough, and smaller API is definitely a good argument.

Re testing:
What speaks against adding a small case to `test_updater_ng` (or `test_fetcher_ng`), which creates updaters (or fetchers) with and without the new setting, does a refresh (or  fetch) and checks if the the session headers contains the expected string?

tuf/ngclient/updater.py
jku: * it had to be moved below self.config and self._fetcher setting anyway and
* seemed more sensible to handle all of configuration first, then load local metadata

jku:
Added the suggested test, that seems reasonable

PR #2613, opened 2024-04-21
Enable rulesets (BLE and RUF)

### LINKED ISSUE #2567: linting: Enable more ruff rulesets

NicholasTanz:
…UF012

Fixes #2567

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,7 @@ line-length=80
 select = [
     "A",   # flake8-builtins
     "B",   # flake8-bugbear
+    "BLE", # flake8-blind-except
     "C4",  # flake8-comprehensions
     "D",   # pydocstyle
     "DTZ", # flake8-datetimez
@@ -94,6 +95,7 @@ select = [
     "N",   # pep8-naming
     "PL",  # pylint
     "RET", # flake8-return
+    "RUF", # ruff-specific rules
     "S",   # flake8-bandit
     "SIM", # flake8-simplify
     "UP",  # pyupgrade
@@ -107,9 +109,10 @@ ignore = [
 
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = [
-    "D",    # pydocstyle: no docstrings required for tests
-    "E501", # line-too-long: embedded test data in "fmt: off" blocks is ok
-    "S",    # bandit: Not running bandit on tests
+    "D",      # pydocstyle: no docstrings required for tests
+    "E501",   # line-too-long: embedded test data in "fmt: off" blocks is ok
+    "S",      # bandit: Not running bandit on tests
+    "RUF012", # ruff: mutable-class-default
 ]
 "examples/*/*" = [
     "D", # pydocstyle: no docstrings required for examples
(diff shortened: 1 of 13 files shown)


jku:
FYI the failure in `with-sslib-main` is expected at the moment and will not prevent merging

NicholasTanz:
> FYI the failure in `with-sslib-main` is expected at the moment and will not prevent merging

ah gotcha, can you take a look at this pr then, when you get a chance @jku.

pyproject.toml
```diff
@@ -107,9 +109,10 @@ ignore = [
 
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = [
-    "D",    # pydocstyle: no docstrings required for tests
-    "E501", # line-too-long: embedded test data in "fmt: off" blocks is ok
-    "S",    # bandit: Not running bandit on tests
+    "D",      # pydocstyle: no docstrings required for tests
+    "E501",   # line-too-long: embedded test data in "fmt: off" blocks is ok
+    "S",      # bandit: Not running bandit on tests
+    "RUF012", # ruff: mutable-class-default
```
jku, line 115: I guess we have a bunch of these? Could mention that in the comment but no need to modify just for that

jku (approved):
Looks good -- or rather the amount of `BLE001` looks absolutely awful but I think I can't blame you for that...

Thanks!

jku:
There's "15 expected checks" listed but that's because the check names changed  in #2610 ...

All of the relevant tests have passed so I'll merge bypassing branch protections.

PR #2627, opened 2024-05-07
linting: Enable all Ruff rulesets by default

jku:
The goal here is to have ruff enable new rulesets when new releases are made without us having to do anything: we can then decide if we disable or not.

* Enable a couple more rulesets (ERA, INP, T20)
* Add a few individual ignores to tests and examples
* Default to enable all, disable the rulesets we don't want


lukpueh (approved):
nice!

PR #2632, opened 2024-05-20
changing user-agent value

### LINKED ISSUE #2631: user-agent should maybe mention "python"?
User agent set in tuf/ngclient/_internal/requests_fetcher.py is `tuf/5.0.0` -- this looks good but I think should be `python-tuf/5.0.0` instead as it should identify the complete project when visible in an access log.

h4l0gen:
It's just changing user agent string to `python-tuf/5.0.0` from `tuf/5.0.0`

Fixes #2631


jku (approved):
Looks great, thank you.

PR #2642, opened 2024-05-25
Updating Contributing guidelines and copy-pastable code

### LINKED ISSUE #2637: Make Implementations command/code copy-pastable

Currently, the [implementation ](https://theupdateframework.readthedocs.io/en/latest/INSTALLATION.html) command/code is not copy-pastable. Giving an `copy` option in top right corner of code's box would address this issue. With this, we can pass the [CNCF assessment criteria](https://github.com/cncf/techdocs/blob/main/assessments/criteria.md#new-user-content).

**Current behavior**:
Not copy-pastable

**Expected behavior**:
Should be copy-pastable.

h4l0gen:
This will make easier to copy commands

Fixes #2637

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -6,4 +6,3 @@
 # install sphinx and its extensions
 sphinx
 sphinx-rtd-theme
-sphinx-copybutton
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -6,3 +6,4 @@
 # install sphinx and its extensions
 sphinx
 sphinx-rtd-theme
+sphinx-copybutton
--- a/docs/CONTRIBUTING.rst
+++ b/docs/CONTRIBUTING.rst
@@ -31,7 +31,7 @@ tox run.
 
 ::
 
-    $ tox
+     tox
 
 Below, you will see more details about each step managed by ``tox``, in case
 you need debug/run outside ``tox``.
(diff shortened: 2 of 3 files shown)


h4l0gen:
Hi @jku , we only need to add `sphinx_copybutton` extension. Also I've changed `CONTRIBUTING.rst` and remove `$` so that we can paste them directly to terminal. PTAL. Thank you!

requirements/docs.txt
```diff
@@ -6,3 +6,4 @@
 # install sphinx and its extensions
 sphinx
 sphinx-rtd-theme
+sphinx-copybutton
```
lukpueh: This does add a new doc build dependency, which @jku expressed his reservations about: https://github.com/theupdateframework/python-tuf/issues/2637#issuecomment-2129522982

I don't have strong objections, but I also don't think the copy button is such a big win.

jku: Yeah no strong feelings either.  It's true it's not a runtime dependency and it's just one small thing that gets installed when you run `tox`. The list of things we install is long already so this doesn't change the situation much... but I don't think we should add developer dependencies without good reason -- and the copy button doesn't feel like a very important one.

The other changes seem good. I would say remove the copybutton references and we'll merge it

h4l0gen: That's okay @lukpueh @jku I understands your POV. BTW I've got this idea just to fulfill [last point criteria](https://github.com/cncf/techdocs/blob/main/assessments/criteria.md#new-user-content) in this heading of CNCF assessment, so that we don't miss this criteria in future. But if you like, I will remove that, and merge other changes, WDYT?
Thank you 👍

h4l0gen: @jku @lukpueh I've made changes, can we merge it now?
Thank you👍

jku (approved):

jku:
thanks, squashed and merged.

PR #2647, opened 2024-05-28
made variable public

### LINKED ISSUE #2598: SimpleEnvelope._DEFAULT_PAYLOAD_TYPE should be public?
ngclient accesses `SimpleEnvelope._DEFAULT_PAYLOAD_TYPE`.

flake8-self linter complains about this, I think correctly. If we use the value from a different module, it should be public

CC @lukpueh for comment

h4l0gen:
It makes _DEAFULT_PAYLOAD_TYPE variable public, so that flake8 linter don't show any error.

Fixes #2598

### DIFF
--- a/tuf/api/dsse.py
+++ b/tuf/api/dsse.py
@@ -52,7 +52,7 @@ class SimpleEnvelope(Generic[T], BaseSimpleEnvelope):
 
     """
 
-    _DEFAULT_PAYLOAD_TYPE = "application/vnd.tuf+json"
+    DEFAULT_PAYLOAD_TYPE = "application/vnd.tuf+json"
 
     @classmethod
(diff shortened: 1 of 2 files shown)


h4l0gen:
I have two choices for this:

- Simply remove underscore from variable to make it public.
- Use @property 

I used first option here, as i can see in codebase there is no special constraint on using DEFAULT_PAYLOAD_TYPE variable, so there is no meaning of using @property. it will just make things complex. 

Please tell me your thoughts?

`side note` flake8 don't show any errors now. see screenshot below
.

image

jku (approved):
lgtm

h4l0gen:
Hi @jku, If it looks good to you, can we approve and merge this? 

Thank you👍

PR #2648, opened 2024-05-29
change_target_file_path

### LINKED ISSUE #2304: Delegated path pattern does not match the target path in basic_repo.py

In the basic_repo.py example when the metadata is created the target ("basic_repo.py) is used as a Targetfile with a target path set of “manual_repo/basic_repo.py”.

But then when the delegated role is created its path pattern is “*.py” which given the spec should not match to be a delegated role for that target path. 

When the generated metadata files are tested with a client, one cannot fetch the target since there's no match.

References:
* https://theupdateframework.github.io/specification/latest/#pathpattern
* https://github.com/theupdateframework/python-tuf/blob/e7cc34a6a5ed6d5d7e3a81cec328888686f59feb/examples/manual_repo/basic_repo.py#L106
* https://github.com/theupdateframework/python-tuf/blob/e7cc34a6a5ed6d5d7e3a81cec328888686f59feb/examples/manual_repo/basic_repo.py#L283

**Current behavior**:

A client cannot fetch the desired target file since there's no match.

**Expected behavior**:

A client should be able to fetch the desired target file.

h4l0gen:
changes target path in basic_repo.py

Fixes #2304

### DIFF
--- a/examples/manual_repo/basic_repo.py
+++ b/examples/manual_repo/basic_repo.py
@@ -104,8 +104,8 @@ roles["targets"] = Metadata(Targets(expires=_in(7)))
 # 'target path', which a client uses to locate the target file relative to a
 # configured mirror base URL.
 #
-#      |----base URL---||-------target path-------|
-# e.g. tuf-examples.org/repo_example/basic_repo.py
+#      |----base artifact URL---||-------target path-------|
+# e.g. tuf-examples.org/artifacts/manual_repo/basic_repo.py
 
 local_path = Path(__file__).resolve()
 target_path = f"{local_path.parts[-2]}/{local_path.parts[-1]}"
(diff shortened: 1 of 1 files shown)


h4l0gen:
IMO, this should works for us, I've used this pattern, as this uses accurate path 
> a [PATHPATTERN](https://theupdateframework.github.io/specification/latest/#pathpattern) of "targets/*.tgz" would match file paths "targets/foo.tgz" and "targets/bar.tgz", but not "targets/foo.txt".

jku (approved):
This does seem to match the target path that gets used, thanks.

This mergeable as is but if you want you can also try making this comment a little more sensible:
```
#      |----base URL---||-------target path-------|
# e.g. tuf-examples.org/repo_example/basic_repo.py
```
could be something like this to match the actual targetpath in the code

```
#      |--- base artifact URL ---||----- target path -----|
# e.g. tuf-examples.org/artifacts/manual_repo/basic_repo.py
```

h4l0gen:
SGTM, made changes. PTAL

jku:
thanks

PR #2650, opened 2024-05-30
addd_root_verification

### LINKED ISSUE #2438: repository: Tweak snapshot/timestamp triggers
Currently `Repository.do_snapshot()` and `Repository.do_timestamp()` decide whether the update is needed by looking at whether the contents are up-to-date.

This bypasses one case where timestamp and snapshot are needed: when the signing keys have changed. So I guess the two methods should also check if the current snapshot/timestamp metadata is verified by root. 

I did not do that originally since I was hoping the methods could be self contained and would not make assumptions about how the repository is generated/stored. This seems to be a good reason to peek at other metadata though: root should be assumed to exist and to be valid if you are calling do_snapshot/do_timestamp

h4l0gen:
Add root verification step in do_snapshot and do_timestamp for catching changes in signing keys. 

Fixes #2438

### DIFF
--- a/tuf/repository/_repository.py
+++ b/tuf/repository/_repository.py
@@ -188,6 +188,12 @@ class Repository(ABC):
         update_version = force
         removed: Dict[str, MetaFile] = {}
 
+        root = self.root()
+        snapshot = self.snapshot()
+
+        if not root.verify_signature(snapshot):
+            update_version = True
+
         with self.edit_snapshot() as snapshot:
             for keyname, new_meta in self.targets_infos.items():
                 if keyname not in snapshot.meta:
@@ -228,6 +234,13 @@ class Repository(ABC):
         """
         update_version = force
         removed = None
+
+        root = self.root()
+        timestampt = self.timestamp()
+
+        if not root.verify_signature(timestamp):
+            update_version = True
+
         with self.edit_timestamp() as timestamp:
             if self.snapshot_info.version < timestamp.snapshot_meta.version:
                 raise ValueError("snapshot version rollback")
--- a/tuf/repository/_repository.py
+++ b/tuf/repository/_repository.py
@@ -9,6 +9,7 @@ from contextlib import contextmanager, suppress
 from copy import deepcopy
 from typing import Dict, Generator, Optional, Tuple
 
+from tuf.api.exceptions import UnsignedMetadataError
 from tuf.api.metadata import (
     Metadata,
     MetaFile,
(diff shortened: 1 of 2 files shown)


h4l0gen:
@jku I don't have strong idea on how to check this locally, but it should works for us. PTAL and tell me if you see any required changes. 
Thanks you 👍

tuf/repository/_repository.py
```diff
@@ -188,6 +188,12 @@ def do_snapshot(
         update_version = force
         removed: Dict[str, MetaFile] = {}
 
+        root = self.root()
+        snapshot = self.snapshot()
+
+        if not root.verify_signature(snapshot):
```
jku: I think this should be something like:
```
snapshot_md = self.open(Snapshot.type)
root = self.root()
if not root.verify_delegate(Snapshot.type, snapshot_md.signed_bytes(), snapshot_md.signatures):
```
`self.snapshot()` returns the Snapshot object but we need the containing Metadata object (which includes the Snapshot object and the signatures).

It's really unfortunate if our linters are unable to see the issues here... we will definitely need to solve the testing situation for this code

jku (commented):
Thanks. The core idea seems valid but you will need the `Metadata` object that contains `Snapshot`: otherwise you can't verify that it is correctly signed.

We'll need to figure out the testing before merging this... either we need a unit test (unsure yet how easy that is) or we need to modify the repository example so that it actually hits this code path. I can try to have a look tomorrow.

tuf/repository/_repository.py
h4l0gen: Oh yes, I verified snapshot and metadata directly, Got it, and sorry I just gone through repository code so made this mistake, let me change it. Thanks for review.

h4l0gen: We have to use try and except block, as there is no return value so Lint is not happy.

Also, using UnsignedMetadataError and changing `bytes()` to `bytes` as lint taking it as function call.

jku:
I've made a first pass for a test set for tuf.repository in #2651.
* Feel free to merge that commit into this PR to test if you want (you'll need to remove the ExpectedFailure mark though) -- this way we could merge both together
* otherwise I'll probably make 2651 into a proper PR next week and you can rebase this after that one is merged

(EDIT: looks like I have some coverage issues to sort out in that PR... I will probably look into those next week)

jku (commented):
I think this is correct, thanks. I'll refrain from approving until we have a test or two

h4l0gen:
Thanks @jku for test set, Yes we can wait for 2651. 👍

jku:
2651 is now merged. When you have the time please rebase on origin/main: the two tests marked `ExpectedFailure` should start "unexpectedly" succeeding:
* if they do, just remove the marks from the two tests
* if not, let me now

thanks!

h4l0gen:
@jku expected failure tests (3.11) is failing.

jku:
> @jku expected failure tests (3.11) is failing.

these look like the "unexpected" successes that we actually expected: you can remove the  `@unittest.expectedFailure` mark from them and we should be good to merge

jku (approved):
This is great, thanks

jku:
squash merged with slightly modified commit message.

PR #2651, opened 2024-05-31
tests: Add some initial tests for tuf.repository

jku:
Hard to tell how easy these are to understand for someone who did not write the module... but here's my attempt at testing the `tuf.repository.Repository` interface:
* TestingRepository is an implementation of the interface (mostly a simplified copy of the repository example)
* that implementation is used in some very simple tests, mostly of do_snapshot and do_timestamp
* Two tests are marked `expectedFailure` because of #2438

### DIFF
--- a/tests/.coveragerc
+++ b/tests/.coveragerc
@@ -10,3 +10,4 @@ exclude_lines =
     pragma: no cover
     def __str__
     if __name__ == .__main__.:
+    @abstractmethod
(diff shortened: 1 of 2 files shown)


jku:
ugh, coverage complains because tuf.repository was never imported in tests before so was not covered by coverage before...

jku:
argh, now the coverage results are different on different pythons...

lukpueh (approved):
Looks great! I think the tests are reasonable and okay easy to understand. The repository is harder and maybe not so intuitive to implement.

PR #2670, opened 2024-07-10
Add a conformance test workflow

jku:
* The conformance test suite is likely to still change quite a bit so the workflow is not enabled on PRs yet (and does not pin the action version)
* The actual conformance client is copied from the tuf-conformance project
* This is mostly a test to see how things should work out, and a demonstration of how the tuf-conformance project should be used

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -1,0 +1,16 @@
+on:
+  # manual dispatch only while the conformance test suite is under rapid development
+  workflow_dispatch:
+
+name: CI
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the client wrapper
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Run test suite
+        uses: theupdateframework/tuf-conformance@main
+        with:
+          entrypoint: ".github/scripts/conformance-client.py"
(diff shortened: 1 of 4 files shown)


jku:
This is draft because it depends on theupdateframework/tuf-conformance#50, otherwise it should be ready

jku:
There is no automated way to run he conformance test suite locally (using your local python-tuf version) but we could potentially add something into tox... it will need a bit of thinking though.

It's not too difficult manually, something like this should work:
```
pip install -e $CONFORMANCE_SUITE_DIR -e $PYTHON_TUF_DIR
pytest "$CONFORMANCE_SUITE_DIR/tuf_conformance" --entrypoint $PYTHON_TUF_DIR/.github/scripts/conformance-client.py
```

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,16 @@
+on:
+  # manual dispatch only while the conformance test suite is under rapid development
+  workflow_dispatch:
+
+name: CI
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the client wrapper
```
lukpueh: nit: We don't call it wrapper elsewhere here. IMO calling it conformance client would be clearer.

lukpueh (approved):
Very cool. Tried it locally and it worked nicely. Does the action also work with `tuf-confromance@main`? AFAICS you tried it against a patched action?
https://github.com/jku/python-tuf/actions/runs/9874869528

jku:
Yeah I used the patched branch for testing but that branch has since been merged -- main should work right now (obviously it will break at some point so we should start pinning the action as soon as we have something that looks like a release)

PR #2679, opened 2024-07-24
Update MAINTAINERS.txt

trishankatdatadog:
Removing myself because, just like with go-tuf, I unfortunately do not have the bandwidth for active maintenance, and do not wish to be in the way. I thank you all very much for the opportunity, and your continued service.


JustinCappos:
Thanks again for all the help!

JustinCappos (approved):

PR #2721, opened 2024-10-16
Python upgrade: Stop testing 3.8, start testing 3.13

jku:
We don't strictly require 3.9 yet but likely should soon as the container annotation features are nice.

I'll file an issue on those annotations


jku:
Once there's a review here, I will change the expected test names (currently the expectation is `Tests (3.12, ...)` but I changed the name to `Tests (3.x, ...)` since we don't really care about the specific version in these two test runs)

kairoaraujo (approved):

PR #2741, opened 2024-11-29
Use future annotations

jku:
This PR achieves two things:
* Makes us compatible with Python 3.8 again  (a non-supported release by now but I figured it's nice to not be the first one to break compatibility...)
* Starts using several nice annotation features from Python 3.10, e.g.:
   * `X | Y` instead of `typing.Union[X, Y]`, 
   * `X | None` instead of `typing.Optional[X]`, 
   * No quotes around class names in factory method return values

Almost all changes come from ruff autofixes: there should be no functional changes here. There's a small revert included to avoid the one thing that was incompatible with python 3.8.

This was tested by setting "--python-version 3.8" in our mypy invocation... Ideally we should be linting using the lowest supported python version but I did not include that change in this PR

### DIFF
--- a/examples/manual_repo/basic_repo.py
+++ b/examples/manual_repo/basic_repo.py
@@ -21,6 +21,8 @@ NOTE: Metadata files will be written to a 'tmp*'-directory in CWD.
 
 """
 
+from __future__ import annotations
+
 import os
 import tempfile
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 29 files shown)


jku:
Ok, there is now a tiny runtime change as well: Repository module changes an exception type to allow for better coverage management. This is not an API change since the exception is not meant to be caught: it really is an assertion purely for defensive programming purposes.

kairoaraujo:
Thank you @jku 
I'm blocking some time to review it.

kairoaraujo (approved):
LGTM

PR #2750, opened 2024-12-11
Update securesystemslib extras

jku:
pynacl is no lo longer either a dependency or "extras" name of securesystemslib 1.0 (which we already require)


kairoaraujo:
Adding on my TODO

jku:
rebased on develop

requirements/pinned.txt
```diff
@@ -1,10 +1,24 @@
...
+idna==3.10
+    # via requests
+pycparser==2.22
+    # via cffi
 requests==2.32.3
-securesystemslib[crypto,pynacl]==1.2.0
-urllib3==2.2.3           # via requests
+    # via -r requirements/main.txt
+securesystemslib==1.2.0
+    # via -r requirements/main.txt
```
jku, line 22: the extras are no longer documented in this file since pip-compile is changing the default output to be compatible with "--constraints" (meaning no extras): this seems sensible.

I'm explicitly using `pip-compile --strip-extras` as the command so we won't get bitten by one version of pip-compile having a different default than another version at some point in future

jku:
rebased on develop

kairoaraujo (approved):

PR #2762, opened 2025-01-06
replace RequestsFetcher for Urllib3Fetcher

### LINKED ISSUE #2751: Investigate switching to urllib3
We currently use requests in the default ngclient fetcher (RequestsFetcher).

Internally requests uses urllib3 and we don't seem to use any functionality outside of what urllib3 provides: connection pool management and streaming are supported. Switching to urllib3 would mean:
* we would have (at least) one less dependency
* the API we would use is arguably better maintained

This could be achieved by first writing an alternative FetcherInterface implementation that uses urllib3: if it looks good we can then make it the default and remove the requests-based implementation.

NicholasTanz:

Fixes #2751 #2776


NicholasTanz:
Hi @jku, I implemented a FetcherInterface that utilizes urllib3, most of the implementation is quite similar to RequestsFetcher. For the tests, I swapped out the RequestsFetcher for the Urllib3Fetcher in `test_fetcher_ng.py`. (I haven't removed the `requests_fetcher.py` yet)

jku:
On first pass this looks really good, thanks -- I was worried there might be some differences that makes this complicated but your work seems very promising. Still, we'll have to be careful not to change any details we don't intend to as this is very core client code that easily affects applications (I mention this in case the review takes a while: it's only because I know this is a potentially tricky change).


I _think_ we can probably even drop the multiple poolmanagers: this manual connection pool management was added at a time when this wasn't part of the http library but now I believe the poolmanager handles exactly the issues the _get_poolmanager() code is working around. I'm going to review this better in the next days so I'll look into this.

tuf/ngclient/updater.py
```diff
@@ -102,7 +102,7 @@ def __init__(
         if fetcher is not None:
             self._fetcher = fetcher
         else:
-            self._fetcher = requests_fetcher.RequestsFetcher(
+            self._fetcher = urllib3_fetcher.Urllib3Fetcher(
```
jku, line 105: _if_ we change the default fetcher already in this PR then we'd want to change the requirements as well (pyprojects.toml, requirements/main.txt and regenerate requirements/pinned.txt) but we can still discuss if there's benefit to shipping this as alternative (non-default) fetcher first. Maybe there isn't -- if we don't think anyone is going to use and test it as an alternative then our only real choice is to swap the defaults

However, I think this change in dependencies would remove certifi from the indirect dependencies -- we'll have to take look at that and decide what is correct for python-tuf:
* what should the default certificate source be (certifi is the status quo but we may need a few lines of code to get the connectionpool to use it)
* do we allow users to change this default? how?

jku: > we may need a few lines of code to get the connectionpool to use it

this is correct: by default urllib3 uses the OS certificate store

NicholasTanz:
> I _think_ we can probably even drop the multiple poolmanagers: this manual connection pool management was added at a time when this wasn't part of the http library but now I believe the poolmanager handles exactly the issues the _get_poolmanager() code is working around. I'm going to review this better in the next days so I'll look into this.

sounds good, I was looking at this [page](https://urllib3.readthedocs.io/en/stable/advanced-usage.html)  and I think it is supported, so I can remove _get_poolmanager() / _poolManagers and just use one of them. I was wondering if I should pass in a different value for the num_pools parameter, or if the default amount of 10 works.

tuf/ngclient/updater.py
NicholasTanz: I think it makes sense to swap out the defaults in this PR, if it isn't too much overhead for changing the requirements. For the certificate source, I think it makes sense to utilize `certifi` instead of the OS certificate source, based on this [page](https://urllib3.readthedocs.io/en/stable/user-guide.html#certificate-verification) it seems like a quick swap.

tuf/ngclient/_internal/urllib3_fetcher.py
```diff
@@ -0,0 +1,152 @@
...
+            raise exceptions.SlowRetrievalError from e
+
+        # Check response status.
+        try:
+            if response.status >= 400:
+                raise urllib3.exceptions.HTTPError
+        except urllib3.exceptions.HTTPError as e:
+            response.close()
+            status = response.status
+            raise exceptions.DownloadHTTPError(str(e), status) from e
```
jku: this does not look right: no point in manually raising a urllib3 exception ourselves just to immediately catch it.

The logic seems reasonable though: if status is >= 400 then raise .

tuf/ngclient/_internal/urllib3_fetcher.py
```diff
@@ -0,0 +1,152 @@
...
+        try:
+            yield from response.stream(self.chunk_size)
+        except (
+            urllib3.exceptions.ConnectionError,
+            urllib3.exceptions.TimeoutError,
+        ) as e:
+            raise exceptions.SlowRetrievalError from e
+
+        finally:
+            response.close()
```
jku: This should be `response.release_conn()` based on the examples and docs.

tuf/ngclient/_internal/urllib3_fetcher.py
```diff
@@ -0,0 +1,152 @@
...
+        Raises:
+            exceptions.SlowRetrievalError: Timeout occurs while receiving
+                data.
+            exceptions.DownloadHTTPError: HTTP error code is received.
+
+        Returns:
+            Bytes iterator
+        """
+        # Get a customized session for each new schema+hostname combination.
+        poolmanager = self._get_poolmanager(url)
```
jku: It seems that  a single poolmanager will do fine: we could set one up in constructor already

tuf/ngclient/_internal/urllib3_fetcher.py
```diff
@@ -0,0 +1,152 @@
...
+        # Always set the timeout. This timeout value is interpreted by
+        # urllib3 as:
+        #  - connect timeout (max delay before first byte is received)
+        #  - read (gap) timeout (max delay between bytes received)
+        try:
+            response = poolmanager.request(
+                "GET",
+                url,
+                preload_content=False,
+                timeout=urllib3.Timeout(connect=self.socket_timeout),
```
jku: this only sets connection timeout but leaves read timeout at default (which might be a fine but isn't what is documented in e.g. comment above). Based on urllib3 docs this sohuld handle both:
```suggestion
                timeout=urllib3.Timeout(self.socket_timeout),
```

tests/test_fetcher_ng.py
```diff
@@ -105,11 +105,12 @@ def test_http_error(self) -> None:
...
 
     # Response read timeout error
-    @patch.object(requests.Session, "get")
+    @patch.object(urllib3.PoolManager, "request")
     def test_response_read_timeout(self, mock_session_get: Mock) -> None:
         mock_response = Mock()
+        mock_response.status = 200
         attr = {
-            "iter_content.side_effect": requests.exceptions.ConnectionError(
+            "stream.side_effect": urllib3.exceptions.ConnectionError(
```
jku: have you verified this is what urllib3 does on read timeout or is this assumption? I think this will be a MaxRetryError (wrapping a NewConnectionError) in our case

I'm not super invested in the whole SlowRetrievalError handling (as I don't think it can be effetively used in practice) but I also don't want us to test something that will not actually happen...

tests/test_fetcher_ng.py
```diff
@@ -118,13 +119,13 @@ def test_response_read_timeout(self, mock_session_get: Mock) -> None:
...
+        mock_response.stream.assert_called_once()
 
     # Read/connect session timeout error
     @patch.object(
-        requests.Session,
-        "get",
-        side_effect=requests.exceptions.Timeout("Simulated timeout"),
+        urllib3.PoolManager,
+        "request",
+        side_effect=urllib3.exceptions.TimeoutError,
```
jku: Same question here: is this the exception that really  happens when connection times out? It seems like the built-in retries could affect this and we might actually get a MaxRetryError wrapping a ReadTimeoutError or something...

jku (changes requested):
Leaving some feedback now:
* let's leave the certificate store decision open for now: William has a good point that certifi is problematic in the linked sigstore-python issue...
* I believe we can drop the manual session management and just use a single PoolManager
* The timeouts need a bit of work: we are supposed to handle timeouts specifically so we can raise SlowRetrievalError... but it looks to me like that will actually mean little more work with urllib3 (since by default we get automatic retries which I believe means any timeouts are actually wapped in MaxRetryError)
   * I think we could add some error parsing in _fetch and _chunks to only raise SlowRetrievalError on these cases
   * but on the other hand I don't want complex code for a feature I don't really think is useful (defence against slow retrieval attacks is not really possible at library level IMO) -- so let's not overdo this

jku:
For the record I was looking at the timeout results with these simple tests:

```
import urllib3
# Test connection timeout: MaxRetryError caused by NewConnectionError
urllib3.request("GET", "https://github.com", timeout=urllib3.Timeout(connect=0.0001))
# Test read timeout: MaxRetryError caused by ReadTimeoutError
urllib3.request("GET", "https://github.com", timeout=urllib3.Timeout(read=0.0001))
```

(note that this simple test reuses the default connection pool so a successful connection will prevent retesting the connection timeout without reloading the module)

tests/test_fetcher_ng.py
NicholasTanz: No, my bad, I had assumed that's what urllib3 did on read timeout, same with the connection timeout

tuf/ngclient/_internal/urllib3_fetcher.py
```diff
@@ -0,0 +1,110 @@
...
+        #  - connect timeout (max delay before first byte is received)
+        #  - read (gap) timeout (max delay between bytes received)
+        try:
+            response = self._poolManager.request(
+                "GET",
+                url,
+                preload_content=False,
+                timeout=urllib3.Timeout(self.socket_timeout),
+            )
+        except urllib3.exceptions.MaxRetryError as e:
```
NicholasTanz, line 82: I did some basic testing, and I believe this will catch the reading timeout & connecting timeout 
```python
import urllib3

try:
    r = urllib3.PoolManager().request('GET', 'https://github.com', timeout=urllib3.Timeout(connect=.0001))

except urllib3.exceptions.MaxRetryError as e:

    if isinstance(e.reason, urllib3.exceptions.TimeoutError):
        print('Timeout occurred')
```

tests/test_fetcher_ng.py
NicholasTanz: I'm honestly not too sure for the `response.stream()` if an error occurs, if it's the MaxRetryError or a TimeoutError; I think the retry logic does get triggered though and a MaxRetryError occurs (I adjusted the `_chunks` function for this)

tuf/ngclient/__init__.py
```diff
@@ -8,14 +8,14 @@
...
+from tuf.ngclient._internal.urllib3_fetcher import Urllib3Fetcher
 from tuf.ngclient.config import UpdaterConfig
 from tuf.ngclient.fetcher import FetcherInterface
 from tuf.ngclient.updater import Updater
 
 __all__ = [  # noqa: PLE0604
     FetcherInterface.__name__,
-    RequestsFetcher.__name__,
+    Urllib3Fetcher.__name__,
     TargetFile.__name__,
```
jku: let's leave both in the API for now: this way we if an application has issues with the new fetcher, they have an easy fix (just set `fetcher=RequestsFetcher`) without reverting to an older version.

That said we should document that RequestsFetcher is now deprecated and will be removed in a future release.

tuf/ngclient/updater.py
jku: I'm leaning towards using the system certs (meaning no further changes needed here):
* [Truststore](https://truststore.readthedocs.io/en/latest/#user-guide) docs make the case for system certs well
* We need to mention this in the changelog, this is a major change
* We could link to truststore as well in our docs: 
  * my current understanding is that system certs should just work on the OSes we care about, but using Truststore would be even better since it uses OS specific APIs that may update certs or revocation lists
   * We as library _probably_ should not depend on Truststore: the application can make that call if they want to

jku (commented):
Added code comments to mention that
* we probably want to provide RequestsFetcher for at least one release
* Using system certs is IMO correct and this feels like a good time to do that change

So:
 * we could add urllib3 to dependencies in `pyproject.toml` for clarity already in this PR -- even if it changes nothing in practice since requests already depends on it
 * we can't remove certifi/requests depencies yet (since RequestsFetcher is still in the API)

jku:
documenting items for changelog:
* Default FetcherInterface implementation is now Urllib3Fetcher instead of RequestFetcher
  * In practice this uses `urllib3` instead of `requests` for HTTP
  * this change also switches the TLS certificate source from `certifi` to system certificate store
  * RequestsFetcher is deprecated but remains available for at least this release

This feels like a major version bump regardless of the actual API changes

jku:
> we probably want to provide RequestsFetcher for at least one release

> we can't remove certifi/requests depencies yet (since RequestsFetcher is still in the API)

Hmm, maybe we could not depend on requests and just document that users who want to use RequestFetcher can do that... If that works in practice that might be a reasonable compromise

Anyway, I'm ok with this PR not touching the requirements: There are other uses of requests still left in the code base (examples etc) that we would want to modify before that anyway

tuf/ngclient/updater.py
NicholasTanz: yeah, after reading WIlliam's points in the linked issue, it makes sense to utilize the system certs; I agree that the library shouldn't depend on Truststore and that if an application wants to utilize it, they can choose to.

NicholasTanz:
> Anyway, I'm ok with this PR not touching the requirements: There are other uses of requests still left in the code base (examples > etc) that we would want to modify before that anyway

gotcha, I omitted `requester_fetcher.py` in `pyproject.toml,` as it was causing code coverage to be < 97%. Also, there was a [deprecation warning](https://github.com/urllib3/urllib3/issues/2920) when running the [test suite](https://github.com/theupdateframework/python-tuf/actions/runs/13065245590/job/36456391643?pr=2762), but the tests in the code base were intentionally using no scheme/host to trigger error handling logic.

jku:
Cheers, I think this looks good for next release
* I've filed the followup issues (#2773, #2776, #2777)
* Would like another reviewer to have a look as well
* I'll try to understand the response release mechanisms a little better -- I'm not sure I still fully understand the `Response.close()` and `Response.release_conn()` details

tuf/ngclient/_internal/urllib3_fetcher.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(
+        self,
+        socket_timeout: int = 30,
+        chunk_size: int = 400000,
+        app_user_agent: str | None = None,
+    ) -> None:
+        # Default settings
+        self.socket_timeout: int = socket_timeout  # seconds
+        self.chunk_size: int = chunk_size  # bytes
+        self.app_user_agent = app_user_agent
```
jku: nit: I suppose there is no need for `self.app_user_agent` as it's never used outside this method

jku (approved):
I think this looks good with the caveat that we have to handle the followup issues already mentioned.
* left one nit but it's not that important
* I would like to see another review
* I think I will try to check with other sigstore-python devs (as a major downstream) and maybe other downstreams like tufup on the timing of this

jku:
@dennisvang, @woodruffw any objections if we switch from `requests+certifi` to `urllib3 (without certifi)` in next major release?

(this would be the default: the old requests fetcher remains usable for maybe one more release)

dennisvang:
> @dennisvang, @woodruffw any objections if we switch from `requests+certifi` to `urllib3 (without certifi)` in next major release?

@jku Thanks for asking. I have no objections. 

Having fewer dependencies sounds good to me, and we need to do some refactoring on tufup anyway.

tests/test_fetcher_ng.py
```diff
@@ -94,7 +94,7 @@ def test_fetch_in_chunks(self) -> None:
     # Incorrect URL parsing
     def test_url_parsing(self) -> None:
         with self.assertRaises(exceptions.DownloadError):
-            self.fetcher.fetch("missing-scheme-and-hostname-in-url")
+            self.fetcher.fetch("http://invalid/")
```
jku, line 97: this changes the underlying error (`No address associated with hostname` vs `Name or service not known`) but that seems fine.

jku:
I left a comment about an issue with proxy support (specifically proxy env var support) https://github.com/theupdateframework/python-tuf/issues/2751#issuecomment-2656054440

I believe we need to deal with that... but I wonder if we should just file that as an issue, merge this, and expect the open issues to get solved before next release.

Pros:
* this PR wouldn't keep waiting for weeks more
* I think the proxy problem is tangible, I have a prototype already
* we can always revert if it turns out the problem is worse than believed

Cons: 
* If we need to make a release quickly in near future, we'd have to do it by branching out from 5.1.0

I think I can live with this plan. Comments welcome

NicholasTanz:
> I left a comment about an issue with proxy support (specifically proxy env var support) [#2751 (comment)](https://github.com/theupdateframework/python-tuf/issues/2751#issuecomment-2656054440)
> 
> I believe we need to deal with that... but I wonder if we should just file that as an issue, merge this, and expect the open issues to get solved before next release.


yeah, implementing and maintaining proxy env var support definitely isn't ideal... but since there's already a prototype, I think it's worth trying out. And as u mentioned, if issues arise, we can always revert back. I agree with the plan 👍

jku:
yolo etc etc. I'm merging this: 
* I believe this is a net positive for the project long term even considering the possibility of regressions
* remaining issues are labeled with [urllib3-migration](https://github.com/theupdateframework/python-tuf/issues?q=state%3Aopen%20label%3A%22urllib3-migration%22).
* if there are more issues, we will consider reverting this

PR #2773, opened 2025-01-29
More porting from  requests to urllib3

jku:
This is  a followup to #2762 (that replaces RequestsFetcher with Urllib3Fetcher):
* remove last uses of requests (examples, verify_release)
* remove requests from dependencies
* Move the fetcher implementations from `tuf.ngclient._internal` to `tuf.ngclient`: this was always the intention anyway and this allows RequestsFetcher to be not imported by default but importable if users want it: `from tuf.ngclient.requests_fetcher import RequestsFetcher`

This is pretty nice:
> $ pip install -e .
> ...
> Successfully installed securesystemslib-1.2.0 tuf-5.1.0 urllib3-2.3.0

two dependencies!

Obviously if you install cryptography you end up with a few more but even in that case we have come down from 17 dependencies in v0.14 to five dependencies in next release. With  this trajectory we're hitting zero before 2028!


jku:
Some additional changes that could be included here:
* replace requests with urllib3 in our dependencies -- this drops 3 more indirect dependencies
* Add documentation to RequestsFetcher
  * class is deprecated and will be removed in future
  * users of RequestsFetcher now need to depend on requests themselves

jku:
@NicholasTanz GitHub doesn't let me mark you as reviewer but I wouldn't mind if you did review: this is a followup on your PR

jku:
> we have come down from 17 dependencies in v0.14 to five dependencies in next release. With this trajectory we're hitting zero before 2028!

singularity is actually projected for christmas 2026:

Screenshot from 2025-02-14 22-42-19

tuf/ngclient/requests_fetcher.py
```diff
@@ -3,11 +3,14 @@
...
 
-# requests_fetcher is public but comes from _internal for now (because
-# sigstore-python 1.0 still uses the module from there). requests_fetcher
-# can be moved out of _internal once sigstore-python 1.0 is not relevant.
+Note that this module is deprecated, and the default fetcher is
+Urllib3Fetcher:
+* RequestsFetcher is still available to make it easy to fall back to
+    previous implementation if issues are found with the Urllib3Fetcher
+* If RequestsFetcher is used, note that `requests` must be explicitly
+    dependended on: python-tuf does not do that.
```
NicholasTanz:
```suggestion
    depended on: python-tuf does not do that.
```
nit - small typo

NicholasTanz:
I looked over the codebase, and it seems that there are no more uses of `requests`, and the changes look good to me 👍

NicholasTanz (approved):
I looked over the codebase, and it seems that there are no more uses of `requests`, and the changes look good to me 👍

tuf/ngclient/updater.py
```diff
@@ -49,7 +49,8 @@
 
 from tuf.api import exceptions
 from tuf.api.metadata import Root, Snapshot, TargetFile, Targets, Timestamp
-from tuf.ngclient._internal import trusted_metadata_set, urllib3_fetcher
+from tuf.ngclient import urllib3_fetcher
```
jku, line 52: hmm, I guess it would be more appropriate to import straight from the sub module to minimize cycles in imports

jku: oops my mistake, that is the whole submodule being imported here, not just a class

mnm678 (approved):

PR #2789, opened 2025-02-14
Handle proxy variables

### LINKED ISSUE #2787: urllib3: Handle proxy environment variables
This is filed in anticipation of #2762 getting merged.

urllib3 made a conscious decision to not handle proxy variables -- this is understandable for them but I don't think we can just hand wave the issue away: loads of users are behind dumb corporate proxies and their only way to get software working is proxy variables.

We can implement a small component above urllib3 PoolManager/ProxyManager that handles this. I have a prototype already

jku:
This PR handles proxy variables (notably `http_proxy`, `https_proxy`, `no_proxy`, `all_proxy`) when using Urllib3Fetcher:
* `ProxyEnvironment` is used like urllib3 `PoolManager`: it has the same `request()` method (we could easily implement the whole RequestMethods mixin but since we only use the one method, that didn't seem useful)
* Internally `ProxyEnvironment` keeps track of typically 1-3 PoolManagers [non-proxied, http-proxied, https-proxied] and chooses between those based on the requested url and the environment variables
* There's a bunch of tests: they do not test the actual proxied connection, only whether the correct pool managers were created based on the proxy environment variable values

I did not want to implement this as I think this component should be in urllib3 but it's not a lot of code (maybe 50 lines of actual code) so I can live with it. As far as I know this puts python-tuf back to feature parity with v5.1: this was the only feature that I know of that we lost with the move from `requests` to `urllib3` only.   

Fixes #2787

### DIFF
--- a/tuf/ngclient/_internal/proxy.py
+++ b/tuf/ngclient/_internal/proxy.py
@@ -3,6 +3,8 @@
 
 """Proxy environment variable handling with Urllib3"""
 
+from __future__ import annotations
+
 from typing import Any
 from urllib.request import getproxies
 
--- a/tuf/ngclient/_internal/proxy.py
+++ b/tuf/ngclient/_internal/proxy.py
@@ -38,8 +38,9 @@ class ProxyEnvironment:
         if no_proxy is None:
             self._no_proxy_hosts = []
         else:
+            # split by comma, remove leading periods
             self._no_proxy_hosts = [
-                h for h in no_proxy.replace(" ", "").split(",") if h
+                h.lstrip(".") for h in no_proxy.replace(" ", "").split(",") if h
             ]
 
     def _get_proxy(self, scheme: str | None, host: str | None) -> str | None:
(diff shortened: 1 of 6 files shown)


tuf/ngclient/_internal/proxy.py
```diff
@@ -0,0 +1,101 @@
+# Copyright New York University and the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""Proxy environment variable handling with Urllib3"""
+
+from __future__ import annotations
+
+from typing import Any
+from urllib.request import getproxies
```
jku, line 9: we could implement this method ourselves as well but I tried to keep this minimal

jku:
rebased on develop after #2773 merged

kairoaraujo:
@jku, this is an impressive work! Thank you for dedicating time on it.
I will start reviewing this weekend.

jku:
I think the `no_proxy` support in requests actually includes specifying ports as well: this implementation does not and I'm thinking of not adding it.

kairoaraujo (approved):

PR #2795, opened 2025-02-19
Enable FA (future annotations) linting ruleset

### LINKED ISSUE #2794: enable FA (future annotations) linting ruleset
We already use future annotations in the code base, just not quite everywhere.
* Remove `"FA"` from ruleset ignore list in pyproject.toml
* `tox -e lint` will start making recommendations: Use `from __future__ import annotations` and use modern annotations like it suggests

pakagronglb:
We already use future annotations (`from __future__ import annotations`) in most of the codebase. This change enables the FA linting ruleset to ensure consistent usage across all files by:

1. Removing "FA" from the ruleset ignore list in pyproject.toml
2. Running linter checks with `tox -e lint` to verify compliance
3. Running full test suite with `tox` to ensure no regressions

The linter passes without any FA-related issues, confirming that our codebase already follows the future annotations best practices consistently.

Testing:
- ✓ Ran `tox -e lint` - all checks passed
- ✓ Ran full test suite with `tox` - all tests passed

### DIFF
--- a/tuf/api/serialization/json.py
+++ b/tuf/api/serialization/json.py
@@ -1,18 +1,13 @@
 # Copyright New York University and the TUF contributors
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-"""``tuf.api.serialization.json`` module provides concrete implementations to
-serialize and deserialize TUF role metadata to and from the JSON wireline
-format for transportation, and to serialize the 'signed' part of TUF role
-metadata to the OLPC Canonical JSON format for signature generation and
-verification.
-"""
+"""JSON de/serialization code."""
 
-# We should not have shadowed stdlib json but that milk spilled already
 # ruff: noqa: A005
 
+from __future__ import annotations
+
 import json
-from typing import Optional
 
 from securesystemslib.formats import encode_canonical
 
@@ -56,7 +51,7 @@ class JSONSerializer(MetadataSerializer):
 
     """
 
-    def __init__(self, compact: bool = False, validate: Optional[bool] = False):
+    def __init__(self, compact: bool = False, validate: bool | None = False):
         self.compact = compact
         self.validate = validate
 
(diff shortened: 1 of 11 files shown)


tuf/api/serialization/json.py
```diff
@@ -1,18 +1,13 @@
...
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-"""``tuf.api.serialization.json`` module provides concrete implementations to
-serialize and deserialize TUF role metadata to and from the JSON wireline
-format for transportation, and to serialize the 'signed' part of TUF role
-metadata to the OLPC Canonical JSON format for signature generation and
-verification.
-"""
-
-# We should not have shadowed stdlib json but that milk spilled already
```
jku: was there a specific reason for removing the rest of the docstring and the comment here?

verify_release
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
jku, line 1: I agree this is correct. Would have preferred to see a  a separate commit (preferably one that fixes all shebangs in the code base, not just this one) but it's ok.

jku (changes requested):
Looks good.
* This project requires a signed-off-by line in the commit message (sorry, CNCF requires that): feel free to do a `git commit --amend --signoff` and a force push. [More info](https://github.com/theupdateframework/python-tuf/blob/develop/docs/CONTRIBUTING.rst#dco). 
* left a question about a docstring change

tuf/api/serialization/json.py
pakagronglb: Oh i think that was my mistake trying to the docstring. Let me restore it real quick. My apologies.

pakagronglb: I've restored the original detailed docstring that explains:
- What the module does (serializes/deserializes TUF role metadata)
- The formats it handles (JSON wireline format and OLPC Canonical JSON)
- The purposes (transportation and signature generation/verification)

verify_release
pakagronglb: Thank you so much for the feedback. I agree with you. Would be better for organisation too.

jku (approved):
LGTM, thanks for the contribution!

PR #2798, opened 2025-02-20
add zizmor for linting workflows.

### LINKED ISSUE #2793: Lint workflows with zizmor
It would be good if our GitHub workflows were statically analyzed with zizmor: https://github.com/woodruffw/zizmor

I think we can integrate zizmor into `tox -e lint` so
1. just add zizmor (with current version) to `requirements/lint.txt` 
2. Add `zizmor .` call in lint section of tox.ini
3. fix issues reported by zizmor when `tox -e lint` runs

### Details  

Feel free to post here with zizmor output if in step three it is not obvious how to fix the listed issues. 

I'm assuming adding zizmor does not slow down lint too much: would appreciate a comparison in time.
  
The only problem with the simple plan listed here is that some zizmor checks require a github token:
* so the local `tox -e lint` run will not run those checks by default anyway -- I think this is acceptable
* we would have to expose GH_TOKEN in the tox environment if we want those specific checks to run in CI

I think it's fine to just not give zizmor the token in the first iteration: we can improve later.

NicholasTanz:

Fixes #2793

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,6 +20,7 @@ jobs:
       - name: Checkout release tag
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          persist-credentials: false
           ref: ${{ github.event.workflow_run.head_branch }}
 
       - name: Set up Python
@@ -66,8 +67,8 @@ jobs:
             res = await github.rest.repos.createRelease({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              name: '${{ github.ref_name }}-rc',
-              tag_name: '${{ github.ref }}',
+              name: process.env.REF_NAME-rc,
+              tag_name: process.env.REF,
               body: fs.readFileSync('changelog', 'utf8'),
             });
 
@@ -81,6 +82,9 @@ jobs:
               });
             });
             return res.data.id
+        env:
+          REF_NAME: ${{ github.ref_name }}
+          REF: ${{ github.ref }}
 
   release:
     name: Release
@@ -108,6 +112,10 @@ jobs:
             github.rest.repos.updateRelease({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              release_id: '${{ needs.candidate_release.outputs.release_id }}',
-              name: '${{ github.ref_name }}',
+              release_id: process.env.RELEASE_ID,
+              name: process.env.REF_NAME,
             })
+        
+        env:
+          REF_NAME: ${{ github.ref_name }}
+          RELEASE_ID: ${{ needs.candidate_release.outputs.release_id }}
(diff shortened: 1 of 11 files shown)


.github/workflows/cd.yml
```diff
@@ -66,8 +67,8 @@ jobs:
             res = await github.rest.repos.createRelease({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              name: '${{ github.ref_name }}-rc',
-              tag_name: '${{ github.ref }}',
+              name: process.env.REF_NAME-rc,
```
jku: this looks suspicious, I would assume you need a template literal or some other string manipulation here

.github/workflows/cd.yml
```diff
@@ -81,6 +82,9 @@ jobs:
               });
             });
             return res.data.id
+        env:
+          REF_NAME: ${{ github.ref_name }}
+          REF: ${{ github.ref }}
```
jku, line 87: I think this approach is fine...  

 `actions/github-script` does make many of the github variables accessible through the `context` object -- so  env section would not be needed here, but I'm not sure if ref_name is included there so this seems like a decent and understandable way to do it.

jku (changes requested):
Nice, thanks.
* left one change request: the release workflow is a bit tricky to test but I'm guessing the rc release name is not correct right now. This seems like the only merge blocking thing
* On personas: I believe "auditor" persona is not appropriate but "pedantic" might be --  feel free to experiment if you want but be aware that some of the suggestions might require discussion. Keeping the default is totally fine for this PR though
* The output is a little verbose maybe...  Can you fiddle with verbosity settings to get the output to roughly match the other tools (in `tox -e lint`) -- one or two lines of output from zizmor would be ideal for successful run

.github/workflows/cd.yml
NicholasTanz: ah yes, my bad. I added string concentration.

NicholasTanz:
Pedantic covered all of the suppressed results (5) and they seem reasonable to me  

4/5 of them were due to unpinned "uses", but those were all unpinned with the reason that security wasn't critical. I tried using the [ignore](https://woodruffw.github.io/zizmor/usage/#with-comments)  with the trailing explanation but it didn't seem to work. (Ignoring without explanation worked fine.) 

1/5 was due to excessive permissions and even though there's just one job in that workflow, I think it makes sense to specify the permissions in that job

NicholasTanz:
> * On personas: I believe "auditor" persona is not appropriate but "pedantic" might be --  feel free to experiment if you want but be aware that some of the suggestions might require discussion. Keeping the default is totally fine for this PR though

switching to pedantic covered all of the suppressed results (5) and they seem fairly reasonable to me

4/5 of them were due to unpinned "uses", but those were all unpinned with the reason that security wasn't critical. ~~I tried using the [ignore](https://woodruffw.github.io/zizmor/usage/#with-comments) with the trailing explanation but it didn't seem to work.~~ 

edit: My bad, the `ignore with trailing explanation` feature isn't out yet as of v1.3.1. 

1/5 was due to excessive permissions and even though there's just one job in that workflow, I think it makes sense to specify the permissions in the job

jku (approved):
Thanks, LGTM

PR #2808, opened 2025-03-07
ngclient: Create directories as needed

### LINKED ISSUE #2807: ngclient: allow metadata dir to not exist
Currently we expect the local metadata dir to exist when Updater initializes. This made sense when we also expected the caller to put the initial root into the local metadata dir, but we no longer require that (as the bootstrap argument was added).  

It should be ok for the metadata dir (and likewise the artifact dir) to not exist when Updater starts up, and the Updater should create those dirs.

jku:
Make sure ngclient creates metadata and artifact directories when needed.

This was not useful when caller needed to populate the metadata dir with the initial root anyway, but now with boostrap argument it becomes a usability improvement.

Fixes #2807

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -58,6 +58,7 @@ import logging
 import os
 import shutil
 import tempfile
+from pathlib import Path
 from typing import TYPE_CHECKING, cast
 from urllib import parse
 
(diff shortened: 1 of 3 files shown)


tuf/ngclient/updater.py
```diff
@@ -267,6 +268,7 @@ def download_target(
 
         if filepath is None:
             filepath = self._generate_target_file_path(targetinfo)
+        Path(filepath).parent.mkdir(exist_ok=True, parents=True)
```
jku, line 271: This is a case where Pathlib is much more ergonomic so I'm using it even though we don't use Paths generally

kairoaraujo (approved):
LGTM

PR #2812, opened 2025-03-14
Include version in docs

### LINKED ISSUE #1568: docs: python-tuf version should be visible in docs (readthedocs)

jku:
* Without this when reading "stable" docs in readthedocs it's not clear what version the docs are for
* This still leaves "latest" a little unclear on readthedocs (since we don't set a -dev version number after release) but this is still an improvement

Fixes #1568

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,5 +1,5 @@
-TUF Developer Documentation
-===========================
+Python-TUF |version| Developer Documentation
+=======================================================================
 
 This documentation provides essential information for those developing software
 with the `Python reference implementation of The Update Framework (TUF)


jku:
https://theupdateframework--2812.org.readthedocs.build/en/2812/

joshuagl (approved):
Nice improvement, thanks.

docs/index.rst
```diff
@@ -1,5 +1,5 @@
-TUF Developer Documentation
-===========================
+Python-TUF |version| Developer Documentation
+=======================================================================
```
kairoaraujo, line 2:
```suggestion
Python-TUF |version| Developer Documentation
============================================
```

jku: I could have mentioned this: The issue is that the line needs to be at least as long as the title -- and we don't know how long the title is because of the variable expansion.

Luckily it's OK for the line to be longer

PR #2815, opened 2025-03-18
Port securesystemslib.hash module

lukpueh:
securesystemslib.hash is a small wrapper around hashlib, which serves two main purposes:
* provide helper function to hash a file
* translate custom hash algorithm name "blake2b-256" to "blake2b" with (digest_size=32).

In preparation for the removal of securesystemslib.hash, this patch ports above behavior to tuf and uses the builtin hashlib directly where possible.

related secure-systems-lab/securesystemslib#943

### DIFF
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -80,7 +80,7 @@ logger = logging.getLogger(__name__)
 
 SPEC_VER = ".".join(SPECIFICATION_VERSION)
 
-_DEFAULT_HASH_ALGORITHM = "sha256"
+_HASH_ALGORITHM = "sha256"
 
 
 @dataclass
@@ -294,9 +294,9 @@ class RepositorySimulator(FetcherInterface):
         self, role: str
     ) -> tuple[dict[str, str], int]:
         data = self.fetch_metadata(role)
-        digest_object = hashlib.new(_DEFAULT_HASH_ALGORITHM)
+        digest_object = hashlib.new(_HASH_ALGORITHM)
         digest_object.update(data)
-        hashes = {_DEFAULT_HASH_ALGORITHM: digest_object.hexdigest()}
+        hashes = {_HASH_ALGORITHM: digest_object.hexdigest()}
         return hashes, len(data)
 
     def update_timestamp(self) -> None:
--- a/tests/repository_simulator.py
+++ b/tests/repository_simulator.py
@@ -45,6 +45,7 @@ Example::
 from __future__ import annotations
 
 import datetime
+import hashlib
 import logging
 import os
 import tempfile
@@ -52,7 +53,6 @@ from dataclasses import dataclass, field
 from typing import TYPE_CHECKING
 from urllib import parse
 
-import securesystemslib.hash as sslib_hash
 from securesystemslib.signer import CryptoSigner, Signer
 
 from tuf.api.exceptions import DownloadHTTPError
@@ -79,6 +79,8 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 SPEC_VER = ".".join(SPECIFICATION_VERSION)
+
+_DEFAULT_HASH_ALGORITHM = "sha256"
 
 
 @dataclass
@@ -292,9 +294,9 @@ class RepositorySimulator(FetcherInterface):
         self, role: str
     ) -> tuple[dict[str, str], int]:
         data = self.fetch_metadata(role)
-        digest_object = sslib_hash.digest(sslib_hash.DEFAULT_HASH_ALGORITHM)
+        digest_object = hashlib.new(_DEFAULT_HASH_ALGORITHM)
         digest_object.update(data)
-        hashes = {sslib_hash.DEFAULT_HASH_ALGORITHM: digest_object.hexdigest()}
+        hashes = {_DEFAULT_HASH_ALGORITHM: digest_object.hexdigest()}
         return hashes, len(data)
 
     def update_timestamp(self) -> None:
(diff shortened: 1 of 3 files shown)


tuf/api/_payload.py
```diff
@@ -45,6 +48,31 @@
 T = TypeVar("T", "Root", "Timestamp", "Snapshot", "Targets")
 
 
+def _hash(algo: str) -> Any:  # noqa: ANN401
```
lukpueh: Python's hashlib does not make the `Hash` type public, hence I'm using `Any` and ask the linter for forgiveness.

tuf/api/_payload.py
```diff
@@ -664,19 +692,15 @@ def _verify_hashes(
         data: bytes | IO[bytes], expected_hashes: dict[str, str]
     ) -> None:
         """Verify that the hash of ``data`` matches ``expected_hashes``."""
-        is_bytes = isinstance(data, bytes)
```
lukpueh, line 667: Moving this further down appeases mypy.

tests/repository_simulator.py
```diff
@@ -80,6 +80,8 @@
 
 SPEC_VER = ".".join(SPECIFICATION_VERSION)
 
+_DEFAULT_HASH_ALGORITHM = "sha256"
```
lukpueh: IMO it's okay to make this a local default, and not re-use the one newly added to _payload.py, so that we can keep the latter internal.

tuf/api/_payload.py
```diff
@@ -45,6 +48,31 @@
...
+
+    return hashlib.new(algo)
+
+
+def _file_hash(f: IO[bytes], algo: str) -> Any:  # noqa: ANN401
+    """Returns hashed file."""
+    f.seek(0)
+    try:
+        # Only availble on Pythons 3.11+
+        digest = hashlib.file_digest(f, lambda: _hash(algo))  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
lukpueh: Hm. mypy running on ci, thinks this ignore is unnecessary, probably because it runs on a version of Python <3.11. :(

tuf/api/_payload.py
```diff
@@ -45,6 +49,30 @@
...
+        return hashlib.blake2b(digest_size=32)
+
+    return hashlib.new(algo)
+
+
+def _file_hash(f: IO[bytes], algo: str) -> Any:  # noqa: ANN401
+    """Returns hashed file."""
+    f.seek(0)
+    if sys.version_info >= (3, 11):
+        digest = hashlib.file_digest(f, lambda: _hash(algo))  # type: ignore[arg-type]
```
lukpueh: Mypy does not allow `IO[bytes]` for f here. This may be related to https://github.com/python/typing/issues/659. Ignore seems fair.

tuf/api/_payload.py
```diff
@@ -731,21 +755,16 @@ def _get_length_and_hashes(
...
                 else:
-                    digest_object = sslib_hash.digest_fileobject(
-                        data, algorithm
-                    )
-            except (
-                sslib_exceptions.UnsupportedAlgorithmError,
-                sslib_exceptions.FormatError,
-            ) as e:
+                    digest_object = _file_hash(data, algorithm)
+            except (ValueError, TypeError) as e:
```
jku: the different functionality is a little annoying here (you need to call update on one digest_object but not the other)... 
* If the `_hash` and `_file_hash` functions directly returned the hexdigest this code would be simpler and the functions would be properly annotated
* on the other hand then _hash() would need a data argument (meaning you would have to split the lambda needed by hashlib.file_digest() into another function)

I don't really know what is the simplest... up to you

tests/repository_simulator.py
jku: sure, I'd be fine with just a magic value as well: it's not really even a default, it's just what this repository decided to use

tuf/api/_payload.py
jku: theoretically less efficient but seems like a fine compromise

jku (approved):
looks fine to me. Left some comments

jku:
> coverage: 96.981% (-0.1%) from 97.076%

coveralls.io is down again so I'm not sure... but I'm guessing this is the "blake2b-256" special case -- otherwise we're probably exercising this code.

tuf/api/_payload.py
lukpueh: Yes, it's a shame. Maybe there's a better way to narrow the type here, but I couldn't find one. Let's come back to the performance question, if it becomes an issue.

tuf/api/_payload.py
lukpueh: Does [535a189](https://github.com/theupdateframework/python-tuf/pull/2815/commits/535a18918bff8896642ac13602aa3c53e3bdb14b) look better? I don't have very strong feelings about either way.

lukpueh:
> > coverage: 96.981% (-0.1%) from 97.076%
> 
> coveralls.io is down again so I'm not sure... but I'm guessing this is the "blake2b-256" special case -- otherwise we're probably exercising this code.

Yep, blake is the culprit. Let me add a test.

tuf/api/_payload.py
jku: No passionate argument here either.  I think I like this new version better.

lukpueh:
@jku, want to re-approve after my latest changes?

jku (approved):
Looks good. The only nit I have left is that multiple docstrings still refer to _"securesystemslib default hash algorithm"_.

lukpueh:
> Looks good. The only nit I have left is that multiple docstrings still refer to "securesystemslib default hash algorithm".

Good eyes! I pushed a fix and will merge when ci returns.

PR #2819, opened 2025-03-26
Start statically analyzing securesystemslib calls

jku:
DRAFT: This requires a new securesystemslib release

The code changes are all in tests which is a good sign.


jku:
oops, there are actual test failures: did not think of checking those :joy:

jku:
will add this in the upgrade PR #2826

PR #2897, opened 2026-01-08
Bump minimum python version

jku:
Require Python 3.10:
* 3.9 is now EOL and a lot of the development dependencies require 3.10 already in latest versions
* We could just bump the minimum tested version (like we did with 3.8 -> 3.9) but I think this time the risk of accidentally breaking 3.9 anyway is higher (since annotation improvements creep in): makes sense to just require 3.10 at this point
* Also start testing with 3.14
* All of the code changes are linter suggestions

unblocks #2895, #2894

### DIFF
--- a/tuf/api/_payload.py
+++ b/tuf/api/_payload.py
@@ -1191,7 +1191,7 @@ class DelegatedRole(Role):
 
         # Every part in the pathpattern could include a glob pattern, that's why
         # each of the target and pathpattern parts should match.
-        for target, pattern in zip(target_parts, pattern_parts, strict=False):
+        for target, pattern in zip(target_parts, pattern_parts, strict=True):
             if not fnmatch.fnmatch(target, pattern):
                 return False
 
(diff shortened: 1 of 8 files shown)


tuf/api/_payload.py
```diff
@@ -1195,8 +1191,8 @@ def _is_target_in_pathpattern(targetpath: str, pathpattern: str) -> bool:
 
         # Every part in the pathpattern could include a glob pattern, that's why
         # each of the target and pathpattern parts should match.
-        for target_dir, pattern_dir in zip(target_parts, pattern_parts):
-            if not fnmatch.fnmatch(target_dir, pattern_dir):
+        for target, pattern in zip(target_parts, pattern_parts, strict=False):
```
jku: I suppose this could be strict=True to be sure... but we do check the lengths a few lines above this

lukpueh (approved):

PR #2903, opened 2026-01-25
feat(ngclient): require explicit bootstrap argument

1seal:
# Description of the changes being introduced by the pull request:

This PR makes the trust anchor choice explicit in tuf.ngclient.Updater().

  - bootstrap is now a required keyword-only argument (no default)
  - Callers must choose:
      - bootstrap=<root_bytes> (recommended: embedded/deployed trusted root)
      - bootstrap=None (explicit opt-in to using cached metadata_dir/root.json as the trust anchor)
  - The fallback now triggers only on bootstrap is None (not on falsy bytes)
  - Tests and examples are updated to pass bootstrap explicitly
  - Adds documentation guidance on secure bootstrap root storage

Migration:

  - Old implicit behavior: Updater(...)
  - Preserve old behavior explicitly: Updater(..., bootstrap=None)
  - Recommended: Updater(..., bootstrap=<embedded root bytes>)

Tests:

  - python -m pytest -c pyproject.toml -q

  Ref: GHSA-9pfj-pjv5-22gj

### DIFF
--- a/docs/INSTALLATION.rst
+++ b/docs/INSTALLATION.rst
@@ -53,6 +53,39 @@ from GitHub, change into the project root directory, and install with pip
    python3 -m pip install -r requirements/dev.txt
 
 
+Bootstrap root metadata
+-----------------------
+
+The initial trusted root metadata (``root.json``) is the trust anchor for all
+subsequent metadata verification. Applications should deploy a trusted root
+with the application and provide it to :class:`tuf.ngclient.Updater`.
+
+Recommended storage locations for bootstrap root metadata include:
+
+* a system-wide read-only path (e.g. ``/usr/share/your-app/root.json``)
+* an application bundle with appropriate permissions
+* a read-only mounted volume in containerized deployments
+
+Not recommended:
+
+* ``metadata_dir`` (the metadata cache) since it is writable by design
+* user-writable install paths (e.g. a user site-packages directory)
+* any location writable by the account running the updater
+
+Example::
+
+   from tuf.ngclient import Updater
+
+   with open("/usr/share/your-app/root.json", "rb") as f:
+       bootstrap = f.read()
+
+   updater = Updater(
+       metadata_dir="/var/lib/your-app/tuf/metadata",
+       metadata_base_url="https://example.com/metadata/",
+       bootstrap=bootstrap,
+   )
+
+
 Verify release signatures
 -------------------------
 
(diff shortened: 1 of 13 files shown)


tests/test_updater_consistent_snapshot.py
jku: this could maybe be removed now that _init_updater() does the right thing?

This probably applies to most test files using the repository simulator

docs/INSTALLATION.rst
```diff
@@ -53,6 +53,39 @@ from GitHub, change into the project root directory, and install with pip
    python3 -m pip install -r requirements/dev.txt
 
 
+Bootstrap root metadata
```
jku: This file is otherwise mostly about developer installs... Maybe we could rename this section to "Application deployment" or something to make it clearer that this is another subject completely?

jku (commented):
This looks good to me, left a couple of suggestions. Let me know what you think.

1seal:
removed the redundant root.json writes in tests where Updater(..., bootstrap=<bytes>) is already passed (incl. test_updater_consistent_snapshot and other repository simulator tests). i kept the distinction for bootstrap=None: tests that require cached root still write it, and ‘no cache + bootstrap=None’ cases keep the cache absent. also renamed the INSTALLATION section to ‘Application deployment’ to clarify the context

jku:
Did you forget to push the branch or do I misunderstand?

1seal:
yes, thanks for the reminder. removed the redundant root.json writes from test files that pass explicit bootstrap and renamed the INSTALLATION section to "Application deployment" as suggested

jku:
apologies for the annoyingly long review cycles at the moment.

* the formatter is complaining (`tox -e lint` will check locally and `ruff format tuf examples tests verify_release .github/scripts` should fix it if you have ruff installed)
* the last commit does not have signed-off-by 

I can also fix these myself (and override DCO) but then we'll wait for another reviewer to have a look, let me know how you want to handle this.

1seal:
@jku i’ve addressed both points:

- ran ruff format on tuf examples tests verify_release .github/scripts (so tox -e lint is clean now)
- amended the last commit to include Signed-off-by (dco)

force-pushed the updated branch to the pr. let me know if you’d prefer i avoid force-pushes and send follow-up commits instead.

jku (approved):
Thanks!

Issue #2228, opened 2022-12-17
repository module: Add tests


jku:
Once #2193 is merged, there is a repository module. The module is not really tested but should be.

It's entirely possible that the module can't really be effectively unit tested (as it's an ABC). But the repository example can be end-to-end tested and maybe even the SimpleRepository class within the example could be unit tested. I think testing the whole repository example is most critical. 

I'm not sure what kind of things we should test but at least
* refreshing a client, downloading a file from the repository
* making sure the repository metadata actually changes, and that the updated metadata is correct (i.e. client that has metadata version N can successfully update to N+M)

jku:
This should include the uploader as well. The basic end-to-end test for all examples would be: 
1. run `repository/repo` in one process
1. wait until http server is up
1. run `uploader/uploader` (in another process) to initialize with trust-on-first-use,  create a delegation and to add a target 
1. wait until upload has finished
1. run `client/client` (in 3rd process or reusing the uploader process) to initialize with trust-on-first-use and download target

This would be useful even without any other checks (like checking which metadata versions are used): if the client can download the file and the contents are correct, that's a really good indication the examples are running pretty well.

PR #2948, opened 2026-06-03
tests: Add end-to-end test for the examples

arpitjain099:

This adds an end-to-end test that runs the `repository`, `uploader` and `client` examples against each other, which is the part of #2228 that was not covered by the `tuf.repository` interface tests already merged in #2651.

The new `tests/test_examples_e2e.py` drives the three examples the way a user runs them from the command line (see their READMEs):

* `examples/repository/repo` is started as a live HTTP server on a free port
* `examples/uploader/uploader` does Trust-On-First-Use, claims a delegation, and uploads target files to that server
* `examples/client/client` does Trust-On-First-Use and downloads the uploaded targets

Because the client downloads through `ngclient.Updater`, a successful download means the metadata chain and the target hash both verified, so the test asserts the downloaded content matches what the uploader published.

It also covers the "client that has metadata version N can successfully update to N+M" case from the issue: after the client is initialized and has downloaded a first target, the uploader adds a second target (which publishes new targets, snapshot and timestamp metadata). The test checks that the snapshot version advanced and that the already-initialized client can then refresh and download the new target.

A few notes on how it stays deterministic and isolated:

* the example scripts import their private helper modules by name (`from _simplerepo import SimpleRepository`), so each example dir is put on `PYTHONPATH`
* the client and uploader store trusted metadata and keys under `Path.home()`, so `HOME` is pointed at a temporary directory; nothing touches the real home dir and every run starts from a clean repository
* readiness is handled by polling for `1.root.json` (reusing `tests.utils.wait_for_server` for the port) rather than sleeping

The examples themselves are not modified; this only wires the existing ones together as a test. Ran locally with `python -m unittest tests.test_examples_e2e` (also green under the project's surrounding example tests), and `ruff check` / `ruff format` / `mypy` are clean on the new file.

I kept this scoped to the examples e2e slice. Whether the remaining idea in the issue (unit-testing `SimpleRepository` directly) is still wanted, and whether #2228 should stay open for it, is up to you.

Addresses #2228

requirements**:

### DIFF
--- a/tests/test_examples_e2e.py
+++ b/tests/test_examples_e2e.py
@@ -1,0 +1,232 @@
+# Copyright 2026, the TUF contributors
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+"""End-to-end test that runs the repository, uploader and client examples.
+
+Unlike test_examples.py (which execs the self-contained 'manual_repo'
+scripts), this test wires the three interacting examples together the way
+a user would run them from the command line:
+
+* 'examples/repository/repo' is started as a live HTTP server,
+* 'examples/uploader/uploader' performs Trust-On-First-Use, claims a
+  delegation and uploads target files to that server,
+* 'examples/client/client' performs Trust-On-First-Use and downloads the
+  uploaded targets, which verifies them against the repository metadata.
+
+The test also uploads a second target after the client has already been
+initialized, then downloads it: this exercises the case from issue #2228
+where a client with metadata version N must update to version N+M.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import shutil
+import socket
+import subprocess
+import sys
+import tempfile
+import time
+import unittest
+from pathlib import Path
+from typing import ClassVar
+from urllib.parse import quote
+
+import urllib3
+
+from tests import utils
+
+# The repository example serves metadata under this URL prefix.
+_METADATA_URL = "metadata"
+
+
+def _get_free_port() -> int:
+    """Return a currently-free localhost TCP port.
+
+    The port is handed to the example repository server via '-p'. There is a
+    small race between closing this socket and the server binding the port,
+    but the readiness poll in setUp() tolerates it.
+    """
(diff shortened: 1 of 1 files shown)


jku:
I'd really like to have this -- without it the examples are likely to break.

From CI:

> TimeoutError: Could not connect to the repository example on port 39201

It's been so long since I touched tests that actually setup http servers that I don't have any useful advice but that issue does reproduce locally for me.

arpitjain099:
Thanks, glad it's wanted. The CI failure is the examples repository server not being ready when the client connects (the TimeoutError on the port). I'll make the test wait until the server is actually reachable before the client step so it's deterministic, rather than relying on a fixed sleep. Updating the branch shortly.

jku (changes requested):
I'll mark this request changes (on account of the CI failure) so it's visible on the PR list -- I don't actually know what is needed here

arpitjain099:
@jku found it, and it was mine rather than anything flaky. Fixed in 82a8f4b.

`setUp` was doing this:

```python
self.env["PYTHONPATH"] = os.pathsep.join(
    str(self.examples_dir / name) for name in ("repository", "uploader", "client")
)
```

That replaces `PYTHONPATH` rather than adding to it. Where the test runner is using `PYTHONPATH` to make `tuf` importable, the repository subprocess then cannot import `tuf` at all. It surfaced as the unhelpful "repository example exited early with code 1", with the real cause only visible in the captured child output:

```text
from tuf.api.exceptions import RepositoryError
ModuleNotFoundError: No module named 'tuf'
```

The example directories are now prepended to any existing `PYTHONPATH` instead of overwriting it.

I reproduced the failure rather than guessing at it: uninstall `tuf`, put the checkout on `PYTHONPATH`, and the test fails with exactly that error before the change and passes after. Full suite locally is 206 passed, 193 subtests passed.

Worth noting the log that put me on to it had already expired, so I updated the branch from develop to re-run CI and read the fresh one.

Separately, `test-with-sslib-main` was failing for the same reason, so that should clear too, though I see it is marked as not a merge blocker.

PR #2964, opened 2026-07-08
ngclient: propagate non-timeout urllib3 connection errors

arpitjain099:

While reading the Urllib3Fetcher I noticed that `_fetch()` only handles the timeout case of `MaxRetryError`. When the reason is anything else (a TLS certificate error, for instance) the except block does not re-raise, so execution continues to the `response.status` check while `response` is still unbound. The caller then gets an `UnboundLocalError` about `response` instead of the actual connection error, which makes real failures hard to diagnose.

This re-raises the original error when the reason is not a timeout, so `fetch()` wraps the meaningful error in a `DownloadError` as documented in FetcherInterface. I added a regression test that drives a non-timeout MaxRetryError and checks the underlying cause is preserved. The full test suite passes locally with tox.

Fixes: non-timeout connection errors surfacing as UnboundLocalError

### DIFF
--- a/tests/test_fetcher_ng.py
+++ b/tests/test_fetcher_ng.py
@@ -137,6 +137,25 @@ class TestFetcher(unittest.TestCase):
             self.fetcher.fetch(self.url)
         mock_session_get.assert_called_once()
 
+    # A non-timeout connection failure (e.g. TLS error) must surface as the
+    # original error, not be masked by an UnboundLocalError on "response".
+    @patch.object(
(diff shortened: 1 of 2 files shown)


kairoaraujo (approved):

arpitjain099:
@kairoaraujo thanks for the review. Does this need a second approval before it can merge, or is it good to go?

jku (approved):
looks right to me, thanks for fixing this

PR #2973, opened 2026-08-03
Fix Role and DelegatedRole hashing

Sachith77:
## Description

While going through `_payload.py`, I noticed that `Role.__hash__()` fails because it tries to hash `keyids` (a list) and `unrecognized_fields` (a dict) directly.

This PR:

* Converts `keyids` to a tuple before hashing.
* Excludes `unrecognized_fields` since it may contain nested, unhashable JSON.
* Fixes `DelegatedRole.__hash__()` using `self.path` instead of `self.paths`.
* Converts `paths` and `path_hash_prefixes` to tuples before hashing.
* Adds regression tests for hashing, equality, and set usage.

The full test suite passes locally with `tox`.

I noticed similar patterns in a few other metadata classes, but kept this PR focused on `Role` and `DelegatedRole`.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -34,6 +34,7 @@ from tuf.api.metadata import (
     Delegations,
     Metadata,
     MetaFile,
+    Role,
     Root,
     RootVerificationResult,
     Signature,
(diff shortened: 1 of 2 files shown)


jku (approved):
yeah, it may have been a mistake to add __hash__() to begin with, I think it only happened because ruff complained about not having it and it seemed easy to do... 

In any case these fixes seem reasonable to me, thanks

Sachith77:
Thanks for the review and the context on why hash() was added.

PR #2976, opened 2026-08-05
fix: update client README with valid tuf-demo target

Phantom9869:
The example command in examples/client/README.md references
demo/succinctly-delegated-1.txt which no longer exists in the
tuf-demo repository. The delegation structure currently only
has jku/*, rdimitrov/*, and kommendorkapten/* paths.

Updated the example to use rdimitrov/artifact-example.md which
is currently available and working.


jku (approved):
Thanks!

PR #2978, opened 2026-08-05
Fix incorrect delegation in diamond-shape case

### LINKED ISSUE #2977: Diamond shape delegation can lead to incorrect results
This comes from https://github.com/theupdateframework/tuf-conformance/pull/405

This looks like a valid bug to me, a diamond shape delegation (where a role is delegated to from multiple delegating roles) can in specific circumstances lead to a an incorrect delegation result:
* if client has already verified a delegation through one branch...
* then when the same client (same process) is verifying a delegation via the other delegation branch, the delegation is accepted even if the role is only signed by the first branch parent role

In my opinion diamond delegations likely have no practical use case and should not be used... but it's still a client bug

jku:
In a diamond delegation (where a role is delegated to from multiple delegating roles), if client has already verified a delegation through one branch and is now verifying the delegation through the other branch, it should still make sure the delegation is correctly signed with the second set of signers.

The fix could be smaller (and isolated in updater.py) but I added the refactoring with `TrustedMetadataSet._add()` and `TrustedMetadataSet.contains()` to make the resulting code clearer.

Fixes #2977

Test is ai generated, other code ai assisted.

### DIFF
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -348,15 +348,24 @@ class TestDelegationsGraphs(TestDelegations):
         )
         self._init_repo(test_case)
 
-        # release metadata is only signed by parent-a
+        # Client happy path: Both delegation paths work since release is signed
+        # by both parents
+        updater = self._init_updater()
+        target_a = updater.get_targetinfo("path-a")
+        self.assertIsNotNone(target_a)
+        target_b = updater.get_targetinfo("path-b")
+        self.assertIsNotNone(target_b)
+
+        # Create release v2: only signed by parent-a
         first_keyid = next(iter(self.sim.signers["release"].keys()))
         self.sim.signers["release"] = {
             first_keyid: self.sim.signers["release"][first_keyid]
         }
+        self.sim.md_delegates["release"].signed.version += 1
         self.sim.update_snapshot()
 
         # Client step 1: Get path-a targetinfo via parent-a -> release
-        # This primes TrustedMetadataSet to contain release metadata
+        # This primes TrustedMetadataSet to contain release v2
         updater = self._init_updater()
         target_a = updater.get_targetinfo("path-a")
         self.assertIsNotNone(target_a)
--- a/tests/test_updater_delegation_graphs.py
+++ b/tests/test_updater_delegation_graphs.py
@@ -330,6 +330,43 @@ class TestDelegationsGraphs(TestDelegations):
         finally:
             self.teardown_subtest()
 
+    def test_shared_delegation_signature_check(self) -> None:
+        """Test that a role delegated by multiple parents is signature-checked
+        for each delegator link.
+        """
+        test_case = DelegationsTestCase(
+            delegations=[
+                TestDelegation("targets", "parent-a", paths=["path-a"]),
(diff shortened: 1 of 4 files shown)


loosebazooka (approved):

tuf/ngclient/_internal/trusted_metadata_set.py
```diff
@@ -127,6 +128,17 @@ def __init__(self, root_data: bytes, envelope_type: EnvelopeType):
         logger.debug("Updating initial trusted root")
         self._load_trusted_root(root_data)
 
+    def contains(self, role: str, delegator: str) -> bool:
+        """Check if ``role`` is in ``TrustedMetadataSet`` and was verified
+        against ``delegator``.
+        """
+        return (delegator, role) in self._trusted_delegations
+
+    def _add(self, role: str, signed: Signed, delegator: str) -> None:
```
loosebazooka, line 137: so _add only matters for the delegated targets, but we're using it to standardize everything? We can't hit the diamond problem in the main roles?

tests/test_updater_delegation_graphs.py
```diff
@@ -330,6 +330,43 @@ def test_invalid_metadata(self, test_data: DelegationsTestCase) -> None:
         finally:
             self.teardown_subtest()
 
+    def test_shared_delegation_signature_check(self) -> None:
```
loosebazooka, line 333: so what's the behavior when it is signed by b? should it be accepted or rejected?

tuf/ngclient/_internal/trusted_metadata_set.py
jku: correct, _add() is used for everything just for consistency: we know the main roles cannot have a delegation issue like this.

tests/test_updater_delegation_graphs.py
jku: then `get_targetinfo("path-b")` should succeed as the metadata is now signed by the correct delegator.

(but I won't blame an implementation that just errors out the moment a diamond delegation is seen, even if all delegators have correctly signed: I can't imagine a real use case)

loosebazooka: sorry I meant to imply we should codify that in test

jku: Sure. Knowing the code this feels impossible to go wrong but it's all complicated enough that it makes sense to include

tests/test_updater_delegation_graphs.py
```diff
@@ -348,15 +348,24 @@ def test_shared_delegation_signature_check(self) -> None:
...
-        # release metadata is only signed by parent-a
+        # Client happy path: Both delegation paths work since release is signed
+        # by both parents
+        updater = self._init_updater()
+        target_a = updater.get_targetinfo("path-a")
+        self.assertIsNotNone(target_a)
+        target_b = updater.get_targetinfo("path-b")
+        self.assertIsNotNone(target_b)
+
+        # Create release v2: only signed by parent-a
```
loosebazooka, line 359: A comment here may help differentiate from the above test, but... I'm okay either way.

loosebazooka (approved):

jku:
well, I'm merging now. Further comments in new issues please

PR #2981, opened 2026-08-07
ngclient: Fix symlink usage on Windows

### LINKED ISSUE #2980: symlinks only work on windows/ntfs  with elevated permissions
We use symlinks in ngclient for root.json to avoid keeping multiple copies -- let's keep copies at least on Windows since symlinks don't work

jku:
Running ngclient on Windows (I believe NTFS only, non-privileged account only) leads to
  OSError: [WinError 1314] A required privilege is not held by the client

symlinking is apparently a high privilege operation: let's add a fallback, a real file will work just as well.

Fixes #2980 

PR is partially AI generated

### DIFF
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -371,9 +371,12 @@ class Updater:
             os.remove(linkname)
         try:
             os.symlink(current, linkname)
-        except OSError:
-            # Fallback for NTFS "required privilege is not held by client"
-            os.link(os.path.join(self._dir, current), linkname)
+        except OSError as e:
+            if getattr(e, "winerror", None) == 1314:
+                # Fallback for NTFS "required privilege is not held by client"
+                os.link(os.path.join(self._dir, current), linkname)
+            else:
+                raise
 
     def _load_root(self) -> None:
         """Load root metadata.
--- a/tuf/ngclient/updater.py
+++ b/tuf/ngclient/updater.py
@@ -372,8 +372,8 @@ class Updater:
         try:
             os.symlink(current, linkname)
         except OSError:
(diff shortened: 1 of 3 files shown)


tuf/ngclient/updater.py
```diff
@@ -369,7 +369,11 @@ def _update_root_symlink(self) -> None:
         current = os.path.join("root_history", f"{version}.root.json")
         with contextlib.suppress(FileNotFoundError):
             os.remove(linkname)
-        os.symlink(current, linkname)
+        try:
+            os.symlink(current, linkname)
+        except OSError:
```
lukpueh: Is this not too broad?

lukpueh (approved):
Thanks for the fix!

tuf/ngclient/updater.py
jku: Kind of... but I don't actually have a Windows/NTFS  machine to test the more complicated code that would check for specific error codes, and the workaround _should_ be harmless even if it happens for other reasons: just an extra copy of root.json

lukpueh: Works for me.

tuf/ngclient/updater.py
```diff
@@ -369,7 +369,14 @@ def _update_root_symlink(self) -> None:
         current = os.path.join("root_history", f"{version}.root.json")
         with contextlib.suppress(FileNotFoundError):
             os.remove(linkname)
-        os.symlink(current, linkname)
+        try:
+            os.symlink(current, linkname)
+        except OSError as e:
+            if getattr(e, "winerror", None) == 1314:
```
lukpueh, line 375: I like that better. Thanks!

lukpueh (approved):

PR #2982, opened 2026-08-07
docs: add Windows note for python3 command

Phantom9869:
On a standard Windows Python.org installation, python3 is not
available — only python is. This causes confusion when users
follow the installation commands in INSTALLATION.rst.

Add a note at the top of the Installation section informing
Windows users to use python or the py launcher instead.


jku (approved):
thank you

PR #2992, opened 2026-08-18
ngclient: add download_target_bytes()

### LINKED ISSUE #1556: ngclient: feature request "download target as bytes"
From #1317: When client downloads files it does not always need them written into an actual file, often it just wants the content: providing API the returns just bytes would be fine. We should still cache the target to disk but client could avoid reading the file if we provided variants of `download_target()` and `find_cached_target()` that returned bytes.

the only complication here might be that we really might want to provide an `iterator[bytes]` (because there could be a lot of bytes)... If that is possible that would be cool but that might be more complicated

ChrisJr404:
Adds `Updater.download_target_bytes()`, which downloads and verifies a target and hands back the content as `bytes` instead of writing it to the cache dir, so callers like sigstore-python that just want the verified bytes in memory don't have to round-trip through a file.

The verification path is shared, not duplicated: I pulled the URL-building out of `download_target()` into a small `_target_file_url()` helper, and both methods then do the same `_fetcher.download_file()` + `targetinfo.verify_length_and_hashes()` against the same downloaded stream. The only difference is what happens after verification succeeds - `download_target()` copies to disk, `download_target_bytes()` reads the stream and returns it. So the length/hash checks are byte-for-byte identical between the two.

On the cache-timing concern from #1556 (re #1168): the new method deliberately does not touch the local target cache at all. It doesn't call `find_cached_target()`, doesn't read an existing cached file, and doesn't persist anything. Since it never consults the cache, there's no cache-hit/miss timing to leak - it always downloads and verifies. Callers who want caching keep using `download_target()` + `find_cached_target()`.

One thing worth flagging: because a target's hash can't be checked until it's fully downloaded, the bytes variant buffers the whole target in memory. That's fine for the small artifacts this is aimed at, but it's not suitable for very large targets - I noted that in the docstring. The streaming/iterator idea from the issue thread would be a separate, larger API and I left it out here.

Tests in `tests/test_updater_fetch_target.py`, mirroring the existing `download_target` tests against the repository simulator:
- `test_fetch_target_bytes` (dataset-driven, same three target cases): returned bytes match expected content, and nothing gets written to the cache dir (`find_cached_target` stays `None`).
- `test_invalid_target_download_bytes`: hash mismatch and length mismatch both still raise `RepositoryError`, and nothing is persisted.

`tox`-equivalent locally: `pytest tests/test_updater_*.py` green (updater suites 27 passed / 13 subtests), `black`/`isort` clean at line-length 80, `mypy` clean on `updater.py`, `pylint` 9.91 (only pre-existing `__init__` arg-count warnings, untouched by this change).

Fixes #1556

### DIFF
--- a/tests/test_updater_fetch_target.py
+++ b/tests/test_updater_fetch_target.py
@@ -119,6 +119,51 @@ class TestFetchTarget(unittest.TestCase):
         self.assertEqual(path, updater.find_cached_target(info))
         self.assertEqual(path, updater.find_cached_target(info, path))
 
+    @utils.run_sub_tests_with_dataset(targets)
+    def test_fetch_target_bytes(self, target: TestTarget) -> None:
+        path = os.path.join(self.targets_dir, target.encoded_path)
+
+        # Add target to repository
+        self.sim.targets.version += 1
+        self.sim.add_target("targets", target.content, target.path)
+        self.sim.update_snapshot()
+
+        updater = self._init_updater()
+        info = updater.get_targetinfo(target.path)
+        assert info is not None
+
+        # download_target_bytes returns the verified content
+        self.assertEqual(target.content, updater.download_target_bytes(info))
+
+        # ...and does not touch the local cache
+        self.assertFalse(os.path.exists(path))
+        self.assertIsNone(updater.find_cached_target(info))
+
+    def test_invalid_target_download_bytes(self) -> None:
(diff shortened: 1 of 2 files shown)


jku (changes requested):
I'll mark this "request changes" just so it's visible on the PR list... no pressure to actually do the experiment I suggested though: I'll try this later if you are not interested.

PR #2994, opened 2026-08-23
tests: improve subtest cleanup

### LINKED ISSUE #1719: Improve subtests clean up

In tests we use the home-made decorator `@utils.run_sub_tests_with_dataset` when we need to execute the same test case consecutively for many datasets.  The decorator works perfectly fine when no clean up is needed after each sub test but in the cases where clean up code needs to be executed, the whole sub test needs to be included in a "try - finally" block in order to run the clean up code even if a sub test fails.

**Expected behavior**:
Improve the decorator by possibly including the clean up logic in it to avoid the big "try" blocks in the test code.
See https://github.com/theupdateframework/python-tuf/pull/1689#issuecomment-989794850

Aryadeepta:
**Description**:

Updates `run_sub_tests_with_dataset` to call `teardown_subtest()` in a
`finally` block when it is defined. This ensures per-subtest cleanup runs
even when a subtest fails.

Removes the now-redundant `try/finally` cleanup blocks from the affected
updater tests and adds regression coverage for cleanup after a failing
subtest.

Testing:
- `tox -e lint`
- `tox -e py`

Fixes #1719

### DIFF
--- a/tests/test_updater_consistent_snapshot.py
+++ b/tests/test_updater_consistent_snapshot.py
@@ -126,23 +126,20 @@ class TestConsistentSnapshot(unittest.TestCase):
     ) -> None:
         # Test if the client fetches and stores metadata files with the
         # correct version prefix, depending on 'consistent_snapshot' config
-        try:
-            consistent_snapshot: bool = test_case_data["consistent_snapshot"]
-            exp_calls: list[Any] = test_case_data["calls"]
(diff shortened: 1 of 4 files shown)


jku (approved):
Yeah looks about as good as we can get without switching to pytest.

Thanks for the patch.

PR #2995, opened 2026-08-24
Fix __hash__ crashing on unhashable dict attributes

Sachith77:
## Description
Follow-up to #2973, which fixed `__hash__` for `Role` and `DelegatedRole`.

Same problem was still there in the other metadata classes: `Signed`, `Root`, `MetaFile`, `Snapshot`, `Delegations`, `TargetFile`, `Targets`, and `Metadata`. Their `__hash__` implementations passed raw dictionaries into `hash()`, which throws `TypeError: unhashable type: 'dict'`.

`Timestamp.__hash__` looked fine on its own, but still breaks because it inherits from `Signed` and `MetaFile`.

All these classes define `__eq__`, so they need a working `__hash__` too if they're going to be used in sets or as dict keys. `test_metadata_eq_.py` covers equality for these classes but never tested hashing, which is probably why this slipped through.

Fix follows the same approach as #2973 — hash a subset of immutable fields instead of the raw dicts.

`unrecognized_fields` is left out since it can hold arbitrary nested JSON. Everything else I just converted to tuples.

## Testing
- `test_metadata_hash` checks that for each affected class, `hash()` doesn't raise, equal objects produce equal hashes, and instances can be used as set members. This fails on `develop` with 8 subtest errors before the fix.
- `test_metadata_hash_ignores_unrecognized_fields` checks the deliberate exclusion of `unrecognized_fields` — objects that only differ in that field are unequal but can share a hash, which is fine given the data model.

### DIFF
--- a/tuf/api/_payload.py
+++ b/tuf/api/_payload.py
@@ -1466,7 +1466,7 @@ class Delegations:
             (
                 tuple(sorted(self.keys)),
                 # Order of the delegated roles matters (see __eq__)
-                tuple(self.roles) if self.roles is not None else None,
+                tuple(self.roles.items()) if self.roles is not None else None,
                 self.succinct_roles,
             )
         )
--- a/tuf/api/_payload.py
+++ b/tuf/api/_payload.py
@@ -188,7 +188,6 @@ class Signed(metaclass=abc.ABCMeta):
                 self.version,
                 self.spec_version,
                 self.expires,
-                self.unrecognized_fields,
             )
         )
 
@@ -565,10 +564,9 @@ class Root(Signed, _DelegatorMixin):
         return hash(
             (
                 super().__hash__(),
-                self.keys,
-                self.roles,
+                tuple(sorted(self.keys)),
+                tuple(sorted(self.roles)),
                 self.consistent_snapshot,
-                self.unrecognized_fields,
             )
         )
 
@@ -848,9 +846,7 @@ class MetaFile(BaseFile):
         )
 
     def __hash__(self) -> int:
-        return hash(
-            (self.version, self.length, self.hashes, self.unrecognized_fields)
-        )
+        return hash((self.version, self.length))
 
     @classmethod
     def from_dict(cls, meta_dict: dict[str, Any]) -> MetaFile:
@@ -1031,7 +1027,7 @@ class Snapshot(Signed):
         return super().__eq__(other) and self.meta == other.meta
 
     def __hash__(self) -> int:
-        return hash((super().__hash__(), self.meta))
+        return hash((super().__hash__(), len(self.meta)))
 
     @classmethod
     def from_dict(cls, signed_dict: dict[str, Any]) -> Snapshot:
@@ -1463,10 +1459,10 @@ class Delegations:
     def __hash__(self) -> int:
         return hash(
             (
-                self.keys,
-                self.roles,
+                tuple(sorted(self.keys)),
+                # Order of the delegated roles matters (see __eq__)
+                tuple(self.roles) if self.roles is not None else None,
                 self.succinct_roles,
-                self.unrecognized_fields,
             )
         )
 
@@ -1592,9 +1588,7 @@ class TargetFile(BaseFile):
         )
 
     def __hash__(self) -> int:
-        return hash(
-            (self.length, self.hashes, self.path, self.unrecognized_fields)
-        )
+        return hash((self.length, self.path))
 
     @classmethod
     def from_dict(cls, target_dict: dict[str, Any], path: str) -> TargetFile:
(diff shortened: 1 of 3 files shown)


tuf/api/_payload.py
```diff
@@ -848,9 +846,7 @@ def __eq__(self, other: object) -> bool:
         )
 
     def __hash__(self) -> int:
-        return hash(
-            (self.version, self.length, self.hashes, self.unrecognized_fields)
-        )
+        return hash((self.version, self.length))
```
jku: what about self.hashes here?

tuf/api/_payload.py
```diff
@@ -1031,7 +1027,7 @@ def __eq__(self, other: object) -> bool:
         return super().__eq__(other) and self.meta == other.meta
 
     def __hash__(self) -> int:
-        return hash((super().__hash__(), self.meta))
+        return hash((super().__hash__(), len(self.meta)))
```
jku: `len(self.meta)` sounds a bit strange

tuf/api/_payload.py
```diff
@@ -1592,9 +1588,7 @@ def __eq__(self, other: object) -> bool:
         )
 
     def __hash__(self) -> int:
-        return hash(
-            (self.length, self.hashes, self.path, self.unrecognized_fields)
-        )
+        return hash((self.length, self.path))
```
jku: what about self.hashes?

tuf/api/_payload.py
```diff
@@ -1740,7 +1734,7 @@ def __eq__(self, other: object) -> bool:
         )
 
     def __hash__(self) -> int:
-        return hash((super().__hash__(), self.targets, self.delegations))
+        return hash((super().__hash__(), len(self.targets), self.delegations))
```
jku: `len(self.targets)` sounds wrong

jku (commented):
yeah I guessed the rest of them had issues after #2973... Left some questions in code.

tuf/api/_payload.py
```diff
@@ -565,10 +564,9 @@ def __hash__(self) -> int:
         return hash(
             (
                 super().__hash__(),
-                self.keys,
-                self.roles,
+                tuple(sorted(self.keys)),
+                tuple(sorted(self.roles)),
```
jku: I *think* this hashes just the role names. How about this:
```suggestion
                tuple(sorted(self.roles.items())),
```

tuf/api/_payload.py
```diff
@@ -1463,10 +1464,10 @@ def __eq__(self, other: object) -> bool:
     def __hash__(self) -> int:
         return hash(
             (
-                self.keys,
-                self.roles,
+                tuple(sorted(self.keys)),
+                # Order of the delegated roles matters (see __eq__)
+                tuple(self.roles) if self.roles is not None else None,
```
jku: same as above but for roles: I think this just includes role names right now.
```suggestion
                tuple(self.roles.items()) if self.roles is not None else None,
```

tuf/api/_payload.py
```diff
@@ -565,10 +564,9 @@ def __hash__(self) -> int:
         return hash(
             (
                 super().__hash__(),
-                self.keys,
-                self.roles,
+                tuple(sorted(self.keys)),
```
jku: this also hashes just key names... unfortunately the correct thing 
```suggestion
                tuple(sorted(self.keys.items())),
```

will keep failing since securesystemslib hashes are still not complete

tuf/api/_payload.py
```diff
@@ -1463,10 +1464,10 @@ def __eq__(self, other: object) -> bool:
     def __hash__(self) -> int:
         return hash(
             (
-                self.keys,
-                self.roles,
+                tuple(sorted(self.keys)),
```
jku: this also hashes just keyids... unfortunately the correct thing 
```suggestion
                tuple(sorted(self.keys.items())),
```

will likely keep failing since securesystemslib hashes are still not complete

tuf/api/metadata.py
```diff
@@ -148,7 +148,7 @@ def __eq__(self, other: object) -> bool:
         )
 
     def __hash__(self) -> int:
-        return hash((self.signatures, self.signed, self.unrecognized_fields))
+        return hash((tuple(self.signatures), self.signed))
```
jku: this is another case that hashes just keyids of the signatures... unfortunately the correct thing 
```suggestion
        return hash((tuple(self.signatures.items()), self.signed))
```

will keep failing since securesystemslib hashing is still not complete

jku (changes requested):
I spent a bit more time and I think it's still not complete, details in code.

Unfortunately several cases will keep failing even with (what I think is) correct code as the securesystemslib hashing seems to still not be complete. I'm fine with either 
* fixing `Key` and `Signature` hashing first in securesystemslib, then completing this
* Completing this PR (even if that means hashing Metadata objects fails until securesystemslib is fixed) -- just need to add some special case in tests

Sachith77:
@jku could you please have a look.

PR #2990, opened 2026-08-16
updates urllib3 dependency specifier

micahculpepper:
This PR updates the project's urllib3 dependency specifier. Previously, tuf's metadata claimed that urllib3 version 1.21.1+ was supported, but actually, tuf v6.0.0 requires urllib3 version 2+

Fixes #2989

PR #3003, opened 2026-08-26
Prepare release 7.0.1

### LINKED ISSUE #2985: release 7.0.1
Should do a bug fix release for #2978 and #2981

jku:
Everything looks like a bug fix, bumping patch number only.
* Fix hash method on Role and DelegatedRole (#2973)
* Fix incorrect dependency range for urllib3 (#2990)
* ngclient: Fix connection exception handling (#2964, #2993)
* ngclient: Fix incorrect delegation in diamond-shape case (#2978)
* ngclient: Avoid use of symlink on Windows: it requires privileges (#2981)

Fixes #2985 

The hash-ability of our classes is still questionable (see discussion in #2995) but I don't think it's critical as they've never been hashable before.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+## v7.0.1
+
+### Fixed
+
+* Fix hash method on Role and DelegatedRole (#2973)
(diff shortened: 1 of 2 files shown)


lukpueh (approved):
looks good. seems okay to not mention test and docs improvements.

PR #3009, opened 2026-09-01
Fix is_expired docstring for reference_time

Sachith77:
The `Signed.is_expired()` docstring says that `reference_time` should be a naive datetime, but passing a naive datetime raises:

 snapshot.is_expired(datetime(2026, 1, 1))
TypeError: can't compare offset-naive and offset-aware datetimes

`self.expires` is timezone-aware, and the default `datetime.now(timezone.utc)` is aware as well, so `reference_time` needs to be timezone-aware too.

This PR only updates the docstring and does not change the behaviour.


jku (approved):
looks correct, thanks

PR #3010, opened 2026-09-03
ngclient: Update root.json link atomically

Sachith77:
`_update_root_symlink()` deletes `root.json` before creating the new link, so if the process dies in that window or the link can't be created, `root.json` just stays gone:

```text
after failed update, root.json exists: False
bootstrap=None now fails: FileNotFoundError
```

Since that's the trust anchor for `Updater(bootstrap=None)`, the client can't start after that. It also runs from a `finally` block in `_load_root()`, so this can happen on failure paths too.

Fixed this by creating the link under a temp name and using `os.replace()` to put it in place, same idea as `_persist_file()`. I used a fixed `.tmp` name instead of `tempfile` since `os.symlink()` needs a path that doesn't exist yet, and multiple Updater instances aren't supported anyway.

Added a test that fails on develop.

### DIFF
--- a/tests/test_updater_ng.py
+++ b/tests/test_updater_ng.py
@@ -385,6 +385,31 @@ class TestUpdater(unittest.TestCase):
         with open(linkname, "rb") as f1, open(target, "rb") as f2:
             self.assertEqual(f1.read(), f2.read())
 
+    @patch("os.link")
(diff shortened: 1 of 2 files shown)


Sachith77:
@jku could you please have a look when you get a chance .

Issue #1562, opened 2021-09-02
ngclient: prevent using rolenames as filenames if they are dangerous


MVrachev:
**Description of issue or feature request**:
Delegation role names are not restricted in any way in the spec, but they are targets metadata role names.
This means they can be any random string. In Sigstore, for example, they are thinking of using delegations role names as URLs to avoid collisions (see https://github.com/theupdateframework/python-tuf/issues/1527#issuecomment-903871213).
We don't want to limit that behavior.

Still, using those role names as filenames is problematic (for both repository and client) as:
- not all strings are actually valid filenames - but the spec at times requires using the role name as a file name. Sanitizing the filename is difficult (as collisions must be prevented)
- Using those role names as filenames is also problematic because it can lead to path traversal with roles like `../../../filename`. This is especially an issue for the client as role name is input from remote that should not be trusted more than is required 
- There are no known security issues with arbitrary role names as files and version numbers, but the area seems ripe for them: is `1.role.json` the metadata for the role `role` or the role `1.role`?

**Current behavior**:
Currently, we are using role names as file names without any checks.

**Expected behavior**:
A possible solution could be validating that delegated role name:
- doesn't contain `.` and `/` in its name
- doesn't start with a number that can be misinterpreted as a metadata file name when consistent_snapshot is used.
See: https://theupdateframework.github.io/specification/latest/#consistent-snapshots

I think the place where this validation should happen is before calling this `Updater._download_metadata()` here:
https://github.com/theupdateframework/python-tuf/blob/e6b41d582a338a1a17ea59f459453f0c71e3c63a/tuf/ngclient/updater.py#L372

For context read: https://github.com/theupdateframework/python-tuf/issues/152

_This issue description is based on @jku comment here:_
https://github.com/theupdateframework/python-tuf/issues/1527#issuecomment-905282226

jku:
Looks good, just wanted to add that we would like to support arbitrary rolenames in ngclient but we just don't have a solution right now that is both:
* clearly spec compliant and
* safe 

So we're trying to compromise here: keep using rolenames as filenames, but error out if those rolenames look like they might be unsafe as filenames. I think we should also at some point prototype with not using rolenames as filenames to show it works but this seems to be in conflict with the specification...

> I think the place where this validation should happen is before calling this Updater._download_metadata() here:

Yeah that will work. The limitations are:
* must happen before the file writing in `_persist_metadata()`
[...]

Issue #1527, opened 2021-08-20
Metadata API: Delegation role names validation


MVrachev:
**Description of issue or feature request**:
Delegation role names are not restricted in any way in the spec, but they are targets metadata role names.
They could be `"."`, `"../../filename"` or `1.role`.
The problem is that at some point those delegation role names are used when constructing an URL used
to download the delegated target metadata file:
https://github.com/theupdateframework/tuf/blob/e9106b59cdb5bbfb4260c5ffc3144e79f8f9596a/tuf/ngclient/updater.py#L287 which is likely to be a problem.

**Current behavior**:
No validation is used for Delegation role names.

**Expected behavior**:
Escape special symbols like `.` or `\`.

trishankatdatadog:
Cc @asraa @mnm678 @joshuagl

jku:
I don't know of specific problems with using these as urls (although those might exist as well), but the obvious problems are in client using the rolename as part of a local file name (this is not just how the reference client is implemented, the spec seems to require doing this): in this client case role names like  `1.root` or `snapshot` or `../../filename` could easily be pretty bad if we assume an attacker having the ability to create new roles and delegations in the server and then client writing files with filenames fromed from these role names.

I don't think think escaping is enough: we probably should not allow rolenames with any of `['.', '/', '\0']` in them and also should not allow `root/snapshot/timestamp/targets` as delegated targets role names.

asraa:
> I don't think think escaping is enough: we probably should not allow rolenames with any of ['.', '/', '\0'] in them and also should not allow root/snapshot/timestamp/targets as delegated targets role names.

Ah, yeah, by and large I agree. In the case of Sigstore delegations to projects, e.g. `github.com/kubernetes/kubernetes`, we were thinking of using the full URL (to avoid e.g. github vs gitlab collisions) as the project name, which would obviously cause problems. I'm really not sure what the right answer is here, if we should sanitize it in the application uniformly by removing the `/`. I definitely think there should be rolename validation.

mnm678:
Maybe this should be part of the filesystem API? That way `.` and `/`' could be disallowed for unix-like filesystems, but could be allowed, with restrictions for sigstore delegations to git repo urls.

Although, these role names are in signed metadata, so it might be reasonable to just trust the delegator to know what a reasonable role name is for a particular situation.

jku:
[...]

Issue #456, opened 2017-06-02
Should the specification restrict the format of rolenames listed in metadata?


vladimir-v-diaz:
The specification does not set any restrictions on the rolenames listed in metadata.  For example, these rolenames may include unicode and other special characters.  We can leave it up to adopters to set these restrictions for their particular implementation, or we can require a set of restrictions by default for cases that we know can lead to problems.

JustinCappos:
This is tough.  Any implementation should be careful about what it permits,
but I don't know that it is our role to force them to not use certain
characters.

Is having best practices around which characters are allowed a good way to
handle this?

jku:
I'll document the current situation in python-tuf and then close this: I agree that the specification could be a lot better in this regard but that's not a python-tuf issue.

python-tuf handles rolenames as follows:
* any string is an acceptable rolename
* rolenames are not expected to be valid parts of a file path: e.g. the client percent encodes the rolename before using it as a filename
* as a clarification to those two points, the "ROLENAME.json" key in targets delegations is still expected to be the raw string with an extension added to the end (I consider this a specification bug, the key should be just "ROLENAME"). This means the key does not necessarily match the filename that is actually used to store the metadata.

PR #3011, opened 2026-09-04
Reject delegated role names that match a top-level role in any case

arpitjain099:
`Delegations.__init__` rejects a delegated role named after a top-level role, but the comparison is exact and `TOP_LEVEL_ROLE_NAMES` is all lowercase. A role named `Root` passes.

That matters because the guard is what keeps a delegated role's cache filename from colliding with a top-level one. `Updater._persist_metadata` builds the filename as `parse.quote(rolename, "")` plus `.json`, and quoting with `safe=""` leaves ASCII letters untouched, so `Root` stays `Root` and `Root.json` is the same directory entry as `root.json` on APFS or NTFS. The issue discussions that settled on percent-encoding (#456, #1527, #1562) reasoned about it as an injective mapping, which it is, but injectivity into filenames also needs the filesystem to distinguish the results.

The effect on a client is that after `get_targetinfo()` walks a delegation to such a role, the `root.json` symlink has been replaced by a regular file holding targets metadata, and a subsequent `Updater(..., bootstrap=None)` raises `RepositoryError: Expected 'root', got 'targets'`. That is the trust-on-first-use mode `examples/client/client` uses. It survives a restart, since nothing reads `root_history/` without a working `root.json`. `Timestamp` and `Snapshot` collide too but recover on the next run, because `_load_timestamp` and `_load_snapshot` refetch on a load failure.

I reproduced it with the repository simulator, delegating from top-level targets to a role named `Root`, and again through a two-level chain. Availability only: `update_delegated_targets` goes through `_load_from_metadata(Targets, ...)`, which rejects anything whose `_type` is not `targets`, so the bytes that land there can never be root metadata.

The change is one `.lower()` in the existing guard, plus a case-variant entry in the invalid-delegations table in `test_metadata_serialization.py`. That test fails without the fix. Full suite is 205 tests, passing on 3.12.

Fixing it at the filename layer instead would also work, and if you would rather have `_persist_metadata` refuse a name whose quoted form collides with a top-level file, I am happy to redo it that way. I went with the validation layer because it keeps the rule in the one place that already states it.


jku (changes requested):
please check the patch contents

